PR #7, opened 2013-12-10
Accelerate lookups

aronsoyol:


davelab6:
@behdad can this be closed?

behdad:
Still want to reevaluate at some other time.

timofonic:
@behdad Have you evaluated it?

behdad:
Closing this.  If we ever decide to take that approach again, should be rewritten.

PR #11, opened 2014-01-08
Some attempt into initial hb-shape documentation

khaledhosny:
Very anaemic, needs to descripe the format understood by
hb_feature_from_string() etc., but it is just start.


behdad:
Merged this manually.  Thanks!

PR #12, opened 2014-01-08
Install hb-unicode-* tools

luisbg:
As discussed in this mailing thread: http://lists.freedesktop.org/archives/harfbuzz/2014-January/004000.html


davelab6:

behdad:
Still want to merge some time.

khaledhosny:
Do we still want to do this now that `hb-shape`/`hb-view` has a `--unicodes` option? Actually do we still use these?

behdad:
I use decode|prettyname for quickly seeing what a sequence is when debugging Indic / East Asian stuff.  But yeah, these are not as useful anymore.

behdad:
I have wanted to implement the diff'ing, including colors in hb-view / hb-shape themselves.

behdad:
Closing this as I cannot make up my mind about it. Just leaving things as is for now.

luisbg:
With this the two meaningful PRs I had in Harfbuzz never merged. Sad because I was looking to contribute back to this awesome project. I understand that since I submitted this to now the need for it has changed.

Let me know if there is anything else I can help with.

PR #14, opened 2014-01-20
Provide hb-compare-shapers tool

jfkthame:
These are the patches I'm using to generate the uniscribe-vs-harfbuzz comparisons - would be nice to integrate this to master sometime.


davelab6:

behdad:
Still want to merge some time.

behdad:
This is outdated and too much work to revive. The buffer-diff API made it into the library long time ago.

PR #41, opened 2014-05-17
Cmake port

pkoshevoy:
Enable native win32 compilation via CMake script
- this is based on qmake project file harfbuzz-ng.pro from Qt
- Ragel is required for pre-processing .rl files
- this is an initial attempt, further improvement may be attempted

### DIFF
--- a/msvcport/CMakeLists.txt
+++ b/msvcport/CMakeLists.txt
@@ -8,6 +8,27 @@ if (NOT WIN32 OR (MINGW OR MSYS OR CYGWIN))
     "For any other platforms please use the autotools ./configure script.")
 endif (NOT WIN32 OR (MINGW OR MSYS OR CYGWIN))
 
+set(FREETYPE_DIR "$ENV{FREETYPE_DIR}"
+  CACHE PATH "root path for freetype lib/ and include/ folders"
+  )
+find_path(FREETYPE_INCLUDE_DIR
+  ft2build.h freetype2/freetype/freetype.h
+  PATHS ${FREETYPE_DIR}/include
+  )
+find_library(FREETYPE_LIBRARY
+  freetype libfreetype
+  PATHS ${FREETYPE_DIR}/lib
+  DOC "freetype library"
+  )
+if (FREETYPE_INCLUDE_DIR)
+  include_directories(AFTER ${FREETYPE_INCLUDE_DIR} ${FREETYPE_INCLUDE_DIR}/freetype2)
+endif (FREETYPE_INCLUDE_DIR)
+
+if (FREETYPE_INCLUDE_DIR AND FREETYPE_LIBRARY)
+  set (THIRD_PARTY_LIBS ${THIRD_PARTY_LIBS} ${FREETYPE_LIBRARY})
+  add_definitions(-DHAVE_FREETYPE=1 -DHAVE_FT_FACE_GETCHARVARIANTINDEX=1)
+endif (FREETYPE_INCLUDE_DIR AND FREETYPE_LIBRARY)
+
 
 find_program(RAGEL "ragel")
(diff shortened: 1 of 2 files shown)


ebraminio:
Honestly, the cmake file introduced on this pull request and the one I am [maintaining](https://github.com/ebraminio/glcourse/blob/master/harfbuzz.cmake) helped me a lot on extending harfbuzz itself as on Windows I was able to access DirectX headers with this cmake file and also I could use Xcode profiler while native project file provided by cmake. Considering usefulness of cmake for myself at least, I thought of a lot of different ways on progressing this but everytime my plan had some hack inside and I didn't came with an interesting sharing list of source files to build. 

If harfbuzz had cmake support, it wouldn't need nmake support (/win32 folder) and now that nmake seems is unable [at least yet] to provide static builds as [vcpkg port files](https://github.com/Microsoft/vcpkg/blob/master/ports/harfbuzz/portfile.cmake#L9) and my little personal tests shows.

Having both autotools nmake is also expensive to maintain and this is in the while that cmake [is getting more love](https://blogs.msdn.microsoft.com/vcblog/2016/10/05/cmake-support-in-visual-studio/) from Microsoft these days. Also Android Studio also is getting better support for cmake native builds). LLVM project, among lots of other [foss projects](https://en.wikipedia.org/wiki/CMake#Applications_that_use_CMake) has switched to cmake as a matter of fact.

So Behdad, I think you should reconsider the possibility of total switch to cmake, probably on a different branch for now. Even the fact I am not fan of cmake but old fashioned autotools based builds with its archaic dependencies like m4 will not be appealing for next generation of developers and this project worth to always have the bests. Even if these are not enough motivation for a total switch, I guess thinking about initial cmake on contrib folder is worthy. Thanks

egorpugin:
Hey, by the way harfbuzz is available at CPPAN (C++ Archive Network) in any config you need - static/shared. Cppan generates cmake projects from simple yaml declarative config.
https://cppan.org/pvt.cppan.demo.behdad.harfbuzz/version/1.3.4
I've tested it mostly on windows at the moment (demo package).

ebraminio:
Thank you Egor for sharing that! Being able to use static build of harfbuzz on Windows was not the point as I can use my own cmake file and I was referring to maintain load of separate nmake project which I couldn't use it for static build with vcpkg (I mean, my intention was to fix vcpkg to make hb static build which at least I was unable to). cppan, however, seems like a cool project which I didn't know about so thanks for letting me know about it. As my involvements with harfbuzz was not about embedding it into my app but extending itself on the aspects I am interested in I still like to see cmake project inside harfbuzz project and being able to just get harfbuzz on my project is not very interesting on this use case but maybe will be later. Thanks :)

behdad:
@ebraminio feel free to push whatever you want into win32/.  As long as you or someone else maintains it I don't care.

Bonus points if you can make it pull the source file list from Makefile.sources instead of duplicating it.

behdad:
@ebraminio what's the status of this?

ebraminio:
Interestingly I wanted to ask yesterday if a better naming also instead "win32/" possible?

behdad:
build? Maybe win32 can move in there as well. Or if it's just one or two files, root dir is fine.

egorpugin:
Some projects use `build(s)` dir and throw all possible build files there, also into subdirs if applicable (e.g. `build/cmake`).

ebraminio:
I went with root dir on #444 as seems you can agree on it and freetype does that also, because I like to make harfbuzz development itself easier (at least for myself) rather than just providing cmake also on tar balls.

ebraminio:
harfbuzz finally got its CMake support on #444 which actually was based on this great effort, hurray! :)

PR #68, opened 2014-12-08
Fix VC++ build warnings

ThePhD:
This PR is a simple PR to fix errors with VC++'s compiler an warnings about conversions from integers to booleans. Otherwise, the build is errorless.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -99,8 +99,7 @@ LOCAL_SHARED_LIBRARIES := \
 	libutils \
 	liblog
 LOCAL_C_INCLUDES += \
-        $(LOCAL_PATH)/src \
-        external/icu/icu4c/source/common
+        $(LOCAL_PATH)/src
 LOCAL_CFLAGS += -DHB_NO_MT -DHAVE_OT -DHAVE_ICU -DHAVE_ICU_BUILTIN
 LOCAL_MODULE:= libharfbuzz_ng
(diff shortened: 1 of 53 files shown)


ThePhD:
Hey! I saw you fix up the warnings. I'm going to run another build, and if things are okay I hope you don't mind me closing this pull request!

ThePhD:
So the only leftover error is C4800 for VC++, which is the `hb_bool_t: forcing value to bool 'true' or 'false' (performance warning)` nonsense.

The best way to fix those is to just compare against 0, I think (apparently the `!!` idiom doesn't make the warning go away).

Here's the build log with the affected files: https://gist.github.com/ThePhD/14c210388fb3e85cea6b .

behdad:
Thanks.  I'll take a look.

ThePhD:
I've updated the `master` branch on my PR. There was an additional hard error that came from harfbuzz recently:
`harfbuzz\src\hb-common.cc(222): error C4996: 'strdup': The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name: _strdup. See online help for details.`

It looks like VC++ supports _strdup, so I added it to this PR's commits. I'm fairly certain g++ and clang++ both have _strdup available...

Finally, I have to close this issue because of some branch and history fixes I had to make to work on the most up-to-date branch rather than my outdated one, so a new Pull Request was opened at #114: sorry if it's confusing!

PR #75, opened 2015-01-04
Fix hb_atomic_ptr_cmpexch -Wunused-value warnings

cpeterso:
When building Firefox's in-tree copy of harfbuzz, the following clang warning is reported for the (nonatomic!) implementation of `hb_atomic_ptr_cmpexch()` used in the Firefox OS simulator builds for OS X:

```
gfx/harfbuzz/src/hb-common.cc:348:5: warning: expression result unused [-Wunused-value]
```


behdad:
Thanks.

PR #76, opened 2015-01-05
Add #ifdef HB_USE_ATEXIT to fix -Wunused-function warnings

cpeterso:
Commit e9171af55cc6a402eb20db4ea74c86a0b1e70e85 marked some functions as static inline to suppress -Wunused-function warnings, but this inline trick doesn't always suppress the warnings. Plus, inline doesn't make sense for function pointers passed to `atexit()`. Instead, this PR wraps these functions in `#ifdef HB_USE_ATEXIT` because they are only called by `atexit()` `#ifdef HB_USE_ATEXIT`.

This change will fix some warnings when building Firefox's in-tree copy of Harfbuzz:
```
gfx/harfbuzz/src/hb-shape.cc:283:6 [-Wunused-function] unused function 'free_static_shaper_list'
gfx/harfbuzz/src/hb-shaper.cc:44:6 [-Wunused-function] unused function 'free_static_shapers'
```

### DIFF
--- a/src/hb-common.cc
+++ b/src/hb-common.cc
@@ -235,7 +235,7 @@ struct hb_language_item_t {
 static hb_language_item_t *langs;
 
 #ifdef HB_USE_ATEXIT
-static inline
+static
 void free_langs (void)
 {
   while (langs) {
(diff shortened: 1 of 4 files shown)


behdad:
Thanks.

PR #82, opened 2015-01-27
Do not leak hb_language_t on hb_language_item_t destruction

KonstantinRitt:


behdad:
Merged manually.  Thanks.

PR #86, opened 2015-02-14
Improve hb-shape/hb-view's help text w.r.t. output options 

cpfair:
Maybe the format and options are self-evident if you know enough about typography - but I certainly didn't when I got started.

New `--help-output`:

	Output destination & format options:
	  --output-file=filename                Set output file-name (default: stdout)
	  --output-format=format                Set output format

	    Supported output formats are: ansi/png/svg/pdf/ps/eps

New `--help-output-syntax` (renamed from `--help-format` since otherwise we have too many definitions of the word "format"):

    Output syntax:
	    text: [<glyph name or index>=<glyph cluster index within input>@<horizontal displacement>,<vertical displacement>+<horizontal advance>,<vertical advance>|...]
	    json: [{"g": <glyph name or index>, "ax": <horizontal advance>, "ay": <vertical advance>, "dx": <horizontal displacement>, "dy": <vertical displacement>, "cl": <glyph cluster index within input>}, ...]

	Output syntax options:
	  --show-text                           Prefix each line of output with its corresponding input text
	  --show-unicode                        Prefix each line of output with its corresponding input codepoint(s)
	  --show-line-num                       Prefix each line of output with its corresponding input line number
	  --verbose                             Prefix each line of output with all of the above
	  --no-glyph-names                      Output glyph indices instead of names
	  --no-positions                        Do not output glyph positions
	  --no-clusters                         Do not output cluster indices

It'd be nice to merge these two help sections, but that looks like it's quite involved - would be good to do at the same time as https://github.com/behdad/harfbuzz/issues/78.

### DIFF
--- a/util/options.cc
+++ b/util/options.cc
@@ -459,7 +459,7 @@ output_options_t::add_options (option_parser_t *parser)
   };
   parser->add_group (entries,
 		     "output",
-		     "Output detination & format options (refer to --help-output-content for content options):",
+		     "Output detination & format options:",
 		     "Options controlling the destination and form of the output",
 		     this);
 }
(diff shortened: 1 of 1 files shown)


behdad:
Thanks.

PR #88, opened 2015-02-26
build: Make it clearer which packages are needed, also check for libtool...

srl295:
* Clarify which packages are needed
* check for libtool (the error message was a little obscure)


srl295:
also cleanup `.gitignore`

behdad:
Thanks.  Pushed out similar fixes.

PR #90, opened 2015-03-22
[bindings] Fix *_from_string functions

khaledhosny:
Without the element-type they will be getting garbage, at least with
Python.

### DIFF
--- a/src/hb-common.cc
+++ b/src/hb-common.cc
@@ -57,7 +57,7 @@ _hb_options_init (void)
 
 /**
  * hb_tag_from_string:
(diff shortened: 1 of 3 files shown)


behdad:
I still can't get this to work for me.  Which versions everything are you using?

khaledhosny:
I’m using Python 2.7, PyGObject 3.14.0, GObject Introspection 1.42.0 and GLib 2.42.2.

behdad:
I tested this again.  They work except for the language one, which is issue 91.

PR #93, opened 2015-03-29
Make hb_test_tools.py compatible with python 3

ebraminio:
On ArchLinux, /usr/bin/python is linked to python 3 so HarfBuzz `make check` is broken there.

This makes hb_test_tools.py compatible with python 3 while no breaking it on python 2.

Of course this is not all good, the change on Unicode.encode is hack and probably needs figuring out what it should be done with it.

### DIFF
--- a/test/shaping/hb_test_tools.py
+++ b/test/shaping/hb_test_tools.py
@@ -1,10 +1,14 @@
 #!/usr/bin/python
 
+from __future__ import print_function
 import sys, os, re, difflib, unicodedata, errno, cgi
 from itertools import *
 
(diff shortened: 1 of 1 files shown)


ebraminio:
I see another approach for this, using /usr/bin/python2 on python scripts shebangs but I am not sure which would be better.

behdad:
I like this.  I prefer the other way around, for chr.  Something like:

```
try:
	unichr
except:
	unichr = chr
```

ebraminio:
Thank you. Fixed. Now it become even more simiar to shim used on other projects so I hope it now look good.

behdad:
Thanks.

PR #96, opened 2015-04-08
Coretext pdf

KonstantinRitt:


behdad:
I saw this before in Qt repo.  This is not well motivated.  CoreText backend doesn't return a glyph entry for any PDF at any location.  Why special case the one at the end of string?  Why just the PDF, and not any other default-ignorable?

To me, this suggests that there's a problem in your client code that make assumptions that are not valid.  For example, if the run text ends in a ligature (eg for Arabic, text ends in لا), then the last character of the text doesn't have any corresponding glyph either.  Does your client code handle that correctly?

KonstantinRitt:
I'll ask Eskil to test this. Thanks.

behdad:
Thanks.

KonstantinRitt:
Ok, tomorrow is a new day... and my fresh-ed head recall what the issue was about.
> Why special case the one at the end of string?
PDF terminates the bidi item, so it will be the last character of the item. We never shape non-itemized text. Without this patch, the cluster of PDF recieves value of the last glyph's cluster.
> Why just the PDF, and not any other default-ignorable?
I think PDF was the only case catched so far.
> For example, if the run text ends in a ligature (eg for Arabic, text ends in لا), then the last character of the text doesn't have any corresponding glyph either.
The combining marks are not the case as they normally recieve continue the cluster of the base character (starter).

behdad:
@KonstantinRitt Does this mean you are all happy about current state of hb-coretext?  I hope that is the case indeed :).

KonstantinRitt:
According to Eskil's last report the issue is gone. With my fingers crossed, throwing this hack from Qt's local copy and closing this MR. Thanks for your work!

PR #97, opened 2015-04-08
[CoreText] Minor optimization

KonstantinRitt:
Squash two for-loops for chars and log_clusters into a single one.


davelab6:

behdad:
Too risky (needs to be tested) with little reward.  :)

PR #98, opened 2015-04-10
Fix build on WEC2013

KonstantinRitt:
Based on patch from Björn Breitmeyer


behdad:
Why was it breaking the build though?

KonstantinRitt:
I think it was redefinition for setlocale

PR #99, opened 2015-04-10
More ntrospection fixes

khaledhosny:
With those small fixes Python bindings are mostly usable.

### DIFF
--- a/src/hb-buffer.cc
+++ b/src/hb-buffer.cc
@@ -967,7 +967,7 @@ hb_buffer_set_language (hb_buffer_t   *buffer,
  *
  * 
  *
- * Return value: 
+ * Return value: (transfer none):
  *
  * Since: 1.0
  **/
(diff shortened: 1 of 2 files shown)


behdad:
Note that hb-language issue is caused by the way I defined it:

```
/* Currently gobject-introspection doesn't understand that hb_language_t 
 * can be passed by-value.  As such we box it up.  May remove in the 
 * future. 
 * 
 *   https://bugzilla.gnome.org/show_bug.cgi?id=707656 
 */ 
GType hb_gobject_language_get_type (void); 
#define HB_GOBJECT_TYPE_LANGUAGE (hb_gobject_language_get_type ()) 
```

khaledhosny:
I guess so, but I’m unable to wrap my head around the hb_language_t code.

behdad:
Thanks.

PR #100, opened 2015-04-23
Don't apply Arabic shaping to vertical text.

jfkthame:


behdad:
I'd be much more comfortable adding a fix here:
https://github.com/behdad/harfbuzz/blob/5eb939ddfe9ef217da2e48d7d0f1f4b7501714ad/src/hb-ot-shape-complex-private.hh#L184

jfkthame:
Sounds fine, I've modified the PR accordingly.

behdad:
Merged manually.  Thanks.

PR #102, opened 2015-04-29
Add OpenType language tag 'NAV ' for Navajo.

roozbehp:
The code is not standardized yet, but is used in some Google fonts.


behdad:
Is it proposed at least?

roozbehp:
No, I would need help with that. It's on our Monday meeting's agenda.

PR #107, opened 2015-05-18
Graphite2 dynamic loading

mhosken:
A number of people have asked me for a mechanism by which graphite may be dynamically loaded only when a Graphite font is used. I've struggled with the notion of this, but I think I understand it now. I hope that this can help everyone to have what they want for minimal cost.

This patch adds dynamic loading of graphite support for graphite fonts in harfbuzz. The three way configure option is now: --with-graphite2=no means no graphite support. --with-graphite2=yes means to build and link against an existing graphite library. --with-graphite2=auto means to build independently of any graphite library but to attempt to dynamically load graphite when a graphite font is encountered.

This patch has been built and tested on linux only at the moment.


mhosken:
Code is there for windows & Mac, but not yet tested on those platforms. I'm willing to work with someone wanting to do such a test.

mhosken:
In response to a flagging of a memory leak around grfuncs, I have removed the alloc and now hold the funcs data structure as part of shaper_face_data, thus removing the leak.

behdad:
Humm. I don't think this is going to happen.  If we ever want to do it, it should be done differently.  See how hb-uniscribe is done for example.  Ie. dl handle should be a global var. face and font data sould be the same as static build. There should be one ifdef only!

PR #110, opened 2015-05-20
Fix unary minus operator applied to unsigned int

kojiishi:
Applying unary minus operator to unsigned int causes the following warning on MSVS:
  warning C4146: unary minus operator applied to unsigned type, result still unsigned

This patch fixes the warning.


behdad:
Thanks.  Fixed.

PR #111, opened 2015-06-01
Fix "Since:" tags

### LINKED ISSUE #103: Put correct "Since:" tag in all API
Right now we have "1.0" for everything.  Put true 0.9.x version they were introduced instead...

brawer:
Based on data from http://upstream-tracker.org/versions/harfbuzz.html
Resolves #103


behdad:
Thanks Sascha.

PR #114, opened 2015-06-23
Fix all VC++ warnings and errors

ThePhD:
This pull request fixes all the errors and warnings that come from VC++'s build (VS 2015 RC, VS 2013). This was originally in PR #68, but github's strange branching mechanism doesn't allow for me to simply change the base of the pull request I'm making.


ThePhD:
This PR has been updated to match the latest as of August 14th, and contains the last comparison-value fix for `hb_bool_t` to `bool`. I believe that's everything!

behdad:
Is this still valid?  Can you try with master and point out the outstanding issues?

behdad:
Thanks.

ThePhD:
Hey @behdad, I went and pulled the latest and did a build. There are no reported errors using this branch of fixes, so I think that's it.

PR #116, opened 2015-06-28
Handle language tags that indicate phonetic IPA transcription

brawer:
The BCP-47 registry defines a variant subtag "fonipa" that can be used
in combination with arbitrary other language tags. For example,
"rm-CH-fonipa-sursilv" indicates the Sursilvan dialect of Romansh
as used in Switzerland, transcribed used the International Phonetic
Alphabet.

http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

### DIFF
--- a/src/hb-ot-tag.cc
+++ b/src/hb-ot-tag.cc
@@ -733,7 +733,6 @@ static const LangTag ot_languages[] = {
 /*{"fuf?",	HB_TAG('F','T','A',' ')},*/	/* Futa */
 /*{"ar-Syrc?",	HB_TAG('G','A','R',' ')},*/	/* Garshuni */
 /*{"cfm/rnl?",	HB_TAG('H','A','L',' ')},*/	/* Halam */
-/*{"fonipa",	HB_TAG('I','P','P','H')},*/	/* Phonetic transcription—IPA conventions */
 /*{"ga-Latg?/Latg?",	HB_TAG('I','R','T',' ')},*/	/* Irish Traditional */
 /*{"krc",	HB_TAG('K','A','R',' ')},*/	/* Karachay */
 /*{"alw?/ktb?",	HB_TAG('K','E','B',' ')},*/	/* Kebena */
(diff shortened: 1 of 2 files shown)


behdad:
How did you come up with the OpenType language tag for this?
Does this imply a Latin-IPA writing?

We need better BCP 47 handling.  This is a good start.  There's also a lot of research that James Clark did around the script/language handling in HarfBuzz.  None of us have got to fully study those yet.  Lets start talking about that.  I can't find the links right now.  Will find later.

brawer:
> How did you come up with the OpenType language tag for this?

It was introduced in the [OpenType specification version 1.6](
http://www.microsoft.com/typography/otspec160/languagetags.htm).

> Does this imply a Latin-IPA writing?

Yes, IPA is always written in Latin characters. For example, [here](https://github.com/brawer/playground/blob/master/cldr/test-si-si_FONIPA.txt) is a file that shows `si` and `si-fonipa` next to each other. Regarding the name of that file: it will eventually move into  [Unicode CLDR](http://www.unicode.org/repos/cldr/trunk/tools/cldr-unittest/src/org/unicode/cldr/unittest/data/transformtest/cs_FONIPA-ko.txt). For historic reasons, CLDR has variant subtags in uppercase; also, its separates subtags by underscores instead of hyphens. `si_FONIPA` is still a valid BCP 47 identifier, but the canonical form (which is used, for example, in W3C’s [intro to language tags](http://www.w3.org/International/articles/language-tags/)) is `si-fonipa`.

> We need better BCP 47 handling

Yes, from glancing over the code, it appears there might be a few issues. For the minor smells, I’ll write test cases and send fixes if needed; but it would be good to chat in person. Also, a few language tags in the [OpenType spec](http://www.microsoft.com/typography/otspec170/languagetags.htm) seem surprising; I’ll look into this with Mark Davis.

> There's also a lot of research that James Clark did around the script/language handling in HarfBuzz

Any pointers beyond [this e-mail](http://lists.freedesktop.org/archives/harfbuzz/2014-March/004183.html)? I’d be super interested.

brawer:
OK to merge? Agree that the handling of language tags should be overhauled for properly handling BCP47; in the meantime, this patch would be an improvement for *-fonipa. (I'm not saying it's super important, just noticed that this pull request has been sticking around for a while. I'm also totally fine abandoning it).

PR #119, opened 2015-07-08
Unicode 8.0 and UCDN update

grigorig:
As requested.


behdad:
Merged manually.  Thanks!

PR #123, opened 2015-08-07
Don't rely on .cluster field in fallback_position

jfkthame:
See https://bugzilla.mozilla.org/show_bug.cgi?id=890156


behdad:
Looks good.  Please merge.  Can you add a micro test for this please?

behdad:
Added test and merged this manually.  Will do the other later.

PR #129, opened 2015-08-25
First two chapters. More to follow.

simoncozens:


behdad:
Fantastic.  Thanks Simon!

behdad:
How do we build these?  How do you edit these?  Can we move into a subdir of docs/ please?

PR #134, opened 2015-09-11
Add working but less prefect DirectWrite backend

ebraminio:
More description on this mailing list [post](http://lists.freedesktop.org/archives/harfbuzz/2015-September/005062.html).

### DIFF
--- a/src/hb-directwrite.cc
+++ b/src/hb-directwrite.cc
@@ -1,0 +1,827 @@
+/*
+ * Copyright © 2015  Ebrahim Byagowi
+ *
+ *  This is part of HarfBuzz, a text shaping library.
+ *
+ * Permission is hereby granted, without written agreement and without
+ * license or royalty fees, to use, copy, modify, and distribute this
+ * software and its documentation for any purpose, provided that the
+ * above copyright notice and the following two paragraphs appear in
+ * all copies of this software.
+ *
+ * IN NO EVENT SHALL THE COPYRIGHT HOLDER BE LIABLE TO ANY PARTY FOR
+ * DIRECT, INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES
+ * ARISING OUT OF THE USE OF THIS SOFTWARE AND ITS DOCUMENTATION, EVEN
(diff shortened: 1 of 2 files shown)


ebraminio:
Squashed.

Behdad, this is far from prefect but may I suggest putting this on some branch?

On anycase, I am open for your code reviews, at least for reaching to a point suitable for having it on a branch, so please have a look at.

KrasnayaPloshchad:
Since M$ making DirectWrite more competitiveness with Uniscribe, and more apps tend to use DirectWrite, the developments on `hb-directwrite` is necessary.

khaledhosny:
Since Uniscribe seems to be just a backward compatibility wrapper around DirectWrite these days, do we gain much be using DirectWrite directly? After all these backends are just for testing and comparing HarfBuzz output to Uniscribe and nothing else.

KrasnayaPloshchad:
Eh, I got wrong materials about DirectWrite, sorry.

ebraminio:
Studying DW API given for Kashida justification was my only motivation for this which I am not saying is even a good one. I don't think I am the right person for this even as the lack of enough familiarity with Windows APIs and the fact that honestly I couldn't even find any clean solution on using mingw/cygwin autotools with directx API headers and actually I used a cmake build script with vs for this. However I hope this would be useful thing for further attempts at someday.

behdad:
Should add buildsystem support for this.  I'll give it a try when I have my Windows 10 machine around.

ebraminio:
Oh, thank you Behdad. I see this as a great step forward. As its known issues I explained on ML I was expecting see this on a branch moreover but this is even better. I doubt there would be any clean way to use dx headers with autotools system but the WIP NMake support seems is fixing this issue also finally.

One suggestion, to prevent misuse of clients from the dx support for now, I think it would be nice to put some compiler directive (maybe `#warning`?) to indicate this is not for use for now. Perhaps that would be useful till I could get access to a machine or someone more familiar with dx and this particular thing could resolve the issues.

Thank you

behdad:
As long as it's not easy to build with it, we should be fine.  But feel free to send a PR.

fanc999:
Hi,

For references, I have it in a commit in PR #164 integration in the autotools build system, which I managed to do successfully with MSYS2/mingw-w64.

With blessings!

PR #147, opened 2015-10-19
[ci] change to docker infrastructure

ebraminio:


ebraminio:
Wow, fascinating how fast is this thing, good to go IMHO.

PR #158, opened 2015-11-04
Enable Visual Studio Compilation

fanc999:
Hi,

This attempts to add Visual Studio compilation support for HarfBuzz via NMake, which allows for a flexible build of HarfBuzz, anything from a simple HarfBuzz DLL containing support for the Uniscribe, OpenType and Fallback backends, to a full-fledged build that includes all the tools and tests in test/api, along with HarfBuzz-GObject (with or without introspection) and HarfBuzz-ICU.  This is done as there are a number of projects (that are built with Visual Studio) which make use of HarfBuzz(-NG), including QT, Mozilla and LibreOffice and more.

To ease maintenance, the source listings are split out from the Makefile.am's, so that they may be shared with the NMake Makefiles and the autotools builds.

Note: This currently supports building the libraries as DLLs, not static libraries.  Was checked with 'make -j8 distcheck'.

With blessings, thank you!


behdad:
Thanks!  Please clean up the branch, removing the merge commits and the reverts.  They make it hard to see what's going on.  You can also send separate pull-requests for fixes to the utils needed for compiling with MSVC, and another for the build system changes.

fanc999:
Hello Behdad,

OK, I will close this PR and do 2 new PR's.

Thanks, with blessings.

PR #159, opened 2015-11-05
Enable the sources to build with various Visual Studio versions

fanc999:
Hi,

I have made some updates to the sources in util/, and also added an implementation of scalbn() and scalbnf() in a header that can be force-included by pre-2013 Visual Studio, so that the tools can be built under Visual Studio as well.

With blessings, thank you!


fanc999:
Hi,

Hmm, it also seems that some more changes are needed... on Visual Studio 2010 and 2008, things went fine, but on 2012 and later, I am getting C2057 (expected constant expression) errors for:

hb-ot-shape-normalize.cc
hb-ot-shape.cc
hb-ot-shape-fallback.cc

When they include hb-ot-layout-private.hh, the error was at line 67 (in the declaration of enum hb_ot_layout_glyph_props_flags_t)

and the C2057 error for hb-glib.cc when it included gnode.h (at line 43 of gnode.h, when the enum GTraverseFlags was declared) from GLib...

There were bitwise or's in the declaration of the enums, so my guess is some header that is being included from within HarfBuzz is defining a macro that conflicts with some other things in the newer Windows headers from the SDKs or so.

PR #163, opened 2015-11-06
Update the sources so they will compile under Visual Studio

fanc999:
Hi,

Sorry for this, was not that good at git and retried a pull request as a result.

I have made some updates to the sources in util/, and also added an implementation of scalbn() and scalbnf() in a header that can be force-included by pre-2013 Visual Studio, so that the tools can be built under Visual Studio as well.

However, it also seems that some more changes are needed... on Visual Studio 2010 and 2008, things went fine, but on 2012 and later, I am getting C2057 (expected constant expression) errors for:

hb-ot-shape-normalize.cc
hb-ot-shape.cc
hb-ot-shape-fallback.cc

When they include hb-ot-layout-private.hh, the error was at line 67 (in the declaration of enum hb_ot_layout_glyph_props_flags_t)

and the C2057 error for hb-glib.cc when it included gnode.h (at line 43 of gnode.h, when the enum GTraverseFlags was declared) from GLib...

There were bitwise or's in the declaration of the enums, so my guess is that the newer Visual Studio compilers are more picky about the overloading of operators, type-wise, or something conflicted in there, as done in hb-private.hh (circa lines 899 to 900)

With blessings, thank you!

### DIFF
--- a/src/hb-buffer-private.hh
+++ b/src/hb-buffer-private.hh
@@ -48,8 +48,8 @@
 ASSERT_STATIC (sizeof (hb_glyph_info_t) == 20);
 ASSERT_STATIC (sizeof (hb_glyph_info_t) == sizeof (hb_glyph_position_t));
 
-template <> class hb_mark_as_flags_t<hb_buffer_flags_t> {};
-template <> class hb_mark_as_flags_t<hb_buffer_serialize_flags_t> {};
+HB_MARK_AS_FLAG_T (hb_buffer_flags_t);
+HB_MARK_AS_FLAG_T (hb_buffer_serialize_flags_t);
 
 enum hb_buffer_scratch_flags_t {
   HB_BUFFER_SCRATCH_FLAG_DEFAULT			= 0x00000000u,
@@ -64,7 +64,7 @@ enum hb_buffer_scratch_flags_t {
   HB_BUFFER_SCRATCH_FLAG_COMPLEX2			= 0x04000000u,
   HB_BUFFER_SCRATCH_FLAG_COMPLEX3			= 0x08000000u,
 };
-template <> class hb_mark_as_flags_t<hb_buffer_scratch_flags_t> {};
+HB_MARK_AS_FLAG_T (hb_buffer_scratch_flags_t);
 
 
 /*
--- a/src/hb-ot-layout-common-private.hh
+++ b/src/hb-ot-layout-common-private.hh
@@ -581,7 +581,7 @@ struct LookupFlag : USHORT
 
 } /* namespace OT */
 /* This has to be outside the namespace. */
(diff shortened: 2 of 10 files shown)


fanc999:
Hi,

For the hb-ot-layout-private.hh case, I can get away by casting the values used in the bitwise or with unsigned int's, such as the following, but it is clearly not an optimal solution (since, we would have to do the same for this case for GLib's gnode.h):

HB_OT_LAYOUT_GLYPH_PROPS_PRESERVE     =
                                          (unsigned int)HB_OT_LAYOUT_GLYPH_PROPS_SUBSTITUTED |
					  (unsigned int)HB_OT_LAYOUT_GLYPH_PROPS_LIGATED |
					  (unsigned int)HB_OT_LAYOUT_GLYPH_PROPS_MULTIPLIED

So this is the take I have for this...

With blessings, thank you!

fanc999:
Hello Behdad,

The thing is, Visual Studio 2013 can be quite picky on the types used in lround() and scalbnf(), so one would get an ambiguous type error as the compiler insists on a type being specified, between double and float, for instance.

fanc999:
Hi Behdad,

I had your suggested changes in there, but since I rebased to make the history look less cluttered, so probably your in-line comments were taken away by this.  Sorry about this-please let me know if you would rather that I just patch on top of a previous patch rather than rebasing.

With blessings, thank you!

fanc999:
Hi,

This patch will make the enum in hb-ot-layout-private.hh strong-typed on MSVC>=2012, but does not cover the scenario where we include the GLib headers, as any enums in there that use bitwise operators within them will break on MSVC>=2012, due to the same problem.  Unlike G++, I don't think there is any switch in the later Visual Studio versions to disable C++-11 items (or, rather, to enable it optionally), this is how life would go in there.  Perhaps, a configure check may be needed for enabling strong-typing and using the bitwise operator overloading conditionally, but this  probably belonga to another PR.

behdad:
Everything looks good except for the last commit, which I left comments about.  Thanks.

fanc999:
Hi, 


For the GLib/GObject related parts, the build will go fine on the newer MSVC (>=2012) if we disable the bitwise operators overloads, but I'm not sure whether this is the best thing to do...

With blessings, thank you.

fanc999:
Hi,

Finally a solution for the newer MSVC builds...  Re-did the last patch for this.

With blessings, thank you!

behdad:
I still don't understand or like the casts before calling lround, etc.  They make the code un-modifiable, as in, when I write new code, I'll never know when I need to put those casts.  But I'm going to commit this since I don't have more time to send this back to you and review again.  The rest looks good.  Thanks.

PR #165, opened 2015-11-06
Get global font metrics from OS/2 table (fallback to hhea table)

simoncozens:


simoncozens:
Wait! This isn't working!

simoncozens:
I don't understand why not, but my function implementations (`hb_ft_get_font_extents` / `hb_ot_get_font_extents`) aren't being called. It just goes `hb_font_get_extents -> hb_font_get_font_extents_parent -> hb_font_get_font_extents_nil`. It certainly worked before the parent/nil split, but now it looks right to me and I can't find any other bits of code to cargo-cult.

simoncozens:
OK, apologies for the git madness but I think I have something working now.

behdad:
Thanks.  Sorry for the split.  I didn't remember that you had an outstanding patch against those parts of the code.

behdad:
Hey, a release was way overdue, and I had to fix things before I could get a release out.  Just managed to finish those (OOM handling, space shaping, etc), so got a release out, 1.1.0 even.  The next item I have on my list is to integrate this work (after splitting it for horizontal and vertical metrics.)  Thanks.

ebraminio:
Perhaps this patch is worth to look, https://codereview.chromium.org/1450303002/

behdad:
I'm almost done with this.

behdad:
Simon,

Can you try this branch:
https://github.com/behdad/harfbuzz/tree/font-extents

simoncozens:
This works for me with the new linespacing package in SILE. Thank you!

PR #166, opened 2015-11-06
include errno.h and strerror where needed

mostynb:
I notice that errno.h and string.h aren't included consistently when errno and strerror are used, this small patch fixes this.

### DIFF
--- a/src/hb-blob.cc
+++ b/src/hb-blob.cc
@@ -41,6 +41,7 @@
 #endif /* HAVE_SYS_MMAN_H */
 
 #include <stdio.h>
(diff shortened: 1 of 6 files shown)


behdad:
In src/, those are included in hb-private.hh that everything includes.  So those changes are not needed.
In util/, options.hh is the catch-all header, so they are there.

Did you observe any compilation issue at all?  If not, I'd rather not touch it.  Thanks. :)

mostynb:
No, this was just pre-emptive cleanup.  No problem- closing this issue.

PR #167, opened 2015-11-07
Micro optimizations to UTF-16 and UTF-32 codecs

KonstantinRitt:


behdad:
Thanks.  Did you actually measure this?

PR #168, opened 2015-11-07
Update opentype language tags

reutenauer:
Synchronised with OpenType 1.7, and suggested some more changes for “difficult“ tags.

### DIFF
--- a/src/hb-ot-tag.cc
+++ b/src/hb-ot-tag.cc
@@ -699,8 +699,8 @@ static const LangTag ot_languages[] = {
   {"smj",	HB_TAG('L','S','M',' ')},	/* Lule Sami */
   {"smn",	HB_TAG('I','S','M',' ')},	/* Inari Sami */
   {"sms",	HB_TAG('S','K','S',' ')},	/* Skolt Sami */
-  {"sna",	HB_TAG('S','N','A','0')},	/* Shona */
   {"sn",	HB_TAG('S','N','A',' ')},	/* Shona */
+  {"sna",	HB_TAG('S','N','A','0')},	/* Shona */
   {"snk",	HB_TAG('S','N','K',' ')},	/* Soninke */
   {"so",	HB_TAG('S','M','L',' ')},	/* Somali */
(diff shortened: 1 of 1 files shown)


behdad:
Hi there.  Thanks for the PR!  Is there any convention on using '0', or you made that up?  I haven't had seen that.

CC @roozbehp @brawer @jclark @jfkthame can you rely on the four of you to review this?  I'm way out of date with language tags...  Thanks.

reutenauer:
Hi Behdad,

On Sat, Nov 07, 2015 at 05:32:06PM -0800, Behdad Esfahbod wrote:
> Hi there.  Thanks for the PR!  Is there any convention on using '0', or you made that up?  I haven't had seen that.

  I can't say for sure that there is a convention, but they all are
standard OpenType language tags, in the more recent versions at least.
From what I can see, the situation was more or less so: there was first
an OpenType tag ABC for a language whose ISO 639-3 code was xyz (hence
not the lowercase of the OpenType tag), and the language with ISO code
abc didn't have an OpenType tag in that version; one was then added
later, with the value ABC0.  It may very well be the case that some
implementations invented the convention to add a '0' for languages in
that case, in fact it sounds like exactly the kind of thing practically
minded implementers would do, but in any case all the tags are now in
the registry (http://www.microsoft.com/typography/otspec/languagetags.htm).

	Best,

		Arthur

behdad:
Ah, ok.  I didn't notice that these were in the updated MS registry.  Thanks.

KrasnayaPloshchad:
@behdad This page including a link that can be used to compare differences between current and previous specifications
http://www.microsoft.com/en-us/Typography/SpecificationsOverview.aspx

behdad:
Apologies for not getting to this earlier.  Is there a chance you can rebase on master and squash fixup commits together, into a few commits that can easily be verified?  Thanks.

khaledhosny:
@reutenauer are you still working on this, I can open a new pull request and squash the commits together if you want.

reutenauer:
Sure, go ahead if you're happy to do that.

khaledhosny:
OK, opened #247, closing this.

PR #177, opened 2015-11-20
Enable use of compiler directives to export symbols

fanc999:
Hi,

This is a follow-up PR from the discussions in PR #164 to use compiler directives to export symbols using compiler directives, which involves:

-Making sure config.h, if included, is included as the first thing in each source file under src/ (it is already so)
-Decorating the public symbols in the public headers with HB_EXTERN, which is extern by default.
-Add items in the autotools files to use visibility attributes to export symbols upon builds, by overriding HB_EXTERN.  This is similar to what is now done in GLib, GTK+, Clutter, etc.
-(In PR #164) add corresponding #defines for HB_EXTERN in the pre-configured config.h.win32.in to do the same for MSVC builds.

With blessings, thank you!

### DIFF
--- a/src/hb-blob.h
+++ b/src/hb-blob.h
@@ -64,7 +64,7 @@ typedef enum {
 
 typedef struct hb_blob_t hb_blob_t;
 
-hb_blob_t *
+HB_EXTERN hb_blob_t *
 hb_blob_create (const char        *data,
 		unsigned int       length,
 		hb_memory_mode_t   mode,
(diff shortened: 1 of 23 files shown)


behdad:
I like the first patch.

Why do I need the second patch?  It sounds to me that I don't need it.  In that case, I prefer to just take the first patch.

fanc999:
Hello Behdad,

I did the second patch to be in line with the things done in GLib and GTK+, but since it seems that it is preferable to keep the def's check and .def file generation around, I dropped the second patch.

I have updated the patches in PR#164 to reflect the changes here, by defining HB_EXTERN as a /D in the build files.

With blessings, thank you!

PR #178, opened 2015-11-20
hb-private.hh needs to #include Windows.h under MSVC

jfkthame:
This is needed to ensure DEFINE_ENUM_FLAG_OPERATORS works as expected. See https://bugzilla.mozilla.org/show_bug.cgi?id=1226175#c10.


behdad:
I'm *just* pushing out a change, changing that piece of code.

Issue #183, opened 2015-11-24
Build fix for OS X


0-wiz-0:
Adam Ciarciński reported the following build problem on OS X:

````
  CXX      libharfbuzz_la-hb-coretext.lo
In file included from hb-coretext.cc:33:
In file included from ./hb-coretext.h:37:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/System/Library/Frameworks/ApplicationServices.framework/Headers/ApplicationServices.h:23:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/System/Library/Frameworks/CoreServices.framework/Headers/CoreServices.h:23:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/System/Library/Frameworks/CoreServices.framework/Frameworks/AE.framework/Headers/AE.h:20:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/System/Library/Frameworks/CoreServices.framework/Frameworks/CarbonCore.framework/Headers/CarbonCore.h:155:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/System/Library/Frameworks/CoreServices.framework/Frameworks/CarbonCore.framework/Headers/UnicodeUtilities.h:304:35: error: expression is not an integral constant expression
  kUCCollateStandardOptions     = kUCCollateComposeInsensitiveMask | kUCCollateWidthInsensitiveMask
                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/System/Library/Frameworks/CoreServices.framework/Frameworks/CarbonCore.framework/Headers/UnicodeUtilities.h:304:68: note: non-constexpr function 'operator|<(anonymous enum at
[...]

PR #184, opened 2015-11-24
hb-coretext.h: include CoreText.h and CoreGraphics.h

moyogo:
Addresses #183.

### DIFF
--- a/src/hb-coretext.h
+++ b/src/hb-coretext.h
@@ -30,12 +30,8 @@
 #include "hb.h"
 
 #include <TargetConditionals.h>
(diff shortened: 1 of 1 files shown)


moyogo:
@astiob Does this break building on 10.6 per #40?

astiob:
Yes, thanks for asking!

It seems the errors in #183 were caused by the templated `operator |` in hb-private.hh getting picked up in a system header, so they have probably already gotten fixed by f94c0ecbb1488138d3ec3b9d6d9fb78113dfee4f.

behdad:
Just released 1.1.1 that should fix this.

PR #192, opened 2015-12-05
[issue 191] Make apply_stch() give a more precise fit

jfkthame:
This aims to make Syriac Abbr Mark sizing more accurate when repeating segments are used, by adding an extra repeat and tightening up the spacing slightly rather than leaving a shortfall corresponding to a partial repeat-width.


behdad:
Thanks.  When we meet in Dec we should also get some hb-shape uniscribe output from Win10 and compare our heuristic to what Windows uses.

PR #196, opened 2015-12-11
AIX fixes

srl295:
- use '-w' instead of '\<...\>' for check-header-guards
  grep manpage says these are the same

- move order of '-q' option to grep

- hb-font-private.hh - use [0] instead of [] for variable array


srl295:
- rebased.
- removed stray `set -x`

srl295:
@behdad squashed, and moved VAR into another header for great visibility.

behdad:
Thanks!

PR #202, opened 2015-12-28
Doc fixes

khaledhosny:
Some fixes to API extraction.


khaledhosny:
No idea what is causing the coveralls error. I’m starting to like how the API docs look after filling them with text, let me know what do you think.

khaledhosny:
OK, coveralls failure is fixed now.

behdad:
Thanks @khaledhosny!  I'm merging now.  Will review the text later.

PR #204, opened 2015-12-29
Fix what looks like a copy-paste error in c79865f9

nico:
Else, nothing uses the functions defined by
```
HB_SHAPER_DATA_ENSURE_DECLARE(coretext_aat, face)
HB_SHAPER_DATA_ENSURE_DECLARE(coretext_aat, font)
```

### DIFF
--- a/src/hb-coretext.cc
+++ b/src/hb-coretext.cc
@@ -1157,7 +1157,7 @@ _hb_coretext_aat_shaper_face_data_create (hb_face_t *face)
   }
   hb_blob_destroy (mort_blob);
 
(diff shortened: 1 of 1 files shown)


behdad:
Hi Nico,

No, the code is correct.  We are delegating to the coretext shaper.  We can remove the unused HB_SHAPER_DATA_ENSURE_DECLARE lines instead.  I don't have a Mac right now, if it builds fine without them, please let me know or send a PR.  Thanks.

nico:
Thanks for the review! Sent https://github.com/behdad/harfbuzz/pull/206 instead, and verified that Chromium's harfbuzz-ng target still builds on OS X with that change applied.

PR #209, opened 2016-01-02
[docs] Fix gtk-doc waring

khaledhosny:
I see a gazillion of warnings like:
Warning: multiple "IDs" for constraint linkend: hb-font-t.

It seems that typedefs like the one below confuses gtk-doc causing it to
generate several elements with the same id:
typedef struct hb_font_t hb_font_t;

GTK seems to prefix such hidden structs with underscore, not really a
fan of it but the warnings are too many to ignore.

### DIFF
--- a/src/hb-blob.cc
+++ b/src/hb-blob.cc
@@ -50,7 +50,7 @@
 #endif
 
 
-struct hb_blob_t {
+struct hb_blob {
   hb_object_header_t header;
   ASSERT_POD ();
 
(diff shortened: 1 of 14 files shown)


behdad:
Can you please file a bug against gtk-doc at least?  While the struct name is not how people should use these, we can't be sure that no one does...

behdad:
Alternately we can drop the `_t`.  Ie, `struct hb_blob`.

khaledhosny:
Reported here https://bugzilla.gnome.org/show_bug.cgi?id=760152.

Dropped the `_t` from the structs instead if prefixing them, as well.

behdad:
This is obsolete, right?

khaledhosny:
Yes, seems to be solved in recent gtk-doc releases.

PR #213, opened 2016-01-12
Android: fix compile on arm64

bog-dan-ro:
Android-21 seems to quite sensible on defining wrong _POSIX_C_SOURCE.
Check https://code.google.com/p/android/issues/detail?id=194631

### DIFF
--- a/src/hb-blob.cc
+++ b/src/hb-blob.cc
@@ -25,7 +25,7 @@
  */
 
 /* http://www.oracle.com/technetwork/articles/servers-storage-dev/standardheaderfiles-453865.html */
-#ifndef _POSIX_C_SOURCE
+#if !defined(_POSIX_C_SOURCE) && !defined(ANDROID)
 #define _POSIX_C_SOURCE 199309L
 #endif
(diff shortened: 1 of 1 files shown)


behdad:
The change definitely doesn't look right.  We should instead update the requested POSIX_C_SOURCE version to what's right.

The bigger question, however, is, why are you seeing FP_NAN in hb-blob.c.  We don't use it, so I'm assuming that Qt is doing something weird?

At any rate, the Android bug, well, is Android bug.  That their libstdc++ doesn't work with certain modes of their libc is a bug for sure.

behdad:
What is the exact error you are seeing?

bog-dan-ro:
Here https://paste.kde.org/pjohgvypi you can find the full compile command line and also the errors.

behdad:
> Here https://paste.kde.org/pjohgvypi you can find the full compile command line and also the errors.

Ok, that's a Qt / KDE bug.  Report to them and ask them to define correct _POSIX_C_SOURCE.

PR #230, opened 2016-02-25
Visual Studio builds: Support NDEBUG/HB_NDEBUG/HB_DEBUG

fanc999:
Hi,

Since HarfBuzz added support for excluding debug code via defining NDEBUG and HB_NDEBUG, probably it is a good idea to update the Visual Studio build files to support them for release builds, as well as supporting building more debug code into the resulting binaries by defining HB_DEBUG=1.

Note that at this point, Visual Studio 2008 builds cannot cleanly build debug builds with HB_DEBUG=1, so this is disabled for Visual Studio 2008, at least for now.  For debug builds using Visual Studio 2010 and later, HB_DEBUG=1 is always defined.

With blessings, thank you!


behdad:
Thanks.  HB_DEBUG should NEVER be turned on by default.  It will bloat the binary AND by default prints LOTS of stuff to stderr.

I made a lot of fuss about NDEBUG initially, and then added HB_NDEBUG, but then I optimized away the performance issues. So, either rework to drop HB_DEBUG references, or completely abandon this change.

Thanks!

fanc999:
Hi Behdad,

Sorry for not re-visiting this issue soon enough.  I think I will just drop the issue as a result.

With blessings, thank you!

PR #239, opened 2016-03-31
Improve DirectWrite backend

ebraminio:
With these changes, dwrite backend becomes much much useful. However there are some issues that AFAICT are dwrite internal bugs:
* "rlig" and perhaps other default enabled features can not be disabled, but from my tests neither it could be on uniscribe backend nor IE/Edge with font-feature-settings
* "هٔ" (and not other Arabic diacritics) has some rendering issue, which is consistent with Edge but not IE! Windows SDK's CustomLayout dwrite sample however doesn't have this issue so I don't reject the idea that there are some chances that some thing can be improved locally.

Our output:
image
Edge's:
image

Other than these, I don't see any other remained functional issue on this backend, other than features range and sideway/vertical text (which uniscribe backend also lacks that).

### DIFF
--- a/src/hb-directwrite.cc
+++ b/src/hb-directwrite.cc
@@ -689,8 +689,11 @@ _hb_directwrite_shape(hb_shape_plan_t    *shape_plan,
 
   bool backward = HB_DIRECTION_IS_BACKWARD(buffer->props.direction);
 
-  wchar_t lang[4];
-  mbstowcs(lang, hb_language_to_string(buffer->props.language), 4);
+  wchar_t lang[4] = {0};
+  if (buffer->props.language != NULL) {
+    mbstowcs(lang, hb_language_to_string(buffer->props.language), 4);
+  }
+
   hr = analyzer->GetGlyphs(pchars, length,
     fontFace, FALSE,
     buffer->props.direction,
@@ -741,7 +744,7 @@ _hb_directwrite_shape(hb_shape_plan_t    *shape_plan,
 
   int font_size = font->face->get_upem();
   if (font_size < 0)
-	  font_size = -font_size;
+    font_size = -font_size;
   int x_mult = (double)font->x_scale / font_size;
 
   hr = analyzer->GetGlyphPlacements(pchars,
--- a/src/hb-directwrite.cc
+++ b/src/hb-directwrite.cc
@@ -25,7 +25,7 @@
 #define HB_SHAPER directwrite
 #include "hb-shaper-impl-private.hh"
 
-#ifdef HB_DIRECWRITE_PRE_WINDOWS8_COMPATIBLE
+#ifndef HB_DIRECTWRITE_EXPERIMENTAL_JUSTIFICATION
   #include <DWrite.h>
 #else
   #include <DWrite_1.h>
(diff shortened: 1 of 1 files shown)


ebraminio:
PTAL

With this change dw backend became much much better, but there are some issue that AFAICT are internal bugs
* rlig and perhaps other enabled features can not be disabled, nor it could be on uniscribe backend nor IE/Edge
* "هٔ" has rendering, which is consistent behaivour with Edge but not IE!
Our output:
image
Edge's:
image

Other than these, I don't see any other remained functional issue for this backend, other than features range and sideway/vertical text I'd guess.

ebraminio:
Now the main plan, just added initial justification support to the back-end...

ebraminio:
Sample justification result:
image

Not bad! I guess there are some issues on `hb_buffer_reverse` but any help would be appreciated.

We should be able to fill lineWidth, `FLOAT lineWidth = 15000;`, but public API. Maybe adding suggested_width or something?

ebraminio:
@behdad, I think this can be landed as is for now until we figure out how we should turn dwrite output into hb buffer correctly. I put whole justification thing behind a flag but if you could have how we should pass width into lineWidth, we can turn most of justification other than GetJustifiedGlyphs which adds Kashida. P.S. Actually, my interest was being able to just test and experiment with MS justification algorithm which even with this visually broken implementation it can do most of the need if we want go for doing Kashida justification someday. P.S.2 Even the dwrite backend implementation now looks useful, it is slow in comparison to both of uniscribe and hb ot, so it is better not be used on any real app. If your need is just having compatible shaping with Windows shaper, uniscribe backend is better choice as nowadays it is just a proxy to dwrite shaper.

ebraminio:
It seems dw backend has some issue with non-BMP char counting but not something added on this patch.

behdad:
Can you describe what the new justification API in DirectWrite looks like?

behdad:
And should I merge, or are you still working on this?

behdad:
In general, the fewer ifdefs we have around, the better.

behdad:
You might be able to use typedefs and polyfill functions, such that the code is cleanly written against new API, but on old API, it fills in the gaps using older API.

ebraminio:
> Can you describe what the new justification API in DirectWrite looks like?
I think [MSDN description](https://msdn.microsoft.com/en-us/library/windows/desktop/hh994452(v=vs.85).aspx#Justification) and [the implemented](https://github.com/behdad/harfbuzz/pull/239/files#diff-bad9a60fe707c3516a7fa7f2bf2d190bR802) which follows that description can explain this well but I can say roughly it gets a line width parameter from user and modifies word spaces and add Kashida if that script have that to reach that shaped glyphs to the user given width.
IMO yes, it should be merged as it fixes various issues of the backend.
> In general, the fewer ifdefs we have around, the better.
Agreed however the API is not supported on prior Win8 and we are not filling [lineWidth](https://github.com/behdad/harfbuzz/pull/239/files#diff-bad9a60fe707c3516a7fa7f2bf2d190bR818) with sensible value so it is just an experiment for now. 
> You might be able to use typedefs and polyfill functions, such that the code is cleanly written against new API, but on old API, it fills in the gaps using older API.
Polyfilling GetJustificationOpportunities and friends with noop can be something not expected from user POV. I think the whole DW backend can be considered just for Win 8 and newer version as the backend target is not harfbuzz users but harfbuzz developers I believe.

ebraminio:
The issue with «هٔ» on Tahoma is resolved on Win10 insider build 10.0.14342, probably the font itself or dw shaper is updated on the newer builds.

PR #241, opened 2016-04-12
Add support for the OpenType MATH table.

fred-wang:
This is a first pull request to add support for the OpenType MATH table.

I think I'm essentially done with the implementation of the API, which provides access to some math values as well as the ability to shape stretchy math characters with the help of the MathVariants table.

I will now wait Behdad's feedback/review before continuing to work on this, but feel free to comment on the proposal as there are certainly aspects of the API to discuss.

Note that I've created math fonts that are used for the unit tests:
https://github.com/fred-wang/MathTestFonts/

Also, I've forked the harfbuzz-tutorial to show the shaping of stretchy math characters:
https://github.com/fred-wang/harfbuzz-tutorial
https://github.com/fred-wang/harfbuzz-tutorial/blob/master/out.png

cc'ing @khaledhosny @behdad @jfkthame @drott


src/hb-ot-layout.cc
```diff
@@ -1137,3 +1159,239 @@ hb_ot_layout_substitute_lookup (OT::hb_apply_context_t *c,
...
+  if (math.has_math_constants()) {
+      hb_position_t value;
+      if (math.get_math_constants().get_value(font, constant, value))
+        return value;
+  }
+  return 0;
+}
+
+/**
+ * hb_ot_layout_get_math_percent:
```
khaledhosny: This does not mach the name of the method. But as I said before, I don’t think the granularity is really needed since the client will need to know any way which constants are percent and which are absolute values and returning `hb_position_t` for both shouldn’t harm (it is just a 32 bit integer).

src/hb-ot-shape-math.cc
```diff
@@ -0,0 +1,717 @@
...
+  { 0x2016, false, 0x2016, 0x2016, 0x0000, 0x0000 }, // DOUBLE VERTICAL LINE
+  { 0x203E, true,  0x203E, 0x203E, 0x0000, 0x0000 }, // OVERLINE
+  { 0x222B, false, 0x23AE, 0x2321, 0x0000, 0x2320 }, // INTEGRAL SIGN
+  { 0x2308, false, 0x23A2, 0x0000, 0x0000, 0x23A1 }, // LEFT CEILING
+  { 0x2309, false, 0x23A5, 0x0000, 0x0000, 0x23A4 }, // RIGHT CEILING
+  { 0x230A, false, 0x23A2, 0x23A3, 0x0000, 0x0000 }, // LEFT FLOOR
+  { 0x230B, false, 0x23A5, 0x23A6, 0x0000, 0x0000 }, // RIGHT FLOOR
+  { 0x23B0, false, 0x23AA, 0x23AD, 0x0000, 0x23A7 }, // \lmoustache
+  { 0x23B1, false, 0x23AA, 0x23A9, 0x0000, 0x23AB }  // \rmoustache
+};
```
khaledhosny: What is this for, some kind of fallback construction?

src/hb-ot-math.h
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * Igalia Author(s): Frédéric Wang
+ */
+
+#ifndef HB_OT_H_IN
+#error "Include <hb-ot.h> instead."
+#endif
+
+#ifndef HB_OT_MATH_HH
+#define HB_OT_MATH_HH
```
khaledhosny: This should be `HB_OT_MATH_H` (fixes one of the make check failures).

src/hb-ot-shape-math.cc
```diff
@@ -0,0 +1,717 @@
...
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ * FITNESS FOR A PARTICULAR PURPOSE.  THE SOFTWARE PROVIDED HEREUNDER IS
+ * ON AN "AS IS" BASIS, AND THE COPYRIGHT HOLDER HAS NO OBLIGATION TO
+ * PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
+ *
+ * Igalia Author(s): Frédéric Wang
+ */
+
+#include "hb-ot-layout-math-table.hh"
+#include "hb-ot-shape-private.hh"
```
khaledhosny: Those two includes need to swapped (fixes another failure).

src/hb-ot-shape-math.cc
```diff
@@ -0,0 +1,717 @@
...
+ * predefined Unicode glyph assembly such as
+ *
+ *                              |LEFT PARENTHESIS UPPER HOOK (U+239B)
+ * LEFT PARENTHESIS (U+0028) => |LEFT PARENTHESIS EXTENSION  (U+239C)
+ *                              |LEFT PARENTHESIS LOWER HOOK (U+239D)
+ *
+ * If the content type of the buffer is HB_BUFFER_CONTENT_TYPE_GLYPHS then that
+ * element is a glyph index and we try and use the OpenType MATH table to
+ * find a larger glyph from a MathGlyphVariantRecord subtable or a glyph
+ * assembly described in a GlyphAssembly subtable.
```
khaledhosny: Not sure I like this behaviour, seems different from general HarfBuzz shaping. I’d expect this API to either accept only `HB_BUFFER_CONTENT_TYPE_UNICODE` and it would then call `hb_shape()` first or (my preference) accept only `HB_BUFFER_CONTENT_TYPE_GLYPHS` and require the client to call `hb_shape()` or `hb_shape_full()` first. The predefined Unicode assembly is unnecessary IMO and if a client really need it (not using fonts with MATH table) then it probably won’t be using this API anyway since there is no much to gain here.

src/hb-ot-shape-math.cc
```diff
@@ -0,0 +1,717 @@
...
+ * either a single glyph or the list of glyph assembly parts:
+ *
+ * - If the base element or large glyph is used, the output contains the
+ * corresponding glyph positioned at the origin and with the glyph advances.
+ *
+ * - If a glyph assembly is used then it can not contain more than
+ * HB_OT_MATH_MAXIMUM_PART_COUNT_IN_GLYPH_ASSEMBLY parts. The order of element
+ * in the output buffer is from left to right glyph for horizontal assembly and
+ * from the bottom to top for vertical assembly. All the glyphs have the same
+ * cluster value as the element from the input buffer. All but the glyphs have
```
khaledhosny: I think something is missing here `All but the glyphs have`?

src/hb-ot-shape-math.cc
```diff
@@ -0,0 +1,717 @@
...
+ * null advances, except the last one which holds the advance of the whole glyph
+ * assembly. The top left glyph is positioned at the origin and the offsets of
+ * the other glyphs are calculated relative to that glyph.
+ *
+ * If none of the option is enough to cover the target size, then the buffer is
+ * set with the best one.
```
khaledhosny: Shouldn’t that be `the last one`?

src/hb-ot-shape-math.cc
fred-wang: Yes, it's described in the API as you noted below.

src/hb-ot-shape-math.cc
fred-wang: Note sure I'm happy with it either ;-)

Again, I'm wondering how much logic we can move from Gecko / WebKit to Harfbuzz and they have these kinds of Unicode fallback.

(Of course, maybe progress with DejaVu/Noto/STIX will make MATH fonts available on Linux/Android/OS X/iOS and this may become unneeded. Actually I think DejaVu already has these Unicode constructions now)

src/hb-ot-shape-math.cc
fred-wang: Fixed

src/hb-ot-shape-math.cc
fred-wang: Fixed too.

fred-wang:
Discussion for Gecko
https://bugzilla.mozilla.org/show_bug.cgi?id=1009582#c18
(Maybe we'd want to move the discussion to the harfbuzz mailing list if that's more convenient for everybody)

src/hb-ot-shape-math.cc
```diff
@@ -0,0 +1,717 @@
...
+          buffer->pos[j + 1].x_offset = 0;
+          buffer->pos[j + 1].y_offset = buffer->pos[j].y_offset + delta_offset;
+        }
+      }
+    }
+  }
+
+  // Now we try and increase the overlap between parts in order to get closer
+  // to the target size. Again, for symmetry reason we do it by browsing the
+  // from the middle and simultaneously towards the start and end.
```
fred-wang: So here we depart a bit from what the MATH table spec says. This preserves the symmetry of the operator but not the condition that things are distributed equally. We should probably do a first round to increase all the overlaps by the same amount. And if that is still not enough maybe do this correction to reduce the size any further.

fred-wang:
Regarding the get_math_constant_* functions, I'm happy to merge them as that would simplify code (as the cost of having bad semantic used for hb_position_t ; maybe we should change the signature to use unsigned int and/or clearly state in the API doc the meaning of the returned value for percent / length constants).

Removing the Unicode fallback too will also simplify the code and also fix the remaining "make check" failure about GlyphAssemblyFromUnicode and GlyphAssemblyFromMathTable being exposed.

src/hb-ot-shape-math.cc
khaledhosny: I’m thinking this wouldn’t be much useful, since I expect the clients to check of the fonts has a math table before using any of these new API calls, and the chance that a font has a MATH table without such basic stretchy glyphs is unlikely and if it does it is a font bug after all. On the other hand clients might intentionally call this method without checking for MATH table to get the fallback, and HarfBuzz does fallback text shaping as well so this is not something new. But why stop here and don’t provide fallback math constants as well?

fred-wang:
@khaledhosny:

I merged get_math_constant_percent into get_math_constant. After discussion on Mozilla's side, I also removed the Unicode fallback. Now Travis build seems to pass on Mac and to build on Linux but the latter still has some errors and I'm not sure how to check src/test-suite.log (that does not happen on my machine)

One remaining thing that we'd probably like to preserve from Mozilla is the ability to find the best match, not just to stop search at the first upper bound. One proposal would be to have three options to pass to hb_ot_shape_math_stretchy and hb_ot_shape_math_stretchy_max_orthogonal_advance:

NEAREST: minimize the distance (on a linear or logarithm scale?) between best_size and target_size 
LOWEST_UPPER_BOUND: same as NEAREST by with best_size <= target_size
GREATEST_LOWER_BOUND: same as NEAREST by with best_size => target_size

khaledhosny:
Great.

For the failure you can temporarily modify .travis.yml to cat the src/test-suite.log and see where the failure is.

I don’t quite understand the best match issue (I have been following the bugzilla discussion as well). I mean how is Mozilla requirements different from TeX’s which AFAIK always returns the variant that is ≥ the requested size, with \delimiterfactor and \delimitershortfall fine tuning it?

fred-wang:
> For the failure you can temporarily modify .travis.yml to cat the src/test-suite.log and see where the failure is.

Thanks, I'll try that.

> I don’t quite understand the best match issue (I have been following the bugzilla discussion as well). I mean how is Mozilla requirements different from TeX’s which AFAIK always returns the variant that is ≥ the requested size, with \delimiterfactor and \delimitershortfall fine tuning it? 

I'll have to check that with Karl again (it seems from his Bugzilla status that he went to vacation).
I don't know exactly if that's a strict requirement (you know, it's verrry old code...) but at the moment Gecko has various way of deciding whether a size is OK and whether one size is better than another ; and it tries all the fonts in the CSS context. Maybe that can change, but let's see what Karl says. Anyway for me it's now clear how I would implement the best match in Harfbuzz if that turns out to be necessary (my only concern is that it may be a bit less efficient to test all sizes to find the best one instead of stopping at the first one).

khaledhosny:
I’m testing `hb_ot_shape_math_stretchy()` and I notices that with XITS Math the parentheses always gives me the base glyph unless I specify a very big target size. Debugging this code I found that `try_base_glyph` is always returning true until a very big size is requested which seems to be because `hb_font_get_glyph_v_advance()` returns such a big size. Not sure what is wrong here, might be that we need to multiply the target size with font size? Any way I think `hb_font_get_glyph_[h,v]_advance()` might not be the most suitable methods since they give the glyph advance for horizontal or vertical writing direction, but what you need is the glyph bounding box since the advance can be bigger or smaller than the bounding box, so `hb_font_get_glyph_extents` might be more suitable here (not that HarfBuzz’s ot font functions does not support glyph extents for CFF table yet, so a client using them will be getting wrong results). Or may be we can just drop `try_base_glyph` altogether unless it is really needed (optimisation?).

fred-wang:
Yes, what we want is the exact bounding box ("ink" metrics). I think it is safe to remove try_base_glyph as anyway the client can just test it before doing the stretch.

hb_font_get_glyph_*_advance calls are used elsewhere but the ink metrics is also we want if we need to align glyphs in an assembly (for now I'm assuming parts are designed in a way that we do not need to shift them in orthogonal directions) and to calculate the max orthogonal advance (where exact bounding box is probably more accurate).

fred-wang:
This afternoon, I wrote an article on my blog to explain the algorithm used to build the glyph assembly. I realized that maybe repeating the extenders in the output buffer is a bit overkill. Instead, we could just position the first copy of each extender and output the overlap value used. Then it would be up to the client to determine the offsets of the other copies during the painting phase. I don't think Gecko/WebKit currently create structures of size the number of glyphs (counting extender repetition) and I don't think that's necessary except to keep closer to hb_shape

This comes back to the issue about providing width/height/italic correction raised in the past. Currently, the x_advance and y_advance on the last glyphs actually gives the size of the whole assembly. A single-glyph output buffer does not necessarily mean we have a variant (a glyph assembly can in rare case be made of a single non-extender glyph if that one is large enough) so maybe for convenience we could also output the italic correction (and then remove hb_shape_get_italic_correction_for_glyph_assembly).

It seems that user data can be attached to hb_buffer_t, which may be a more convenient way to output info like italic correction, glyph assembly overlap, kind of stretching used etc instead of adding new arguments to hb_shape_math_stretchy.

khaledhosny:
> I realized that maybe repeating the extenders in the output buffer is
> a bit overkill. Instead, we could just position the first copy of each
> extender and output the overlap value used. Then it would be up to the
> client to determine the offsets of the other copies during the
> painting phase.

Can you elaborate a bit more about this, not sure I understand how this
would work.

> This comes back to the issue about providing width/height/italic
> correction raised in the past. Currently, the x_advance and y_advance
> on the last glyphs actually gives the size of the whole assembly. A
> single-glyph output buffer does not necessarily mean we have a variant
> (a glyph assembly can in rare case be made of a single non-extender
> glyph if that one is large enough) so maybe for convenience we could
> also output the italic correction (and then remove
> hb_shape_get_italic_correction_for_glyph_assembly).

Yes, I think this would be better.

> It seems that user data can be attached to hb_buffer_t, which may be a
> more convenient way to output info like italic correction, glyph
> assembly overlap, kind of stretching used etc instead of adding new
> arguments to hb_shape_math_stretchy.

I think user data is for the clients and HarfBuzz itself shouldn’t be
using it. May be we can add new fields to `hb_buffer_t` and have
something like `hb_buffer_get_math_italic_correction` etc.

khaledhosny:
> I realized that maybe repeating the extenders in the output buffer is
> a bit overkill. Instead, we could just position the first copy of each
> extender and output the overlap value used. Then it would be up to the
> client to determine the offsets of the other copies during the
> painting phase.

Can you elaborate a bit more about this, not sure I understand how this
would work.

> This comes back to the issue about providing width/height/italic
> correction raised in the past. Currently, the x_advance and y_advance
> on the last glyphs actually gives the size of the whole assembly. A
> single-glyph output buffer does not necessarily mean we have a variant
> (a glyph assembly can in rare case be made of a single non-extender
> glyph if that one is large enough) so maybe for convenience we could
> also output the italic correction (and then remove
> hb_shape_get_italic_correction_for_glyph_assembly).

Yes, I think this would be better.

> It seems that user data can be attached to hb_buffer_t, which may be a
> more convenient way to output info like italic correction, glyph
> assembly overlap, kind of stretching used etc instead of adding new
> arguments to hb_shape_math_stretchy.

I think user data is for the clients and HarfBuzz itself shouldn’t be
using it. May be we can add new fields to `hb_buffer_t` and have
something like `hb_buffer_get_math_italic_correction` etc.

fred-wang:
> Can you elaborate a bit more about this, not sure I understand how this would work.

So here is a simplified example if the assembly is made of 3 parts all of advance a=1000 and if the calculated connector overlap is o = 100, we currently output something like

-------
Left
x_offset[0] = 0

Extender
x_offset[1] = a - o = 900

Extender
x_offset[2] = 1800

...

Extender
x_offset[r] = 900r

Right
x_offset[r+1]= 900 * (r + 1)
x_advance[r+1] = 900 * (r + 1) + 1000
------------

i.e. a buffer of size (r + 2). Instead we could just output a buffer with 3
elements + the value of o:

------------
Left
x_offset[0]=0

Extender
x_offset[1] = a - o = 900

Right
x_offset[2] = 900 * (r + 1)
x_advance[2] = 900 * (r + 1) + 1000
------------

and when the client see that x_offset[1] + (a - o) != x_offset[2], it deduce that it is an extender that should be repeated K times

x_offset[1] + K * (a - o) = x_offset[2]

(of course K = r)

But now I realize that the client does not know the advances of the output glyphs (at least not the values taken from the MathVariant table)...

khaledhosny:
I’d keep it as it is now, it is simple for the client and matches how HarfBuzz does shaping in general and I don’t think repeating the extenders is that bad.

fred-wang:
Yes, I think in practice the repetition of extenders will remain small so it's probably not worth moving too far away from HarfBuzz's normal shaping behavior.

khaledhosny:
`hb_ot_shape_math_stretchy()` asserted on me a couple of times, here is a small program to reproduce it:
````c
#include <assert.h>
#include <stdio.h>

#include <hb.h>
#include <hb-ot.h>
#include <hb-ft.h>

int
main(int argc, char **argv)
{
  if (argc < 2)
  {
    printf ("Usage: bug xits-math.otf\n");
    return 1;
  }
  
  FT_Library library;
  assert (!FT_Init_FreeType (&library));

  FT_Face face;
  assert (!FT_New_Face (library, argv[1], 0, &face));
  FT_Set_Char_Size (face, 24, 0, 0, 0);

  hb_font_t *font = hb_ft_font_create (face, NULL);

  hb_buffer_t *buffer = hb_buffer_create ();
  hb_buffer_set_direction (buffer, HB_DIRECTION_LTR);
  hb_buffer_set_script (buffer, hb_script_from_string ("Math", -1));
  hb_buffer_add_utf8 (buffer, "(", 1, 0, 1);

  hb_shape (font, buffer, NULL, 0);
  hb_ot_shape_math_stretchy (font, buffer, 0, 4220);
}
````

This seems to happen when `set_glyph_assembly()` returns false after changing the buffer length and the subsequent call to `set_single_glyph()` asserts the buffer length is 1.

fred-wang:
Thanks, the initial intention was that set_glyph_assembly always true after setting the buffer size. However, in my last refactoring I added this line:

     if (connectorOverlap < minConnectorOverlap) return false;

Probably the best would to set the buffer properties only after having determined connectorOverlap i.e. just before "Add the glyph to the buffer and position them.".

However, I'm not sure why connectorOverlap < minConnectorOverlap could happen with XITS, I'll need to debug that.

fred-wang:
@khaledhosny OK, I've moved the setting of the buffer after the calculation of max overlap so the assert should no longer happen. That calculation was also badly implemented and did not match what I indicated on my blog so the overlap ended up being 0 < minOverlap. This should be fixed now.

behdad:
I really hope to start reviewing your work late next week :).

khaledhosny:
`hb_ot_shape_math_stretchy()` seems to work fine now in my limited testing (apart from the `try_base_glyph()` issue).

fred-wang:
> hb_ot_shape_math_stretchy() seems to work fine now in my limited testing (apart from the try_base_glyph() issue).

Great.

> I really hope to start reviewing your work late next week :)

No problem. Anyway, I think Karl is back next week, so I'd like to wait his opinion on Gecko's needs. BTW, I initially wanted to make a clean pull request but there have been a few more commits and it's likely that more changes will happen... If you wish a stashed commit or split differently to help review, please tell me...

fred-wang:
I'm just adding where the Mozilla discussion stopped here:

1) It seems that we need more flexibility for the size variant selection, so the idea would be instead to add functions to expose the size variants (count, glyph id and advance).

2) Then the shaping function will only be used for the glyph assembly (i.e. the most complex part).

3) The max orthogonal advance function can finally be removed since they can do the calculation from 1) and 2).

@khaledhosny Do you think that will work for you?

@behdad any update on this?

khaledhosny:
I think I like the current math shaping API, it makes things simpler for the client. To be honest I still don’t understand what Mozilla requirements are here, and I hop that there a simpler way to achieve what you want.

robin33n:
Will this feature be merged with master in the near future?

fred-wang:
> Will this feature be merged with master in the near future? 

@robin33n The plan is to work on this during the web engines hackfest at the end of September (http://www.webengineshackfest.org/).

> I think I like the current math shaping API, it makes things simpler for the client. To be honest I still don’t understand what Mozilla requirements are here, and I hope that there a simpler way to achieve what you want.

@khaledhosny I think Karl gave the current features in Gecko and what would need in HarfBuzz to preserve the current behavior.  I'm not sure whether or not all the hacks and workaround should be preserved in the long term but it's probably safer to keep them in the short term. So I guess HarfBuzz should give some flexibility to the clients. I think we could generalize the current proposal to have:

1) One API to expose the whole MATH data (hence including the MathVariants table).

2) One API to shape a stretched operator, given a target size. We could have some generic functions (that receive data for size variants and/or the glyph assembly) and a simple one (taking the base glyph as a parameter and automatically deducing the stretch data to use).

The exact C API for this still has to be designed, though.

fred-wang:
@khaledhosny  @behdad I've rebased, squashed and split into more logical commits so that it will be easier to work with:

https://github.com/behdad/harfbuzz/compare/master...fred-wang:MATH-3

Basically, I think the three first commits (API for math constants and glyph info) are ready for review and we agree on the way to expose these data.

The two other commits still need to be discussed and improved during the hackfest:

- For the 4th one I'm suggesting to extend the API to expose the whole MathVariants table so that Mozilla, WebKit or any other clients implementing its own table parsing can just replace it with HarfBuzz's one.

- The 5th will add a shaping API that could be used by clients to implement general operator stretching, without having to mess up with the MathVariants table themselves. However, I'm suggesting to make it flexible so that clients could still pass their own constructions if they wish.

khaledhosny:
An API that exposes MathVariants sounds fine, flexible shaping is fine by me as well as long as getting the default behavior is straight forward.

behdad:
Thanks Fred.  I'm really busy till after ATypI.  Will try to study this before the hackfest.

fred-wang:
Moved to https://github.com/behdad/harfbuzz/pull/326

PR #244, opened 2016-04-25
Support CPAL table

brawer:


khaledhosny:
Wow, I was planning to git CPAL in HarfBuzz a try in the coming weeks :)

ebraminio:
> Wow, I was planning …
Me too, except the fact I didn't know if I could to :)

brawer:
@behdad, mind having a look?

src/hb-ot-color.h
```diff
@@ -0,0 +1,156 @@
...
+ *
+ * Returns: an identifier within @face's `name` table.
+ * If the requested palette has no name, or if @face has no color,
+ * or if @palette is not between 0 and hb_ot_color_get_palette_count(),
+ * the result is zero.
+ *
+ * Since: 1.2.8
+ */
+HB_EXTERN unsigned int
+hb_ot_color_get_palette_name (hb_face_t *face, unsigned int palette);
```
khaledhosny: I think `_name_id` would be better here, since we are not returning the actual name string.

src/hb-ot-color.h
```diff
@@ -0,0 +1,156 @@
...
+ */
+HB_EXTERN hb_ot_color_palette_flags_t
+hb_ot_color_get_palette_flags (hb_face_t *face, unsigned int palette);
+
+
+/**
+ * hb_ot_color_get_palette_colors:
+ * @face: a font face
+ * @palette: the index of the color palette whose colors are being requested
+ * @start_offset: the index of the first color being requested.
```
khaledhosny: This should be annotated with `(in out)` for introspection (see https://wiki.gnome.org/Projects/GObjectIntrospection/Annotations).

src/hb-ot-color.h
```diff
@@ -0,0 +1,156 @@
...
+ * @color_count: how many colors can be maximally stored into the @colors array
```
khaledhosny: This should be annotated with `(in out)` for introspection (see https://wiki.gnome.org/Projects/GObjectIntrospection/Annotations).

src/hb-ot-color.h
```diff
@@ -0,0 +1,156 @@
...
+ * @colors: an array of #hb_ot_color_t records. After calling this function,
```
khaledhosny: And `(out)` here.

src/hb-ot-color.h
brawer: Done.

src/hb-ot-color.h
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * hb_ot_color_get_palette_count:
+ * @face: a font face
+ *
+ * Returns: the number of color palettes in @face, or zero if @face has
+ * no color
+ *
+ * Since: 1.2.8
+ */
+HB_EXTERN unsigned int
```
khaledhosny: We usually have the method documentation in the `.cc` files, can you please move it there for consistency?

src/hb-ot-layout.cc
```diff
@@ -60,6 +61,9 @@ _hb_ot_layout_create (hb_face_t *face)
   layout->gpos_blob = OT::Sanitizer<OT::GPOS>::sanitize (face->reference_table (HB_OT_TAG_GPOS));
   layout->gpos = OT::Sanitizer<OT::GPOS>::lock_instance (layout->gpos_blob);
 
+  layout->cpal_blob = OT::Sanitizer<OT::CPAL>::sanitize (face->reference_table (HB_OT_TAG_CPAL));
+  layout->cpal = OT::Sanitizer<OT::CPAL>::lock_instance (layout->cpal_blob);
+
```
khaledhosny: I’m just wondering if we should load the `CPAL` table lazily like #241 does for the `MATH` table, though I’m not sure if it makes any practical difference.

brawer: Sure, done.

src/hb-ot-color.h
brawer: Sure, done.

src/hb-ot-color.h
brawer: From the GObject introspection doc, (out caller-allocates) (array length=color_count) (optional) seemed most appropriate; please tell if it’s wrong.

src/hb-ot-color.h
brawer: From the GObject introspection doc, `(inout) (optional)` seemed most appropriate; please tell if it’s wrong.

src/hb-ot-layout.cc
brawer: Done.

brawer: Done.

brawer:
@khaledhosny, thanks for the review! Mind having another look? I’ve rebased my branch on the current master, and squashed my various commits into one single change.

src/hb-ot-color.cc
```diff
@@ -0,0 +1,213 @@
...
+ * @face: a font face.
+ * @palette: the index of the color palette whose name is being requested.
+ *
+ * Retrieves the name id of a color palette. For example, a color font can
+ * have themed palettes like "Spring", "Summer", "Fall", and "Winter".
+ *
+ * Returns: an identifier within @face's `name` table.
+ * If the requested palette has no name, or if @face has no colors,
+ * or if @palette is not between 0 and hb_ot_color_get_palette_count(),
+ * the result is zero.
```
khaledhosny: I’m just wondering since zero is a valid name id (that is the copyright of course), wouldn’t it be better to return an invalid id in case of error, e.g. -1, to be clear this is not a valid value?

src/hb-ot-color.h
khaledhosny: I’m not an expert either, but when trying this method in Python it always returns the number of colors and an empty colours list. No idea how this is best handled.

src/hb-ot-color.cc
brawer: Done.

src/hb-ot-color.cc
```diff
@@ -0,0 +1,213 @@
...
+    return HB_OT_COLOR_PALETTE_FLAG_DEFAULT;
+  }
+
+  const OT::CPALV1Tail& cpal1 = OT::StructAfter<OT::CPALV1Tail>(cpal);
+  if (unlikely (&cpal1 == &OT::Null(OT::CPALV1Tail) ||
+		cpal1.paletteFlags.is_null())) {
+    return HB_OT_COLOR_PALETTE_FLAG_DEFAULT;
+  }
+
+  const OT::ULONG* flags = &cpal1.paletteFlags(&cpal);
```
brawer: @khaledhosny and @behdad, with my current implementation, I’m not sure if the sanitizer will catch attacks where paletteFlags (which contains an offset relative to the beginning of the CPAL table) points beyond the limits of the CPAL structure. Does this line here (or the dereferencing on line 141) need to be changed? Or the sanitizer? Thanks for double-checking.

brawer:
r?

brawer:
Quarterly ping?

behdad:
Will get there after variations stuff is finished.

ebraminio:
Wine has just released a version with COLR/CPAL color Emoji rendering support, the code is probably worth to read https://github.com/wine-mirror/wine/blob/master/dlls/dwrite/opentype.c#L759

ebraminio:
Now that CBDT is in, I think having this in also would be fun, specially on the same release, sorry for spamming.

behdad:
I pushed a release out right now.  But this is next in my review queue indeed.

KrasnayaPloshchad:
@brawer I have found a notice that this branch has conflicts, can you fix it?

KrasnayaPloshchad:
@brawer Can you fix the conflicts?

behdad:
No need to resolve conflicts.  I'll do that when I get to review this.

brawer:
@behdad, have a look? The review shouldn’t take long, and then we’ll have one less thing on the plate...

test/api/test-ot-color.c
```diff
@@ -0,0 +1,318 @@
...
+  } \
+  if (colors[i].blue != blue) { \
+    g_assertion_message_cmpnum (G_LOG_DOMAIN, __FILE__, __LINE__, G_STRFUNC, \
+				"colors[" #i "].blue", colors[i].blue, "==", blue, 'x'); \
+  } \
+  if (colors[i].alpha != alpha) { \
+    g_assertion_message_cmpnum (G_LOG_DOMAIN, __FILE__, __LINE__, G_STRFUNC, \
+				"colors[" #i "].alpha", colors[i].alpha, "==", alpha, 'x'); \
+  } \
+} G_STMT_END
```
dscorbett: This macro currently has an effort at safety but doesn’t go all the way. To make it safer, the red condition should use `red` instead of `r`; the green, blue, and alpha conditions should use `_colors[_i]` instead of `colors[i]`; and the green, blue, and alpha assertions should also use `_colors[_i]` instead of `colors[i]`.

src/hb-ot-color.cc
```diff
@@ -0,0 +1,213 @@
...
+
+#include "hb-ot-layout-private.hh"
+#include "hb-shaper-private.hh"
+
+HB_MARK_AS_FLAG_T (hb_ot_color_palette_flags_t)
+HB_SHAPER_DATA_ENSURE_DECLARE(ot, face)
+
+
+static inline const OT::CPAL&
+_get_cpal (hb_face_t *face)
```
behdad: We have a simpler way to do this now. Check how MATH table is accessed.

src/hb-ot-color.cc
```diff
@@ -0,0 +1,213 @@
...
+ * Returns: the number of color palettes in @face, or zero if @face has
+ * no colors.
+ *
+ * Since: 1.2.8
+ */
+unsigned int
+hb_ot_color_get_palette_count (hb_face_t *face)
+{
+  const OT::CPAL& cpal = _get_cpal(face);
+  return &cpal != &OT::Null(OT::CPAL) ? cpal.numPalettes : 0;
```
behdad: You don't need the Null check.  Just access cpal.numPalettes freely.  Same in other places.

src/hb-ot-color.cc
```diff
@@ -0,0 +1,213 @@
...
+unsigned int
+hb_ot_color_get_palette_name_id (hb_face_t *face, unsigned int palette)
+{
+  const OT::CPAL& cpal = _get_cpal(face);
+  if (unlikely (&cpal == &OT::Null(OT::CPAL) || cpal.version == 0 ||
+		palette >= cpal.numPalettes)) {
+    return 0;
+  }
+
+  const OT::CPALV1Tail& cpal1 = OT::StructAfter<OT::CPALV1Tail>(cpal);
```
behdad: Logic should be moved to cpal table implementation.  The API should be one-line call into the `cpal.get_name_id()`.

Also, function name should be `hb_ot_color_palette_get_name_id`.

behdad:
> @behdad, have a look? The review shouldn’t take long, and then we’ll have one less thing on the plate...

I haven't seen a short review in a long time :).  Left enough comments to keep you busy.  Plus, there are many higher priority issues on my plate and should be on your plate too!  SVG table for example.

timofonic:
@behdad Any plans to add CPAL table support?

ebraminio:
How about dropping this and trying to progress http://savannah.nongnu.org/bugs/?44689 as freetype is better conceptual match for this. I remember, @Behdad, that you were saying as the platforms using harfbuzz now also shipping freetype also anyway (shipping of freetype on Windows/macOS Chrome) there is not much point for progressing this inside harfbuzz.

behdad:
I have to think about it... I know I also want to implement SVG and probably CBDT/CBLC / sbix at some point...

behdad:
But yeah, making progress on the FreeType patch would be great.  I haven't got to work on it yet.

twardoch:
This is a good question. I still think COLR/CPAL support in HarfBuzz would be beneficial in HB  for clients that use HB but don't use FreeType — for example Firefox or XeTeX, which could do their own painting. 

BTW, the COLR format was clearly inferior to SVG for static fonts but I think it'll stay with us for a while because it's the only format that can currently be used to realize variable color fonts (I mean it's implicitly possible already to do variable color fonts via COLR).

twardoch:
Variable OT+SVG fonts aren't even planned and it'd be hard to say how to do them. OT+SVG glyphs don't just contain points but can contain primitives such as circles or rectangles, and of course bitmaps. There of course approaches in SVG proper that allow animation, and I think those techniques could be used to power variable OT+SVG fonts, but that stuff is years ahead. Variable COLR fonts are, on the other hand, possible today.

twardoch:
which reminds need that I don't think CSS has a method of selecting or specifying a color font pallette. There should be a way to do it, with syntax possibly akin to how gradients are doing in CSS: 
```
color: palette(2);
color: palette(black, yellow, green, rgba(244, 235, 5, 0.8), #2244a7);
```

Instead of `palette`, it could be `font-palette`.

jfkthame:
See https://drafts.csswg.org/css-fonts-4/#color-font-support (currently in draft).

khaledhosny:
`SVG` table might be superior to `COLR` for emoji, but emoji is not the only application for color fonts. There are uses like [this](http://www.amirifont.org/fatiha-colored.html) and [this](http://www.babelstone.co.uk/Fonts/Irk_Bitig_Test.html) where color is used for regular text where you want to retain hinting and other text rasterization features not supported with SVG.

twardoch:
I forgot that the OpenType+SVG spec internally uses CSS variables: 
```
--color0: rgba(255,0,0, 0.5);
```

I don't quite understand whether users are expected to be using such constructed variables themselves and then the font would pick them up.

PR #247, opened 2016-04-26
Update opentype language tags

khaledhosny:
This is rebased version of #168 with the commits squashed together.


behdad:
Looks like the list is not sorted.  Please run the test suite and fix.

behdad:
Feel free to merge after.

khaledhosny:
The order should be fixed now, there were a few duplicates so I moved them down to the commented out section for further consideration.

PR #257, opened 2016-05-23
Introduce hb-view-skia.

aam:
To follow-up on https://bugs.chromium.org/p/skia/issues/detail?id=4742#c14 this is working sketch of using Skia as backend for hb-view. Curious what you think, Behdad!

---

To build one needs to have Skia headers and libraries. Easiest way seem to be get Skia sources from https://skia.googlesource.com/skia, configure using cmake and build using ninja (cd ~/skia/skia/cmake && cmake . && ninja).
After that one can point to generated compile and link option files to auto-configure Harfbuzz with Skia:

```
SKIA_CFLAGS=`cat ~/skia/skia/cmake/skia_compile_arguments.txt | tr '\n' ' '` SKIA_LIBS=`cat ~/skia/skia/cmake/skia_link_arguments.txt | tr '\n' ' '` ./autogen.sh  --with-skia
```

To run:

```
cd util
make && ./hb-view-skia --output-format=pdf --font-file=${HOME}/fonts/DejaVuSans.ttf --font-size=64 < view-skia.cc > test.pdf && xdg-open test.pdf
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -459,7 +459,27 @@ if test "$os_win32" = no && ! $have_pthread; then
 	AC_SEARCH_LIBS(sched_yield,rt,AC_DEFINE(HAVE_SCHED_YIELD, 1, [Have sched_yield]))
 fi
 
+dnl ==========================================================================
+
+AC_ARG_WITH(skia,
+       [AS_HELP_STRING([--with-skia=@<:@yes/no/auto@:>@],
+                       [Use skia @<:@default=auto@:>@])],,
+       [with_skia=auto])
+have_skia=false
+if test "x$with_skia" = "xyes" -o "x$with_skia" = "xauto"; then
+       PKG_CHECK_MODULES(SKIA, skia >= 1.8.0, have_skia=true, :)
+fi
+if test "x$with_skia" = "xyes" -a "x$have_skia" != "xtrue"; then
+       AC_MSG_ERROR([skia support requested but not found])
+fi
+if $have_skia; then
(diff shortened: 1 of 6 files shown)


ebraminio:
Hey @aam, skia build doesn't seem to work, can you update your instruction for building this? Too many skia cmake build instruction are out there, is there one still working?

aam:
I think skia moved away from cmake to gn about year ago, so doubt you will find working cmake-based build instructions for skia.
I had to update instructions for building https://github.com/aam/skiaex about six months ago.

ebraminio:
Ha, so I believe maintaining load of this on harfbuzz project (e.g. putting it on CI) would be a little hard to have, lets defer this to sometime in future.

behdad:
Yeah agree this can be done later.  Also, ideally we should just build it into hb-view itself, with a commandline argument to choose graphics library.

twardoch:
Indeed, it would be cool if Harfbuzz had a configurable choice of rendering backends similar to how you can conditionally enable and disable the shapers (ot, graphite, coretext, uniscribe and so on). 

It would be useful to be able to use a minimal stack (only libpng and freetype), or cairo, or skia etc. — be able to include or exclude these renderers at compile time and then choose one  at runtime. 

"shapers" and "renderers" are the two essential stacks used during text layout, and it would be nice to treat them in a similar, pluggable way.

khaledhosny:
That is about the command line `hb-view` utility. HarfBuzz itself is not tied to any graphics library of any kind and knows nothing about them, HarfBuzz output is glyph indices, advances, and offsets.

PR #266, opened 2016-06-15
Blacklist GDEF tables of more fonts

jfkthame:
This is based on bug reports that have been filed against Firefox since it updated to a version of harfbuzz that uses zeroing by GDEF rather than by Unicode. I'm sure there are a bunch more font versions that should also be included; these are just the ones I have on hand and have confirmed as having bad GDEF data.

Given how the list here is growing, I think we should reconsider the approach, and perhaps revert to zeroing by Unicode instead.

### DIFF
--- a/src/hb-ot-layout.cc
+++ b/src/hb-ot-layout.cc
@@ -93,6 +93,45 @@ _hb_ot_layout_create (hb_face_t *face)
      if (3 == layout->gdef->get_glyph_class (5))
        layout->gdef = &OT::Null(OT::GDEF);
     }
+    else if (0
+      /* sha1sum:96eda93f7d33e79962451c6c39a6b51ee893ce8c  tahoma.ttf from Windows 8 */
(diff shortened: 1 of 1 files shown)


behdad:
Any idea why the Tibetan doesn't reproduce with IE and other Uniscribe-based systems?  Or it does?

At this point I think the question is do we want to be forward-looking, or backward.  In the past, these bugs went unnoticed on Windows, because windows didn't do shaping for Latin, at least in many situations.  The question is: is that still the case wit DirectWrite?  I like to align with what DirectWrite is doing moving forward and have a limited blacklisting of past broken fonts, rather than keep diverging from Windows forever.

jfkthame:
> Any idea why the Tibetan doesn't reproduce with IE and other Uniscribe-based systems? Or it does?

I'm away from home at the moment and don't have my Windows systems with me, but I'll try to check next week.

behdad:
https://github.com/behdad/harfbuzz/issues/264

PR #269, opened 2016-06-17
use unicode codepoints instead of byte arrays

kelvin13:


src/sample.py
```diff
@@ -39,7 +40,7 @@ def message (self, buf, font, msg, data, _x_what_is_this):
 		return True
 debugger = Debugger()
 hb.buffer_set_message_func (buf, debugger.message, 1, 0)
-hb.buffer_add_utf8 (buf, text.encode('utf-8'), 0, -1)
+hb.buffer_add_codepoints (buf, codepoints, 0, -1)
```
khaledhosny: `hb_buffer_add_codepoints()` does not validate the text and i’m not sure if Python strings can never contain invalid Unicode characters, it is better to use the `_utf32` method as it will work with the same input but HarfBuzz will also make sure the input is valid.

kelvin13: I don’t think python can ever create an invalid string can it?

````
>>> a = "ab\xfgXa"
  File "<stdin>", line 1
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in
position 2-4: truncated \xXX escape
````

Comment (khaledhosny) [inline src/sample.py]: It accepts surrogate characters for one, which are invalid in UTF-32:
```python
a = "\uD800\uDFFF"
```
This is a valid string in Python, but not valid UTF-32 string and if you pass such code units to `_add_utf32()` it will replace them with the replacement character but `_add_codepoints()` won’t.

kelvin13: Would it work if I just changed the function to add_utf32? and still use
ord() ? I don’t understand unicode encoding that well sorry

khaledhosny: It should.

kelvin13: updated pull request

kelvin13: if add_utf32() does everything add_codepoints() does and better, why are
there two functions?

khaledhosny: See bcba8b45024e1eca8be77ca2657de1dc44dbf8fb.

kelvin13: so it does the exact same thing except not validate? is it worth pointing
that out in the/any documentation or something since 99.9% of the time
you’re not going to get invalid strings in python at least

khaledhosny: The [documentation](http://behdad.github.io/harfbuzz/harfbuzz-Buffers.html#hb-buffer-add-codepoints) mentions this, do you think it is not clear enough?

src/sample.py
```diff
@@ -39,7 +40,7 @@ def message (self, buf, font, msg, data, _x_what_is_this):
 		return True
 debugger = Debugger()
 hb.buffer_set_message_func (buf, debugger.message, 1, 0)
-hb.buffer_add_utf8 (buf, text.encode('utf-8'), 0, -1)
+hb.buffer_add_utf32 (buf, codepoints, 0, -1)
```
khaledhosny: `-1` should be `len(codepoints)` since your list is not NULL-terminated.

kelvin13: so that’s what the -1 does…what does the 0 do then?

kelvin13: and do you need len(codepoints) for the add_codepoints() function too?

khaledhosny: See http://behdad.github.io/harfbuzz/harfbuzz-Buffers.html#hb-buffer-add-codepoints

khaledhosny:
Can you rebase on master (to fix Travis failure) and squash the commits, so that I can merge this?

kelvin13:
@khaledhosny like this https://github.com/behdad/harfbuzz/pull/271 ?

khaledhosny:
You can update the same pull request (by pulling, rebasing, then forced push), but a new pull request is fine. Closing this then.

PR #270, opened 2016-06-18
Try to see why Travis make check is failing

khaledhosny:


behdad:
Printing out test suite logs in case of failure sounds useful.

khaledhosny:
I couldn’t find a straight forward way to prevent the test log on failure, the cat trick won’t work on general since we have several test suites.

The failure here is just a “narrow” Python build issue.

PR #273, opened 2016-06-21
	[dwrite] Use stream font loader instead GDI interop

ebraminio:
This patch changes DirectWrite backend to use dwrite font file stream API instead falling back to Uniscribe suited solution, and also improves the backend to not create dwrite font object each shape time which have led to a significant performance boost on my tests. As the result of these changes the incorporated tricks to create HFONT objects are no longer needed on this backend so most of GDI backend shared codes are also removed.

In case no review would be given, I think this is ready to merge.


ebraminio:
Also all the ifdefs are removed now as [you wished so](https://github.com/behdad/harfbuzz/pull/239#issuecomment-207028088) 😊. Even the fact it now needs newer SDK in order to build but executable output is usable on older Windows versions (I've tested it on Win 7)

behdad:
Thanks Ebrahim!

PR #279, opened 2016-06-27
[dwrite] Provide an experimental API for justification

ebraminio:
This is just created to start brain storming around the way we could have some initial justification API of at least directwrite backend on harfbuzz.

Ideally, from the start, we should think about how current harfbuzz can be developed in a way to provide a justification API that would be possible to be integrated with caching word shapers solutions where the "line" concept itself is less relevant...

Obviously harfbuzz API differs from paragraph like API such as DirectWrite and that's why integrating justification with the current state of harfbuzz and its clients usage needs some considerate and innovative thinking :)

### DIFF
--- a/src/hb-directwrite.cc
+++ b/src/hb-directwrite.cc
@@ -209,8 +209,8 @@ _hb_directwrite_shaper_face_data_destroy(hb_directwrite_shaper_face_data_t *data
     data->fontFile->Release ();
   if (data->dwriteFactory) {
     if (data->fontFileLoader)
-      data->dwriteFactory->UnregisterFontFileLoader(data->fontFileLoader);
-    data->dwriteFactory->Release();
+      data->dwriteFactory->UnregisterFontFileLoader (data->fontFileLoader);
+    data->dwriteFactory->Release ();
   }
   if (data->fontFileLoader)
     delete data->fontFileLoader;
@@ -542,12 +542,13 @@ static inline uint32_t hb_uint32_swap (const uint32_t v)
  * shaper
  */
 
-hb_bool_t
-_hb_directwrite_shape(hb_shape_plan_t    *shape_plan,
+static hb_bool_t
+_hb_directwrite_shape_full(hb_shape_plan_t    *shape_plan,
   hb_font_t          *font,
   hb_buffer_t        *buffer,
   const hb_feature_t *features,
-  unsigned int        num_features)
+  unsigned int        num_features,
+  float               lineWidth)
 {
   hb_face_t *face = font->face;
   hb_directwrite_shaper_face_data_t *face_data = HB_SHAPER_DATA_GET (face);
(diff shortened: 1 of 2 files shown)


ebraminio:
@behdad: Lets change direction of this and assume we just want have some way to set lineWidth, what do you suggest for the name of API to show it meant to be experimental and will broke in future? Where should lineWidth parameter should go, inside font_data or face_data or buffer?

ebraminio:
This failed to gain any attention and as I used master branch for the pull request, it somehow blocks my other patches so let's close it for now.

behdad:
> @behdad: Lets change direction of this and assume we just want to have some way to set lineWidth, what do you suggest for the name of API to show it meant to be experimental and will likely be broken in future and where lineWidth parameter should go, inside font_data or face_data or buffer?

The way I always thought about justification is that we will have a different `hb_shape()` call that takes a desired-width parameter, and perhaps a justification-strategy...

khaledhosny:
The bigger problem, IMHO, is how the clients will use such an API? If you have more than one run per line, how do you distribute the desired width on them and what if some runs can be stretched more than the others. It gets even more interesting for applications doing (variation of) Knuth and Plass whole paragraph justification, who will need to know the stretchability and shrinkability of each run (word?) before doing the line breaking.

ebraminio:
> The way I always...

Is it [now](https://github.com/behdad/harfbuzz/pull/279/commits/193de7e55a1f1d8aac04412f543d3f25ab845aac) more similar (hb_shape_dwrite_experimental_width) to what it should as a start point?

> The bigger problem, ...

Indeed it is, lets not call it "lineWidth" but fragment width, aren't clients able to pass some destined width at least on second round of shaping? We don't have concept of line so that is only thing to move forward I can think about. I know [not very confidentially of course] clients like Chromium are currently doing two or more step shaping, first for measuring and line breaking then for actual shapes, result is improved by caching word shaping of the font doesn't involve space on GPOS table or such but if a client destinies to have full justification, second round of shaping is something worth to do I guess anyway.

The idea here primarily is to first provide some way to play with directwrite API.

ebraminio:
> It gets even more interesting for applications doing (variation of) Knuth and Plass whole paragraph justification, who will need to know the stretchability and shrinkability of each run (word?) before doing the line breaking.

I didn't think about that of course, I guess that would be something we should ignore for now 😅 (or let clients to somehow figure out based on changed widths)

behdad:
It's a hard problem indeed, and that's why no one has implemented it properly so far...

Last time I talked to @jfkthame, our conclusion was that clients will probably want to shape normally first, decide line break, then reshape the runs in the line, with prorated width per run and adjust as they reshape.  Something as handwavy as that...

ebraminio:
Can we merge this Behdad? Since hb-directwrite is not provided on releases this is not a harmful thing to have I guess. I should also note that result of justification itself is far from perfect and the intention of the API is not providing it to hb clients, just like the development of the backend as whole.

image

behdad:
Feel free to merge.  Thanks.

PR #281, opened 2016-06-29
WIP [gobject] Use a identifier-filter

behdad:
Based on filter in
https://bugzilla.gnome.org/show_bug.cgi?id=706898

Doesn't quite work.  We now get hb.buffer.buffer_create()
instead of the old hb.buffer_create().  The filter itself
is also crude; not sure what it should do.

This is as much progress as I made tonight.  To be refined.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -9,6 +9,8 @@ DISTCLEANFILES =
 MAINTAINERCLEANFILES =
 DISTCHECK_CONFIGURE_FLAGS = --enable-introspection
 
+PYTHON = python
+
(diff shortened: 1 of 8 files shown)


khaledhosny:
`hb-gobject-identfilter.py` file is missing.

behdad:
Added.

timofonic:
@behdad What happened to this? :O

behdad:
Should we abandon this?  I don't know of anyone using our GI API, but kinda leaving it as is sounds better?

khaledhosny:
Closing in favor of #2462.

Issue #274, opened 2016-06-22
harfbuzz only works once


kelvin13:
well ain’t this the weirdest thing

when I do `make install` on harfbuzz, I can use the library, but only once. Then when I use it again, every codepoint it puts out is 0 so I just get a bunch of rectangles. Every time I import it I get
````
** (kt.py:5410): WARNING **: Failed to load shared library 'libharfbuzz-gobject.so.0' referenced by the typelib: libharfbuzz-gobject.so.0: cannot open shared object file: No such file or directory
````

even the first time when harfbuzz works normally. Every time I make install it resets harfbuzz and it works one more time. I tried doing the entire sequence starting with ./autogen.sh too but it doesn’t help things

kelvin13:
anyone know why this is happening?? Even on a basic test file it spit out zeros about every one in eight times

````
from os import environ

environ['GI_TYPELIB_PATH'] = 'HB/lib/girepository-1.0'
environ['LD_LIBRARY_PATH'] = 'libraries/HB/lib/girepository-1.0'
import gi
gi.require_version('HarfBuzz', '0.0')

from gi.repository import HarfBuzz as hb

from gi.repository.GLib import Bytes

path = 'amiri-regular.ttf'

text = 'AVefficiency Today'

with open(path, 'rb') as fi:
    fontdata = fi.read()

face = hb.face_create(hb.glib_blob_create(Bytes.new(fontdata)), 0)
font = hb.font_create(face)

upem = hb.face_get_upem(face)
hb.font_set_scale(font, upem, upem)
hb.ot_font_set_funcs(font)

buf = hb.buffer_create()
hb.buffer_add_utf8(buf, text.encode('utf-8'), 0, -1)
hb.buffer_guess_segment_properties(buf)

hb.shape(font, buf, [])

infos = hb.buffer_get_glyph_infos(buf)
positions = hb.buffer_get_glyph_positions(buf)

x = 0
y = 0
for I, P in zip(infos, positions):
    print(I.codepoint, x + P.x_offset, y + P.y_offset, I.cluster)
#    print(P.x_offset, P.y_offset)
    x += P.x_advance
    y += P.y_advance
````

output
````
** (process:3008): WARNING **: Failed to load shared library 'libharfbuzz-gobject.so.0' referenced by the typelib: libharfbuzz-gobject.so.0: cannot open shared object file: No such file or directory
0 0 0 0
0 500 0 1
0 1000 0 2
0 1500 0 3
[...]

PR #282, opened 2016-06-30
set the environment variables

kelvin13:
right now this script doesn’t work at all; also made the output a bit more readable

### DIFF
--- a/src/sample.py
+++ b/src/sample.py
@@ -2,21 +2,29 @@
 # -*- coding: utf-8 -*-
 
 from __future__ import print_function
+
+from os import environ
+environ['LD_LIBRARY_PATH'] = '/usr/local/lib/girepository-1.0'
+environ['GI_TYPELIB_PATH'] = '/usr/local/lib/girepository-1.0'
+
+from gi import require_version
+require_version('HarfBuzz', '0.0')
+
 import sys
 from gi.repository import HarfBuzz as hb
 from gi.repository import GLib
 
 # Python 2/3 compatibility
 try:
-	unicode
+    unicode
 except NameError:
-	unicode = str
+    unicode = str
 
 def tounicode(s, encoding='utf-8'):
-	if not isinstance(s, unicode):
-		return s.decode(encoding)
-	else:
-		return s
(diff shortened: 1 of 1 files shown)


src/sample.py
```diff
@@ -2,21 +2,29 @@
 # -*- coding: utf-8 -*-
 
 from __future__ import print_function
+
+from os import environ
+environ['LD_LIBRARY_PATH'] = '/usr/local/lib/girepository-1.0'
+environ['GI_TYPELIB_PATH'] = '/usr/local/lib/girepository-1.0'
```
khaledhosny: This is only needed if HarfBuzz is installed there, which is not always the case.

kelvin13: this is the default install directory, and in fact the *only* install directory that harfbuzz works in, see #274

behdad:
First, no mixing of two separate issues into one PR.

Second, env-vars don't belong here.  It's not "the _only_ install directory that harfbuzz works in".  I always use ~/.local for example.

kelvin13:
trying to install harfbuzz in a custom directory never worked for me, you get sporadic failures & `Failed to load shared library 'libharfbuzz-gobject.so.0' referenced by the typelib: libharfbuzz-gobject.so.0: cannot open shared object file: No such file or directory`

plus it should at least be worth it to put "`<your girepository folder location here>`" or something so people realize that’s why it’s not working out when they download this file

behdad:
> trying to install harfbuzz in a custom directory never worked for me, you get sporadic failures & Failed to load shared library 'libharfbuzz-gobject.so.0' referenced by the typelib: libharfbuzz-gobject.so.0: cannot open shared object file: No such file or directory

That's because you failed to set `LD_LIBRARY_PATH`, but that's basic Linux / Unix issue, not HarfBuzz-related.

kelvin13:
I did set LD_LIBRARY_PATH

khaledhosny:
Did you try to set the environment variable before calling python?

kelvin13:
what do you mean before? I thought you set the environment variable in the code

PR #284, opened 2016-07-01
fix OSAtomic calls for AArch64

robUx4:


behdad:
I don't understand why this is required.  What are you trying to fix?

PR #285, opened 2016-07-01
clang from MSVC needs to include intrin.h before winnt.h

robUx4:
I haven't found a better workaround. It we do this after `hb-private.hh` it doesn't work.


behdad:
But we never include `winnt.h`!  And it definitely doesn't belong before `hb-private.h`.

What is "clang from MSVC" anyway?

PR #286, opened 2016-07-01
[introspection] Fix hb_tag_from_string annotation

khaledhosny:


behdad:
Thanks!  There are more places like this that we need to annotate...

PR #288, opened 2016-07-06
[coretext] Speculative fix for CoreText nullptr access

ebraminio:
Merge back of a committed on Chromium patch: https://codereview.chromium.org/2118263002/

### DIFF
--- a/src/hb-coretext.cc
+++ b/src/hb-coretext.cc
@@ -164,21 +164,25 @@ create_ct_font (CGFontRef cg_font, CGFloat font_size)
        * system locations that we cannot access from the sandboxed renderer
        * process in Blink. This can be detected by the new file URL location
        * that the newly found font points to. */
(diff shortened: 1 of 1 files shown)


ebraminio:
Additionally, I put some spaces here and there in order to conform more with harfbuzz code style.

ebraminio:
Interesting that I can not use coretext backend without the patch on even simplest cases so it is not just speculative for me at least.

drott:
So the patch fixes this for you? Which Mac OS version, Sierra?

LGTM for merging this.

ebraminio:
Yes. Actually it is an El Capitan, 10.11.3.

ebraminio:
Do you think it would be better to remove some of you comments there? BTW, I've profiled and inspected more on memory usage of the backend and fixed [one of the issues](https://github.com/ebraminio/harfbuzz/commit/5462882e807107bcd4d7bf7844b46c7ef92c603a). Does the fix sounds right to you?

PR #292, opened 2016-07-12
[coretext] Turn all tabs to spaces

ebraminio:
Mix of tabs and spaces is highly annoying for editing and debugging the code, lets fix it once and forever.


behdad:
Can we go the other way please?  The rest of HarfBuzz uses tabs that are 8 spaces each.

ebraminio:
Oh I didn't know that, I've configured xcode and now I don't see much thing to do here. Sorry for the trouble, thank you.

PR #296, opened 2016-07-21
Don’t build UCDN support when building with GLib

khaledhosny:
Regression from: b424b6c372dfe4c0ed75a49761eb34a416819446.


behdad:
Ah, good catch.  Thanks.

Issue #287, opened 2016-07-02
Make harfbuzz give glyph output in a more performant format


kelvin13:
In horizontal typesetting, it is convenient to iterate through glyph output in the form `(cluster, codepoint, x_advance, x_offset)` or something similar. However harfbuzz returns this info in two lists of glyph objects, from `buffer_get_glyph_infos` and `hb.buffer_get_glyph_positions`. That means that, in python at least, code like 

```
glyph_iterator = ((N.cluster, N.codepoint, P.x_advance, P.x_offset) for N, P in zip(infos, positions))
```

is necessary to bridge that step. However my tests show that this generator actually takes about _8.1_ times longer than the shaping operation itself, including calling the two buffer functions that give glyph infos and glyph positions. That means that the python unpacking operation is responsible for almost **90%** of the total processing time. Further testing shows that most of this is due to the attribute lookups on `.cluster`, `.codepoint`, `.x_advance`, and `.x_offset`, not `zip()`.
This is quite shocking to me, and I think it would be _very, very_ helpful if harfbuzz could just build the glyphs in such a format to begin with so this bottleneck could be avoided.

PR #299, opened 2016-08-05
305% performance improvement with python bindings

kelvin13:
(Addresses #287)

There are probably many better ways of doing this than adding a field to `hb_glyph_position_t` but as a proof of concept, it is possible to achieve a roughly 300% increase in speed by having the glyph info/position lookups resolved in C rather than in Python through GObject.

````
def shape_new_buffer(extractor=lambda b: b):
    buf = hb.buffer_create ()
    hb.buffer_add_utf32 (buf, array.array('I', text.encode('utf-32')), 0, -1)
    hb.buffer_guess_segment_properties (buf)
    hb.shape (font, buf, [])
    return extractor(buf)

def unpack_hb_buffer1(HBB):
    return [(N.cluster, N.codepoint, P.x_advance, P.x_offset) for N, P in zip(hb.buffer_get_glyph_infos(HBB), hb.buffer_get_glyph_positions(HBB))]

def unpack_hb_buffer2(HBB):
    hb.buffer_compact_glyphs(HBB)
    return [P.compact for P in hb.buffer_get_glyph_positions(HBB)]

import timeit
t1 = timeit.timeit("shape_new_buffer(unpack_hb_buffer1)", number=1000, setup="from __main__ import shape_new_buffer, unpack_hb_buffer1")

t2 = timeit.timeit("shape_new_buffer(unpack_hb_buffer2)", number=1000, setup="from __main__ import shape_new_buffer, unpack_hb_buffer2")

print(t1)
print(t2)
print(t1/t2)

shaped_buffer = shape_new_buffer()

t3 = timeit.timeit("unpack_hb_buffer1(shaped_buffer)", number=1000, setup="from __main__ import shaped_buffer, unpack_hb_buffer1")

t4 = timeit.timeit("unpack_hb_buffer2(shaped_buffer)", number=1000, setup="from __main__ import shaped_buffer, unpack_hb_buffer2")

print(t3)
print(t4)
print(t3/t4)
````

````
0.5077813390016672
0.1881202860022313
2.6992375452567856 (overall stack performance improvement factor)
````
````
0.5054081650014268
0.16407813999830978
3.0802894584655407 (glyph extraction performance improvement factor)
````

This reduces the number of python attribute lookups required from 4 to 1 per glyph. Since sticking the values into the `hb_glyph_position_t` struct probably isn’t the best way to go, if there was a way to generate the array independently, testing shows it would probably achieve close to a 1000% percent performance improvement (since the bottleneck is still the one python attribute lookup).

### DIFF
--- a/src/hb-buffer-private.hh
+++ b/src/hb-buffer-private.hh
@@ -45,7 +45,7 @@
 #define HB_BUFFER_MAX_LEN_DEFAULT 0x3FFFFFFF /* Shaping more than a billion chars? Let us know! */
 #endif
 
-ASSERT_STATIC (sizeof (hb_glyph_info_t) == 20);
+ASSERT_STATIC (sizeof (hb_glyph_info_t) == 36);
 ASSERT_STATIC (sizeof (hb_glyph_info_t) == sizeof (hb_glyph_position_t));
 
 HB_MARK_AS_FLAG_T (hb_buffer_flags_t);
--- a/src/hb-buffer.cc
+++ b/src/hb-buffer.cc
@@ -1380,6 +1380,31 @@ hb_buffer_get_glyph_positions (hb_buffer_t  *buffer,
 }
 
 /**
+ * hb_buffer_compact_glyphs:
+ * @buffer: an #hb_buffer_t.
+ *
+ * Since: ???
+ **/
+void
+hb_buffer_compact_glyphs (hb_buffer_t * buffer)
+{
+    assert (buffer->have_positions);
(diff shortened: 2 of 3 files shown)


behdad:
You can convert the glyph positions and glyph infos into Python array.array objects and access those.  Let me make it absolutely clear: no change to the library is going to be accepted for this purpose.

kelvin13:
@behdad how to convert &lt;HarfBuzz.glyph_info_t object&gt; into python array?

PR #302, opened 2016-08-16
[travis] Fix MacOS libtool issue

khaledhosny:
Reveals another MacOS build breakage in the Core Text shaper.

PR #304, opened 2016-08-18
Fix the Travis build on MacOS X

brawer:


khaledhosny:
Is this different from #302?

brawer:
Oh, I didn't seen yours! Can we merge #302?

khaledhosny:
I was waiting for someone with an actual Mac to conform the Core Text fix.

brawer:
#302 looks better.

PR #112, opened 2015-06-02
Add Appveyor CI support

ebraminio:
The test suite is far from prefect on this setup (it will eventually work) but for now but the building and part that does matter more, compile of hb-uniscribe path, works. Lets [ping](https://sourceforge.net/p/msys2/tickets/163/) msys2 guys to see if is there any clean way to avoid hard coded msys2 base release date, of course it was much more nice if msys2 had [native support on appveyor](http://help.appveyor.com/discussions/suggestions/615-support-for-msys2) but this also seem good for now.

PR #308, opened 2016-09-01
[ci] Add Windows CI support, provided by AppVeyor

ebraminio:
The world is changed a lot since my last try for making this happen on #112, ragel added to msys2 ([by me](https://github.com/Alexpux/MINGW-packages/pull/645)), msys2 added on appveyor (I pushed this a bit also 😊) and since then both appveyor support for msys2 and msys2 itself perhaps are polished, @fanc999 added a nmake build project file for native win32 compiles, and I added DirectWrite backend to harfbuzz; and now it is the time for adding a deserving Windows CI to harfbuzz which compiles the project against MSVC compiler and gives us more confidence on harfbuzz development. And, [here](https://ci.appveyor.com/project/ebraminio/harfbuzz) is its result :)

Please take a look at it, thank you :)

### DIFF
--- a/README
+++ b/README
@@ -1,4 +1,5 @@
 [![Build Status](https://travis-ci.org/behdad/harfbuzz.svg)](https://travis-ci.org/behdad/harfbuzz)
+[![Build Status](https://ci.appveyor.com/api/projects/status/4oaq58ns2h0m2soa?svg=true)](https://ci.appveyor.com/project/behdad/harfbuzz)
 [![Coverage Status](https://img.shields.io/coveralls/behdad/harfbuzz.svg)](https://coveralls.io/r/behdad/harfbuzz)
 [ABI Tracker](http://abi-laboratory.pro/tracker/timeline/harfbuzz/)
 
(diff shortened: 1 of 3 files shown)


behdad:
Thanks Ebrahim!

Can we also run tests?  Also, I assume Uniscribe backend was enabled by default?  Maybe make it explicit doesn't hurt?

ebraminio:
AFAIK tests are not ready to use on nmake project files and Uniscribe backend is [https://github.com/behdad/harfbuzz/blob/master/win32/config.h.win32.in#L118](enabled by default) on nmake project files currently which is not super good idea IMO (e.g. even the fact dwrite backend now is much better than the other backends, mainly from memory consumption perspective, I don't like encourage its use at all).

ebraminio:
But I am going to see if can I enable tests on msys2 itself, better than nothing I'd guess :)

ebraminio:
> is much better than the other backends,

Hmm, of course I didn't mean "ot", which is highly superior on this area also :)

PR #313, opened 2016-09-07
Do reconfigure the cascade list for Emoji font on OS X 10.9.5

drott:
This seems to fix crash issues on 10.9.5 reported on Chrome, compare
crbug.com/549610

### DIFF
--- a/src/hb-coretext.cc
+++ b/src/hb-coretext.cc
@@ -149,9 +149,16 @@ create_ct_font (CGFontRef cg_font, CGFloat font_size)
   /* crbug.com/576941 and crbug.com/625902 and the investigation in the latter
    * bug indicate that the cascade list reconfiguration occasionally causes
    * crashes in CoreText on OS X 10.9, thus let's skip this step on older
-   * operating system versions. */
-  if (&CTGetCoreTextVersion != NULL && CTGetCoreTextVersion() <= kCTVersionNumber10_9)
-    return ct_font;
+   * operating system versions. Except for the emoji font, where _not_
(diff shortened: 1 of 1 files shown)


src/hb-coretext.cc
```diff
@@ -149,9 +149,16 @@ create_ct_font (CGFontRef cg_font, CGFloat font_size)
   /* crbug.com/576941 and crbug.com/625902 and the investigation in the latter
    * bug indicate that the cascade list reconfiguration occasionally causes
    * crashes in CoreText on OS X 10.9, thus let's skip this step on older
-   * operating system versions. */
-  if (&CTGetCoreTextVersion != NULL && CTGetCoreTextVersion() <= kCTVersionNumber10_9)
-    return ct_font;
+   * operating system versions. Except for the emoji font, where _not_
+   * reconfiguring the cascade list causes CoreText crashes. For details, see
+   * crbug.com/549610 */
+  if (&CTGetCoreTextVersion != NULL && CTGetCoreTextVersion() < kCTVersionNumber10_10) {
```
khaledhosny: The use of `kCTVersionNumber10_10` breaks compiling on older mac/xcode versions, see a0f1b44b6c0e3b1e6a1ba63dde2be22e0425353f.

behdad: Ugh.  I just made a release before checking email :(

khaledhosny: It is always good idea to wait for Travis to become green before merging/releasing 😃

behdad: Yeah, my bad :(.  @khaledhosny Can you send a PR please?

PR #315, opened 2016-09-09
[coretext] Use intended coretext version check logic

ebraminio:
Fix per [this](https://github.com/behdad/harfbuzz/commit/48677345281a93d8829dd37d0480a6062945416a#commitcomment-18945709) comment and the following.


behdad:
Thanks.

PR #333, opened 2016-09-28
Merge OpenType Variation Fonts support

behdad:
This is not complete yet.  'fvar' and 'avar' are missing.  But posting to request review.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -344,6 +344,10 @@ if test "x$with_freetype" = "xyes" -a "x$have_freetype" != "xtrue"; then
 fi
 if $have_freetype; then
 	AC_DEFINE(HAVE_FREETYPE, 1, [Have FreeType 2 library])
+	save_libs=$LIBS
+	LIBS="$LIBS $FREETYPE_LIBS"
+	AC_CHECK_FUNCS(FT_Get_Var_Blend_Coordinates)
+	LIBS=$save_libs
 fi
 AM_CONDITIONAL(HAVE_FREETYPE, $have_freetype)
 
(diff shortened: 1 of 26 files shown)


behdad:
@jfkthame can you review API changes please?

jfkthame:
@behdad I'm not seeing any API to query a font and find out what variation coordinate axes it supports? (And the corresponding user-scale min/default/max values.) Seems like that's probably desirable. Or did I just overlook it?

behdad:
> @behdad I'm not seeing any API to query a font and find out what variation coordinate axes it supports? (And the corresponding user-scale min/default/max values.) Seems like that's probably desirable. Or did I just overlook it?

Correct.  That's what I meant saying 'fvar' and 'avar' are not yet implemented.  Will add those.  Should I use float for those or 16.16?

jfkthame:
>  That's what I meant saying 'fvar' and 'avar' are not yet implemented

Oh, right -- I saw that when I first looked at this, but then had forgotten that comment by the time I came back here, sorry!

I think I'd lean towards exposing the values as 16.16, leaving it for clients to convert to float (or double, or whatever) if they prefer to use that.

behdad:
TODO items (not necessarily blocker for merge):

Implement 'fvar'/'avar' tables:
- [ ] Implement tables themselves.
- [ ] Add `hb_font_set_var_coords()` (user-space values).
- [ ] Add `hb_font_set_var_coord()` (singular) and `_coord_normalized()`.
- [ ] Possibly add getter version of `hb_font_set_var_*`.
- [ ] Add axis enumeration API.  Need to add name table API as well?

hb-ot-font:
- [ ] Implement HVAR/VVAR for glyph advance and possibly glyph metrics.
- [ ] Implement MVAR for font-extents.

Caching and performance tuning: TBD later.

PR #334, opened 2016-09-30
Support building with older glib versions

glandium:
It is desirable to be able to build against older versions of glib.

fd7a245 changed the configure check to require glib > 2.38 for
G_TEST_DIST. Before that, version 2.16 was required, but in fact,
since aafe395, G_PASTE is being used, which was introduced in 2.19.1.
And since 0ef179e2, hb-glib uses GBytes, which were introduced in
2.31.10.

2.19.1 is rather old, but 2.38 is rather new. For Firefox, building
against 2.22 is still supported, although we could probably get away
with bumping that to 2.28. Either way, GBytes is not available.

Arguably, if you build against a glib that doesn't support GBytes,
you're not going to use the hb_glib_blob_create function, so we hide
the function when building against such a glib.

As for G_TEST_DIST, when building against versions of glib that don't
support it, we can fallback to the previous behavior, which, AIUI, was
just making the test not work when building in a separate directory.


behdad:
We typically don't make library API dependent on versions of other libraries.  This is problematic when one uses harfbuzz compiled against older glib but compiling their code against newer glib.  But I'm happy to ignore that and merge this.  Thanks.

behdad:
Really merged this time.

PR #340, opened 2016-10-09
Make Uniscribe optional on nmake win32 builds

ebraminio:
Most of harfbuzz clients don't need uniscribe and directwrite support at all
as harfbuzz is a replacement for Uniscribe so Chromium and Firefox are using
harfbuzz without this on Windows.

This removal helps win32 nmake users which even don't use the backend to have
harfbuzz on a reduced binary size and libs count.

### DIFF
--- a/win32/config-msvc.mak
+++ b/win32/config-msvc.mak
@@ -20,6 +20,9 @@ CAIRO_LIB = cairo.lib
 # Graphite2 is needed for building SIL Graphite2 support
 GRAPHITE2_LIB = graphite2.lib
 
+# Uniscribe is needed for Uniscribe shaping support
+UNISCRIBE_LIB = usp10.lib gdi32.lib rpcrt4.lib user32.lib
+
 # Directwrite is needed for DirectWrite shaping support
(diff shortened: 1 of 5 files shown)


win32/config-msvc.mak
```diff
@@ -31,17 +34,15 @@ HB_UCDN_CFLAGS = /I..\src\hb-ucdn
...
 HB_HEADERS =	\
 	$(HB_BASE_headers)		\
 	$(HB_NODIST_headers)	\
-	$(HB_OT_headers)		\
-	$(HB_UNISCRIBE_headers)
+	$(HB_OT_headers)
 
 # Minimal set of (system) libraries needed for the HarfBuzz DLL
-HB_DEP_LIBS = usp10.lib gdi32.lib rpcrt4.lib user32.lib
+HB_DEP_LIBS = gdi32.lib user32.lib
```
ebraminio: I should test if we can reduce this even more.

behdad:
sgtm.

ebraminio:
So lets do it :)

PR #344, opened 2016-10-10
[ci] Whitelist more symbols and unbreak mac CI

ebraminio:


src/check-symbols.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+if [ "$TRAVIS_OS_NAME" == "osx" ]; then exit; fi
```
khaledhosny: I don’t think this is right fix, we need to find out why we are exporting private symbols (if this is the case) or find why the check is giving us a false positive.

khaledhosny:
BTW, the test passes fine on my local Mac build.

ebraminio:
Yes, I don't know why CI is failing like this, that's why I first tested something and disabling check-symbols.sh on Mac Travis, I can't also repro this locally.

ebraminio:
Lets close it till I (or someone else) get some clue on what is wrong.

src/check-symbols.sh
ebraminio: Absolutely.

khaledhosny:
Just add a line on the travis file to print the test suit log, it should give a clue what is going on.

ebraminio:
Good idea :)

ebraminio:
[Here](https://travis-ci.org/ebraminio/harfbuzz/jobs/166725226#L669) it is, any idea?

khaledhosny:
I think we need to update the list of ignored symbols, we already ignore `__gcov_flush` and `llvm_`, but the symbols shown here have an extra `_` prefix.

Something like:
```diff
diff --git a/src/check-symbols.sh b/src/check-symbols.sh
index b2bf43f..a8b472a 100755
--- a/src/check-symbols.sh
+++ b/src/check-symbols.sh
@@ -20,7 +20,7 @@ for suffix in so dylib; do
        so=.libs/libharfbuzz.$suffix
        if ! test -f "$so"; then continue; fi
        
-       EXPORTED_SYMBOLS="`nm "$so" | grep ' [BCDGINRSTVW] ' | grep -v ' _fini\>\| _init\>\| _fdata\>\| _ftext\>\| _fbss\>\| __bss_start\>\| __bss_start__\>\| __bss_end__\>\| _edata\>\| _end\>\| _bss_end__\>\| __end__\>\| __gcov_flush\>\| llvm_' | cut -d' ' -f3`"
+       EXPORTED_SYMBOLS="`nm "$so" | grep ' [BCDGINRSTVW] ' | grep -v ' _fini\>\| _init\>\| _fdata\>\| _ftext\>\| _fbss\>\| __bss_start\>\| __bss_start__\>\| __bss_end__\>\| _edata\>\| _end\>\| _bss_end__\>\| __end__\>\| __gcov_flush\>\| ___gcov_flush\>\| llvm_ \| _llvm_' | cut -d' ' -f3`"
 
        prefix=`basename "$so" | sed 's/libharfbuzz/hb/; s/-/_/g; s/[.].*//'`
 ````

File: src/check-symbols.sh [hunk 2]
```diff
@@ -20,7 +20,7 @@ for suffix in so dylib; do
 	so=.libs/libharfbuzz.$suffix
 	if ! test -f "$so"; then continue; fi
 	
-	EXPORTED_SYMBOLS="`nm "$so" | grep ' [BCDGINRSTVW] ' | grep -v ' _fini\>\| _init\>\| _fdata\>\| _ftext\>\| _fbss\>\| __bss_start\>\| __bss_start__\>\| __bss_end__\>\| _edata\>\| _end\>\| _bss_end__\>\| __end__\>\| __gcov_flush\>\| llvm_' | cut -d' ' -f3`"
+	EXPORTED_SYMBOLS="`nm "$so" | grep ' [BCDGINRSTVW] ' | grep -v ' _fini\>\| _init\>\| _fdata\>\| _ftext\>\| _fbss\>\| __bss_start\>\| __bss_start__\>\| __bss_end__\>\| _edata\>\| _end\>\| _bss_end__\>\| __end__\>\| __gcov_flush\>\| ___gcov_flush\>\| llvm_ \| _llvm_' | cut -d' ' -f3`"
```
khaledhosny: There is an extra space after `llvm_`. I think that is what is breaking Linux clang builds.

ebraminio: Done

ebraminio: > I think that is what is breaking Linux clang builds.

No problem Khaled, thanks for patient help :)

PR #349, opened 2016-10-18
[ci] Enable hb-freetype on Windows build tests

ebraminio:
This merges back @vlj's local patches of vcpkg for enabling [hb-freetype](https://github.com/Microsoft/vcpkg/pull/144) there. Lets apply them locally so they would be unnecessary when the next version of harfbuzz released.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,15 +4,19 @@ environment:
   matrix:
     - compiler: msvc
       ARCH: amd64
+      VCPKG_ARCH: x64-windows
       CFG: release
     - compiler: msvc
       ARCH: x86
+      VCPKG_ARCH: x86-windows
       CFG: release
     - compiler: msvc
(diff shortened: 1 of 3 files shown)


vlj:
thanks !

KrasnayaPloshchad:
Is this helpful to load font using FreeType?

behdad:
Feel free to fix and merge.  I don't review win32 stuff (except for hb-uniscribe)

ebraminio:
Thank you Behdad, for sure :)

ebraminio:
Behdad, can you see why appveyor ci is down? It seems it is gone since last github [dns] outage and needs some restart or something.

behdad:
There was a syntax error in the config file.  Should be fixed.

ebraminio:
Very sorry... Thank you

Issue #361, opened 2016-11-02
Consider moving to a CTFont based face constructor on Mac


drott:
In Chrome, we're creating a `CGFont` specifically for shaping, see `FontPlatformData::cgFont`, then internally, HarfBuzz creates a `CTFont` out of it again. This code path covers the dreaded `CTFontCreateWithGraphicsFont` mentioned in issue #291, so we might get this potential leak out of the way as well.

hb-coretext, IIRC, has two functions, one making tables available for the OpenType shaper, one, providing the fallback AAT shaper.

`hb_face_create` conceptually accepts an argument that does not contain a font size. A `CGFont` in that sense may be conceptually closer.

However, if we want to fix trak table support, issue #360, we might need to recreate a `CTFont` for each new font size shaping run, profiting less from caching a size independent face.

ebraminio:
I think we decided to not go on this path, if not, #367 simply can be revived.

PR #367, opened 2016-11-27
[coretext] Add CTFontRef constructor

### LINKED ISSUE #291: CoreText backend memory leak
CoreText backend as its use on Chrome on Mac is an important part of harfbuzz [at least till we get a native AAT implementation] so we should take seriously its possible bugs.

Continuing from [this not so good way to go patch](https://github.com/ebraminio/harfbuzz/commit/5462882e807107bcd4d7bf7844b46c7ef92c603), it seems hb-coretext suffers from memory leaks. As a proof, if you put `int t = CFGetRetainCount(data->ct_font); int g = CFGetRetainCount(data->cg_font);` in start of `_hb_coretext_shaper_face_data_destroy` you get two "3" which doesn't seem right.

My first finding is the fact `CTFontRef ct_font = CTFontCreateWithGraphicsFont (cg_font, font_size, NULL, NULL);` is adding a ref to cg_font but its release doesn't seem to cause cg_font ref count decrease and so it prevents cg_font eventual release (which is a serious issue). It seems this is a known [issue](http://stackoverflow.com/questions/37833225/how-can-i-stop-memory-being-leaked-by-ctfontcreatewithgraphicsfont) with coretext.

The main issue, however, seems lies into `_hb_coretext_shape` as if you put the same snippet, `int t = ...` on its third line and last line, you will get these pairs: (1, 2) and (3, 3). This doesn't seem right, I think logically `_hb_coretext_shape` should clean whatever reference it added to cg_font and ct_font.

ebraminio:
An attempt to fix #291 and #361, currently `hb_coretext_face_create`, `hb_coretext_face_create_from_ctfont` are broken, however the easy path works fine and with very low amount of memory leak.

image
(compare the second row to #291 screenshot with ~14mb leak)

### DIFF
--- a/src/hb-coretext.cc
+++ b/src/hb-coretext.cc
@@ -1,6 +1,7 @@
 /*
  * Copyright © 2012,2013  Mozilla Foundation.
  * Copyright © 2012,2013  Google, Inc.
+ * Copyright © 2016  Ebrahim Byagowi
  *
  *  This is part of HarfBuzz, a text shaping library.
  *
@@ -47,8 +48,8 @@ release_table_data (void *user_data)
 static hb_blob_t *
 reference_table  (hb_face_t *face HB_UNUSED, hb_tag_t tag, void *user_data)
 {
-  CGFontRef cg_font = reinterpret_cast<CGFontRef> (user_data);
-  CFDataRef cf_data = CGFontCopyTableForTag (cg_font, tag);
+  CTFontRef ct_font = reinterpret_cast<CTFontRef> (user_data);
+  CFDataRef cf_data = CTFontCopyTable (ct_font, tag, kCTFontTableOptionNoOptions);
   if (unlikely (!cf_data))
     return NULL;
 
@@ -62,10 +63,34 @@ reference_table  (hb_face_t *face HB_UNUSED, hb_tag_t tag, void *user_data)
 			 release_table_data);
 }
 
+static void
+hb_face_ct_font_destroyer (void *user_data)
+{
+  CTFontRef ct_font = reinterpret_cast<CTFontRef> (user_data);
+  CFRelease (ct_font);
+}
+
+/*
+ * Since: 0.9.27
+ */
 hb_face_t *
 hb_coretext_face_create (CGFontRef cg_font)
 {
-  return hb_face_create_for_tables (reference_table, CGFontRetain (cg_font), (hb_destroy_func_t) CGFontRelease);
+  return hb_face_create_for_tables (reference_table,
+                                    (void*) CTFontCreateWithGraphicsFont (cg_font, 36., 0, 0),
+                                    hb_face_ct_font_destroyer);
+
+}
+
+/*
+ * Since: 1.3.4
+ */
+hb_face_t *
+hb_coretext_face_create_from_ct_font (CTFontRef ct_font)
+{
+  return hb_face_create_for_tables (reference_table,
+                                    (void*) CFRetain (ct_font),
+                                    hb_face_ct_font_destroyer);
 }
 
 
@@ -100,49 +125,12 @@ get_last_resort_font_desc (void)
   return font_desc;
 }
 
-static void
-release_data (void *info, const void *data, size_t size)
-{
-  assert (hb_blob_get_length ((hb_blob_t *) info) == size &&
-          hb_blob_get_data ((hb_blob_t *) info, NULL) == data);
-
-  hb_blob_destroy ((hb_blob_t *) info);
-}
-
-static CGFontRef
-create_cg_font (hb_face_t *face)
-{
-  CGFontRef cg_font = NULL;
-  if (face->destroy == (hb_destroy_func_t) CGFontRelease)
-  {
-    cg_font = CGFontRetain ((CGFontRef) face->user_data);
-  }
-  else
-  {
-    hb_blob_t *blob = hb_face_reference_blob (face);
-    unsigned int blob_length;
-    const char *blob_data = hb_blob_get_data (blob, &blob_length);
-    if (unlikely (!blob_length))
-      DEBUG_MSG (CORETEXT, face, "Face has empty blob");
-
-    CGDataProviderRef provider = CGDataProviderCreateWithData (blob, blob_data, blob_length, &release_data);
-    if (likely (provider))
-    {
-      cg_font = CGFontCreateWithDataProvider (provider);
-      if (unlikely (!cg_font))
-	DEBUG_MSG (CORETEXT, face, "Face CGFontCreateWithDataProvider() failed");
-      CGDataProviderRelease (provider);
-    }
-  }
-  return cg_font;
-}
-
 static CTFontRef
-create_ct_font (CGFontRef cg_font, CGFloat font_size)
+create_ct_font (CTFontDescriptorRef ct_font_descriptor, CGFloat font_size)
 {
-  CTFontRef ct_font = CTFontCreateWithGraphicsFont (cg_font, font_size, NULL, NULL);
+  CTFontRef ct_font = CTFontCreateWithFontDescriptor (ct_font_descriptor, font_size, NULL);
   if (unlikely (!ct_font)) {
-    DEBUG_MSG (CORETEXT, cg_font, "Font CTFontCreateWithGraphicsFont() failed");
+    DEBUG_MSG (CORETEXT, ct_font_descriptor, "Font CTFontCreateWithFontDescriptor() failed");
     return NULL;
   }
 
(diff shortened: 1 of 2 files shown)


ebraminio:
@drott: Have a look if you got some time.

ebraminio:
Ready to review.

drott:
I had a discussion with Behdad on this, perhaps he can better summarize. 

As far I understood, a `hb_face` does not have a size and as such, the `CGFont` would be the better architectural match.  The difference is also the number of instantiations - for each `hb_font_t` we would then need to instantiate a new `CTFont`, instead of deriving the shaping results from the `CGFont` as right now.

Nevertheless, in general, I would be in favor of such a change, since we probably need different CTFont instances for fixing trak table support. Compare comments in https://bugs.chromium.org/p/chromium/issues/detail?id=627609#c30

The other important point is - we can most likely not retire the `CGFont` based constructor for backwards compatibility reasons - we would need to still provide both.

One question I had on your comments: "and with very low amount of memory leak." - what do you mean by "low amount"? We would need to get to a zero amount of leaks, wouldn't we?

ebraminio:
> The other important point is - we can most likely not retire the CGFont based constructor for backwards compatibility reasons - we would need to still provide both.

It still supports CGFontRef constructor, I just removed deprecated hints.

> One question I had on your comments: "and with very low amount of memory leak." - what do you mean by "low amount"? We would need to get to a zero amount of leaks, wouldn't we?

AFAIK remained ones are on shape process.

ebraminio:
> One question I had on your comments: "and with very low amount of memory leak." - what do you mean by "low amount"? We would need to get to a zero amount of leaks, wouldn't we?

I am getting constant amount of persistent memory on 100 and 1000 runs reports so I can say it doesn't have any leak, or at least, the ones I was experiencing before.

src/hb-coretext.cc
```diff
@@ -62,10 +63,32 @@ reference_table  (hb_face_t *face HB_UNUSED, hb_tag_t tag, void *user_data)
 			 release_table_data);
 }
 
+static void
+hb_face_from_ct_font_destroyer(void *user_data)
+{
+}
```
behdad: How come not destroying the reference here?

ebraminio: Fixed.

ebraminio: Just want to explain the cause and the fix, I was not doing "retain" somewhere else so I could free these on a different place but after your mention I put a retain on the custom constructors and trigged a ct_font "release" here.

ebraminio:
I've uploaded a [cl](https://codereview.chromium.org/2544283003/) of this on Chromium and it looks better with this patch: https://blink.gs/results/2544283003

ebraminio:
Before:
image
After:
image

ebraminio:
Any thought guys?

behdad:
With this, now we are creating hb-face from CTFont at arbitrary size (whatever was set on the CTFont) instead of the previously hardcoded 36pt.  That's why you see a difference.

FWIW, I talked to Ned the other day and asked if in any way they are deprecating CGFont.  He said no, and agrees with me that CGFont is the abstraction for the face data itself, and CTFont has size and other parameters.

In short, if this approach helps fix leaks we don't know how to fix otherwise, sure, I'll take it, but the original premise that CGFont is deprecated is simply not true.

ebraminio:
> With this, now we are creating hb-face from CTFont at arbitrary size (whatever was set on the CTFont) instead of the previously hardcoded 36pt. That's why you see a difference.

So that would be a good thing by its own as current spacing of Mac doesn't look good for me as an user and I guess that would help bugs related to handling "trak" table (or `mort` and `morx`?) on Mac.

> In short, if this approach helps fix leaks we don't know how to fix otherwise, sure, I'll take it, but the original premise that CGFont is deprecated is simply not true.

I can say at least it helps overall memory consumption, if not resolves leaks:

Default hb_face constructor (not  hb_coretext_face_create and hb_coretext_face_create_from_ct_font), (1000 iteration, before the change:

image

But after the change with exactly the same user code:

image

ebraminio:
hb_coretext_face_create, before the change:
image


hb_coretext_face_create, after the change:
image

hb_coretext_face_create_from_ct_font, introduced after the change:
image

ebraminio:
Also here is the (not clean but able to test all the cases) testing code I used for the analyzing, I am not sure if it itself is a correct code or my fix is in the right but I just tried to help the situation.

```C++
    const char *shapers[] = { "coretext", 0 };
    hb_buffer_t *hb_buffer = 0;
    hb_face_t *hb_face = 0;
    hb_font_t *hb_font = 0;
    CGFontRef cg_font = 0;
    CGDataProviderRef provider = 0;
    CTFontRef ct_font = 0;
    CFDataRef data = 0;
    CTFontDescriptorRef desc = 0;
    
    
    FILE *f = fopen("/Library/Fonts/Tahoma.ttf", "rb");
    fseek(f, 0, SEEK_END);
    long fsize = ftell(f);
    fseek(f, 0, SEEK_SET);
    
    char* font = (char*)malloc(fsize);
    fread(font, fsize, 1, f);
    fclose(f);
    
    
    for (int i = 0; i < 1000; ++i) {
        
        if (false) { // change this to true if you like test default constructors
            if (hb_font) {
                hb_font_destroy (hb_font);
            }
            hb_font = hb_ft_font_create (ft_face, NULL);
        } else {
            if (hb_buffer) {
                hb_buffer_destroy(hb_buffer);
                hb_face_destroy (hb_face);
                hb_font_destroy (hb_font);
                if (cg_font) {
                    CFRelease(cg_font);
                    CFRelease(provider);
                }
                if (ct_font) {
                    CFRelease(ct_font);
                    CFRelease(desc);
                    CFRelease(data);
                }
            }
            
            if (true) { // change this 
                data = CFDataCreateWithBytesNoCopy(NULL, (UInt8*)font, fsize, kCFAllocatorNull);
                desc = CTFontManagerCreateFontDescriptorFromData(data);
                ct_font = CTFontCreateWithFontDescriptor(desc, 36., NULL);
                hb_face = hb_coretext_face_create_from_ct_font(ct_font);
            } else {
                provider = CGDataProviderCreateWithData(NULL, font, fsize, NULL);
                cg_font = CGFontCreateWithDataProvider(provider);
                hb_face = hb_coretext_face_create(cg_font);
            }
            
            hb_font = hb_font_create(hb_face);
        }
        // Create hb-buffer and populate.
        hb_buffer = hb_buffer_create ();
        hb_buffer_add_utf8(hb_buffer, text, -1, 0, -1);
        hb_buffer_guess_segment_properties (hb_buffer);
        
        // Shape it!
        hb_shape_full (hb_font, hb_buffer, NULL, 0, shapers);
    }
```

drott:
> In short, if this approach helps fix leaks we don't know how to fix otherwise, sure, I'll take it, but the original premise that CGFont is deprecated is simply not true.

Got it. FWIW, the other positive aspect of having a CTFont based constructor is that we don't convert back and forth between CTFont->CGFont->CTFont when interfacing with Blink.

behdad:
> Got it. FWIW, the other positive aspect of having a CTFont based constructor is that we don't convert back and forth between CTFont->CGFont->CTFont when interfacing with Blink.

But we still have to go CTFont->new CTFont.  Anyway. What do you guys want me to do?  Merge this? Or try to fix it properly so we we create CTFont at proper size for each hb-font?

drott:
> But we still have to go CTFont->new CTFont. Anyway. What do you guys want me to do? Merge this? Or try to fix it properly so we we create CTFont at proper size for each hb-font?

IIUC, creating a font at the right size will be needed for trak support, right? So for proper optical sizing, we have to select the right variant of the San Francisco font, plus then have HarfBuzz' CoreText backend apply track at the output size, so I think yes, we want to fix this properly. 

How about we put this on the agenda for our February meeting? And in the meantime I'll try to find information about whether I can get screen dpi information in Blink.

ebraminio:
My really just 2cent, as [Chrome UI migration from MacView](http://crbug.com/603386) what would be more important is not CoreText proxying (this patch and hb-coretext) but having AAT on harfbuzz and if "Or try to fix it properly so we we create CTFont at proper size for each hb-font" would be in line it would be great. Also I think first step would be trying merge of [morx branch on fonttools](https://github.com/fonttools/fonttools/tree/morx) and trying to just parse morx table on harfbuzz afterward as the first step.

behdad:
> Also I think first step would be trying merge of morx branch on fonttools,

That branch is nowhere near complete.  But yeah, I like to get back to that.

ebraminio:
Turned out this can also cause functionality regression where Emoji characters width are getting better under this on LoDPI screen but getting broken on HiDPI so lets close it as that functionality improvements of this is not that obvious.

ebraminio:
Interesting FYI for both of Chrome and HarfBuzz, http://stackoverflow.com/a/43216750/1414809

PR #374, opened 2016-12-06
ICU_BUILTIN option for NMake build

elmarb:
I'm not sure when or why it would be sensible to build ICU support in a separate DLL on Windows. So I've left the old "ICU" build option, which did just that, alone, and added a new option "ICU_BUILTIN" for single-DLL builds.

Also fixed a few spelling mistakes, and commented a weird CFLAGS choice.


khaledhosny:
i think it should be safe to always build builtin ICU when it is enabled in Windows, unless @fanc999 has a different opinion.

elmarb:
If nobody objects to having ICU always builtin, I can make a pull request for that tomorrow.

fanc999:
Hi,

I think I am fine with this as well, so please go ahead.

With blessings, thank you!

khaledhosny:
Superseded by #375.

PR #377, opened 2016-12-14
Minor fixes to pacify Coverity code scan

simonis:
Hi Behdad,

we regularly run Coverity code scans on the OpenJDK sources and recently discovered two issues with HarfBuzz. While the discovered issues are not real errors, we think that fixing them my be nevertheless worthwile in order to increase the readability of the source code.

We just wanted to ask, if you are willing to accept these changes in the upstream HarfBuzz repository because only then it would make sense to also fix them in the OpenJDK copy of HarfBuzz.

The first issue found by Coverity is the last of the following four lines from src/hb-ot-font.cc:

    if (!subtable) subtable = cmap->find_subtable (0, 2);
    if (!subtable) subtable = cmap->find_subtable (0, 1);
    if (!subtable) subtable = cmap->find_subtable (0, 0);
    if (!subtable)(subtable = cmap->find_subtable (3, 0)) && (symbol = true);

From the whole context it really took me some time to understand that 'symbol' should only be set to true if 'subtable' is set from 'cmap->find_subtable (3, 0)'. Coverity reports an "assignment instead of compare" which is a false positive, but we think the could would be much more readable if changed to look as follows:

    if (!subtable)
    {
      subtable = cmap->find_subtable (3, 0);
      if (subtable) symbol = true;
    }

The second issue is related to the following definition in src/hb-ot-layout-gpos-table.hh:

 ```
 ValueFormat  valueFormat1;           /* Defines the types of data in
                                        * ValueRecord1--for the first glyph
                                        * in the pair--may be zero (0) */
  ValueFormat  valueFormat2;           /* Defines the types of data in
                                        * ValueRecord2--for the second glyph
                                        * in the pair--may be zero (0) */
```

Throughout hb-ot-layout-gpos-table.hh, '&valueFormat1' is used as if it were an array of two ValueFormat objects. While extremely unlikely, a compiler could theoretically insert padding between 'valueFormat1' and 'valueFormat2' which would make the code incorrect. We would therefore propose to simply change the previous definiton into a real array:

  ```
ValueFormat  valueFormat[2];         /* [0] Defines the types of data in
                                        * ValueRecord1--for the first glyph
                                        * in the pair--may be zero (0) */
                                       /* [1] Defines the types of data in
                                        * ValueRecord2--for the second glyph
                                        * in the pair--may be zero (0) */
```

and change the code which uses 'valueFormat' accordingly.

Thank youand best regards,
Volker

### DIFF
--- a/src/hb-ot-font.cc
+++ b/src/hb-ot-font.cc
@@ -355,7 +355,11 @@ struct hb_ot_face_cmap_accelerator_t
     if (!subtable) subtable = cmap->find_subtable (0, 2);
     if (!subtable) subtable = cmap->find_subtable (0, 1);
     if (!subtable) subtable = cmap->find_subtable (0, 0);
-    if (!subtable)(subtable = cmap->find_subtable (3, 0)) && (symbol = true);
+    if (!subtable)
+    {
+      subtable = cmap->find_subtable (3, 0);
+      if (subtable) symbol = true;
+    }
     /* Meh. */
     if (!subtable) subtable = &OT::Null(OT::CmapSubtable);
 
--- a/src/hb-ot-layout-gpos-table.hh
+++ b/src/hb-ot-layout-gpos-table.hh
@@ -689,7 +689,7 @@ struct PairPosFormat1
     (this+coverage).add_coverage (c->input);
     unsigned int count = pairSet.len;
     for (unsigned int i = 0; i < count; i++)
-      (this+pairSet[i]).collect_glyphs (c, &valueFormat1);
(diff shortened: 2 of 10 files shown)


behdad:
Dear Volker.  Thanks for the patch. Both are indeed very good suggestions.  I'll merge.

simonis:
Thanks a lot Behdad, it's always a pleasure contributing to your project :)

Best regards,
Volker

Behdad Esfahbod <notifications@github.com> schrieb am So., 18. Dez. 2016 um
07:12:

PR #391, opened 2017-01-09
Make UCDN default enabled (fixes #389)

### LINKED ISSUE #389: hb_ucdn_get_unicode_funcs is referenced but not compiled
v1.4.1, Windows, MSVC, compiling with glib/gobject support.

```nmake /f Makefile.vc FREETYPE=1 GOBJECT=1```

fails at link time with

```
hb-unicode.obj : error LNK2001: unresolved external symbol _hb_ucdn_get_unicode_funcs
```

`hb-ucdn.cc` is not getting compiled, so the error is justified.

Probably introduced by 53543f8b3e19456279c86f711feba42ce976668c

[Currently `HB_UCDN_sources` is only included in non-glib builds.](https://github.com/behdad/harfbuzz/blob/af596a5655ee2ed09da17a7de79f619e1547dcd9/win32/config-msvc.mak#L199) Perhaps you want to include it for all builds now.

ebraminio:
Lets see what CI thinks about this.


behdad:
Thanks Ebrahim!

PR #401, opened 2017-01-17
[indic] Add support for Grantha marks that may be used in Tamil to th…

### LINKED ISSUE #400: Use of Grantha marks within Tamil breaks shaping
A couple of marks that are encoded in the Grantha block may also be used in Tamil, as confirmed by ScriptExtensions.txt. However, because the Indic shaper doesn't know about them, they don't have the right Indic category assignments and will break shaping behavior.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1331339 for testcases, example font, etc. After fixing the script-run itemization issue discussed there, the Graphite version of the font shapes successfully, but OpenType still fails because of this issue.

jfkthame:
…e Indic table. Fixes #400.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1331339.

Testcase: U+0BA4,U+0BC6,U+1133c,U+0BAA,U+1133c,U+0BC6,U+1133c


behdad:
Shall we do a more systematic review of ScriptExtensions.txt to include anything that is sanctioned for use in Indic scripts?

behdad:
Also, can you add a test in tests/shaping?

jfkthame:
> Shall we do a more systematic review of ScriptExtensions.txt to include anything that is sanctioned for use in Indic scripts?

I did read through the whole file, and AFAICS these are the only two (Indic-related) characters in there that aren't already handled by the Indic table. (But if I missed something, please do add it!)

jfkthame:
@behdad OK to go ahead and merge this?

behdad:
@jfkthame Yes. Would be nice to have a test.

jfkthame:
@behdad see d024787dca62a6e665ee5e41cbab6bde94ddcbd8

behdad:
Thanks.

PR #407, opened 2017-01-26
Add Win7 version of himalaya.ttf to the GDEF table blacklist.

jfkthame:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1279693.


behdad:
No idea why bots are failing...

BTW, do you think we should record and check checksum of the GDEF table as well?

PR #412, opened 2017-02-02
Add Win10 Anniversary Update version of Tahoma to GDEF blacklist.

jfkthame:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1279925.


behdad:
Humm. Should we reach out to MS?

PR #413, opened 2017-02-08
hb-font: Fix a potentially undefined use of memcmp()

pwithnall:
While it’s fine to call memcmp(x, 0, 0) in practice, the C99 standard
explicitly says that this is not allowed: even if the length is zero,
the pointer arguments must be valid.

http://stackoverflow.com/a/16363034

Coverity ID: 141178

Signed-off-by: Philip Withnall <withnall@endlessm.com>

### DIFF
--- a/src/hb-font.cc
+++ b/src/hb-font.cc
@@ -1596,7 +1596,8 @@ _hb_font_adopt_var_coords_normalized (hb_font_t *font,
 				      unsigned int coords_length)
 {
   if (font->num_coords == coords_length &&
-      0 == memcmp (font->coords, coords, coords_length * sizeof (coords[0])))
+      (coords_length == 0 ||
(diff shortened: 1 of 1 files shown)


behdad:
Thanks. These quirks in the standards are really annoying. :)

behdad:
Humm. I have no idea why check-defs.sh is failing all in a sudden.

khaledhosny:
> Humm. I have no idea why check-defs.sh is failing all in a sudden.

See https://travis-ci.org/khaledhosny/harfbuzz/jobs/199785332#L1065

behdad:
> See https://travis-ci.org/khaledhosny/harfbuzz/jobs/199785332#L1065

Thanks. Let's see if I managed to fix it.

khaledhosny:
This fixed the Linux build, the Mac build seems to be failing for a different reason.

PR #419, opened 2017-02-16
Blacklist another instance of Padauk

drott:
In https://crbug.com/681813 another instance of Padauk was identified
triggering collapsed glyphs. Blacklist this version by running ttx from fonttools like
     ttx -l Padauk.ttf
then add the values those to the list of blacklisted table lengths.

### DIFF
--- a/src/hb-ot-layout.cc
+++ b/src/hb-ot-layout.cc
@@ -133,6 +133,9 @@ _hb_ot_layout_create (hb_face_t *face)
       || (1330 == gdef_len && 57938 == gpos_len && 109904 == gsub_len)
       /* 91fcc10cf15e012d27571e075b3b4dfe31754a8a padauk-3.0/Padauk-bookbold.ttf */
       || (1330 == gdef_len && 58972 == gpos_len && 109904 == gsub_len)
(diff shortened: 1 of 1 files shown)


drott:
CC @kojiishi

drott:
After patching `hb-ot-layout.cc` you can run  `utils/hb-view Padauk.ttf "test"` to observe the identified table length values.

jfkthame:
On 16/02/2017 11:23, Dominik Röttsches wrote:
> After patching |hb-ot-layout.cc| you can run |utils/hb-view Padauk.ttf
> "test"| to observe the identified table length values.

Rather than temporarily patching hb code, it may be simpler to use TTX 
(from fonttools) to list font table info:

     ttx -l Padauk.ttf

will report the lengths needed.

drott:
@jfkthame , I tried using ttx first but didn't know about the `-l` option, good to know, thanks!

jfkthame:
Is there a corresponding Bold font that should also be listed?

drott:
I don't know, @kojiishi, did you see one on the device?

kojiishi:
> I don't know, @kojiishi, did you see one on the device?

No, the device doesn't have bold, [fonts.xml](https://bugs.chromium.org/p/chromium/issues/attachmentText?aid=271414) from the device.

PR #421, opened 2017-02-16
Guard against underflow when adjusting length

jfkthame:
With the fuzz-testcase in mozilla bug 1295299, we end up with a recursed lookup that removes 3 items, when `match_positions[idx]` is 0, which results in (unsigned) `end` wrapping to a huge value.

Making `end` a signed int is probably the simplest route to a fix.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1295299.


behdad:
Excellent. Thanks for taking care of this. Can you drop the don't into the fuzzed.tests as well?

PR #422, opened 2017-02-17
[ci] Fix Windows CI

ebraminio:
Blind patch, lets see...


behdad:
lgtm if it fixes

ebraminio:
PTAL again, it seems vcpkg package installation is in trouble and is facing timeout, perhaps it is unable to fetch freetype tarball or is unstable for now. For more stable CI build I think it is better to disable vcpkg at least till [proper support from appveyor](https://github.com/appveyor/ci/issues/1076).

behdad:
lgtm whatever you decide.

PR #430, opened 2017-03-01
Fix Visual Studio build files for UCDN, and fix build on older Visual Studio

fanc999:
Hi,

Since now UCDN is the default unicode implementation in HarfBuzz, meaning that it is enabled for all builds unless explicitly disabled, the NMake Makefiles need to be adjusted to fix the build when ICU or GLib is enabled for the build,  and so HAVE_UCDN cannot be hardcoded to be 1 as one may still choose to disable it.

Also, in src/hb-common.cc, strtof() is used, but since that is only supported in Visual Studio 2013 and later, use strtod() in the earlier Visual Studio versions.

With blessingss, thank you!


src/hb-common.cc
```diff
@@ -667,7 +667,11 @@ parse_float (const char **pp, const char *end, float *pv)
   float v;
 
   errno = 0;
+#if defined (_MSC_VER) && (_MSC_VER < 1800)
+  v = strtod (p, &pend);
+#else
   v = strtof (p, &pend);
+#endif
   if (errno || p == pend)
```
khaledhosny: @behdad can we just use `strtod()` here instead of the ifdef?

khaledhosny:
@ebraminio any idea why msys2 build is suddenly unable to find glib? Seems not related to the changes here, though.

ebraminio:
No idea and no Windows around but if removing `--with-glib` from https://github.com/behdad/harfbuzz/blob/master/appveyor.yml#L49 will fix the CI I would be in favor of its removal.

fanc999:
Hi @ebraminio,

I think --with-glib is enabled so that the unit tests can be run on AppVeyor CI, as the unit tests depend on it.  What I did do on my own build is I had GLib enabled (as I normally do) and the C/C++ test programs did pass, sans the fuzzy ones that rely on *NIX shell scripts.

This would probably mean a AppVeryor configuration issue.

With blessings, thank you!

ebraminio:
> This would probably mean an AppVeryor configuration issue.

Couldn't you replicate the issue with installing msys2 locally and applying this? https://github.com/behdad/harfbuzz/blob/master/appveyor.yml#L48-L49

ebraminio:
If this is blocked on msys2 CI, I think disabling [it](https://github.com/behdad/harfbuzz/blob/master/appveyor.yml#L22-L29) is OK for now.

khaledhosny:
I’m just waiting for @behdad’s answer to my comment above, the msys2 issue is unrelated.

src/hb-common.cc
behdad: Sure, works for me.

fanc999:
Hi ebramnio,

Sorry, I don't think that it is convenient on my work system to install MSYS2 here.  I did note however that when AppVeyor was trying to install GLib, it became a NOP as the schema file was not found (probably a MSYS2 packman issue?)

Thanks for the feedback though.

With blessings, thank you!

PR #442, opened 2017-03-22
[ci] Fix msys2 build on AppVeyor

khaledhosny:
For whatever reason the env variables need to be the Windows way or they
end up being empty.


ebraminio:
Interesting. Thank you Khaled.

PR #444, opened 2017-03-25
Add CMake build support

ebraminio:
Just like freetype project, lets add a cmake project file on the root of the project.

Based on @pkoshevoy work on #41 but improved much and updated.

Lets see how this goes…


ebraminio:
No longer python script helper, all Makefile.sources reader logic went into the CMake itself.

ebraminio:
(the just added "Disallow in-source builds" bits are from freetype)

Now I believe this also can replace nmake build entirely also.

khaledhosny:
I think the cmake CI build should be done separately using build matrix or so, having to do each build twice is a waste of time. Or just don’t do it in CI builds at all, if it breaks people can just fix it, I don’t think Behdad is going to look into such build breakage anyway.

ebraminio:
@khaledhosny: Removed and limited to Windows, lets defer that.

ebraminio:
PTAL

behdad:
Preprocessing ragel files sounds good.  I don't know what to review.

Should we move `RAGEL_GENERATED` to Makefile.sources and read it from cmake as well?  Feel free to push changes to cmake and appveyor without my review.  I don't know them enough to comment.

ebraminio:
Great! So I will merge this and do that on a separate issue. Thanks

PR #445, opened 2017-03-26
Test fixes: fix record-test.sh for current fonttools, doc update

domschl:
This introduces changes to the test-tools and a documentation update:
* record-test.sh is incompatible with current (3.9.1) fonttools, the subset-font-file is now called font.subset.ttf (instead of font.ttf.subset), this updates for the new fonttools behavior.
* Documentation: record_it.sh is now called record_test.sh

(Former change of shebangs is reverted, see discussion below)

### DIFF
--- a/test/shaping/README.md
+++ b/test/shaping/README.md
@@ -14,7 +14,7 @@ FontTools from <https://github.com/behdad/fonttools>.
 
 To use `record-test.sh`, just put it right before the `hb-shape` invocation:
 ```sh
-$ ./hb-unicode-encode 41 42 43 627 | ./record-it.sh ../../util/hb-shape font.ttf
+$ ./hb-unicode-encode 41 42 43 627 | ./record-test.sh ../../util/hb-shape font.ttf
 ```
 what this does is:
   * Subset the font for the sequence of Unicode characters requested,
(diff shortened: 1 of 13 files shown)


khaledhosny:
> since the tools won't work with python3 on platforms that have python3 as default python.

They worked for me rather recently, what part did not work with Python 3, we should probably just fix it.

domschl:
Running the existing tests does work without any changes, however the test-recording tools do not. 
Example of a failure with python3:

`$ ./hb-unicode-encode F68 F71 F74 | ./record-test.sh ../../util/hb-shape DDC_Uchen.ttf`

`File "./hb-unicode-decode", line 5, in <module>
    UtilMains.filter_multiple_strings_or_stdin (Unicode.decode, "UNICODE_STRING")
  File ".../git/Tibetan/harfbuzz/test/shaping/hb_test_tools.py", line 432, in filter_multiple_strings_or_stdin
    print (callback (line))
  File ".../git/Tibetan/harfbuzz/test/shaping/hb_test_tools.py", line 448, in decode
    return u','.join ("U+%04X" % ord (u) for u in unicode (s, 'utf-8')).encode ('utf-8')
NameError: name 'unicode' is not defined`

It would be some work to fix this for both python2 and 3 due to differences in unicode handling between the two versions.

khaledhosny:
We are almost Python 3 compatible, let me fix the small remaining bit.

khaledhosny:
The Python 3 part should be fixed now.

domschl:
Hi, your changes work fine, and this is a much better solution, so I've undone the insertion of the shebangs. Is that the right approach, or should I generate a new pull request?

khaledhosny:
You can use interactive rebase `git rebase -i` to drop the unwanted commits, or you can create a new pull request. Whatever works for you.

domschl:
Git took already care of it. The patch doesn't affect the python-files anymore. Tx!

khaledhosny:
But there are still 5 commit when there should only be 2. Fixed it and pushed manually.

PR #443, opened 2017-03-25
Fix for reordering of Tibetan vowel u

domschl:
Following the discoveries of @eroux, discussion in #417, this minimal change to 2f560ee44b7b08683e8e6955fe3b95187e8bda32 seems to fix the rendering of long Tibetan vowel U, fixes #417.

Issue #417, opened 2017-02-11
Rendering for Tibetan long vowel U is broken


domschl:
On Linux and Android, the Tibetan syllable U (ཨཱུ) is rendered incorrectly.

The following site shows the browser-output when rendering hUM and U, together with bitmaps for the correctly rendered glyphs: [Test-site](http://digitaltibetan.org/index.php/Tibetan_Unicode_Test_Page#Harfbuzz_test)

Affected are:
* Android (tested with 7.1.2beta)
* Linux (tested with harfbuzz 1.4.2)
* macOS & Chrome: rendering seems ok within macOS fonts Kokonor and Kailasa, but breaks with all other Tibetan fonts tested.

KrasnayaPloshchad:
Confirmed on Windows 10. 

Firefox 51.0.1
![1](https://cloud.githubusercontent.com/assets/10364351/22862948/1829ad28-f173-11e6-8341-6606ed1ba142.PNG)
![2](https://cloud.githubusercontent.com/assets/10364351/22862946/13de5a98-f173-11e6-845b-5ec7032132e4.PNG)
BabelPad 9.0.0.3
![3](https://cloud.githubusercontent.com/assets/10364351/22863010/141e2492-f174-11e6-8a70-cd7513c1d968.PNG)
![4](https://cloud.githubusercontent.com/assets/10364351/22863012/14303902-f174-11e6-965b-7adfa3baa449.PNG)
LibreOffice 5.3
![5](https://cloud.githubusercontent.com/assets/10364351/22863013/14b95728-f174-11e6-974c-43a3b3df07bb.PNG)
![6](https://cloud.githubusercontent.com/assets/10364351/22863011/141eee90-f174-11e6-8044-3ef131eccbd1.PNG)

What I got surprised are: with Microsoft Himalaya the vowel signs are upside down, with Jomolhari one vowel sign failed to join. Since HarfBuzz is also intergrated in LibreOffice, this problem become more wide spread that need to pay more attentions to.

eroux:
With latest git, running

```
./hb-view --output-format=png --output-file=U.png --language=tibt --script=tibt /usr/share/fonts/truetype/noto/NotoSansTibetan-Regular.ttf ཨཱུ
```

gives

![u](https://cloud.githubusercontent.com/assets/60868/24321480/1e662568-114e-11e7-80fc-9a2a3b21d369.png)

(which is wrong)

eroux:
[...]

PR #446, opened 2017-03-26
Shaping tests for Tibetan vowels

domschl:
This implements tests for Tibetan vowels e, i, u, ei, au, I, U, aH, aM, oM.
Font used is [DDC_Uchen by Chris Fynn (Open Font License)](https://github.com/mozilla-b2g/moztt/tree/master/DDCUchen-1.000)
This test will currently break on U (ཨཱུ), if pull-request #443 is not applied. (Discussion in #417 )


eroux:
Thanks, that looks good! I think the initial change was to handle some of the weird abbreviations of http://www.babelstone.co.uk/Tibetan/Contractions.html although it's not clear what particular abbreviations were supposed to be handled... maybe you could add a few more things from these abbreviations in the tests? There are also some in http://www.dzongkha.gov.bt/publications/publication_pdf/A%20handbook%20of%20Dzongkha%20and%20Ch%C3%B6k%C3%A9%20abbreviations.pdf but I think only very few specific ones should be picked up

domschl:
I'll have a look and add more tests.
However, to my knowledge, the only application for supporting those shorthands would be to print a text about shorthands ;-) The shorthands were only used in dbu med hand-writings, and -- as far as I know -- are usually always expanded during type-setting.
Anyway, it's useful to find out how the initial change came about, so I will add more tests from those sources...

eroux:
Well, in Dzongkha these shorthands can be frequent, even in dbu can; `དྲགོས` for instance is really considered normal (it's a shorthand for `དྲགོས`)

domschl:
Ah, that I didn't know. 
Can I change contractions.txt? Sometimes, by re-ordering the vowel-signs, the standard shaper could work, e.g. ཀླུ་ཟློག་ can be written as kluog ཀློུག་ (looks ok) or  kloug ུཀློུག་. (looks bad)
Generally, in case of multiple vowels with u, using u first, makes it easier for the shaper.

eroux:
Yes, I suspect this was the original intent of the patch, I think we can probably modify it so that u is always used first, except when there is an achung (which would cause the bug to come back)

domschl:
Hi,
yes, that was most probably what happened.

I started to test with a different rewrite-strategy in hb-unicode-private.hh:
/* Tibetan */
#define HB_MODIFIED_COMBINING_CLASS_CCC129 129 /* sign aa */
#define HB_MODIFIED_COMBINING_CLASS_CCC130 132 /* sign i */
#define HB_MODIFIED_COMBINING_CLASS_CCC132 131 /* sign u */
#define HB_MODIFIED_COMBINING_CLASS_CCC138 133 /* sign e */
#define HB_MODIFIED_COMBINING_CLASS_CCC140 134 /* sign o */

This seems to make the standard vowels and shorthands happy.
I'll test some more...

eroux:
Thanks! interesting, why u before i?

domschl:
if you have ཀིུས་, it's important to shape ཀུ་ first, (because there's a special glyph in the font), and then add the kiku (which doesn't require a custom-glyph). The reordering-rule just tells the shaper, that the sequence kuis is fine. Without the rule, the shaper would not use the glyph for ku, and the result would look smashed.

So basically, it seems the original idea of moving the u-vowel earlier was just the right idea, they just moved it one place too far (in front of achung).

This now just needs *lots* of testing.

eroux:
Oh ok, I was just confused by the fact that the line about i is before the line about u, but the numbers are in the correct order, sorry! I think it should be fine this way! There are indeed many tests to do (although I think you should be able to make all the permutations of all vowels programmatically quite simply), but you're definitely on right tracks, thanks a lot!

domschl:
I checked the output for the 612 entries in contractions.txt, and with:

```
/* Tibetan
 * 
 * In case of multiple vowel-signs, use u first (but after achung) 
 * this allows Dzongkha multi-vowel shortcuts to render correctly 
 */
#define HB_MODIFIED_COMBINING_CLASS_CCC129 129 /* sign aa */
#define HB_MODIFIED_COMBINING_CLASS_CCC130 132 /* sign i */
#define HB_MODIFIED_COMBINING_CLASS_CCC132 131 /* sign u */
#define HB_MODIFIED_COMBINING_CLASS_CCC138 138 /* sign e */
#define HB_MODIFIED_COMBINING_CLASS_CCC140 140 /* sign o */
```

609 Testcases are ok with DDC_Uchen. The missing 3 would require patching DDC_Uchen, because of missing glyphs that cannot be compensated by the shaper (e.g.: rdz+ra) . I am unsure on how many testcases we should add to the automatic tests (I started just adding all of them, when noticing that the other languages just use a few tests)

eroux:
great! let's wait for some input from @behdad or @khaledhosny and see what kind of tests are necessary... I think the minimum would be the U stack and also something like kui... but for me your code is ok, thanks a lot!

domschl:
Ok, I've added a few "best-of-contractions" tests for the multiple-vowel cases that where broken before the new-reordering patch. That way we can check that the "conventional" Tibetan with all vowels work, and that we support the contraction-shortcuts. I guess that's it so far, and let's wait for feed-back.

Reminder: the test for tibetan-vowels.tests will break on checking long-u (ཨཱུ), until #443 is applied.

KrasnayaPloshchad:
I found the following resource is suitable for test:
http://yalasoo.com/English/docs/yalasoo_en_font.html

KrasnayaPloshchad:
> The missing 3 would require patching DDC_Uchen, because of missing glyphs that cannot be compensated by the shaper (e.g.: rdz+ra) .

I suggest you can also try to test with this font:
[himalaya.zip](https://github.com/behdad/harfbuzz/files/871315/himalaya.zip)

domschl:
Good news: MS Himalaya renders all vowel- and contractions-test-cases correctly with the new shaper patch. 
The only cosmetic issue I could find is a slightly irregular rendering of a triple-drengbu vowel-sign, that's sometimes uses as a shortcut for repetitions of three syllables. The three vowel-signs are not quite equidistant:
image
All other testcases (>600) are flawless.

behdad:
Adding all is good if the resulting font is not too big.  Thanks!

behdad:
> I suggest you can also try to test with this font:
> himalaya.zip

Don't post fonts you don't have license to distribute.  I'm going to delete your comment.

domschl:
All 600 test-cases require a subset-font that is 384kB (the original complete DDC_Uchen is 861kB).

Would that be too large? If yes, what would be a reasonable upper limit? So far, the largest test-font in the repo is 74kB.

I guess it is possible to make the Tibetan contractions test-set smaller (and less redundant), depending on how big test-fonts-files we want to allow.

domschl:
I've extended the test-cases to cover all 113 cases from texts/in-tree/shaper-tibetan/script-tibetan/misc/contractions.txt that actually use multiple vowel-signs.
Resulting tests have been split into two files to work around size-limitations of hb-view.
The test-font files are about 120kb, 100kb which is quite large.

PR #458, opened 2017-04-06
Re-write make rule to not use absolute paths

attente:
https://github.com/behdad/harfbuzz/issues/455


behdad:
I explained why I generate them in srcdir:

https://bugzilla.gnome.org/show_bug.cgi?id=780163#c2

I'll try to fix the bug in a different way, though looks like might need to use GNU make substitutions.

attente:
I originally had something that used the srcdir explicitly, would that work better?

https://github.com/behdad/harfbuzz/pull/467

Issue #345, opened 2016-10-10
'OSMemoryBarrier' is deprecated in macOS 10.12


khaledhosny:
Building with latest XCode I got lots of these warnings:

```
warning: 'OSMemoryBarrier' is deprecated: first deprecated in macOS 10.12 - Use std::atomic_thread_fence() from <atomic> instead [-Wdeprecated-declarations]
warning: 'OSAtomicAdd32Barrier' is deprecated: first deprecated in macOS 10.12 - Use std::atomic_fetch_add() from <atomic> instead [-Wdeprecated-declarations]
warning: 'OSAtomicCompareAndSwap64Barrier' is deprecated: first deprecated in macOS 10.12 - Use std::atomic_compare_exchange_strong() from <atomic> instead [-Wdeprecated-declarations]
```

behdad:
Feel free to patch conditionally.  Thanks.

behdad:
Maybe we should support C++11 and use atomic on all platforms?

khaledhosny:
But this would make us dependent on the C++ standard library on all platforms. C11 has a similar C API, may be we can use it if supported by platform/compiler and fallback to existing code otherwise?

behdad:
> But this would make us dependent on the C++ standard library on all platforms.

Right.  Not sure if that's still a concern.

>  C11 has a similar C API, may be we can use it if supported by platform/compiler and fallback to existing code otherwise?

Whatever works.  I do prefer not changing this much as it seems to have been working.

behdad:
What's suggested work item here?

khaledhosny:
Someone familiar with the new C/C++ atomics adds support for them. It should be e.g. used by default when supported and else fall back to the existing implementations. I might give it a try at some point, unless some one beat me to it.

behdad:
There's an alternative approach in https://bugs.freedesktop.org/show_bug.cgi?id=90305

Let's give that one a try?  I'll find some time to look into https://github.com/behdad/harfbuzz/pull/486 as well.

behdad:
Also, to make TSan happy, we should replace `get_unsafe` to do an atomic fetch with `memory_order_relaxed` ordering instead of being a non-atomic read.

waywardmonkeys:
My changes in #676 remove this warning by using the compiler primitives rather than the macOS APIs.

behdad:
Let's see how the bots like my implementation.

PR #462, opened 2017-04-12
[cmake] Remove NO_MT flag

ebraminio:
Is this logic sounds right to you or it should be better be like "if (HB_HAVE_GLIB OR HB_HAVE_ICU)" and maybe HB_HAVE_CORETEXT and other OS provided APIs also?

(With having this in mind that the goal of this cmake is I think to be a simple but correct (thats why I removed -DHB_NO_UNICODE_FUNCS) and easily embeddable harfbuzz distribution that maybe replaces third party distributions like [these](https://www.google.com/search?q=cmake+harfbuzz), and our internal nmake win32 port also entirely someday, if @fanc999 agrees also, other than my personal goal that is making harfbuzz development itself easier with OS provided Xcode or VS tools and extra APIs/SDKs.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,7 +38,6 @@ include_directories(AFTER
 
 # feel free to discuss these or more with the maintainers
 add_definitions(-DHAVE_OT)
-add_definitions(-DHB_NO_MT)
 add_definitions(-DHB_DISABLE_DEPRECATED)
 
 if (BUILD_SHARED_LIBS)


ebraminio:
Also any other comment on the naming and flags would be welcomed, specially before harfbuzz release so we can fix it before people rely on them.

ebraminio:
(I rethought of this flag just when I saw [this](https://github.com/tamaskenez/harfbuzz-cmake/blob/master/CMakeLists.txt) (and invited [the maintainer](https://github.com/tamaskenez/harfbuzz-cmake/issues/4)) that tries to port some of atomic operations to bsd/solaris derivatives that can be avoided if only thing one need is having a simple working harfbuzz)

behdad:
Humm? Why?

Please don't make it easy to build with `NO_MT`. Definitely don't make it default in any situation.  Those kinds of situations are why I was afraid of adding an alternate build system.

behdad:
IMO don't add an option for `NO_MT`.  It exists in the code for the likes of Firefox and Android who only use HB from one thread.

ebraminio:
No problem, removed.

> Those kinds of situations are why I was afraid of adding an alternate build system.

Better to let third parties doing that I believe when can be done somewhere right, so quick question, do you see anything [here](https://github.com/tamaskenez/harfbuzz-cmake/blob/master/CMakeLists.txt#L130-L150) worth to be merged back into harfbuzz itself? I don't know how they work and why they are there honestly.

behdad:
> do you see anything here worth to be merged back into harfbuzz itself? I don't know how they work and why they are there honestly.

They are ported from HB's configure.ac. They are feature-detection code to detect what atomic ops the compiler supports...

behdad:
I've got to say. I find the cmake as unreadable as configure.ac :). Guess it's more portable though.

ebraminio:
Now without NO_MT I faced #345, hope to be fixed soon.

> They are feature-detection code to detect what atomic ops the compiler supports...

So, I have to study them.

> I find the cmake as unreadable as configure.ac :).

Agreed :) Though now I am accustomed to it even the fact I learned and use it just time to time for harfbuzz.

ebraminio:
(And just to make directwrite backend happen for kashida thing)

PR #470, opened 2017-04-17
[cmake] Remove HB_DISABLE_DEPRECATED as it seems needed for pango build

ebraminio:
See https://github.com/Microsoft/vcpkg/pull/943#issuecomment-294447460


behdad:
Yes, thanks.  Deprecated symbols are still part of the ABI.  Though, we should fix Pango.

behdad:
> we should fix Pango.

Done.

Issue #471, opened 2017-04-18
clang -Wcomma compiler warnings


cpeterso:
clang's new -Wcomma compiler option warns about possible misuse of the comma operator such as between two statements or to call a function for side effects within an expression. When enabling -Wcomma and compiling Firefox, I see the following warnings. Looking at the harfbuzz code, these warnings do not point to real bugs, but the code is a little confusing because it is using the comma operator to squeeze two statements into one line of code.

```
src/hb-common.cc:190:9 [-Wcomma] possible misuse of comma operator here
src/hb-ot-layout-gsubgpos-private.hh:345:30 [-Wcomma] possible misuse of comma operator here
src/hb-shape-plan.cc:438:26 [-Wcomma] possible misuse of comma operator here
```

behdad:
Fixed already.

PR #472, opened 2017-04-18
Fix clang -Wcomma warnings (#471)

cpeterso:
clang's new -Wcomma compiler option warns about possible misuse of the
comma operator such as between two statements.

hb-common.cc:190:9 [-Wcomma] possible misuse of comma operator here
hb-ot-layout-gsubgpos-private.hh:345:30 [-Wcomma] possible misuse of
comma operator here
hb-shape-plan.cc:438:26 [-Wcomma] possible misuse of comma operator here


behdad:
Thanks.

PR #475, opened 2017-04-20
Graphite2 bug fix for rtl

mhosken:
This PR fixes a bug in converting data from graphite structures to harfbuzz structures for rtl segments.


behdad:
Maybe we should add tests as well?  We currently don't have conditional tests, but shouldn't be hard to add those in the Makefile.am

PR #484, opened 2017-05-09
[cmake] Add framework build support

ebraminio:


ebraminio:
The framework build works on my tests but if someone is more familiar with framework build practices, or can use it in some project, please put your comment here.

PR #487, opened 2017-05-16
Do not skip TAG characters in glyph substitution

khaledhosny:
Hide them like Mongolian Free Variation Selectors instead.

Fixes https://github.com/behdad/harfbuzz/issues/463


src/hb-ot-layout-private.hh
```diff
@@ -274,6 +275,10 @@ _hb_glyph_info_set_unicode_props (hb_glyph_info_t *info, hb_buffer_t *buffer)
        * https://github.com/behdad/harfbuzz/issues/234
        */
       if (unlikely (hb_in_range (u, 0x180Bu, 0x180Du))) props |= UPROPS_MASK_FVS;
+      /* TAG cgaracters used in emoji flags:
```
jfkthame: typo, "cgaracters"

src/hb-ot-layout-private.hh
```diff
@@ -375,7 +380,7 @@ _hb_glyph_info_is_default_ignorable (const hb_glyph_info_t *info)
 static inline hb_bool_t
 _hb_glyph_info_is_default_ignorable_and_not_fvs (const hb_glyph_info_t *info)
```
jfkthame: Should we update the name of this function to add "_or_tag"? Or maybe use an alternative such as _hb_glyph_info_is_ignorable_for_skipping_iterator?

src/hb-ot-layout-private.hh
```diff
@@ -241,6 +241,7 @@ enum hb_unicode_props_flags_t {
   UPROPS_MASK_GEN_CAT	= 0x001Fu,
   UPROPS_MASK_IGNORABLE	= 0x0020u,
   UPROPS_MASK_FVS	= 0x0040u, /* MONGOLIAN FREE VARIATION SELECTOR 1..3 */
+  UPROPS_MASK_TAG       = 0x0080u, /* TAG characters */
```
jfkthame: AFAICS, this doesn't really need to be a separate bit, we could just rename FVS to FVS_OR_TAG. I guess it's up to Behdad whether he prefers to keep them separate for clarity, or merge to leave the other bit free for some future use.

behdad: Yeah, I was thinking of changing the FVS to something like HIDDEN instead.  Let's keep the 0x0080 flag for future.

khaledhosny: I was split about this, couldn’t come up with a good name 😄. Reused and renamed the FVS flag.

src/hb-ot-layout-private.hh
khaledhosny: Fixed.

behdad:
Thanks!

khaledhosny:
Should I merge this now?

behdad:
> Should I merge this now?

I thought I did.  Doing now.

PR #488, opened 2017-05-22
Avoid using strdup inside library.

sebras:
If an application provides a malloc replacement through
hb_malloc_impl() it is important that it is used to allocate
everything, but the use of strdup() circumvents this and
causes system malloc() to be called instead. This pairs
badly with the custom hb_free_impl() being called later.

### DIFF
--- a/src/hb-common.cc
+++ b/src/hb-common.cc
@@ -221,7 +221,13 @@ struct hb_language_item_t {
   }
 
   inline hb_language_item_t & operator = (const char *s) {
-    lang = (hb_language_t) strdup (s);
+    /* If a custom allocated is used calling strdup() pairs
+    badly with a call to the custom free() in finish() below.
+    Therefore don't call strdup(), implement its behavior.
+    */
+    size_t len = strlen(s) + 1;
+    lang = (hb_language_t) malloc(len);
+    memcpy((unsigned char *) lang, s, len);
     for (unsigned char *p = (unsigned char *) lang; *p; p++)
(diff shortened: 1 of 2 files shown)


behdad:
Thanks. Makes sense. Care to add a comment at the site of use, as well as not call strlen twice?

behdad:
We should add a check for this as well.  I can do that part later.

sebras:
I attempted to address your comments, let me know if I missed anything.

src/hb-common.cc
```diff
@@ -221,7 +221,13 @@ struct hb_language_item_t {
   }
 
   inline hb_language_item_t & operator = (const char *s) {
-    lang = (hb_language_t) strdup (s);
+    /* If a custom allocated is used calling strdup() pairs
+    badly with a call to the custom free() in finish() below.
+    Therefore don't call strdup(), implement its behavior.
+    */
+    int len = strlen(s) + 1;
```
maksqwe: Please, use size_t instead of int, because strlen() returns size_t result type
and memcpy() accepts size with size_t type.

sebras: D'oh, I should have noticed that myself. Fixed, sorry for the noise.

behdad:
Thanks. You made me notice that we do not handle allocation failure here.  I can fix that separately.

behdad:
> Thanks. You made me notice that we do not handle allocation failure here. I can fix that separately.

Done.

PR #491, opened 2017-06-01
Support branch prediction helpers on clang compiles

ebraminio:
And drop GCC 2 support (is that still relevant?)

I was [reading](https://chadaustin.me/2017/05/writing-a-really-really-fast-json-parser/) about [sajson](https://github.com/chadaustin/sajson/blob/3cb7ba16ea6ff9184878f6b77f2899d95c26d270/include/sajson.h#L39) and noticed this seems missing on harfbuzz, hope would be useful.

### DIFF
--- a/src/hb-private.hh
+++ b/src/hb-private.hh
@@ -74,10 +74,9 @@ extern "C" void  hb_free_impl(void *ptr);
 /* Compiler attributes */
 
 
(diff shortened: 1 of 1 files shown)


ebraminio:
Also do you see use of "!!(x)" instead ternary operator needed/cool?

behdad:
I like "!!(x)".

Any other place we are not handling clang?

ebraminio:
> Any other place we are not handling clang?

Not to my knowledge.

Issue #479, opened 2017-04-28
Remove win32/ folder and direct nmake support


ebraminio:
First of all I should mention that all @fanc999 works on harfbuzz project is highly appreciated on providing nmake and that was very helpful on making cmake port also happen (namely and used on CMake port items, the creation and separation of Makefile.sources and annotating harfbuzz API with HB_EXPORT, and probably much more I am just missing).

As just a suggestion, now that we have cmake port in and that is able to generate nmake build files also, I like to just propose removal of nmake support and win32 folder entirely from harfbuzz project. I think cmake is just easier to use for Windows people as its better integration with Visual Studio (also even better with [native vs support](https://blogs.msdn.microsoft.com/vcblog/2016/10/05/cmake-support-in-visual-studio/)) and this will reduce maintenance load of harfbuzz project and less confusion of people want to use harfbuzz on their Windows projects.

I should mention however gir/gobject implementation of the cmake port is just not done yet and I probably wouldn't able to reach to it as lack of experience on it, but I am also not sure how much that would be needed for third party projects that was provided on nmake port, and there are possibly more needed tweaks for our just new and early cmake to make it really useful for other projects which probably @fanc999 knows better than me so there is no rush on this and there is also a chance this will be just closed if we conclude removing nmake is not a good move for harfbuzz project. Thanks all.

KrasnayaPloshchad:
👍 I agree with you, to make the build easier on Windows is valuable to be considered. Althrough MS Windows has Uniscribe and DirectWrite to give native support for OT text layout, several Windows developers has expected to intergrate HarfBuzz to get their own support, or free from some technical limits, if we can make some efforts to reduce their fears, we would get great achievement, also get more support from Windows developers whose want to point out what’s wrong with us.

ebraminio:
I can say CMake support is complete except the gir/gobject part, which I am not sure who is interested on it. Have a look at this example, https://github.com/ebraminio/kashida which uses DirectWrite backend of HarfBuzz with a CMake based build system.

behdad:
What's the status of this?

ebraminio:
I suggest just removing win32 folder and redirect all Windows users to https://github.com/Microsoft/vcpkg which is provided by Microsoft so no random Windows compile issues anymore.

behdad:
sgtm.

PR #498, opened 2017-06-19
Win32/NMake builds: Support builds from GIT

fanc999:
Hi,

To ease development on Windows using NMake, I thought it might be a good idea to support the following:

1. Allow one to generate win32\config.h.win32 and src\hb-version.h, which are distributed in a release
   tarball, from a GIT checkout.
2. Allow one to generate (and re-generate) the Ragel-generated .hh headers during the NMake build 
  process.

This pull request attempts to address these issues by:
-Adding Python 2.7.x+/3.x scripts to generate win32\config.h.win32 and src\hb-version.h, which is not
 distributed in a release tarball, for (1).  Note that replace.py and pc_base.py are common utility Python
 scripts that are the same with the ones in GLib, and will also be re-used at a later stage for generating
 the HarfBuzz pkg-config files, as Meson becomes prevalent in GNOME.
-Add rules to use Ragel to generate the .hh headers (and to clean and regenerate them), for (2).

With blessings, thank you!

With blessings, thank you!

### DIFF
--- a/win32/README.txt
+++ b/win32/README.txt
@@ -11,7 +11,22 @@ backends are enabled, and this is the base configuration that is built if no
 options (see below) are specified.  A 'clean' target is provided-it is recommended
 that one cleans the build and redo the build if any configuration option changed.
 An 'install' target is also provided to copy the built items in their appropriate
-locations under $(PREFIX), which is described below.
+locations under $(PREFIX), which is described below.  A 'reallyclean' target is
+provided that not only does what is done for the 'clean' target, but also removes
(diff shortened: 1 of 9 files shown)


behdad:
lgtm. @ebraminio ?

ebraminio:
i wanted to go for #479 in the future but as our cmake doesn't support gobject introspection yet, why not, lets go for this also

ebraminio:
Thank you very much for you contribution @fanc999 😊

PR #508, opened 2017-07-14
[coretext] Allow to disable kern

anthrotype:
This partly fixes https://github.com/behdad/harfbuzz/issues/338.

Note that after kerning has been disabled (i.e. by setting `kCTKernAttributeName` to zero), CoreText doesn't let us enable it again for subsequent ranges of characters in the same attributed string.

https://developer.apple.com/documentation/coretext/kctkernattributename

> If this attribute is not present, standard kerning is used.

Removing the `kCTKernAttributeName` attribute does not seem to do anything: kerning stays disabled.


anthrotype:
cc @nedley

nedley:
Will get this fixed, thanks.

PR #529, opened 2017-09-01
Add release checklist

ebraminio:
Based on emails we had on the subject, MINGW's scripts should be polished maybe (or moved to the repository it self? why not?) and the whole thing should be tested by multiple people I believe.


ebraminio:
WDYT?

ebraminio:
@behdad: Any thought?

PR #562, opened 2017-10-11
Remove nmake port of harfbuzz

### LINKED ISSUE #479: Remove win32/ folder and direct nmake support

ebraminio:
Closes #479 

and as suggested there by Behdad.

This encourages use of cmake but our cmake port lacks gir support for now so that should be considered.

As a note for Windows users, please consider just using https://github.com/Microsoft/vcpkg which builds harfbuzz for you among many other libraries and don't get involve with the detail of building harfbuzz thus asking support here.


ebraminio:
@Behdad, I guess you should consider bumping a minor version at least for the removal.

ebraminio:
We should also put some notice to discourage Windows build bugs something like what is done [here](https://github.com/Homebrew/brew/issues/new) by [this](https://github.com/Homebrew/brew/tree/master/.github) folder.

khaledhosny:
What about GTK Windows builds that I think depend on this, can they use vcpkg?

ebraminio:
> What about GTK Windows builds that I think depend on this, can they use vcpkg?

They should use cmake directly of course but I guess gir is also needed, so no idea honestly.

PR #564, opened 2017-10-11
CMakelists.txt fix for Freetype builds

khaledhosny:
Instead of searching for freetype using pkg-config, use the FindFreetype
feature of CMake. This allows for better integration with other projects
that make use of CMake.

Fixes https://github.com/behdad/harfbuzz/issues/518


ebraminio:
Thanks Khaled.

PR #568, opened 2017-10-16
[ot-font] Implement reading glyph names

khaledhosny:
Turns out we already have support for “post” table, it just needed to be activated and put in use.

Might not be the best way to handle this, as I don’t really think I totally understand how parsing tables in HarfBuzz works.

### DIFF
--- a/src/hb-ot-font.cc
+++ b/src/hb-ot-font.cc
@@ -327,6 +327,15 @@ struct hb_ot_face_post_accelerator_t
 
     return this->post->get_glyph_name (glyph, name, size, this->post_len);
   }
+
+  inline bool get_glyph_from_name (const char *name, int len,
+                                   hb_codepoint_t *glyph) const
+  {
+    if (unlikely (!name) || unlikely(!len))
+      return false;
+
+    return this->post->get_glyph_from_name (name, len, glyph, this->post_len);
+  }
 };
 
 typedef bool (*hb_cmap_get_glyph_func_t) (const void *obj,
@@ -578,6 +587,17 @@ hb_ot_get_glyph_name (hb_font_t *font HB_UNUSED,
 }
 
 static hb_bool_t
+hb_ot_get_glyph_from_name (hb_font_t *font HB_UNUSED,
+                           void *font_data,
+                           const char *name, int len,
+                           hb_codepoint_t *glyph,
+                           void *user_data HB_UNUSED)
+{
+  const hb_ot_font_t *ot_font = (const hb_ot_font_t *) font_data;
+  return ot_font->post->get_glyph_from_name (name, len, glyph);
+}
+
+static hb_bool_t
 hb_ot_get_font_h_extents (hb_font_t *font HB_UNUSED,
 			  void *font_data,
 			  hb_font_extents_t *metrics,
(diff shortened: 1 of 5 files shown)


twardoch:
In CFF-based fonts, glyph names are stored inside the CFF table, not the post table. It seems that this PR doesn't take it into account, right??

twardoch:
Also: decoding CFF may a bit be expensive, so I'm not sure if it should be done in HB (especially since it will most likely be parsed again during the imaging step)

khaledhosny:
Yes, HarfBuzz’f ot font functions do not currently parse CFF table.

behdad:
Thanks for this!  I'll review it soon.

BTW, CFF doesn't have to be slow...  We'll eventually implement it.

src/hb-ot-post-table.hh
```diff
@@ -77,6 +117,52 @@ struct post
...
 
+  inline bool get_glyph_name (hb_codepoint_t glyph,
+                              char *buffer, unsigned int buffer_length,
+                              unsigned int blob_len) const
+  {
+    buffer[0] = '\0';
+
+    if (version.to_int () == 0x00010000)
+    {
+      strncpy(buffer, format1_glyph_names[glyph], buffer_length);
```
behdad: You don't check `glyph` before accessing `format1_glyph_names` array.

src/hb-ot-post-table.hh
```diff
@@ -77,6 +117,52 @@ struct post
...
+    }
+    else if (version.to_int () == 0x00020000)
+    {
+      const postV2Tail &v2 = StructAfter<postV2Tail>(*this);
+
+      if (unlikely (glyph >= v2.numberOfGlyphs))
+        return false;
+
+      unsigned int index = v2.glyphNameIndex[glyph];
+      if (index > 257)
```
behdad: I'd write >= 258...

src/hb-ot-post-table.hh
```diff
@@ -77,6 +117,52 @@ struct post
...
+      {
+        unsigned int offset = min_size + v2.min_size + 2 * v2.numberOfGlyphs;
+        char* data = (char*) this + offset;
+        for (unsigned int i = 0; data < (char*) this + blob_len; i++)
+        {
+          unsigned int name_length = data[0];
+          if (i == index - 258)
+          {
+            name_length = MIN (name_length, buffer_length);
+            memcpy (buffer, data + 1, name_length);
```
behdad: This can access past the blob end...

khaledhosny: Hopefully fixed now.

src/hb-ot-font.cc
```diff
@@ -301,6 +301,34 @@ struct hb_ot_face_cbdt_accelerator_t
...
+
+  inline void fini (void)
+  {
+    hb_blob_destroy (this->post_blob);
+  }
+
+  inline bool get_glyph_name (hb_codepoint_t glyph,
+                              char *name, unsigned int size) const
+  {
+    if (unlikely (!name) || unlikely(!size))
```
behdad: We just need to check for size.

src/hb-ot-post-table.hh
```diff
@@ -77,6 +119,60 @@ struct post
...
+  inline bool get_glyph_name (hb_codepoint_t glyph,
+                              char *buffer, unsigned int buffer_length,
+                              unsigned int blob_len) const
+  {
+    if (version.to_int () == 0x00010000)
+    {
+      if (glyph >= NUM_FORMAT1_NAMES)
+        return false;
+
+      strncpy(buffer, format1_names[glyph], buffer_length);
```
behdad: From `man strncpy`:
```       If  there  is  no  terminating  null  byte in the first n bytes of src,
       strncpy() produces an unterminated string in dest.  If buf  has  length
       buflen, you can force termination using something like the following:

           strncpy(buf, str, buflen - 1);
           if (buflen > 0)
               buf[buflen - 1]= '\0';
```

src/hb-ot-font.cc
behdad: Humm.  Returning false doesn't sound quite right. Don't know.  Maybe doesn't matter.

src/hb-ot-post-table.hh
```diff
@@ -77,6 +119,60 @@ struct post
...
+      buffer[buffer_length] = '\0';
```
behdad: Ok I see you have this.  But it should be `buffer_length - 1`, no?  I prefer checking for zero here as well.  Maybe just here, not in the caller.

behdad:
I merged and am fixing the rest myself.  Thanks Khaled, this has been long overdue!

khaledhosny:
That was quick 😄

PR #569, opened 2017-10-17
Switch from ICU deprecated unorm_normalize to unorm2_normalize.

roubert:
The new unorm2_normalize has been public API since ICU 4.4, while
the old unorm_normalize has been deprecated since ICU 56.

Now, beginning with ICU 60, unorm_normalize will also be annotated
U_DEPRECATED and trigger the compiler -Wdeprecated-declarations warning.

### DIFF
--- a/src/hb-icu.cc
+++ b/src/hb-icu.cc
@@ -34,7 +34,7 @@
 #include "hb-unicode-private.hh"
 
 #include <unicode/uchar.h>
-#include <unicode/unorm.h>
+#include <unicode/unorm2.h>
 #include <unicode/ustring.h>
 #include <unicode/utf16.h>
 #include <unicode/uversion.h>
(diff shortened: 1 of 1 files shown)


roubert:
@behdad @roozbehp 

If this patch looks good to you, it'd be very nice if you could cherry-pick it to external/harfbuzz_ng in Android aosp/master as well. We could add -Wno-deprecated-declarations when upgrading to ICU 60, but just fixing the root cause is so much nicer.

behdad:
Actually, if you read the code, we already use unorm2 stuff for ICU >= 49.  The old code is just fallback there, with this comment:
```
  /* We don't ifdef-out the fallback code such that compiler always 
   * sees it and makes sure it's compilable. */ 
```

How old is ICU 49?  We can remove the old code at some point.

roubert:
Oh, I see. As there was no #include <unicode/unorm2.h> I didn't even think about looking for any already existing use of it.

ICU 49 is from 2012 and ICU 60 is going to start triggering the -Wdeprecated-declarations warning for the fallback code.

roubert:
Btw., ICU 4.4 is from 2010 so if you'd like to change your code as little as possible and remain as backwards compatible as possible at the moment, you could merge this pull request to just move your fallback code to ICU 4.4 minimum while still getting rid of the usage of deprecated API.

PR #572, opened 2017-10-18
builds: Support HarfBuzz-GObject and introspection in CMake builds

fanc999:
Hi,

Since the NMake Makefiles were removed from the source tree, we should keep the CMake Makefiles more up-to-date by supporting builds of Harfbuzz-GObject from within it, especially for Visual Studio builds.  I will also look at (later) on how we may integrate introspection builds, but that will most likely be another PR.

This also fixes a typo in the CMake build files so that DirectWrite builds can be properly enabled.

Note that I have no idea on how the Mac frameworks are built, so I am not doing the framework support for Harfbuzz-GObject here.

With blessings, thank you!


ebraminio:
@fanc999: this is great, thanks for your great work. Is it possible to do replace-enum-strings inside the cmake file and not adding a new file to the project? Thanks again.

ebraminio:
> Note that I have no idea on how the Mac frameworks are built, so I am not doing the framework support for Harfbuzz-GObject here.

Honestly I just learned framework build part from freetype and don't know of any particular use so feel free to change the whole cmake build in any way you like :)

fanc999:
Hi @embraminio,

I am afraid that it is not possible, from my searches on Google, and I am pretty new to CMake.  Any suggestions on this would be great.

Oh, I thought, perhaps I would just go ahead to integrate the introspection builds in this PR anyways, so here it comes.

With blessings, thank you!

anthrotype:
Thanks @fanc999 for working on this! I've long been waiting to be able to build the harfbuzz python bindings for Windows with MSVC.
It would be nice to pass `HB_HAVE_GOBJECT` and `HB_HAVE_INTROSPECTION` to cmake in the appveyor.yml, to test that these can be built on the CI.

fanc999:
Hi @anthrotype,

For the CI part, I think we need to find a way for AppVeyor to download to use MSVC-compatible builds of GLib and GObject-Introspection.

The (former) NMake Makefiles also provided a way to build the introspection files as well, anyways...

One thing though, since this is a common situation on Windows where multiple Python versions may be installed, the CMake PythonInterp module will use the latest Python release series (i.e. 2.7, 3.5, 3.6...) (and system-matching architecture) that is available on the system, unless one has a Python interpretor in the PATH.

It would not matter if you are building Harfbuzz-GObject only, but it will matter if you are building the introspection files for Harfbuzz, as the Python release series (and architecture) must be the same as the one that was used to build _giscanner.pyd (the Python C-Module that is used by g-ir-scanner.

Hope this piece of info is useful for you.

With blessings, and cheers!

ebraminio:
Just let me know when you feel this can be merged :)

anthrotype:
> The (former) NMake Makefiles also provided a way to build the introspection files as well, anyways...

Yes, but I was never able to make that work. See https://github.com/behdad/harfbuzz/pull/164#issuecomment-183285126

I'll try again using the new method and see it this time it will work, thanks.

ebraminio:
@fanc999 Hi there. So gir is back and title should be updated? Can we merge this?

fanc999:
Hi @ebraminio,

Oops, I forgot to change the title back again...

I think this should be ready on my side to merge.  Maybe it would be a good idea (in another PR) to build the test programs (at least in test/api) when we enable Glib support so that we can include it in the AppVeyor CI, but the issue is, we need to use a corresponding build of Glib/G-I (which I have but needs to become distributed/downloaded for AppVeyor).  Note that all the API tests pass.

We can't do this for ARM yet, since the Glib stack is not tested/supported there, and I don't have an ARM system to test it.  We will need to see how Windows 10 on ARM will go for that regard.

ebraminio:
Thank you!

ebraminio:
@fanc999 as we are going to use meson, described on #2252 more, can make sure meson gobject support is on par with autotools and introspection? Have asked the same here https://github.com/harfbuzz/harfbuzz/pull/2173#issuecomment-598129188 also. Thanks

PR #575, opened 2017-10-22
Clarify the last step of release

ebraminio:


ebraminio:
@behdad, you've forgotten it, so :) It is in use for [vcpkg](https://github.com/Microsoft/vcpkg/blob/master/ports/harfbuzz/portfile.cmake) and [vitasdk](https://github.com/vitasdk/packages/blob/master/harfbuzz/VITABUILD), both added by me but movable to freedesktop releases if you wish so, but other than those, I remember we learned this was needed for some others also.

behdad:
Done. :)

BTW, I don't need to create a release.  After I push the tag, I just go edit the tag and upload binaries.

ebraminio:
Oh, so please fix it. Thanks

On a separate issue, https://github.com/behdad/harfbuzz/blob/master/harfbuzz.doap#L16 is referring to freedesktop, I am not sure how the whole thing is still relevant but maybe needs to be updated with current status of the project.

PR #582, opened 2017-10-26
Extend hb_set test for issue 579, and fix broken hb_set operations

jfkthame:
The inputs I reported in this issue were incomplete, as it turns out -- they were derived by using hb_set_next to enumerate the contents of the sets in question, but that was itself broken! So here's a complete example that fails with the 1.6.2 implementation of hb_set (but passed previously).

Obviously, this can't be merged right now as it'll make the test fail.

### DIFF
--- a/src/hb-set-private.hh
+++ b/src/hb-set-private.hh
@@ -313,13 +313,13 @@ struct hb_set_t
     b = nb;
     for (; a && b; )
     {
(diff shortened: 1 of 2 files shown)


behdad:
Ugh.  Thanks!  So I broke it in 30a591e3cef67818ef1ffec92c6a022b5a5f5e77

PR #583, opened 2017-10-26
Bugfix for iteration with hb_set_next, and add failing case to hb_set tests

jfkthame:
This fixes the bug that led to confusion about the sets involved in issue 579, because `hb_set_next` was returning incorrect results, leading to wrongly-printed set contents.

The breakage occurs when `hb_set_t::next` needs to cross a page boundary, because the new declaration of `j` in the inner loop here in `hb_set_t::page_t::next` goes out of scope when we exit with `goto found`, and we then return a result based on `j == ELT_BITS` from the termination of the preceding loop.

### DIFF
--- a/src/hb-set-private.hh
+++ b/src/hb-set-private.hh
@@ -95,7 +95,7 @@ struct hb_set_t
 	  goto found;
       for (i++; i < len (); i++)
         if (v[i])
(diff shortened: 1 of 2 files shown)


jfkthame:
Note, however, that this does not fix the `hb_set_intersect` bug in issue 579; there's something else wrong, too.

PR #584, opened 2017-10-26
Include all relevant glyphs when subsetting a font for testing

dscorbett:
record-test.sh subsets the input based on the characters in the input text. pyftsubset only includes glyphs for the characters in the input text, so if the test is of a string where OpenType inserts a dotted circle, the subsetted font will not include a dotted circle. The solution is to use `--glyphs` to tell pyftsubset to include all the glyphs HarfBuzz emits for the input text.

### DIFF
--- a/test/shaping/record-test.sh
+++ b/test/shaping/record-test.sh
@@ -41,6 +41,7 @@ if test $? != 0; then
 	echo "hb-shape failed." >&2
 	exit 2
 fi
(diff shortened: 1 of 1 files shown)


behdad:
Thanks for this.  Note that the subsetter is rather smart and follows OpenType lookups to gather glyphs.  But it doesn't know about script-specific shaper stuff that inserts dottedcircle for example.

PR #594, opened 2017-11-01
Enable PSVita (arm-none-eabi) compile again and add it to CI

ebraminio:
Does the src/hb-* modifications make sense?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,5 @@
 version: 2
 
-branches:
-  ignore:
-      - gh-pages
 jobs:
 
   base:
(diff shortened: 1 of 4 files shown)


behdad:
Yes, they look good.  Thanks.  Do you know what the sizeof(int) is on that platform?

ebraminio:
Don't have the device around but, https://github.com/RIOT-OS/RIOT/wiki/Family:-ARM#data-type-sizes as far as I know, matches [this](https://github.com/behdad/harfbuzz/issues/451#issuecomment-291421348) also.

PR #596, opened 2017-11-01
[ci] Fix branch limitation, hopefully

ebraminio:
follow up to https://github.com/behdad/harfbuzz/pull/592#issuecomment-340894585, circleci documentation feels a little vague


ebraminio:
Hmm, I am not about the effect of this.

PR #597, opened 2017-11-01
Fix clusters test differently

khaledhosny:


test/shaping/tests/cluster.tests
```diff
@@ -1,2 +1,2 @@
-fonts/sha1sum/6466d38c62e73a39202435a4f73bf5d6acbb73c0.ttf:--cluster-level=2:U+0078,U+030A,U+0058,U+030A:[gid2=0+1083|gid4=1@-555,-8+0|gid1=2+1200|gid4=3@-614,349+0]
+fonts/sha1sum/4fac3929fc3332834e93673780ec0fe94342d193.ttf:--cluster-level=2:U+0078,U+030A,U+0058,U+030A:[gid2=0+1083|gid3=1@-1131,-8+0|gid1=2+1200|gid3=3@-1190,349+0]
```
behdad: I don't get this.  Is this the composite flag as well?  The difference is huge.

khaledhosny: The composite component was shifted horizontally by -589 units, but I didn’t shift the simple glyph the same way.

behdad: I see.  I was suggesting just changing the composite flag to remove 0x4.  Anyway, doesn't matter.  We never visualize these results anyway.

khaledhosny: I tried removing the flag first but it didn’t work, I used `0x1000` instead of `0x1004`, but it is quite possible that I got confused fiddling with the bits 🙂

PR #598, opened 2017-11-01
[ci] Add Fedora builder

ebraminio:
As https://github.com/behdad/harfbuzz/issues/589#issuecomment-341161406 I thought adding a fedora based compile also would be nice. Unlike others, this does `make check` also.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,14 @@ version: 2
 
 jobs:
 
+  fedora:
+    docker:
+      - image: fedora
+    steps:
+      - checkout
+      - run: dnf install -y pkg-config ragel gcc gcc-c++ automake autoconf libtool make which glib2-devel freetype-devel || true
+      - run: ./autogen.sh && make && make check
+
   freebsd9:
     docker:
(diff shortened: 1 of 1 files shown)


behdad:
Thanks!

BTW, I'm still getting emails re gh-pages.

ebraminio:
There is a possible fix here #596 but I am not sure about it, it is ignoring top level branch filter and is a bug it seems, couldn't follow the documentations.

behdad:
Looks like the `yum`-based ones need '-y' to be passed to yum.  It's waiting for user to confirm installation and times out I think. https://circleci.com/gh/ebraminio/harfbuzz/111

ebraminio:
My attention there was on `No package pkg-config available. No package ragel available.` I added -y later.

ebraminio:
@behdad, Hmm, asked [here](https://discuss.circleci.com/t/excluding-a-branch-from-ci/17811) but I guess you should put a dummy .circleci/config.yml

behdad:
Works for me.

ebraminio:
Done on #599

PR #601, opened 2017-11-02
[test] Fix another instance of component rounding

khaledhosny:
See https://github.com/behdad/harfbuzz/pull/590#issuecomment-341194673


behdad:
Oh I thought we merged this.

khaledhosny:
This was the 2nd one, the 1st one was #597 🙂

PR #605, opened 2017-11-02
Simplify tracing code when tracing is disabled.

m0hamed:
If tracing is disabled, instead of using hb_auto_trace_t<0, ret_t>, a
new struct is used instead hb_no_trace<ret_t>. Using the simpler struct
reduces the binary size for release builds by removing unneaded debug
code.

These are the results of the difference (from chromium build with and without patch):
Section Sizes (Total=-101kb (-103544 bytes)):
    .bss: 0 bytes (0 bytes) (not included in totals)
    .data: 0 bytes (0 bytes) (-0.0%)
    .data.rel.ro: 0 bytes (0 bytes) (-0.0%)
    .rel.dyn: 0 bytes (0 bytes) (-0.0%)
    .rodata: -73.5kb (-75264 bytes) (72.7%)
    .text: -27.6kb (-28280 bytes) (27.3%)

MonochromePublic.apk_Breakdown (-106,496 bytes)
        +2 bytes Zip Overhead
  -106,496 bytes Native code size
        -2 bytes Package metadata size
MonochromePublic.apk_Specifics
  -106,496 bytes normalized apk size
  -106,496 bytes main lib size

For reference the compiled library size in full is 247,609 bytes, in the chrome for android apk.

### DIFF
--- a/src/hb-open-type-private.hh
+++ b/src/hb-open-type-private.hh
@@ -179,10 +179,14 @@ struct hb_dispatch_context_t
 #endif
 
 
+#if HB_DEBUG_SANITIZE == 0
+#define TRACE_SANITIZE(this) hb_no_trace_t<bool> trace;
+#else
 #define TRACE_SANITIZE(this) \
 	hb_auto_trace_t<HB_DEBUG_SANITIZE, bool> trace \
 	(&c->debug_depth, c->get_name (), this, HB_FUNC, \
 	 "");
+#endif
 
 /* This limits sanitizing time on really broken fonts. */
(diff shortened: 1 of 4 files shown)


m0hamed:
Here is the [chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=779572) about the issue.

behdad:
Thanks Mohamed.  We noticed this independently as well, but only on Android with clang.  To me, this looks like a clang bug.  I'll comment in the Chromium bug.

PR #609, opened 2017-11-06
Categorize U+20F0 COMBINING ASTERISK ABOVE for use in Grantha

dscorbett:
U+20F0 COMBINING ASTERISK ABOVE can be used as a svara marker in Grantha, but since Unicode does not define its Indic categories, a dotted circle is inserted between it and a following svara marker. For example, the string ⟨𑌊⃰𑍧𑍧⟩ from [L2/09-372](http://www.unicode.org/L2/L2009/09372-grantha.pdf#page=15) gets a dotted circle.


jfkthame:
What about the dash and caret marks that also appear in L2/09-372 samples -- should something similar be done about those?

jfkthame:
Hmm, I see there are U+A8FA and U+A8FB in the Devanagari Extended block, but I don't know enough about them to judge whether these are the same characters as seen in the Grantha examples, or behave differently in some way.

dscorbett:
That document specifies that they are U+1CD0 VEDIC TONE KARSHANA and U+1CD2 VEDIC TONE PRENKHA.

jfkthame:
Ah, right - I forgot about those, sorry!

behdad:
Waiting for @jfkthame to advise how to fix this.

jfkthame:
Given this seems to be a valid use case, @dscorbett's fix seems reasonable to me; I think we can just merge it.

behdad:
Thanks David.  Can you please update the patch to insert URL to this PR in the code as well?

src/gen-use-table.py
```diff
@@ -303,6 +304,9 @@ def map_to_use(data):
 		# TODO: https://github.com/behdad/harfbuzz/issues/525
 		if U == 0x1A7F: UISC = Consonant_Final; UIPC = Bottom
 
+		# TODO: Neither in Unicode 10.0 nor in USE's override list
```
behdad: Please add the PR URL here.

dscorbett: Done.

PR #610, opened 2017-11-06
Mark glyphs involved in 'stch' as unsafe to break

dscorbett:


behdad:
Awesome, thanks!

I'm curious to know how you came up with this.  Are you using unsafe-to-break?

dscorbett:
No, I was just reading the code to figure out what unsafe-to-break even is, and I wondered if an obscure feature like 'stch' had fallen through the cracks.

behdad:
Great.  Thanks :).  Really appreciate all your contributions.

Issue #485, opened 2017-05-15
Chinese Characters not well aligned with 'vpal' feature enabled


KrasnayaPloshchad:
I have tested hb-view with Source Han Serif font, if 'vpal' feature enabled, then Chinese Characters not centered, and the height have not been changed.
![test](https://cloud.githubusercontent.com/assets/10364351/26042436/10729d8c-3968-11e7-93dd-c4d9f447816f.png)
They should works as this (found at [Type is Beautiful](http://www.typeisbeautiful.com/2017/04/11961/)):
![sample](http://www.typeisbeautiful.com/wp-content/uploads/2017/04/SHSPropMetrics.png)

behdad:
Assigning to @kenlunde to see if he can figure out what's going on.  I have no idea and no time to debug any time soon.

kenlunde:
@behdad: The glyph for a single ideograph, U+4E00 一, is included within the scope of the 'vpal' GPOS feature. The following is my raw (using Unicode-based working glyph names) "features" file code for this glyph:

`position uni4E00-CN <0 140 0 -280>;`

The second element adjusts lowers the Y-axis origin by 140 units, whose effect is to raise the glyph by 140 units. The fourth element trims 280 units from the vertical advance, making it 720 units. The end result is that 140 units are trimmed from both the top and bottom of the glyph.

KrasnayaPloshchad:
So HarfBuzz need to handle all four elements to clip the glyph when this feature is activated. BTW I found HarfBuzz has the similar problem for 'vhal' feature.

kenlunde:
The 'vert' GPOS (not GSUB) feature also uses the same four elements, so it needs to be checked, too. However, given that U+3001 、 IDEOGRAPHIC COMMA seems to behave correctly and is handled via the same feature ('vpal'), I wonder whether something else it at work here.

behdad:
We seems to be applying GPOS just fine.  However, neither FreeType nor hb-ot-font implement VORG table.  I don't know what the reporter's screenshots are coming from, but probably same issue.  Without VORG, it's not possible to correctly get the vertical glyph origin..

KrasnayaPloshchad:
> I don't know what the reporter's screenshots are coming from, but probably same issue. 

According to TIB, this screenshot came from Adobe Illusruator to show the effect when enabling 'vpal' feature in SHS font.

KrasnayaPloshchad:
[...]

PR #619, opened 2017-11-17
attempt to offset vertical glyphs (#575) (#485)

marklh9:
This patch would fix incorrect glyph position
highlighted in https://imgur.com/a/I0jJU as well
as in #485. I try to explain my point of view
and would like to know if I'm correct.

Note: for harfbuzz I have do some cairo translate.
In LibreOffice I don't. I don't know why.

Original issue is to fix text offset issue of
vertical writing in LibreOffice (#575)
after LibreOffice update its HarfBuzz from 1.3.3 to 1.4.8. 
I tried to figure out how to use those values empirically.

Y_offset was 0 with HarfBuzz 1.3.3 and has
become -1160 (roughly 1EM upward) after upgrade to
1.4.8, I suspect that the offsets returned by Harfbuzz
are always repect to h_origin:

subtract_glyph_v_origin()
-> get_glyph_v_origin_with_fallback
-> guess_v_origin_minus_h_origin

But underlying renderer might expect positions relative
to the vertical origin. To compensate the result
I try to translate them back by calling
hb_font_add_glyph_origin_for_direction(),
to add the horizontal origin, hence make offset
relative to the v origin.

### DIFF
--- a/util/helper-cairo.cc
+++ b/util/helper-cairo.cc
@@ -466,7 +466,8 @@ helper_cairo_line_from_buffer (helper_cairo_line_t *l,
 			       const char          *text,
 			       unsigned int         text_len,
 			       int                  scale_bits,
-			       hb_bool_t            utf8_clusters)
+			       hb_bool_t            utf8_clusters,
+			       hb_font_t           *font)
 {
   memset (l, 0, sizeof (*l));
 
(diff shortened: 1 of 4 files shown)


behdad:
I'll try to review this soon. I've learned a lot about what's going on while debugging a similar issue on the Pango side. See https://github.com/harfbuzz/harfbuzz/issues/2765

behdad:
I don't think this is correct. Our vertical support should be correct already.

PR #626, opened 2017-11-21
Categorize U+A8B4 as Consonant_Medial

dscorbett:
U+A8B4 SAURASHTRA CONSONANT SIGN HAARU can come between a consonant and a vowel sign or virama ([L2/05-222R2](http://www.unicode.org/L2/L2005/05222r2-n2969r2-saurashtra.pdf#page=2)). It is really a nukta-like modifying letter, not a medial consonant, but Consonant_Medial seems like the best fit for its glyphic behavior.


behdad:
Should it e reported to Unicode as well?  Also, Nukta category wouldn't work better?

dscorbett:
I’ve been reporting all of these to Unicode.

The Nukta category corresponds to the USE class CONS_MOD, which only has the subclasses CONS_MOD_ABOVE and CONS_MOD_BELOW. The haaru has a post-base glyph, so it does not fit in either subclass, so it shouldn’t be in the Nukta category.

behdad:
Thank you very much David. :)

Issue #628, opened 2017-11-29
[coretext] Add hb_coretext_font_create()


behdad:
https://bugs.chromium.org/p/chromium/issues/detail?id=785522

ebraminio:
Maybe reviving parts of #367 instead?

behdad:
> Maybe reviving parts of #367 instead?

Same comment as https://github.com/harfbuzz/harfbuzz/pull/367#issuecomment-263238372
That is: a CGFont is a font face, CTFont is a font face plus font size. Same in HB: hb_face_t is a font face, hb_font_t is a font face plus size.  So the CGFont=hb_face_t / CTFont=hb_font_t is the right abstraction.

Plus, hb-coretext is extremely fragile.  I prefer this small change.

ebraminio:
Definitely fragile, my change also had some unexpected results, I hope we implement AAT shaper someday.

behdad:
> I hope we implement AAT shaper someday.

Sascha and I are sitting down together to work on that second week of January.

PR #486, opened 2017-05-15
Use stdatomic.h atomic primitives

ebraminio:
Tries to fix #345, `hb_atomic_ptr_impl_cmpexch` part is not ok

PR #629, opened 2017-11-29
[coretext] Add hb_coretext_font_create()

behdad:
Uncompiled, untested.

Fixes https://github.com/harfbuzz/harfbuzz/issues/628

New API:
hb_coretext_font_create()

### DIFF
--- a/src/hb-coretext.cc
+++ b/src/hb-coretext.cc
@@ -81,20 +81,11 @@ _hb_cg_font_release (void *data)
   CGFontRelease ((CGFontRef) data);
 }
 
-hb_face_t *
-hb_coretext_face_create (CGFontRef cg_font)
-{
-  return hb_face_create_for_tables (reference_table, CGFontRetain (cg_font), _hb_cg_font_release);
-}
 
 HB_SHAPER_DATA_ENSURE_DEFINE(coretext, face)
 HB_SHAPER_DATA_ENSURE_DEFINE_WITH_CONDITION(coretext, font,
 	fabs (CTFontGetSize((CTFontRef) data) - coretext_font_size (font->ptem)) <= .5
 )
-
-/*
- * shaper face data
- */
 
 static CTFontDescriptorRef
 get_last_resort_font_desc (void)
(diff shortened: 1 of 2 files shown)


src/hb-coretext.cc
```diff
@@ -306,6 +299,35 @@ _hb_coretext_shaper_font_data_destroy (hb_coretext_shaper_font_data_t *data)
...
+  CGFontRef cg_font = CTFontCopyGraphicsFont (ct_font, 0);
+  hb_face_t *face = hb_coretext_face_create (cg_font);
+  CFRelease (cg_font);
+  hb_font_t *font = hb_font_create (face);
+  hb_face_destroy (face);
+
+  if (unlikely (hb_object_is_inert (font)))
+    return font;
+
+  /* Let there be dragons here... */
```
drott: ?

behdad: Normally that value (ctfont) is automatically created on-demand by HarfBuzz generic layer.  We are sneaking in and setting it here.  It works and is correct, but is fragile if the rest of the code changes....

behdad:
@ebraminio since when we started -Werror'ing? So, Mac build has been broken since, because of the deprecation warnings?

behdad:
Ok I see I'm the one who added that...

behdad:
Can we add -Wno-deprecated-declarations to OS X builds?

behdad:
https://github.com/harfbuzz/harfbuzz/issues/345

ebraminio:
> Can we add -Wno-deprecated-declarations to OS X builds?

Does having access to a macOS terminal help you perhaps to try that or #486? If so, I guess CircleCI provides macOS machines with SSH access.

behdad:
> Can we add -Wno-deprecated-declarations to OS X builds?
> 
> Does having access to a macOS terminal help you perhaps to try that or #486? If so, I guess CircleCI provides macOS machines with SSH access.

Seems like I fixed it.  ssh to mac would be nice if it works.  But if it's too much work to figure out every time I needed every six months, easier to just sit next to drott physically somewhere in the world and do it together.

ebraminio:
Turned out CircleCI's macOS support needs some fee unlike TravisCI :(

PR #632, opened 2017-12-04
[hb-ft] Fix build when Multiple Master font support is disabled in freetype

blino:
FT_Set_Var_Blend_Coordinates() is not available when "Multiple Master font interface" is disabled in freetype's modules.cfg

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -354,6 +354,7 @@ if $have_freetype; then
 	save_libs=$LIBS
 	LIBS="$LIBS $FREETYPE_LIBS"
 	AC_CHECK_FUNCS(FT_Get_Var_Blend_Coordinates)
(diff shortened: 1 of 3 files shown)


behdad:
Is the check for `FT_Get_Var_Blend_Coordinates` not enough?

behdad:
Ok, sure. I like your patch.

blino:
`FT_Get_Var_Blend_Coordinates()` has been introduced in FreeType 2.7.1:
https://sourceforge.net/projects/freetype/files/freetype2/2.7.1/
while `FT_Get_Var_Blend_Coordinates()` is much older.

To make sure older versions of FreeType are handled, a separate check is needed.

Thanks for the quick review!

PR #633, opened 2017-12-04
Fix string matching in parse_bool

dscorbett:


behdad:
How did *that* happen I have no idea!

PR #638, opened 2017-12-05
Always hyphenate "pre-base-reordering", to avoid ambiguity

n8willis:
Replaces occurrences of "pre-base reordering" with "pre-base-reordering" in the comments.

There is a potential ambiguity for new readers in "pre-bse reordering" (one hyphen); it could be read as "pre-base, reordering" — i.e., a character in pre-base position that gets reordered. 

The term actually refers to a characters **in** a post-base position that gets reordered **to** a pre-base position, so "pre-base-reordering" is somewhat clearer.


behdad:
Thanks.  Merging master into your work is not good.  Can you rebase on top of master please?

n8willis:
Argh; rebase. Yep....

PR #647, opened 2017-12-06
Rewrite run-tests in python

ebraminio:
@khaledhosny I guess this helps on progressing your work on bringing tests to cmake.


behdad:
Thanks!

khaledhosny:
I was actually going to explore using [ctest scripting mode](https://cmake.org/Wiki/CMake_Scripting_Of_CTest) for the cmake build to avoid Python dependency, and keep the existing bash script for the autotools build.

test/shaping/run-tests.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+import sys, os, subprocess
+
+def cmd(command):
+	p = subprocess.Popen (
+		command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
```
anthrotype: Do you actually need shell=True? It spawns a sub shell. Only use it if you really use commands that only work inside a shell

test/shaping/run-tests.py
```diff
@@ -0,0 +1,86 @@
...
+	p.wait ()
+	return p.stdout.read ().decode ('utf-8'), p.returncode
+
+srcdir = os.environ.get ("srcdir", '.')
+builddir = os.environ.get ('builddir', '.')
+top_builddir = os.environ.get ('top_builddir', '../..')
```
anthrotype: Maybe you should use os.path.join() instead of using the literal forward slash, for better windows compatibility

test/shaping/run-tests.py
```diff
@@ -0,0 +1,86 @@
...
+fails = 0
+args = sys.argv[1:]
+
+reference = False
+if len (args) and args[0] == "--reference":
+	reference = True
+	args = args[1:]
+
+if not len (args):
+	args = ['/dev/stdin']
```
anthrotype: That’s not a portable way to read from standard input in python. You should use sys.stdin

test/shaping/run-tests.py
```diff
@@ -0,0 +1,86 @@
...
+
+for f in args:
+	if not reference:
+		print ("Running tests in " + f)
+
+	for line in open(f).read().strip().splitlines():
```
anthrotype: You don’t need to do splitlines to iterate over the lines of a file object, you can just do 

for line in fileobj:

ebraminio:
I guess not relying to cmake and sharing more codes across builds can be considered as advantages of this approach. Currently fails on msys2 however.

test/shaping/run-tests.py
ebraminio: It is in use for "../.." part, as this is a replacement for a bash script, hmm let me ask @behdad, is this script in use on projects outside harfbuzz? Guessed that because of smilingly unused `--reference` defined on the script.

anthrotype: you can do `os.path.join("..", "..")`

even the two dots have their portable alias in `os.path.pardir`

ebraminio: Done

test/shaping/run-tests.py
ebraminio: Done

test/shaping/run-tests.py
ebraminio: Done

test/shaping/run-tests.py
ebraminio: Thanks, done.

test/shaping/run-tests.py
```diff
@@ -0,0 +1,93 @@
...
+	print (p.stderr.read(), file=sys.stderr)
+	return p.stdout.read ().decode ("utf-8"), p.returncode
+
+srcdir = os.environ.get ("srcdir", ".")
+builddir = os.environ.get ("builddir", ".")
+top_builddir = os.environ.get ("top_builddir", os.path.normpath(os.path.join(os.getcwd(), "..", "..")))
+EXEEXT = os.environ.get ("EXEEXT", "")
+
+extra_options = "--verify"
+hb_shape = top_builddir + "/util/hb-shape" + EXEEXT
```
anthrotype: shouldn't you normalize this to `os.path.join(top_builddir, "util", "hb-shape" + EXEXT)`?

ebraminio: Thanks! :) Done

ebraminio:
Also your review and my changes seem to fixed msys2 failure, thanks @anthrotype! ;)

test/shaping/run-tests.py
behdad: > It is in use for "../.." part, as this is a replacement for a bash script, hmm let me ask @behdad, is this script in use on projects outside harfbuzz? Guessed that because of smilingly unused --reference defined on the script.

Script is not used outside, except that one can use --reference to regenerate the expected-output file, when things change in harfbuzz, eg. glyph-names are implemented in ot-fonts or other things that invalidate the output.  Or just to write test manually and generate expected output.

ebraminio:
> to avoid Python dependency

Honestly I missed this part of your comment @khaledhosny. From the issue I had #93 I had still unclear assumption that tests are already python depended but I couldn't find it now. I am just OK if you guys see the change unfit and want to revert it as I thought tests are already python dependent.

khaledhosny:
Python is not an issue on Linux or similar systems, but it would be an issue on Windows I presume, but I don’t do much development on Windows myself.

PR #648, opened 2017-12-07
[ci] Add an after distribution test runner

ebraminio:


behdad:
What does this do?

ebraminio:
Checks whether `make dist` folder reliable to use. For example #650 was failing on this as the newly introduced file, `run-fuzzer-tests.py` was not added to distribution.

PR #659, opened 2017-12-21
[ci] Add a DOS/djgpp compiler

ebraminio:
Not a super useful bot I guess but has interesting compile warnings https://circleci.com/gh/ebraminio/harfbuzz/374

Btw, main.exe result of harfbuzz compile really works on a dosbox.

WDYT?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,6 +107,14 @@ jobs:
       - run: apt update && apt install ragel
       - run: cmake -Bbuild -H. -GNinja && ninja -Cbuild
 
+  dos:
+    docker:
+      - image: quay.io/ebraminio/djgpp
+    steps:
+      - checkout
+      - run: apt update && apt install -y ragel pkg-config libtool autoconf
+      - run: ./autogen.sh --prefix=/usr/local/djgpp --host=i586-pc-msdosdjgpp && make
+
 workflows:
   version: 2
   build:
@@ -123,3 +131,4 @@ workflows:
       - linux-arm64
       - linux-mips
       - windows-x64
+      - dos


behdad:
Hahahahahahaha.  Sure.

Last line of your patch is missing a new line.

twardoch:
Doesn't hb-view have this ASCII output mode?

ebraminio:
It is more about what OS the executable will run on rather than what output mode would have.

Please note that DJGPP output is a 32bit executable rather than 16bit

> DJGPP is a complete 32-bit C/C++ development system for Intel 80386 (and higher) PCs running DOS. It includes ports of many GNU development utilities. The development tools require a 80386 or newer computer to run, as do the programs they produce.

> Programs compiled with DJGPP, and all development tools provided as part of DJGPP, look on the outside like normal DOS programs, and they rely on MS-DOS and BIOS for file I/O and other basic functions such as keyboard input, screen cursor position, etc. However, the bulk of the code in a DJGPP program is 32-bit protected-mode code; DJGPP programs use DPMI (the DOS Protected Mode Interface) to allow DOS/BIOS calls from protected mode. Therefore, any environment that can run DOS programs and provides DPMI services, will run DJGPP programs as well.

but even the fact it runs on 32bit protected mode, it is still fun to have as specially it runs on a dos machine and dosbox.

Apparently there are folks running competition for creation of new CGA or DOS games, [CGA JAM](https://itch.io/jam/cga-jam) like this one [Castle de Viana](https://itch.io/jam/cga-jam/rate/150219) and its [source](https://github.com/TheFakeMontyOnTheRun/castelo-de-viana) and they use DJGPP.

ebraminio:
Also before forgetting it, this is source of Dockerfile https://quay.io/repository/ebraminio/djgpp which is based on https://github.com/foone/docker-djgpp work:

```docker
FROM ubuntu:trusty

RUN apt-get update 
RUN DEBIAN_FRONTEND=noninteractive apt-get install -y build-essential bison flex curl gcc g++ make texinfo zlib1g-dev g++ unzip curl git

WORKDIR /root
RUN git clone https://github.com/andrewwutw/build-djgpp.git
WORKDIR /root/build-djgpp
RUN ./build-djgpp.sh 7.2.0
VOLUME /root/src
WORKDIR /root/src
ENV PATH /usr/local/djgpp/i586-pc-msdosdjgpp/bin/:$PATH
ENV GCC_EXEC_PREFIX /usr/local/djgpp/lib/gcc/
```

behdad:
Can you add -Wno-attributes to this bot to filter down the errors?

ebraminio:
@behdad: Good idea, #663 

Also have a look at warnings of the other bots, I guess some interesting stuff can be found there. 

Also any idea for some other approachable compilers for testing?

PR #660, opened 2017-12-21
Move Brahmi Number Joiner 1107F up so it actually returns HN

punchcutter:
USE cluster validation wasn't working on Brahmi numeral clusters using BNJ. Moving the USE category assignment for BNJ up makes it work correctly.


behdad:
This is generated code. The generator needs to be updated.

behdad:
Fixed. Would be nice to add a test for it.

PR #661, opened 2017-12-22
Categorize U+1107F as Number_Joiner

dscorbett:
Supersedes #660. It has the same effect, but this way the change won’t be overwritten next time the USE table is generated.


behdad:
See https://github.com/harfbuzz/harfbuzz/pull/660

PR #663, opened 2017-12-24
[ci] Add -Wno-attributes to djgpp, add alpine, format

ebraminio:


ebraminio:
Hmm, actually lets also fix psvita by switching tis compile to autotools, and format the whole recipe add an Alpine linux based compile. Alpine is a Linux distro with a different libc so that might be interesting but moreover I wanted to list the packages needed to build harfbuzz on Alpine somewhere so am ok with commenting this bot (or some of the other may felt unnecessary or broken) later.

PR #665, opened 2017-12-25
[cmake] Use GNUInstallDirs values

ebraminio:


ebraminio:
@fanc999 Hi there, can you quickly check whether this is OK with your use case sometime? Thank you very much.

PR #666, opened 2017-12-29
[cmake] Run fuzzing and shaping tests

ebraminio:
PTAL


test/CMakeLists.txt
```diff
@@ -1 +1,6 @@
 add_subdirectory(api)
+add_subdirectory(shaping)
+# Not ready yet
+if (NOT WIN32)
+  add_subdirectory(fuzzing)
```
khaledhosny: This should be checking for MSVC, the fuzzing is a clang/gcc feature.

ebraminio: Done.

ebraminio:
Hi @khaledhosny, can you give me a hand on msys2 Windows issues? I guess hb-fuzzer executable is not on the expected place or something.

ebraminio:
My guess is msys2 is using system's python instead msys2 one so it generates incorrect paths but don't know how to resolve if that is the case.

ebraminio:
I guess I figured it out (by dropping `python` and letting env decide about it), lets see if that make it working.

Other than this, fuzzing and shaping test are disabled on MSVC CI tests as we don't have HB_BUILD_UTILS enabled on appveyor yet and I guess they have still path issues. I suggest fixing their issue sometime later.

ebraminio:
> I guess I figured it out (by dropping python and letting env decide about it), lets see if that make it working.

Didn't work :(

PR #668, opened 2017-12-30
[cmake] Run rest of the tests on Windows

ebraminio:


ebraminio:
@khaledhosny As you were saying Khaled, I should have go for a cmake based solution, my python scripts are just losing real path of directories and binaries on Windows.

ebraminio:
Finally! Now we have a full featured cmake builder that can be relied on.

@behdad, consider experimentally using cmake for your own development sometime also, just test once how that is faster by running this command
```bash
cmake -DHB_HAVE_FREETYPE=ON -DHB_HAVE_GRAPHITE2=ON -DHB_BUILTIN_UCDN=ON -DHB_HAVE_GLIB=ON -DHB_BUILD_UTILS=ON -Bbuild -H. -GNinja && ninja -Cbuild && CTEST_OUTPUT_ON_FAILURE=1 ninja -Cbuild test
```
you won't be disappointed.

ebraminio:
screen shot 2018-01-01 at 1 43 55 pm

whole harfbuzz compilation and tests in 21 seconds in my 2011 Air machine...

ebraminio:
And 5 seconds, in some other machine.

behdad:
> cmake -DHB_HAVE_FREETYPE=ON -DHB_HAVE_GRAPHITE2=ON -DHB_BUILTIN_UCDN=ON -DHB_HAVE_GLIB=ON -DHB_BUILD_UTILS=ON -Bbuild -H. -GNinja && ninja -Cbuild && CTEST_OUTPUT_ON_FAILURE=1 CTEST_PARALLEL_LEVEL=8 ninja -Cbuild test

Sure, but this is an awful lot to remember...

behdad:
Ok, I like it so far. :)

ebraminio:
Great :) We can auto enable build options or put the command on some script if you agree to do so (on ./dev.sh or something).

behdad:
Just running ninja is fine. It uses optimal number of parallel jobs.  So it's just the initial cmake's parameters indeed.  Would be nice if they default to the same behavior as the configure equivalents, and the naming is consistent between the two.

behdad:
cmake still doesn't run the tests in src/. I like those because they catch many errors.  Any way we can hook those up too, if on a Unix-like system?

ebraminio:
> It uses optimal number of parallel jobs.

It seems however `CTEST_PARALLEL_LEVEL=8` was needed for making test run in parallel.

> Would be nice if they default to the same behavior as the configure equivalents, and the naming is consistent between the two.

Like turning `ninja test` to `ninja check`?

> cmake still doesn't run the tests in src/. I like those because they catch many errors. 

Honestly I wasn't aware of them at that moment :)

> Any way we can hook those up too, if on a Unix-like system?

Lets try that also :)

ebraminio:
#675

PR #674, opened 2018-01-03
[cmake] Colorize output when using cmake + ninja.

waywardmonkeys:
Due to how ninja runs clang and gcc, it doesn't colorize the output
like make does. This forces color output in this situation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -699,6 +699,16 @@ if (NOT SKIP_INSTALL_LIBRARIES AND NOT SKIP_INSTALL_ALL)
   endif ()
 endif ()
 
+if (UNIX AND CMAKE_GENERATOR STREQUAL "Ninja")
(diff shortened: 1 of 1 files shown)


ebraminio:
I guess it should be applied to CMAKE_C_FLAGS as we use it for some of units, and can this applied to upstream's cmake ninja generator also do you believe?

Btw, can you give us a hand on #675? I've put the command can be used for compilation, the issue is harfbuzz's cmake exporting unnecessary symbols and some static initializing things that tests that currently working fine with the output of autotools are catching them.

ebraminio:
Tried to resolve the conflict and add CMAKE_C_FLAGS but seem to broke it somehow?

waywardmonkeys:
I've gone ahead and rebased this forward, added setting the C compiler flags as well and force pushed it here.

ebraminio:
Great, as a bikeshedding, please put the new part before `# Test` section :)

waywardmonkeys:
Done.

PR #676, opened 2018-01-04
atomics: Favor compiler primitives over macOS APIs.

waywardmonkeys:
In macOS 10.12, the `OSMemoryBarrier` and related APIs were deprecated
in favor of using `std::atomic`. On the way to supporting `std::atomic`,
we can favor using the "Intel primitives" which are also available on
macOS.

### DIFF
--- a/src/hb-atomic-private.hh
+++ b/src/hb-atomic-private.hh
@@ -70,6 +70,29 @@ typedef LONG hb_atomic_int_impl_t;
 #define hb_atomic_ptr_impl_cmpexch(P,O,N)	(InterlockedCompareExchangePointer ((void **) (P), (void *) (N), (void *) (O)) == (void *) (O))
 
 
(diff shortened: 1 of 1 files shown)


ebraminio:
This is a good idea I believe, I went for std::atomic on #486 and not much success but feel free to try that also.

waywardmonkeys:
I was looking at using `std::atomic` as well ... but figured this was a much nicer first step and solves the immediate issue without all of the other pain that you saw in #486.

ebraminio:
Went ahead and merged this, without in detail knowing pros of cons of the approach, as our main clients on macOS at leasts, Chrome and Firefox, are using harfbuzz without these and this was just giving the developers on macOS compile headache. Of course I don't know much about build configuration of Qt and GTK, I don't believe that would hurt much as updates like this https://www.theregister.co.uk/2018/01/02/intel_cpu_design_flaw/ going to :)

Issue #678, opened 2018-01-04
Improving CMake and its tests


ebraminio:
First run CMake like this:

```
cmake -DHB_CHECK=ON -Bbuild -H. -GNinja && ninja -Cbuild && ninja -Cbuild test
```

<s>As #675 turned out, CMake, or its testing need improvements from various aspects:</s> almost done

* [x] Exact path for working directory of first test should be fixed in a portable way, `${PROJECT_BINARY_DIR}/CMakeFiles/harfbuzz.dir/src` should be replaced with some CMake magic (Improved but no much luck as https://stackoverflow.com/q/48296061)
* [x] `check-defs.sh` has some `make` stuff in inside that should be corrected before enabling it (#710)
* [x] `check-libstdc++.sh` is not happening on macOS at least it seems? (almost done on #707)
* [x] `check-static-inits.sh` is failing as it says there are objects with static initialization (not on Apple, but it is fine I guess)
* [x] After fixing the tests, some ci bot (or all of them) should test cmake build with `-DBUILD_SHARED_LIBS=ON`, https://github.com/harfbuzz/harfbuzz/blob/master/.circleci/config.yml
* [x] As `check-symbols.sh` there are private symbols exposed on CMake build? (fixed on #705, but maybe there are some better ways?)
* [x] `check-c-linkage-decls.sh` and `check-header-guards.sh` are not working, are they autotools specific?

Remained ones, for future:
* porting some of test scripts on src/ to python so we can run them on Visual C++ builds also (and test hb_check on appveyor, use `objdump` instead `nm` on Windows?)
* Autotools builds whole harfbuzz again on building hb-fuzzer, https://github.com/harfbuzz/harfbuzz/blob/master/test/fuzzing/Makefile.am but our cmake's `hb-fuzzer` is not doing the same. <s>Why autotools does that on building hb-fuzzer?</s> (update: explained by behdad below, done but commented: https://github.com/harfbuzz/harfbuzz/blob/master/test/fuzzing/CMakeLists.txt#L9)

behdad:
> As check-symbols.sh there are private symbols exposed on CMake build?

That's weird. `HB_INTERNAL` is supposed to take care of those.

>     check-static-inits.sh is failing as it says there are objects with static initialization

Yeah not sure this works on Apple platforms.

>     check-defs.sh has some make stuff in inside that should be corrected before enabling it

Right. It just makes sure harfbuzz.def is built.

[...]

PR #675, opened 2018-01-03
[cmake] Run src/ tests

ebraminio:

PR #679, opened 2018-01-04
Improve tests

ebraminio:
Some of #678 and https://github.com/harfbuzz/harfbuzz/pull/675#discussion-diff-159685424R7

### DIFF
--- a/src/check-header-guards.sh
+++ b/src/check-header-guards.sh
@@ -6,7 +6,7 @@ export LC_ALL
 test -z "$srcdir" && srcdir=.
 stat=0
 
-test "x$HBHEADERS" = x && HBHEADERS=`cd "$srcdir"; find . -maxdepth 1 -name 'hb*.h'`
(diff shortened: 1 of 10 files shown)


src/check-header-guards.sh
```diff
@@ -11,6 +11,7 @@ test "x$HBSOURCES" = x && HBSOURCES=`cd "$srcdir"; find . -maxdepth 1 -name 'hb-
 
 for x in $HBHEADERS $HBSOURCES; do
 	test -f "$srcdir/$x" -a ! -f "$x" && x="$srcdir/$x"
+	test "$x" = "./hb-gobject-structs.h" && continue;
```
ebraminio: @behdad, hadn't chance to fully digest the check, can you see what has went wrong `./hb-gobject-structs.h`?

ebraminio: Ok, couldn't digest it at 1AM but now can :) This seem OK, clever use of compiler directive and definition for preventing duplication on headers btw

ebraminio:
Added a HB_CHECK flag that enables most of optional parts and shared library compilation.
```
cmake -DHB_CHECK=ON -Bbuild -H. -GNinja && ninja -Cbuild && ninja -Cbuild test
```

PR #685, opened 2018-01-09
[cmake] Add header existence tests

ebraminio:
Hmm, but why isatty is detected on macOS but couldn't be used?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -138,8 +138,8 @@ workflows:
   version: 2
   build:
     jobs:
-      - oracledeveloperstudio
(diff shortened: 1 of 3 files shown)


ebraminio:
Similar goes to sys/mman.h, why cmake detects it on macOS but can't use it?

ebraminio:
Lots of CI breakage, help please!

khaledhosny:
I’m afraid I don’t know enough of CMake to be of any help here.

ebraminio:
No problem Khaled. I will try to sort it out somehow :) It is so weird, something interesting going on perhaps

PR #688, opened 2018-01-10
[test] Run run-tests.py in parallel

### LINKED ISSUE #687: [test] Speed up test/shaping/run-tests.py
Right now we spawn hb-shape for each line.  We can do just once and input multiple lines. Easier now that we are in Python.

ebraminio:
Closes #687

But I am not sure how much gain this have

### DIFF
--- a/test/shaping/run-tests.py
+++ b/test/shaping/run-tests.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 from __future__ import print_function
-import sys, os, subprocess
+import sys, os, subprocess, multiprocessing
 
 
 try:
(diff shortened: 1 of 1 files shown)


ebraminio:
It is of course faster but feels it should be even more faster? Have a look @behdad.

behdad:
Thanks Ebrahim. Sorry if I confused you.  That's not what I meant.  I meant, we should invoke one hb-shape and feed it all lines of text one at a time...

Also, don't touch it now. I have a reworked branch that I will push out later this week. It has touched the test suite a lot.

ebraminio:
3AM here, awaked by some shakes I felt is some minor earthquake! Now I see :) Well the gain here is about on 8seconds on my devices and it makes it complex so lets close this to see what we can do instead

ebraminio:
> so complex

Not so complex with pressing Ctrl+C in order to exit the test freeze the console and fixing that needs complicated workarounds.

PR #691, opened 2018-01-10
[cmake] Another try on making sure feature testing is working

ebraminio:
and some sorting on ci configs


ebraminio:
@Behdad, I've changed names of ci bots in hope understanding their propose become easier,

* crosscompile build are not running checks, thus having both "notest" and "crosscompile" in their names
* there are bots with cmake in their name and they use cmake configs, feel free to disable them when your autotools changes broke those, or just wait someone fix them :)
* cmake-gcc is a base cmake builder for the system
* distcheck is simply checking the repo is in state of releasing

If you have any naming suggestion or some bots you feel are overkill please let me know.

behdad:
Thank you!

PR #692, opened 2018-01-10
[WIP][test] Speed up test runner

behdad:
Keep one hb-shape process open and stream text to it.

Not ready to merge. I disabled the font-funcs=ot testing.  By keeping a pool of two or any other limited number of processes open we can fix this properly.

If I disable font-funcs=ot, then a single-process automake run of the tests in test/shaping take 13s without this and 8s with it.

Fixes https://github.com/harfbuzz/harfbuzz/issues/687

### DIFF
--- a/test/shaping/run-tests.py
+++ b/test/shaping/run-tests.py
@@ -1,16 +1,32 @@
 #!/usr/bin/env python
 
 from __future__ import print_function
-import sys, os, subprocess
+import sys, os, subprocess, re
 
+p = None
+cm = None
 
-def cmd(command):
-	p = subprocess.Popen (
-		command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-	p.wait ()
-	print (p.stderr.read (), file=sys.stderr, end='')
-	return p.stdout.read ().decode ("utf-8").strip(), p.returncode
+def cmd(command, line):
+	global p, cm
+	if command != cm:
+		if p:
+			p.stdin.close()
+		p = subprocess.Popen (command, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+		cm = command
+	p.stdin.write(line+'\n')
+	p.stdin.flush()
+	p.poll()
+	return p.stdout.readline ().decode ("utf-8").strip(), 0# p.returncode
 
+def parse (s):
+	s = re.sub (r"0[xX]", " ", s)
+	s = re.sub (r"[<+>{},;&#\\xXuUnNiI\n\t]", " ", s)
+	return [int (x, 16) for x in s.split ()]
+
+def utf8 (s):
+	s = u''.join (unichr (x) for x in parse (s))
+	if sys.version_info[0] == 2: s = s.encode ('utf-8')
+	return s
 
 args = sys.argv[1:]
 if not args or sys.argv[1].find('hb-shape') == -1 or not os.path.exists (sys.argv[1]):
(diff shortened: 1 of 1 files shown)


behdad:
I have no idea why these are failing...

ebraminio:
> I have no idea why these are failing...

Hmm, tests are also working fine locally but not on bots, maybe this is Linux only issue? It also happens on Travis bots but fortunately you can diagnose circleci bots remotely

screen shot 2018-01-10 at 11 56 32 pm

ebraminio:
One test doesn't work locally also on cmake:

```
 3/88 Test  #3: test-blob ..................................***Exception: Other  0.03 sec
/blob/empty: OK
/blob/duplicate: OK
/blob/readonly: OK
/blob/writable: OK
/blob/readonly-may-make-writable: **
ERROR:../test/api/test-blob.c:129:get_pagesize: assertion failed: (pagesize != (uintptr_t) -1)
```

ebraminio:
So will this eventually become something like grouping tests of similar fonts and run them at once? That would be great! I believe passing entries straightly to groupby https://docs.python.org/3/library/itertools.html#itertools.groupby even without sorting (as they are already are sorted) by their font name as key function would be nice.

Maybe font functions comparison can be moved to hb-shape?

behdad:
> One test doesn't work locally also on cmake (not related to this patch but the recent changes):

Can you debug this? It probably has to do with some feature detection stuff.  Looks like you have not defined any of these:
```
#if defined(HAVE_SYSCONF) && defined(_SC_PAGE_SIZE) 
  pagesize = (uintptr_t) sysconf (_SC_PAGE_SIZE); 
#elif defined(HAVE_SYSCONF) && defined(_SC_PAGESIZE) 
  pagesize = (uintptr_t) sysconf (_SC_PAGESIZE); 
#elif defined(HAVE_GETPAGESIZE) 
  pagesize = (uintptr_t) getpagesize (); 
```

behdad:
> Can you debug this? It probably has to do with some feature detection stuff. Looks like you have not defined any of these:

That's the problem with having multiple build systems. All but one are always lagging behind :)

behdad:
> So will this eventually become something like grouping tests of similar fonts and run them at once? That would be great! I believe passing entries straightly to groupby https://docs.python.org/3/library/itertools.html#itertools.groupby even without sorting (as they are already sorted usually) by their font name as key function would be nice.

We already have tests using the same font grouped together.

> Also, just a thought, maybe font functions comparison can be moved to hb-shape?

I thought about this more. Doing comparison stuff in hb-shape or another binary is something I have had wanted to do.  So, yes.  Syntax still to be determined.  I think as a first step for speed-up, I like a mode for hb-shape that would read command-line arguments from stdin and each line can set new arguments. So it's like invoking hb-shape once per line except that we skip the overhead and the font can also be kept open...  That would fix our test suite use-case. Comparison still to be determined.  But then we can keep that in the test runner and run it once with each font func.

ebraminio:
> That's the problem with having multiple build systems. All but one are always lagging behind :)

‌You are right, after all these tries on proposing autotools replacement I feel also more comfy with it nowadays :)

behdad:
> ‌You are right, after all these tries on proposing autotools replacement I feel also more comfy with it nowadays :)

Looking at what you are doing with cmake, it all looks more adhoc to me... :)

khaledhosny:
> Looking at what you are doing with cmake, it all looks more adhoc to me... :)

That is precisely why I never been a fan of CMake as it seems to lack any of the standardized ways autotools have for common development tasks and each project has to come with it adhoc solutions. The ugly command line invocation is the other thing.

Autotools biggest weakness is building on Windows with MSVC, but there are gcc-like wrappers to get around this. I use one in OTS and LibreOffice has one as well.

anthrotype:
> there are gcc-like wrappers to get around this

yeah but how reliable is using cccl for a complex project such as harfbuzz? anybody tried?

anthrotype:
> ugly command line invocation

you mean, the fact that you have to pass all these `-DHAVE_SOMETHING` options? I'm pretty sure there should be a way to enable those by default with the same values as in ./configure.

ebraminio:
Before going to remove or replace cmake, I hope two things considered, some of the APIs specially needed for Kashida still are missing on msys2 dwrite header https://github.com/Alexpux/mingw-w64/blob/master/mingw-w64-headers/include/dwrite_1.h and the other thing is https://github.com/Microsoft/vcpkg and https://github.com/wingtk/gvsbuild support. Meson as it generates MSVC project files can do both.

khaledhosny:
> you mean, the fact that you have to pass all these `-DHAVE_SOMETHING` options? I'm pretty sure there should be a way to enable those by default with the same values as in ./configure.

Not just the defaults, they are also undiscoverable. With `configure` I get auto-completion; `./configure --<TAB><TAB>` gives me all the options, or I can just call `./configure --help`, with CMake I have neither. Also `--with-something` is much easier to type and remember than `-DHAVE_SOMETHING` which might be `-DENABLE_SOMETHING` or `-DWITH_SOMETHING` in a different CMake-using project.

anthrotype:
https://cmake.org/cmake/help/latest/manual/cmake.1.html

```
-L[A][H]
List non-advanced cached variables.

List cache variables will run CMake and list all the variables from the CMake cache that are not marked as INTERNAL or ADVANCED. This will effectively display current CMake settings, which can then be changed with -D option. Changing some of the variables may result in more variables being created. If A is specified, then it will display also advanced variables. If H is specified, it will also display help for each variable
```

khaledhosny:
You need run CMake first before calling this 🙂

behdad:
Closing this.  Let's continue in https://github.com/harfbuzz/harfbuzz/issues/687

PR #703, opened 2018-01-12
[ci] Update msvc builder

ebraminio:


ebraminio:
Lets forget about this right now

PR #704, opened 2018-01-12
[aat] First dig on 'kerx'

ebraminio:
Doc: https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6kerx.html

Far far from anything useful... based on `hb-ot-kern-table.hh` stock on implementing KerxSubTableFormat0, perhaps you should give me a hand, and no format 4 and 6 and `Glyph Coverage Tables` yet. Not sure if this has anything useful, but at least it builds and as there is no AAT vs OT duality (AFAIK), lots of unnecessary things are cleaned and tweaked some of the types that are different between the two (Offset to LOffset, uint16 to uint32) whenever I could.

### DIFF
--- a/src/hb-ot-layout-private.hh
+++ b/src/hb-ot-layout-private.hh
@@ -131,6 +131,7 @@ namespace OT {
 }
 
 namespace AAT {
+  struct kerx;
   struct morx;
 }
 
@@ -169,6 +170,7 @@ struct hb_ot_layout_t
   OT::hb_lazy_table_loader_t<struct OT::MATH> math;
   OT::hb_lazy_table_loader_t<struct OT::fvar> fvar;
   OT::hb_lazy_table_loader_t<struct OT::avar> avar;
+  OT::hb_lazy_table_loader_t<struct AAT::kerx> kerx;
   OT::hb_lazy_table_loader_t<struct AAT::morx> morx;
 
   unsigned int gsub_lookup_count;
--- a/src/Makefile.sources
+++ b/src/Makefile.sources
@@ -82,6 +82,7 @@ HB_FALLBACK_sources = \
 HB_OT_sources = \
 	hb-aat-layout.cc \
 	hb-aat-layout-common-private.hh \
+	hb-aat-layout-kerx-table.hh \
 	hb-aat-layout-morx-table.hh \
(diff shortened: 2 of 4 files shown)


ebraminio:
[devongovett/fontkit#127](https://github.com/devongovett/fontkit/issues/127), [kerx.js](https://github.com/devongovett/fontkit/blob/aat-kern/src/tables/kerx.js)

behdad:
Thanks Ebrahim! Let me finish morx and turn to this after.

ebraminio:
Implementing logics for all KerxSubTableFormatX versions would be hard I guess, is having some statistics would be useful (that which version is more used) or will we (you in particular :P) be able to implement all of them on one version?

Is Apple's [`kern`](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6kern.html) also `morx` like state machine? If so is harfbuzz implemented that or it just implements OT's one? (which isn't?) I was wondering despite the huge difference and complicatedness of `hb-ot-kern-table.hh` shouldn't it separated into two different file, maybe `hb-ot-kern-table.hh` and `hb-aat-layout-kern-table.hh` with having different logics to deal with them?

And as I like to get more experience from these, can you suggest some task on this ongoing AAT project?

behdad:
```
./hb-aat-layout-kerx-table.hh:111:3: error: control reaches end of non-void
      function [-Werror,-Wreturn-type]
  }
```
Please fix that so bots are happy and then merge. I'll take it from there.

src/hb-ot-layout-private.hh
```diff
@@ -169,6 +170,7 @@ struct hb_ot_layout_t
   OT::hb_lazy_table_loader_t<struct OT::MATH> math;
   OT::hb_lazy_table_loader_t<struct OT::fvar> fvar;
   OT::hb_lazy_table_loader_t<struct OT::avar> avar;
+  //OT::hb_lazy_table_loader_t<struct AAT::kerx> kerx;
```
behdad: You can uncomment this.

ebraminio:
> Please fix that so bots are happy and then merge. I'll take it from there.

Done. Thats an honor.

PR #706, opened 2018-01-16
Fix for CMake 3.0 and below

rdb:
This is a trivial change that fixes compilation in CMake 2.8.

The "SOURCES" option to `try_compile` is not available in the stated minimum CMake version 2.8.  However, it is only necessary if more than once source file needs to be compiled, which isn't the case here.


ebraminio:
Excellent! Thank you! Consider having a look at some of the issues #678 also :)

PR #714, opened 2018-01-18
Hatran script is right-to-left

anthrotype:
I just stumbled on this ancient script from Iraq. Apparently HarfBuzz thinks it's left-to-right, but it turns out:

"Hatran is written from right to left horizontally"

http://www.unicode.org/L2/L2012/12312-n4324-hatran.pdf

This script was added with Unicode 8.0.

Also this spreadsheet (referenced in an inline comment in the source code) has it as RTL: http://goo.gl/x9ilM


ebraminio:
@anthrotype I know you were more confident if Behdad was merging this but as I found this low risk and making sense at the same time, I did it :)

anthrotype:
Very low risk! Thanks :)

PR #716, opened 2018-01-18
Update record-test.sh to the new directory layout

dscorbett:
Since the test directories were reorganized a week ago, record-test.sh was out of date. Maybe there should be a continuous integration test for the test framework. :)


ebraminio:
> Maybe there should be a continuous integration test for the test framework. :)

We won't have any issue with that :P

PR #718, opened 2018-01-20
[ci] unbreak bots with outdated freetype version

### LINKED ISSUE #701: Bots failing
I think we might want to compiler our own freetype. Or maybe I should update the test suite to know how to skip some tests... That's why travis is failing for example.

ebraminio:
Fixes #701


ebraminio:
Hmm, no idea what to do honestly, closing till figuring it out.

PR #724, opened 2018-01-22
[cmake] fix symbol exporting issue on newer gcc

### LINKED ISSUE #723: CMake builds exposing "roundf" as a "weak" symbol
Not that important (or at all) but as [this bot log](https://circleci.com/gh/harfbuzz/harfbuzz/5035) "_ZSt5roundf" is exposed as a weak symbol but this just happens on Ubuntu 17.10 and was not happening on an older version and it only happens on CMake builds. Steps to reproduce:

1. `cmake -DHB_CHECK=ON -Bbuild -H. -GNinja && ninja -Cbuild && CTEST_OUTPUT_ON_FAILURE=1 CTEST_PARALLEL_LEVEL=8 ninja -Cbuild test`
2. `nm build/libharfbuzz.so | grep ' W '`

It can't see why why just this symbol is exposed on this way. [This](https://github.com/harfbuzz/harfbuzz/commit/8b60e7c2c301a24906d591c80c42fd5e506072d2#diff-af3b638bc2a3e6c650974192a53c7291) can be reverted if this could be fixed.

ebraminio:
Closes #723

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -269,7 +269,7 @@ if (HB_HAVE_FREETYPE)
 
   list(APPEND THIRD_PARTY_LIBS ${FREETYPE_LIBRARIES})
   include_directories(AFTER ${FREETYPE_INCLUDE_DIRS})
-  add_definitions(-DHAVE_FREETYPE=1 -DHAVE_FT_FACE_GETCHARVARIANTINDEX=1)
+  add_definitions(-DHAVE_FREETYPE=1)
 
   list(APPEND project_sources ${PROJECT_SOURCE_DIR}/src/hb-ft.cc)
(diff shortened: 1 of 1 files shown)


behdad:
What's the "fix"?

ebraminio:
Adding -D__CORRECT_ISO_CPP11_MATH_H_PROTO_FP, bringing back links to libc libm.

ebraminio:
Of course I am not sure why -D__CORRECT_ISO_CPP11_MATH_H_PROTO_FP was needed but that was the only way to bring back disabled exported symbols tests on CMake.

behdad:
Where did you get that from?
This cmake experiment just made me appreciate the good old autotools a lot more :).

ebraminio:
From looking at its header, hehe.

PR #729, opened 2018-01-26
Use KernSubTableFormat2::rightClassTable

bungeman:
Issue originally reported at http://cppfiles.com/chromium.txt .

This fixes what appears to be a copy - paste issue which causes both
KernSubTableFormat2::rightClassTable and the 'right' argument to
KernSubTableFormat2::get_kerning to go unused.


behdad:
Thanks.

PR #730, opened 2018-01-28
Refactor script and language tags

### LINKED ISSUE #362: Erroneous conversion of fonipa to IPPH

dscorbett:
This pull request updates everything related to parsing script and language tags. It fixes #362.

A private use subtag beginning with “hbsc” is parsed as a script override. See #645, which this pull request subsumes.

The functions that determine the script and language of a buffer can return any number of results, not just two scripts and one language. Among the results, the map builder chooses the first script and language found in the font.

The mappings between BCP 47 and OpenType tags are automatically generated. Doing it manually has resulted in missing tags (BSK), typos (BER for BBR), and overgeneration (kvd → KUI). There are still manual overrides where necessary. `diff <(git show language-tags~8:src/hb-ot-tag.cc) <(git show language-tags:src/hb-ot-tag-table.hh)` shows the mapping changes.


behdad:
Wow, this is amazing.  Thank you!

I did a quick review. Looks great.  I'll do a more thorough review of the API changes and merge after making a release with existing master.

behdad:
On the generated table:
- hb_ot_ambiguous_tag_to_language() should be converted to a table and bsearch().
- I'm slightly worried about performance of hb_ot_tags_from_complex_language().  Can you at least do a switch based on the first letter of the tag to compare?

behdad:
Can you rebase on master please? That should fix the bots.

behdad:
> Where the old functions return the default tags 'DFLT' and 'dflt',
`hb_ot_tags` returns an empty array. The caller is responsible for
using the default tag in that case.

I don't think I like this part, but can live with it.  Will do a more thorough review.

I was thinking, instead of `hb_ot_tags()` name, we need `hb_ot_tags_from_bcp47` and `hb_ot_tags_to_bcp47`. No?

Also, we like to have `hb_scripts_from_language()` and `hb_language_likely_for_script()` kind of functionality. Any way you can add those (possibly separately) as well?

behdad:
Also, opening braces "{" go on a new line. :) Thanks.

dscorbett:
> I'm slightly worried about performance of hb_ot_tags_from_complex_language(). Can you at least do a switch based on the first letter of the tag to compare?

I’ll fix this.

> hb_ot_ambiguous_tag_to_language() should be converted to a table and bsearch().

I can do this, but is it necessary? It is just a switch statement on a `uint32_t`, so wouldn’t it get compiled to something like binary search anyway?

> we like to have hb_scripts_from_language() and hb_language_likely_for_script() kind of functionality. Any way you can add those (possibly separately) as well?

This requires CLDR. I’d like to do it as a separate pull request.

> I was thinking, instead of `hb_ot_tags()` name, we need `hb_ot_tags_from_bcp47` and `hb_ot_tags_to_bcp47`.

Sure, though I would prefer `bcp_47` over `bcp47`, since it is called “BCP 47”, not “BCP47”. `hb_ot_tags_to_bcp47` would just be a wrapper around `hb_ot_tag_to_script` and `hb_ot_tag_to_language`; should those be deprecated?

> opening braces "{" go on a new line. :)

HarfBuzz is inconsistent about this, but okay, I will change it.

> `HB_OT_MAX_TAGS_PER_SCRIPT` we want to increase since `dev3` is planned. Maybe both of them set to 8?

That seems wasteful. My idea was that `HB_OT_MAX_TAGS_PER_SCRIPT` would be exactly the maximum, and that when 'dev3' was implemented, it would be increased to 3. Otherwise, you’ll get users allocating space for 8 scripts when there are only 2 or 3.

> in the codebase itself, we use `bool`, not `hb_bool_t`.

The codebase is inconsistent. I propose that I use `bool` except when a function is a wrapper around, or a close analogue of, an existing function that uses `hb_bool_t`.

behdad:
> hb_ot_ambiguous_tag_to_language() should be converted to a table and bsearch().
> 
> I can do this, but is it necessary? It is just a switch statement on a uint32_t, so wouldn’t it get compiled to something like binary search anyway?

I'm not sure they do. Do they?

>     we like to have hb_scripts_from_language() and hb_language_likely_for_script() kind of functionality. Any way you can add those (possibly separately) as well?
> 
> This requires CLDR. I’d like to do it as a separate pull request.

Sounds good.

>     I was thinking, instead of hb_ot_tags() name, we need hb_ot_tags_from_bcp47 and hb_ot_tags_to_bcp47.
> 
> Sure, though I would prefer bcp_47 over bcp47, since it is called “BCP 47”, not “BCP47”.

We already have things like hb_script_from_iso15924_tag, so I prefer bcp47.  Also, to me, the underscores are semantic separators, not necessarily just space replacements.  We also use "codepoint", not "code point", etc.

> hb_ot_tags_to_bcp47 would just be a wrapper around hb_ot_tag_to_script and hb_ot_tag_to_language; should those be deprecated?

No, not deprecated.

>     opening braces "{" go on a new line. :)
> 
> HarfBuzz is inconsistent about this, but okay, I will change it.

You are absolutely right.  I started with putting it on same line.  Over the years, found it unreadable and switched style.  Whenever touching code, it's encouraged to update the surrounding (but not far away) to new style.

>     HB_OT_MAX_TAGS_PER_SCRIPT we want to increase since dev3 is planned. Maybe both of them set to 8?
> 
> That seems wasteful. My idea was that HB_OT_MAX_TAGS_PER_SCRIPT would be exactly the maximum, and that when 'dev3' was implemented, it would be increased to 3. Otherwise, you’ll get users allocating space for 8 scripts when there are only 2 or 3.

This gets compiled into application code. So next time HarfBuzz is updated to add dev3, application code would also need updating.  Since we are talking about 8*4 bytes allocated on the stack, I don't think it's a problem.  But sure, I'd be fine with 3.

>     in the codebase itself, we use bool, not hb_bool_t.
> 
> The codebase is inconsistent. I propose that I use bool except when a function is a wrapper around, or a close analogue of, an existing function that uses hb_bool_t.

Again, correct.  Started as hb_bool_t; over the years switched to bool when C API was not getting in the way.

Thanks

dscorbett:
Switching on the first character made `hb_ot_tags_from_complex_language` much faster. `"en"` went from an average of 0.586 μs per call to 0.056 μs. `"cdo"` got a little worse (from 0.588 μs to 0.687 μs) but it is worth it. I’ll push this soon.

To determine whether a switch statement is compiled to binary search, I ran `objdump -rwCS -Mintel libharfbuzz.so.0`. The first line of `hb_ot_ambiguous_tag_to_language` is:
```objdump
   188e0:   81 ff 20 49 56 4c       cmp    edi,0x4c564920
```
which corresponds to 'LVI ', which would indeed be the first pivot, given this set of tags. So it’s probably doing something like binary search, at least with g++ on CentOS.

After thinking about it some more, I don’t think it makes sense to rename `hb_ot_tags` to `hb_ot_tags_from_bcp47`. The function takes an `hb_language_t` and an `hb_script_t`. An `hb_language_t` is a wrapper around a BCP 47 string. The name `hb_ot_tags_from_bcp47` would be synonymous with `hb_ot_tags_from_language`, which is confusingly similar to the existing `hb_ot_tag_from_language`, and would ignore the script parameter. How about `hb_ot_tags_from_script_and_language`?

behdad:
Thanks. Switches seem to be compiled fine to bsearch indeed.

behdad:
> After thinking about it some more, I don’t think it makes sense to rename hb_ot_tags to hb_ot_tags_from_bcp47. The function takes an hb_language_t and an hb_script_t. An hb_language_t is a wrapper around a BCP 47 string. The name hb_ot_tags_from_bcp47 would be synonymous with hb_ot_tags_from_language, which is confusingly similar to the existing hb_ot_tag_from_language, and would ignore the script parameter. How about hb_ot_tags_from_script_and_language?

You are right. `hb_ot_tags_from_script_and_language` sounds great.  Do we need the inverse?

dscorbett:
No, the inverse is not needed, because there are already `hb_ot_tag_to_script` and `hb_ot_tag_to_language`. It might be convenient to include it though.

behdad:
> No, the inverse is not needed, because there are already hb_ot_tag_to_script and hb_ot_tag_to_language. It might be convenient to include it though.

But you can't encode, say, deva vs dev2 distinction in the old API.

dscorbett:
On `master`, `hb_ot_tag_to_script` returns `HB_SCRIPT_DEVANAGARI` for both 'deva' and 'dev2', so I am not sure what you mean.

behdad:
> On master, hb_ot_tag_to_script returns HB_SCRIPT_DEVANAGARI for both 'deva' and 'dev2', so I am not sure what you mean.

Exactly.  One of the reasons this new API is being added is to be able to choose 'deva' when the font has both; using the -hbsc extension.  Right?  I think if the script tag passed to the reverse function is NOT the first tag for script, we should encode it in the language tag.

Also, if the script is not the most likely script for the language, encode it in the language tag as well? Probably not.

Ok, I'm not sure about these, but my first point about deva makes roundtripping much more likely.

dscorbett:
Now I see what you mean. That makes sense.

dscorbett:
I think I’ve made all the requested changes. I don’t know why Codacy is finding fault with the character buffer code in `hb_ot_tags_to_script_and_language`.

behdad:
> I think I’ve made all the requested changes.

Any reason you moved `hb_ot_tags_from_script()` from hb-ot-tag.h to hb-ot-layout.h?

Also, hardcoding max script tags to 2 is not good.  We want at least 3, since dev3 etc will be added in the future.  By setting it to 3 now, applications wouldn't need to be recompiled.

behdad:
I still have to sit down and do a thorough review of this.

behdad:
@dscorbett Is this ready from your point of view?  I like to get it in before next release, so I'll do a review tonight.

dscorbett:
From my point of view, this is ready. clang-everything is failing, though, because of a `-Wunreachable-code-return` violation. I’ll fix that.

dscorbett:
There are now a couple of deprecation warnings. It is not clear how to fix `_hb_graphite2_shape`, which calls `hb_ot_tags_from_script`, which I deprecated.

https://github.com/harfbuzz/harfbuzz/blob/4035158de46ce373b7521daf61c5b6df83312968/src/hb-graphite2.cc#L281-L287

This code assumes that `script_tag[1]` might be `HB_TAG_NONE`, but `hb_ot_tags_from_script` actually uses `HB_OT_TAG_DEFAULT_SCRIPT` as the default tag instead. It seems there is a bug here, but I don’t know what this code is supposed to do, so I am wary of fixing it.

behdad:
This looks REALLY good.  Thank you!!!  I'm merging now.  We'll deal with breakages later.

behdad:
All should be fixed now!

ebraminio:
Is this fixed also https://github.com/harfbuzz/harfbuzz/blob/c93d379bf2414eb6ee6d38fb9753ddf818a777ea/TODO#L18 ?

behdad:
> Is [this](https://github.com/harfbuzz/harfbuzz/blob/c93d379bf2414eb6ee6d38fb9753ddf818a777ea/TODO#L18) fixed also?

Probably as much as we'd do.  I don't think we want to expose API for it.  Or maybe one or two API entries to better compare languages is fine.

Getting most-likelly language from script, and scripts from language is still interesting API.  Maybe open an issue and assign to @dscorbett ?

PR #731, opened 2018-01-28
Fix Travis build on macOS

dscorbett:
Travis builds on macOS have been failing lately because check-defs.sh does not fully ignore the prefix “_” when comparing symbol lists. When the test fails, Travis does not print the log, because it tries to run find with a syntax that does not work on macOS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,7 @@ matrix:
         - ./autogen.sh
         - ./configure $CONFIGURE_OPTS --with-coretext
         - make
-        - make check || (cat `find -name '*.log'` && false)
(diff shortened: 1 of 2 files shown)


ebraminio:
Thank you very much!

ebraminio:
Any idea why "check-libstdc++.sh" is getting skipped? Locally the test works fine on macOS on both autotools and cmake builds and I am not sure why it doesn't work.

dscorbett:
Do you have ldd installed locally? I don’t, and the test gets skipped for me too.

ebraminio:
Oh, great help, I had it from somewhere unrelated, otool is its replacement on macOS, #732

PR #735, opened 2018-01-31
[coretext] Fix memory leak.

waywardmonkeys:
In `reference_table`, if the data is empty and we return early,
we still need to release the data object.


waywardmonkeys:
Force pushed this with 2 additional memory leak fixes.

PR #742, opened 2018-02-01
Documentation for HB_GLYPH_FLAG_UNSAFE_TO_BREAK.

bungeman:
HB_GLYPH_FLAG_UNSAFE_TO_BREAK means that the glyph with this flag is somehow affected by the previous logical glyph (the previous index in the buffer if ltr and the next index if the buffer is rtl). If these two glyphs are separated by a break (line or otherwise) then the underlying text should be re-shaped on both sides up to corresponding position in the text of some glyph not marked with this flag.

### DIFF
--- a/src/hb-buffer.h
+++ b/src/hb-buffer.h
@@ -71,9 +71,15 @@ typedef struct hb_glyph_info_t {
   hb_var_int_t   var2;
 } hb_glyph_info_t;
 
+/**
+ * hb_glyph_flags_t:
+ * @HB_GLYPH_FLAG_UNSAFE_TO_BREAK: If this glyph is split from the previous logical glyph
+ *                                 then both sides should be re-shaped.
+ */
 typedef enum { /*< flags >*/
   HB_GLYPH_FLAG_UNSAFE_TO_BREAK		= 0x00000001,
 
+  /*< private >*/
   HB_GLYPH_FLAG_DEFINED			= 0x00000001 /* OR of all defined flags */
 } hb_glyph_flags_t;
 


bungeman:
Note that I derived this information by tracking down and then reading https://github.com/harfbuzz/harfbuzz/blob/c6dbf6e77cae30772ffa110c651cb4287ab3b0dc/util/options.hh#L329 and trying to figure out the right way to explain what it's doing. It's possible this is completely wrong, so please take a close look. This is somewhat an issue report that I had no idea how to use this flag since it doesn't appear very well documented yet.

behdad:
Thanks Ben. I expanded it:
```
- * @HB_GLYPH_FLAG_UNSAFE_TO_BREAK: If this glyph is split from the previous logical glyph
- *                                 then both sides should be re-shaped.
+ * @HB_GLYPH_FLAG_UNSAFE_TO_BREAK: Indicates that if input text is broken at the
+ *                                beginning of the cluster this glyph is part of,
+ *                                then both sides need to be re-shaped, as the
+ *                                result might be different.  On the flip side,
+ *                                it means that when this flag is not present,
+ *                                then it's safe to break the glyph-run at the
+ *                                beginning of this cluster, and the two sides
+ *                                represent the exact same result one would get
+ *                                if breaking input text at the beginning of
+ *                                this cluster ans shaping the two sides
+ *                                separately.  This can be used to optimize
+ *                                paragraph layout, by avoiding re-shaping
+ *                                of each line after line-breaking, or limiting
+ *                                the reshaping to a small piece around the
+ *                                breaking point only.
```

bungeman:
nit: s/ans/and

PR #743, opened 2018-02-02
[cmake] Improve emscripten support

ebraminio:
Done as just an experiment, it specifically uses the WASM support of emscripten also.

0. Install emscripten, java and nodejs
1. `emcmake cmake -Bbuild -H. -GNinja  && ninja -Cbuild`
2. `cd build && node -e "require('./harfbuzz.js')().then(module => console.log(module.cwrap('hb_version_string', 'string', [])()))"`

alternatively,

2. `echo "<script src='harfbuzz.js'></script><script>Module().then(module => alert(module.cwrap('hb_version_string', 'string', [])()))</script>" > build/a.html`
3. open a.html on an http server like `python3 -m http.server`
4. http://localhost:8000/build/a.html

harfbuzz-js however was doing also an [auto wrapping](https://github.com/prezi/harfbuzz-js/blob/master/src/main/coffee/connector.coffee#L301) of C functions, like to see how to do the same here.

The WASM output is around 543kb.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,29 +92,32 @@ add_definitions(-DHAVE_FALLBACK)
 
 
 ## Functions and headers
-include (CheckFunctionExists)
-include (CheckIncludeFile)
-macro (check_funcs) # Similar to AC_CHECK_FUNCS of autotools
-  foreach (func_name ${ARGN})
-    string(TOUPPER ${func_name} definiton_to_add)
-    check_function_exists(${func_name} HAVE_${definiton_to_add})
-    if (${HAVE_${definiton_to_add}})
-      add_definitions(-DHAVE_${definiton_to_add})
-    endif ()
-  endforeach ()
-endmacro ()
-check_funcs(atexit mprotect sysconf getpagesize mmap isatty newlocale strtod_l)
-check_include_file(unistd.h HAVE_UNISTD_H)
-if (${HAVE_UNISTD_H})
-  add_definitions(-DHAVE_UNISTD_H)
-endif ()
-check_include_file(sys/mman.h HAVE_SYS_MMAN_H)
-if (${HAVE_SYS_MMAN_H})
-  add_definitions(-DHAVE_SYS_MMAN_H)
-endif ()
-check_include_file(xlocale.h HAVE_XLOCALE_H)
-if (${HAVE_XLOCALE_H})
-  add_definitions(-DHAVE_XLOCALE_H)
+if (NOT EMSCRIPTEN) # time consuming and not needed for Emscripten builds
+  include (CheckFunctionExists)
+  include (CheckIncludeFile)
+  macro (check_funcs) # Similar to AC_CHECK_FUNCS of autotools
+    foreach (func_name ${ARGN})
+      string(TOUPPER ${func_name} definiton_to_add)
+      CHECK_FUNCTION_EXISTS(${func_name} HAVE_${definiton_to_add})
+      if (${HAVE_${definiton_to_add}})
+        add_definitions(-DHAVE_${definiton_to_add})
+      endif ()
+    endforeach ()
+  endmacro ()
+
+  check_funcs(atexit mprotect sysconf getpagesize mmap isatty newlocale strtod_l)
+  check_include_file(unistd.h HAVE_UNISTD_H)
+  if (${HAVE_UNISTD_H})
+    add_definitions(-DHAVE_UNISTD_H)
+  endif ()
+  check_include_file(sys/mman.h HAVE_SYS_MMAN_H)
+  if (${HAVE_SYS_MMAN_H})
+    add_definitions(-DHAVE_SYS_MMAN_H)
+  endif ()
+  check_include_file(xlocale.h HAVE_XLOCALE_H)
+  if (${HAVE_XLOCALE_H})
+    add_definitions(-DHAVE_XLOCALE_H)
+  endif ()
 endif ()
 
 
@@ -484,42 +487,56 @@ endif ()
 
 
 ## Atomic ops availability detection
-file(WRITE "${PROJECT_BINARY_DIR}/try_compile_intel_atomic_primitives.c"
-"		void memory_barrier (void) { __sync_synchronize (); }
-		int atomic_add (int *i) { return __sync_fetch_and_add (i, 1); }
-		int mutex_trylock (int *m) { return __sync_lock_test_and_set (m, 1); }
-		void mutex_unlock (int *m) { __sync_lock_release (m); }
-		int main () { return 0; }
-")
-try_compile(HB_HAVE_INTEL_ATOMIC_PRIMITIVES
-  ${PROJECT_BINARY_DIR}/try_compile_intel_atomic_primitives
-  ${PROJECT_BINARY_DIR}/try_compile_intel_atomic_primitives.c)
-if (HB_HAVE_INTEL_ATOMIC_PRIMITIVES)
-  add_definitions(-DHAVE_INTEL_ATOMIC_PRIMITIVES)
-endif ()
+if (NOT EMSCRIPTEN)
+  file(WRITE "${PROJECT_BINARY_DIR}/try_compile_intel_atomic_primitives.c"
+  "		void memory_barrier (void) { __sync_synchronize (); }
+      int atomic_add (int *i) { return __sync_fetch_and_add (i, 1); }
+      int mutex_trylock (int *m) { return __sync_lock_test_and_set (m, 1); }
+      void mutex_unlock (int *m) { __sync_lock_release (m); }
+      int main () { return 0; }
+  ")
+  try_compile(HB_HAVE_INTEL_ATOMIC_PRIMITIVES
+    ${PROJECT_BINARY_DIR}/try_compile_intel_atomic_primitives
+    ${PROJECT_BINARY_DIR}/try_compile_intel_atomic_primitives.c)
+  if (HB_HAVE_INTEL_ATOMIC_PRIMITIVES)
+    add_definitions(-DHAVE_INTEL_ATOMIC_PRIMITIVES)
+  endif ()
 
-file(WRITE "${PROJECT_BINARY_DIR}/try_compile_solaris_atomic_ops.c"
-"		#include <atomic.h>
-		/* This requires Solaris Studio 12.2 or newer: */
-		#include <mbarrier.h>
-		void memory_barrier (void) { __machine_rw_barrier (); }
-		int atomic_add (volatile unsigned *i) { return atomic_add_int_nv (i, 1); }
-		void *atomic_ptr_cmpxchg (volatile void **target, void *cmp, void *newval) { return atomic_cas_ptr (target, cmp, newval); }
-		int main () { return 0; }
-")
-try_compile(HB_HAVE_SOLARIS_ATOMIC_OPS
-  ${PROJECT_BINARY_DIR}/try_compile_solaris_atomic_ops
-  ${PROJECT_BINARY_DIR}/try_compile_solaris_atomic_ops.c)
-if (HB_HAVE_SOLARIS_ATOMIC_OPS)
-  add_definitions(-DHAVE_SOLARIS_ATOMIC_OPS)
+  file(WRITE "${PROJECT_BINARY_DIR}/try_compile_solaris_atomic_ops.c"
+  "		#include <atomic.h>
+      /* This requires Solaris Studio 12.2 or newer: */
+      #include <mbarrier.h>
+      void memory_barrier (void) { __machine_rw_barrier (); }
+      int atomic_add (volatile unsigned *i) { return atomic_add_int_nv (i, 1); }
+      void *atomic_ptr_cmpxchg (volatile void **target, void *cmp, void *newval) { return atomic_cas_ptr (target, cmp, newval); }
+      int main () { return 0; }
+  ")
+  try_compile(HB_HAVE_SOLARIS_ATOMIC_OPS
+    ${PROJECT_BINARY_DIR}/try_compile_solaris_atomic_ops
+    ${PROJECT_BINARY_DIR}/try_compile_solaris_atomic_ops.c)
+  if (HB_HAVE_SOLARIS_ATOMIC_OPS)
+    add_definitions(-DHAVE_SOLARIS_ATOMIC_OPS)
+  endif ()
+else ()
+  # emscripten doesn't need multithreading support, so
+  add_definitions(-DHB_NO_MT)
 endif ()
 
 
 ## Define harfbuzz library
-add_library(harfbuzz ${project_sources} ${project_extra_sources} ${project_headers})
+if (NOT EMSCRIPTEN) # define emscripten as a executable on emscripten mode
+  add_library(harfbuzz ${project_sources} ${project_extra_sources} ${project_headers})
+else ()
+  add_executable(harfbuzz ${project_sources} ${project_extra_sources} ${project_headers})
+  add_definitions("-DHB_EXTERN=__attribute__((used))")
+  add_definitions(-DHB_NO_MT)
+  set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -s MODULARIZE=1 \
+    -s 'EXTRA_EXPORTED_RUNTIME_METHODS=[\"ccall\", \"cwrap\"]' -s WASM=1 \
+    -fno-threadsafe-statics -fno-rtti -fno-exceptions -Oz")
+endif ()
 target_link_libraries(harfbuzz ${THIRD_PARTY_LIBS})
 
-if (UNIX OR MINGW)
+if ((UNIX OR MINGW) AND NOT EMSCRIPTEN)
   # Make symbols link locally
   link_libraries(-Bsymbolic-functions)
 
@@ -785,20 +802,24 @@ endif ()
 
 
 ## src/ executables
-foreach (prog main test test-would-substitute test-size-params test-buffer-serialize hb-ot-tag)
-  set (prog_name ${prog})
-  if (${prog_name} STREQUAL "test")
-    # test can not be used as a valid executable name on cmake, lets special case it
-    set (prog_name test-test)
-  endif ()
-  add_executable(${prog_name} ${PROJECT_SOURCE_DIR}/src/${prog}.cc)
-  target_link_libraries(${prog_name} harfbuzz ${THIRD_PARTY_LIBS})
-endforeach ()
-set_target_properties(hb-ot-tag PROPERTIES COMPILE_FLAGS "-DMAIN")
+# as harfbuzz should be built with add_executable for Emscripten builds, it is not
+# possible to use it as a library for the executables
+if (NOT EMSCRIPTEN)
+  foreach (prog main test test-would-substitute test-size-params test-buffer-serialize hb-ot-tag)
+    set (prog_name ${prog})
+    if (${prog_name} STREQUAL "test")
+      # test can not be used as a valid executable name on cmake, lets special case it
+      set (prog_name test-test)
+    endif ()
+    add_executable(${prog_name} ${PROJECT_SOURCE_DIR}/src/${prog}.cc)
+    target_link_libraries(${prog_name} harfbuzz ${THIRD_PARTY_LIBS})
+  endforeach ()
+  set_target_properties(hb-ot-tag PROPERTIES COMPILE_FLAGS "-DMAIN")
+endif ()
 
 
 ## Tests
-if (UNIX OR MINGW)
+if ((UNIX OR MINGW) AND NOT EMSCRIPTEN)
   if (BUILD_SHARED_LIBS)
     # generate harfbuzz.def after build completion
     string(REPLACE ";" " " space_separated_headers "${project_headers}")
@@ -829,4 +850,6 @@ endif ()
 
 # Needs to come last so that variables defined above are passed to
 # subdirectories.
-add_subdirectory(test)
+if (NOT EMSCRIPTEN)
+  add_subdirectory(test)
+endif ()


ebraminio:
Closure Tools flags was troublesome for CI thus removed but here was its code:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 5b52d84..6e2c54f 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -531,6 +531,14 @@ else ()
   add_definitions("-DHB_EXTERN=__attribute__((used))")
   add_definitions(-DHB_NO_MT)
   set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -s 'EXTRA_EXPORTED_RUNTIME_METHODS=[\"ccall\", \"cwrap\"]' -s WASM=1 -fno-rtti -fno-exceptions -Oz")
+
+  # If Java exists, use Google Closure Tool support
+  # Doesn't have any effect on harfbuzz.wasm output but on harfbuzz.js which
+  # is just an interface for the former.
+  find_package(Java)
+  if (Java_FOUND)
+    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -s MODULARIZE=1 --closure 1")
+  endif ()
 endif ()
 target_link_libraries(harfbuzz ${THIRD_PARTY_LIBS})
 
```

It seems the future is loading .wasm files without need of any JS interface but it needs more tweaks to run that.

```
<script>
    fetch('harfbuzz.wasm').then(response =>
        response.arrayBuffer()
    ).then(bytes => {
        return WebAssembly.instantiate(bytes, {
            'global': {},
            'env': {
                'memoryBase': 0,
                'tableBase': 0,
                'abort': () => {},
                'setTempRet0': () => {},
                'getTempRet0': () => {},
                '___assert_fail': () => {},
                '___divdi3': () => {},
                '___errno_location': () => {},
                '___muldi3': () => {},
                '_bitshift64Shl': () => {},
                '_bsearch': () => {},
                '_calloc': () => {},
                '_free': () => {},
                '_getenv': () => {},
                '_i64Add': () => {},
                '_i64Subtract': () => {},
                '_llvm_ctpop_i32': () => {},
                '_malloc': () => {},
                '_memcmp': () => {},
                '_memcpy': () => {},
                '_memmove': () => {},
                '_memset': () => {},
                '_qsort': () => {},
                '_realloc': () => {},
                '_roundf': () => {},
                '_setlocale': () => {},
                '_snprintf': () => {},
                '_strchr': () => {},
                '_strcmp': () => {},
                '_strlen': () => {},
                '_strncmp': () => {},
                '_strncpy': () => {},
                '_strstr': () => {},
                '_strtod': () => {},
                '_strtol': () => {},
                '_strtoul': () => {},
                '_vsnprintf': () => {},
                'memory': new WebAssembly.Memory({initial: 400, limit: 200}),
                'table': new WebAssembly.Table({initial: 400, element: 'anyfunc'})
            }
        });
    }).then(module => {
        console.log(module.instance.exports._hb_version_string());
    });
</script>
```

ebraminio:
Btw, why I have underscore prefixed symbols? I hope that is not because I am running Emscripten on macOS...

behdad:
Nice. Have you been able to bring up a harfbuzz emscripten demo? Would be great to have that.

kripken:
Running just a wasm file without JS is not practical currently, given wasm limitations. See the current docs [here](https://github.com/kripken/emscripten/wiki/WebAssembly-Standalone). The SIDE_MODULE option is also not yet recommended for general use.

The `_` prefixes are an emscripten convention (similar to some assembly language conventions).

ebraminio:
Thank you @kripken. The intention here to have a convenient support for possible js users (like the file opened [on emscripten](https://github.com/kripken/emscripten/issues/4754)) by setting a recommended set of flags, a lightweight wrapper and recommended way of building the repo (like the ones you have aggregated on @emscripten-ports, I like to have harfbuzz there also 😊), something like in-tree [harfbuzz-js](https://github.com/prezi/harfbuzz-js). I don't know yet how harfbuzz-js was invoking [cwrap](https://github.com/prezi/harfbuzz-js/blob/master/src/main/coffee/connector.coffee#L301) automatically on their work (or did they?) and that is also something we like to have (but with minimum change from our side :P).

@behdad, I could run hb_version_string as noted on first post, I've seen your works on glyphy and emscripten, what I however imagine is porting this https://github.com/ebraminio/kashida/blob/master/kashida.cc#L31-L63 (minus dwrite part :P) to this work, does that sound good?

ebraminio:
Also went for merging `add_executable(harfbuzz` with `add_library(harfbuzz` but that seems is very hacky.

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index f4200d5..ac329de 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -524,17 +524,22 @@ endif ()
 
 
 ## Define harfbuzz library
-if (NOT EMSCRIPTEN) # define emscripten as a executable on emscripten mode
-  add_library(harfbuzz ${project_sources} ${project_extra_sources} ${project_headers})
-else ()
-  add_executable(harfbuzz ${project_sources} ${project_extra_sources} ${project_headers})
+add_library(harfbuzz ${project_sources} ${project_extra_sources} ${project_headers})
+target_link_libraries(harfbuzz ${THIRD_PARTY_LIBS})
+
+
+## Emscripten
+if (EMSCRIPTEN)
   add_definitions("-DHB_EXTERN=__attribute__((used))")
   add_definitions(-DHB_NO_MT)
   set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} \
     -s 'EXTRA_EXPORTED_RUNTIME_METHODS=[\"ccall\", \"cwrap\"]' -s WASM=1 \
     -fno-threadsafe-statics -fno-rtti -fno-exceptions -Oz")
+  set(CMAKE_AR ${CMAKE_C_COMPILER})
+  set(CMAKE_STATIC_LIBRARY_SUFFIX ".js")
+  set(CMAKE_CXX_CREATE_STATIC_LIBRARY "<CMAKE_AR> -o <TARGET> <LINK_FLAGS> ${CMAKE_CXX_FLAGS} <OBJECTS>")
 endif ()
-target_link_libraries(harfbuzz ${THIRD_PARTY_LIBS})
+
 
 if ((UNIX OR MINGW) AND NOT EMSCRIPTEN)
   # Make symbols link locally
@@ -804,7 +809,7 @@ endif ()
 ## src/ executables
 # as harfbuzz should be built with add_executable for js build, it is not
 # possible to use it as a library for the executables
-if (NOT EMSCRIPTEN)
+#if (NOT EMSCRIPTEN)
   foreach (prog main test test-would-substitute test-size-params test-buffer-serialize hb-ot-tag)
     set (prog_name ${prog})
     if (${prog_name} STREQUAL "test")
@@ -815,7 +820,7 @@ if (NOT EMSCRIPTEN)
     target_link_libraries(${prog_name} harfbuzz ${THIRD_PARTY_LIBS})
   endforeach ()
   set_target_properties(hb-ot-tag PROPERTIES COMPILE_FLAGS "-DMAIN")
-endif ()
+#endif ()
 
 
 ## Tests
```

ebraminio:
> I don't know yet how harfbuzz-js was invoking cwrap automatically on their work (or do they?)

Not automatic: 
https://github.com/prezi/harfbuzz-js/blob/8310305c343a187172843a5e25ce24574fc768f1/src/main/coffee/hb-js.coffee#L233-L261 I wish it was however.

ebraminio:
Ok, making a complete JS interface and maintaining it is just to much of hassle and we should import these two first https://github.com/prezi/harfbuzz-js/tree/master/src/main/coffee or write our own to handle hb structs on JS which is somehow out of scope of harfbuzz project for now I guess.

So about this PR, we should either close it or if it does anything good (does it?), resolve its issues merge it, I guess. Which one do you suggest, @behdad and @kripken?

ebraminio:
Or maybe reduce all this into `if (EMSCRIPTEN) add_definitions(-fno-threadsafe-statics -fno-rtti -fno-exceptions -DHB_NO_MT) endif ()` the flags maybe not that obvious for an someone touches the library for the first time.

waywardmonkeys:
You should consider using `target_compile_definitions` instead of `add_definitions` perhaps ?

ebraminio:
We are using it [here](https://github.com/harfbuzz/harfbuzz/blob/72bec1cd09f919f3826345db4832752ec96ff5c5/test/fuzzing/CMakeLists.txt#L14) but in here we have only one target so that doesn't differ much. The thing I like to know is what we can and should do for making use of harfbuzz for [possible users of emscripten](https://github.com/kripken/emscripten/issues/4754) easier so they don't skip the use of harfbuzz. With cmake project it is already possible with `emcmake cmake -Bbuild -H. -GNinja && ninja -Cbuild` without any change but I had no success with autotools (perhaps I should try more). I like to see what can be done more than this and eventually having and keeping harfbuzz like the projects hosted on @emscripten-ports.

kripken:
> emscripten-ports, I like to have harfbuzz there also

That could make sense, yeah. Is this emitting an executable output or a library, though? (Ports contains libraries, to make it easy to link them in; for executables, there's not much benefit of putting them on ports.)

ebraminio:
The current harfbuzz cmake emits a library, a .a file rather than .js. If that is all what we can do, it is already there :)

kripken:
Sounds good to add it to ports, then. See previous ports for how to add it, e.g. https://github.com/kripken/emscripten/commit/5bd93efc75bc8b09d7b5faf253de37613db3ef4c I also created a team and repo for it in ports and invited you.

behdad:
Thank you all for working on this. :)

ebraminio:
Thank you @kripken, that is an honor!

Very little familiar with that embuxilder, I believe what would best is having all the needed changes in here so I started with this:

```diff
diff --git a/embuilder.py b/embuilder.py
index b2a1f73..c26c4a9 100755
--- a/embuilder.py
+++ b/embuilder.py
@@ -243,6 +243,8 @@ if operation == 'build':
       build_port('sdl2-net', 'libsdl2_net.bc', ['-s', 'USE_SDL=2', '-s', 'USE_SDL_NET=2'])
     elif what == 'freetype':
       build_port('freetype', 'libfreetype.a', ['-s', 'USE_FREETYPE=1'])
+    elif what == 'harfbuzz':
+      build_port('harfbuzz', 'libharfbuzz.a', ['-s', 'USE_HARFBUZZ=1'])
     elif what == 'sdl2-ttf':
       build_port('sdl2-ttf', 'libsdl2_ttf.bc', ['-s', 'USE_SDL=2', '-s', 'USE_SDL_TTF=2', '-s', 'USE_FREETYPE=1'])
     elif what == 'binaryen':
diff --git a/src/settings.js b/src/settings.js
index 04e1840..a1f9260 100644
--- a/src/settings.js
+++ b/src/settings.js
@@ -749,6 +749,7 @@ var USE_BULLET = 0; // 1 = use bullet from emscripten-ports
 var USE_VORBIS = 0; // 1 = use vorbis from emscripten-ports
 var USE_OGG = 0; // 1 = use ogg from emscripten-ports
 var USE_FREETYPE = 0; // 1 = use freetype from emscripten-ports
+var USE_HARFBUZZ = 0; // 1 = use harfbuzz from emscripten-ports
 var USE_COCOS2D = 0; // 3 = use cocos2d v3 from emscripten-ports
 
 var SDL2_IMAGE_FORMATS = []; // Formats to support in SDL2_image. Valid values: bmp, gif, lbm, pcx, png, pnm, tga, xcf, xpm, xv
diff --git a/tools/ports/__init__.py b/tools/ports/__init__.py
index c9a70be..64d55f1 100644
--- a/tools/ports/__init__.py
+++ b/tools/ports/__init__.py
@@ -1,7 +1,7 @@
-from . import binaryen, bullet, cocos2d, freetype, libpng, ogg, sdl, sdl_image, sdl_ttf, sdl_net, vorbis, zlib
+from . import binaryen, bullet, cocos2d, freetype, harfbuzz, libpng, ogg, sdl, sdl_image, sdl_ttf, sdl_net, vorbis, zlib
 
 # If port A depends on port B, then A should be _after_ B
-ports = [zlib, libpng, sdl, sdl_image, ogg, vorbis, bullet, freetype, sdl_ttf, sdl_net, binaryen, cocos2d]
+ports = [zlib, libpng, sdl, sdl_image, ogg, vorbis, bullet, freetype, harfbuzz, sdl_ttf, sdl_net, binaryen, cocos2d]
 
 ports_by_name = {}
 for port in ports:
diff --git a/tools/ports/harfbuzz.py b/tools/ports/harfbuzz.py
new file mode 100644
index 0000000..109772d
--- /dev/null
+++ b/tools/ports/harfbuzz.py
@@ -0,0 +1,38 @@
+import os, shutil, logging, subprocess
+
+TAG = '1.7.5'
+
+def get(ports, settings, shared):
+  if settings.USE_HARFBUZZ == 1:
+    ports.fetch_project('harfbuzz', 'https://github.com/harfbuzz/harfbuzz/archive/' + TAG + '.zip', 'harfbuzz-' + TAG)
+    def create():
+      logging.info('building port: harfbuzz')
+
+      source_path = os.path.join(ports.get_dir(), 'harfbuzz', 'harfbuzz-' + TAG)
+      dest_path = os.path.join(shared.Cache.get_path('ports-builds'), 'harfbuzz')
+
+      shutil.rmtree(dest_path, ignore_errors=True)
+      shutil.copytree(source_path, dest_path)
+
+      cmake_build_type = 'Release'
+      subprocess.check_call(['cmake', '-Bbuild', '-H.', '-DCMAKE_BUILD_TYPE=' + cmake_build_type])
+      subprocess.check_call(['make', '-Cbuild'])
+      return os.path.join(ports.get_build_dir(), 'harfbuzz', 'harfbuzz', 'build', 'libharfbuzz.a')
+    return [shared.Cache.get('harfbuzz', create, what='port')]
+  else:
+    return []
+
+
+def process_dependencies(settings):
+  #if settings.USE_HARFBUZZ == 1:
+  #  settings.USE_FREETYPE = 1
+  pass
+
+def process_args(ports, args, settings, shared):
+  if settings.USE_HARFBUZZ == 1:
+    get(ports, settings, shared)
+    args += ['-Xclang', '--sysroot' + os.path.join(shared.Cache.get_path('ports-builds'), 'harfbuzz')]
+  return args
+
+def show():
+  return 'harfbuzz (USE_HARFBUZZ=1; MIT license)'
```

But I am getting with `./embuilder.py build harfbuzz`:
```
OSError: [Errno 2] No such file or directory: '/Users/ebrahim/.emscripten_ports/harfbuzz/harfbuzz-1.7.5'
  File "./embuilder.py", line 247, in <module>
    build_port('harfbuzz', 'libharfbuzz.a', ['-s', 'USE_HARFBUZZ=1'])
  File "./embuilder.py", line 136, in build_port
    build(C_BARE, [os.path.join('ports-builds', port_name, lib_name)] if lib_name else None, params)
  File "./embuilder.py", line 126, in build
    shared.Building.emcc(temp, args, output_filename=temp_js)
  File "/Users/ebrahim/Desktop/emscripten/tools/shared.py", line 1986, in emcc
    assert os.path.exists(output_filename), 'emcc could not create output file: ' + output_filename
AssertionError: emcc could not create output file: /tmp/tmpjClsek.js
```

I have several issues with using harfbuzz releases here, I can't use our .tar.bz2 releases which do not need `ragel` be installed so using GitHub generated source git which need `ragel` and harfbuzz cmake doesn't support in-tree cmake make file generation (so it should be done on a separate folder like build/)

I know if I use https://github.com/emscripten-ports/HarfBuzz I can resolve most of the issues however I like to keep our repository also updated with emscripten needs (as we are also running a CI job with emscripten) so what do you suggest?

CMakeLists.txt
```diff
@@ -92,29 +92,32 @@ add_definitions(-DHAVE_FALLBACK)
...
+  include (CheckIncludeFile)
+  macro (check_funcs) # Similar to AC_CHECK_FUNCS of autotools
+    foreach (func_name ${ARGN})
+      string(TOUPPER ${func_name} definiton_to_add)
+      CHECK_FUNCTION_EXISTS(${func_name} HAVE_${definiton_to_add})
+      if (${HAVE_${definiton_to_add}})
+        add_definitions(-DHAVE_${definiton_to_add})
+      endif ()
+    endforeach ()
+  endmacro ()
```
juj: The `include` and `macro` lines only add generic functions, so may be worthwhile to hoist them out from the `if (NOT EMSCRIPTEN)` path and restrict the Emscripten-specific changes to as few lines as possible - probably only over the `check_funcs` lines? That way there won't be any misassumptions that something e.g. in the `check_funcs` macro would be incompatible with Emscripten.

Also if this was done like the comment suggests only to speed up Emscripten CMake runs, then I think it might be desirable to remove this `if (NOT EMSCRIPTEN)` part altogether, and arrive to the same speed wins by pre-specifying all the `-DHAVE_UNISTD_H=1` etc. from the CMake command line. If I understand correctly, CMake only does these checks if the variables have not yet been checked, so pre-setting the results on the command line should have the same effect.

There is a chance that in the distant future the Emscripten environment changes and one of the headers, like `xlocale.h` might begin to exist, so there's also a chance that avoiding all this configuration would lead to incorrect results?

CMakeLists.txt
```diff
@@ -484,42 +487,56 @@ endif ()
...
-		int atomic_add (volatile unsigned *i) { return atomic_add_int_nv (i, 1); }
-		void *atomic_ptr_cmpxchg (volatile void **target, void *cmp, void *newval) { return atomic_cas_ptr (target, cmp, newval); }
-		int main () { return 0; }
-")
-try_compile(HB_HAVE_SOLARIS_ATOMIC_OPS
-  ${PROJECT_BINARY_DIR}/try_compile_solaris_atomic_ops
-  ${PROJECT_BINARY_DIR}/try_compile_solaris_atomic_ops.c)
-if (HB_HAVE_SOLARIS_ATOMIC_OPS)
-  add_definitions(-DHAVE_SOLARIS_ATOMIC_OPS)
+if (NOT EMSCRIPTEN)
```
juj: This was also set to be avoided in Emscripten, but there's no mention as to if the reason is the same as above, i.e. to speed up CMake, or if there is some incompatibility?

In Emscripten the atomic functions below are supported, so I wonder if there was some build issue or similar?

CMakeLists.txt
```diff
@@ -484,42 +487,56 @@ endif ()
...
+      int main () { return 0; }
+  ")
+  try_compile(HB_HAVE_SOLARIS_ATOMIC_OPS
+    ${PROJECT_BINARY_DIR}/try_compile_solaris_atomic_ops
+    ${PROJECT_BINARY_DIR}/try_compile_solaris_atomic_ops.c)
+  if (HB_HAVE_SOLARIS_ATOMIC_OPS)
+    add_definitions(-DHAVE_SOLARIS_ATOMIC_OPS)
+  endif ()
+else ()
+  # emscripten doesn't need multithreading support, so
```
juj: There is multithreading support in Emscripten, enabled when `-s USE_PTHREADS=1` is specified. In such mode, Harfbuzz could be compiled multithreaded as well. It may be worthwhile to allow multithreading support to be configured on the command line for this?

CMakeLists.txt
```diff
@@ -484,42 +487,56 @@ endif ()
...
+  add_definitions(-DHB_NO_MT)
 endif ()
 
 
 ## Define harfbuzz library
-add_library(harfbuzz ${project_sources} ${project_extra_sources} ${project_headers})
+if (NOT EMSCRIPTEN) # define emscripten as a executable on emscripten mode
```
juj: Hmm, what is this about library vs executable about?

CMakeLists.txt
```diff
@@ -484,42 +487,56 @@ endif ()
...
+  add_library(harfbuzz ${project_sources} ${project_extra_sources} ${project_headers})
+else ()
+  add_executable(harfbuzz ${project_sources} ${project_extra_sources} ${project_headers})
+  add_definitions("-DHB_EXTERN=__attribute__((used))")
+  add_definitions(-DHB_NO_MT)
```
juj: This line looks like duplicated to above, adding `-DHB_NO_MT` twice?

CMakeLists.txt
```diff
@@ -484,42 +487,56 @@ endif ()
...
+  set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -s MODULARIZE=1 \
+    -s 'EXTRA_EXPORTED_RUNTIME_METHODS=[\"ccall\", \"cwrap\"]' -s WASM=1 \
+    -fno-threadsafe-statics -fno-rtti -fno-exceptions -Oz")
```
juj: This line looks odd from Emscripten perspective: In CMake `CMAKE_CXX_FLAGS` specifies compilation time flags, and I think these flags will not be present at link time(?)

In Emscripten when `-s 'EXTRA_EXPORTED_RUNTIME_METHODS=[\"ccall\", \"cwrap\"]'` or `-s MODULARIZE=1` are present at compile time, they have no effect, instead these two are only link time flags.

The other flags, especially `-s WASM=1` apply at both compile and link time.

Specifying `-Oz` here seems slightly nonidiomatic use of CMake. For CMake, there's four canonical CMake build modes, Debug, RelWithDebInfo, Release and MinSizeRel. The CMake default iirc is that Debug is `-O0`, and Release and RelWithDebInfo are `-O3`, and MinSizeRel is `-Oz`. Specifying `-Oz` here will override the optimization mode even for debug and release builds. I wonder if this was intentional here to avoid some codegen bug, or performance issue, or similar? Perhaps debug builds still should not apply `-Oz`, but would be better with the default `-O0`?

CMakeLists.txt
```diff
@@ -484,42 +487,56 @@ endif ()
...
+endif ()
 target_link_libraries(harfbuzz ${THIRD_PARTY_LIBS})
 
-if (UNIX OR MINGW)
+if ((UNIX OR MINGW) AND NOT EMSCRIPTEN)
```
juj: This happens since in Emscripten we define `UNIX` in CMake to make Emscripten look like an easy drop-in replacement to native Linux GCC-style environments, so the extra `NOT EMSCRIPTEN` check is indeed unfortunately needed :/

CMakeLists.txt
```diff
@@ -785,20 +802,24 @@ endif ()
...
-  set (prog_name ${prog})
-  if (${prog_name} STREQUAL "test")
-    # test can not be used as a valid executable name on cmake, lets special case it
-    set (prog_name test-test)
-  endif ()
-  add_executable(${prog_name} ${PROJECT_SOURCE_DIR}/src/${prog}.cc)
-  target_link_libraries(${prog_name} harfbuzz ${THIRD_PARTY_LIBS})
-endforeach ()
-set_target_properties(hb-ot-tag PROPERTIES COMPILE_FLAGS "-DMAIN")
+# as harfbuzz should be built with add_executable for Emscripten builds, it is not
```
juj: It should be possible to build Harfbuzz as a static library with Emscripten, though I wonder if the use case here was something that wanted to do different? In general I'd recommend having the same static library vs shared library vs executable structures on all platforms that CMake is targeting, and, let's say if Emscripten wants to do an executable, then do it in a way that other platforms can also build executables, i.e. decouple the target type from the platform. Or was there some specific issue here?

CMakeLists.txt
ebraminio: Thanks for you review! Done.

CMakeLists.txt
ebraminio: It is not that needed for harfbuzz but if will be useful you think. Ok, lets leave it alone :)

CMakeLists.txt
ebraminio: Makes sense. Thanks

CMakeLists.txt
ebraminio: You are right. Whole thing gone.

CMakeLists.txt
ebraminio: I was thinking about some harfbuzz-js like in-tree support and providing things like this https://github.com/prezi/harfbuzz-js/blob/master/src/main/coffee/connector.coffee but lets provide just a libharfubzz.a for now. Whole thing gone.

CMakeLists.txt
ebraminio: Honestly it was just an experiment, whole thing removed.

CMakeLists.txt
ebraminio: No problem actually lets reuse emscripten here, I added an implicit EMSCRIPTEN also let me know if that doesn't look good to you :)

CMakeLists.txt
ebraminio: This was just experiment to see what we else do here, removed.

ebraminio:
@juj: Thank you for all of your reviews. You know the idea to this patch was to see what we can offer to emscripten more than what we have already right now, something like harfbuzz-js and https://github.com/prezi/harfbuzz-js/tree/master/src/main/coffee

With applying all you reviews, nothing remains from this patch except

```diff
-if (UNIX OR MINGW)
+if (UNIX OR MINGW OR EMSCRIPTEN) # not needed just to emphasize emscripten will go this way
   # Make symbols link locally
   link_libraries(-Bsymbolic-functions)

@@ -554,9 +552,7 @@ if (UNIX OR MINGW)
   endif ()

   # No threadsafe statics as we do it ourselves
-  if (BUILD_SHARED_LIBS)
-    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-threadsafe-statics")
-  endif ()
+  set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-threadsafe-statics")
```

which is not that essential obviously.

What we can do is adding harfbuzz to emscripten's embuilder with this patch https://github.com/harfbuzz/harfbuzz/pull/743#issuecomment-363899141 but honestly I can not run that and am not that familiar with embuilder, lets close this [unnecessary] changes and continue on https://github.com/kripken/emscripten/pull/6222 :)

PR #745, opened 2018-02-03
Interface for hb_subset, skeleton for the hb-subset cli, and basic testing rigging.

garretrieger:
Adds an interface for hb_subset and a minimal implementation (just returns input font unmodified). Also adds the start of a cli tool which calls hb_subset. Lastly, adds the start of a unit test for hb_subset and an integration test which can compare the results of subsetting to the output of fontTools.


ebraminio:
Kodus and ❤️ for not ignoring CMake :) I know it may feel as an extra work however we wouldn't have a [VS](https://ci.appveyor.com/project/harfbuzz/harfbuzz/build/1.0.384/job/gd08fvf96qsbrq0q#L1147) build without it which can be used on Chrome/Chromium eventually.

ebraminio:
For Windows builds, please rebase the change with master (because of #747) and wrap unistd inclusion with `#ifdef HAVE_UNISTD_H`.

PR #750, opened 2018-02-04
Switch from using custom version.xml to using gtkdocentities.

waywardmonkeys:
`docs/xml/gtkdocentities.ent` is already created by `gtkdoc.make`
and contains the version number among other things. We can just
generate this single file and use it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -501,7 +501,6 @@ test/shaping/data/Makefile
 test/shaping/data/in-house/Makefile
 test/shaping/data/text-rendering-tests/Makefile
 docs/Makefile
-docs/version.xml
 ])
 
 AC_OUTPUT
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
This usage of `gtkdocentities.ent` requires gtk doc 1.25 and higher. Not sure if that is an issue or not.

khaledhosny:
We use Trusty on Travis which [seems to have](https://launchpad.net/ubuntu/+source/gtk-doc) gtk-doc 1.20, so I depending on 1.25 would break generating the online documentation, and the [build logs](https://travis-ci.org/harfbuzz/harfbuzz/jobs/337144806#L1676) suggests so.

waywardmonkeys:
I will look for a more past-proof version of this change then... or I knew why running gtkdoc  is cmake generated such different output (which args, etc)

behdad:
> We use Trusty on Travis which seems to have gtk-doc 1.20, so I depending on 1.25 would break generating the online documentation, and the build logs suggests so.

Humm. How come the bots are all green now?

Issue #424, opened 2017-02-19
autotools warnings


eroux:
when running `autoreconf -fi` (or `autogen.sh`) with `autoreconf (GNU Autoconf) 2.69` I get several

```
configure.ac:22: warning: AC_RUN_IFELSE was called before AC_USE_SYSTEM_EXTENSIONS
../../lib/autoconf/specific.m4:368: AC_USE_SYSTEM_EXTENSIONS is expanded from...
configure.ac:22: the top level
```

working on a PR

behdad:
Ignore it.  I fixed that before. Broke automake on my version of Ubuntu.  Will eventually fix it when we don't have to support Ubuntu 14.04

eroux:
oh ok thanks!

behdad:
Closing since not actionable currently, and works.

PR #751, opened 2018-02-04
Fix autoconf warning re: AC_USE_SYSTEM_EXTENSIONS

waywardmonkeys:
`AC_USE_SYSTEM_EXTENSIONS` must be set before anything using the
system compilers, so move the initialization a bit earlier.


khaledhosny:
Fixing this breaks it with old autotools versions, see https://github.com/harfbuzz/harfbuzz/issues/424.

coltongit:
perhaps a better way to fix this would be to check for the last version of autotools that this breaks make distcheck on, and have the AC_USE_SYSTEM_EXTENSIONS be defined in different spots dependinf on said version

behdad:
We're moving to meson so it feels like no action needed here still.

PR #758, opened 2018-02-06
[test] Fix test-subset on old distros

ebraminio:
https://developer.gnome.org/glib/stable/glib-Testing.html#g-assert-cmpmem


ebraminio:
Travis revived!

PR #769, opened 2018-02-10
Fix symbol exporting and libstdc++ linkage of hb-subset

### LINKED ISSUE #767: Update src/check-*.sh for harfbuzz-subset.so
Some of those checks currently only check libharfbuzz.so; they should be updated to also check harfbuzz-subset, and possibly harfbuzz-icu as well.

For example, harfbuzz-subset has many internal symbols leaking into the binary ABI right now. The tests are not catching that.

ebraminio:
Fixes #767.

TODO:
* src/check-defs.sh
* libharfbuzz_gobject_la_LINK = $(libharfbuzz_la_LINK)
* Separate -icu and -subset on CMake also.

### DIFF
--- a/src/check-symbols.sh
+++ b/src/check-symbols.sh
@@ -17,24 +17,26 @@ fi
 
 echo "Checking that we are not exposing internal symbols"
 tested=false
-for suffix in so dylib; do
-	so=$libs/libharfbuzz.$suffix
-	if ! test -f "$so"; then continue; fi
+for soname in harfbuzz harfbuzz-icu harfbuzz-subset; do
+	for suffix in so dylib; do
+		so=$libs/lib$soname.$suffix
+		if ! test -f "$so"; then continue; fi
 
-	EXPORTED_SYMBOLS="`nm "$so" | grep ' [BCDGINRSTVW] .' | grep -v ' _fini\>\| _init\>\| _fdata\>\| _ftext\>\| _fbss\>\| __bss_start\>\| __bss_start__\>\| __bss_end__\>\| _edata\>\| _end\>\| _bss_end__\>\| __end__\>\| __gcov_flush\>\| ___gcov_flush\>\| llvm_\| _llvm_' | cut -d' ' -f3`"
+		EXPORTED_SYMBOLS="`nm "$so" | grep ' [BCDGINRSTVW] .' | grep -v ' _fini\>\| _init\>\| _fdata\>\| _ftext\>\| _fbss\>\| __bss_start\>\| __bss_start__\>\| __bss_end__\>\| _edata\>\| _end\>\| _bss_end__\>\| __end__\>\| __gcov_flush\>\| ___gcov_flush\>\| llvm_\| _llvm_' | cut -d' ' -f3`"
 
-	prefix=`basename "$so" | sed 's/libharfbuzz/hb/; s/-/_/g; s/[.].*//'`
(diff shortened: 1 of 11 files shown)


src/check-symbols.sh
```diff
@@ -17,24 +17,26 @@ fi
...
+		# On macOS, C symbols are prefixed with _
+		if test $suffix = dylib; then prefix="_$prefix"; fi
 
-	echo "Processing $so"
-	if echo "$EXPORTED_SYMBOLS" | grep -v "^${prefix}_"; then
-		echo "Ouch, internal symbols exposed"
-		stat=1
-	fi
+		echo "Processing $so"
+		if echo "$EXPORTED_SYMBOLS" | grep -v "^${prefix}\(_\|$\)"; then
```
ebraminio: Either end of line or _, allows hb_subset function on hb-subset

behdad: Agreed.

behdad:
Thanks for working on this.

behdad:
> libharfbuzz_gobject_la_LINK = $(libharfbuzz_la_LINK)

Was there any difficulty with that one?

behdad:
Please merge if you feel this is ready.

ebraminio:
> Was there any difficulty with that one?

Honestly no, I wanted to unify all the linker and get feedback from the bots.

Btw, with all the bots fails, it seems I've opened a can of worm :)

behdad:
Yeah there's so many of them now, one or the other is failing all the time...

ebraminio:
> Yeah there's so many of them now, one or the other is failing all the time...

I believe that is not a bad sign :) Specially some of them are telling some interesting things others won't. It seems you have just applied that break more bots. So lets merge this and leave libstdc++ test to you guys (is the only thing this introduce AFAICT) :)  It doesn't happen on macOS (thus locally for me) so you guys will be more efficient on working on it I believe and instead I will work on src/check-defs.sh :)

ebraminio:
But I am also thinking about this, if all these addition of bots (in retrospect, mostly added by me :P) are making more issues, just feel free to kill some or all of them. What I like about harfbuzz is it can be developed without need of a beefy hardware thus making me able to compile it everyday using my 2011 device and we should keep the tradition of keeping things easy to use :)

behdad:
Yeah totally. I only use a light laptop...
Bots are catching all kinds of interesting things.  I wonder why gcc doesn't catch those. Or why each bot seem to catch a different issue :)).

behdad:
My biggest problem with the CircleCI bots is that the failure email I get doesn't clearly specify which bot it is at the top.

ebraminio:
> Or why each bot seem to catch a different issue

I saw this exactly on this PR, alpine (which had a symbol exposing issue last time) hadn't issue, the rest of clan, had. (was about [this](https://github.com/harfbuzz/harfbuzz/pull/769/files#diff-4dc80722f5281fa888134afdbfca3de1L111) if I remember correctly)

> My biggest problem with the CircleCI bots is that the failure email I get doesn't clearly specify which bot it is at the top.

Oh just go to https://circleci.com/gh/harfbuzz/harfbuzz and press unfollow at top right, I hope that wouldn't have a bad effect, lets test at least.

behdad:
> Oh just go to https://circleci.com/gh/harfbuzz/harfbuzz and press unfollow at top right, I hope that wouldn't have a bad effect, lets test at least.

Oh I want the emails.  I just want to immediately see which bot it is.

PR #770, opened 2018-02-10
Add --with-stdcpp for ossfuzzer and mupdf use

ebraminio:
@behdad, not exactly an expert on autotools, is this even a correct path? Should I use AS_IF here?

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -30,6 +30,14 @@ HBSOURCES += $(HB_BASE_RAGEL_GENERATED_sources)
 HBHEADERS = $(HB_BASE_headers)
 HBNODISTHEADERS = $(HB_NODIST_headers)
 
+if !WITHSTDCPP
+# Make sure we don't link to libstdc++
+HBCFLAGS += -fno-rtti -fno-exceptions
+
+# No threadsafe statics and C++ as we do it ourselves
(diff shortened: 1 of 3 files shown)


behdad:
Something like that. I'd phrase the help as "Do not try suppressing linkage to libstdc++".
We probably need an AM_CONDITIONAL and not try linking with C linker in hb/src/Makefile.am.

ebraminio:
Not testes yet even but I guess now I am on right track?

ebraminio:
Ok, I guess now everything theoretically should work.

ebraminio:
PTAL

src/Makefile.am
```diff
@@ -116,13 +124,17 @@ export_symbols = -export-symbols harfbuzz.def
 harfbuzz_def_dependency = harfbuzz.def
 libharfbuzz_la_LINK = $(CXXLINK) $(libharfbuzz_la_LDFLAGS)
 else
-# Use a C linker for GCC, not C++; Don't link to libstdc++
+if WITHSTDCPP
+libhb_ucdn_la_LINK = $(CXXLINK)
```
behdad: hb_ucdn line is wrong here.

behdad:
Thanks. Left a couple of comments. Lgtm otherwise.

PR #771, opened 2018-02-10
Hide more symbols of hb-subset

ebraminio:


ebraminio:
@behdad, is this the correct path? or I should've used `inline` having the same effect with `-fvisibility-inlines-hidden`?

ebraminio:
Our "DOS" bot also has a different complains :) https://circleci.com/gh/harfbuzz/harfbuzz/6894

ebraminio:
Doesn't revive much but moves main complains to https://travis-ci.org/harfbuzz/harfbuzz/jobs/339949244#L2480

behdad:
I pushed more fixes.

Also, you should have just used static here, not HB_INTERNAL. Fixed.

PR #775, opened 2018-02-11
Initialize bad_indices on hb-subset

### LINKED ISSUE #774: Hitting a GCC compiler optimization?
Some of our Linux bots are failing on test-subset and it seems if you build the project with -O0 it compiles just fine:

    CFLAGS="-g -O0" CXXFLAGS="-g -O0" ./autogen.sh --with-freetype --with-glib --with-gobject --with-cairo --with-icu --with-graphite2 && make check

but not when it is compiled with -O1 (or default, -O2). Maybe something should be marked as volatile?

One of them I am checking (ArchLinux) has GCC 7.3.

After building with O1, `gdb test/api/.libs/lt-test-subset` says:

```
[root@777c592dcd67 project]# gdb test/api/.libs/lt-test-subset
GNU gdb (GDB) 8.1
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from test/api/.libs/lt-test-subset...done.
(gdb) run
Starting program: /root/project/test/api/.libs/lt-test-subset 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
/subset: 
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff792ea9f in _populate_gids_to_retain (old_gids_sorted=..., old_gids=..., codepoints=..., face=<optimized out>) at hb-subset-plan.cc:95
95	    unsigned int i = bad_indices[bad_indices.len - 1];
(gdb) print i
$1 = <optimized out>
(gdb) print bad_indices
$2 = {len = 1433796640, allocated = <optimized out>, array = <optimized out>, static_array = {<optimized out> <repeats 16 times>}}
```

(well, just on double look, spotted the interesting len :)

ebraminio:
Fixes #774 and hopefully brings back some of the bots.

### DIFF
--- a/src/hb-subset-plan.cc
+++ b/src/hb-subset-plan.cc
@@ -76,6 +76,7 @@ _populate_gids_to_retain (hb_face_t *face,
   cmap.init (face);
 
   hb_prealloced_array_t<unsigned int> bad_indices;
+  bad_indices.init ();
 
   old_gids.alloc (codepoints.len);
   bool has_zero = false;


ebraminio:
I can see why this happened (https://github.com/harfbuzz/harfbuzz/commit/c31fcf4c58d96eb7d9781a986991b1a79ac7be44), how `hb_auto_array_t` could do this automatically, and I guess there are some other places we have uninitialized variables (like [this](https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ot-cmap-table.hh#L594), I don't see any indication of initialization). Hopefully we will address all of them.

PR #776, opened 2018-02-11
Fix SunCC/ODS compile on CMake

ebraminio:
This obviously doesn't make sense but I tried other combinations with no success.

### DIFF
--- a/src/hb-private.hh
+++ b/src/hb-private.hh
@@ -122,6 +122,7 @@ extern "C" void  hb_free_impl(void *ptr);
 #ifndef HB_INTERNAL
 # ifdef __SUNPRO_CC
 #  define HB_INTERNAL __hidden
+#  define HB_NO_VISIBILITY 1
 # elif !defined(__MINGW32__) && !defined(__CYGWIN__)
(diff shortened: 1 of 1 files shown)


behdad:
I was wondering why the __hidden isn't working...

ebraminio:
It doesn't work without __hidden. Interestingly it is similar with MSVC fail, maybe `extern HB_INTERNAL const void * const _hb_NullPool[HB_NULL_POOL_SIZE / sizeof (void *)];` are confusing them?

ebraminio:
So should this be merged or we should get more confident what is going on before the merge? (or ask Oracle guys?)

behdad:
This patch doesn't look right.

Ok, maybe the Sun compiler does not hide data members.  I have another idea for how to fix this and the Windows failure.  For Windows, though, wouldn't using the harfbuzz.def thing fix it?  We also need to produce harfbuzz-subset.def by the way.

behdad:
I ssh'ed and checked. __hidden doesn't seem to do anything on this compiler on this platform (Linux). I'm removing it.

PR #778, opened 2018-02-11
Add Travis based Coverity trigger

ebraminio:


ebraminio:
The caveat, IIUC, is you should push coverity_scan manually as:

    # Pattern to match selecting branches that will run analysis. We recommend leaving this set to 'coverity_scan'.
    # Take care in resource usage, and consider the build frequency allowances per
    #   https://scan.coverity.com/faq#frequency
    branch_pattern: coverity_scan

Is using master is OK with coverity guys? If not I am not sure whether current time to time is better than pushing to that branch.

behdad:
Looks like we are allowed 3 builds a day. How about we were some code for our Travis to push master to coverity_scan branch once a day? Oh, I read we can even set cron jobs on Travis now? Or in confusing with something else?

ebraminio:
Ok, I guess now everything theoretically should work.

ebraminio:
PTAL

behdad:
Thanks! Please go ahead and set this up.

ebraminio:
Lets see how it goes

pwithnall:
Assuming you’re using the old Coverity project for results, rather than the new one (I strongly suggest using the old one, in order to keep your result, triage and statistical history), then everything seems set for you to maintain the Coverity scans for Harfbuzz yourselves. I’ll step back from providing them. :-)

PR #780, opened 2018-02-12
Apply check-defs.sh to -icu and -subset, toward mingw fix

ebraminio:


ebraminio:
I guess it should but src/gen-def.py should be adopted with it.

ebraminio:
And I was worried that it currently doesn't. Doesn't anyone noticed any issue with -icu.dll?

behdad:
Quite possibly no one ever used -icu.dll. I think people only use hb-shape / hb-view from our Windows builds. All other uses integrate the sources into their project directly (Firefox, Chrome, ...)

PR #781, opened 2018-02-12
Fix symbol exporting issue on MINGW

ebraminio:
This brings all the bots, [hopefully], yay! Of course this part can be refactored to look better.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -126,19 +126,23 @@ export_symbols_icu = -export-symbols harfbuzz-icu.def
 harfbuzz_icu_def_dependency = harfbuzz-icu.def
 export_symbols_subset = -export-symbols harfbuzz-subset.def
 harfbuzz_subset_def_dependency = harfbuzz-subset.def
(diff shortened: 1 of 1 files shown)


ebraminio:
I wonder also if our libharfbuzz-icu.0.dll has ever correct symbol exporting. The same should be done about -gobject.dll also AFAIK.

src/Makefile.am
```diff
@@ -126,20 +126,24 @@ export_symbols_icu = -export-symbols harfbuzz-icu.def
...
-libharfbuzz_la_LINK = $(LINK) $(libharfbuzz_la_LDFLAGS)
+choosed_linker = $(LINK)
 else
-libharfbuzz_la_LINK = $(CXXLINK) $(libharfbuzz_la_LDFLAGS)
+choosed_linker = $(CXXLINK)
 endif
 endif
 endif
 
+libharfbuzz_la_LINK = $(choosed_linker) $(libharfbuzz_la_LDFLAGS)
```
behdad: Can we put these lines back next to the rest of their respective library definitions?

PR #782, opened 2018-02-12
Rename --with-stdcpp → --with-libstdc++

khaledhosny:
Also improve wording a bit and fix make check in `--with-libstdc++=yes`.


ebraminio:
I am OK with this I just saw some weirdness with use + on argument, as the autocomplete feature (my zsh shell) couldn't work well with this I guess. Honestly I didn't know about autocompletion of configure files even and learned that from you :) I see something like this:

screen shot 2018-02-12 at 11 09 32 pm

khaledhosny:
I don’t use zsh, but auto-completion in bash is working just fine here.

PR #783, opened 2018-02-12
Move linker specifying to a better place

ebraminio:


src/Makefile.am
```diff
@@ -207,7 +206,6 @@ EXTRA_DIST += harfbuzz-icu.pc.in
 
 if HAVE_GOBJECT
 lib_LTLIBRARIES += libharfbuzz-gobject.la
-libharfbuzz_gobject_la_LINK = $(libharfbuzz_la_LINK)
```
ebraminio: I added this but am worried that it makes more issue than solve. Perhaps CI bots can show whether it is useful a change or better to reverted to its previous state.

ebraminio: Removed this part of the patch. I am still not sure how harfbuzz-gobject.dll will work.

PR #789, opened 2018-02-14
Add .collect_logs and use it on CI bots, fixes #786

### LINKED ISSUE #786: Better printing of logs for CI bots
Right now when a build fails we do ```(cat `find -name '*.log'` && false)```. We should instead write a script `.collect_logs` or something and run that. That script should print the log filename before the contents. Also, not print configure.log which seems to be unnecessary most of the time.

ebraminio:


ebraminio:
@behdad, is this looking good? Some bikeshedding, `echo '====' $f '===='` is good enough?

behdad:
Don't you need `./.collect_logs` instead of `.collect_logs`?

Also, we mostly use dash instead of underscores in file names.

behdad:
lgtm otherwise.

ebraminio:
> Don't you need ./.collect_logs instead of .collect_logs?

How I didn't see it! Very sorry, I wanted to do some smoke test with didn't due to laziness!

PR #794, opened 2018-02-15
 Simplify calls to collect-logs.sh 

khaledhosny:


ebraminio:
While the script is on `.ci/` folder, this is a good idea ✅.

The working directory will be reset on each `- run:` command, probably this is the issue with most of the bots.

.circleci/config.yml
```diff
@@ -37,7 +42,10 @@ jobs:
       - checkout
       - run: dnf install -y pkg-config ragel gcc gcc-c++ automake autoconf libtool make which glib2-devel freetype-devel cairo-devel libicu-devel gobject-introspection-devel graphite2-devel redhat-rpm-config python || true
       - run: NOCONFIGURE=1 ./autogen.sh --with-freetype --with-glib --with-gobject --with-cairo --with-icu --with-graphite2
-      - run: mkdir build && cd build && ../configure && make && (make check || (.ci/collect-logs.sh && false))
+      - run: mkdir build && cd build
+      - run: ../configure
```
ebraminio: Working directory won't be preserved thus better to keep this on one line.

.circleci/config.yml
```diff
@@ -9,7 +9,10 @@ jobs:
       - checkout
       - run: apt update && apt install -y ninja-build binutils libtool autoconf automake make cmake gcc g++ pkg-config ragel gtk-doc-tools libfreetype6-dev libglib2.0-dev libcairo2-dev libicu-dev libgraphite2-dev python python-pip
       - run: pip install fonttools
-      - run: ./autogen.sh && (make distcheck || (cd harfbuzz-* && ../.ci/collect-logs.sh && false)) && rm -rf harfbuzz-*
+      - run: ./autogen.sh
+      - run: make
```
ebraminio: I think `make` before would be unnecessary or better to test `distcheck` before running a `make`.

behdad:
Do "make distcheck" before "make" first. That shouldn't fail. We want to test for that.

.circleci/config.yml
khaledhosny: Didn’t know that. Fixed.

.circleci/config.yml
khaledhosny: Plain `make` can fail, but there is no point in trying to show the test logs for such failure.

behdad:
Let's finish and merge this.

khaledhosny:
I think I’m done now.

behdad:
Thanks. Please merge after bots are happy.

PR #796, opened 2018-02-16
Add a continuous builder using fswatch

ebraminio:
Have a look, uses cmake/ninja/fswatch. I guess we can add more features or anything useful to it. Also feel free to suggest anything about the naming or so.

I tested it with: `src/dev-run.sh "/Library/Fonts/Tahoma.ttf" سََََََََََلام` (don't run it from src/ folder, run it from root of the project, that perhaps should checked by the script also)

Caveat: Doesn't support space on font name and has not a trivial fix apparently other than double quoting.

https://twitter.com/ebraminio/status/965333488969965568

b

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,6 +63,7 @@ if (HB_HAVE_INTROSPECTION)
   set (HB_HAVE_GLIB ON)
 endif ()
 
+option(HB_DISABLE_TEST_PROGS OFF "Do not build some of the test programs, useful for continuous builds")
 option(HB_CHECK OFF "Do a configuration suitable for testing (shared library and enable all options)")
 if (HB_CHECK)
   set (BUILD_SHARED_LIBS ON)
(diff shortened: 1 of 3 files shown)


ebraminio:
So fancy not worth considering or `fswatch` is not available on your distro package manager? Maybe we can use `entr` if that is available.

behdad:
Feel free to add. It probably is available on my distro, I haven't looked thoroughly.

ebraminio:
I like to see if it can be really useful to anyone first :)

behdad:
If it's easy to use, I might use it. Though, not just with hb-shape / hb-view.  My routine is a lot of "make -j5 -s lib && ./...." where the last part is either hb-shape, hb-view, or 'make check' in one of the test dirs.

ebraminio:
So some testing would be nice :)

1. `git fetch origin pull/796/head:dev-run`
2. src/dev-run.sh "/Library/Fonts/Tahoma.ttf" سََََََََََلام
3. change a file to see the result
4. Ctrl+C to exit and run the tests (but limiting to some specific test also can be considered)

Currently uses cmake as it is super fast even on my old system to use but can be replaced with autotools if it is really useful.

behdad:
Tried installing `fswatch`, but make failed. That's all my try for today.

ebraminio:
Ok, is `entr` available on your distro package manager? If so I will go for that.

PR #797, opened 2018-02-16
Remove checking macros __OPTIMIZE__ for MSVC

KindDragon:
Visual C++ compiler doesn't define macros __OPTIMIZE__  when optimizations enabled

### DIFF
--- a/src/hb-private.hh
+++ b/src/hb-private.hh
@@ -355,9 +355,9 @@ _hb_bit_storage (unsigned int number)
 {
 #if defined(__GNUC__) && (__GNUC__ >= 4) && defined(__OPTIMIZE__)
   return likely (number) ? (sizeof (unsigned int) * 8 - __builtin_clz (number)) : 0;
-#elif defined(_MSC_VER) && defined(__OPTIMIZE__)
+#elif defined(_MSC_VER)
   unsigned long where;
-  if (_BitScanReverse(&where, number)) return 1 + where;
+  if (_BitScanReverse (&where, number)) return 1 + where;
   return 0;
 #else
   unsigned int n_bits = 0;
@@ -375,11 +375,10 @@ _hb_ctz (unsigned int number)
 {
 #if defined(__GNUC__) && (__GNUC__ >= 4) && defined(__OPTIMIZE__)
   return likely (number) ? __builtin_ctz (number) : 0;
-#elif defined(_MSC_VER) && defined(__OPTIMIZE__)
+#elif defined(_MSC_VER)
   unsigned long where;
-  if (_BitScanForward(&where, n)) return where;
+  if (_BitScanForward (&where, number)) return where;
   return 0;
-}
 #else
   unsigned int n_bits = 0;
   if (unlikely (!number)) return 0;


behdad:
Thanks.

BTW, do you know of an intrinsic MSVC has that we can use in _hb_popcount?

src/hb-private.hh
```diff
@@ -355,7 +355,7 @@ _hb_bit_storage (unsigned int number)
 {
 #if defined(__GNUC__) && (__GNUC__ >= 4) && defined(__OPTIMIZE__)
   return likely (number) ? (sizeof (unsigned int) * 8 - __builtin_clz (number)) : 0;
-#elif defined(_MSC_VER) && defined(__OPTIMIZE__)
+#elif defined(_MSC_VER)
   unsigned long where;
   if (_BitScanReverse(&where, number)) return 1 + where;
```
ebraminio: From Behdad's comment on previous PR, please put a space after `_BitScanReverse`.

src/hb-private.hh
```diff
@@ -375,9 +375,9 @@ _hb_ctz (unsigned int number)
 {
 #if defined(__GNUC__) && (__GNUC__ >= 4) && defined(__OPTIMIZE__)
   return likely (number) ? __builtin_ctz (number) : 0;
-#elif defined(_MSC_VER) && defined(__OPTIMIZE__)
+#elif defined(_MSC_VER)
   unsigned long where;
-  if (_BitScanForward(&where, n)) return where;
+  if (_BitScanForward(&where, number)) return where;
```
ebraminio: Ditto

KindDragon:
> BTW, do you know of an intrinsic MSVC has that we can use in _hb_popcount?

They  support SSE4 instruction __popcnt, [`__popcnt64`](https://docs.microsoft.com/en-us/cpp/intrinsics/popcnt16-popcnt-popcnt64), but you need manually check `__cpuid` before

behdad:
> They support SSE4 instruction __popcnt, __popcnt64, but you need manually check __cpuid before

Right. We use popcount intrinsic from gcc. Will just leave the fallback impl as is then.

behdad:
AppVeyor is acting funny today. Can't tell if this failure is legit: https://ci.appveyor.com/project/harfbuzz/harfbuzz/build/1.0.595/job/68gr335o8w0737hs

Anyway; I merged this and am playing with it using mingw32. @rsheeter is buying us paid AppVeyor.

KindDragon:
> Right. We use popcount intrinsic from gcc. Will just leave the fallback impl as is then.

I mean MSVC intrinsic doesn't have fallback impl

ebraminio:
> @rsheeter is buying us paid AppVeyor.

Wow, cool. https://www.appveyor.com/docs/how-to/rdp-to-build-worker/

src/hb-private.hh
```diff
@@ -375,11 +375,10 @@ _hb_ctz (unsigned int number)
 {
 #if defined(__GNUC__) && (__GNUC__ >= 4) && defined(__OPTIMIZE__)
   return likely (number) ? __builtin_ctz (number) : 0;
-#elif defined(_MSC_VER) && defined(__OPTIMIZE__)
+#elif defined(_MSC_VER)
   unsigned long where;
-  if (_BitScanForward(&where, n)) return where;
+  if (_BitScanForward (&where, number)) return where;
   return 0;
-}
```
ebraminio: Hmm, this doesn't seem innocent, probably the build fail reason @behdad

PR #798, opened 2018-02-16
Do not mark the first glyph as unsafe to break

### LINKED ISSUE #791: HB_GLYPH_FLAG_UNSAFE_TO_BREAK is set on the first glyph
When not using the OT shaper, all the glyphs are marked as unsafe to break, including the first one; however, it is always safe to break before the first glyph.

dscorbett:
Fixes #791.

### DIFF
--- a/src/hb-buffer-private.hh
+++ b/src/hb-buffer-private.hh
@@ -344,8 +344,7 @@ struct hb_buffer_t {
   inline void
   unsafe_to_break_all (void)
   {
-    for (unsigned int i = 0; i < len; i++)
-      info[i].mask |= HB_GLYPH_FLAG_UNSAFE_TO_BREAK;
+    unsafe_to_break_impl (0, len);
   }
   inline void
   safe_to_break_all (void)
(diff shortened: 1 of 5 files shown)


src/hb-coretext.cc
```diff
@@ -1244,8 +1244,6 @@ _hb_coretext_shape (hb_shape_plan_t    *shape_plan,
 	pos->x_offset = info->var1.i32;
 	pos->y_offset = info->var2.i32;
 
-	info->mask = HB_GLYPH_FLAG_UNSAFE_TO_BREAK;
```
behdad: Hah! I didn't even remember that I have had added these lines to these shapers. Thanks for cleaning.

src/hb-buffer-private.hh
```diff
@@ -344,7 +344,7 @@ struct hb_buffer_t {
   inline void
   unsafe_to_break_all (void)
   {
-    for (unsigned int i = 0; i < len; i++)
```
behdad: This is not correct. It must scan over the buffer first and find minimum cluster number... scratch that. Remove the loop and just call `unsafe_to_break_impl (0, buffer->len)`.

dscorbett: I’ve made this change.

PR #800, opened 2018-02-17
Optimize set through optimized bit operations

behdad:


ebraminio:
Interesting fail just on MSVC x64. I was thinking about merging those two bots but now I see their usefulness. Do you need any help here btw?

behdad:
> Interesting fail just on MSVC x64. I was thinking about merging those two bots but now I see their usefulness. Is any help need here btw?

Yes. We have WIN64-specific code now, so it's good to have it.

behdad:
[I'm going to disable Appveyor for branches; PR's are enough.]

behdad:
> [I'm going to disable Appveyor for branches; PR's are enough.]

Weird. It was already set to only build master; but it clearly is building feature branches. Anyway; disabled build of feature branches with open PR existing.  Also enabled rolling builds, which automatically cancels queued builds when new commits are pushed to the same branch / PR.

PR #808, opened 2018-02-19
[aat] First dig on 'trak'

ebraminio:
Sorry that this is not that useful and obviously is erroneous, be careful when developing it. Based on Apple's [specification](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6trak.html) and [trak.js](https://github.com/devongovett/fontkit/blob/aat-kern/src/tables/trak.js)


ebraminio:
https://bug1398802.bmoattachments.org/attachment.cgi?id=8912161 ( https://github.com/qdot/gecko-hg/commit/2abdbb027f5a2f8a0c632544e148e763cbb5189f )

PR #810, opened 2018-02-20
Create .editorconfig

ebraminio:
I guess this is the convention we have on harfbuzz.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+root = true
+
+[*]
+charset = utf-8
+trim_trailing_whitespace = true
+end_of_line = lf
+insert_final_newline = true
+
+[*.{c,cc,h,hh}]
+indent_size = 2
+indent_style = space
+tab_width = 8
+
+[*.{py,sh}]
+indent_style = tab
+
+[{CMakeLists.txt,*.cmake}]
+indent_size = 2


khaledhosny:
How can one use this?

behdad:
I'm curious as well.  Does vim respect this? If yes, please let's add one to fonttools as well, so we stop fighting about tab vs spaces.

Also, if it makes code @rsheeter and @garretrieger write adhere to our conventions, then +10000000.

ebraminio:
Oh sorry, I forgot to put more information. It has a builtin support or using a plugin by these http://editorconfig.org/#download software.

ebraminio:
Vim has a plugin for it, https://github.com/editorconfig/editorconfig-vim that depends on what editor they use :)

ebraminio:
Its tab width setting however is not working. At least on VSCode and Atom it has this issue https://github.com/editorconfig/editorconfig-vscode/issues/190

jedmao:
Looks useful. It will probably be 100% useful for most if not all non-Atom editors. Even for Atom editors, at least it defines what the rules are for documentation purposes. The only reason it doesn't work on those editors is because "weird" indentation settings are enforced against and for good reason.

behdad:
I gave it a try (installed vim plugin, checked out this branch); didn't see it immediately change my workflow. LGTM.

PR #811, opened 2018-02-20
[util] Implement ppem/ptem on hb-{shape,view}

ebraminio:
I should have missed something...

    $ build/hb-shape --font-ptem 4 HelveticaNeueDeskInterface.ttc Hi
    [H=0+722|I=1+222]

    $ build/hb-shape --font-ptem 10 HelveticaNeueDeskInterface.ttc Hi
    [H=0+722|I=1+221]

So it has some effect, it seems.

### DIFF
--- a/util/options.cc
+++ b/util/options.cc
@@ -481,6 +481,25 @@ parse_font_size (const char *name G_GNUC_UNUSED,
       return false;
   }
 }
+
+static gboolean
+parse_font_ppem (const char *name G_GNUC_UNUSED,
+		 const char *arg,
+		 gpointer    data,
+		 GError    **error G_GNUC_UNUSED)
+{
+  font_options_t *font_opts = (font_options_t *) data;
+  switch (sscanf (arg, "%d%*[ ,]%d", &font_opts->x_ppem, &font_opts->y_ppem)) {
+    case 1: font_opts->y_ppem = font_opts->x_ppem;
+    case 2: return true;
+    default:
+      g_set_error (error, G_OPTION_ERROR, G_OPTION_ERROR_BAD_VALUE,
+		   "%s argument should be one or two space-separated numbers",
+		   name);
+      return false;
+  }
+}
+
 void
 font_options_t::add_options (option_parser_t *parser)
 {
@@ -513,12 +532,13 @@ font_options_t::add_options (option_parser_t *parser)
 
   GOptionEntry entries[] =
   {
-    {"font-file",	0, 0, G_OPTION_ARG_STRING,	&this->font_file,		"Set font file-name",			"filename"},
-    {"face-index",	0, 0, G_OPTION_ARG_INT,		&this->face_index,		"Set face index (default: 0)",		"index"},
+    {"font-file",	0, 0, G_OPTION_ARG_STRING,	&this->font_file,		"Set font file-name",				"filename"},
+    {"face-index",	0, 0, G_OPTION_ARG_INT,		&this->face_index,		"Set face index (default: 0)",			"index"},
     {"font-size",	0, default_font_size ? 0 : G_OPTION_FLAG_HIDDEN,
-			      G_OPTION_ARG_CALLBACK,	(gpointer) &parse_font_size,	font_size_text,				"1/2 numbers or 'upem'"},
-    /* TODO Add font-ppem / font-ptem. */
-    {"font-funcs",	0, 0, G_OPTION_ARG_STRING,	&this->font_funcs,		text,					"impl"},
+			      G_OPTION_ARG_CALLBACK,	(gpointer) &parse_font_size,	font_size_text,					"1/2 integers or 'upem'"},
+    {"font-ppem",	0, 0, G_OPTION_ARG_CALLBACK,	(gpointer) &parse_font_ppem,	"Set x,y pixels per EM (default: 0; disabled)",	"1/2 integers"},
+    {"font-ptem",	0, 0, G_OPTION_ARG_DOUBLE,	&this->ptem,			"Set font point-size (default: 0; disabled)",	"point-size"},
+    {"font-funcs",	0, 0, G_OPTION_ARG_STRING,	&this->font_funcs,		text,						"impl"},
     {nullptr}
   };
   parser->add_group (entries,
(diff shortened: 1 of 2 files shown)


util/options.cc
```diff
@@ -517,7 +517,9 @@ font_options_t::add_options (option_parser_t *parser)
     {"face-index",	0, 0, G_OPTION_ARG_INT,		&this->face_index,		"Set face index (default: 0)",		"index"},
     {"font-size",	0, default_font_size ? 0 : G_OPTION_FLAG_HIDDEN,
 			      G_OPTION_ARG_CALLBACK,	(gpointer) &parse_font_size,	font_size_text,				"1/2 numbers or 'upem'"},
-    /* TODO Add font-ppem / font-ptem. */
+    {"font-y-ppem",	0, 0, G_OPTION_ARG_INT,		&this->x_ppem,			"Set 'x pixels per EM'",		"96"},
```
behdad: Please do it the way font-size is done. Ie. no separate X / Y option.

Also, default is NOT 96. Defaults to zero, meaning not set / size-independent.

util/options.cc
```diff
@@ -517,7 +517,9 @@ font_options_t::add_options (option_parser_t *parser)
...
+    {"font-x-ppem",	0, 0, G_OPTION_ARG_INT,		&this->y_ppem,			"Set 'y pixels per EM'",		"96"},
+    {"font-ptem",	0, 0, G_OPTION_ARG_INT,		&this->ptem,			"Set font point-size",			"72"},
```
behdad: Same here. Default is 0 meaning not set / size-independent.

util/options.cc
```diff
@@ -691,6 +693,9 @@ font_options_t::get_font (void) const
   if (font_size_y == FONT_SIZE_UPEM)
     font_size_y = hb_face_get_upem (face);
 
+  if (x_ppem && y_ppem) hb_font_set_ppem (font, x_ppem, y_ppem);
```
behdad: Set if either is set. Or, always set.

util/options.cc
```diff
@@ -691,6 +693,9 @@ font_options_t::get_font (void) const
...
+  if (ptem) hb_font_set_ptem (font, ptem);
```
behdad: Same. Always set.

util/options.hh
```diff
@@ -478,6 +481,9 @@ struct font_options_t : option_group_t
   hb_variation_t *variations;
   unsigned int num_variations;
   int default_font_size;
+  int x_ppem;
+  int y_ppem;
+  int ptem;
```
behdad: The API takes ptem as float. Please make float and update argparsing to accept float.

behdad:
LGTM with comments. Thanks.

behdad:
Later, we might want to add hinting setting / FreeType load flags as well.
CC @lemzwerg

ebraminio:
@behdad, great, thanks for reviewing, can you have another look?

util/options.cc
```diff
@@ -691,6 +692,9 @@ font_options_t::get_font (void) const
   if (font_size_y == FONT_SIZE_UPEM)
     font_size_y = hb_face_get_upem (face);
 
+  hb_font_set_ppem (font, ppem, ppem);
```
behdad: I didn't mean not having separate xppem and yppem. Just that both will be parsed from `--font-ppem`. Again, see what `--font-size` is doing for parsing.

util/options.cc
```diff
@@ -517,7 +517,8 @@ font_options_t::add_options (option_parser_t *parser)
     {"face-index",	0, 0, G_OPTION_ARG_INT,		&this->face_index,		"Set face index (default: 0)",		"index"},
     {"font-size",	0, default_font_size ? 0 : G_OPTION_FLAG_HIDDEN,
 			      G_OPTION_ARG_CALLBACK,	(gpointer) &parse_font_size,	font_size_text,				"1/2 numbers or 'upem'"},
-    /* TODO Add font-ppem / font-ptem. */
+    {"font-ppem",	0, 0, G_OPTION_ARG_INT,		&this->ppem,			"Set pixels per EM",			"0"},
```
behdad: Looking at other options, looks like last argument is NOT default. It's the metavariable. Just see what the font-size and face-index have. Ie. the default goes into the argument help text. Eg. `"Set font point-size (default: 0; disabled)"`

ebraminio:
I guess I resolved the issues, what would be be instead "72" or "3.0" below? Just zero?

```
Font options:
  --font-file=filename                   Set font file-name
  --face-index=index                     Set face index (default: 0)
  --font-size=1/2 numbers or 'upem'      Font size (default: 256)
  --font-ppem=72 or 72,72                Set x,y pixels per EM (default: 0; disabled)
  --font-ptem=3.0                        Set font point-size (default: 0; disabled)
  --font-funcs=impl                      Set font functions implementation to use (default: ft)
```

util/options.cc
```diff
@@ -691,6 +711,9 @@ font_options_t::get_font (void) const
   if (font_size_y == FONT_SIZE_UPEM)
     font_size_y = hb_face_get_upem (face);
 
+  hb_font_set_ppem (font, x_ppem, y_ppem);
+  hb_font_set_ptem (font, ptem);
```
ebraminio: Should I guard this against default value? As comparing double to zero is not that nice, what would be the best way? As tests are not failing that perhaps not that needed?

behdad: The defaults match the font's defaults, so it's harmless to set. Also, no problem comparing double to zero when you specifically set it to zero yourself...

behdad:
> I guess I resolved the issues, what would be nice instead "72" or "3.0" below? Just zero?

Maybe "1/2 integers" and "point size"? Maybe change font-size from "1/2 numbers..." to "1/2 integers..." as well.

ebraminio:
Can you edit
```
Font options:
  --font-file=filename                   Set font file-name
  --face-index=index                     Set face index (default: 0)
  --font-size=1/2 numbers or 'upem'      Font size (default: 256)
  --font-ppem=72 or 72,72                Set x,y pixels per EM (default: 0; disabled)
  --font-ptem=3.0                        Set font point-size (default: 0; disabled)
  --font-funcs=impl                      Set font functions implementation to use (default: ft)
```
To what you see correct? That way I can just apply it :)

ebraminio:
@behdad ping

PR #812, opened 2018-02-20
Clean up of glyf subsetting.

garretrieger:
Move glyf subsetting code out of hb-subset.cc and into hb-ot-glyf-table.hh to be consistent with the other table subsetting code. Switch glyf to use hb_blob_copy_writable_or_fail instead of malloc.


behdad:
Please merge after bots are green.

PR #815, opened 2018-02-21
[subset] Add hb_map_t as a harfbuzz-like facade around std::unordered_map. Use it in subsetter.

rsheeter:
For subsetting only given the `std::unordered_map` lurking underneath.

Fails check-static-inits.sh; not entirely sure how to fix.


src/hb-subset-glyf.cc
```diff
@@ -153,9 +156,10 @@ _write_glyf_and_loca_prime (hb_subset_plan_t              *plan,
     _update_components (plan, glyf_prime_data_next, end_offset - start_offset);
 
     glyf_prime_data_next += length;
+    i++;
   }
 
-  success = success && _write_loca_entry (glyph_ids.len,
+  success = success && _write_loca_entry (hb_set_get_population (plan->gids_to_retain),
```
behdad: We should also cache set_population in plan as glyph_count. Can be done later.

src/hb-ot-cmap-table.hh
```diff
@@ -521,21 +521,22 @@ struct cmap
 			       hb_prealloced_array_t<CmapSubtableLongGroup> *groups) const
   {
     CmapSubtableLongGroup *group = nullptr;
-    for (unsigned int i = 0; i < plan->codepoints.len; i++) {
-
-      hb_codepoint_t cp = plan->codepoints[i];
-      if (!group || cp - 1 != group->endCharCode)
+    hb_codepoint_t cp = HB_SET_VALUE_INVALID;
+    while (hb_set_next (plan->codepoints, &cp))
```
behdad: I actually like the for() loop of this instead of while. But feel free to leave as is.

src/hb-subset-map.cc
```diff
@@ -0,0 +1,54 @@
...
+
+hb_map_t *
+hb_map_create_or_fail ()
+{
+  hb_map_t *map = hb_object_create<hb_map_t> ();
+  if (map)
+  {
+    map->map = new (std::nothrow) std::unordered_map<hb_codepoint_t, hb_codepoint_t>();
+  }
+  if (!map->map)
```
behdad: This is invalid access if first map allocation failed. Instead, just return fast if allocation failed.

src/hb-subset-map.cc
```diff
@@ -0,0 +1,54 @@
...
+  {
+    hb_object_destroy (map);
+    map = nullptr;
```
behdad: Same here. Just return.

src/hb-subset-map.cc
```diff
@@ -0,0 +1,54 @@
...
+HB_INTERNAL void
+hb_map_put (hb_map_t *map, hb_codepoint_t key, hb_codepoint_t value)
+{
+  (*map->map)[key] = value;
+}
+
+HB_INTERNAL hb_bool_t
+hb_map_get (hb_map_t *map, hb_codepoint_t key, hb_codepoint_t *value /* OUT */)
+{
+  const auto it = map->map->find(key);
```
behdad: We'll see how our bots like auto...

src/hb-subset-map.cc
rsheeter: Fixed.

src/hb-subset-map.cc
rsheeter: Fixed.

behdad:
I'm happy. Bots not quite.

behdad:
We have `hb_map_t` implemented natively now.

PR #816, opened 2018-02-21
[aat] Implement trak logic

ebraminio:
Not to forget (may or may not on this issue):
* SANITZE, SANITZIE, SANTIZE
* > Assign the UnsizedArrays to interim variables before indexing them
* > bound-check
* > Also, you need a track number, which defaults to 0. We should add API on the buffer or shape later to set traking / letter-spacing.
* > We should possibly auto-disable ligatures if traking is set.

  (`ff` on Helvetica Neue)
* #808

### DIFF
--- a/src/hb-aat-layout-trak-table.hh
+++ b/src/hb-aat-layout-trak-table.hh
@@ -1,6 +1,6 @@
 /*
- * Copyright © 2018  Google, Inc.
  * Copyright © 2018  Ebrahim Byagowi
+ * Copyright © 2018  Google, Inc.
  *
  *  This is part of HarfBuzz, a text shaping library.
  *
@@ -45,10 +45,15 @@ struct TrackTableEntry
     return_trace (c->check_struct (this));
   }
 
+  inline float get_value (const void *base, unsigned int index) const
+  {
+    return (base+values)[index];
+  }
+
   protected:
   Fixed			track;		/* Track value for this record. */
   HBUINT16		trackNameID;	/* The 'name' table index for this track */
-  OffsetTo<UnsizedArrayOf<Fixed> >
+  OffsetTo<UnsizedArrayOf<HBINT16> >
 			values;		/* Offset from start of tracking table to
 					 * per-size tracking values for this track. */
 
@@ -61,7 +66,40 @@ struct TrackData
   inline bool sanitize (hb_sanitize_context_t *c) const
   {
     TRACE_SANITIZE (this);
+    /* TODO */
     return_trace (c->check_struct (this));
+  }
+
+  inline float get_tracking (const void *base, float ptem) const
+  {
+    /* CoreText points are CSS pixels (96 per inch),
+     * NOT typographic points (72 per inch).
+     *
+     * https://developer.apple.com/library/content/documentation/GraphicsAnimation/Conceptual/HighResolutionOSX/Explained/Explained.html
+     */
+    float csspx = ptem * 96.f / 72.f;
+    Fixed fixed_size;
+    fixed_size.set_float (csspx);
+
+    // TODO: Make indexing work and use only an entry with zero track
+    const TrackTableEntry trackTableEntry = trackTable[0];
+
+    unsigned int size_index;
+    for (size_index = 0; size_index < nSizes; ++size_index)
+      if ((base+sizeTable)[size_index] >= fixed_size)
+        break;
+
+    // We don't attempt to extrapolate to larger or smaller values
+    if (size_index == nSizes)
+      return trackTableEntry.get_value (base, nSizes - 1);
+    if (size_index == 0 || (base+sizeTable)[size_index] == fixed_size)
+      return trackTableEntry.get_value (base, size_index);
+
+    float s0 = (base+sizeTable)[size_index - 1].to_float ();
+    float s1 = (base+sizeTable)[size_index].to_float ();
+    float t = (csspx - s0) / (s1 - s0);
+    return t * trackTableEntry.get_value (base, size_index) +
+      (1.0 - t) * trackTableEntry.get_value (base, size_index - 1);
   }
 
   protected:
@@ -69,7 +107,7 @@ struct TrackData
   HBUINT16		nSizes;		/* Number of point sizes included in this table. */
   LOffsetTo<UnsizedArrayOf<Fixed> >
 			sizeTable;
-  TrackTableEntry	trackTable[VAR];/* Array[nSizes] of size values. */
+  UnsizedArrayOf<TrackTableEntry>	trackTable;/* Array[nSizes] of size values. */
 
   public:
   DEFINE_SIZE_ARRAY (8, trackTable);
@@ -82,7 +120,41 @@ struct trak
   inline bool sanitize (hb_sanitize_context_t *c) const
   {
     TRACE_SANITIZE (this);
+    /* TODO */
     return_trace (c->check_struct (this));
+  }
+
+  inline bool apply (hb_aat_apply_context_t *c) const
+  {
+    TRACE_APPLY (this);
+    const float ptem = c->font->ptem;
+    if (ptem > 0.f)
+    {
+      hb_buffer_t *buffer = c->buffer;
+      if (HB_DIRECTION_IS_HORIZONTAL (c->buffer->props.direction))
+      {
+        const TrackData trackData = this+horizOffset;
+        float tracking = trackData.get_tracking (this, ptem);
+        hb_position_t advance_to_add = c->font->em_scalef_x (tracking / 2);
+        foreach_grapheme (buffer, start, end)
+        {
+          buffer->pos[start].x_advance += advance_to_add;
+          buffer->pos[end].x_advance += advance_to_add;
+        }
+      }
+      else
+      {
+        const TrackData trackData = this+vertOffset;
+        float tracking = trackData.get_tracking (this, ptem);
+        hb_position_t advance_to_add = c->font->em_scalef_y (tracking / 2);
+        foreach_grapheme (buffer, start, end)
+        {
+          buffer->pos[start].y_advance += advance_to_add;
+          buffer->pos[end].y_advance += advance_to_add;
+        }
+      }
+    }
+    return_trace (false);
   }
 
   protected:
--- a/src/hb-aat-layout.cc
+++ b/src/hb-aat-layout.cc
@@ -55,6 +55,40 @@ _get_morx (hb_face_t *face, hb_blob_t **blob = nullptr)
   return morx;
 }
 
+static inline const AAT::kerx&
+_get_kerx (hb_face_t *face, hb_blob_t **blob = nullptr)
+{
+  if (unlikely (!hb_ot_shaper_face_data_ensure (face)))
+  {
+    if (blob)
+      *blob = hb_blob_get_empty ();
+    return OT::Null(AAT::kerx);
+  }
+  hb_ot_layout_t * layout = hb_ot_layout_from_face (face);
+  /* XXX this doesn't call set_num_glyphs on sanitizer. */
+  const AAT::kerx& kerx = *(layout->kerx.get ());
+  if (blob)
+    *blob = layout->kerx.blob;
+  return kerx;
+}
+
+static inline const AAT::trak&
+_get_trak (hb_face_t *face, hb_blob_t **blob = nullptr)
+{
+  if (unlikely (!hb_ot_shaper_face_data_ensure (face)))
+  {
+    if (blob)
+      *blob = hb_blob_get_empty ();
+    return OT::Null(AAT::trak);
+  }
+  hb_ot_layout_t * layout = hb_ot_layout_from_face (face);
+  /* XXX this doesn't call set_num_glyphs on sanitizer. */
+  const AAT::trak& trak = *(layout->trak.get ());
+  if (blob)
+    *blob = layout->trak.blob;
+  return trak;
+}
+
 static inline void
 _hb_aat_layout_create (hb_face_t *face)
 {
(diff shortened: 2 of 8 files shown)


src/hb-aat-layout.cc
```diff
@@ -78,3 +112,15 @@ hb_aat_layout_substitute (hb_font_t *font, hb_buffer_t *buffer)
...
+
+void
+hb_aat_layout_position (hb_font_t *font, hb_buffer_t *buffer)
+{
+  hb_blob_t *blob;
+  const AAT::trak& trak = _get_trak (font->face, &blob);
+  const AAT::kerx& kerx = _get_kerx (font->face, &blob);
+
+  AAT::hb_aat_apply_context_t c (font, buffer, blob);
+  trak.apply (&c);
```
behdad: I think trak should go after kerx.

src/hb-aat-layout-trak-table.hh
```diff
@@ -58,6 +65,13 @@ struct TrackTableEntry
 
 struct TrackData
 {
+  inline bool apply (hb_aat_apply_context_t *c) const
```
behdad: We probably don't need apply here or in TrackTableEntry. Basically just add them when you hook them up.

src/hb-aat-layout.cc
ebraminio: Obviously, done.

src/hb-aat-layout-trak-table.hh
ebraminio: Makes sense. Thanks

src/hb-aat-layout-trak-table.hh
```diff
@@ -45,10 +45,10 @@ struct TrackTableEntry
     return_trace (c->check_struct (this));
   }
 
-  protected:
+  public:
   Fixed			track;		/* Track value for this record. */
   HBUINT16		trackNameID;	/* The 'name' table index for this track */
-  OffsetTo<UnsizedArrayOf<Fixed> >
+  OffsetTo<UnsizedArrayOf<HBINT16> >
```
behdad: Why?

src/hb-aat-layout-trak-table.hh
```diff
@@ -64,21 +64,71 @@ struct TrackData
...
   }
 
-  protected:
+  public:
   HBUINT16		nTracks;	/* Number of separate tracks included in this table. */
   HBUINT16		nSizes;		/* Number of point sizes included in this table. */
   LOffsetTo<UnsizedArrayOf<Fixed> >
 			sizeTable;
-  TrackTableEntry	trackTable[VAR];/* Array[nSizes] of size values. */
+  UnsizedArrayOf<TrackTableEntry>	trackTable;/* Array[nSizes] of size values. */
```
behdad: Humm, yeah, this UnsizedArrayOf<> thing is new. Guess there's a bunch of places we can switch to it.

src/hb-aat-layout-trak-table.hh
```diff
@@ -64,21 +64,71 @@ struct TrackData
     return_trace (c->check_struct (this));
   }
 
-  protected:
+  public:
```
behdad: It's preferred that you add useful methods and keep things protected, than public.

src/hb-aat-layout-trak-table.hh
```diff
@@ -64,21 +64,71 @@ struct TrackData
...
 {
   static const hb_tag_t tableTag = HB_AAT_TAG_trak;
 
+  inline bool apply (hb_aat_apply_context_t *c) const
+  {
+    TRACE_APPLY (this);
+
+    hb_unicode_funcs_t *unicode = c->buffer->unicode;
+    unsigned int u = c->buffer->info->codepoint;
+    unsigned int gen_cat = (unsigned int) unicode->general_category (u);
```
behdad: We have this cached on the buffer already. You don't need to get it again. Check hb-ot-layout-private.hh.

src/hb-aat-layout-trak-table.hh
```diff
@@ -64,21 +64,71 @@ struct TrackData
...
 struct trak
 {
   static const hb_tag_t tableTag = HB_AAT_TAG_trak;
 
+  inline bool apply (hb_aat_apply_context_t *c) const
+  {
+    TRACE_APPLY (this);
+
+    hb_unicode_funcs_t *unicode = c->buffer->unicode;
+    unsigned int u = c->buffer->info->codepoint;
```
behdad: At this point, codepoint is a glyph, not Unicode character, so you cannot do this.

src/hb-aat-layout-trak-table.hh
```diff
@@ -64,21 +64,71 @@ struct TrackData
...
+    unsigned int gen_cat = (unsigned int) unicode->general_category (u);
+    printf(">>> %d %d <<<", u, gen_cat);
+
+    const TrackData trackData = this+horizOffset;
+    const TrackTableEntry trackTableEntry = trackData.trackTable[0];
+
+    for (unsigned int i = 0; i < trackData.nSizes; ++i)
```
behdad: This fetching of trak should be methods on the respective structs.

Also, you need a track number, which defaults to 0. We should add API on the buffer or shape later to set traking / letter-spacing.

src/hb-aat-layout-trak-table.hh
```diff
@@ -64,21 +64,71 @@ struct TrackData
...
   LOffsetTo<UnsizedArrayOf<Fixed> >
 			sizeTable;
-  TrackTableEntry	trackTable[VAR];/* Array[nSizes] of size values. */
+  UnsizedArrayOf<TrackTableEntry>	trackTable;/* Array[nSizes] of size values. */
 
   public:
   DEFINE_SIZE_ARRAY (8, trackTable);
 };
 
+#define HB_CORETEXT_DEFAULT_FONT_SIZE 12.f
```
behdad: If ptem is <= 0 / not set, I *think* we should simply not apply trak.

src/hb-aat-layout-trak-table.hh
ebraminio: Apparently I was wrong

src/hb-aat-layout-trak-table.hh
ebraminio: Sure

src/hb-aat-layout-trak-table.hh
ebraminio: Definitely, make sense, its logic is copied from hb-coretext, perhaps we can change that also to share the logic sometime.

src/hb-aat-layout-trak-table.hh
ebraminio: Oh these are just my playing but thanks.

src/hb-aat-layout-trak-table.hh
ebraminio: Ops, shouldn't upload this either

src/hb-aat-layout-trak-table.hh
ebraminio: Oh, sure

src/hb-aat-layout-trak-table.hh
ebraminio: Done

src/hb-aat-layout-trak-table.hh
ebraminio: Done

src/hb-aat-layout-trak-table.hh
ebraminio: > Also, you need a track number, which defaults to 0. We should add API on the buffer or shape later to set traking / letter-spacing.

That would be nice also.

ebraminio:
Finally this has become something I can be proud of :)

src/hb-aat-layout-trak-table.hh
```diff
@@ -61,15 +66,48 @@ struct TrackData
...
+    float csspx = ptem * 96.f / 72.f;
+    Fixed fixed_size;
+    fixed_size.set_float (csspx);
+
+    // TODO: Figure out when to use others
+    const TrackTableEntry trackTableEntry = trackTable[0];
+
+    unsigned int size_index;
+    for (size_index = 0; size_index < nSizes; ++size_index)
+      if ((base+sizeTable)[size_index] >= fixed_size)
```
ebraminio: Is comparing "Fixed" supposed to be OK? It seems is working however I am worried about it.

src/hb-aat-layout-trak-table.hh
```diff
@@ -82,9 +120,38 @@ struct trak
...
+    const float ptem = c->font->ptem;
+    if (ptem > 0.f)
+    {
+      hb_buffer_t *buffer = c->buffer;
+      if (HB_DIRECTION_IS_HORIZONTAL (c->buffer->props.direction))
+      {
+        const TrackData trackData = this+horizOffset;
+        float tracking = trackData.get_tracking (this, ptem);
+        int x_scale = c->font->x_scale / c->face->upem;
+        // TODO: Do we anything else for RTL?
```
ebraminio: https://github.com/qdot/gecko-hg/blob/ab17432d9bede83cf25be231f4bfba44478561fb/gfx/thebes/gfxFont.cpp#L743 for example

src/hb-aat-layout-trak-table.hh
```diff
@@ -82,9 +120,38 @@ struct trak
...
+        foreach_grapheme (buffer, start, end)
+          buffer->pos[end].x_advance += tracking * x_scale;
+      }
+      else
+      {
+        const TrackData trackData = this+vertOffset;
+        float tracking = trackData.get_tracking (this, ptem);
+        int y_scale = c->font->y_scale / c->face->upem;
+        foreach_grapheme (buffer, start, end)
+          buffer->pos[end].y_advance += tracking * y_scale;
```
ebraminio: Anything to be worried about TTB/BTT distinction?

src/hb-aat-layout-trak-table.hh
```diff
@@ -61,15 +66,48 @@ struct TrackData
...
+     * NOT typographic points (72 per inch).
+     *
+     * https://developer.apple.com/library/content/documentation/GraphicsAnimation/Conceptual/HighResolutionOSX/Explained/Explained.html
+     */
+    float csspx = ptem * 96.f / 72.f;
+    Fixed fixed_size;
+    fixed_size.set_float (csspx);
+
+    // TODO: Figure out when to use others
+    const TrackTableEntry trackTableEntry = trackTable[0];
```
ebraminio: https://github.com/qdot/gecko-hg/commit/2abdbb027f5a2f8a0c632544e148e763cbb5189f#diff-2ddaa57c670bc5ab02a79e074435d908R552

src/hb-aat-layout-trak-table.hh
```diff
@@ -61,15 +66,48 @@ struct TrackData
...
+
+    unsigned int size_index;
+    for (size_index = 0; size_index < nSizes; ++size_index)
+      if ((base+sizeTable)[size_index] >= fixed_size)
+        break;
+
+    // We don't attempt to extrapolate to larger or smaller values
```
ebraminio: This is what Mozilla [does](https://github.com/qdot/gecko-hg/blob/2abdbb027f5a2f8a0c632544e148e763cbb5189f/gfx/thebes/gfxMacPlatformFontList.mm#L595) but not per spec, `If the point size and track number required are not found in the tracking table, interpolation or extrapolation is performed to determine the adjustment to the advance width that is required for the font size and tracking number specified.`

src/hb-aat-layout-trak-table.hh
ebraminio: `The tracking table you create should have per-size tracking values for track numbers -1, 0, and +1 at a minimum of two font sizes. Additional font sizes can also be included in the table. Tracking numbers are arbitrary, but the values -1, 0, and +1 should be used to indicate loose, normal, and tight tracking, respectively.`

src/hb-aat-layout-trak-table.hh
behdad: Yes, they are integers.

src/hb-aat-layout-trak-table.hh
```diff
@@ -82,9 +120,38 @@ struct trak
...
+    TRACE_APPLY (this);
+    const float ptem = c->font->ptem;
+    if (ptem > 0.f)
+    {
+      hb_buffer_t *buffer = c->buffer;
+      if (HB_DIRECTION_IS_HORIZONTAL (c->buffer->props.direction))
+      {
+        const TrackData trackData = this+horizOffset;
+        float tracking = trackData.get_tracking (this, ptem);
+        int x_scale = c->font->x_scale / c->face->upem;
```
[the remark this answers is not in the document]
behdad: Theres font->em_scale_x or something like that.

src/hb-open-type-private.hh
```diff
@@ -692,8 +692,8 @@ struct F2DOT14 : HBINT16
 /* 32-bit signed fixed-point number (16.16). */
 struct Fixed: HBINT32
 {
-  //inline float to_float (void) const { return ???; }
-  //inline void set_float (float f) { v.set (f * ???); }
+  inline float to_float (void) const { return ((int32_t) v) / 65536.0; }
+  inline void set_float (float f) { v.set ((int) (f * 65536.0 + 0.5)); }
```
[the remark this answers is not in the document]
behdad: We started using round() I think, so you can use that.

src/hb-aat-layout-trak-table.hh
ebraminio: Do we need 2d track/font size interpolation eventually?

behdad: I think we have a bfind() that works for this.

behdad: > Do we need 2d track/font size interpolation eventually?

Yes. Bilinear interpolation.

src/hb-aat-layout-trak-table.hh
behdad: In general I think we want to add half of the tracking to each edge of the glyph...

src/hb-aat-layout-trak-table.hh
ebraminio: Thank you very much :)

src/hb-aat-layout-trak-table.hh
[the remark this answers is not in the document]
ebraminio: Done

src/hb-open-type-private.hh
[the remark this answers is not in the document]
ebraminio: Done

src/hb-aat-layout-trak-table.hh
ebraminio: Done

src/hb-aat-layout-trak-table.hh
ebraminio: Ditto below

PR #821, opened 2018-02-21
Clean up check-defs and check-symbols

behdad:
I probably broken Mac test. I'm going to merge the two tests and bot-happy
it after.

Also, call c++filt on their results.


behdad:
@ebraminio Can I get your help updating cmake for this please? check-defs.sh is gone, and harfbuzz-gobject is also tested by check-symbols.sh.

ebraminio:
Done

behdad:
Thanks Ebrahim. I made the distcheck bot happy, but please check my last commit here, see if Cmake needs any adjusting.  I like to merge this soon. Thanks!

ebraminio:
Still failing bots :)

behdad:
@khaledhosny looks like fail.sh is not working? 
https://travis-ci.org/harfbuzz/harfbuzz/jobs/345373422

behdad:
I have set " Do not build feature branches with open Pull Requests" in AppVeyor. No idea why it still builds /pr and /branch.

ebraminio:
Merged this as Appveyor had a long queue not worth to wait.

> Maybe I should keep hb-gobject-enum.* in git as well and we'd have a full source distribution in tree...

Good idea as they are not touched that much and that specially some py2/py3 issue on ArchLinux IIRC but somehow we should test them that they are not outdated...

behdad:
Ok, let's go all the way then!  I'll merge the hb-version branch and then do the enums. Makes a lot of build stuff simpler.

behdad:
The rebuild rules in autotools will still be there, so we'll notice if they change.  The only risk is glib changes the generated code it can cause noise when different developers with different glib versions touch code.

PR #823, opened 2018-02-22
[subset] Add hdmx subsetting.

garretrieger:


behdad:
Curious. Why hdmx? No one uses that...

src/hb-ot-hdmx-table.hh
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  HBUINT16 version;
+  HBINT16  num_records;
+  HBINT32  size_device_record;
+
+  DeviceRecord records[VAR];
+
+  DEFINE_SIZE_MIN (8);
+
```
behdad: Remove extra lines. End of the struct should look like others, with public: and protected: and comments and ...

src/hb-ot-hdmx-table.hh
```diff
@@ -0,0 +1,77 @@
...
+#define HB_OT_TAG_hdmx HB_TAG('h','d','m','x')
+
+
+struct DeviceRecord
+{
+  HBUINT8 pixel_size;
+  HBUINT8 max_width;
+  HBUINT8 widths[VAR];
+};
```
behdad: This needs DEFINE_SIZE_STATIC

behdad: and sanitize.

src/hb-ot-hdmx-table.hh
```diff
@@ -0,0 +1,77 @@
...
+  {
+    if (unlikely (i >= num_records)) return Null(DeviceRecord);
+    return StructAtOffset<DeviceRecord> (this, min_size + i * size_device_record);
+  }
+
+  inline bool sanitize (hb_sanitize_context_t *c) const
+  {
+    TRACE_SANITIZE (this);
+    return_trace (likely (c->check_struct (this)
+                          && version == 0));
```
behdad: This is wrong. Should check the array.

garretrieger:
Fonts subsetted via harfbuzz were failing to pass ots due to hdmx being invalid.

src/hb-ot-hdmx-table.hh
garretrieger: Done.

src/hb-ot-hdmx-table.hh
garretrieger: Done.

src/hb-ot-hdmx-table.hh
garretrieger: Done.

Issue #817, opened 2018-02-21
Chrome DEPS based build fails with missing files


drott:
For a build without Ragel and running `autogen.h` we still need the following three files in the repo:
`hb-buffer-deserialize-text.hh`
`hb-buffer-deserialize-json.hh`
and 
`hb-version.h`

drott:
Thanks, we still need a parsed version of `hb-version.h`, filled in by autotools, or some alternative to that.

PR #825, opened 2018-02-23
Include hb-version.h in git

behdad:
Probably needs Cmake updates.

https://github.com/harfbuzz/harfbuzz/issues/817


behdad:
@ebraminio Can I get some help from you here as well? I'm moving to keeping hb-version.h in git as well. Don't know what's best to do in Cmake. You can remove regenerating it maybe. I don't understand the bot failures enough to fix. Thanks.

ebraminio:
This seems to be OK as well and needs #821. Of course we better to remove hb-version/ragel generation from CMake sometime but for now failings are not related to that.

behdad:
Ah ok. I'm hopeful that one will go through this time.

behdad:
@ebraminio HELP :D

PR #827, opened 2018-02-23
Add BASE table

behdad:


behdad:
Humm. Did you forget to add the main file in the end? This looks, not right.

ebraminio:
Behdad, what is missing here? I could run https://github.com/eroux/harfbuzz-base-demo after some minor struggles.

behdad:
Don't know. The diff doesn't show the main piece of code to me (hb-ot-layout-base-table.hh); and then we need an API.

ebraminio:
Just click on "Load diff" :P

ebraminio:
This look to have a good quality also, how about merging it but marking it somehow for further security review or something like?

behdad:
> Just click on "Load diff" :P

Hah! UI fail...



> This look to have a good quality also, how about merging it but marking it somehow for further security review or something like?

I'll merge and review now.

behdad:
Following up in https://github.com/harfbuzz/harfbuzz/pull/838

PR #828, opened 2018-02-23
[subset] Fix several memory leaks in subsetting.

garretrieger:


behdad:
lgtm. Please merge when bots agree.

PR #831, opened 2018-02-24
Remove ragel use on CI

ebraminio:
The main goal is to remove ragel installation on most of the bots (keep them on Travis), that is somehow beneficial as reduces bots build time and configs complexity. What should be used to make `AM_MISSING_PROG([RAGEL], [ragel])` to not fail the builds when it couldn't find it? Or maybe this is totally a wrong approach? It would be nice if you could give me a hand @behdad.


behdad:
AM_MISSING_PROG doesn't fail.

behdad:
I'm not sure why the bot is trying to rebuild the ragel-generated file. Timestamps should have taken care of this.

It's weird that $(RAGEL) is expanding to nothing. Let me remember how this is supposed to work.

behdad:
This is what AM_MISSING_PROG is supposed to do:
```
AM_MISSING_PROG(name, program)
Find a maintainer tool program and define the name environment variable with its location. If program is not detected, then name will instead invoke the missing script, in order to give useful advice to the user about the missing maintainer tool. See maintainer-mode, for more information on when the missing script is appropriate.
```

ebraminio:
nvm

PR #836, opened 2018-02-25
[aat] trak sanitization

ebraminio:
harfbuzz really works beautifully :)

See also #816, #808

### DIFF
--- a/src/hb-aat-layout-trak-table.hh
+++ b/src/hb-aat-layout-trak-table.hh
@@ -39,11 +39,15 @@ namespace AAT {
 
 struct TrackTableEntry
 {
-  inline bool sanitize (hb_sanitize_context_t *c) const
+  inline bool sanitize (hb_sanitize_context_t *c, const void *base, uint16_t size) const
   {
     TRACE_SANITIZE (this);
-    /* XXX Sanitize values */
-    return_trace (c->check_struct (this));
+    return_trace (c->check_struct (this) && ((base+values).sanitize (c, size)));
+  }
+
+  inline Fixed get_track_value () const
+  {
+    return track;
   }
 
   inline float get_value (const void *base, unsigned int index) const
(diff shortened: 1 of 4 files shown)


src/hb-aat-layout-trak-table.hh
```diff
@@ -39,11 +39,15 @@ namespace AAT {
 
 struct TrackTableEntry
 {
-  inline bool sanitize (hb_sanitize_context_t *c) const
+  inline bool sanitize (hb_sanitize_context_t *c, const void *base, uint16_t size) const
   {
     TRACE_SANITIZE (this);
-    /* XXX Sanitize values */
```
ebraminio: Any special thing here?

behdad: The value array is not sanitized...

I'll try to clean up sanitize in trak soon.

In the future, it would be much better to first write the table structs and finish sanitize, before moving on to actually using the struct. The time the sanitize is written is the best time to make sure all memory access is accounted for correctly. Failing that, an XXX comment is to be added, to make sure we fix things before calling into that code. It's much harder to audit code after the sanitize is written and checked in without an XXX.

src/hb-aat-layout-trak-table.hh
```diff
@@ -124,8 +157,28 @@ struct trak
...
+
+    if (!(c->check_struct (this)))
+      return_trace (false);
+
+    if ((format != 0) || (reserved != 0))
+      return_trace (false);
+
+    if (horizData)
+    {
+      const TrackData &trackData = this+horizData;
```
behdad: The `&` here is important. Failing to put it there generates memory-wrong code when the structures reference other structs. I tried to make it illegal to assign these structs, but couldn't get it working.

src/hb-aat-layout-trak-table.hh
ebraminio: That makes sense. For here, isn't `((base+values).sanitize (c, size))` added below enough?

behdad: Humm. The code in front of me is this:
```
struct TrackTableEntry 
{ 
  { 
    TRACE_SANITIZE (this); 
    /* XXX Sanitize values */ 
  } 
 
  inline float get_value (const void *base, unsigned int index) const 
  { 
    return (base+values)[index]; 
  } 
 
```

ebraminio: !!! This is before the change, it should followed by `get_track_value` also.

behdad: Sorry I thought you merged already.

src/hb-aat-layout-trak-table.hh
behdad: Public link about design of our structs? Not yet. I'm slowly trying to work on one.

ebraminio: I guess I no longer need that documentation, its too late for me :)

About what was that attempt on making struct uncopyable, I know it starts with making copy method private but I like to know what stopped the move and learn from that.

behdad: I can make it by creating a new empty type to be derived from by all font structs. Eg.:
```
struct FontStruct 
{ 
  private: 
  FontStruct (const FontStruct &); 
  FontStruct & operator = (const FontStruct &); 
}; 
```
I don't know What else we can do with that. Also not sure if the DEFINE_* should check that we inherit from this. Note that for basic types (HBINT*, Tag, GlyphID, ...) we don't want to enforce this.  So maybe we can have two different types, one for ValueTypes and one for others.  I like the idea. Can you open an issue for it?

If we make the base type a template that takes the inherited type (https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern) we might be able to do more useful things as well.

ebraminio: #843

PR #841, opened 2018-02-27
Support CPAL table

ebraminio:
In continuation to and closes #244, addressed some of the review there and (unsure about if `hb_ot_color_get_palette_colors` is best API we can provide?) however the related test fails. I want to see if my changes are in right direction? (making fields private, doing checks at the related places, ...)


src/hb-ot-cpal-table.hh
```diff
@@ -0,0 +1,174 @@
...
+    		  paletteEntryLabel.sanitize (c, this));
+  }
+
+  private:
+  inline hb_ot_color_palette_flags_t get_palette_flags (const void *base, unsigned int palette) const
+  {
+    if (paletteFlags.is_null ())
+      return HB_OT_COLOR_PALETTE_FLAG_DEFAULT;
+
+    const HBUINT32* flags = (const HBUINT32 *) (const void*) &paletteFlags (base);
```
ebraminio: I wish I could use UnsizedArrayOf...

behdad: Yeah, will move UnsizedArrayOf to hb-open-type.hh some time.

behdad:
Ok let me finish the other stuff. If this is clean and adds no API for now, feel free to merge.

ebraminio:
I improved it as far as I could from things learned from the previous experiences, like sanitizing arrays, bringing the logic on the right place and making data private and in the process I was careful about not breaking the tests so hopefully it won't give you headache when you want use the API or not before disabling the tests altogether. However I didn't refactor `hb_ot_color_get_palette_colors` as I wasn't sure if that is right abstraction to have, but I can understand its benefits of course.

I think Wine implementation is a good place to be inspired about what would be needed for a real implementation to be exposed, https://github.com/wine-mirror/wine/blob/5ec6b8f807f61ee77b9a96d94798c8e3f3db7af4/dlls/dwrite/dwrite_private.h#L220-L238

behdad:
Thanks. Let me finish BASE (while also keeping eye on subsetter, and redoing fonttools table packing) and aat, then review this.

ebraminio:
Behdad, if you got sometime, just review [hb-ot-cpal-table.hh](https://github.com/harfbuzz/harfbuzz/pull/841/commits/41aa43a7c80cd3fe02e22a188a21a5eb79357402#diff-93799e2d0b4092493e62f1e8d54a4457) and let me know things I missed as I've commented public API of it and for now the goal is to close a stale PR.

ebraminio:
Ok, this adds no API now (except the some enums which I guess are not issue) and is clean, so lets have it as your previous suggestion.

PR #845, opened 2018-02-28
[subset] drop hints from composites

rsheeter:
Add a unit test, fix todo's and bugs in drop hints for composites.

We inspect flags repeatedly. I am disinclined to optimize this until it pops out on a perf profile.


src/hb-subset-glyf.cc
```diff
@@ -180,7 +180,17 @@ _write_glyf_and_loca_prime (hb_subset_plan_t              *plan,
...
+      {
+        /* remove WE_HAVE_INSTRUCTIONS from flags in dest */
+        OT::glyf::CompositeGlyphHeader::Iterator composite_it;
+        if (unlikely (!OT::glyf::CompositeGlyphHeader::get_iterator (glyf_prime_data_next, length, &composite_it))) return false;
+        const OT::glyf::CompositeGlyphHeader *glyph;
+        do {
+          glyph = composite_it.current;
+          OT::HBUINT16 *flags = const_cast<OT::HBUINT16 *> (&glyph->flags);
+          flags->set ( (uint16_t) *flags & ~OT::glyf::CompositeGlyphHeader::WE_HAVE_INSTRUCTIONS);
+        } while (composite_it.move_to_next());
```
garretrieger: Consider moving this block into a helper method.

rsheeter: Done

PR #847, opened 2018-02-28
[colr] Implement

ebraminio:
Now that we have CPAL table, lets have COLR also, hopefully we can reach to both and providing things tests, public API and such on the next release. I like to port my color emoji code to it https://github.com/ebraminio/glcourse/blob/master/emojidump.cc someday on that process hopefully I can test and make sure everything works as intended also.

### DIFF
--- a/src/Makefile.sources
+++ b/src/Makefile.sources
@@ -97,6 +97,7 @@ HB_OT_sources = \
 	hb-ot-layout-jstf-table.hh \
 	hb-ot-layout-private.hh \
 	hb-ot-color.cc \
+	hb-ot-colr-table.hh \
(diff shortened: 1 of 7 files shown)


behdad:
The API needs to be redone. That adds more work before I can get next release out...  Typically we haven't done this. But I'm fine as a one-time thing.

ebraminio:
Opened #849 for discussing that, no problem, just before making a release please make sure I haven't exposed anything harmless, AFAIK there as some enum and structs that was used by the table also, they obviously are not part of ABI but it is possible "ABI Tracker" would complain about them if removed someday (as a `-Og -g` compile can contain some info about enums even)

PR #850, opened 2018-02-28
[aat/kerx] Make parsing of the table actually work

ebraminio:


behdad:
Please separate the CPAL change.

ebraminio:
Done. Actually this is has not the best title, I just want to make sure table parsing code is correct here, which is also WIP anyway and I hope be able to do it.

ebraminio:
The subtables are now hand tested with Apple fonts but of course their field datatype, logic and sanitization are not prefect. I guess that is as far as I could for now, so.

PR #851, opened 2018-02-28
[CPAL] Fix version checking logic on sanitizer

ebraminio:


behdad:
lgtm. Please add a commented-out V1Tail at the end of the CPAL struct. Check post table for what I mean.

Also, no new-line before the "public:" before "DEFINE_..."

PR #854, opened 2018-03-01
[COLR] More sanitization

ebraminio:
I guess with this, COLR implementation can be considered something useful.

### DIFF
--- a/src/hb-ot-colr-table.hh
+++ b/src/hb-ot-colr-table.hh
@@ -42,10 +42,10 @@ struct LayerRecord
   inline bool sanitize (hb_sanitize_context_t *c) const
   {
     TRACE_SANITIZE (this);
-    return_trace (true);
+    return_trace (c->check_struct (this));
   }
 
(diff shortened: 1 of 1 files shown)


src/hb-ot-colr-table.hh
```diff
@@ -74,9 +76,17 @@ struct COLR
...
+
+    const BaseGlyphRecord *base_glyph_records = &baseGlyphRecords (this);
+    for (unsigned int i = 0; i < numBaseGlyphRecords; ++i)
+      if (!(base_glyph_records[i].sanitize (c, this)))
+        return_trace (false);
+
+    return_trace (true);
   }
 
   protected:
```
ebraminio: Anything could be used instead 
```
   HBUINT16	numBaseGlyphRecords;	/* Number of Base Glyph Records */
   LOffsetTo<BaseGlyphRecord>
 		baseGlyphRecords;	/* Offset to Base Glyph records. */
```
?

PR #858, opened 2018-03-03
[ot-color] Move the related tables to hb-ot-color-*

### LINKED ISSUE #856: Renaming to hb-ot-color-*
Let's rename the color tables to have a hb-ot-color-* prefix. I'm still making my mind re the public API. It probably will be hb-ot-color.h, NOT hb-color.h

ebraminio:
Fixes #856


behdad:
Maybe move cbdt as well, if you feel like.

ebraminio:
I do, #860 so we know that we just lack `SVG ` and `sbix` now :)

Issue #849, opened 2018-02-28
Public API for COLR/CPAL


ebraminio:
Now that we have COLR/CPAL tables parsing, it would be nice if we expose some of it as a public API for clients so they won't need to write their own font parsing logic. All what is needed here is signature of public APIs we like to have (which of course is not that easy...). Some useful links:

* https://codereview.chromium.org/1984943002/
* https://bugzilla.mozilla.org/show_bug.cgi?id=889401
* https://bug889401.bmoattachments.org/attachment.cgi?id=8424617
* http://msdn.microsoft.com/en-us/goglobal/bb688099.aspx
* http://msdn.microsoft.com/en-us/library/windows/desktop/hh802480%28v=vs.85%29.aspx
* http://www.getcodesamples.com/src/712F9685/B41B1957
* https://gist.github.com/egtra/6045976
* https://github.com/wine-mirror/wine/blob/5ec6b8f807f61ee77b9a96d94798c8e3f3db7af4/dlls/dwrite/dwrite_private.h#L220-L238
* http://savannah.nongnu.org/bugs/?44689
* https://github.com/ShaoYuZhang/freetype/commit/27d9e6f6012b55d4001c33e3d46f6814c94ebc0d
* https://github.com/ShaoYuZhang/freetype/commit/27d9e6f6012b55d4001c33e3d46f6814c94ebc0d#diff-b4b664c297e0aefa0ecc9dd995dffdf2R30

behdad:
I think I like a hb_color_t type that is either a typedef to uint32_t and stored as ARGB; or a union type that has uint32_t as well as four uint8_t b, g, r, a channels.

ebraminio:
There is something similar developed by @brawer on https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ot-color.h#L51-L54 (that struct and enum are the only things currently exposed publicly)

ebraminio:
Also #859

ebraminio:
Interestingly it seems macOS has added COLR/CPAL and Chrome on macOS is currently befitting from it (but with the switch to freetype for sbix rendering, that will be regressed I guess).

ebraminio:
The fonts related to COLR/CPAL test on Mozilla, [emoji-yellow.zip](https://github.com/harfbuzz/harfbuzz/files/1790427/emoji-yellow.zip) https://bug889401.bmoattachments.org/attachment.cgi?id=8428570

ebraminio:
[...]

PR #859, opened 2018-03-03
[ot-color] Improve CPAL table structure

ebraminio:


src/hb-ot-color-cpal-table.hh
```diff
@@ -151,22 +134,17 @@ struct CPAL
     return numPalettes;
   }
 
-  inline void get_color_record (int palette_index, uint8_t &r, uint8_t &g,
-                                              uint8_t &b, uint8_t &a) const
+  inline uint32_t get_color_record_argb (int palette_index) const
   {
-    // We should check if palette_index is in range as it is not done on COLR sanitization
```
behdad: Wait wait. It's NEVER ok to have such code committed to HarfBuzz. Even if not hooked up. That's how bugs creep in.
As a minimum there should be a big "XXX" comment, which must be resolved before this code is ever called.  See, I cannot check every single array access later. It's the responsibility of original developer to clearly annotate any access checks that must be done but currently are not.

Also, because colorRecords is a C struct that needs access check (as opposed to an ArrayOf<> template which does the access check for you), it must be renamed to colorRecordsZ to signify that and make code review possible.

behdad: Lol. I just saw it *is* ArrayOf. So it's safe. colorRecordIndices however must be commented out and name end in X because it's not correctly placed and needs to be accessed through StructAfter<>.

src/hb-ot-color.h
```diff
@@ -50,7 +50,7 @@ HB_BEGIN_DECLS
  */
 typedef struct
 {
-  uint8_t red, green, blue, alpha;
+  uint8_t alpha, red, green, blue;
```
behdad: I'm now reminded that we cannot have a union of int and four uint8_t's as that would become endianness-dependent. So, I think we want it as a uint32_t. Kinda like a hb_tag_t. Essentially you treat the color as an integer.

behdad:
We can add convenience functions for creating a color from rgba values and back.

src/hb-ot-color-cpal-table.hh
ebraminio: Yes, just checked ArrayOf implementation before removing my own comment :)

src/hb-ot-color.h
ebraminio: So the removal hb_{ot_}color_t? To be honest I like that!

ebraminio:
> We can add convenience functions for creating a color from rgba values and back.

Where to put them?

```
  int r = color & 0xFF;
  int g = (color >> 8) & 0xFF;
  int b = (color >> 16) & 0xFF;
  int alpha = (color >> 24) & 0xFF;
```

behdad:
A typedef, like hb_tag_t.

src/hb-ot-color-cpal-table.hh
ebraminio: I notice now whole thing should be rewritten.

src/hb-ot-color-cpal-table.hh
```diff
@@ -151,22 +126,27 @@ struct CPAL
...
+  }
+
+  inline uint32_t get_color_record_argb (unsigned int color_index, unsigned int palette) const
   {
-    // We should check if palette_index is in range as it is not done on COLR sanitization
-    r = colorRecords[palette_index].red;
-    g = colorRecords[palette_index].green;
-    b = colorRecords[palette_index].blue;
-    a = colorRecords[palette_index].alpha;
+    if (color_index > numPaletteEntries || palette > numPalettes)
```
ebraminio: All comparisons should be looked deeper, ">" or ">="? ">=" makes more sense but on COLR ">" was only working.

ebraminio:
PTAL

ebraminio:
Replaced get_base_glyph_record with a bsearch, don't know of a better way (whether a currently available struct featuring bsearch could be used instead).

ebraminio:
#849

PR #868, opened 2018-03-07
[subset] Drop glyph names from post when subsetting.

garretrieger:


src/hb-ot-post-table.hh
```diff
@@ -82,6 +82,34 @@ struct post
     return_trace (true);
   }
 
+  inline bool subset (hb_subset_plan_t *plan) const
+  {
+    hb_blob_t *post_blob = OT::Sanitizer<post>().sanitize (hb_face_reference_table (plan->source, HB_OT_TAG_post));
+    post *post_table = (post *) hb_blob_get_data (post_blob, nullptr);
+
+    post *post_prime = (post *) malloc (post::static_size);
```
behdad: Please allocate fixed-size objects using calloc(..., 1).

src/hb-ot-post-table.hh
```diff
@@ -82,6 +82,34 @@ struct post
...
+    if (unlikely (!post_prime))
+    {
+      hb_blob_destroy (post_blob);
+      DEBUG_MSG(SUBSET, nullptr, "Unable to alloc %lu for post subset output.", (unsigned long) post::static_size);
+      return false;
+    }
+
+    memcpy (post_prime, post_table, post::static_size);
```
behdad: This actually does out-of-memory access if post_table is too short. Let me think what I suggest to use instead.

An idiom instead of malloc+memcpy is to create_sub_blob().

src/hb-ot-post-table.hh
garretrieger: Done. Reworked to use create sub blob.

src/hb-ot-post-table.hh
garretrieger: Fixed. Reworked to use create sub blob.

PR #871, opened 2018-03-08
[subset] Add a Japanese font to the subset integration tests.

garretrieger:


test/subset/run-tests.py
```diff
@@ -77,6 +78,7 @@ def run_test(test, should_check_ots):
 	return 0
 
 def run_ttx(file):
+	print ("ttx %s" % file)
```
rsheeter: I think it would be fine to print the command for all cmd() calls; always useful when troubleshooting

garretrieger: Good idea, I'll do that separate from this pr.

PR #874, opened 2018-03-09
[SVG ] Implementation

ebraminio:
Works on https://github.com/symbolset/ss-emoji-beta/blob/master/ss-emoji-adobe-mozilla.ttf and I was able to dump the SVGs inside the font with it!

### DIFF
--- a/src/Makefile.sources
+++ b/src/Makefile.sources
@@ -18,7 +18,6 @@ HB_BASE_sources = \
 	hb-open-file-private.hh \
 	hb-open-type-private.hh \
 	hb-ot-color-cbdt-table.hh \
-	hb-ot-color-sbix-table.hh \
 	hb-ot-cmap-table.hh \
 	hb-ot-glyf-table.hh \
 	hb-ot-hdmx-table.hh \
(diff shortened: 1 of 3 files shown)


ebraminio:
Works on https://github.com/adobe-fonts/emojione-color but needs a gz extractor in order to work.

KrasnayaPloshchad:
So there is a necessary to add hb-gz.cc in order to allow HarfBuzz build with gz, isn’t it?

ebraminio:
I think the short/medium plan is to provide the very gzipped blob to users so they can decode the blob for themselves and use their own SVG renderer (as librsvg or anything available on the client) with the provided blobs.

ebraminio:
Oh I think I was unclear on what I mean, I provided a temporarily and commented dump function that saves blob of the font on files. The functionality works great on the first font but on the Adobe fonts, it extracts the SVGs as well but they are not in correct format as the need to be extracted first, basically they are SVGZ instead SVG.

behdad:
Yeah, we don't decode it. Just return.

ebraminio:
The added tables are starving for public API Behdad :) I believe we can start with COLR/CPAL.

PR #875, opened 2018-03-10
[WIP][glyf] Implement glyph coordinates extractor

ebraminio:
WIP...

### DIFF
--- a/src/hb-ot-glyf-table.hh
+++ b/src/hb-ot-glyf-table.hh
@@ -133,9 +133,19 @@ struct glyf
     DEFINE_SIZE_STATIC (10);
   };
 
+  struct SimpleGlyph
+  {
+    //endPtsOfContours
+    //ArrayOf<HBUINT16> instructions;
+      // const GlyphHeader &glyph_header = StructAtOffset<GlyphHeader> (glyph_data, 0);
+      // if (glyph_header.numberOfContours < 0)
+    //DEFINE_SIZE_ARRAY (4);
+  };
+
   struct CompositeGlyphHeader
   {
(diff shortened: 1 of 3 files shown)


src/hb-ot-glyf-table.hh
```diff
@@ -275,16 +317,19 @@ struct glyf
 						 composite);
     }
 
+    static const int FLAG_ON_CURVE = 0x01;
```
behdad: Please use (anonymous) enum instead of "static const int". We're not Java :D.

ebraminio: :D cc: @garretrieger

ebraminio:
Going lights off for this and decrease opened PRs number.

behdad:
It's still a valid issue and we want to fix it. Please don't close valid issue and PRs just because of inactivity.

ebraminio:
Useful links,

* https://github.com/behdad/glyphy/blob/73d7fbd96975249c4eb806be94b9889df839335b/src/glyphy-freetype.h#L44-L95
* https://developer.apple.com/documentation/coretext/1510921-ctfontcreatepathforglyph?language=objc
* https://docs.microsoft.com/en-us/windows/desktop/api/wingdi/nf-wingdi-getglyphoutlinew
* https://docs.microsoft.com/en-us/windows/desktop/api/wingdi/nf-wingdi-getoutlinetextmetricsw

behdad:
Can we talk about the API here, in comments, first?

ebraminio:
> Can we talk about the API here, in comments, first?

#1686

ebraminio:
let's start a new PR

PR #877, opened 2018-03-12
Fix CMake builds on Windows, MSVC in particular

fanc999:
Hi,

The CMake build files evolved with more testing support, but also resulted in builds (and tests) failing on Windows, in particular the older versions of Visual Studio.  Please bear with me, as I am not sure of the compiler support policy, but this attempts to fix building and testing on Visual Studio.

Note that after this patchset all the tests (including shaping, fuzzing and subset, in addition to the API tests) pass on Visual Studio 2008/x86 and 2017/x64 builds, except for the tests/basics.tests (Failed) and tests/full-font.tests (Failed), as the which command could not be found, and it is only usable on Windows via MSYS or Cygwin builds, AFAICT, which I did not (yet) try to address here.

Note also that as we gained more source files for the main HarfBuzz/HarfBuzz-GObject DLLs, the command line that we need to build the introspection files becomes too long for Windows, so I am also currently looking for a way to fix this here, by using the --filelist option for g-ir-scanner, which I will try to address a little bit later.

With blessings, thank you!


fanc999:
Hmm,

Somehow I was having problems getting the image on Travis to recognize that it does indeed have round(), but nothing could convince it, even if I tried looking for round in the 'm' library, although the circleci had no complaints...

I am in no way knowledgeable in the math library stuff on *NIX, so I think I need some advice for this.

With blessings, thank you!

ebraminio:
> as I am not sure of the compiler support policy

I feel the policy is like if you can put it on CI so collaborators know they are breaking something with their change, then it is a supported compiler for a foreseeable future; we have a compiler for DOS on our CI (djgpp, which compiles 32bit executable of course) and gaming console related (psvita) compiler, nothing can be beat those I'd guess :)

ebraminio:
Tried to reproduce the issue locally on macOS with no success. I hope these help https://github.com/jedisct1/libpuzzle/blob/master/configure.ac#L45 and https://github.com/dajobe/rasqal/blob/master/configure.ac#L1214 and https://github.com/named-data/PyNDN2/blob/master/configure.ac#L66

fanc999:
Hi @ebraminio,

Thanks for the tips...

Seems that the GCC/CLang installation on Travis is older, so it didn't pick up the round() from math.h as it didn't support C99 flags when checking somehow (see the outputs).  So, as a workaround, I had it look for round()'s declaration in math.h if AC_CHECK_FUNCS([round]) did not work even as we do -lm...

Does not look the nicest, but at least it worked on Travis... :|

With blessings, thank you!

ebraminio:
Thank you very much@fanc999, this looks great! @behdad if the autotools change part make good I think this can be merged

ebraminio:
I really don't know how you change has broken msys2... I've added gen-def.py so I feel responsible for it somehow, can you see why it is failing? I guess maybe a change from `io.open(h, encoding='utf8')` to `io.open(h, 'rb', encoding='utf8')`. If not (and you had no another idea) we will merge this as is fix that later. Thank you

anthrotype:
> maybe a change from io.open(h, encoding='utf8') to io.open(h, 'rb', encoding='utf8') can fix it

The encoding argument is for reading text, can’t be used with rb, which is for reading bytes.

ebraminio:
Thank you, do you have any other idea? I couldn't reproduce this locally on my machine with a fairly similar configuration...

anthrotype:
you should check that the text file your are trying to read in as UTF-8 was actually encoded as UTF-8.
Maybe it was written with some different encoding. On Windows, the default encoding might not be UTF-8.

ebraminio:
The failure happens regardless of this patch and I think I have a solution for this so lets merge this first and then fix that

ebraminio:
Some essential part of this is reverted on https://github.com/harfbuzz/harfbuzz/commit/86a0ac284fc4c2ce96354bfdf32878b6db5c77bd @behdad

PR #879, opened 2018-03-12
[ot-color/sbix] imageOffsets range check

ebraminio:


src/hb-ot-color-sbix-table.hh
```diff
@@ -27,6 +27,12 @@
 
 #include "hb-open-type-private.hh"
 
+/*
+ * sbix -- Standard Bitmap Graphics Table
+ * https://docs.microsoft.com/en-us/typography/opentype/spec/sbix
+ * https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6sbix.html
```
ebraminio: There are more things on Apple's specification but currently no fonts on macOS to test them.

behdad: We'll just support the OpenType version. The rest is deprecated and there are no fonts out there.

behdad:
We don't want to do this check.  One of the design criteria is that we don't do O(num-glyphs) or any other heavy-weight checks at sanitize() time if they can be easily done at run time.  Ie. we need a sbix_accelerator_t that does the check.  That's actually why the glyf table sanitize is empty. You asked about it the other day...

(*If* we wanted though, we should have checked that the offsets are non-decreasing.)

src/hb-ot-color-sbix-table.hh
```diff
@@ -58,10 +64,17 @@ struct SBIXStrike
...
+    if (!(c->check_struct (this) &&
 		  c->check_array (imageOffsetsZ,
 				  sizeof (HBUINT32),
-				  1 + c->num_glyphs));
+				  1 + c->num_glyphs)))
+      return_trace (false);
+
+    for (unsigned int i = 0; i < 1 + c->num_glyphs; ++i)
+      if (!c->check_range (this, imageOffsetsZ[i]))
+        return_trace (false);
```
ebraminio: How about first finding the maximum imageOffsetsZ value then checking if that is in range.

ebraminio:
Thank you Behdad, I wanted to ask also what is the point of having accelerator_t and I am gradually understanding its role :)

src/hb-ot-color-sbix-table.hh
behdad: How is that not O(num-glyphs)? ;)

behdad:
> Thank you Behdad, I wanted to ask also what is the point of having accelerator_t and I am gradually understanding its role :)

The accelerator pattern is used when we need to carry extra information for efficient use of the table. Ie. when just having a pointer to the table itself is not enough.

PR #884, opened 2018-03-13
[dwrite] GCC/mingw/msys2 compatibility

ebraminio:
It now compiles on GCC also but msys2 needs to be patched like this https://github.com/Alexpux/mingw-w64/pull/5

### DIFF
--- a/src/hb-directwrite.cc
+++ b/src/hb-directwrite.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2015-2016  Ebrahim Byagowi
+ * Copyright © 2015-2018  Ebrahim Byagowi
  *
  *  This is part of HarfBuzz, a text shaping library.
  *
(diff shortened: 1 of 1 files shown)


src/hb-directwrite.cc
```diff
@@ -919,7 +924,7 @@ hb_directwrite_shape_experimental_width(hb_font_t          *font,
   unsigned int        num_features,
   float               width)
 {
-  static char *shapers = "directwrite";
+  static char *shapers = (char *) "directwrite";
```
behdad: This is wrong. You should just declare "static const char *...".

behdad:
Can you explain what this change does?

src/hb-directwrite.cc
ebraminio: Oh, thanks.

ebraminio:
> Can you explain what this change does?

hb-directwrite was never compiled on GCC, I fixed headers missing of mingw64 project and applied fixes needed to get it working on GCC/msys2.

ebraminio:
* Replaced new/delete with plain malloc to get it working on autotools.
* Replaced the places false meant `nullptr` actually but MSVC didn't complain.
* Fixed the places that compiler raised warnings, of course using const was more appropriate and I just wanted to make it happy, and I wasn't enlightened with `const` as the time originally writing this! :)

There are more improvement needed for hb-directwrite of course now that I see but as hb-directwrite wasn't considered that important from the beginning I just wanted to fix it on GCC and no more.

behdad:
I see.  Thanks. That's a lot more helpful :).

PR #885, opened 2018-03-13
[ci] Try to fix gen-def issue on appveyor

ebraminio:


src/gen-def.py
```diff
@@ -7,7 +7,7 @@
 headers_content = []
 for h in os.environ["headers"].split (' '):
 	if h.endswith (".h"):
-		with io.open(h, encoding='utf8') as f: headers_content.append (f.read ())
+		with io.open (h, encoding='ISO-8859-1') as f: headers_content.append (f.read ())
```
behdad: Can you explain why this is needed? Why not ASCII? Our headers should be UTF-8. latin1 definitely doesn't sound like has any place here.

ebraminio: I know that this doesn't make sense but a msys2 bot didn't like that.

anthrotype: try to print the name of the header that was failing with UnicodeDecodeError when being read as UTF-8, and make sure it is indeed encoded as UTF-8.

ebraminio: The issue is `glib-mkenums` generated `hb-gobject-enums.h` which is in ASCII on msys2 apparently but I am not sure about the fix

anthrotype:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=870310

anthrotype:
probably fixed in the latest glib, see https://bugzilla.gnome.org/show_bug.cgi?id=785113

ebraminio:
As far as I can remember the version that was on appveyor had this patch applied https://bug785113.bugzilla-attachments.gnome.org/attachment.cgi?id=356410 msys2 packages, shaped like ArchLinux usually has updated packages so probably `getpreferredencoding` is the key to the issue I guess.

anthrotype:
are we piping to stdout? Try setting `PYTHONIOENCODING=UTF-8` and see if that works

PR #886, opened 2018-03-14
CMake: hide symbols by default on non-MSVC, define export attribute

madig:
This change makes sure that non-MSVC builds using CMake properly hide
non-exported functions.

We haven't reached symbol-parity yet though:
Autotools symbol list on my machine: https://paste.fedoraproject.org/paste/VndugIDtEGsXj~UDZJitzw
CMake build symbol list: https://paste.fedoraproject.org/paste/LmK7tRcdWAYAg-Y~BGq1Rw

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,7 @@
 cmake_minimum_required(VERSION 2.8.0)
+# Allow symbol hiding for both static and shared libs.
+cmake_policy(SET CMP0063 NEW)
+
 project(harfbuzz)
 
 enable_testing()
@@ -524,23 +527,19 @@ endif ()
 
 ## Define harfbuzz library
 add_library(harfbuzz ${project_sources} ${project_extra_sources} ${project_headers})
+set_target_properties(harfbuzz PROPERTIES
+  C_VISIBILITY_PRESET hidden
+  CXX_VISIBILITY_PRESET hidden
+  VISIBILITY_INLINES_HIDDEN TRUE)
 target_link_libraries(harfbuzz ${THIRD_PARTY_LIBS})
-
-if (BUILD_SHARED_LIBS)
-  if (WIN32 AND NOT MINGW)
-    add_definitions("-DHB_EXTERN=__declspec(dllexport) extern")
-  else ()
-    add_definitions("-DHB_EXTERN=__attribute__(( visibility( \"default\" ) )) extern")
-    set_target_properties(harfbuzz PROPERTIES
-      C_VISIBILITY_PRESET hidden
-      CXX_VISIBILITY_PRESET hidden
-      VISIBILITY_INLINES_HIDDEN TRUE)
-  endif ()
-endif ()
 
 ## Define harfbuzz-subset library
 add_library(harfbuzz-subset ${subset_project_sources} ${subset_project_headers})
 add_dependencies(harfbuzz-subset harfbuzz)
+set_target_properties(harfbuzz-subset PROPERTIES
+  C_VISIBILITY_PRESET hidden
+  CXX_VISIBILITY_PRESET hidden
+  VISIBILITY_INLINES_HIDDEN TRUE)
 target_link_libraries(harfbuzz-subset harfbuzz ${THIRD_PARTY_LIBS})
 
 if (UNIX OR MINGW)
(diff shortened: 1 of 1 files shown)


ebraminio:
Hi there. Thanks for the effort! It seems CMake's exported symbols test (which is common with autotools) is not happy about the change.

madig:
Shouldn't HB_EXTERN be defined in a source header instead of the build system? Like shown on https://gcc.gnu.org/wiki/Visibility.

ebraminio:
I think that make sense! Of course @behdad should know why the project moved that but feel free to apply that also if you like :)

madig:
Can do, unless Behdad is against it :)

ebraminio:
I say go ahead and put it as a PR. It is something worthy and worth the try I'd say :)

behdad:
Please don't push any cmake-only visibility changes in.

Let's first agree on which direction we want to go. I don't have time to restudy everything right now, so if you can explain it to me in simple terms and show library size before/after, that would be helpful.

The way we are currently set up is using default visibility, and hiding internal symbols. You want to go the other direction, which is nicer for static builds but shouldn't matter for shared builds as the two should produce the same results.

madig:
What prompted this was an uncontrolled symbol export when statically linking Harfbuzz into FT for distribution of freetype-py, noticed when using the CMake build of FreeType which didn't hide symbols (whereas the autotools build does by defaulting to hidden and using a symbol file to further weed out symbols). MSVC didn't care because it hides by default and needs to be told what to export.

The patch makes it so that everything is hidden, unless marked with the HB_EXPORT macro and should bring roughly symbol parity with MSVC output. It is supposed to be a consistency improvement so to speak :) Isn't HB de facto using hidden visibility by default?

Before this patch:
Autotools default shared build (Fedora 27 x64):
- File size: 791040 bytes
- Symbols: https://paste.fedoraproject.org/paste/zc5cSrKpb1a9oCWhGNCUUw

CMake default build:
- File size: 1295320 bytes (no idea why)
- Symbols: https://paste.fedoraproject.org/paste/lEsMbI3TiRI9lEdo5xvq1Q

Symbol diff: https://paste.fedoraproject.org/paste/ySuzj4e9hc7JsOl52WnwNw -- so CMake isn't picking up GLib and FT by default.

After patch:
Autotools default shared build:
- File size: 791040 bytes
- Symbols: https://paste.fedoraproject.org/paste/95a~a-TvV3WVTpI-xw9fmg

CMake default build:
- File size: 1295320 bytes
- Symbols: https://paste.fedoraproject.org/paste/suBl2SqXr1tXazu9TA0NeA

Symbol diff: https://paste.fedoraproject.org/paste/VCwreU539Lnf-EtXlBphKA -- the same.

I can't compare static Autotools and CMake output because master with autotools --enable-static -- disable-shared fails me with:
```
/home/nikolaus/Entwicklung/harfbuzz/src/.libs/libharfbuzz.a(libharfbuzz_la-hb-ot-layout.o):(.rodata+0x280): multiple definition of `OT::_hb_NullPool'
../src/.libs/libharfbuzz-subset.a(libharfbuzz_subset_la-hb-subset.o):(.rodata+0x5e0): first defined here
```

madig:
A quick comparison using my [FT bundler script](https://github.com/madig/freetype-py/blob/setup-ci/setup-build-freetype.py), comparing final FT shared object file size with a statically linked in HB: 2384952 bytes before this change, 2083544 bytes after. That's a free 300 KB shaved off. No idea why a MSVC compile is half of that though...

madig:
Isn't it just a semantic flip? Instead of being careful to hide everything (blacklisting), you have to be careful to export what you want (whitelisting). The latter should be easier.

Size before:
```
   text	   data	    bss	    dec	    hex	filename
1938886	  46224	     88	1985198	 1e4aae	build/local/lib/libfreetype.so
```
After:
```
   text	   data	    bss	    dec	    hex	filename
1652304	  31312	     88	1683704	 19b0f8	build/local/lib/libfreetype.so
```
After, with just `VISIBILITY_INLINES_HIDDEN TRUE` on all libs:
```
   text	   data	    bss	    dec	    hex	filename
1673836	  32320	     88	1706244	 1a0904	build/local/lib/libfreetype.so
```
This actually takes care of the rampant symbolage, so if you're uncomfortable with hiding everything, revert the commit and I'll make it use just VISIBILITY_INLINES_HIDDEN TRUE.
If you use my FT build script, keep in mind that the FT 2.9 release CMake build is broken insofar it uses default visibility instead of hidden like the autotools build.

ebraminio:
Thanks all, proposed #891.

PR #887, opened 2018-03-15
[subset] Add a fuzzing target for hb-subset.

garretrieger:


behdad:
Thanks! Can you also rename hb-fuzzer to hb-shape-fuzzer? Also, any way to reduce duplication of changes in src/Makefile.am?

behdad:
Also need to figure out how to convince oss-fuzz to run both. You need to send a PR to this: https://github.com/google/oss-fuzz/tree/master/projects/harfbuzz

behdad:
Humm. So you disabled it actually? Please push any followups to a separate PR.

Also need oss-fuzz changes to go with it. Please CC me on those.

garretrieger:
Yes sorry, I was trying to debug the failing msys2 bot by figuring out what
was causing hb-subset-fuzzer to fail to run. I'll fix it up and send
another pr.

Issue #883, opened 2018-03-13
Build Fails on QNX


INRIX-joel-winarske:
[ 68%] Linking CXX executable test-test
cc: unknown option: -Bsymbolic-functions

This is the fix, should I create a PR?

    diff --git a/CMakeLists.txt b/CMakeLists.txt
    index defd5d6..99526cc 100644
    --- a/CMakeLists.txt
    +++ b/CMakeLists.txt
    @@ -533,7 +533,9 @@ target_link_libraries(harfbuzz-subset harfbuzz ${THIRD_PARTY_LIBS})
 
     if (UNIX OR MINGW)
       # Make symbols link locally
    -  link_libraries(-Bsymbolic-functions)
    +  if(NOT(${CMAKE_SYSTEM_NAME} STREQUAL "QNX"))
    +      link_libraries(-Bsymbolic-functions)
    +  endif()
 
       if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
         # Make sure we don't link to libstdc++

ebraminio:
Sure, go ahead :) What is the used compiler btw? I wish also we could have a CI bot testing harfbuzz with the used compiler.

behdad:
Please keep cmake and autotools in sync.
I'm starting to regret allowing cmake in :)))

ebraminio:
Well, for some platforms it seems possible autotools can't be used, I guess we could be relax about that.

jwinarske:
It's based on GCC 5.4.0, with rename as QCC:

    -- The C compiler identification is QCC 5.4.0
    -- The CXX compiler identification is QCC 5.4.0

My fix resolves the problem in a compiler agnostic way, by testing the flag.
The dependency selection seems more granular in the CMake build, over autotools.  More changes would be required to build with autotools using the QNX toolchain, that are not related.

jwinarske:
[PR #889](https://github.com/harfbuzz/harfbuzz/pull/889)

ebraminio:
Done on both cmake and autotools

PR #889, opened 2018-03-15
Add compiler check for -Bsymbolic-functions

jwinarske:
Fix for reported issue 883 - [Build Fails on QNX](https://github.com/harfbuzz/harfbuzz/issues/883)


ebraminio:
Considering Behdad's opinion on keeping cmake and autotools in sync, this I believe just needs the same check for autotools. If you couldn't or had't time for that, please let me know so I will try to implement the similar logic for autotools :)

ebraminio:
Autotools part uploaded on #892.

PR #890, opened 2018-03-16
[dwrite] Use new again and enable the build on msys2 bots

ebraminio:
#884 fixed most of the GCC compile issues, this enables it on msys2 bots.

<s>The way I am checking validity of headers may don't seem good. The issue is msys2 and mingw-w64 headers are not complete and I am checking that using that `#ifdef`. I proposed a fix to upstream also but that can take years to have here so I am applying this workaround.</s> this now just downloads the needed header

Also applied https://github.com/harfbuzz/harfbuzz/pull/884#discussion_r174845362 now that it doesn't have any other hb-directwrite.cc perhaps should be separated but don't know if worths to

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,11 +20,11 @@ environment:
 
 
     - compiler: msys2
-      MINGW_PREFIX: /c/msys2/mingw64/
+      MINGW_PREFIX: /mingw64
       MINGW_CHOST: x86_64-w64-mingw32
       MSYS2_ARCH: x86_64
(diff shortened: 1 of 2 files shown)


ebraminio:
I should also note that hb-shape crashes if I issue `--shaper=directwrite`. I should fix that also.

ebraminio:
The crashing issue was because of lack of vtable initialization with plain malloc that I couldn't understand without @andlabs help. Thanks!

PR #891, opened 2018-03-16
[cmake] Don't hide symbols by default

ebraminio:
but keep use of cmake idiomatic way of making inlines hidden. Partially reverts #886 but keeps the good part.


behdad:
Thank you. We can figure out if there's a better symbol export system that is more efficient later and switch both build systems to it.

PR #893, opened 2018-03-16
hb-private.hh: Add fallback implementation for round()

fanc999:
(Please see discussion at commit 86a0ac2 for more details for re-adding
this patch.)

With blessings, thank you!


ebraminio:
Related to #877

PR #894, opened 2018-03-16
Appropriate fix for msys2 bot fail on gen-def.py

ebraminio:
Reverts the hack happened here #885 with @anthrotype help. Thanks!

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -234,7 +234,7 @@ DISTCLEANFILES += \
 	$(HB_GOBJECT_ENUM_headers) \
 	$(NULL)
 hb-gobject-enums.%: hb-gobject-enums.%.tmpl $(HBHEADERS)
-	$(AM_V_GEN) $(GLIB_MKENUMS) \
+	$(AM_V_GEN) PYTHONIOENCODING=UTF-8 $(GLIB_MKENUMS) \
 		--identifier-prefix hb_ --symbol-prefix hb_gobject \
 		--template $^ | \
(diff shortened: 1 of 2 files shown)


anthrotype:
I hope that doesn't break other stuff :)
it also changes the encoding of the stdout/stderr streams, so if we print anything to the console and the latter uses some weird ANSI codepage.. 

Maybe you should only set that env variable when calling glib-mkenums

anthrotype:
https://ci.appveyor.com/project/harfbuzz/harfbuzz/build/1.0.1076/job/ici5gkiglm79meo6#L556

still getting UnicodeDecodeError.

Are we sure that cmake's execute_process that calls the glib-mkenums script is reading the PYTHONIOENCODING variable properly?

maybe you could try doing `SET(ENV{PYTHONIOENCODING} "UTF-8")` just before `EXECUTE_PROCESS(...)`? 
https://cmake.org/pipermail/cmake/2011-August/045665.html

ebraminio:
msys2 bot which is failing uses autotools, Behdad's territory, I fear to touch it and then get blamed :D

behdad:
Um. No blame. :) I'll check.

ebraminio:
Hehehe, done, lets see.

anthrotype:
> msys2 bot which is failing uses autotools

oh I see. I presume it would give the same error if one attempted to run glib-mkenums under cmake on msys2, if that env variable isn't set.

ebraminio:
@anthrotype does it look good now? If so I think this can be merged now.

PR #895, opened 2018-03-16
[subset] Restore subset to hb-subset-fuzzer.

garretrieger:


garretrieger:
@ebraminio Hey I'm having some issues with a single windows bot failing on this PR. I'm not able to replicate on a windows machine locally. Could you help me track down why it's failing?

ebraminio:
msys2 bots usually complain about an interesting aspect of the project and this one is a 32bit one and IIRC it is the only autotools bot that runs tests on 32bit so the bug can be about some 32-bit `int` variable not enough to contain all the needed range. It has one issue for me, msys2's `gdb`, at least on my machine, doesn't seem to work so lets make sure your setup is correct,

1. Install msys2 from https://www.msys2.org
2. Open "MSYS2 MingGW 32-bit" shortcut
3. `pacman --noconfirm -S mingw-w64-i686-{freetype,cairo,icu,gettext,gobject-introspection,gcc,gcc-libs,glib2,graphite2,pkg-config,python2,ragel}`
4. `./autogen.sh --with-uniscribe --with-freetype --with-glib --with-gobject --with-cairo --with-icu --with-graphite2`
5. `make`
6. `make check`

If this couldn't reproduce the issue let me know.

ebraminio:
Weird, it didn't have issue when I ran the instruction inside the bot, so my argument was not valid, maybe it just needs rebuild (which if is the case we should still be concerned but at least we know it doesn't happen always).

ebraminio:
Restarted the bot but the fail was consistent :/

garretrieger:
I tried your steps and was also unable to reproduce. Any other ideas of what the cause could be?

I think I'll just start commenting out parts of the code and checking that against the bots to try and narrow down where specifically is causing issues.

ebraminio:
Oh interesting, any idea how it is fixed now? That doesn't matter of course but is just interesting :)

garretrieger:
hmmm, all I did was rebase against master and re-ran the bots. I suppose some recent change must have fixed something?

PR #898, opened 2018-03-21
[subset] Various fixes for problems reported by oss-fuzz

garretrieger:


src/hb-ot-glyf-table.hh
```diff
@@ -232,9 +232,11 @@ struct glyf
   {
     inline void init (hb_face_t *face)
     {
+      memset (this, 0, sizeof (accelerator_t));
+
       hb_blob_t *head_blob = Sanitizer<head>().sanitize (face->reference_table (HB_OT_TAG_head));
       const head *head_table = Sanitizer<head>::lock_instance (head_blob);
-      if ((unsigned int) head_table->indexToLocFormat > 1 || head_table->glyphDataFormat != 0)
+      if (!head_table || (unsigned int) head_table->indexToLocFormat > 1 || head_table->glyphDataFormat != 0)
```
behdad: Humm. I thought head_table will be the null table in case of failure?

behdad:
lgtm if bots happy. Thanks!

src/hb-ot-glyf-table.hh
garretrieger: Fixed.

src/hb-ot-glyf-table.hh
```diff
@@ -236,7 +236,7 @@ struct glyf
 
       hb_blob_t *head_blob = Sanitizer<head>().sanitize (face->reference_table (HB_OT_TAG_head));
       const head *head_table = Sanitizer<head>::lock_instance (head_blob);
-      if (!head_table || (unsigned int) head_table->indexToLocFormat > 1 || head_table->glyphDataFormat != 0)
+      if (head_table == &Null(head) || (unsigned int) head_table->indexToLocFormat > 1 || head_table->glyphDataFormat != 0)
```
behdad: Technically we shouldn't need the first check. If head_table is Null, then indexToLocFormat will be 0. So, just change the second check to check for == 1?

behdad: Oh, my bad. Yeah maybe this is find.

garretrieger: Fixed.

PR #899, opened 2018-03-22
Fix llvm-gcc-4.2 compile issue and add a macOS bot to test it

### LINKED ISSUE #853: Build: Undefined symbols for architecture x86_64 -- "OT::KernOT::SubTableWrapper::min_size" and "OT::KernAAT::SubTableWrapper::min_size"
When building HarfBuzz 1.7.5, `make` was terminated with error messages saying that "OT::KernOT::SubTableWrapper::min_size" and "OT::KernAAT::SubTableWrapper::min_size" are undefined symbols for architecture x86_64.

### I did the following

export PKG_CONFIG='/opt/pkg-config/bin/pkg-config'
export PKG_CONFIG_PATH='/opt/pkg-config/lib/pkgconfig'
./configure
make

### Actual Result

`make` was terminated with the following error messages.

      GEN      libharfbuzz.la
    Undefined symbols for architecture x86_64:
      "OT::KernOT::SubTableWrapper::min_size", referenced from:
          OT::KernTable<OT::KernOT>::sanitize(OT::hb_sanitize_context_t*) const in libharfbuzz_la-hb-ot-font.o
      "OT::KernAAT::SubTableWrapper::min_size", referenced from:
          OT::KernSubTableWrapper<OT::KernAAT::SubTableWrapper>::sanitize(OT::hb_sanitize_context_t*) const  in libharfbuzz_la-hb-ot-font.o
    ld: symbol(s) not found for architecture x86_64
    collect2: ld returned 1 exit status
    make[4]: *** [libharfbuzz.la] Error 1
    make[3]: *** [all-recursive] Error 1
    make[2]: *** [all] Error 2
    make[1]: *** [all-recursive] Error 1
    make: *** [all] Error 2

### Environment

HarfBuzz 1.7.5
Mac OS X
Intel 64-bit

ebraminio:
Fixes #853

### DIFF
--- a/src/hb-open-file-private.hh
+++ b/src/hb-open-file-private.hh
@@ -295,11 +295,13 @@ struct OpenTypeFontFile
 {
   static const hb_tag_t tableTag	= HB_TAG ('_','_','_','_'); /* Sanitizer needs this. */
 
-  static const hb_tag_t CFFTag		= HB_TAG ('O','T','T','O'); /* OpenType with Postscript outlines */
-  static const hb_tag_t TrueTypeTag	= HB_TAG ( 0 , 1 , 0 , 0 ); /* OpenType with TrueType outlines */
-  static const hb_tag_t TTCTag		= HB_TAG ('t','t','c','f'); /* TrueType Collection */
-  static const hb_tag_t TrueTag		= HB_TAG ('t','r','u','e'); /* Obsolete Apple TrueType */
-  static const hb_tag_t Typ1Tag		= HB_TAG ('t','y','p','1'); /* Obsolete Apple Type1 font in SFNT container */
+  enum {
+    CFFTag		= HB_TAG ('O','T','T','O'), /* OpenType with Postscript outlines */
+    TrueTypeTag	= HB_TAG ( 0 , 1 , 0 , 0 ), /* OpenType with TrueType outlines */
+    TTCTag		= HB_TAG ('t','t','c','f'), /* TrueType Collection */
+    TrueTag		= HB_TAG ('t','r','u','e'), /* Obsolete Apple TrueType */
+    Typ1Tag		= HB_TAG ('t','y','p','1')  /* Obsolete Apple Type1 font in SFNT container */
+  };
 
   inline hb_tag_t get_tag (void) const { return u.tag; }
 
--- a/src/hb-ot-glyf-table.hh
+++ b/src/hb-ot-glyf-table.hh
@@ -134,18 +134,20 @@ struct glyf
 
   struct CompositeGlyphHeader
   {
-    static const uint16_t ARG_1_AND_2_ARE_WORDS =      0x0001;
-    static const uint16_t ARGS_ARE_XY_VALUES =         0x0002;
-    static const uint16_t ROUND_XY_TO_GRID =           0x0004;
-    static const uint16_t WE_HAVE_A_SCALE =            0x0008;
-    static const uint16_t MORE_COMPONENTS =            0x0020;
-    static const uint16_t WE_HAVE_AN_X_AND_Y_SCALE =   0x0040;
-    static const uint16_t WE_HAVE_A_TWO_BY_TWO =       0x0080;
-    static const uint16_t WE_HAVE_INSTRUCTIONS =       0x0100;
-    static const uint16_t USE_MY_METRICS =             0x0200;
-    static const uint16_t OVERLAP_COMPOUND =           0x0400;
-    static const uint16_t SCALED_COMPONENT_OFFSET =    0x0800;
-    static const uint16_t UNSCALED_COMPONENT_OFFSET =  0x1000;
+    enum composite_glyph_flag_t {
+      ARG_1_AND_2_ARE_WORDS =      0x0001,
+      ARGS_ARE_XY_VALUES =         0x0002,
+      ROUND_XY_TO_GRID =           0x0004,
+      WE_HAVE_A_SCALE =            0x0008,
+      MORE_COMPONENTS =            0x0020,
+      WE_HAVE_AN_X_AND_Y_SCALE =   0x0040,
+      WE_HAVE_A_TWO_BY_TWO =       0x0080,
+      WE_HAVE_INSTRUCTIONS =       0x0100,
+      USE_MY_METRICS =             0x0200,
+      OVERLAP_COMPOUND =           0x0400,
+      SCALED_COMPONENT_OFFSET =    0x0800,
+      UNSCALED_COMPONENT_OFFSET =  0x1000
+    };
 
     HBUINT16 flags;
     HBUINT16 glyphIndex;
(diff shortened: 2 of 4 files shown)


ebraminio:
Interestingly llvm-gcc-4.2 hates `static const int` so much (with different grades) which is a good code style check also!

src/hb-ot-glyf-table.hh
```diff
@@ -134,18 +134,20 @@ struct glyf
...
+      WE_HAVE_A_SCALE =            0x0008,
+      MORE_COMPONENTS =            0x0020,
+      WE_HAVE_AN_X_AND_Y_SCALE =   0x0040,
+      WE_HAVE_A_TWO_BY_TWO =       0x0080,
+      WE_HAVE_INSTRUCTIONS =       0x0100,
+      USE_MY_METRICS =             0x0200,
+      OVERLAP_COMPOUND =           0x0400,
+      SCALED_COMPONENT_OFFSET =    0x0800,
+      UNSCALED_COMPONENT_OFFSET =  0x1000
+    };
```
ebraminio: This wasn't essential but now that we have the simple_glyph_flag_t change lets have this also.

src/hb-ot-glyf-table.hh
```diff
@@ -280,16 +282,18 @@ struct glyf
 						 composite);
     }
 
+    enum simple_glyph_flag_t {
+      FLAG_X_SHORT = 0x02,
+      FLAG_Y_SHORT = 0x04,
+      FLAG_REPEAT = 0x08,
+      FLAG_X_SAME = 0x10,
+      FLAG_Y_SAME = 0x20
+    };
```
ebraminio: This was needed on llvm-gcc-4.2 exported symbols tests, without it was saying these are exported.

src/hb-open-file-private.hh
```diff
@@ -295,11 +295,13 @@ struct OpenTypeFontFile
...
-  static const hb_tag_t TTCTag		= HB_TAG ('t','t','c','f'); /* TrueType Collection */
-  static const hb_tag_t TrueTag		= HB_TAG ('t','r','u','e'); /* Obsolete Apple TrueType */
-  static const hb_tag_t Typ1Tag		= HB_TAG ('t','y','p','1'); /* Obsolete Apple Type1 font in SFNT container */
+  enum {
+    CFFTag		= HB_TAG ('O','T','T','O'), /* OpenType with Postscript outlines */
+    TrueTypeTag	= HB_TAG ( 0 , 1 , 0 , 0 ), /* OpenType with TrueType outlines */
+    TTCTag		= HB_TAG ('t','t','c','f'), /* TrueType Collection */
+    TrueTag		= HB_TAG ('t','r','u','e'), /* Obsolete Apple TrueType */
+    Typ1Tag		= HB_TAG ('t','y','p','1')  /* Obsolete Apple Type1 font in SFNT container */
+  };
```
ebraminio: This was needed otherwise linker had this issue,

```
  CXXLD    libharfbuzz-subset.la
Undefined symbols for architecture x86_64:
  "OT::OpenTypeFontFile::TrueTypeTag", referenced from:
      _hb_subset_face_reference_table(hb_face_t*, unsigned int, void*)in libharfbuzz_subset_la-hb-subset.o
  "OT::OpenTypeFontFile::CFFTag", referenced from:
      _hb_subset_face_reference_table(hb_face_t*, unsigned int, void*)in libharfbuzz_subset_la-hb-subset.o
ld: symbol(s) not found for architecture x86_64
```

src/hb-ot-kern-table.hh
```diff
@@ -205,7 +205,7 @@ struct KernSubTableWrapper
   {
     TRACE_SANITIZE (this);
     return_trace (c->check_struct (thiz()) &&
-		  thiz()->length >= thiz()->min_size &&
+		  thiz()->length >= T::min_size &&
```
ebraminio: Needed as the original issue linking issue,

```
  GEN      libharfbuzz.la
Undefined symbols for architecture x86_64:
  "OT::KernOT::SubTableWrapper::min_size", referenced from:
      OT::KernSubTableWrapper<OT::KernOT::SubTableWrapper>::sanitize(OT::hb_sanitize_context_t*) const  in libharfbuzz_la-hb-ot-font.o
  "OT::KernAAT::SubTableWrapper::min_size", referenced from:
      OT::KernSubTableWrapper<OT::KernAAT::SubTableWrapper>::sanitize(OT::hb_sanitize_context_t*) const  in libharfbuzz_la-hb-ot-font.o
ld: symbol(s) not found for architecture x86_64
collect2: ld returned 1 exit status
```

src/hb-open-file-private.hh
behdad: That's some really stupid compiler...

I'm not sure what to feel about this change. We have many other places that we define `static const hb_tag_t` values. This is now inconsistent.

Do we care about that particular compiler? Honest question.

src/hb-ot-glyf-table.hh
behdad: These are good. I was actually going to suggest you do.

src/hb-ot-kern-table.hh
behdad: Ok, good. So you didn't have to duplicate sanitize after all.

src/hb-open-file-private.hh
khaledhosny: I’d say we shouldn’t, it is an ancient compiler for an ancient, unsupported by its vendor, platform. Downstream can maintain there own patch-set if they want, it is not like HarfBuzz is the only package that will break on it.

ebraminio: The fact,

* It provides useful insights and it just needs some trivial changes which some are considered nice,
* There are people complaining about it, #853 and https://github.com/macports/macports-ports/pull/1466
* We can have it on our CI and we can have SSH access to the bot when you don't know what is wrong

makes it eligible for short term support. If you guys don't agree, at least lets have one release with it, write some compiler policy somewhere about it and leave the support those don't fit (removing their bots).

ebraminio: For example see https://bugs.chromium.org/p/chromium/issues/detail?id=820422 and https://groups.google.com/a/chromium.org/forum/#!topic/cxx/N7weI2XylMo

> I assume we would move to a community-supported model for MSVC in Chromium's code – no bot coverage, but patches to keep things building are accepted – just like our current gcc setup.

Even on not supported compilers, it would be nice to accept needed patches at least.

And from, http://blog.llvm.org/2018/03/clang-is-now-used-to-build-chrome-for.html

> Building your project with several compilers is generally considered good for code health, but in Chrome we found that Clang’s diagnostics found most problems and we were mostly battling compiler bugs (and if another compiler has a great new diagnostic, we can add that to Clang).

If we follow Chromium lead here [sarcastically], we can drop both MSVC and GCC support and bots :)

behdad: I understand the value of different compilers and bots. As long as we know why we are doing it and the workaround are not ugly. In this case, we clearly are working around a compiler bug.

I just miss the days that all our commits were bug fixes or feature implementations. :) Guess that's the price of popularity... :D

PR #900, opened 2018-03-23
CMake: Policy 0063 must be enabled to have hidden inline visibility on static builds

madig:
Tested with `cmake -E make_directory build && cmake -E chdir build cmake .. && cmake --build build --config Release` and `cmake -E remove_directory build && cmake -E make_directory build && cmake -E chdir build cmake -DBUILD_SHARED_LIBS=ON .. && cmake --build build --config Release`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,6 @@
 cmake_minimum_required(VERSION 2.8.0)
+cmake_policy(SET CMP0063 NEW)
+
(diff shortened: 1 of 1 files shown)


ebraminio:
Oh I see. Thanks! Btw, you could use `cmake -Bbuild -H. && cmake --build build --config Release` also :)

madig:
Thanks, those aren't documented in --help :)

PR #909, opened 2018-03-25
Implement an internal emojis dumper tool

ebraminio:
The plan is to have some opinionated but private developed tool to dump all the glyphs using the original format or well formed SVGs. Currently the only `sbix` is somehow OK but that also doesn't do runtime sanitizing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -543,7 +543,7 @@ if (UNIX OR MINGW)
   check_cxx_compiler_flag(-Bsymbolic-functions CXX_SUPPORTS_FLAG_BSYMB_FUNCS)
   if(CXX_SUPPORTS_FLAG_BSYMB_FUNCS)
     link_libraries(-Bsymbolic-functions)
-  endif()  
+  endif()
 
   if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     # Make sure we don't link to libstdc++
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -368,11 +368,15 @@ dist_check_SCRIPTS += \
 endif
 
 check_PROGRAMS += \
+	dump-emoji \
 	dump-indic-data \
 	dump-khmer-data \
 	dump-myanmar-data \
 	dump-use-data \
 	$(NULL)
+dump_emoji_SOURCES = dump-emoji.cc
+dump_emoji_CPPFLAGS = $(HBCFLAGS)
+dump_emoji_LDADD = libharfbuzz.la $(HBLIBS)
 dump_indic_data_SOURCES = dump-indic-data.cc hb-ot-shape-complex-indic-table.cc
 dump_indic_data_CPPFLAGS = $(HBCFLAGS)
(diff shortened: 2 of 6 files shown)


src/dump-glyphs.cc
```diff
@@ -0,0 +1,69 @@
...
+  char *font_blob = (char*) malloc (font_file_size);
+  fread (font_blob, font_file_size, 1, font_file);
+  fclose (font_file);
+
+  hb_blob_t *blob = hb_blob_create (font_blob, font_file_size,
+    HB_MEMORY_MODE_READONLY, nullptr, nullptr);
+  hb_face_t *face = hb_face_create (blob, 0);
+  hb_font_t *font = hb_font_create (face);
+
+  OT::hb_lazy_loader_t<OT::sbix::accelerator_t> sbix;
```
ebraminio: Doesn't need lazy loading obviously...

behdad:
Take a look at src/main.cc. We should rename it to dump or something, and for every table, dump interesting internal stuff.  Also, good idea to sanitize each table and print the sanitization result.  Also, that can become hb-sanitize.

ebraminio:
Does everything now look good (except its not careful memory use, which I am not sure if needed)? It now supports SVG/sbix/CBDT dumping but the CBDT parts looks weird, on the recent NotoColorEmoji almost every glyph is using [indexFormat](https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ot-color-cbdt-table.hh#L174) of 0 or 2 and only 15 one of them is format 1; and there is no version 0 [on the spec](https://docs.microsoft.com/en-us/typography/opentype/spec/eblc#indexsubtable1-variable-metrics-glyphs-with-4-byte-offsets)!

ebraminio:
CBDT's implementation is outdated and feels to be wrong, just run this on NotoColorEmoji.ttf

```diff
diff --git a/src/hb-ot-color-cbdt-table.hh b/src/hb-ot-color-cbdt-table.hh
index cf1c69c..1392d5b 100644
--- a/src/hb-ot-color-cbdt-table.hh
+++ b/src/hb-ot-color-cbdt-table.hh
@@ -115,6 +115,7 @@ struct IndexSubtableFormat1Or3
   inline bool sanitize (hb_sanitize_context_t *c, unsigned int glyph_count) const
   {
     TRACE_SANITIZE (this);
+    printf ("%d \n", (int) header.imageFormat);
     return_trace (c->check_struct (this) &&
                  c->check_array (offsetArrayZ, offsetArrayZ[0].static_size, glyph_count + 1));
   }
```

Only 18 ones are correct, which TTX tool is OK here, and the current implementation (before this PR) didn't have GlyphBitmapDataFormat18 even.

```
18
43662
16731
45019
18
43662
16731
45019
18
43662
16731
```

ebraminio:
Behdad what are essential for merging this? I tried to not touch the places used on production of `CBDT` and its dump doesn't work mostly. Things needed before or after this,

* Figure why CBDT dumping doesn't work
* Complete the WIP `glyf` work and turn this into dump-glyphs and also support CPAL/COLR rendering
* CFF/CFF2
* Implement hb_blob_from_file?
* hb-sanitize as you suggested

PR #921, opened 2018-03-26
[aat] Unprefix not essentially layout tables

### LINKED ISSUE #920: hb-aat-layout vs hb-aat-
Most tables are *not* layout per se. fmtx being an example. Should probably be renamed.

ebraminio:
Fixes #920


behdad:
Thanks.

PR #924, opened 2018-03-26
[subset] Couple of fixes for fuzzer discovered issues.

garretrieger:


src/hb-ot-hdmx-table.hh
```diff
@@ -196,9 +196,18 @@ struct hdmx
   inline bool sanitize (hb_sanitize_context_t *c) const
   {
     TRACE_SANITIZE (this);
-    return_trace (c->check_struct (this) && version == 0 &&
-		  !_hb_unsigned_int_mul_overflows (num_records, size_device_record) &&
-		  c->check_range (this, get_size()));
+    if (!c->check_struct (this) || version != 0 ||
```
behdad: Add `unlikely()` around the condition.

src/hb-ot-hdmx-table.hh
```diff
@@ -196,9 +196,18 @@ struct hdmx
...
+	_hb_unsigned_int_mul_overflows (num_records, size_device_record) ||
+	!c->check_range (this, get_size()))
+    {
+      return_trace (false);
+    }
+    for (unsigned int i = 0; i < this->num_records; i++)
```
behdad: Since it's a compact array, I suggest you just add one check_range. Indeed, that's what you were doing. I think what you were missing was a check that `size_device_record >= DeviceRecord::min_size`. No?

src/hb-subset-glyf.cc
```diff
@@ -214,7 +213,7 @@ _write_glyf_and_loca_prime (hb_subset_plan_t              *plan,
                                             loca_prime_size);
     _update_components (plan, glyf_prime_data_next, length);
 
-    glyf_prime_data_next += length;
+    glyf_prime_data_next += length + (length % 2); // Align to 2 bytes for short loca.
```
behdad: Do we always do this, or only if using short loca? It's fine if we always do it for now. Just add a TODO item to not doing for long loca, since that would be more compact.

garretrieger: Currently we always do it. I've added a TODO.

src/hb-ot-hdmx-table.hh
garretrieger: You're correct. I've removed the for loop and just added size_device_record >= DeviceRecord::min_size.

src/hb-ot-hdmx-table.hh
```diff
@@ -196,18 +196,10 @@ struct hdmx
...
-    for (unsigned int i = 0; i < this->num_records; i++)
-    {
-      if (!(*this)[i].sanitize (c, size_device_record))
-	return_trace (false);
-    }
-    return_trace (true);
+    return_trace (c->check_struct (this) && version == 0 &&
+		  !_hb_unsigned_int_mul_overflows (num_records, size_device_record) &&
+		  c->check_range (this, get_size()) &&
+		  size_device_record >= DeviceRecord::min_size);
```
behdad: Please move the size_device_record check before the check_range. Idea is that if that is wrong, we shouldn't even call get_size(); even though in this case doesn't matter.

behdad:
Humm. No idea why Travis is failing.

PR #926, opened 2018-03-27
Add hb_blob_create_from_file, a new API

ebraminio:
I guess there was some reason that this is not done earlier so no problem if rejected but I guess in case accepted, it will be a useful thing.

### DIFF
--- a/src/hb-blob.cc
+++ b/src/hb-blob.cc
@@ -43,6 +43,10 @@
 
 #include <stdio.h>
 #include <errno.h>
+#include <stdlib.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
 
 
 struct hb_blob_t {
@@ -273,7 +277,7 @@ hb_blob_destroy (hb_blob_t *blob)
  * @destroy: callback to call when @data is not needed anymore.
  * @replace: whether to replace an existing data with the same key.
  *
- * Return value: 
+ * Return value:
  *
  * Since: 0.9.2
  **/
@@ -292,9 +296,9 @@ hb_blob_set_user_data (hb_blob_t          *blob,
  * @blob: a blob.
  * @key: key for data to get.
  *
- * 
  *
- * Return value: (transfer none): 
+ *
+ * Return value: (transfer none):
  *
  * Since: 0.9.2
  **/
@@ -310,7 +314,7 @@ hb_blob_get_user_data (hb_blob_t          *blob,
  * hb_blob_make_immutable:
  * @blob: a blob.
  *
- * 
+ *
  *
  * Since: 0.9.2
  **/
@@ -327,7 +331,7 @@ hb_blob_make_immutable (hb_blob_t *blob)
  * hb_blob_is_immutable:
  * @blob: a blob.
  *
- * 
+ *
  *
  * Return value: TODO
  *
@@ -344,7 +348,7 @@ hb_blob_is_immutable (hb_blob_t *blob)
  * hb_blob_get_length:
  * @blob: a blob.
  *
- * 
+ *
  *
  * Return value: the length of blob data in bytes.
  *
@@ -361,9 +365,9 @@ hb_blob_get_length (hb_blob_t *blob)
  * @blob: a blob.
  * @length: (out):
  *
- * 
  *
- * Returns: (transfer none) (array length=length): 
+ *
+ * Returns: (transfer none) (array length=length):
  *
  * Since: 0.9.2
  **/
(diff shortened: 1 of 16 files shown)


behdad:
Thanks.

As you guessed, there's a reason this was not done earlier: we don't want to link libharfbuzz to glib.

So, you have to replace the g_mapped_file implementation by mmap and the Windows equivalent.  You can look at the glib implementation.

util/options.cc
```diff
@@ -631,74 +631,29 @@ font_options_t::get_font (void) const
...
+    while (!feof (stdin)) {
+      size_t ret = fread (buf, 1, sizeof (buf), stdin);
+      if (ferror (stdin))
+	fail (false, "Failed reading font from standard input: %s",
+	      strerror (errno));
+      g_string_append_len (gs, buf, ret);
     }
-
-    if (debug)
-      mm = HB_MEMORY_MODE_DUPLICATE;
```
ebraminio: The only difference is removal of this. If it is vital, I can revert my change on options.cc and options.hh.

ebraminio:
PTAL

util/options.cc
```diff
@@ -631,74 +631,29 @@ font_options_t::get_font (void) const
...
 
   /* Create the blob */
-  {
-    char *font_data;
-    unsigned int len = 0;
-    hb_destroy_func_t destroy;
-    void *user_data;
-    hb_memory_mode_t mm;
-
-    /* This is a hell of a lot of code for just reading a file! */
```
ebraminio: No longer it is :)

util/options.cc
ebraminio: Hmm, but I should see what should be done instead

behdad: No this is fine.

src/hb-blob.cc
```diff
@@ -502,3 +506,113 @@ _try_writable (hb_blob_t *blob)
...
+{
+  // Adopted from glib's gmappedfile.c with Matthias Clasen and
+  // Allison Lortie permission but changed to suit our need.
+  bool writable = false;
+  hb_memory_mode_t mm = HB_MEMORY_MODE_READONLY_MAY_MAKE_WRITABLE;
+
+  int fd = open (file_name, (writable ? O_RDWR : O_RDONLY) | _O_BINARY, 0);
+  if (fd == -1) return hb_blob_get_empty ();
+
+  hb_mapped_file_t *file = (hb_mapped_file_t *) malloc (sizeof (hb_mapped_file_t));
```
behdad: Use `calloc(..., 1)` to allocate objects.

src/hb-blob.cc
```diff
@@ -502,3 +506,113 @@ _try_writable (hb_blob_t *blob)
...
+  if (!file->contents) goto fail;
+  read (fd, file->contents, file->length);
+  mm = HB_MEMORY_MODE_WRITABLE;
+#endif
+
+  close (fd);
+  return hb_blob_create (file->contents, file->length, mm, (void *) file,
+			 (hb_destroy_func_t) _hb_mapped_file_destroy);
+
+fail:
```
behdad: I think if memory-mapping fails we should still try reading the file. That would make it work for stdin, etc, no? Or there might be other reasons that mmap fails?

behdad: In particular, accommodate for this piece of code in options.cc:
```

-      if (!len) {
-	/* GMappedFile is buggy, it doesn't fail if file isn't regular.
-	 * Try reading.
-	 * https://bugzilla.gnome.org/show_bug.cgi?id=659212 */
```

As long as running on a font file of `/dev/null` or `/dev/stdin` works, I'm fine. Thanks.

src/hb-blob.cc
ebraminio: Obviously, done.

src/hb-blob.cc
ebraminio: > In particular, accommodate for this piece of code in options.cc:

I've picked the patch related to that comment https://bug659212.bugzilla-attachments.gnome.org/attachment.cgi?id=196855 

> As long as running on a font file of /dev/null or /dev/stdin works, I'm fine. Thanks.

They do, AFAICS

util/options.cc
```diff
@@ -631,74 +628,29 @@ font_options_t::get_font (void) const
...
-	  user_data = (void *) font_data;
-	  mm = HB_MEMORY_MODE_WRITABLE;
-	} else {
-	  fail (false, "%s", error->message);
-	  //g_error_free (error);
-	}
-      }
+    while (!feof (stdin)) {
+      size_t ret = fread (buf, 1, sizeof (buf), stdin);
+      if (ferror (stdin))
```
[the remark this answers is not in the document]
ebraminio: I've kept "stdin" reading untouched and as it workarounds Windows input, lean no do to. To be honest however I see a difference,

image

And that shouldn't related to my changes as the untouched ones also do,

image

ebraminio:
PTAL

src/hb-blob.cc
ebraminio: Tested on Linux also.

util/options.cc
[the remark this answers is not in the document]
behdad: Weird. We should figure out what's going on..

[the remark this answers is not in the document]
behdad: I can reproduce. It's the '-' case that is not working. We are getting cairo's internal font. Ouch.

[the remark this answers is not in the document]
ebraminio: So maybe merging this first and then looking into that? As said I've kept stdin reading of hb-{view,shape} untouched and if is broken I believe it shouldn't be here.

behdad:
Agreed.

src/hb-blob.cc
```diff
@@ -502,3 +506,114 @@ _try_writable (hb_blob_t *blob)
...
+  int fd = open (file_name, (writable ? O_RDWR : O_RDONLY) | _O_BINARY, 0);
+  if (unlikely (fd == -1)) return hb_blob_get_empty ();
+
+  hb_mapped_file_t *file = (hb_mapped_file_t *) calloc (sizeof (hb_mapped_file_t), 1);
+
+  struct stat st;
+  if (unlikely (fstat (fd, &st) == -1)) goto fail;
+
+  // If the file size is 0 and isn't regular, give up
+  if (unlikely (st.st_size == 0 && S_ISREG (st.st_mode))) goto fail;
```
behdad: The comment and the code don't agree?

src/hb-blob.cc
behdad: Thanks.

src/hb-blob.cc
ebraminio: You are right.

ebraminio: https://github.com/harfbuzz/harfbuzz/commit/5ad87a93fefcb94ee5f28de8d75903b2550c8d94

PR #927, opened 2018-03-27
[WIP] COLR/CPAL renderer

ebraminio:
Not in final state, I probably want to do output SVG instead render PNG and remove temporary blending code.


src/dump-emoji.cc
```diff
@@ -117,12 +131,96 @@ int main(int argc, char **argv)
...
+
+  OT::Sanitizer<OT::CPAL> sanitizerCPAL;
+  hb_blob_t* cpal_blob = sanitizerCPAL.sanitize (face->reference_table (HB_OT_TAG_CPAL));
+  const OT::CPAL *cpal = OT::Sanitizer<OT::CPAL>::lock_instance (cpal_blob);
+
+  FT_Library library;
+  FT_Init_FreeType(&library);
+  FT_Face ftface;
+  FT_New_Face(library, argv[1], 0, &ftface);
+  cairo_font_face_t *cairo_face = cairo_ft_font_face_create_for_ft_face (ftface, 0);
```
ebraminio: freetype is not used directly on this script but just for creating Cairo face, I hope it has some better way to load from the blob or the original path?

src/dump-emoji.cc
```diff
@@ -133,6 +272,27 @@ int main(int argc, char **argv)
...
+  const OT::CPAL *cpal = OT::Sanitizer<OT::CPAL>::lock_instance (cpal_blob);
+
+  cairo_font_face_t *cairo_face;
+  {
+    FT_Library library;
+    FT_Init_FreeType(&library);
+    FT_Face ftface;
+    FT_New_Face(library, argv[1], 0, &ftface);
+    cairo_face = cairo_ft_font_face_create_for_ft_face (ftface, 0);
+  }
```
ebraminio: freetype is not used directly on this script but just for creating Cairo face, does it have some better way to load from the blob or the original path?

PR #930, opened 2018-03-28
Make atexit callbacks threadsafe

### LINKED ISSUE #923: atexit callbacks need to be threadsafe
Apparently POSIX doesn't specify whether atexit callbacks are called before or after threads are terminated during process shutdown.  Following suggests that in fact, they are called while threads still running!

  https://stackoverflow.com/questions/39655868/what-does-the-posix-standard-say-about-thread-stacks-in-atexit-handlers-what

We should make our atexit callbacks threadsafe, simply by using cmpexch to clean things up.

ebraminio:
Fixes #923 hopefully, wanted to go for some macro but felt better to keep it simple, let me know if is wrong or anything.

### DIFF
--- a/src/hb-common.cc
+++ b/src/hb-common.cc
@@ -60,12 +60,12 @@ _hb_options_init (void)
 
 /**
  * hb_tag_from_string:
- * @str: (array length=len) (element-type uint8_t): 
- * @len: 
+ * @str: (array length=len) (element-type uint8_t):
+ * @len:
  *
- * 
  *
- * Return value: 
+ *
+ * Return value:
  *
  * Since: 0.9.2
  **/
@@ -90,10 +90,10 @@ hb_tag_from_string (const char *str, int len)
 
 /**
  * hb_tag_to_string:
- * @tag: 
- * @buf: (out caller-allocates) (array fixed-size=4) (element-type uint8_t): 
+ * @tag:
+ * @buf: (out caller-allocates) (array fixed-size=4) (element-type uint8_t):
  *
- * 
+ *
  *
  * Since: 0.9.5
  **/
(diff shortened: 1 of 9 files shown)


src/hb-common.cc
```diff
@@ -60,12 +60,12 @@ _hb_options_init (void)
 
 /**
  * hb_tag_from_string:
- * @str: (array length=len) (element-type uint8_t): 
- * @len:
```
behdad: Why all the whitespace noise? Or do you have an editor that automatically does these?

behdad:
No this is not how you do it.  It's not that multiple copies of the atexit can run. It's that it can race with other code using these. Let me fix one and you see what's needed. No new variables are needed.

behdad:
See this: https://github.com/harfbuzz/harfbuzz/commit/5aa2c6e194bc75637d3f4862a096c0e7100072e2

src/hb-common.cc
ebraminio: Yes, that is the case with the our current `.editorconfig` which requests it.

src/hb-shape.cc
```diff
@@ -51,7 +51,12 @@ static const char **static_shaper_list;
 static
 void free_static_shaper_list (void)
 {
-  free (static_shaper_list);
+retry:
+  char **shaper_list = (char **) hb_atomic_ptr_get (&static_shaper_list);
+  if (!hb_atomic_ptr_cmpexch (static_shaper_list, *shaper_list, nullptr))
+    goto retry;
```
ebraminio: Hmm, this one doesn't seem right.

ebraminio:
PTAL

behdad:
Thanks. Looks good.

It occurred to me that this is rather futile. If we free these while another thread might use them, there's an unfixable race condition. I'm happy pushing this in though.

ebraminio:
> It occurred to me that this is rather futile. If we free these while another thread might use them, there's an unfixable race condition.

I wanted to ask about the benefit on that situation also.

PR #931, opened 2018-03-28
[graphite] Make get_table threadsafe

### LINKED ISSUE #922: [graphite] get_table is not threadsafe
Fix this comment:
```
    /* TODO Not thread-safe, but fairly harmless.
     * We can do the double-checked pointer cmpexch thing here. */
```
It's prepending to a prepend-only singly-linked list, so should use pointer cmpexch like we do in other places.

ebraminio:
Fixes #922 hopefully

### DIFF
--- a/src/hb-graphite2.cc
+++ b/src/hb-graphite2.cc
@@ -79,10 +79,12 @@ static const void *hb_graphite2_get_table (const void *data, unsigned int tag, s
     p->blob = blob;
     p->tag = tag;
 
-    /* TODO Not thread-safe, but fairly harmless.
-     * We can do the double-checked pointer cmpexch thing here. */
-    p->next = face_data->tlist;
-    face_data->tlist = p;
+retry:
+    hb_graphite2_tablelist_t *tlist = (hb_graphite2_tablelist_t *) hb_atomic_ptr_get (&face_data->tlist);
+    p->next = tlist;
+
+    if (!hb_atomic_ptr_cmpexch (&face_data->tlist, tlist, p))
+      goto retry;
   }
 
   unsigned int tlen;
--- a/src/hb-graphite2.h
+++ b/src/hb-graphite2.h
@@ -1,6 +1,6 @@
 /*
- * Copyright (C) 2011  Martin Hosken
- * Copyright (C) 2011  SIL International
+ * Copyright © 2011  Martin Hosken
+ * Copyright © 2011  SIL International
  *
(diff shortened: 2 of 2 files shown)


src/hb-graphite2.cc
```diff
@@ -79,10 +79,9 @@ static const void *hb_graphite2_get_table (const void *data, unsigned int tag, s
     p->blob = blob;
     p->tag = tag;
 
-    /* TODO Not thread-safe, but fairly harmless.
-     * We can do the double-checked pointer cmpexch thing here. */
-    p->next = face_data->tlist;
-    face_data->tlist = p;
+    p->next = (hb_graphite2_tablelist_t *) hb_atomic_ptr_get (&face_data->tlist);
+    if (!hb_atomic_ptr_cmpexch (&face_data->tlist, nullptr, p))
+      free (p);
```
behdad: We should "goto retry". See other examples of cmpexch use.

ebraminio:
Now I guess I got it right, have a look please

src/hb-graphite2.cc
```diff
@@ -79,10 +80,13 @@ static const void *hb_graphite2_get_table (const void *data, unsigned int tag, s
     p->blob = blob;
     p->tag = tag;
 
-    /* TODO Not thread-safe, but fairly harmless.
-     * We can do the double-checked pointer cmpexch thing here. */
-    p->next = face_data->tlist;
-    face_data->tlist = p;
+    hb_graphite2_tablelist_t *tlist = (hb_graphite2_tablelist_t *) hb_atomic_ptr_get (&face_data->tlist);
```
ebraminio: Or should this be `hb_graphite2_tablelist_t **tlist = (hb_graphite2_tablelist_t **) hb_atomic_ptr_get (&face_data->tlist);`?

behdad:
Any idea why coveralls is being flaky?

ebraminio:
It is saying the newly added code is not covered by the test runs and that's why it is giving a minus score and symbolically failing and I guess we should get to use it :)

behdad:
Can we configure it to not fail if the coverage decrease is small? Or not fail at all?

ebraminio:
Asked on https://github.com/exercism/exercism.io/issues/3139

ebraminio:
Oh its setting is here, https://coveralls.io/github/harfbuzz/harfbuzz/settings and "COVERAGE DECREASE THRESHOLD FOR FAILURE" to 2 percent and disabled its leaving comment option.

PR #932, opened 2018-03-28
Make gen-use-table.py py3 compatible

ebraminio:


src/gen-use-table.py
```diff
@@ -352,44 +353,44 @@ def map_to_use(data):
...
 		if u % 16 == 0:
-			print
-			print "  /* %04X */" % u,
+			print ()
+			print ("  /* %04X */" % u, end='')
 		if u in data:
 			num += 1
 		d = data.get (u, defaults)
-		sys.stdout.write ("%6s," % d[0])
+		print ("%6s," % d[0], end='')
```
ebraminio: Only part that needed for more change was this one `print_block`, I modified it (using end='') to it the result unchanged.

behdad:
If you convert, then add this to the beginning of the file:
```
from __future__ import print_function, division, absolute_import
```
Also, please make all similar changes to other gen- scripts por favor!

PR #933, opened 2018-03-28
[WIP] Route Tibetan through USE

### LINKED ISSUE #861: Tibetan reordering does not match Uniscribe
U+0F3F TIBETAN SIGN MAR TSHES is a spacing mark that goes to the left of digits. Uniscribe reorders it before digits but HarfBuzz does not.

ebraminio:
Hopefully to fix #861


src/gen-use-table.py
```diff
@@ -338,6 +338,9 @@ def map_to_use(data):
 		if 0x1CF2 <= U <= 0x1CF3: UIPC = Right
 		if 0x1CF8 <= U <= 0x1CF9: UIPC = Top
 
+		if not (UIPC in [Not_Applicable, Visual_Order_Left] or USE in use_positions):
+			print ("%s %s %s %s %s" % (hex(U), UIPC, USE, UISC, UGC), file=sys.stderr)
+			continue
```
ebraminio: I should figure out what is going on to know why this is failing, which of course it has some reason.

dscorbett:
This breaks clusters with multiple vowels like ⟨ཀིུ⟩: it needs the USE’s Tibetan overrides.

src/gen-use-table.py
dscorbett:
```
0xf18 Bottom O Other Mn
0xf19 Bottom O Other Mn
0xf3e Right O Other Mc
0xf3f Left O Other Mc
0xf86 Top O Other Mn
0xf87 Top O Other Mn
```
It’s failing because Unicode assigns each of those characters an Indic_Positional_Category but not an Indic_Syllabic_Category.

ebraminio: And any hint what should be here also? :)

dscorbett: Set `UISC` for these characters after line 297.

src/gen-use-table.py
```diff
@@ -293,6 +293,9 @@ def map_to_use(data):
 
 		# Resolve Indic_Syllabic_Category
 
+		# TODO: There are Tibetan signs that don't have UISC
+		if U in [0xF18, 0xF19, 0xF3E, 0xF3F, 0xF86, 0xf87]: UISC = Cantillation_Mark
```
behdad: Always minimum four digits for Unicode hex values. Also, last one has lowercase 'f' while others are uppercase.

src/gen-use-table.py
```diff
@@ -338,9 +341,6 @@ def map_to_use(data):
 		if 0x1CF2 <= U <= 0x1CF3: UIPC = Right
 		if 0x1CF8 <= U <= 0x1CF9: UIPC = Top
 
-		if not (UIPC in [Not_Applicable, Visual_Order_Left] or USE in use_positions):
```
behdad: Why this change?

ebraminio: Oh that was just added, I probably should squash the changes.

src/gen-use-table.py
ebraminio: Sure, can you help on finding correct syllable category for those also?

src/hb-ot-shape-complex-private.hh
```diff
@@ -233,7 +232,7 @@ hb_ot_shape_complex_categorize (const hb_ot_shape_planner_t *planner)
     /* Unicode-2.0 additions */
     case HB_SCRIPT_TIBETAN:
```
behdad: This case should be removed and other commented-out TIBETAN uncommented.

src/gen-use-table.py
behdad: Yeah, looked at the squashed diff and looks good.

src/gen-use-table.py
behdad: Let's ask @jfkthame @PeterCon 
Typically those need to be discussed on an issue by themselves.

behdad:
Also, did you test the reported bug at all?

src/gen-use-table.py
behdad: @dscorbett

behdad: @PeterCon Peter, can you double check the overrides for Tibetan against the list posted on the spec page? Better yet, can you refresh the override list for all scripts?

ebraminio:
> Also, did you test the reported bug at all?

Yep, and this by its own wasn't fix it IIRC. I thought this as just a starting point...

ebraminio:
No response to what to do with the signs and I guess I can't reach to this soon so probably better to go lights off for this.

behdad:
It's still a valid issue and we want to fix it. Please don't close valid issue and PRs just because of inactivity.

ebraminio:
Well, I though I can't reach to them soon and better to focus on some limited ones, but if you want to, sure :)

src/hb-ot-shape-complex-use-table.cc
```diff
@@ -194,7 +194,24 @@ static const USE_TABLE_ELEMENT_TYPE use_table[] = {
   /* 0DE0 */     O,     O,     O,     O,     O,     O,     B,     B,     B,     B,     B,     B,     B,     B,     B,     B,
   /* 0DF0 */     O,     O,  VPst,  VPst,     O,     O,     O,     O,
 
-#define use_offset_0x1000u 1360
+#define use_offset_0x0f18u 1360
+
+
+  /* Tibetan */
+                                                                     VMBlw, VMBlw,     O,     O,     O,     O,     O,     O,
```
eroux: @ebraminio where do these values come from? I can't find them in the MS spec

ebraminio: This is the result of applying of the change above on "gen-use-table.py" and re-running it. According to Behdad every one of them should be decided to what category they fall into and are hard to decide.

eroux: I see, thanks a lot!

PR #940, opened 2018-03-28
Make more gen-* scripts py3 compatible

ebraminio:
Don't merge yet however, should be tested more before the merge.


behdad:
lgtm. Thanks.

Did you get to verify that their outputs didn't change?

ebraminio:
Yes, that why I was saying so. Arabic script update seems to have not any change but have a look again at `src/Makefile.am` and `src/hb-ot-shape-complex-indic-table.cc`

dscorbett:
Some of these scripts use `file` instead of `open` and therefore do not work in Python 3.

ebraminio:
Ops! #941

PR #941, opened 2018-03-29
Actual py3 compatibility making on gen-* scripts

ebraminio:
This is tested on both py2 and py3. There is no difference between the previous version and the current on python2 but on python3 currently there is just a minor difference. This time I changed shebangs for my python3 testing. The generated files on this commit are by python2 by running this command,

`make -Cbuild/src arabic-table indic-table use-table && mv build/src/*.{cc,hh} src/`

for my out-of-tree build.

### DIFF
--- a/src/gen-arabic-table.py
+++ b/src/gen-arabic-table.py
@@ -9,7 +9,7 @@ if len (sys.argv) != 4:
 	print ("usage: ./gen-arabic-table.py ArabicShaping.txt UnicodeData.txt Blocks.txt", file=sys.stderr)
 	sys.exit (1)
 
-files = [file (x) for x in sys.argv[1:]]
+files = [open (x) for x in sys.argv[1:]]
 
 headers = [[files[0].readline (), files[0].readline ()], [files[2].readline (), files[2].readline ()]]
 headers.append (["UnicodeData.txt does not have a header."])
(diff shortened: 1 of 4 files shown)


ebraminio:
Python 3 generated files actually look a little better also, if you like I can put those instead, or maybe not.

dscorbett:
These scripts fail in Python 3 when `LC_ALL=C`.
```
$ python3 ./gen-use-table.py ~/.unicode/IndicSyllabicCategory.txt ~/.unicode/IndicPositionalCategory.txt ~/.unicode/UnicodeData.txt ~/.unicode/Blocks.txt >hb-ot-shape-complex-use-table.cc
  File "./gen-use-table.py", line 15, in <module>
    headers = [[f.readline () for i in range (2)] for j,f in enumerate(files) if j != 2]
  File "./gen-use-table.py", line 15, in <listcomp>
    headers = [[f.readline () for i in range (2)] for j,f in enumerate(files) if j != 2]
  File "./gen-use-table.py", line 15, in <listcomp>
    headers = [[f.readline () for i in range (2)] for j,f in enumerate(files) if j != 2]
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 79: ordinal not in range(128)
```

ebraminio:
Interesting! Have a look #942 and let me know if anything is needed.

Issue #919, opened 2018-03-26
Implement Zapf table parsing


ebraminio:
Now that we have an issue for every other AAT table, lets have for this one also,

> The 'Zapf' table—named with permission after legendary type designer Hermann Zapf—contains information about the individual glyphs in the font.  Conceptually, this information consists of:
>
> A text string which the glyph represents
> A set of identifiers
> A set of font features
> A set of glyph collections

https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6Zapf.html

ebraminio:
List of Apple fonts on macOS 10.13.3 with `Zapf` table: (on their first face)

```
/Library/Fonts/AlBayan.ttc
/Library/Fonts/Zapfino.ttf
/Library/Fonts/Silom.ttf
/Library/Fonts/Nadeem.ttc
/Library/Fonts/KufiStandardGK.ttc
/Library/Fonts/Papyrus.ttc
/Library/Fonts/Farisi.ttf
/Library/Fonts/Mishafi.ttf
/Library/Fonts/BigCaslon.ttf
/Library/Fonts/Al Nile.ttc
/Library/Fonts/Baghdad.ttc
/Library/Fonts/AmericanTypewriter.ttc
/Library/Fonts/Muna.ttc
/Library/Fonts/Apple Chancery.ttf
/Library/Fonts/Farah.ttc
/Library/Fonts/Diwan Kufi.ttc
/Library/Fonts/Hoefler Text Ornaments.ttf
/Library/Fonts/Beirut.ttc
/Library/Fonts/Diwan Thuluth.ttf
/Library/Fonts/Ayuthaya.ttf
/Library/Fonts/Sana.ttc
/Library/Fonts/Hoefler Text.ttc
/Library/Fonts/Sathu.ttf
/Library/Fonts/Krungthep.ttf
/Library/Fonts/Al Tarikh.ttc
/Library/Fonts/Damascus.ttc
/Library/Fonts/Waseem.ttc
/Library/Fonts/DecoTypeNaskh.ttc
/System/Library/Fonts/PingFang.ttc
/System/Library/Fonts/LucidaGrande.ttc
/System/Library/Fonts/Thonburi.ttc
/System/Library/Fonts/Palatino.ttc
/System/Library/Fonts/Apple Symbols.ttf
/System/Library/Fonts/GeezaPro.ttc
/System/Library/Fonts/Helvetica.ttc
```

[...]

Issue #907, opened 2018-03-25
Implement lcar table parsing


KrasnayaPloshchad:
This table is used to indicate ligature carets in AAT fonts.
See: https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6lcar.html

ebraminio:
(just a personal note)

List of Apple fonts on macOS 10.13.3 with `lcar` table:

```
/Library/Fonts/Apple Chancery.ttf
/Library/Fonts/Hoefler Text.ttc
```

ebraminio:
@behdad Only two fonts have it, maybe it is a sign that this is being replaced with GDEF's or the reverse way? The number is not changed between Sierra and High Sierra.

ebraminio:
Needed code for this,

```C
struct lcar
{
  static const hb_tag_t tableTag = HB_AAT_TAG_lcar;

  inline bool sanitize (hb_sanitize_context_t *c) const
  {
    TRACE_SANITIZE (this);
    return_trace (c->check_struct (this) && lookup.sanitize (c));
  }

  protected:
  FixedVersion<>version;	/* Version number of the ligature caret table */
  HBUINT16	format;		/* Format of the ligature caret table. */
  Lookup<OffsetTo<ArrayOf<HBINT16> > >
		lookup;		/* data Lookup table associating glyphs */
  public:
  DEFINE_SIZE_MIN (8);
};
```

But `Lookup<OffsetTo<ArrayOf<HBINT16> > >` won't work so it needs runtime check and in order to make that happen and provide it seamlessly with OT's API this needs control points data from `glyf` table (format 1) which isn't there also and too much work just for supporting just two fonts and for an API (`hb_ot_layout_get_ligature_carets`) not used on Firefox and Chromium even.

ebraminio:
A possible fix is https://github.com/harfbuzz/harfbuzz/pull/1358

behdad:
> But `Lookup<OffsetTo<ArrayOf<HBINT16> > >` won't work as is and in order to make that happen and provide it seamlessly with OT's API this needs control points data from `glyf` table (format 1) which isn't there also and too much work just for supporting just two fonts and for an API (`hb_ot_layout_get_ligature_carets`) not used on Firefox and Chromium even.

I don't understand what you're saying.

behdad:
Why did you close?  Is https://github.com/harfbuzz/harfbuzz/pull/1358 still relevant?

[...]

PR #946, opened 2018-03-31
[aat] Implement lcar and Zapf (incomplete) tables parsing

ebraminio:
Created to be closed and archived. Related to #907 #919 but is incomplete.

"src/hb-aat-layout-common-private.hh" changes was needed for both lcar and Zapf but didn't apply the changes `LookupFormat4` because it was different.

### DIFF
--- a/src/Makefile.sources
+++ b/src/Makefile.sources
@@ -88,7 +88,9 @@ HB_OT_sources = \
 	hb-aat-layout-trak-table.hh \
 	hb-aat-fmtx-table.hh \
 	hb-aat-gcid-table.hh \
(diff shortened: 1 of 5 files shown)


behdad:
lcar is actually relevant to shaping, and is equivalent to data from GDEF. But we can add it later when we have API for it. Thanks.

ebraminio:
I liked to add `lcar` but only two fonts are having it and the number is not changed between Sierra and High Sierra :/ https://github.com/harfbuzz/harfbuzz/issues/907#issuecomment-377624222

behdad:
Right. So, not for now.

PR #947, opened 2018-03-31
[aat] Implement feat table parsing

### LINKED ISSUE #904: Implement feat table parsing
This table is used to attribute font features for AAT fonts, and intimately connected with the `morx` table, this table has own font feature resistry. 
See: https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6feat.html

ebraminio:
Fixes #904, tested manually with Skia.ttf

### DIFF
--- a/src/hb-aat-feat-table.hh
+++ b/src/hb-aat-feat-table.hh
@@ -1,0 +1,122 @@
+/*
+ * Copyright © 2018  Ebrahim Byagowi
+ *
+ *  This is part of HarfBuzz, a text shaping library.
+ *
+ * Permission is hereby granted, without written agreement and without
+ * license or royalty fees, to use, copy, modify, and distribute this
+ * software and its documentation for any purpose, provided that the
+ * above copyright notice and the following two paragraphs appear in
+ * all copies of this software.
+ *
+ * IN NO EVENT SHALL THE COPYRIGHT HOLDER BE LIABLE TO ANY PARTY FOR
+ * DIRECT, INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES
+ * ARISING OUT OF THE USE OF THIS SOFTWARE AND ITS DOCUMENTATION, EVEN
+ * IF THE COPYRIGHT HOLDER HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ *
+ * THE COPYRIGHT HOLDER SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ * FITNESS FOR A PARTICULAR PURPOSE.  THE SOFTWARE PROVIDED HEREUNDER IS
+ * ON AN "AS IS" BASIS, AND THE COPYRIGHT HOLDER HAS NO OBLIGATION TO
+ * PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
+ */
+
+#ifndef HB_AAT_FEAT_TABLE_HH
+#define HB_AAT_FEAT_TABLE_HH
+
+#include "hb-aat-layout-common-private.hh"
+
+#define HB_AAT_TAG_feat HB_TAG('f','e','a','t')
+
+
+namespace AAT {
+
+struct FeatureName {
+
+  inline bool sanitize (hb_sanitize_context_t *c, const void *base) const
+  {
+    TRACE_SANITIZE (this);
+    return_trace (c->check_struct (this) &&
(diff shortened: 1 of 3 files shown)


behdad:
Please rename to `hb-aat-layout-feat-table.hh`.

The sanitize here looks incomplete. I don't want any more incomplete sanitize in the code, so, when not sure, please do not merge.  There's some formatting issues as well. When we break line mid-expression, the indentation should align logically with the other operators. Also, we don't check for things like nameID > 255 etc in sanitize as those are irrelevant to the parsing of tables. Sanitize only checks that it's safe to access memory.

src/hb-aat-feat-table.hh
```diff
@@ -0,0 +1,122 @@
...
+
+#include "hb-aat-layout-common-private.hh"
+
+#define HB_AAT_TAG_feat HB_TAG('f','e','a','t')
+
+
+namespace AAT {
+
+struct FeatureName {
+
```
behdad: Opening `{` goes on new line.

src/hb-aat-feat-table.hh
```diff
@@ -0,0 +1,122 @@
...
+  inline bool sanitize (hb_sanitize_context_t *c, const void *base) const
+  {
+    TRACE_SANITIZE (this);
+    return_trace (c->check_struct (this) &&
+      (base+settingTable).sanitize (c, nSettings, base));
```
behdad: Indentation should align to `c->...`.

src/hb-aat-feat-table.hh
```diff
@@ -0,0 +1,122 @@
...
+				 * as the default. */
+  };
+
+  struct SettingName
+  {
+    inline bool sanitize (hb_sanitize_context_t *c, const void *base HB_UNUSED) const
+    {
+      TRACE_SANITIZE (this);
+      return_trace (c->check_struct (this) &&
+	nameIndex > 255 && nameIndex < 32768);
```
behdad: Remove the range check. Also indentation was wrong.

src/hb-aat-feat-table.hh
```diff
@@ -0,0 +1,122 @@
...
+				 * which is to be taken as the default. */
+    IndexMask = 0x00FF		/* If bits 30 and 31 are set, then these sixteen bits
+				 * indicate the index of the setting in the setting name
+				 * array for this feature which should be taken
+				 * as the default. */
+  };
+
+  struct SettingName
+  {
+    inline bool sanitize (hb_sanitize_context_t *c, const void *base HB_UNUSED) const
```
behdad: What's the unused base for?

src/hb-aat-feat-table.hh
```diff
@@ -0,0 +1,122 @@
...
+				 * (if one is required). If set, then bits 0-15 of this
+				 * featureFlags field contain the index of the setting
+				 * which is to be taken as the default. */
+    IndexMask = 0x00FF		/* If bits 30 and 31 are set, then these sixteen bits
+				 * indicate the index of the setting in the setting name
+				 * array for this feature which should be taken
+				 * as the default. */
+  };
+
+  struct SettingName
```
behdad: I don't think I like this struct being inside the other. We don't do that in other places except for accelerators, iterators, etc.

src/hb-aat-feat-table.hh
```diff
@@ -0,0 +1,122 @@
...
+				 * The nameIndex must be greater than 255 and less than 32768. */
+    public:
+    DEFINE_SIZE_STATIC (4);
+  };
+
+  protected:
+  HBUINT16	feature;	/* Feature type. */
+  HBUINT16	nSettings;	/* The number of records in the setting name array. */
+  LOffsetTo<UnsizedArrayOf<SettingName> >
+	settingTable;		/* Offset in bytes from the beginning of this table to
```
behdad: Alignment.

src/hb-aat-feat-table.hh
ebraminio: Sure

src/hb-aat-feat-table.hh
ebraminio: Oh sure.

src/hb-aat-feat-table.hh
ebraminio: Oh ok. Just wanted to avoid name collision but it we will deal with whenever happens so why not

src/hb-aat-feat-table.hh
ebraminio: So we can do the sanitization.

src/hb-aat-feat-table.hh
ebraminio: It is stated by the spec, "the nameIndex must be greater than 255 and less than 32768." but sure why not.

src/hb-aat-feat-table.hh
ebraminio: Sure.

PR #948, opened 2018-03-31
[aat] Further improvements on feat

ebraminio:
Please let me know if any thing else is needed. AFAIK my codes were sanitized carefully at least when they were committed (except the first ones that I tried to fix later)

### DIFF
--- a/src/hb-aat-layout-feat-table.hh
+++ b/src/hb-aat-layout-feat-table.hh
@@ -22,8 +22,8 @@
  * PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
  */
 
-#ifndef HB_AAT_FEAT_TABLE_HH
-#define HB_AAT_FEAT_TABLE_HH
+#ifndef HB_AAT_LAYOUT_FEAT_TABLE_HH
+#define HB_AAT_LAYOUT_FEAT_TABLE_HH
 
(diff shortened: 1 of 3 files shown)


src/hb-aat-feat-table.hh
```diff
@@ -32,13 +32,28 @@
...
+};
+
+struct FeatureName
+{
   inline bool sanitize (hb_sanitize_context_t *c, const void *base) const
   {
     TRACE_SANITIZE (this);
     return_trace (c->check_struct (this) &&
-      (base+settingTable).sanitize (c, nSettings, base));
+		  (base+settingTable).sanitize (c, nSettings));
```
ebraminio: This is changed from the original code. Previous code was passing just to make sure the inner sanitization.

src/hb-aat-layout-feat-table.hh
```diff
@@ -32,13 +32,28 @@
...
+{
+  inline bool sanitize (hb_sanitize_context_t *c) const
+  {
+    TRACE_SANITIZE (this);
+    return_trace (c->check_struct (this));
+  }
 
+  protected:
+  HBUINT16	setting;	/* The setting. */
+  NameID	nameIndex;	/* The name table index for the setting's name. */
```
ebraminio: The spec says `nameIndex` is an `int16` but ours is unsigned. It doesn't matter eventually anyway but just saying.

PR #950, opened 2018-04-01
Add hb_sanitize API and use it on a new tool, hb-sanitize

ebraminio:
Have a look. Maybe it should be called validator, as Apple's ftxvalidator and [the Microsoft one](https://github.com/Microsoft/Font-Validator)?

### DIFF
--- a/TODO
+++ b/TODO
@@ -19,8 +19,6 @@ API additions
 
 - Add hb-cairo glue
 
-- Add sanitize API (and a cached version, that saves result on blob user-data)
-
 - BCP 47 language handling / API (language_matches?)
 
 - Add hb_font_create_unscaled()?
--- a/src/hb-sanitize.cc
+++ b/src/hb-sanitize.cc
@@ -1,0 +1,132 @@
+/*
+ * Copyright © 2018  Ebrahim Byagowi
+ *
+ *  This is part of HarfBuzz, a text shaping library.
+ *
+ * Permission is hereby granted, without written agreement and without
+ * license or royalty fees, to use, copy, modify, and distribute this
+ * software and its documentation for any purpose, provided that the
+ * above copyright notice and the following two paragraphs appear in
+ * all copies of this software.
+ *
+ * IN NO EVENT SHALL THE COPYRIGHT HOLDER BE LIABLE TO ANY PARTY FOR
+ * DIRECT, INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES
+ * ARISING OUT OF THE USE OF THIS SOFTWARE AND ITS DOCUMENTATION, EVEN
+ * IF THE COPYRIGHT HOLDER HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ *
+ * THE COPYRIGHT HOLDER SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ * FITNESS FOR A PARTICULAR PURPOSE.  THE SOFTWARE PROVIDED HEREUNDER IS
+ * ON AN "AS IS" BASIS, AND THE COPYRIGHT HOLDER HAS NO OBLIGATION TO
+ * PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
+ */
+
+#include "hb-private.hh"
+
+#include "hb-aat-layout-ankr-table.hh"
+#include "hb-aat-layout-bsln-table.hh"
+#include "hb-aat-layout-feat-table.hh"
+#include "hb-aat-layout-kerx-table.hh"
+#include "hb-aat-layout-morx-table.hh"
+#include "hb-aat-layout-trak-table.hh"
+#include "hb-aat-fmtx-table.hh"
+#include "hb-aat-gcid-table.hh"
+#include "hb-aat-ltag-table.hh"
+#include "hb-ot-layout-base-table.hh"
+#include "hb-ot-layout-gdef-table.hh"
+#include "hb-ot-layout-gpos-table.hh"
+#include "hb-ot-layout-gsub-table.hh"
+#include "hb-ot-layout-jstf-table.hh"
+#include "hb-ot-color-cbdt-table.hh"
+#include "hb-ot-color-colr-table.hh"
+#include "hb-ot-color-cpal-table.hh"
+#include "hb-ot-color-sbix-table.hh"
+#include "hb-ot-color-svg-table.hh"
+#include "hb-ot-math-table.hh"
+#include "hb-ot-var-avar-table.hh"
+#include "hb-ot-var-fvar-table.hh"
+#include "hb-ot-var-hvar-table.hh"
+#include "hb-ot-var-mvar-table.hh"
+#include "hb-ot-cmap-table.hh"
+#include "hb-ot-glyf-table.hh"
+#include "hb-ot-hdmx-table.hh"
+#include "hb-ot-head-table.hh"
+#include "hb-ot-hhea-table.hh"
+#include "hb-ot-hmtx-table.hh"
+#include "hb-ot-kern-table.hh"
+#include "hb-ot-maxp-table.hh"
+#include "hb-ot-name-table.hh"
+#include "hb-ot-os2-table.hh"
+#include "hb-ot-post-table.hh"
+
+
+/**
+ * hb_sanitize: Sanitizes a table specified by a tag
+ * @face:
+ * @tag:
+ *
+ *
+ *
+ * Return value: Sanitizer result, nullptr if we don't support that table at all
+ * Since: REPLACEME
+ **/
+hb_blob_t *
+hb_sanitize (hb_face_t *face, hb_tag_t tag)
+{
+  unsigned int num_glyphs = hb_face_get_glyph_count (face);
+  hb_blob_t *blob = hb_face_reference_table (face, tag);
+  hb_blob_t *result = nullptr;
+#define DEFINE_SANITIZER(table) \
+  case table::tableTag: { \
+    OT::Sanitizer<table> sanitizer; \
+    sanitizer.set_num_glyphs (num_glyphs); \
+    result = sanitizer.sanitize (blob); \
+    break; \
+  }
+  switch (tag)
+  {
+  DEFINE_SANITIZER(AAT::ankr);
+  DEFINE_SANITIZER(AAT::bsln);
+  DEFINE_SANITIZER(AAT::feat);
+  DEFINE_SANITIZER(AAT::kerx);
+  DEFINE_SANITIZER(AAT::morx);
+  DEFINE_SANITIZER(AAT::trak);
+  DEFINE_SANITIZER(AAT::fmtx);
+  DEFINE_SANITIZER(AAT::gcid);
+  DEFINE_SANITIZER(AAT::ltag);
+  DEFINE_SANITIZER(OT::BASE );
+  DEFINE_SANITIZER(OT::GDEF );
+  DEFINE_SANITIZER(OT::GPOS );
+  DEFINE_SANITIZER(OT::GSUB );
+  DEFINE_SANITIZER(OT::JSTF );
+  DEFINE_SANITIZER(OT::CBDT );
+  DEFINE_SANITIZER(OT::CBLC );
+  DEFINE_SANITIZER(OT::COLR );
+  DEFINE_SANITIZER(OT::CPAL );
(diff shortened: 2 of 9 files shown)


src/hb-private.hh
```diff
@@ -123,7 +123,7 @@ extern "C" void  hb_free_impl(void *ptr);
 #endif
 
 #ifndef HB_INTERNAL
-# if !defined(HB_NO_VISIBILITY) && !defined(__MINGW32__) && !defined(__CYGWIN__) && !defined(_MSC_VER) && !defined(__SUNPRO_CC)
+# if !defined(HB_NO_VISIBILITY) && !defined(__MINGW32__) && !defined(__CYGWIN__) && !defined(_MSC_VER) && !defined(__SUNPRO_CC) && !defined(__DJGPP__) && defined(__unix__)
```
ebraminio: As https://stackoverflow.com/q/43287103 we can remove some other if is harmless

ebraminio:
Something fun to check is to see what is the result of sanitization of current fonts so went ahead with `for i in /Library/Fonts/*.tt{c,f} /System/Library/Fonts/*.tt{c,f}; do echo $i; build/util/hb-sanitize "$i" 1>/dev/null; echo $?; done` and saw `/Library/Fonts/Gujarati Sangam MN.ttc` fails on just recent implemented table, `feat`, I should what has went wrong with it.

ebraminio:
About the previous comment, it seems it passes the predefined borders of `feat` table and fonttools also don't like it and not hb-sanitize nor `feat` table implementation fault.

```
ERROR: An exception occurred during the decompilation of the 'feat' table
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/__init__.py", line 409, in __getitem__
    return self.tables[tag]
KeyError: 'feat'

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/__init__.py", line 424, in __getitem__
    table.decompile(data, self)
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/tables/otBase.py", line 42, in decompile
    self.table.decompile(reader, font)
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/tables/otBase.py", line 618, in decompile
    table[conv.name] = conv.read(reader, font, table)
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/tables/otConverters.py", line 347, in read
    table.decompile(reader, font)
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/tables/otBase.py", line 614, in decompile
    table[conv.name] = conv.readArray(reader, font, table, countValue)
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/tables/otConverters.py", line 110, in readArray
    l.append(self.read(reader, font, tableDict))
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/tables/otConverters.py", line 347, in read
    table.decompile(reader, font)
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/tables/otBase.py", line 618, in decompile
    table[conv.name] = conv.read(reader, font, table)
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/tables/otConverters.py", line 436, in read
    table.decompile(reader, font)
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/tables/otBase.py", line 614, in decompile
    table[conv.name] = conv.readArray(reader, font, table, countValue)
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/tables/otConverters.py", line 110, in readArray
    l.append(self.read(reader, font, tableDict))
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/tables/otConverters.py", line 347, in read
    table.decompile(reader, font)
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/tables/otBase.py", line 618, in decompile
    table[conv.name] = conv.read(reader, font, table)
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/tables/otConverters.py", line 183, in read
    return reader.readUShort()
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/tables/otBase.py", line 134, in readUShort
    value, = struct.unpack(">H", self.data[pos:newpos])
struct.error: unpack requires a buffer of 2 bytes
```

ebraminio:
PTAL

behdad:
Sanitize is fine as a name.

In general this looks good.  Specifically though, it's not valid to include private headers from the utils. We need to add a `hb_sanitize` API and call it. Maybe give it a try?

behdad:
I suppose it can be `hb_blob_t *hb_sanitize (hb_blob_t *, hb_tag_t)`.

ebraminio:
PTAL

TODO
```diff
@@ -19,8 +19,6 @@ API additions
 
 - Add hb-cairo glue
 
-- Add sanitize API (and a cached version, that saves result on blob user-data)
```
ebraminio: Is the rest something I can do here or should I turn that into a GitHub issue?

src/hb-sanitize.cc
```diff
@@ -0,0 +1,131 @@
...
+
+
+/**
+ * hb_sanitize: Sanitizes a table specified by a tag
+ * @face:
+ * @tag:
+ *
+ *
+ *
+ * Return value: Sanitizer result, nullptr if we don't support that table at all
```
ebraminio: I guess this is not its idiomatic way. Please let me know what should be done instead.

ebraminio:
PTAL

What do you suggest for testing? Maybe some tool to run the tool over the entire the repository and bots font files?

src/hb-sanitize.cc
```diff
@@ -0,0 +1,131 @@
...
+ * Since: REPLACEME
+ **/
+hb_blob_t* hb_sanitize (hb_face_t* face, hb_tag_t tag)
```
ebraminio: Behdad, your suggested signature had hb_blob_t instead hb_face_t but what should be done with `num_glyphs`? Passing that also thought the API?

TODO
behdad: No, don't want the caching.

src/hb-sanitize.cc
behdad: Yeah cannot make up my mind so far.

src/hb-sanitize.cc
```diff
@@ -0,0 +1,132 @@
...
+ * Return value: Sanitizer result, nullptr if we don't support that table at all
+ * Since: REPLACEME
+ **/
+hb_blob_t *
+hb_sanitize (hb_face_t *face, hb_tag_t tag)
+{
+  unsigned int num_glyphs = hb_face_get_glyph_count (face);
+  hb_blob_t *blob = hb_face_reference_table (face, tag);
+  hb_blob_t *result = nullptr;
+#define DEFINE_SANITIZER(a, b, c, d, table) \
```
behdad: You don't need the tag explicitly. You can use `table::tableTag`.

src/hb-sanitize.cc
```diff
@@ -0,0 +1,132 @@
...
+  case HB_TAG(a,b,c,d): { \
+    OT::Sanitizer<table> sanitizer; \
+    sanitizer.set_num_glyphs (num_glyphs); \
+    result = sanitizer.sanitize (blob); \
+    break; \
+  }
+  switch (tag)
```
behdad: We should special-case tag 0 to mean sanitise as font file.

src/hb-sanitize.h
```diff
@@ -0,0 +1,45 @@
...
+#ifndef HB_H_IN
+#error "Include <hb.h> instead."
+#endif
+
+#ifndef HB_SANITIZE_H
+#define HB_SANITIZE_H
+
+#include "hb-common.h"
+#include "hb-buffer.h"
+#include "hb-font.h"
```
behdad: Why these three? Looks like you need hb-common, hb-face, and hb-blob.

util/hb-sanitize.cc
```diff
@@ -0,0 +1,100 @@
...
+  hb_blob_t *blob = hb_blob_create_from_file (argv[1]);
+  unsigned int faces = 1; // hb_face_count (blob); #1002
+
+  unsigned int fails = 0;
+  for (unsigned int face_index = 0; face_index < faces; ++face_index)
+  {
+    printf ("Face index: %d\n", face_index);
+    hb_face_t *face = hb_face_create (blob, face_index);
+
+    hb_tag_t tags[100];
```
behdad: Use a number like 16 or 32.

util/hb-sanitize.cc
```diff
@@ -0,0 +1,100 @@
...
+    unsigned int table_count = sizeof (tags) / sizeof (hb_tag_t);
+    hb_face_get_table_tags (face, 0, &table_count, tags);
+
+    for (unsigned int table_index = 0; table_index < table_count; ++table_index)
```
behdad: This is wrong. This doesn't go over all tables.  Look at other places about how to iterate using an api like `get_table_tags` and offsets.

util/hb-sanitize.cc
```diff
@@ -0,0 +1,100 @@
...
+    {
+      char buf[4];
+      hb_tag_to_string (tags[table_index], buf);
+      hb_blob_t *sanitized = hb_sanitize (face, tags[table_index]);
+
+      if (!sanitized)
+	printf ("%.*s: %s\n", 4, buf, "SKIPPED!");
+      else if (hb_blob_get_length (sanitized))
```
behdad: Check that returned blob is equal to input blob. That's definition of sanitization passing.

src/hb-sanitize.cc
behdad: Ok I see that doesn't make sense when you are sending face in.  I'll think about it more...

behdad:
Ok, regarding the API:
- It's unfortunate that we need num_glyphs. That's needed by the AAT::Lookup structure.

Maybe we can add these two APIs:
- `hb_sanitize_file(hb_blob_t, ...)`: This will sanitize as a OpenTypeFontFile.
- `hb_sanitize_table(hb_face_t, tag)`: This does what you have right now.

The results is a returned blob. If the blob is different from the input blob then sanitization had failed. Ouch, ok I see in the second API the caller does not have the blob.

Maybe we should return an enum instead?  Checking for size 0 is not what I would suggest.

behdad:
Also, later, would be nice to have hb-sanitize tool optionally get output filename and save a font with the sanitized tables. That would need the subset_face to become face_builder first.

behdad:
Also, I tested this and it passes even if sanitization failed. That's because sanitizer copies table and fixes things up, if the blob allows that.  Apparently we can do that by making the blob immutable before calling sanitize.

src/hb-sanitize.cc
ebraminio: Done

src/hb-sanitize.h
ebraminio: Done

util/hb-sanitize.cc
ebraminio: Done

util/hb-sanitize.cc
ebraminio: I think if the input buffer is big enough it go through all of it and fonts table usually don't exceed 32 (or 100?)? I guess I can implement a [double loop] iteration (if that is what you want?) but please give a specific place how that is used. Thanks

ebraminio:
> It's unfortunate that we need num_glyphs. That's needed by the AAT::Lookup structure.

How about passing that also? That feels bad as what if we need other things in future (so we should provide another API maybe), but at least we can provide sanitization API independently from hb_face_t so it can be used on font editor software or something but I am not sure.

behdad:
> It's unfortunate that we need num_glyphs. That's needed by the AAT::Lookup structure.
> 
> How about passing that also? That feels bad as what if we need other things in future (so we should provide another API maybe), but at least we can provide sanitization API independently from hb_face_t so it can be used on font editor software or something but I am not sure.

Yeah I don't like that for the reason you mention.

Alternative is that sanitize would not require this, but we pass down table size to every use of AAT:Lookup and do a runtime check. That was too intrusive to the code, so I did it this way.

ebraminio:
> This is wrong. This doesn't go over all tables. Look at other places about how to iterate using an api like get_table_tags and offsets.

So should I do something like this? https://github.com/freedesktop/libreoffice-core/commit/fa1ec9461e2f78655e9aa17473dba085b0b6015d#diff-edeb470f1693db8a762fae32df27343bR841

ebraminio:
As a part of PR cleanup I am doing these days, as https://github.com/harfbuzz/harfbuzz/pull/950#issuecomment-384802100 and the fact this needs the face itself so can get its glyph count so this patch at least with its current approach feels unlikely to go anywhere, let's close and since it exists on https://github.com/harfbuzz/harfbuzz/blob/master/TODO#L16 we can think about it later. Thanks :)

behdad:
Yeah don't need this. Thanks.

PR #952, opened 2018-04-04
Fix build with CoreText on OS X 10.7 and earlier

ryandesign:
Hi, I'm the maintainer of harfbuzz in MacPorts. I was recently asked to enable CoreText support for harfbuzz in MacPorts. In attempting to do so, I found that although CoreText was introduced in Mac OS X 10.5, harfbuzz's CoreText support doesn't work there:

```
hb-coretext.cc: In function “const __CTFont* create_ct_font(CGFont*, CGFloat)”:
hb-coretext.cc:171: error: “kCTFontUIFontSystem” was not declared in this scope
hb-coretext.cc:173: error: “kCTFontUIFontEmphasizedSystem” was not declared in this scope
hb-coretext.cc:209: error: “kCTFontURLAttribute” was not declared in this scope
hb-coretext.cc: In function “hb_bool_t _hb_coretext_shape(hb_shape_plan_t*, hb_font_t*, hb_buffer_t*, const hb_feature_t*, unsigned int)”:
hb-coretext.cc:948: error: “kCTTypesetterOptionForcedEmbeddingLevel” was not declared in this scope
```

According to Apple documentation, `kCTTypesetterOptionForcedEmbeddingLevel` was introduced in 10.5, so I'm not sure why on 10.5 it says it's not declared, but here is someone else's patch for that:

https://github.com/jralls/gtk-osx-build/blob/master/patches/harfbuzz-kCTTypesetterOptionForcedEmbeddingLevel-Leopard.patch

`kCTFontURLAttribute` was introduced in 10.6; I found and adapted some compatibility code I found for that.

`kCTFontUIFontSystem` and `kCTFontUIFontEmphasizedSystem` were introduced in 10.8, as new names for older constants that existed before. So on older systems, I defined these new constants to the old ones.

This builds for me now with `--with-coretext` on 10.5 and later and the test suite passes, though I wasn't sure if the tests cover the CoreText functionality.

### DIFF
--- a/src/hb-coretext.cc
+++ b/src/hb-coretext.cc
@@ -168,6 +168,10 @@ create_ct_font (CGFontRef cg_font, CGFloat font_size)
   if (CFStringHasPrefix (cg_postscript_name, CFSTR (".SFNSText")) ||
       CFStringHasPrefix (cg_postscript_name, CFSTR (".SFNSDisplay")))
   {
+#if MAC_OS_X_VERSION_MIN_REQUIRED < 1080
+# define kCTFontUIFontSystem kCTFontSystemFontType
+# define kCTFontUIFontEmphasizedSystem kCTFontEmphasizedSystemFontType
+#endif
     CTFontUIFontType font_type = kCTFontUIFontSystem;
     if (CFStringHasSuffix (cg_postscript_name, CFSTR ("-Bold")))
       font_type = kCTFontUIFontEmphasizedSystem;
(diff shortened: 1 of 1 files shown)


ebraminio:
That's great! Feel free to test with `util/hb-view /System/Library/Fonts/Times.ttc ABC --shaper=coretext` (or any other font available) then I guess we are ready to merge. Thanks!

behdad:
Thanks Ryan. What Ebrahim said.

ebraminio:
Just checked this on my 10.6 machine and found the fails are just with `.ttc` and `.dfont` but not `.ttf` fonts which makes this good enough for the merge. About `.ttc` and `.dfont`, we should see what font loader is used for them (which I guess is freetype) and that probably is the source of issues not the portable codes harfbuzz have. So, thank you all

PR #962, opened 2018-04-11
[ot-color] Cosmetic changes

ebraminio:


src/hb-open-type-private.hh
```diff
@@ -909,7 +909,6 @@ struct UnsizedArrayOf
     return_trace (true);
   }
 
-  private:
```
ebraminio: The only interesting part of this patch is this I guess, making `sanitize_shallow` public to use in `sbix` so we don't have to sanitize all the glyphs theirselves also.

PR #971, opened 2018-04-12
Always enable atexit on Android

ebraminio:
Obviously one can use a newer NDK for building an updated HarfBuzz
instead that now pretty old version however I am concerned if that
version checking is working with clang that is used with the newer
NDK versions.

### DIFF
--- a/src/hb-private.hh
+++ b/src/hb-private.hh
@@ -226,11 +226,9 @@ static int errno = 0; /* Use something better? */
  * mingw32 headers say atexit is safe to use in shared libraries.
  */
 #    define HB_USE_ATEXIT 1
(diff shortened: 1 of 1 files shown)


ebraminio:
Ok, so we really want to support that old NDK versions? If so I can revert this and just add `|| defined(__clang__)`

ebraminio:
For the context, on clang `__GNUC__ __GNUC_MINOR__` is `4 2` so that check was for example passing on gcc Chrome build but not clang Chrome builds and that is not expected I guess.

Other than that, there is no real users on old NDK version that are locked in to use old version AFAICU. NDK is usually used by third parties for building minor C libraries and not packaging harfbuzz itself into their assets, which even if they use, they can update their NDK version easily, and Google itself uses brand new NDK for compiling Chrome which we don't need to worry anyway.

Issue #432, opened 2017-03-01
Consider exposing access to 'name' table on hb_face_t or hb_font_t


drott:
For fixing the `@font-face { src: local(xyz) }` matching, we need to look at postscript name or full name in the `name` table. It seems like `hb-ot-name.hh` already has some parsing for this. 

If accessing such information would be available on `hb_face_t`, that would be easiest for reusing in Blink and allow us to at least validate a match for `local()`.

roozbehp:
Useful in Android too, although for a different reason (debugging).

KrasnayaPloshchad:
What happened if this table works with 'stat' table?

behdad:
My current thinking on the API for this:

- We want the actual name fetcher to have two variants, one for UTF-8, another for UTF-16.  The cleanest, though not most user-friendly, would be the fetcher only takes an index into the list of names. The problem is, this is NOT the nameID, but an otherwise useless index that the user first needs to fetch using following API,

- The matching API takes nameID, platformID, platEncID, languageID.  Each of platformID, platEncID, and languageID can each be "wildcard" (0xFFFF), to let HarfBuzz choose best.  Easiest would be for this API to return the index to be used by the fetching API, to avoid having to expose separate UTF-8 / UTF-16 versions,

- An enumeration API. My currently thinking is that this one simply takes start_offset and count and returns the nameID/platformID/platEncID/languageID of the name table entry at those indices.  Ie. this is very raw.

What this API does not make easy is to list all languages of a nameID. Maybe the enumeration API can have another counterpart that takes a nameID?

cc @khaledhosny 

- API to fetch name by an index into the list of names.

behdad:
I thought more about this. I'm leaning towards NOT exposing the platform/encoding/language madness, and provide a higher-level API that works with hb_language_t only.

twardoch:
Just 4 remarks (more observations then recommendations, as I don't have opinions on the API but perhaps it'll be useful): 

- I agree that smart stuff like falling back from name ID 17 to 2 may be out of scope, so reading each name ID should be treated the same

- What the API could do is to simplify the language matching. The 3.1 name branch uses the old Win 32 API language numbers, the 1.0 or 0.3 branches use something else; it would be nice if the API tied them into HarfBuzz language tags

- In reality, the "English" names are often treated as default, so something like "undefined language" or "generic language" should fetch the English names

[...]

PR #976, opened 2018-04-12
Add two APIs for getting stylistic set label

### LINKED ISSUE #425: get StyleSet names if available
Scribus uses HarfBuzz to get font features, but right now harfbuzz gives only the tags. This is enough for almost font features expect Stylistic Set 1 - Stylistic Set 20. In [one of scribus bugs](https://bugs.scribus.net/view.php?id=14621), a user requests a feature to display StyleSet names if available instead of just displaying ss01, ss02, ss03, …, ss20. 

Is this possible with current api?

ebraminio:
Fixes #425, *highly unready* patch that just works but needs to be improved a lot but like to hear your comment.

Snippets to make it work

```
  unsigned int label_id = hb_ot_layout_get_stylistic_set_label (font->face, HB_TAG ('s','s','0','1'));
  char str[256] = {0};
  hb_get_name_entry (font->face, 1, 0, 0, label_id, str, 255);
  printf ("%s", str);
```


src/hb-ot-layout-common-private.hh
```diff
@@ -479,6 +479,13 @@ struct FeatureParams
     return Null(FeatureParamsSize);
   }
 
+  inline int get_stylistic_set_name_id (hb_tag_t tag) const
```
khaledhosny: Does this work with `cvXX` features, they can have name IDs as well. Also I’d rename it `get_feature_name_id` to be more general.

src/hb-ot-layout.cc
```diff
@@ -1047,6 +1055,48 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
 }
 
 
+/**
+ * hb_ot_layout_get_stylistic_set_label:
+ *
+ * Since: REPLACEME
+ **/
+int
+hb_ot_layout_get_stylistic_set_label (hb_face_t    *face,
```
khaledhosny: It should be `_get_feature_label_name_id` or something since it returning the name IDs of the label, not the actual label itself.

src/hb-ot-layout.cc
```diff
@@ -1047,6 +1055,48 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
...
+				      hb_tag_t      tag)
+{
+  const OT::GSUB &gsub = _get_gsub (face);
+
+  unsigned int num_features = gsub.get_feature_count ();
```
khaledhosny: Shouldn’t this handle GPOS features as well?

src/hb-ot-layout.cc
```diff
@@ -1047,6 +1055,48 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
...
+  return -1;
+}
+
+/**
+ * hb_get_name_entry:
+ *
+ * Since: REPLACEME
+ **/
+int
+hb_get_name_entry (hb_face_t    *face,
```
khaledhosny: I think this should be done in a separate PR, returning name ID here should be enough for many clients, and an API for accesing name tables should be good on its own independent of this PR.

src/hb-ot-layout-common-private.hh
ebraminio: There are multiple name id associated with them https://docs.microsoft.com/en-us/typography/opentype/spec/features_ae#cv01-cv99 What do you suggest?

ebraminio: Used featUILableNameID but probably some API is needed to get all of them.

src/hb-ot-layout-common-private.hh
```diff
@@ -479,6 +479,15 @@ struct FeatureParams
     return Null(FeatureParamsSize);
   }
 
+  inline unsigned int get_name_id (hb_tag_t tag) const
+  {
+    if ((tag & 0xFFFF0000u) == HB_TAG ('s','s','\0','\0')) /* ssXX */
+      return u.stylisticSet.uiNameID;
+    if ((tag & 0xFFFF0000u) == HB_TAG ('c','v','\0','\0')) /* cvXX */
+      return u.characterVariants.featUILableNameID;
```
ebraminio: One can ask why just returning this and not the other Name IDs. I am asked to do so but I guess that can be done using a separate API and this can support just `featUILableNameID` (or maybe not?)

khaledhosny: What about something like:
```C
hb_bool_t
hb_ot_layout_get_feature_name_ids (unsigned int* label_id,
                                   unsigned int* tooltip_id,
                                   unsigned int* sample_id,
                                   unsigned int* param_id)
{
  hb_bool_t result = false;

  if (*label_id)
      …;
  if (*tooltip_id)
      …;

  …

  return result;
}
```

khaledhosny: May be pass also `numNamedParameters` and `character[charCount]` and rename it `hb_ot_layout_get_feature_parameters ()` or something.

ebraminio: Done

src/hb-ot-layout.cc
```diff
@@ -1009,43 +1027,109 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
...
   return false;
 }
 
+/**
+ * hb_ot_layout_get_feature_parameters:
+ *
+ * Since: REPLACEME
+ **/
+hb_bool_t
+hb_ot_layout_get_feature_parameters (hb_face_t       *face,
```
khaledhosny: I forgot that we have `hb_ot_layout_get_size_params()`, sorry. It should be `hb_ot_layout_get_feature_params()` for consistency then.

src/hb-ot-layout.cc
```diff
@@ -1009,43 +1027,109 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
...
+    _get_gsubgpos_matched_feature_params (face, tag);
+  if (&feature_params != &OT::Null (OT::FeatureParams))
+  {
+    if ((tag & 0xFFFF0000u) == HB_TAG ('s','s','\0','\0')) /* ssXX */
+    {
+      const OT::FeatureParamsStylisticSet& params =
+        feature_params.get_stylistic_set_params (tag);
+
+#define PARAM(a, A) if (a) *a = params.A
+#define UNPARAM(a) if (a) *a = 0
```
khaledhosny: `0` is a valid name ID (copyright), I think we better return something different for missing IDs.

khaledhosny: In #244 `0xFFFF` was returned for such cases, I think we should do the same here and document it.

src/hb-ot-layout.cc
```diff
@@ -1009,43 +1027,109 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
 			      unsigned int *range_start,       /* OUT.  May be nullptr */
 			      unsigned int *range_end          /* OUT.  May be nullptr */)
 {
```
ebraminio: Extracted _get_gsubgpos_matched_feature_params from hb_ot_layout_get_size_params. The difference is however it searches gsub table also which is negligible I guess but easily can be fixed.

src/hb-ot-layout.cc
```diff
@@ -1009,43 +1027,109 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
...
+				     unsigned int    *first_param_id       /* OUT.  May be nullptr */,
+				     unsigned int    *char_count           /* OUT.  May be nullptr */,
+				     hb_codepoint_t **characters           /* OUT.  May be nullptr */)
+{
+  const OT::FeatureParams &feature_params =
+    _get_gsubgpos_matched_feature_params (face, tag);
+  if (&feature_params != &OT::Null (OT::FeatureParams))
+  {
+    if ((tag & 0xFFFF0000u) == HB_TAG ('s','s','\0','\0')) /* ssXX */
+    {
```
khaledhosny: Instead of repeating the tag comparison that `get_stylistic_set_params()` already does, what about calling it unconditionally and checking if the result is not `Null`?.

src/hb-ot-layout.cc
```diff
@@ -1009,43 +1027,109 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
...
+      UNPARAM(tooltip_id);
+      UNPARAM(sample_id);
+      UNPARAM(num_named_parameters);
+      UNPARAM(first_param_id);
+      UNPARAM(char_count);
+      UNPARAM(characters);
+      return true;
+    }
+    if ((tag & 0xFFFF0000u) == HB_TAG ('c','v','\0','\0')) /* cvXX */
+    {
```
khaledhosny: Same as above.

src/hb-ot-layout.cc
```diff
@@ -1009,43 +1027,109 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
...
   }
 
-#define PARAM(a, A) if (a) *a = 0
-  PARAM (design_size, designSize);
-  PARAM (subfamily_id, subfamilyID);
-  PARAM (subfamily_name_id, subfamilyNameID);
-  PARAM (range_start, rangeStart);
-  PARAM (range_end, rangeEnd);
-#undef PARAM
+#define UNPARAM(a) if (a) *a = 0
```
ebraminio: Changed it to UNPARAM to match with the `hb_ot_layout_get_feature_parameters` counterpart.

ebraminio:
@khaledhosny, is this looks good to you now?

src/hb-ot-layout.cc
```diff
@@ -1009,43 +1027,109 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
...
+    {
+      const OT::FeatureParamsCharacterVariants& params =
+        feature_params.get_character_variants_params (tag);
+
+      PARAM(label_id, featUILableNameID);
+      PARAM(tooltip_id, featUITooltipTextNameID);
+      PARAM(sample_id, sampleTextNameID);
+      PARAM(num_named_parameters, numNamedParameters);
+      PARAM(first_param_id, firstParamUILabelNameID);
+      PARAM(char_count, characters.len);
```
khaledhosny: What if any of these was `NULL` would we just return zero here?

src/hb-ot-layout.cc
ebraminio: Done

src/hb-ot-layout.cc
ebraminio: Done

src/hb-ot-layout.cc
ebraminio: I guess no? It follows the idea of `hb_ot_layout_get_size_params`.

ebraminio:
@khaledhosny PTAL

src/hb-ot-layout.cc
```diff
@@ -1009,43 +1027,109 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
...
 }
 
+/**
+ * hb_ot_layout_get_feature_params:
+ *
+ * Returns bunch of NameIDs and other related information for a tag,
+ * and sets NameIDs parameters (suffixed with _id) to 0xFFFF in case
+ * the requested tag not found or the tag doesn't support it.
+ *
+ * Since: REPLACEME
```
khaledhosny: Please document the function arguments as well, otherwise gi-introspection will be broken.

src/hb-ot-layout.cc
```diff
@@ -1009,43 +1027,109 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
...
+hb_bool_t
+hb_ot_layout_get_feature_params (hb_face_t       *face,
+				 hb_tag_t         tag,
+				 unsigned int    *label_id             /* OUT.  May be nullptr */,
+				 unsigned int    *tooltip_id           /* OUT.  May be nullptr */,
+				 unsigned int    *sample_id            /* OUT.  May be nullptr */,
+				 unsigned int    *num_named_parameters /* OUT.  May be nullptr */,
+				 unsigned int    *first_param_id       /* OUT.  May be nullptr */,
+				 unsigned int    *char_count           /* OUT.  May be nullptr */,
+				 hb_codepoint_t **characters           /* OUT.  May be nullptr */)
```
khaledhosny: The `OUT` comments will be redundant with the documentation above, so I suggest dropping them (they make the lines too wide as well).

src/hb-ot-layout.cc
ebraminio: Done

src/hb-ot-layout.cc
ebraminio: Done

ebraminio:
@khaledhosny PTAL, the documentation says `May be NULL` so maybe zero was a good NameID anyway?

ebraminio:
Done

ebraminio:
I guess we should somehow notify the users to `free` the `characters` string when they are done with on the documentation?

src/hb-ot-layout.cc
```diff
@@ -1046,6 +1064,97 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
...
+    if (&character_variants_params != &OT::Null (OT::FeatureParamsCharacterVariants)) /* cvXX */
+    {
+      PARAM(label_id, character_variants_params.featUILableNameID);
+      PARAM(tooltip_id, character_variants_params.featUITooltipTextNameID);
+      PARAM(sample_id, character_variants_params.sampleTextNameID);
+      PARAM(num_named_parameters, character_variants_params.numNamedParameters);
+      PARAM(first_param_id, character_variants_params.firstParamUILabelNameID);
+      PARAM(char_count, character_variants_params.characters.len);
+      unsigned int len = character_variants_params.characters.len;
+      if (len != 0 && characters)
```
khaledhosny: This should be `&& *chracters` I think.

src/hb-ot-layout.cc
```diff
@@ -1046,6 +1064,97 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
...
+      {
+	hb_codepoint_t *chars = (hb_codepoint_t*) malloc (len * sizeof (hb_codepoint_t));
+	for (unsigned int i = 0; i < len; ++i)
+	  chars[i] = character_variants_params.characters[i];
+	*characters = chars;
```
khaledhosny: May be instead of copying, just return a pointer to `character_variants_params.characters`, this way the caller does not need to free this (and note in the documentation that this pointer is only valid as long as the face is valid, so the caller should copy itself).

src/hb-ot-layout.cc
ebraminio: I think `characters` is correct but I am not sure.

src/hb-ot-layout.cc
ebraminio: I like that to be honest but it seems this is not the convention as it is done here [here](https://github.com/harfbuzz/harfbuzz/blob/f24b0b9728f87d7599867a4ba6ec7d30da43b869/src/hb-ot-name-table.hh#L100) also apparently but the main problem is it uses [3 bytes characters](https://github.com/harfbuzz/harfbuzz/blob/435b1878e710b67e21872572ad9fd5ed1369e97f/src/hb-ot-layout-common-private.hh#L452) and I guess it would be more nice to resolve it for the client.

src/hb-ot-layout.cc
khaledhosny: Of course it is, silly me! I need my morning tea.

src/hb-ot-layout.cc
ebraminio: Also does `Unicode Scalar Value` correspond to hb_codepoint_t? If not, what should be done instead?

khaledhosny: The alternative is to have the client give us an allocated array to fill, but this makes the API a bit more complex. Allocating internally using `malloc` and expecting the client to call `free` is bad design IMHO, since we might be using different memory allocation methods, especially since we allow overriding them.
 
Unicode Scalar Value is another name for UTF-32 code points, so yes they are the same.

khaledhosny: I don’t think we use `name` table `get_name()` any where, but if you look closely you will see it doesn’t do any allocation and takes in an allocated buffer and length.

You can do the same here, which is even a better alternative. Allocating internally using `malloc` and expecting the client to call `free` is bad design IMHO, since we might be using different memory allocation methods, especially since we allow overriding them.
 
Unicode Scalar Value is another name for UTF-32 code points, so yes they are the same.

ebraminio: > Allocating internally using malloc and expecting the client to call free is bad design IMHO

Oh OK, so will follow that

ebraminio:
```
hb_ot_layout_get_feature_params (hb_face_t       *face,
				 hb_tag_t         tag,
				 unsigned int    *label_id,
				 unsigned int    *tooltip_id,
				 unsigned int    *sample_id,
				 unsigned int    *num_named_parameters,
				 unsigned int    *first_param_id,
				 unsigned int    *char_count,
				 unsigned int     buffer_length,
				 hb_codepoint_t **characters)
```

Should the signature should look like this?

khaledhosny: I haven’t seen one in the wild, but you can make one with FontTools, I guess.

ebraminio:
I *think* I got its signature and implementation right this time and I updated the snippet of how it should be used on the first comment. PTAL

src/hb-ot-layout.cc
ebraminio: Using your tips and looking at `hb_face_get_table_tags` implementation I guess I've got it right this time, have a look please.

src/hb-ot-layout.cc
ebraminio: Too much hassle :D

src/hb-ot-layout.cc
```diff
@@ -1046,6 +1064,93 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
...
+      PARAM(label_id, stylistic_set_params.uiNameID);
+      // ssXX features don't have the rest
+      PARAM(tooltip_id, 0);
+      PARAM(sample_id, 0);
+      PARAM(num_named_parameters, 0);
+      PARAM(first_param_id, 0);
+      PARAM(char_count, 0);
+      return true;
+    }
+    const OT::FeatureParamsCharacterVariants& character_variants_params =
```
khaledhosny: Small nit; I’d personaly prefer shorter variable names for `stylistic_set_params` and `character_variants_params` variables, may be `ss_params` and `cv_params`.

ebraminio:
Done, added a start_offset also like [hb_face_get_table_tags](https://github.com/harfbuzz/harfbuzz/blob/0ed69c946247feb5bbb5be06d5314c5e0c0fec67/src/hb-face.cc#L489) please check if its logic and its documentation wording is correct (or was it needed anyway). Thanks

ebraminio:
Given an API a similar API is used here, https://github.com/freedesktop/libreoffice-core/commit/fa1ec9461e2f78655e9aa17473dba085b0b6015d#diff-edeb470f1693db8a762fae32df27343bR841 should this also return characters count also?

behdad:
Ok, my review:

- Please `#define HB_NO_NAME_ID 0xFFFF` and return that instead of 0.  I'm not sure where to define that macro.  Maybe hb-ot-layout.h is fine for now.

- I suggest we split the API in two: `hb_ot_layout_get_feature_name_ids` and `hb_ot_layout_get_feature_characters`.  The latter should return `unsigned int` like the other APIs that use a `start_offset` model.

The rest looks good!  Thanks!

mhosken:
Request to advance this and perhaps merge this PR. Libreoffice want to use it (whatever "it" ends up being).

ebraminio:
PTAL

ebraminio:
Or maybe defer `hb_ot_layout_get_feature_characters`, the untested part, to sometime later?

mhosken:
The latest (v5) of the Charis font from SIL has ssxx and cv0x features with
featoarams and names. Available from
https://software.sil.org/downloads/r/charis/CharisSIL-5.000.zip


GB,
Martin

ebraminio:
Thanks. There is no sample characters on it however, may I just remove `hb_ot_layout_get_feature_characters` as it is not used anyway?

Behdad, I've applied 0xFFFF as default but `cv01` name ids itself [can be NULL](https://docs.microsoft.com/en-us/typography/opentype/spec/features_ae), so may I detect and change them or use that as default?

ebraminio:
WDYT Behdad?

ebraminio:
But just feel free to revert this (or request to) before the release

behdad:
I actually did review this on phone.  Will type comments.  Am going to request changes to the API and impl.  Will post soon. Thanks.

behdad:
> But just feel free to revert this (or request to) before the release

Let's not do this (adding API before review) again.  I like to keep master in releasable state all the time, and APIs are one thing that should never be in without agreement.  It's just too messy to revert and what not.

Also, please include the full file "Fixes GITHUB-URL" in the commit message, not just "#123".  Who knows where the project will be hosted ten years from now...

behdad:
As to review:

 - Please add back `hb_ot_layout_get_feature_characters`.  Just because no font doesn't use it is not enough to not implement part of a feature.  It's chicken and egg... Plus, you can build a test font using FontTools.

src/hb-ot-layout.cc
```diff
@@ -1066,6 +1084,74 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
...
+				   hb_tag_t        feature,
+				   unsigned int   *label_id,             /* OUT.  May be nullptr  */
+				   unsigned int   *tooltip_id,           /* OUT.  May be nullptr  */
+				   unsigned int   *sample_id,            /* OUT.  May be nullptr  */
+				   unsigned int   *num_named_parameters, /* OUT.  May be nullptr  */
+				   unsigned int   *first_param_id        /* OUT.  May be nullptr  */)
+{
+  const OT::FeatureParams &feature_params =
+    _get_gsubgpos_matched_feature_params (face, feature);
+  if (&feature_params != &Null (OT::FeatureParams))
```
behdad: This should not be needed.  Null objects are designed to be used directly...  What we want however is `DEFINE_NULL_DATA` for the two `FeatureParams` types in this change to use 0xFFFF instead of zero bytes.

src/hb-ot-layout.cc
```diff
@@ -1066,6 +1084,74 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
...
+  {
+    const OT::FeatureParamsStylisticSet& ss_params =
+      feature_params.get_stylistic_set_params (feature);
+    if (&ss_params != &Null (OT::FeatureParamsStylisticSet)) /* ssXX */
+    {
+#define PARAM(a, A) if (a) *a = A
+      PARAM(label_id, ss_params.uiNameID);
+      // ssXX features don't have the rest
+      PARAM(tooltip_id, 0);
+      PARAM(sample_id, 0);
```
behdad: Why do these still use 0?

src/hb-ot-layout.cc
```diff
@@ -1066,6 +1084,74 @@ hb_ot_layout_get_size_params (hb_face_t    *face,
...
+{
+  const OT::FeatureParams &feature_params =
+    _get_gsubgpos_matched_feature_params (face, feature);
+  if (&feature_params != &Null (OT::FeatureParams))
+  {
+    const OT::FeatureParamsStylisticSet& ss_params =
+      feature_params.get_stylistic_set_params (feature);
+    if (&ss_params != &Null (OT::FeatureParamsStylisticSet)) /* ssXX */
+    {
+#define PARAM(a, A) if (a) *a = A
```
behdad: I don't think the macro is quite justified :)).

behdad:
> @khaledhosny PTAL, the documentation says May be NULL so maybe zero was a good NameID anyway?

No.  Please no 0. 0 is a valid name ID.

ebraminio:
> Let's not do this (adding API before review) again.

As you initial review I thought we agree on API. Ok. I am sorry.

> No. Please no 0. 0 is a valid name ID.

So NULL should turned into 0 also?

> I don't think the macro is quite justified :)).

It is used on hb_ot_layout_get_size_params like this, I am removing it.

> Plus, you can build a test font using FontTools.

If someone creates one I will bring it back :)

behdad:
> No. Please no 0. 0 is a valid name ID.
> 
> So NULL should turned into 0 also?

No. Those are two different things.  If user passes NULL, it means they don't want to fetch this nameID.  If they pass non-NULL, but the font doesn't have this name, then we store 0xFFFF.  If the font has it, we store it.

>     I don't think the macro is quite justified :)).
> 
> It is used on hb_ot_layout_get_size_params like this, I am removing it.

Ah, ok. Hard to argue with that. Either way I'm fine.

>     Plus, you can build a test font using FontTools.
> 
> If someone creates one I will bring it back :)

Bring it back. Even untested it's fine. The way the HarfBuzz structs work, it's near-impossible to get it wrong.

behdad:
Now I remember.  I also have concern about the API and how it finds a feature to pull names out of.  We did it that way for size feature, because spec says all size features should agree on the size parameters.  For stylistic sets / character variants, it's very reasonable for different script / language-systems to have different values.  As such, I think the API should take a feature-index only.  User should fetch feature-index using existing API before calling this API.

behdad:
Err, table_index and feature_index.

ebraminio:
Ok. So lets revert this and apply it again?

behdad:
> Ok. So lets revert this and apply it again?

Whichever is easier.

Khaled, what do you think about proposed API change?

khaledhosny:
Makes sense. I don’t think I have seen a font where `ssXX` features differ across scripts, but it is a possibility and the API should handle it indeed. Returning `0xFFFF` for missing IDs makes sense too, but I’m not sure if `hb_ot_layout_get_size_params()` does this for its name IDs and whether we want to change it to match.

khaledhosny:
Here is a sample font generated with FontTools from the [feature files spec example](https://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#8.d):
[Test.zip](https://github.com/harfbuzz/harfbuzz/files/2249438/Test.zip)

behdad:
Let's not get stuck here.

ebraminio:
I agree also but am very busy on releasing new version of [my app](https://github.com/ebraminio/DroidPersianCalendar) with Arabic support also :)

ebraminio:
#1239

ebraminio:
@mhosken This is reverted but merged again eventually, is LibreOffice able name table itself? If not have a look at #432

PR #977, opened 2018-04-13
Don't decompose Bengali Rra and Rha

### LINKED ISSUE #779: Bangla font rendering problem with letter ড় ঢ় in LibreOffice 6.0 with some bengali fonts
I reported this issue on https://bugs.documentfoundation.org/show_bug.cgi?id=115421 on 3rd Feb, 2018. (Bug Id- 115421). In reply of this bug report, Mr. Khaled Hosny told me to report this bug here, so I did so. The problem as I told early is problem with Nukta sign. Libre Office use this character to create few bengali alphabet like  ড় ঢ় where the round dots of those letters become slightly displaced with some bengali fonts. But I test those fonts in alternative office packages like OpenOffice, MS Office where the Nukta character doesn't use to form those alphabets. My request is please fix this problem of LibreOffice in next update. I attaced few screenshots and few font which are used

.
27332268_1789682217708983_7452483768167553570_n
27336372_1789682214375650_3125569294339190204_n

ebraminio:
Fixes #779, the original font is copyrighted and I wanted to create a subset using hb-subset or pyftsubset but I need a hand here as their result doesn't look correct, so I guess I need to do it manually?

The fix itself looks legit as the same is done for `DEVANAGARI LETTER RRA` also apparently.

After the change,
```
build/util/hb-view "Bangla Akademi.ttf" --unicodes="09DC 0020 09DD 0020 09A1 09BC 0020 09A2 09BC"  --shapers=ot
```

a


ebraminio:
The fix itself looks legit as the same is done for `DEVANAGARI LETTER RRA` also apparently.

src/hb-ot-shape-complex-indic.cc
```diff
@@ -1472,6 +1472,8 @@ decompose_indic (const hb_ot_shape_normalize_context_t *c,
     case 0x0931u  : return false; /* DEVANAGARI LETTER RRA */
     case 0x0B94u  : return false; /* TAMIL LETTER AU */
 
+    case 0x09DCu  : return false; /* BENGAlI LETTER RRA */
+    case 0x09DDu  : return false; /* BENGAlI LETTER RHA */
```
dscorbett: “BENGAlI” should be in all caps.

ebraminio: Opps! Done

kuntalkundu:
thanks for response developers. Now what can I do please tell me. You said this font has a copyright issue so If you want I can send you some free Bengali Unicode font which has too same problem with those particular letters. Please give me some suggestions.

kuntalkundu:
Here I attached few bengali fonts which has same issue in Nuqta dot in Libre Office. 
[Nikosh.zip](https://github.com/harfbuzz/harfbuzz/files/1948503/Nikosh.zip)

behdad:
Thanks @ebraminio. The commit adding the existing two (b01d9b3d90e892341ee4463f2eda4600850b97d8) says it does because uniscribe doesn't. So I'm happy adding these two if uniscribe doesn't either. Would be nice to discover more.

src/hb-ot-shape-complex-indic.cc
```diff
@@ -1472,6 +1472,8 @@ decompose_indic (const hb_ot_shape_normalize_context_t *c,
     case 0x0931u  : return false; /* DEVANAGARI LETTER RRA */
     case 0x0B94u  : return false; /* TAMIL LETTER AU */
```
behdad: No need for the new-line here. Just sort these two with the previous two. Also add link to github issue.

ebraminio:
@kuntalkundu, thanks!

ebraminio:
@behdad, applied the changes, have a look if now looks ok.

kuntalkundu:
attaching few more fonts from same font family, Please check
[Desktop.zip](https://github.com/harfbuzz/harfbuzz/files/1950147/Desktop.zip)

Issue #681, opened 2018-01-06
Support "Traditional Arabic Windows 3.1 font page"


khaledhosny:
(This comment is edited by Ebrahim) Currently it needs Uniscribe support as the client needs to use `ScriptGetCMap()` or so since the fonts don’t have an Arabic `cmap` subtable.

However these fonts are still rather popular among Arabic users and there are hundreds of them floating around the web and people often complain about them not working on HarfBuzz-using applications (for example https://github.com/libass/libass/issues/292).

I don’t know if we can just do PUA shaping similar to Thai or if there is someway to identify these fonts like the Windows API do.

behdad:
I like to do this, but don't think we'd have ANY way to detect these.

dscorbett:
It should be possible to detect them. ssee1256.fon declares its [`dfCharSet`](https://support.microsoft.com/en-my/help/65123/windows-developers-notes-font-file-format) as 178 ([Arabic](https://msdn.microsoft.com/en-us/library/windows/desktop/bb322881(v=vs.85).aspx)), which [corresponds to Windows-1256](https://msdn.microsoft.com/en-us/library/cc194829.aspx). The font doesn’t fully comply with Windows-1256 but it is similar. Compare with ssee1255.fon, which uses a modification of Windows-1255.

behdad:
Right. But we don't support .fon files. Client font-funcs should support them before we can shape anything, and they should synthesize the cmap.

I'm open to coming up with a way to implement .fon and Type 1 in HarfBuzz itself if it's not too intrusive and does not distract from OpenType stuff.

anoosi:
right now i used libass+harfbuzz last ver.
how can i enable Windows-1256 private shaping in windows to get the old fonts work..

ebraminio:
Does the implementation have to be cross platform (and afterward support big endian machines also)? If so I guess we will have so fun with parsing EXE resources (FON is just an alias) and then FNT, writing a limited set of little endian int parsers. And do we need just to implement something equivalent to `ScriptGetCMap` or we need more? I mean the other font rendering stack layers are ready for it? It will be like my .dfont work but way cooler! (but I don't if I can reach to it)

ebraminio:
[...]

PR #981, opened 2018-04-13
Add dump-fon, a prototype on how to parse .fon/.fnt files

ebraminio:
Hopefully eventually to workaround #681, useful links,
* https://github.com/wine-mirror/wine/blob/master/tools/sfnt2fon/sfnt2fon.c
* https://github.com/andrewshadura/gbdfed/blob/master/bdffnt.c
* https://anolis.codeplex.com/
* https://github.com/ebraminio/PDB-Downloader/blob/master/main.cc
* https://web.archive.org/web/20120215123301/http://support.microsoft.com/kb/65123
* http://www.csn.ul.ie/~caolan/publink/winresdump/winresdump/doc/resfmt.txt
* https://github.com/juanitogan/mkwinfont/blob/master/python/dewinfont.py
* https://github.com/juanitogan/mkwinfont/blob/master/python/mkwinfont.py
* http://font.gohu.org (.fon, would be useful)

I will try eventually to contain all the needed changes (including LE ints) PE/NE exe/fnt structs into one file.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -362,11 +362,15 @@ dist_check_SCRIPTS += \
 endif
 
 check_PROGRAMS += \
+	dump-fon \
 	dump-indic-data \
 	dump-khmer-data \
 	dump-myanmar-data \
 	dump-use-data \
 	$(NULL)
+dump_fon_SOURCES = dump-fon.cc
+dump_fon_CPPFLAGS = $(HBCFLAGS)
+dump_fon_LDADD = libharfbuzz.la $(HBLIBS)
 dump_indic_data_SOURCES = dump-indic-data.cc hb-ot-shape-complex-indic-table.cc
 dump_indic_data_CPPFLAGS = $(HBCFLAGS)
 dump_indic_data_LDADD = libharfbuzz.la $(HBLIBS)
--- a/src/dump-fon.cc
+++ b/src/dump-fon.cc
@@ -1,0 +1,475 @@
+/*
+ * Copyright © 2018  Ebrahim Byagowi
+ *
+ *  This is part of HarfBuzz, a text shaping library.
+ *
+ * Permission is hereby granted, without written agreement and without
+ * license or royalty fees, to use, copy, modify, and distribute this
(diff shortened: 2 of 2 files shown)


ebraminio:
Now can extract ".fnt"s from .fon of New EXE format, the next step is extracting parsing them and supporting PE EXE format.

khaledhosny:
#681 isn’t about about Windows bitmap fonts, but old, pre-OpenType, TrueType fonts that used some sort of PUA encoding that I failed to find any documentation for and that works only on Windows. I thought the code mentioned in the issue helps shaping these fonts on Windows, but I might have been mistaken as I didn’t actually test any of this with HarfBuzz.

ebraminio:
To be honest I just followed Behdad comment on "Right. But we don't support .fon files." So what we should do with this implementation?

ebraminio:
There is no .fon files on Win 3.1 or Win 10 (!) that is using PE instead NE so I guess we can go without it.

behdad:
> #681 isn’t about about Windows bitmap fonts, but old, pre-OpenType, TrueType fonts that used some sort of PUA encoding that I failed to find any documentation for and that works only on Windows. I thought the code mentioned in the issue helps shaping these fonts on Windows, but I might have been mistaken as I didn’t actually test any of this with HarfBuzz.

AFAIU the font in question was `ssee1256.fon`. I'm not aware of any non-Unicode Arabic TTF fonts on Windows.

behdad:
Not sure how to detect those. Maybe the cmap platform etc has some clues?

khaledhosny:
I just accidentally found something interesting in FontValidator that might help solving this “mystery”: If `OS/2` version is 0, some of the reserved bits seem to indicate “win 3.1 font page”: https://github.com/Microsoft/Font-Validator/blob/master/OTFontFileVal/val_OS2.cs#L644-L681

The font in question seems to have `TRAD_ARABIC_FONT_PAGE` set.

ebraminio:
Cooool, I will write its detection part.

Issue #539, opened 2017-09-18
Indy3! Indic scripts through USE shaper


behdad:
Support 'xxx3' script tags for Indic scripts through USE shaper.  Apple does that and we want to put it in OpenType itself.  According to Ned, here's the overrides they use:
```
# Indic_Syllabic_Category=Nukta
0F71       ; Nukta          # Mn       TIBETAN VOWEL SIGN AA
# ================================================
# Indic_Syllabic_Category=Tone_Mark
1CED       ; Tone_Mark      # Mn       VEDIC SIGN TIRYAK
A982       ; Tone_Mark      # Mn       JAVANESE SIGN LAYAR
# ================================================
# Indic_Syllabic_Category=Consonant_Dead
0F7F       ; Consonant_Dead # Mc       TIBETAN SIGN RNAM BCAD
# ================================================
# Indic_Syllabic_Category=Symbol_Combining
1B6B..1B73 ; Symbol_Combining # Mn [9] BALINESE MUSICAL SYMBOL COMBINING TEGEH..BALINESE MUSICAL SYMBOL COMBINING GONG
# ================================================
# Indic_Syllabic_Category=Cantillation_Mark
1CE2..1CE8 ; Cantillation_Mark # Mn [7] VEDIC SIGN VISARGA SVARITA..VEDIC SIGN VISARGA ANUDATTA WITH TAIL
# ================================================
# Indic_Syllabic_Category=Consonant
A8F2..A8F7 ; Consonant      # Lo [6]   DEVANAGARI SIGN SPACING CANDRABINDU..DEVANAGARI SIGN CANDRABINDU AVAGRAHA
A8FB       ; Consonant      # Lo       DEVANAGARI HEADSTROKE
A8FD       ; Consonant      # Lo       DEVANAGARI JAIN OM

# Indic_Positional_Category=Top
0F74       ; Top     # Mn      TIBETAN VOWEL SIGN U
1A18       ; Top     # Mn      BUGINESE VOWEL SIGN U
1B6B       ; Top     # Mn      BALINESE MUSICAL SYMBOL COMBINING TEGEH
1B6D..1B73 ; Top     # Mn  [7] BALINESE MUSICAL SYMBOL COMBINING KEMPUL..BALINESE MUSICAL SYMBOL COMBINING GONG
# ================================================
# Indic_Positional_Category=Bottom
0F72       ; Bottom  # Mn      TIBETAN VOWEL SIGN I
[...]

PR #982, opened 2018-04-13
Order Chakma split vowels in accordance with NFC

dscorbett:
U+1112E CHAKMA VOWEL SIGN O and U+1112F CHAKMA VOWEL SIGN AU decompose to a below-base vowel sign followed by an above-base vowel sign, the opposite of the usual USE order. Currently, HarfBuzz reverses the decomposition, which is not a complete solution: it causes a dotted to be inserted when the input buffer is in NFD. My fix recategorizes all the below-base Chakma vowels as above-base and vice versa, as in Tibetan.


punchcutter:
This doesn't make any sense to me and breaks more than it fixes. What needs to be done is to update gen-use-table.py with the overrides listed at https://docs.microsoft.com/en-us/typography/script-development/use#overrides-to-indicpositionalcategory
I thought we had already done that, but I guess not yet. Also, like Behdad mentioned at https://github.com/harfbuzz/harfbuzz/issues/539#issue-258544458 we should move the overrides to data files.

dscorbett:
You are right: that broke U+11134 CHAKMA MAAYYAA followed by a below-base vowel. I have fixed that.

I know about the USE’s override list (that is what I was referring to when I mentioned Tibetan): it overrides all Chakma vowels to be below the base, which means that both U+11131 U+11127 and U+11131 U+11127 are valid, even though they look identical. It is preferable to avoid ambiguity by enforcing one canonical order, so I recommend not following the USE list exactly.

PR #991, opened 2018-04-17
Enable Windows-1256 private shaping on other platforms

ebraminio:
Not complete yet (and not completely confident that I am able to), but I guess it is the starting point and I like to see llvm-gcc opinion on it also to be honest.


ebraminio:
@khaledhosny @behdad I've decrypted (!) the hand-coded DSL using compiler preprocessor but still need your help I guess.

Issue #956, opened 2018-04-05
No distinction between failed shaper and nonexistent shaper


ryandesign:
Hello, with harfbuzz 1.7.6 on macOS I have:

```
$ hb-view --list-shapers
graphite2
coretext_aat
ot
coretext
fallback
```

There isn't any difference in output between a shaper that exists but doesn't work:

```
$ hb-view /System/Library/Fonts/LucidaGrande.ttc ABC --shaper=graphite2; echo $?
hb-view: all shapers failed.
1
```

...and a shaper that doesn't exist at all:

```
$ hb-view /System/Library/Fonts/LucidaGrande.ttc ABC --shaper=foobar; echo $?
hb-view: all shapers failed.
1
```

This has caused me to have to be very careful in my recent testing, and double-check all of my typing, to make sure I wasn't inadvertently mistyping a shaper name. I had expected to see a message like "shaper foobar doesn't exist".

behdad:
I suppose we can add code in hb-shape / hb-view to do that just to help user.

PR #993, opened 2018-04-17
[util] Check all specified shapers are known

khaledhosny:
A bit brute force and requires all shapers to be known, not just one.

Fixes https://github.com/harfbuzz/harfbuzz/issues/956


behdad:
Ok let's get this in and see if anyone complains.

PR #994, opened 2018-04-17
Do sanitization before the use on 'main' mini-program

ebraminio:


ebraminio:
Merged early to adopt #926 with it

PR #999, opened 2018-04-19
 [subset] Check that instruction offsets in glyph are in bounds

garretrieger:
Also add a drop hints subsetting operation to the subset fuzzer.


src/hb-ot-glyf-table.hh
```diff
@@ -433,9 +433,23 @@ struct glyf
...
+	unsigned int start = instruction_length_offset + 2;
+	unsigned int end = start + (uint16_t) instruction_length;
+	if (unlikely (end > end_offset)) // Out of bounds of the current glyph
+	{
+	  DEBUG_MSG(SUBSET, nullptr, "The instructions array overruns the glyph's boundaries.");
+	  return false;
+	}
+
+	*instruction_start = start;
+        *instruction_end = end;
```
khaledhosny: The indentation is inconsistent in this change.

PR #1002, opened 2018-04-20
Add hb_face_get_count, a new API

ebraminio:
To be used on hb-sanitize (#950). Does that have a correct name?


TODO
```diff
@@ -27,8 +27,6 @@ API additions
 
 - Add query / enumeration API for aalt-like features?
 
-- SFNT api? get_num_faces?
```
ebraminio: Do we another things like `ot.get_tag` maybe? That particular one is just a detail I guess

src/hb-face.cc
```diff
@@ -178,10 +178,34 @@ hb_face_create (hb_blob_t    *blob,
...
+{
+  if (unlikely (!blob))
+    return 0;
+
+  hb_blob_t *sanitized = OT::Sanitizer<OT::OpenTypeFontFile> ().sanitize (blob);
+  const char *base = hb_blob_get_data (sanitized, nullptr);
+  if (unlikely (!base))
+    return 0;
+
+  return OT::CastP<OT::OpenTypeFontFile> (base)->get_face_count ();
```
behdad: No cast. Use the sanitizer LockInstance.

behdad:
I see how `hb_face_get_empty()` is badly named...  If doing again I would have renamed `hb_face_empty()`

TODO
behdad: No. We're good now that there's get_table_tags and this.

src/hb-face.cc
ebraminio: Done

src/main.cc
```diff
@@ -54,17 +54,14 @@ main (int argc, char **argv)
...
 
-  Sanitizer<OpenTypeFontFile> sanitizer;
-  hb_blob_t *font_blob = sanitizer.sanitize (blob);
-  const OpenTypeFontFile* sanitized = Sanitizer<OpenTypeFontFile>::lock_instance (font_blob);
-  if (sanitized == &Null (OpenTypeFontFile))
+  unsigned int num_fonts = hb_face_count (blob);
+  if (num_fonts == 0)
   {
-    printf ("Sanitization of the file wasn't successful. Exit");
+    printf ("No face has found on the file, or, sanitization of it wasn't successful. Exit\n");
```
ebraminio: Please see if the wording is OK, thanks.

src/hb-face.cc
```diff
@@ -34,6 +34,31 @@
...
+ **/
+unsigned int
+hb_face_count (hb_blob_t *blob)
+{
+  if (unlikely (!blob))
+    return 0;
+
+  hb_blob_t *sanitized = OT::Sanitizer<OT::OpenTypeFontFile> ().sanitize (blob);
+  const OT::OpenTypeFontFile& ot = *OT::Sanitizer<OT::OpenTypeFontFile>::lock_instance (sanitized);
+  if (unlikely (&ot == &OT::Null (OT::OpenTypeFontFile)))
```
behdad: You should not need this, right? The whole idea of the Null objects is that you don't need to check for them constantly. Null object should already return 0 face count.

ebraminio: Done. I guess it was used to make more sense on previous versions of the patch or not. Thanks

ebraminio:
PTAL

ebraminio:
Not fan of merging API/ABI changes but given its simplicity and is already reviewed I went ahead.

ebraminio:
Please let me know if you see any issue before the release.

behdad:
Does it compile?!

behdad:
lock_instance was moved to blob->as() IIRC.

ebraminio:
Fixed on 8220ef8a54, I should've rebased the change before the merge, my bad.

PR #1021, opened 2018-05-07
Set inline hidden flag only on shared library building

ebraminio:
To avoid need of CMP0063 which is not available on older CMake versions


ebraminio:
We should have a CMake 2.8.0 bot to check the compatibility.

PR #1027, opened 2018-05-10
Make hb_blob_create_from_file more portable

ebraminio:
It makes it compatible with ARMCC which I could have access in a collaboration with @imgtec, thanks!

TBR, @behdad

### DIFF
--- a/src/hb-blob.cc
+++ b/src/hb-blob.cc
@@ -1,5 +1,6 @@
 /*
  * Copyright © 2009  Red Hat, Inc.
+ * Copyright © 2018  Ebrahim Byagowi
  *
  *  This is part of HarfBuzz, a text shaping library.
  *
@@ -43,10 +44,6 @@
 #include <stdio.h>
 #include <errno.h>
 #include <stdlib.h>
-#include <sys/types.h>
-#include <sys/stat.h>
-#include <fcntl.h>
-
 
 
 /**
@@ -484,6 +481,12 @@ hb_blob_t::try_make_writable (void)
  * Mmap
  */
 
+#ifdef HAVE_MMAP
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#endif
+
 #if defined(_WIN32) || defined(__CYGWIN__)
 #include <windows.h>
 #include <io.h>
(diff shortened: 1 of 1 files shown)


src/hb-blob.cc
```diff
@@ -545,53 +548,115 @@ hb_blob_create_from_file (const char *file_name)
...
+  rewind (fd);
+  file->contents = (char *) malloc (file->length);
+  if (unlikely (!file->contents))
+  {
+    DEBUG_MSG (BLOB, nullptr, "Failed to allocate enough space to read the file");
+    free (file);
+    fclose (fd);
+    return hb_blob_get_empty ();
+  }
+  assert (file->length == fread (file->contents, file->length, 1, fd));
```
ebraminio: Undecided if needs this highly unlikely assertion

src/hb-blob.cc
```diff
@@ -545,53 +548,115 @@ hb_blob_create_from_file (const char *file_name)
...
+  rewind (fd);
+  file->contents = (char *) malloc (file->length);
+  if (unlikely (!file->contents))
+  {
+    DEBUG_MSG (BLOB, nullptr, "Failed to allocate enough space to read the file");
+    free (file);
+    fclose (fd);
+    return hb_blob_get_empty ();
+  }
+  assert (file->length == fread (file->contents, 1, file->length, fd));
```
ebraminio: Undecided if needs this highly unlikely assertion

src/hb-blob.cc
```diff
@@ -545,53 +548,116 @@ hb_blob_create_from_file (const char *file_name)
   // Allison Lortie permission but changed to suit our need.
   bool writable = false;
   hb_memory_mode_t mm = HB_MEMORY_MODE_READONLY_MAY_MAKE_WRITABLE;
+  hb_mapped_file_t *file = (hb_mapped_file_t *) calloc (1, sizeof (hb_mapped_file_t));
+  if (unlikely (!file))
```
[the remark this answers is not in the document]
behdad: It's not about being likely or not. It's about correctness. We ALWAYS check. We don't crash.

src/hb-blob.cc
behdad: This is by no means unlikely. Read can be interrupted for variety of reasons.

behdad:
Can you describe what this actually does?

I don't like the change in error handling style.  The goto style avoids repetition introduced in this change.

src/hb-blob.cc
ebraminio: Thanks, so it should be turned into "if" rather instead assertion or is it enough? (just want to know when to use `assert` also anyway)

ebraminio:
> Can you describe what this actually does?

It makes `hb_blob_create_from_file` more portable by fallback to `fopen`, `fread`, ... family when `mmap` or Win32 API is not available.

We had a third implementation before this but it still relied on `open` which is not available on ARMCC. Interestingly we could use `open` on Windows but it is emulated by MSVCRT so better to split the path from `mmap` now we have a complete separated path for no `mmap` code.

> I don't like the change in error handling style. The goto style avoids repetition introduced in this change.

I know but the problem each part (`mmap`/Win32/`fopen`) use fairly different cleaner and not much to gain by continuing to use the pattern :/

behdad:
> Can you describe what this actually does?
> 
> It makes hb_blob_create_from_file more portable by fallback to fopen, fread, ... family when mmap or Win32 API is not available.
>
> We had a third implementation before this but it still relied on open which is not available on ARMCC. Interestingly we could use open on Windows but it is emulated by MSVCRT so better to split the path from mmap now we have a complete separated path for no mmap code.

This ARMCC thing, is it actually useful? We don't support non-Unix-like-non-MSVC platforms... Though I'm fine with adding a POSIX path.

Also, what you said above, should be in commit message.
 
>     I don't like the change in error handling style. The goto style avoids repetition introduced in this change.
> 
> I know but the problem each part (mmap/Win32/fopen) use fairly different cleaner and not much to gain by continuing to use the pattern :/

I'm sure there's a middle-ground.  Three different goto targets?

src/hb-blob.cc
behdad: Correct.

`assert` should be used to check program logic invariants. Eg. that ref_count > 0... if it fails, it's a programmer error.

For fread, there's a variety of reasons it might return short. From the man page:
```
       On  success,  fread()  and  fwrite() return the number of items read or
       written.  This number equals the number of bytes transferred only  when
       size  is 1.  If an error occurs, or the end of the file is reached, the
       return value is a short item count (or zero).

       fread() does not distinguish between end-of-file and error, and callers
       must use feof(3) and ferror(3) to determine which occurred.
```

That's just input error. We should not crash on it.

ebraminio:
> This ARMCC thing, is it actually useful?

An embedded device is created by @imgtec to display text which is working and uses freetype and this function was the only blocker to port harfbuzz there so I was requested to help :)

> I'm sure there's a middle-ground. Three different goto targets?

Ok. Have a look

src/hb-blob.cc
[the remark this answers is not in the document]
ebraminio: I feared being too defensive, glad that harfbuzz is that kind of project :)

src/hb-blob.cc
```diff
@@ -545,53 +548,116 @@ hb_blob_create_from_file (const char *file_name)
...
+  }
 
+#ifdef HAVE_MMAP
   int fd = open (file_name, (writable ? O_RDWR : O_RDONLY) | _O_BINARY, 0);
-  if (unlikely (fd == -1)) return hb_blob_get_empty ();
-
-  hb_mapped_file_t *file = (hb_mapped_file_t *) calloc (1, sizeof (hb_mapped_file_t));
+  if (unlikely (fd == -1))
+  {
+    DEBUG_MSG (BLOB, nullptr, "Failed to open the file (mmap)");
```
ebraminio: Can I pass something else the `DEBUG_MSG`? Or these debug messages are needed anyway?

behdad: Not really needed.

ebraminio: Gone, have a look

Issue #1035, opened 2018-05-21
Bhaiksuki and Grantha USE overrides


punchcutter:
For Bhaiksuki the two gap fillers U+11C44 and U+11C45 should be treated like letters so they can allow a following visarga. Here's manuscript and Noto as reference:
![bhks_gapfiller_manuscript](https://user-images.githubusercontent.com/11257204/40296001-402fd20a-5c90-11e8-825d-6c5139d22b19.png)
![bhks_gapfiller2_noto](https://user-images.githubusercontent.com/11257204/40295975-2d4abd30-5c90-11e8-8953-a7b2ec6f6237.png)

For Grantha the anusvara U+11302 currently has a USE category VMPst, but it needs to allow Vedic marks like udatta U+0951 and anudatta U+0952 to follow. Those are VMAbv and VMBlw. It also needs to allow the Sama Vedic svaras which are all VMAbv so I suggest setting the anusvara to VMAbv as well.

ebraminio:
Let me see if I can improve this :)

Can you put the exact text also?

punchcutter:
For Bhaiksuki the string above is: 𑰧𑱅𑱅𑰾𑱁 11C27,11C45,11C45,11C3E,11C41
You can use Noto Bhaiksuki to see that work. 
Grantha with anusvara and Vedic mark:𑌷𑌂𑍧 11337,11302,11367 or 𑌕𑍍𑌷𑌂॒ 11315,1134D,11337,11302,0952
Noto Grantha isn't out yet to test that, but I verified setting U+11302 to VMAbv works. I updated hb-ot-shape-complex-use-table.cc and tested. I was going to just send a pull request updating gen-use-table.py, but as Behdad says in https://github.com/harfbuzz/harfbuzz/issues/539 these overrides should really be put into data files.

ebraminio:
Ok. I see that changing 11C3E from VMPst to either of {H,B,O} fixes this. So should we just wait for the next unicode update or is there anything we need to do here?

dscorbett:
The fix for Bhaiksuki should not be to change U+11C3E (the visarga) but to change the two gap fillers to `Indic_Syllabic_Category=Consonant_Placeholder`.

punchcutter:
Yes, U+11C3E is not the problem. I submitted a pull request. Some tests configs seem to be failing, but that seems unrelated.

ebraminio:
Ok so. Thanks!

PR #1037, opened 2018-05-25
USE overrides for Bhaiksuki Gap Fillers and Grantha Anusvara

punchcutter:
Addresses https://github.com/harfbuzz/harfbuzz/issues/1035

### DIFF
--- a/src/gen-use-table.py
+++ b/src/gen-use-table.py
@@ -47,6 +47,9 @@ defaults = ('Other', 'Not_Applicable', 'Cn', 'No_Block')
 data[0][0x034F] = defaults[0]
 data[0][0x2060] = defaults[0]
 data[0][0x20F0] = defaults[0]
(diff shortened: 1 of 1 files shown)


src/gen-use-table.py
```diff
@@ -47,6 +47,8 @@
 data[0][0x034F] = defaults[0]
 data[0][0x2060] = defaults[0]
 data[0][0x20F0] = defaults[0]
+data[0][0x11C44] = 'Consonant'
+data[0][0x11C45] = 'Consonant'
```
behdad: Shouldn't these be `Consonant_Placeholder`?

behdad: Please add github URL in comments as well.
Also, best if you also file a bug against https://github.com/roozbehp/unicode-data as well.

punchcutter: Absolutely. `Consonant_Placeholder` makes more sense. I'll update and add URL.

PR #1047, opened 2018-06-06
[subset] Add memoization of GSUB lookup closures.

garretrieger:


behdad:
Thanks!  The memoize with checking number of glyphs makes sense. But does the Python code actually do that too? I don't remember implementing it that way.

garretrieger:
No, it keys the memoization on (lookup id, glyph set), see: https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/subset/__init__.py#L1156.

If we assume that the glyph set is only ever added to then using the size of the set to uniquely identify the set should be fine.

behdad:
> No, it keys the memoization on (lookup id, glyph set), see: https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/subset/__init__.py#L1156.

Oh. That's actually something different. `cur_glyphs` is not current set of collected glyphs. It's the (smaller) set of glyphs at current-position, when lookup is applied. This is used to implement the finer-grained closure that I mentioned is not implemented in HarfBuzz.

The glyph-set size thing you implemented looks like might optimize things slightly, but not necessary I think.

behdad:
I should add a has() to set implementation so you can do it like Python.

In the mean time, I suggest you add the set size thing and use a set for done lookups.

behdad:
Or maybe instead of memoizing on 'lookup,cur_glyphs_set`, we can map lookup->cur_glyphs_set which should be better.  I can actually make the Python do that as well, if I felt like touching it...

PR #1048, opened 2018-06-06
Convert Consonant_Initial_Postfixed to CONS_FINAL

dscorbett:
Consonant_Initial_Postfixed was split off of Consonant_Succeeding_Repha, so it should correspond to the same USE class, CONS_FINAL.


behdad:
Thanks. The documentation on the file is not useful.

PR #1049, opened 2018-06-06
[subset] add a new closure call to hb-ot-layout that can compute the …

garretrieger:
…closure over multiple lookups.

### DIFF
--- a/src/hb-ot-layout.cc
+++ b/src/hb-ot-layout.cc
@@ -952,6 +952,32 @@ hb_ot_layout_lookup_substitute_closure (hb_face_t    *face,
   l.closure (&c, lookup_index);
 }
 
+/**
+ * hb_ot_layout_lookups_substitute_closure:
+ *
+ * Compute the transitive closure of glyphs needed for all of the
+ * provided lookups.
+ *
+ * Since: 1.8.1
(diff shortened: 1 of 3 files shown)


src/hb-ot-layout.cc
```diff
@@ -952,6 +952,32 @@ hb_ot_layout_lookup_substitute_closure (hb_face_t    *face,
...
+ * provided lookups.
+ *
+ * Since: 1.8.1
+ **/
+void
+hb_ot_layout_lookups_substitute_closure (hb_face_t      *face,
+                                         const hb_set_t *lookups,
+                                         hb_set_t       *glyphs)
+{
+  hb_auto_t<hb_map_t> done_lookups;
```
behdad: I believe done_lookups should be cleared or reinitialized withiin the while loop. No?

garretrieger: That's what I was thinking initially too but I don't believe it's necessary. We have a function closure(lookup, glyph set) -> glyph set'. Given the same input set, glyph set' should be always be the same. So therefore it's not harmful to re-use the existing done_lookups. Which saves from having to create a new map (or clear the existing one) on each loop.

PR #1050, opened 2018-06-07
[subset] Add gsub closure.

garretrieger:


src/hb-subset-input.cc
```diff
@@ -117,3 +118,19 @@ hb_subset_input_drop_hints (hb_subset_input_t *subset_input)
...
+
+/**
+ * hb_subset_input_drop_ot_layout:
+ * @subset_input: a subset_input.
+ *
+ * If enabled ot layout tables will be dropped as part of
+ * the subsetting operation. Currently this defaults to
+ * true.
+ *
+ * Since: 1.8.0
```
ebraminio: You can use REPLACEME here instead

garretrieger: Thanks, done.

src/hb-subset-input.cc
```diff
@@ -117,3 +118,19 @@ hb_subset_input_drop_hints (hb_subset_input_t *subset_input)
...
+
+/**
+ * hb_subset_input_drop_ot_layout:
+ * @subset_input: a subset_input.
+ *
+ * If enabled ot layout tables will be dropped as part of
+ * the subsetting operation. Currently this defaults to
+ * true.
+ *
+ * REPLACEME
```
ebraminio: `Since: REPLACE` would be better for the next time :)

garretrieger: Fixed.

src/hb-subset-input.cc
```diff
@@ -45,6 +45,7 @@ hb_subset_input_create_or_fail (void)
 
   input->unicodes = hb_set_create ();
   input->glyphs = hb_set_create ();
+  input->drop_ot_layout = true;
```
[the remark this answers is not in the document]
ebraminio: Or, the bot is drunk! (I don't see this reasonable)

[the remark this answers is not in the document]
garretrieger: Agreed.

Issue #630, opened 2017-11-30
1.7.1 Compilation errors on Solaris with SunStudio C++ 12.4


prrace:
There were several issues here and I made source code changes .. probably
all of these could be wrapped in  #ifdef __SUNPRO_CC if so desired.

1)
harfbuzz/hb-face.cc", line 493: Error: The operation "extern "C" void(*)(void*) != void(*)(void*)" is illegal.

The line it did not like is
  if (face->destroy != _hb_face_for_data_closure_destroy)

Quite why that comparison is considered illegal is not apparent to me.
I don't know what the elegant solution is here. I simply hacked it to be
  if ((void*)(face->destroy) != (void*)_hb_face_for_data_closure_destroy)

and the compiler was happy, but I don't think of this as a long term solution.

2)
hb-dsalgs.hh", line 80: Error: Badly formed expression.
hb-dsalgs.hh", line 82: Error: Use ";" to terminate declarations.
hb-dsalgs.hh", line 82: Error: "," expected instead of ")".
hb-dsalgs.hh", line 82: Error: Use ";" to terminate declarations.
hb-dsalgs.hh", line 82: Error: A declaration was expected instead of ",".
hb-dsalgs.hh", line 83: Error: "," expected instead of ")".
hb-dsalgs.hh", line 85: Error: Use ";" to terminate declarations.
hb-dsalgs.hh", line 85: Error: A declaration was expected instead of ",".
hb-dsalgs.hh", line 85: Error: a is not defined.
hb-dsalgs.hh", line 85: Error: w is not defined.
hb-dsalgs.hh", line 86: Error: A declaration was expected instead of "if".
hb-dsalgs.hh", line 86: Error: a is not defined.

It turns out that the SunStudio compiler does not support "restrict" so this line was not parsed correctly :
static int sort_r_cmpswap(char *__restrict a, char *__restrict b, size_t w,

I changed it to
#ifndef __SUNPRO_CC
/* __restrict is same as restrict but better support on old machines */
static int sort_r_cmpswap(char *__restrict a, char *__restrict b, size_t w,
#else
static int sort_r_cmpswap(char *a, char *b, size_t w,
#endif

We lose the verification of __restrict but at least it compiles.

3) hb-string-array.hh", line 51: Error: An array cannot have zero size unless you use the option -features=zla.

That turns out to be the union member str in this code :
#undef _S
  } st;
  char str[0];

Since you need 'str' just to get the offset of the beginning of the array that is the other
[...]

PR #1052, opened 2018-06-08
Fix SunStudio 12.6 build

prrace:
This should resolve the outstanding issues with https://github.com/harfbuzz/harfbuzz/issues/630


prrace:
Need to fix typo first

PR #1053, opened 2018-06-08
Fix SunStudio 12.6 build

prrace:
This should resolve the outstanding issues with https://github.com/harfbuzz/harfbuzz/issues/630.

Testing build with pull request.


ebraminio:
We have ODS12.6 on our CI so I guess you meant 12.4 but anyway we are OK with it :)

prrace:
I really do mean SunStudio 12.6, and I don't know why SunStudio 12.6 for SPARC is different
than the ODS 12.6 you are using ..

Issue #1054, opened 2018-06-12
hb_set_intersect is broken when the number of pages in the hb_set is reduced


jfkthame:
Noticed as a regression in 1.7.7/1.8.0 because of 33d6f46bbbb4fdee110f1cb024a37ed2e46e5e22, which made it much more likely to occur, but in fact the code was already broken prior to that; it just would have required a larger testcase to expose the bug.

In hb_set_t::process, the call to `resize (count)` may shrink the `pages` and `page_map` vectors in the case where `Op::passthru_left` was `false`; but if this happens, the trailing entries in those vectors will become invalid *before* the backward-iterating loop has had a chance to read/copy them.

In the case where these vectors are shrinking, we must defer the resize until after the "Process in-place backward" loop has done its work.

behdad:
Gah. I'm embaraced by how many bugs I introduced by implementing a set from scratch. But at least this one is one of the language's design flaws, not mine...

Thanks for debugging. Looking into it.  Is there a simple test case you can add to test-set.c perhaps?

behdad:
Err. I take that back about being language flaw. Obviously just a bug. I thought you meant the resize can reallocate and make existing pointers invalid. Apparently that's not what you meant.  Never mind.

jfkthame:
We saw failures when we tried to update to 1.8.0 in Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=1467108); investigating that led to the testcase here.

Although we don't actually shrink the vector's allocated space, trying to access the original last element (`pages[na-1]` or `page_map[na-1]`) will fail because `operator[]` validates the index and returns `Crap (Type)` for out-of-bounds.

behdad:
Ah ok so it was the out-of-bounds check that exposed the bug. Glad it did.  Thanks.

PR #1055, opened 2018-06-12
Fix hb_set_intersect bug when shrinking vectors

jfkthame:
See #1054.

### DIFF
--- a/src/hb-set-private.hh
+++ b/src/hb-set-private.hh
@@ -424,7 +424,7 @@ struct hb_set_t
     unsigned int nb = other->pages.len;
     unsigned int next_page = na;
 
-    unsigned int count = 0;
+    unsigned int count = 0, newCount = 0;
     unsigned int a = 0, b = 0;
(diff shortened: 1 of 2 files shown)


behdad:
Thanks. Just came up with this before you submitted PR:
```patch
diff --git a/src/hb-set-private.hh b/src/hb-set-private.hh
index ed753e68..1a5cff48 100644
--- a/src/hb-set-private.hh
+++ b/src/hb-set-private.hh
@@ -452,7 +452,8 @@ struct hb_set_t
     if (Op::passthru_right)
       count += nb - b;
 
-    if (!resize (count))
+    unsigned int final_count = count;
+    if (count > na && !resize (count))
       return;
 
     /* Process in-place backward. */
@@ -506,6 +507,8 @@ struct hb_set_t
        page_at (count).v = other->page_at (b).v;
       }
     assert (!count);
+
+    resize (final_count);
   }
 
   inline void union_ (const hb_set_t *other)
```

PR #1059, opened 2018-06-17
Verbose fail when something is wrong with hb-shape/hb-view input font file

### LINKED ISSUE #1058: Calling hb-shape/hb-view with non-existent font no longer fail
Calling `hb-shape` or `hb-view` with non-existent font used to give an error as expected, but not anymore:
```
$ hb-shape foo bar
[gid0=0+1000|gid0=1+1000|gid0=2+1000]
```

ebraminio:
Fixes #1058 

In addition to check whether it exists or isn't empty or checks if the input file is a font, which is something not used to happen before.

### DIFF
--- a/util/options.cc
+++ b/util/options.cc
@@ -671,6 +671,17 @@ font_options_t::get_font (void) const
     blob = hb_blob_create_from_file (font_file);
   }
 
+  if (hb_blob_get_length (blob) == 0)
+    fail (false, "No such file or directory, or is empty");
+
+  unsigned int face_count = hb_face_count (blob);
+
+  if (face_count == 0)
+    fail (false, "Not a font file"); // most likely
+
+  if (face_index > face_count)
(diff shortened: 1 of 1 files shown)


behdad:
> In addition to check whether it exists or isn't empty

Humm. I'm leaning towards changing the API to return NULL if file does not exist. Or, better, check if `blob == hb_blob_get_empty ()`, instead of checking length being zero. On the creation side, obviously, make it return nil blob only on error.

> or checks if the input file is a font, which is something not used to happen before.

That was intentional.  Try using the tools on Type1 or other non-SFNT-based font files. They used to work with FreeType font-funcs. That was intentional. Even if HarfBuzz doesn't understand the font file, the font functions allow for simple shaping.  That's why Pango, others, don't have to special-case OpenType vs other font formats as long as FreeType+HarfBuzz are used together.  Please revert the check and TTC index check.

PR #1060, opened 2018-06-17
Fix reading fonts from stdin

### LINKED ISSUE #1001: hb-{view,shape} should support reading font file from stdin
It should both support, `hb-{shape,view} {/dev/stdin,-} <font.ttf` and `cat font.ttf | hb-{shape,view} {/dev/stdin,-}`. More nice if it uses `hb_blob_create_from_file` underneath. It should be tested on both Linux and macOS (their implementation differs), and even better, Windows also (\r thing). AIX doesn't have `/dev/stdin` (not the most used platform, but well) so we should have some workaround for that in mind also (specially not redirecting `-` to `/dev/stdin`)

ebraminio:
Fixes #1001 

We were passing the font path directly to freetype so rendering
was broken when we are getting the font from stdin.

This fixes it by using FT_New_Memory_Face instead.

This fixes:
* build/util/hb-view /dev/stdin text < font.ttf
* build/util/hb-view - text < font.ttf
* cat font.ttf | build/util/hb-view - text

but this doesn't work on
* cat font.ttf | build/util/hb-view /dev/stdin text

Which I will try to fix separately.

### DIFF
--- a/util/helper-cairo.cc
+++ b/util/helper-cairo.cc
@@ -89,10 +89,16 @@ helper_cairo_create_scaled_font (const font_options_t *font_opts)
       atexit (free_ft_library);
 #endif
     }
-    FT_New_Face (ft_library,
-		 font_opts->font_file,
-		 font_opts->face_index,
-		 &ft_face);
+
+    unsigned int blob_length;
+    const char *blob_data = hb_blob_get_data (font_opts->blob, &blob_length);
+
+    if (FT_New_Memory_Face (ft_library,
+			    (const FT_Byte *) blob_data,
+                            blob_length,
+			    font_opts->face_index,
+			    &ft_face))
+      fail (false, "FT_New_Memory_Face fail");
   }
   if (!ft_face)
   {
--- a/util/options.cc
+++ b/util/options.cc
@@ -643,8 +643,6 @@ font_options_t::get_font (void) const
   if (font)
     return font;
 
-  hb_blob_t *blob = nullptr;
-
   /* Create the blob */
   if (!font_file)
     fail (true, "No font file set");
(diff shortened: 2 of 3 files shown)


util/helper-cairo.cc
```diff
@@ -89,10 +89,16 @@ helper_cairo_create_scaled_font (const font_options_t *font_opts)
       atexit (free_ft_library);
 #endif
     }
-    FT_New_Face (ft_library,
-		 font_opts->font_file,
-		 font_opts->face_index,
-		 &ft_face);
+
+    unsigned int blob_length;
+    const char *blob_data = hb_blob_get_data (font_opts->blob, &blob_length);
```
ebraminio: We shouldn't have pass file path directly to the freetype, that was causing shaping with one font and rendering with another, the weirdness you were describing Behdad.

util/options.cc
```diff
@@ -663,8 +661,9 @@ font_options_t::get_font (void) const
 	      strerror (errno));
       g_string_append_len (gs, buf, ret);
     }
+    unsigned int len = gs->len;
     char *font_data = g_string_free (gs, false);
-    blob = hb_blob_create (font_data, gs->len,
```
ebraminio: My fault, as it was broken, I didn't notice I broke it even more.

util/options.hh
```diff
@@ -471,6 +472,7 @@ struct font_options_t : option_group_t
   hb_font_t *get_font (void) const;
 
   char *font_file;
+  mutable hb_blob_t *blob;
```
ebraminio: Have a look if is needed.

util/options.hh
```diff
@@ -483,7 +485,7 @@ struct font_options_t : option_group_t
   mutable double font_size_y;
   char *font_funcs;
 
-  private:
+private:
```
behdad: We keep private:, etc, indented.

util/helper-cairo.cc
behdad: Right. Thanks.

PR #1061, opened 2018-06-17
Support pipe and friends on hb_blob_create_from_file

ebraminio:
With this hb-shape/hb-view support `cat font.ttf | build/util/hb-view /dev/stdin text` also.

Please have a look and see what can be improved on c53697d3 and aa0c5df41 also (I merged those first for more convince on working on this but like to get feedback on them).

Another possible solution for fixing this is to use third path of `hb_blob_create_from_file` which uses ftell/fopen/fread but currently unable to read from /dev/stdin when piped. In order to make `/dev/stdin` work there I should make that path just like our [current stdin reader](https://github.com/harfbuzz/harfbuzz/blob/c53697d3f2a/util/options.cc#L651-L668) (and possibly use hb_array_t also?) but as we can't turn `-` to `/dev/stdin` on Windows and AIX as the lack of support, I guess that doesn't worth the hassle anyway.

### DIFF
--- a/src/hb-blob.cc
+++ b/src/hb-blob.cc
@@ -499,6 +499,36 @@ hb_blob_t::try_make_writable (void)
 # define MAP_NORESERVE 0
 #endif
 
+static hb_blob_t *
+_fread_to_end (FILE *file)
+{
+  int allocated = BUFSIZ * 16;
+  char *blob = (char *) malloc (allocated);
+  if (blob == nullptr) return hb_blob_get_empty ();
+
+  int len = 0;
+  while (!feof (file))
+  {
+    if (allocated - len < BUFSIZ)
+    {
+      allocated *= 2;
+      char *new_blob = (char *) realloc (blob, allocated);
+      if (unlikely (!new_blob)) goto fail;
+      blob = new_blob;
+    }
+
+    len += fread (blob + len, 1, allocated - len, file);
+    if (ferror (file)) goto fail;
+  }
+
+  return hb_blob_create (blob, len, HB_MEMORY_MODE_WRITABLE, blob,
+			 (hb_destroy_func_t) free);
+
+fail:
+  free (blob);
+  return hb_blob_get_empty ();
+}
+
 struct hb_mapped_file_t
 {
   char *contents;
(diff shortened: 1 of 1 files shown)


util/options.cc
```diff
@@ -647,7 +647,9 @@ font_options_t::get_font (void) const
   if (!font_file)
     fail (true, "No font file set");
 
-  if (0 == strcmp (font_file, "-")) {
+  if (0 == strcmp (font_file, "-") ||
+      0 == strcmp (font_file, "/dev/stdin") ||
+      0 == strcasecmp (font_file, "STDIN")) {
```
ebraminio: Not sure to support this always but as `stdin` and `STDIN` are both valid on DOS, I used `strcasecmp` instead.

util/options.cc
```diff
@@ -647,7 +647,10 @@ font_options_t::get_font (void) const
   if (!font_file)
     fail (true, "No font file set");
 
-  if (0 == strcmp (font_file, "-")) {
+  if (0 == strcmp (font_file, "-") ||
+      0 == strcmp (font_file, "/dev/stdin") ||
+      0 == strcmp (font_file, "stdin") ||
+      0 == strcmp (font_file, "STDIN")) {
```
ebraminio: Not sure to support this always or behind ifdef macros but as `stdin` and `STDIN` are both valid on DOS, I used `strcasecmp` instead.

behdad:
Two questions:

1. Why do you / we care about stdin and STDIN?  Anyone still uses DOS? :D 

2. Can't we use fcntl or something to detect such strings and handle them correctly (eg. reading from a named pipe as well), instead of hardcoding the names? The '-' is special in that we want to assign a different meaning to it than is otherwise meant. Special-casing "/dev/stdin" is not the same; it's just special-casing a pipe. We should know how to handle all pipes.  Checking for mmap() failure and reading instead sounds right to me.

ebraminio:
> Why do you / we care about stdin and STDIN? Anyone still uses DOS? :D

Hehe, ok. With the new version of the patch we can still possibly support STDIN also but I don't have a working machine with Windows now to add the support.

> Can't we use fcntl or something to detect such strings and handle them correctly (eg. reading from a named pipe as well), instead of hardcoding the names? The '-' is special in that we want to assign a different meaning to it than is otherwise meant. Special-casing "/dev/stdin" is not the same; it's just special-casing a pipe. We should know how to handle all pipes. Checking for mmap() failure and reading instead sounds right to me.

Used https://stackoverflow.com/a/28774534

src/hb-blob.cc
```diff
@@ -499,6 +499,43 @@ hb_blob_t::try_make_writable (void)
 # define MAP_NORESERVE 0
 #endif
 
+static hb_blob_t *
+_fread_to_end (FILE *file)
```
ebraminio: I learned from https://codereview.stackexchange.com/q/9421 for this but hardly there is any connection between the two IMO.

ebraminio:
Something feels wrong, apparently we are using no MMAP on most of our bots :/

ebraminio:
I'm going to merge if you don't object with it but the situation doesn't feel good, most of our bots (including Travis) aren't using `mmap` and they were falling back to fread/ftell/fopen anyway so my merge of this with them was causing them to fail. The only bot that wasn't failing was `clang-O3-O0` so I wonder if you locally also have `HAVE_MMAP` (so the recent releases are fine?)?

PR #1065, opened 2018-06-19
Treat - just as /dev/stdin and remove one extra file reader

### LINKED ISSUE #1064: Remove stdin code from util/options.cc
Now that blob_from_file is supposed to handle pipes, we should remove this code:
```C
  if (0 == strcmp (font_file, "-")) { 
    /* read it */ 
    GString *gs = g_string_new (nullptr); 
    char buf[BUFSIZ]; 
#if defined(_WIN32) || defined(__CYGWIN__) 
    setmode (fileno (stdin), O_BINARY); 
#endif 
    while (!feof (stdin)) { 
      size_t ret = fread (buf, 1, sizeof (buf), stdin); 
      if (ferror (stdin)) 
↦       fail (false, "Failed reading font from standard input: %s", 
↦             strerror (errno)); 
      g_string_append_len (gs, buf, ret); 
    } 
    unsigned int len = gs->len; 
    char *font_data = g_string_free (gs, false); 
    blob = hb_blob_create (font_data, len, 
↦       ↦       ↦          HB_MEMORY_MODE_WRITABLE, font_data, 
↦       ↦       ↦          (hb_destroy_func_t) g_free); 
  } else { 
    blob = hb_blob_create_from_file (font_file); 
  } 
```

Also, probably setmode should always be called if available?

ebraminio:
Fixes #1064.

Or some other token would be needed as AIX and Windows issue.

Interestingly fopen or CreateFile didn't support "STDIN" as string at least on my tests on Wine on our current CreateFile configs.

Bad idea? I am out of idea for #1064 so.

### DIFF
--- a/util/options.cc
+++ b/util/options.cc
@@ -647,28 +647,19 @@ font_options_t::get_font (void) const
   if (!font_file)
     fail (true, "No font file set");
 
-  if (0 == strcmp (font_file, "-")) {
-    /* read it */
-    GString *gs = g_string_new (nullptr);
-    char buf[BUFSIZ];
+  const char *font_path = font_file;
+
+  if (0 == strcmp (font_path, "-"))
+  {
 #if defined(_WIN32) || defined(__CYGWIN__)
     setmode (fileno (stdin), O_BINARY);
+    font_path = "STDIN";
+#else
+    font_path = "/dev/stdin";
 #endif
-    while (!feof (stdin)) {
-      size_t ret = fread (buf, 1, sizeof (buf), stdin);
-      if (ferror (stdin))
-	fail (false, "Failed reading font from standard input: %s",
-	      strerror (errno));
-      g_string_append_len (gs, buf, ret);
-    }
-    unsigned int len = gs->len;
-    char *font_data = g_string_free (gs, false);
-    blob = hb_blob_create (font_data, len,
-			   HB_MEMORY_MODE_WRITABLE, font_data,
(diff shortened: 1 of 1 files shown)


behdad:
Treating nullptr as stdin is definitely not an API we like.  What's wrong with just replacing it with /dev/stdin or STDIN or other things based on platform and let the existing machinery handle it?  Alternatively, we can special-case "/dev/stdin" in the open function instead of nullptr. But I rather not havve to special-case anything.

That said, I don't like that then the client would have to special-case what the call the function with to open stdin... Humm. Let me think more about it.

behdad:
Maybe we don't care about handling "-" as stdin?  I'm leaning towards that. Users on Unix-like can always pass `/dev/stdin` manually...

ebraminio:
Interestingly I added dash to node.js https://github.com/nodejs/node/pull/13012 for one special use case I had so it feels weird now to me to remove it from another project! :)

Have a look at the updated version, I redirected `-` to `/dev/stdin` for Unix and to `STDIN` to non-cygwin Windows. It probably doesn't work on Windows (didn't actually tested) and definitely doesn't on AIX but I guess this is good enough to have and much better than removing it altogether. If someone really care about Windows support (which is weird, who is going to pipe a font on Windows?!) they can try to see why STDIN doesn't work on `hb_blob_create_from_file`.

behdad:
Ok thanks.

behdad:
Lgtm.

PR #1068, opened 2018-06-24
Merge two fread readers of hb_blob_create_from_file

### LINKED ISSUE #1063: Merge two fread readers of hb_blob_create_from_file
Simply, these two, [1](https://github.com/harfbuzz/harfbuzz/blob/941f2b85489b/src/hb-blob.cc#L502-L530) [2](https://github.com/harfbuzz/harfbuzz/blob/941f2b85489b/src/hb-blob.cc#L625-L638) having many things in common and should be merged (so we would have a `fread` reader without `ftell` at the end). I tried to do that initially but it revealed most of our bots are relying on `hb_blob_create_from_file`'s `fread` reader and don't have HAVE_MMAP which feels very wrong (and possibly shows just added `fread` reader, `_fread_to_end`, is doing something wrong also).

Filing this as a bug that I can't see any of these locally :/ I can try logging into the bots however but I think I should get some help from others.

ebraminio:
Fixes #1063.

And assign our Alpine CI bot to always use fallback reader.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,14 +40,14 @@ jobs:
       - run: rm -rf harfbuzz-*
       - run: make distdir && cd harfbuzz-* && cmake -DHB_CHECK=ON -Bbuild -H. -GNinja && ninja -Cbuild && CTEST_OUTPUT_ON_FAILURE=1 ninja -Cbuild test && ninja -Cbuild install
 
-  alpine-O3:
+  alpine-O3-NOMMAP:
     docker:
       - image: alpine
     steps:
       - checkout
       - run: apk update && apk add ragel make pkgconfig libtool autoconf automake gettext gcc g++ glib-dev freetype-dev cairo-dev
       # C??FLAGS are not needed for a regular build
-      - run: CFLAGS="-O3" CXXFLAGS="-O3" ./autogen.sh
+      - run: CFLAGS="-O3" CXXFLAGS="-O3 -DNOMMAPFILEREADER" ./autogen.sh
       - run: make
       - run: make check || .ci/fail.sh
 
@@ -196,7 +196,7 @@ workflows:
       - distcheck
 
       # autotools based builds
-      - alpine-O3
+      - alpine-O3-NOMMAP
       - archlinux-debug-O0-py3
       - clang-O3-O0
       - fedora-outoftreebuild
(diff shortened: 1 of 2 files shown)


src/hb-blob.cc
```diff
@@ -621,32 +589,49 @@ hb_blob_create_from_file (const char *file_name)
...
+  if (unlikely (fp == nullptr)) goto fread_fail_without_close;
+
+  while (!feof (fp))
+  {
+    if (allocated - len < BUFSIZ)
+    {
+      allocated *= 2;
+      // Don't allocate more than 200MB, our mmap reader still
+      // can cover files like that but lets limit our fallback reader
+      if (unlikely (allocated > 200000000)) goto fread_fail;
```
ebraminio: So no more fun with /dev/zero /dev/random /dev/urandom yes and friend :)

src/hb-blob.cc
```diff
@@ -568,45 +538,43 @@ hb_blob_create_from_file (const char *file_name)
...
+
+  close (fd);
+
+  return hb_blob_create (file->contents, file->length, mm, (void *) file,
+			 (hb_destroy_func_t) _hb_mapped_file_destroy);
+
+fail:
+  close (fd);
+fail_without_close:
+  free (file);
```
ebraminio: So this will always fallback on failures.

.circleci/config.yml
```diff
@@ -40,14 +40,14 @@ jobs:
...
-  alpine-O3:
+  alpine-O3-NOMMAP:
     docker:
       - image: alpine
     steps:
       - checkout
       - run: apk update && apk add ragel make pkgconfig libtool autoconf automake gettext gcc g++ glib-dev freetype-dev cairo-dev
       # C??FLAGS are not needed for a regular build
-      - run: CFLAGS="-O3" CXXFLAGS="-O3" ./autogen.sh
+      - run: CFLAGS="-O3" CXXFLAGS="-O3 -DNOMMAPFILEREADER" ./autogen.sh
```
ebraminio: Is "NOMMAPFILEREADER" good?

src/hb-blob.cc
ebraminio: Is this a good limitation?

ebraminio:
WDYT Behdad?

.circleci/config.yml
behdad: I prefer always a `HB_` prefix.  I'd go for `HB_NOMMAP`.

src/hb-blob.cc
behdad: For multi-line comments please use C-style.  I personally always use C-style comments for, well, comments. And use C++-style for code that is commented out.

PR #1078, opened 2018-06-26
configure.ac: remove fontconfig configure option

maxinbjohn:
Comment out fontconfig option in configure.ac to normalize the support
in cmake and autotools

Signed-off-by: Maxin B. John <maxin.john@intel.com>


maxinbjohn:
Ok, Thanks for the input. Will redo it.

PR #1079, opened 2018-06-26
Minor touches on hb_blob_create_from_file

ebraminio:
* Handle EINTR on fallback reader
* Increase fallback reader limitation size limitation to 2 << 28
* Ensure _O_BINARY does exist if MMAP is used on Windows
  (maybe superfluous but makes me more confident)


ebraminio:
Anything to be improved here Behdad?

behdad:
Lgtm. Thanks. Calling something other than hb_blob_t a blob is weird, but cosmetic.

Issue #651, opened 2017-12-12
FT_Load_Glyph speedup


molind:
Hello,

We use harfbuzz inside of mobile application Galileo along with SDF renderer. Extremely satisfied by result.

Recently I've looked in profiler to find new bottlenecks and noticed that FT_Load_Glyph causes huge performance slow down on devices with slow internal memory (e.g. iPhone 6). It' takes up to 2 seconds to compute extents and load glyphs information. Mostly because FT_Load_Glyph is slow and calls to FT_Load_Glyph is not cached.

I suggest to use [FreeType cache system](https://www.freetype.org/freetype2/docs/reference/ft2-cache_subsystem.html#FTC_ImageCache_Lookup) instead of reading the same data over and over again. What do you think?

https://github.com/harfbuzz/harfbuzz/blob/20e69c950d431391f92335f8845e5d4291aca8f3/src/hb-ft.cc#L250-L275

https://github.com/harfbuzz/harfbuzz/blob/20e69c950d431391f92335f8845e5d4291aca8f3/src/hb-ft.cc#L294-L322

https://github.com/harfbuzz/harfbuzz/blob/20e69c950d431391f92335f8845e5d4291aca8f3/src/hb-ft.cc#L324-L349

khaledhosny:
Most HarfBuzz clients do some sort of caching so I think you should consider doing the same. Last I tried, just caching `hb_font_t`s resulted in a measurable performance boost.

molind:
@khaledhosny Yes, we do cache `hb_font_t`s. Issue happened when we process some significant amount of text to render e.g. 500 - 5000 words for example. At that moment we have to get glyph extents and other information to draw glyphs correctly.

khaledhosny:
HarfBuzz shouldn’t normally need glyph extents, are you calling `hb_font_get_glyph_extents()` yourself? If so, may be you should cache these.

Anyway, I don’t think we can use Freetype cache API as we don’t create the `FT_Face` ourselves but get it from the client.

destman:
I will try to explain performance problem that we face with more details:
We have 95 fonts that contain all glyphs we need. All this fonts occupy 24MB so instead of loading them into memory we load them from files. In this case FreeType use memory mapping. With new devices all is fine. 

But device with not very fast flash drive have problems with performance.
At our reference test ~4s was spend for text processing. In profiler we find out that our calls of hb_font_get_glyph_extents spend ~2s and shaping spend ~1.9s. 

[...]

PR #1082, opened 2018-06-29
Added caching into hb_ft_get_glyph_h_advance function.

molind:
Added caching mentioned in #651. 

In our tests hb_shape time decreased from 260ms to 10ms. 

I have almost no experience with autoconf, please check if it's correct.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -464,6 +464,18 @@ AM_CONDITIONAL(HAVE_CORETEXT, $have_coretext)
 
 dnl ===========================================================================
 
+AC_ARG_ENABLE(h_advance_cache,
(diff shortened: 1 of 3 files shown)


ebraminio:
Behdad, couldn't we always have this instead introducing a build option?

behdad:
> Behdad, couldn't we always have this instead introducing a build option?

Yep, agree.

behdad:
I need to figure out how to invalidate the cache if font size changes though... That's, complicated.

behdad:
Done.  I also added `--ft-load-flags` to hb-shape and hb-view to allow measuring performance.  Looks good now.

PR #1084, opened 2018-06-30
Improve nommap naming and use C style comments on create_from_file

ebraminio:


behdad:
Really sorry for changing my mind all the time.  Please make it `HB_NO_MMAP` so it's aligned with `HB_NO_MT`.

Issue #1046, opened 2018-06-04
hb-shape with default shaper (not uniscribe) not shaping Sinhala at all on Windows


cjchapman:
Follow up to my previous bug: Behdad and I also noticed that hb-shape with its default OpenType shaper (i.e. not using Uniscribe) seemed to not shape a line of Sinhala at all (using the text & font from https://github.com/harfbuzz/harfbuzz/issues/967) -- i.e. the glyphs were all the nominal glyphs for the characters, not the shaped glyphs.

ebraminio:
On an updated master,
```bash
➜  harfbuzz git:(master) ✗ wine winbuild/util/hb-shape.exe iskpota.ttf -u U+0DBB,U+0DCA,U+200D,U+0DBA,U+0DCA,U+200D,U+0DBA,U+0A,U+0DBB,U+0DCA,U+200D,U+0DBA,U+0DCA,U+200D,U+0DBA
[gid736=0+1343|gid600=0+1130]
[gid736=0+1343|gid600=0+1130]
```

So no idea how to repro.

behdad:
I can't repro using wine either. Issue only exhibited when we tried the binaries on a Windows 10 VM.

ebraminio:
So Sinhala test is working on our CI on both mingw and MSVC. Weird :/

HinTak:
Are you talking about win32 harfbuzz or win64 harfbuzz? I assume the VM is likey 64-bit, but the wine instruction is 32-bit.

khaledhosny:
I can’t reproduce this (pre-built binaries, Windows 10), please re-open if it is still an issue.

PR #1087, opened 2018-07-03
Use CreateFileW explicitly and pass wchar_t to it

### LINKED ISSUE #1086: error: no matching function for call to 'CreateFileW'
Chrome Windows build has this issue:
https://logs.chromium.org/v/?s=chromium%2Fbuildbucket%2Fcr-buildbucket.appspot.com%2F8942035310612106064%2F%2B%2Fsteps%2Fcompile__with_patch_%2F0%2Fstdout

cc @drott

ebraminio:
Fixes #1086 and I guess #1046 also (hopefully)

### DIFF
--- a/src/hb-blob.cc
+++ b/src/hb-blob.cc
@@ -569,9 +569,15 @@ fail_without_close:
   hb_mapped_file_t *file = (hb_mapped_file_t *) calloc (1, sizeof (hb_mapped_file_t));
   if (unlikely (!file)) return hb_blob_get_empty ();
 
-  HANDLE fd = CreateFile (file_name, GENERIC_READ, FILE_SHARE_READ, nullptr,
-			  OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL|FILE_FLAG_OVERLAPPED,
-			  nullptr);
+  HANDLE fd;
+  unsigned int size = strlen (file_name) + 1;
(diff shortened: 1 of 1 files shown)


dscorbett:
`mbstowcs` uses the current locale. Could that unpredictability make this fix ever do the wrong thing?

ebraminio:
> `mbstowcs` uses the current locale. Could that unpredictability make this fix ever do the wrong thing?

:/

We have one here already https://github.com/harfbuzz/harfbuzz/blob/159ddb872986f121818e816d2ea75d271075ba1f/util/options.cc#L152

Worst case I'd guess would be fallback to `fopen` reader.

dscorbett:
On Windows, `_mbstowcs_l` is the locale-sensitive version. I am not sure this is a real problem: [the locale is always initially set to `"C"`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/setlocale-wsetlocale#generic-text-routine-mappings), but maybe a client of HarfBuzz can override that and mess something up. You should confirm with someone who knows C++ locales.

ebraminio:
Ok. Lets merge this and think about our two instance of `mbstowcs` on the project (the other is on hb-directwrite, we is also added by me). At least we have `fopen` around as a fallback for this cases, for now.

ebraminio:
For the reference, I asked my C/C++ expert friend and he was undecided about what to do here but I try to fix this and will apply any solution you will find @dscorbett also. Thank you

PR #1090, opened 2018-07-07
Regenerate the USE table for Grantha and Bhaiksuki

### LINKED ISSUE #1035: Bhaiksuki and Grantha USE overrides

dscorbett:
Completes #1037 and fixes #1035.

This is the danger of keeping generated files in version control. Maybe you could add a CI test to verify that all generated files are up to date.

### DIFF
--- a/src/hb-ot-shape-complex-use-table.cc
+++ b/src/hb-ot-shape-complex-use-table.cc
@@ -560,7 +560,7 @@ static const USE_TABLE_ELEMENT_TYPE use_table[] = {
 
   /* Grantha */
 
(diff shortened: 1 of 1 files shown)


behdad:
When one builds with autotools, the generated files get updated. Is there a cmake issue here?

PR #1091, opened 2018-07-09
gen-def.py: pass headers as arguments so that msys2 can convert posix paths

anthrotype:
When one is not using the msys2 python to run the `gen-def.py` script, the header files that are passed in as environment variable cannot be found:

https://ci.appveyor.com/project/fonttools/ttfautohint-py/build/1.0.65/job/rkremny4jjid9nl2#L803

This is because msys2 shell and GNU make use POSIX paths (e.g. `/c/Users/clupo/...`) whereas non-msys2 python.exe uses native Windows paths (e.g. `C:\Users\clupo\...`).

Msys2 will automatically convert command line arguments (but not environment variables) from POSIX to Windows paths when calling a native win32 executable, so we need to pass the header paths as arguments instead of environment variables.

This way the gen-def.py script can support both mingw python running in an MSYS2 shell, and native win32 python.

This is required for building the embedded libttfautohint.dll (statically linked to libharfbuzz) that ships with https://github.com/fonttools/ttfautohint-py (python bindings for the ttfautohint library).

In the build environment for ttfautohint-py, we must use the regular Windows python.exe instead of the msys2 one, so the gen-def.py script should be able to run on the former as well as the latter.

Here is a build of ttfautohint-py that passes after applying the batch below:
https://ci.appveyor.com/project/fonttools/ttfautohint-py/build/1.0.68/job/ty4jqbncx04289uo

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -836,7 +836,7 @@ if (UNIX OR MINGW)
     # generate harfbuzz.def after build completion
     string(REPLACE ";" " " space_separated_headers "${project_headers}")
     add_custom_command(TARGET harfbuzz POST_BUILD
-      COMMAND ${CMAKE_COMMAND} -E env "headers=${space_separated_headers}" python ${PROJECT_SOURCE_DIR}/src/gen-def.py ${PROJECT_BINARY_DIR}/harfbuzz.def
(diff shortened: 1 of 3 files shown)


anthrotype:
oops, forgot to also fix up the CMakeLists.txt to match the updated Makefile.am (fix coming)

behdad:
Lgtm

anthrotype:
bots are all green, going to merge...

PR #1093, opened 2018-07-10
appveyor: don't install msys2 'python2' package; try using Appveyor python

anthrotype:
follow up from https://github.com/harfbuzz/harfbuzz/commit/bf9e9676dda686f5b76826b4e3148f4a0b512e3c#commitcomment-29653698

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -43,7 +43,7 @@ build_script:
   - 'if "%compiler%"=="msvc" msbuild harfbuzz.sln /p:Configuration=%configuration% /p:Platform=%platform%'
   - 'if "%compiler%"=="msvc" if not "%platform%"=="ARM" ctest --output-on-failure -C %configuration%'
 
-  - 'if "%compiler%"=="msys2" C:\msys64\usr\bin\bash -lc "pacman --noconfirm -S mingw-w64-$MSYS2_ARCH-{freetype,cairo,icu,gettext,gobject-introspection,gcc,gcc-libs,glib2,graphite2,pkg-config,python2}"'
+  - 'if "%compiler%"=="msys2" C:\msys64\usr\bin\bash -lc "pacman --noconfirm -S mingw-w64-$MSYS2_ARCH-{freetype,cairo,icu,gettext,gobject-introspection,gcc,gcc-libs,glib2,graphite2,pkg-config}"'
   - 'if "%compiler%"=="msys2" C:\msys64\usr\bin\bash -lc "curl https://raw.githubusercontent.com/mirror/mingw-w64/023eb04c396d4e8d8fcf604cfababc53dae13398/mingw-w64-headers/include/dwrite_1.h > %MINGW_PREFIX%/%MINGW_CHOST%/include/dwrite_1.h"'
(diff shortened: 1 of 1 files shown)


anthrotype:
@ebraminio I noticed that Appveyor is running 2 builds in parallel. I had never seen that before. Are we on a premium plan, or has something changed lately?

anthrotype:
pacman failed to install some package with a "maximum file size exceeded".. weird:
https://ci.appveyor.com/project/harfbuzz/harfbuzz/build/1.0.1832/job/afkaap7bp1gq77b2#L72

@ebraminio can you try restart the Appveyor build? I can't seem to be able to do it from my own Appveyor account.

anthrotype:
I noticed something.
The test/fuzzing/Makefile.am are calling the fuzzer test scripts with a `builddir="$(builddir)"` variable:

https://github.com/harfbuzz/harfbuzz/blob/bf9e9676dda686f5b76826b4e3148f4a0b512e3c/test/fuzzing/Makefile.am#L84-L85

however, the scripts themselves don't seem to be reading that environment variable; instead they are attempting to get another one with a similar name `top_builddir`:

https://github.com/harfbuzz/harfbuzz/blob/bf9e9676dda686f5b76826b4e3148f4a0b512e3c/test/fuzzing/run-subset-fuzzer-tests.py#L9
https://github.com/harfbuzz/harfbuzz/blob/bf9e9676dda686f5b76826b4e3148f4a0b512e3c/test/fuzzing/run-shape-fuzzer-tests.py#L9

Are they supposed to be using `$builddir` instead?

(git blame says @garretrieger wrote that)

ebraminio:
You shuold be able to by logging out and logging in to harfbuzz account on appveyor but I also did it

ebraminio:
I think I can say feel free to apply any improvements that makes sense and don't break our CI :)

anthrotype:
apparently my user (@anthrotype, email cosimo [at] anthrotype [dot] com) is not part of the `harfbuzz` Appveyor organization, it does not show up in the list when I log in:

image

anthrotype:
the mingw 64-bit build keeps failing, with  
https://ci.appveyor.com/project/harfbuzz/harfbuzz/build/1.0.1834/job/26v56v60egbxj6q5#L66

> failed retrieving file 'mingw-w64-x86_64-python3-markupsafe-1.0-2-any.pkg.tar.xz' from repo.msys2.org : Maximum file size exceeded

the i686 one seems to pass...

I don't understand why, by simply _not_ installing `python2` msys2 package, now it pulls in python3 plus other related deps like that markupsafe.

Maybe it's not worth the effort, let's keep it like it was.

anthrotype:
the pacman dowload failure may only be temporary, it seems connected to this
https://github.com/Alexpux/MSYS2-packages/issues/942#issuecomment-357359208

it'll probably fix itself at some point, we can retry another time.

anthrotype:
@ebraminio it looks like the Appveyor MSYS2 x86_64 build is failing on master, with the same error I encountered here:

> failed retrieving file 'mingw-w64-x86_64-python3-markupsafe-1.0-2-any.pkg.tar.xz' from repo.msys2.org : Maximum file size exceeded

https://ci.appveyor.com/project/harfbuzz/harfbuzz/build/job/ky7lao1ii1bi7ew4#L71

I was able to fix it locally by doing an update of the msys2 package database and core system packages `pacman -Syu`. But then I had to close and reopen the mintty shell, as requested by the warning message printed at the bottom after the upgrade:

> close your terminal window instead of calling exit

Not sure how that can be done in the Appveyor environment though..

ebraminio:
> it does not show up in the list when I log in

I fixed it I guess but forgot to leave a comment :/

Not sure and sad we could do this at the end but thanks for the effort :)

PR #1094, opened 2018-07-16
appveyor.yml: try updating msys2 to fix failing mingw-w64-x86_64 build

anthrotype:
https://github.com/harfbuzz/harfbuzz/pull/1093#issuecomment-405201903
https://ci.appveyor.com/project/harfbuzz/harfbuzz/build/job/ky7lao1ii1bi7ew4#L71

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -45,7 +45,7 @@ build_script:
   - 'if "%compiler%"=="msvc" msbuild harfbuzz.sln /p:Configuration=%configuration% /p:Platform=%platform%'
   - 'if "%compiler%"=="msvc" if not "%platform%"=="ARM" ctest --output-on-failure -C %configuration%'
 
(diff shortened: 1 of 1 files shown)


anthrotype:
looks like it fixes it, though updating msys2 makes the build a bit slower..

anthrotype:
@ebraminio I wonder, have you noticed this error messages (but actually they seem more like warnings since it doesn't exit with failure)

```
libtool:   error: Could not determine the host path corresponding to
libtool:   error:   
'C:/msys64/mingw64/lib:C:/msys64/mingw64/bin:/c/projects/harfbuzz/src/.libs:/mingw64/lib:/mingw64/bin'
libtool:   error: Continuing, but uninstalled executables may not work.
```

they happen when running `make check` from the msys2 builds.
I have no idea what they means, but the test suite seems to pass despite those warnings.

anthrotype:
so, without upgrading the Appveyor's msys2, the i686 and x86_64 builds would take 8-9 minutes each. With this patch, they take 14 minutes each, but at least they pass.

anthrotype:
ok, it's again below 10 minutes now.

PR #1106, opened 2018-07-23
Fix fuzzer timeout.

garretrieger:
Limit the number of iterations of the closure algorithm and don't reprocess glyphs added during single subst closure.

### DIFF
--- a/src/hb-ot-layout-common-private.hh
+++ b/src/hb-ot-layout-common-private.hh
@@ -41,6 +41,13 @@
 #ifndef HB_MAX_CONTEXT_LENGTH
 #define HB_MAX_CONTEXT_LENGTH	64
 #endif
+#ifndef HB_CLOSURE_MAX_STAGES
+/*
+ * The maximum number of times a lookup can be applied during shaping.
+ * Used to limit the number of iterations of the closure algorithm.
+ */
+#define HB_CLOSURE_MAX_STAGES	8
+#endif
 
 
 namespace OT {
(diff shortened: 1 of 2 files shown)


src/hb-ot-layout.h
```diff
@@ -37,6 +37,11 @@
 
 HB_BEGIN_DECLS
 
+/*
+ * The maximum number of times a lookup can be applied during shaping.
+ * Used to limit the number of iterations of the closure algorithm.
+ */
+#define HB_CLOSURE_MAX_STAGES 8
```
behdad: Please move this to `hb-ot-layout-common-private.hh` after the couple of other defines there, and also guard with `ifndef` like them.

src/hb-ot-layout-gsub-table.hh
```diff
@@ -41,14 +41,22 @@ struct SingleSubstFormat1
   {
     TRACE_CLOSURE (this);
     Coverage::Iter iter;
+
+    hb_set_t closure_glyphs;
+    closure_glyphs.init ();
```
behdad: I was actually thinking about doing this at the closure_context_t level (adding separate glyphs and out members) and update all closure implementations.  Can be done separately.  I don't mind doing it myself either.  Probably a swap() method as well that unions the out to the glyphs.  Humm, however, this will increase set operations we do significantly. I doubt it be a problem though.  Feel free to remove this commit so we merge the loop fix.

src/hb-ot-layout.h
garretrieger: Done.

src/hb-ot-layout-gsub-table.hh
garretrieger: Sounds good. I've removed it from this pr.

behdad:
Thanks. Feel free to merge.  Is there a small test we can add?

garretrieger:
Unfortunately there isn't any existing unit tests for the closure operation (or the individual lookup closure calls). I've made a note to myself to add some.

PR #1108, opened 2018-07-24
Closure enhancements

behdad:


src/hb-ot-layout-gsub-table.hh
```diff
@@ -1261,13 +1266,6 @@ struct SubstLookup : Lookup
   template <typename context_t>
   static inline typename context_t::return_t dispatch_recurse_func (context_t *c, unsigned int lookup_index);
 
-  static inline hb_closure_context_t::return_t dispatch_closure_recurse_func (hb_closure_context_t *c, unsigned int lookup_index)
```
garretrieger: I believe this is still needed. My read of the code is that hb_closure_context_t::recurse doesn't call SubstLookup::closure , but instead ends up calling SubstLookup::dispatch directly and bypasses SubstLookup::closure. Thus we need to this method to check should_visit_lookup.

PR #1113, opened 2018-07-26
CFF/CFF2 subsetter

blueshade7:
Index, Dict structs
hooked up to hb-subset (takes CFF2, outputs empty CFF2)


src/hb-ot-cff2-table.hh
```diff
@@ -0,0 +1,386 @@
...
+
+struct cff2
+{
+  static const hb_tag_t tableTag        = HB_OT_TAG_cff2;
+
+  inline bool sanitize (hb_sanitize_context_t *c) const
+  {
+    TRACE_SANITIZE (this);
+    return_trace (c->check_struct (this) &&
+                  version.sanitize (c) &&
```
garretrieger: No need to call sanitize on version since it's of static size. The check_struct above will cover it.

src/hb-ot-cff2-table.hh
```diff
@@ -0,0 +1,386 @@
...
+        CFF2TopDict_Interpreter top_interp;
+        top_interp.init ();
+        top_interp.interpret (topDictStr, top);
+        top_interp.fini ();
+      }
+      
+      varStore = &OT::StructAtOffset<VariationStore>(cff2, top.vstoreOffset);
+      charStrings = &OT::StructAtOffset<CharStrings>(cff2, top.charStringsOffset);
+      fdArray = &OT::StructAtOffset<FDArray>(cff2, top.FDArrayOffset);
+      fdSelect = &OT::StructAtOffset<FDSelect>(cff2, top.FDSelectOffset);
```
garretrieger: As the comment below mentions you need to sanitize these to ensure it's safe to access the memory they point too.

src/hb-ot-cff2-table.hh
```diff
@@ -0,0 +1,386 @@
...
+      const OT::cff2 *cff2 = this->blob->as<OT::cff2> ();
+
+      if (cff2 == &Null(OT::cff2))
+      {
+        hb_blob_destroy (blob);
+        return;
+      }
+
+      { /* parse top dict */
+        ByteStr topDictStr (cff2 + cff2->topDict, cff2->topDictSize);
```
garretrieger: Needs to be sanitized to ensure memory is within the blob.

garretrieger:
The bots are failing with:

"Ouch, header file hb-ot-cff-common-private.hh does not have correct preprocessor guards
FAIL check-header-guards.sh (exit status: 1)"

Please fix.

src/hb-ot-cff2-table.hh
blueshade7: Added a sanitize call

src/hb-ot-cff2-table.hh
blueshade7: Added a sanitize call to each if not Null

src/hb-ot-cff2-table.hh
blueshade7: fixed

src/hb-ot-cff-common-private.hh
```diff
@@ -0,0 +1,475 @@
...
+
+/* CFF INDEX */
+struct Index
+{
+  inline bool sanitize (hb_sanitize_context_t *c) const
+  {
+    TRACE_SANITIZE (this);
+    return_trace (c->check_struct (this) && offSize >= 1 && offSize <= 4 &&
+                  c->check_array (offsets, offSize, count + 1) &&
+                  c->check_array (data_base (), 1, offset_at (count)));
```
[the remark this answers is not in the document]
blueshade7: Added & used a function max_offset as suggested, but later realized better to verify offsets are in the increasing order which should make a better check.

ebraminio:
Please remove "build/*" folder from your patch, thanks.

PR #1116, opened 2018-07-31
Update lookup collection to first collect features

garretrieger:
Then collect lookups from those features. This allows us to use a more optimized add_array call on the feature set.


behdad:
LGTM.

PR #1119, opened 2018-08-01
[subset] limit the max codepoint value to the unicode limit.

garretrieger:
When collecting all codepoints in the cmap avoid using large amount of memory for fonts that declare coverage over all 32 bit integers.


src/hb-ot-cmap-table.hh
```diff
@@ -437,8 +440,10 @@ struct CmapSubtableLongSegmented
   {
     for (unsigned int i = 0; i < this->groups.len; i++) {
       hb_set_add_range (out,
-			this->groups[i].startCharCode,
-			this->groups[i].endCharCode);
+			MIN ((unsigned int) this->groups[i].startCharCode,
+                             (unsigned int) HB_MAX_UNICODE_CODEPOINT_VALUE),
+                        MIN ((unsigned int) this->groups[i].endCharCode,
+                             (unsigned int) HB_MAX_UNICODE_CODEPOINT_VALUE));
```
behdad: Indentation seems off.

garretrieger: Fixed.

PR #1120, opened 2018-08-01
Apply reviews of hb_ot_layout_get_feature_name_ids

ebraminio:
Remained things,
* "What we want however is DEFINE_NULL_DATA for the two FeatureParams types in this change to use 0xFFFF instead of zero bytes." (If I figure out how)
* Bringing back hb_ot_layout_get_feature_characters (If someone create a test font for it)


ebraminio:
> "What we want however is DEFINE_NULL_DATA for the two FeatureParams types in this change to use 0xFFFF instead of zero bytes."

How to detect which one we are dealing with?

ebraminio:
The API should be altered itself https://github.com/harfbuzz/harfbuzz/pull/976#issuecomment-409460172

Issue #1130, opened 2018-08-09
handling _ in feature tags


mhosken:
Currently harfbuzz will not parse a feature which contains _ in its id. This isn't a problem with OT because there are no registered features with _ in the id. But it is a problem for a number of published fonts that use _ in graphite features.

Any chance the ISALNUM(**pp) could be relaxed to include (ISALNUM(**pp) || **pp == "_") in hb_common.cc in parse_tag()?

behdad:
I'm fine with that.  Send a PR.

khaledhosny:
PR was merged.

PR #1132, opened 2018-08-09
Support _ in feature ids

mhosken:
Addresses issue #1130.


mhosken:
The build fails on clang-O3-O0 due to apt problems and nothing to do with this PR.

The build fails on cmake-oracledeveloperstudio due to lack of a library libnsl and again has nothing to do with this PR (by my understanding).

PR #1136, opened 2018-08-14
Guard free_static function definitions with macro

eaenet:
This is required to compile chrome with 1.8.8 as we treat warnings as errors.


eaenet:
Err, I only want to latest change there. Let me try to recreate this pull request.

PR #1147, opened 2018-08-29
[DON'T MERGE] C++11 solution for previous commit

behdad:
Let's see which gods (new term we use to refer to build bots) we anger.
Will revert after.


behdad:
Okay, most angry.

PR #1148, opened 2018-08-29
Delete pkg.m4

rossburton:
pkg.m4 is provided by pkg-config, which is a requirement to build, and this copy is ancient.  Delete it to avoid it being used instead of the host copy.


behdad:
Thanks. Good to hear from you!

PR #1158, opened 2018-09-13
Few fixes for Visual Studio

fanc999:
Hi,

There are some fixes that are needed for Visual Studio builds when:
-We are building the tools and we are building HarfBuzz as a DLL, as Visual Studio does not support GCC visibility attributes as well.
-Building the test-subset* sources on older Visual Studio, as some variables are not at the top of the block.

This PR is for the fixes that are needed.

With blessings, thank you!

### DIFF
--- a/src/hb.hh
+++ b/src/hb.hh
@@ -39,6 +39,10 @@
 #define _POSIX_C_SOURCE 200809L
 #endif
 
+#if defined (_MSC_VER) && defined (HB_DLL_EXPORT)
+#define HB_EXTERN __declspec (dllexport) extern
+#endif
+
 #include "hb.h"
 #define HB_H_IN
 #ifdef HAVE_OT
@@ -178,8 +182,10 @@ struct _hb_alignof
 # if !defined(HB_NO_VISIBILITY) && !defined(__MINGW32__) && !defined(__CYGWIN__) && !defined(_MSC_VER) && !defined(__SUNPRO_CC)
 #  define HB_INTERNAL __attribute__((__visibility__("hidden")))
 # elif defined(__MINGW32__)
-   /* We use -export-symbols on mingw32, since it does not support visibility
-    * attribute. */
+   /* We use -export-symbols on mingw32, since it does not support visibility attributes. */
+#  define HB_INTERNAL
+# elif defined (_MSC_VER) && defined (HB_DLL_EXPORT)
+   /* We do not try to export internal symbols on Visual Studio */
 #  define HB_INTERNAL
 #else
 #  define HB_INTERNAL
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -574,7 +574,7 @@ if (HB_HAVE_GOBJECT)
 endif ()
 
 if (BUILD_SHARED_LIBS AND WIN32 AND NOT MINGW)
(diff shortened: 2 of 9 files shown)


src/hb.hh
```diff
@@ -173,9 +173,8 @@ struct _hb_alignof
 #ifndef HB_INTERNAL
 # if !defined(HB_NO_VISIBILITY) && !defined(__MINGW32__) && !defined(__CYGWIN__) && !defined(_MSC_VER) && !defined(__SUNPRO_CC)
 #  define HB_INTERNAL __attribute__((__visibility__("hidden")))
-# elif defined(__MINGW32__)
-   /* We use -export-symbols on mingw32, since it does not support visibility
-    * attribute. */
+# elif defined(__MINGW32__) || defined (_MSC_VER)
+   /* We use -export-symbols on mingw32 and __declspec(dllexport), since they do not support visibility attributes. */
```
behdad: I don't like this change.  Instead, add the code to define HB_INTERNAL to __declspec(dllexport) *here*, instead of putting it in the build file or something.

src/hb.hh
```diff
@@ -174,9 +178,11 @@ struct _hb_alignof
 # if !defined(HB_NO_VISIBILITY) && !defined(__MINGW32__) && !defined(__CYGWIN__) && !defined(_MSC_VER) && !defined(__SUNPRO_CC)
 #  define HB_INTERNAL __attribute__((__visibility__("hidden")))
 # elif defined(__MINGW32__)
-   /* We use -export-symbols on mingw32, since it does not support visibility
-    * attribute. */
+   /* We use -export-symbols on mingw32, since it does not support visibility attributes. */
 #  define HB_INTERNAL
+# elif defined (_MSC_VER) && defined (HB_BUILD_DLL)
+   /* Export the symbols for harfbuzz-subset.dll */
+#  define HB_INTERNAL __declspec (dllexport)
```
behdad: No, we don't want this.  This is wrong.

src/hb.hh
```diff
@@ -35,6 +35,10 @@
 #include "config.h"
 #endif
 
+#if defined (_MSC_VER) && defined (HB_BUILD_DLL)
```
behdad: I wonder if we should check for DLL_EXPORT?  Is that standard name?

fanc999:
Hi @behdad,

Sorry for not coming back to this as I was looking at other things... So, I'd like to find out, should I just define HB_INTERNAL as nothing or as __declspec (dllexport) when we define HB_BUILDING_DLL, as I am a bit confused.

With blessings, thank you!

src/hb.hh
fanc999: DLL_EXPORT seems to me more of a MinGW thing, so I don't think it is exactly standard.

behdad:
HB_INTERNAL should definitely NOT be __declspec (dllexport).  What's the way to NOT export symbols?  Leave empty I guess?

behdad:
I prefer HB_DLL_EXPORT as a name.

fanc999:
Hi @behdad,

I updated HB_BUILD_DLL to HB_DLL_EXPORT as requested, and HB_INTERNAL is indeed defined to nothing when building on Visual Studio and HB_DLL_EXPORT is defined.

With blessings, thank you!

PR #1161, opened 2018-09-14
Fix installation of cmake config when building with cmake

tronical:
When building with the auto tools, the manually mantained
harfbuzz-config.cmake is installed. When building with cmake, we can use
cmake to generate the correct config files for us and install them.


ebraminio:
Sure, why not, thanks :)

behdad:
Good to hear from you, Simon!

ebraminio:
> Good to hear from you, Simon!

https://github.com/harfbuzz/harfbuzz-old/commits/master?before=3ab7b37bdebf0f8773493a1fee910b151c4de30f+175 same here...

Issue #1168, opened 2018-09-21
Add bots with address- and thread-sanitizer


behdad:
This is how ICU is doing it:

https://github.com/unicode-org/icu/blob/master/.travis.yml#L33

behdad:
Continuing here.
For address-sanitizer, maybe add some flags.  From https://clang.llvm.org/docs/AddressSanitizer.html maybe "-O1 -g -fsanitize=address -fno-omit-frame-pointer"

For thread-sanitizer to have any use, we need to add multi-threaded test, and then just run that one with thread-sanitizer.

Also useful is undefined-behavior sanitizer, and memory sanitizer. Unfortunately doesn't seem like they can be combined...  But one bot should be fine.

behdad:
Also, how did you read the reports?  I don't see any symbols.  Check this: https://clang.llvm.org/docs/AddressSanitizer.html#symbolizing-the-reports

behdad:
Also these?

Use-after-return (runtime flag ASAN_OPTIONS=detect_stack_use_after_return=1)
Use-after-scope (clang flag -fsanitize-address-use-after-scope)

behdad:
(I used asan_symbolize.)

behdad:
Oh, I just saw that you added msan, etc as well.

ebraminio:
'-O1 -g -fsanitize=address -fno-omit-frame-pointer' it was making symbol exporting test to fail

ebraminio:
> Also these?

Default enabled AFAIK.

> (I used asan_symbolize.)

Let me know exact build environment you have used locally and I will put that on the bot.

ebraminio:
Ok, lets close this.

PR #1169, opened 2018-09-22
Add bots with address- and thread-sanitizer

ebraminio:
Fixes https://github.com/harfbuzz/harfbuzz/issues/1168


ebraminio:
https://circleci.com/gh/harfbuzz/harfbuzz/33002 over to you guys I guess @behdad @garretrieger @rsheeter

ebraminio:
Fixed some of them, including subset related one so nothing in particular about hb-subset for now. Lets see how to fix the rest.

ebraminio:
> But yeah, I think one bot is fine.

Will do after we have all the bots green.

behdad:
> > But yeah, I think one bot is fine.
> 
> Will do after we have all the bots green.

Since you set it up already, let's leave as is and see how long it takes them...  I don't know how much setting up the bot takes in comparison to just running our stuff...

Also, @garretrieger will buy us an additional container...

PR #1178, opened 2018-09-24
[circleci] Add an obsessive clang bot

### LINKED ISSUE #869: Add more compiler warnings/errors into CI or build
There are some useful compiler flags https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html and https://clang.llvm.org/docs/ClangCommandLineReference.html can be enabled on the CI or build. It would be nice if we pick some of them and add them on our CI or build script.

ebraminio:
Fixes #869

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,6 +79,22 @@ jobs:
       - run: make
       - run: LD_LIBRARY_PATH="$PWD/freetype-2.9/objs/.libs" make check || .ci/fail.sh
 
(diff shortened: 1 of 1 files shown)


behdad:
Wow!  That's a long list :D

ebraminio:
Yes, we should bring back the ones makes more sense and fix their warning gradually, this however ensure we know about the brand new clang issue we easily can ignore.

PR #1179, opened 2018-09-25
CoreText: fix iOS build

jbnoel-gpsw:
1.7.6 builds fine for iOS (cross-compilation on macOS).
1.7.7 version introduced a new call to CoreText 'CTFontGetPlatformFont' in file src/hb-coretext.cc
This function 'CTFontGetPlatformFont' is not available on iOS (only on macOS):
https://developer.apple.com/documentation/coretext/1510544-ctfontgetplatformfont

From 1.7.7 to 1.9.0, harfbuzz can not be build for iOS for this reason.
The commit based on 1.9.0 version goes back to old code path of the 1.7.6 :
Is it the right way to fix the iOS build ?

For your information, we have also a patch to support arm64 architecture on iOS in our build environment:
```--- config.sub	2014-11-03 15:12:46.000000000 +0100
+++ config.sub	2014-11-03 15:11:32.000000000 +0100
@@ -253,7 +253,7 @@
 	| alpha64 | alpha64ev[4-8] | alpha64ev56 | alpha64ev6[78] | alpha64pca5[67] \
 	| am33_2.0 \
 	| arc | arceb \
-	| arm | arm[bl]e | arme[lb] | armv[2-8] | armv[3-8][lb] | armv7[arm] \
+	| arm | arm[bl]e | arme[lb] | armv[2-8] | armv[3-8][lb] | armv7[arm]  | arm64 \
 	| avr | avr32 \
 	| be32 | be64 \
 	| bfin \
@@ -367,7 +367,7 @@
 	| alpha-* | alphaev[4-8]-* | alphaev56-* | alphaev6[78]-* \
 	| alpha64-* | alpha64ev[4-8]-* | alpha64ev56-* | alpha64ev6[78]-* \
 	| alphapca5[67]-* | alpha64pca5[67]-* | arc-* | arceb-* \
-	| arm-*  | armbe-* | armle-* | armeb-* | armv*-* \
+	| arm-*  | armbe-* | armle-* | armeb-* | armv*-* | arm64-* \
 	| avr-* | avr32-* \
 	| be32-* | be64-* \
 	| bfin-* | bs2000-* \```

Command line used to build in our build environment:
```CPPFLAGS='-arch arm64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS11.4.sdk -miphoneos-version-min=7.0 -O2'
[...]

PR #1180, opened 2018-09-25
Fix iOS builds

jgranick:
Some of the older macOS CoreText code does not work with the current iOS SDKs, and since I believe `MAC_OS_X_VERSION_MIN_REQUIRED` evaluates to 0 on an iOS build, it falls into the older code rather than using the newer APIs. Thanks!


khaledhosny:
Duplicate of #1179

jgranick:
Hah! Weird timing

PR #1183, opened 2018-09-26
Fix infinite loop when walking up the directory hierarchy

vkrause:
A single find_package(harfbuzz) line in user code resulted in this loop
getting stuck when _harfbuzz_libdir_iter became "/".


ebraminio:
Thank you :)

PR #1184, opened 2018-09-26
Add a multithreaded test

### LINKED ISSUE #1173: Write a multi-threaded test
From https://github.com/harfbuzz/harfbuzz/issues/1168#issuecomment-423852823:

> For thread-sanitizer to have any use, we need to add multi-threaded test, and then just run that one with thread-sanitizer.

It will be fun.

ebraminio:
Fixes #1173

### DIFF
--- a/test/api/test-multithread.c
+++ b/test/api/test-multithread.c
@@ -1,0 +1,146 @@
+/*
+ * Copyright © 2018  Ebrahim Byagowi
+ *
+ *  This is part of HarfBuzz, a text shaping library.
+ *
+ * Permission is hereby granted, without written agreement and without
+ * license or royalty fees, to use, copy, modify, and distribute this
+ * software and its documentation for any purpose, provided that the
+ * above copyright notice and the following two paragraphs appear in
+ * all copies of this software.
+ *
+ * IN NO EVENT SHALL THE COPYRIGHT HOLDER BE LIABLE TO ANY PARTY FOR
+ * DIRECT, INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES
+ * ARISING OUT OF THE USE OF THIS SOFTWARE AND ITS DOCUMENTATION, EVEN
+ * IF THE COPYRIGHT HOLDER HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ *
+ * THE COPYRIGHT HOLDER SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ * FITNESS FOR A PARTICULAR PURPOSE.  THE SOFTWARE PROVIDED HEREUNDER IS
+ * ON AN "AS IS" BASIS, AND THE COPYRIGHT HOLDER HAS NO OBLIGATION TO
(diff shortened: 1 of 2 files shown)


test/api/test-multithread.c
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  return 0;
+}
+
+int
+main (int argc, char **argv)
+{
+  int i;
+  int num_threads = 10;
```
ebraminio: Do we need it configurable? The code works locally only if is set to 1, showing something apparently.

test/api/test-multithread.c
```diff
@@ -0,0 +1,119 @@
...
+  pthread_t *threads = calloc (num_threads, sizeof (pthread_t));
+  hb_buffer_t **buffers = calloc (num_threads, sizeof (hb_buffer_t *));
+
+  pthread_mutex_lock (&mutex);
+
+  hb_blob_t *blob = hb_blob_create_from_file (path);
+  hb_face_t *face = hb_face_create (blob, 0);
+  font = hb_font_create (face);
+  hb_ot_font_set_funcs (font);
```
ebraminio: To ensure we don't use not instrumented fonts.

test/api/test-multithread.c
```diff
@@ -0,0 +1,119 @@
...
+    {
+      hb_buffer_t *buffer = buffers[i];
+      hb_buffer_diff_flags_t diff = hb_buffer_diff (ref_buffer, buffer, (hb_codepoint_t) -1, 0);
+      if (diff)
+      {
+        fprintf (stderr, "One of the buffers (%d) was different from the reference.\n", i);
+        char out[255];
+
+        hb_buffer_serialize_glyphs (buffer, 0, hb_buffer_get_length (ref_buffer),
+            out, sizeof (out), NULL,
```
ebraminio: Spaces, I know.

ebraminio:
Where the test should be located? Apparently feeding argument is not that easy on current folder, will do "Just take font file and text and num-threads / num-iters...." after that decision

behdad:
> Where the test should be located? Apparently feeding argument is not that easy on current folder, will do "Just take font file and text and num-threads / num-iters...." after that decision

Not sure either.

Another option is to make it read .tests file data from stdin and run each of those in a separate thread.  This will exercise all our fonts and tests in one mega process.  Is probably better / much faster than running each separately.

test/api/test-multithread.c
ebraminio: We can build freetype on that bot.

ebraminio:
> I like to run it and see if it catches anything.

I am merging it with a "int num_threads = 1; // FIXME: Increase this and fix the issue" so you can test and see locally if is valid. Thanks.

test/api/test-multithread.c
```diff
@@ -0,0 +1,119 @@
...
+
+const char *text = "طرح‌نَما";
+const char *path =
+#if defined(__linux__)
+		"/usr/share/fonts/truetype/dejavu/DejaVuSans.ttf";
+#elif defined(_WIN32) || defined(_WIN64)
+		"C:\\Windows\\Fonts\\tahoma.ttf";
+#elif __APPLE__
+		"/Library/Fonts/Tahoma.ttf";
+#endif
```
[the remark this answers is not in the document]
khaledhosny: Why not just one of the fonts we have on the repo?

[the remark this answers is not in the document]
ebraminio: Sorry for not seeing this sooner. You are right Khaled. Better to remove all the hard coding.

PR #1187, opened 2018-09-26
Test freetype funcs on test-multithread

ebraminio:


ebraminio:
Trolls tsan and ubsan bots, probably going to be fun :)

PR #1197, opened 2018-09-28
Fedora+64bit wine

### LINKED ISSUE #1157: Status of win64 harfbuzz , and 64-bit uniscribe.
I finally got round to spend a few days looking at a rather old bug in win64 FontVal ( https://github.com/HinTak/Font-Validator/issues/27 ) . It turns out the FreeType backend crashed when calling `hb_ft_font_create` in the autofit module. I probably will have to dive into harfbuzz myself, but meanwhile, it would be useful to know:

- the status of win64 harfbuzz

- update the wine README for 64-bit wine and 64-bit uniscribe, since 64-bit windows is rather the norm now.

ebraminio:
@HinTak work. Looks great, lets review and merge it, fixes #1157 also


README.wine.md
```diff
@@ -1,40 +1,98 @@
...
 implemenetation and that specially is important where OpenType specification
 is or wasn't that clear. For having access to Uniscribe on Linux/macOS these
 steps are recommended:
 
-1. Install Wine from your favorite package manager.
+1. Install Wine and mingw32/mingw64 from your package manager. This likely
+   means, at the very least, running this:
+
+```
+sudo dnf install -y wine.i686 wine.x86_64 mingw32-gcc-c++ mingw64-gcc-c++
```
ebraminio: Can you put its Debian (Ubuntu to be precise equivalent of it) also? Or make the guide that general to cover that also

README.wine.md
```diff
@@ -1,40 +1,98 @@
...
+```
+mkdir win32build
+cd win32build && ../mingw32.sh --with-uniscribe && cd ..
+make -Cwin32build
+```
+
+4. Build the win64 version of harbuzz with:
 
-2. And `mingw-w64` compiler.
-   With `brew` on macOS, you can have it like `brew install mingw-w64`
```
ebraminio: Covering macOS also would be nice, we sometime need to test it also and on mac all the three shapers (harfbuzz, uniscribe, coretext) can be present at the same time :)

README.wine.md
```diff
@@ -1,40 +1,98 @@
...
+```
+mkdir win64build
+cd win64build && ../mingw64.sh --with-uniscribe && cd ..
+make -Cwin64build
+```
 
-3. Download and put [this](https://drive.google.com/open?id=0B3_fQkxDZZXXbWltRGd5bjVrUDQ)
```
ebraminio: Can you generalize the guide to cover that? That packages is not available on Ubuntu and macOS so having this fallback will be nice I believe

mingw32.sh
```diff
@@ -14,9 +14,7 @@ unset PKG_CONFIG_PATH
 # Removed -static from the following
 export CFLAGS="-static-libgcc"
 export CXXFLAGS="-static-libgcc -static-libstdc++"
-export CPPFLAGS="-I$HOME/.local/$target/include -O2"
-export LDFLAGS=-L$HOME/.local/$target/lib
-export PKG_CONFIG_LIBDIR=$HOME/.local/$target/lib/pkgconfig
-export PATH=$HOME/.local/$target/bin:$PATH
+export CPPFLAGS="-O2"
+export PKG_CONFIG_LIBDIR=/usr/i686-w64-mingw32/sys-root/mingw/lib/pkgconfig/
```
ebraminio: Only available on Fedora, can you somehow generalize it to cover the way we used it

HinTak: I just checked, PKG_CONFIG_LIBDIR can be a list separated by colon also; so the originals of these 4 are harmless on Fedora if .local/target is not there.

README.wine.md
HinTak: I'll need to look up a bit on this.

README.wine.md
HinTak: Okay.

behdad:
As long as doesn't break the mingw32 script for me and instructions don't get overly complicated, fine with me.

HinTak:
To be honest I removed a lot of stuff in wine.readme because they are not needed/too complicated and I don't particularly want to include instructions about Ubuntu, etc. I don't use currently use it and do not want to spend time researching and merging information which are of no use to me. I am okay of letting the fork stay indefinitely, and not merging. Having different branches of that document for different systems seems cleaner.

ebraminio:
@HinTak I didn't see that but OK, let me see if I can reach to it sometimes. Thanks

PR #1211, opened 2018-10-02
[circleci] Fix some of warnings from clang-everything bot

ebraminio:
Fixes -Wshift-sign-overflow and -Wmissing-prototypes


src/hb-buffer-serialize.cc
```diff
@@ -368,7 +367,6 @@ hb_buffer_serialize_glyphs (hb_buffer_t *buffer,
 					       buf, buf_size, buf_consumed,
 					       font, flags);
 
-    default:
     case HB_BUFFER_SERIALIZE_FORMAT_INVALID:
       return 0;
```
ebraminio: It says it is not needed to put default when all the cases are available on the switch.

src/hb-ot-shape.cc
```diff
@@ -632,7 +631,7 @@ hb_ot_substitute_complex (hb_ot_shape_context_t *c)
 
   c->plan->substitute (c->font, buffer);
 
-  if (0) /* XXX Call morx instead. */
+  if ((0)) /* XXX Call morx instead. */
```
ebraminio: Apparently the way clang prefers for disabled codes.

src/hb-ucdn.cc
```diff
@@ -274,6 +274,9 @@ void free_static_ucdn_funcs (void)
 #endif
 
 extern "C" HB_INTERNAL
+hb_unicode_funcs_t *
+hb_ucdn_get_unicode_funcs (void);
```
ebraminio: Not available on any header and clang doesn't like it

util/ansi-print.cc
```diff
@@ -71,7 +71,7 @@ struct color_t
 {
   static color_t from_ansi (unsigned int x)
   {
-    color_t c = {(0xFF<<24) | ((0xFF*(x&1))<<16) | ((0xFF*((x >> 1)&1))<<8) | (0xFF*((x >> 2)&1))};
+    color_t c = {(0xFFu<<24) | ((0xFF*(x&1))<<16) | ((0xFF*((x >> 1)&1))<<8) | (0xFF*((x >> 2)&1))};
```
ebraminio: To make it unsigned, as clang wants

src/hb-buffer-serialize.cc
khaledhosny: I don’t think removing `default` case is right here, since this is handling user input and it can be anything, so one needs to catch invalid values. What is the warning message this is fixing?

src/hb-glib.cc
```diff
@@ -246,7 +246,7 @@ hb_glib_unicode_compose (hb_unicode_funcs_t *ufuncs HB_UNUSED,
 {
 #if GLIB_CHECK_VERSION(2,29,12)
   return g_unichar_compose (a, b, ab);
-#endif
+#else
```
khaledhosny: See the comment below, this is intentionally left **not** `ifdef`’d out.

khaledhosny: Alternatively we need a bot this builds with an old GLib that hits the `else` branch, and remove the comment that is no longer true.

src/hb-glib.cc
```diff
@@ -283,7 +284,7 @@ hb_glib_unicode_decompose (hb_unicode_funcs_t *ufuncs HB_UNUSED,
 {
 #if GLIB_CHECK_VERSION(2,29,12)
   return g_unichar_decompose (ab, a, b);
-#endif
+#else
```
khaledhosny: Same here.

src/hb-buffer-serialize.cc
behdad: Right.  Since this is input, we should actually cast it to int or unsigned in the switch ().  Otherwise the compiler is "within its rights" to assume default never happens.  Even with such a cast, it probably is technically still undefined-behavior if someone passed in another value.  Not sure, need to contact C/C++ lawyers.  Anyway, adding the cast means we won't get warnings for missing case items if we add new values to the enum.  I think I like the cast though.

src/hb-glib.cc
behdad: Don't think we need to build, given that the code doesn't change.  But yeah, I prefer to compile the old code.  Then again, since this code won't ever change, I'm fine not compiling it.

src/hb-icu.cc
```diff
@@ -173,14 +173,12 @@ hb_icu_unicode_compose (hb_unicode_funcs_t *ufuncs HB_UNUSED,
...
-    *ab = ret;
-    return true;
-  }
-#endif
+  const UNormalizer2 *normalizer = (const UNormalizer2 *) user_data;
+  UChar32 ret = unorm2_composePair (normalizer, a, b);
+  if (ret < 0) return false;
+  *ab = ret;
+  return true;
+#else
```
behdad: Why remove the braces?

src/hb-ot-math-table.hh
```diff
@@ -148,8 +148,8 @@ struct MathConstants
     case HB_OT_MATH_CONSTANT_RADICAL_DEGREE_BOTTOM_RAISE_PERCENT:
       return radicalDegreeBottomRaisePercent;
 
-    default:
-      return 0;
+    // default:
```
behdad: Nope.

src/hb-ucdn.cc
behdad: Okay.  Not sure why I don't get the warning.

util/ansi-print.cc
behdad: Add u to all of the 0xFF's please.

behdad:
I think I added the default's because gcc was warning about missing ones at some point!

ebraminio:
thank you guys, will try to submit each removal separately

src/hb-ucdn.cc
ebraminio: The bot uses a brand clang compiler, so no one can come from Chrome project and say they need something :)

src/hb-ot-shape.cc
ebraminio: The same here https://stackoverflow.com/a/29177815 and apparently like that instead

PR #1218, opened 2018-10-03
[test] Use an in-repo font for test-multithread

ebraminio:


behdad:
This wouldn't work in out-of-tree builds.  See other tests that use `g_test_build_filename`.

behdad:
Also should check what asan fail is about.

behdad:
Also, you probably should change this line:
```
  hb_buffer_add_utf8 (buffer, text, sizeof (text), 0, sizeof (text)); 
```
to
```
  hb_buffer_add_utf8 (buffer, text, sizeof (text), 0, -1); 
```

test/api/hb-subset-test.h
```diff
@@ -57,6 +57,11 @@ hb_subset_test_open_font (const char *font_path)
 #endif
 
   hb_blob_t* blob = hb_blob_create_from_file (path);
+  if (hb_blob_get_length (blob) == 0)
+  {
+    printf ("The test font is not found.");
+    exit (1);
+  }
```
ebraminio: Adding the check here, does it look good?

behdad: Not sure what you mean.

test/fuzzing/main.cc
```diff
@@ -1,12 +1,18 @@
...
 
 int main(int argc, char **argv) {
   hb_blob_t *blob = hb_blob_create_from_file (argv[1]);
   unsigned int len;
   const char *font_data = hb_blob_get_data (blob, &len);
+  if (len == 0)
+  {
+    printf ("The test font is not found.");
+    exit (1);
+  }
```
ebraminio: And here.

test/api/hb-subset-test.h
ebraminio: The same as https://github.com/harfbuzz/harfbuzz/commit/f72b748371da2d7ce327a5d15feea46960aa7dd3#diff-7e54590da7a9eb1143ac164a13c1f300L9

test/api/test-multithread.c
```diff
@@ -135,16 +127,34 @@ test_body (void)
 int
 main (int argc, char **argv)
 {
+  g_test_init (&argc, &argv, NULL);
```
ebraminio: Is this look good?

behdad:
Just do whatever is needed :)

behdad:
I'm trying to figure out the asan issue.  It's wrong to pass sizeof(text) to buffer API, but it shouldn't result in bad memory access.

behdad:
> I'm trying to figure out the asan issue. It's wrong to pass sizeof(text) to buffer API, but it shouldn't result in bad memory access.

Ah, text is a pointer, not array.  Makes sense.  Using -1 is correct fix.

PR #1234, opened 2018-10-10
better logic for introspection check, and fails on introspection with…

HinTak:
…out gobject


behdad:
Humm.  Doesn't seem to work: https://circleci.com/gh/harfbuzz/harfbuzz/39139?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

HinTak:
Argh, it is because of the auto case...

HinTak:
Missing quotes too.

HinTak:
Close for now, reopening when I figure it out

PR #1238, opened 2018-10-11
Parse Indic3 tags

dscorbett:
All the Indic3 tags are mapped to the same scripts as the Indic2 tags. This includes 'mym3', though 'mym2' doesn’t use the Indic shaper.


src/hb-ot-tag.cc
```diff
@@ -143,7 +143,11 @@ hb_ot_all_tags_from_script (hb_script_t   script,
 
   hb_tag_t new_tag = hb_ot_new_tag_from_script (script);
   if (unlikely (new_tag != HB_OT_TAG_DEFAULT_SCRIPT))
-    tags[i++] = new_tag;
+  {
+    tags[i++] = new_tag | '3';
```
behdad: Hahahaha.

PR #1240, opened 2018-10-12
Fix Indic script tags in Graphite

dscorbett:
`_hb_graphite2_shape` is supposed to use the same script tags as OpenType, except that it prefers e.g. 'deva' to 'dev2'. If the call to `hb_ot_tags_from_script_and_language` is restricted to 2 script tags, the tags will be e.g. 'dev3' and 'dev2'.


behdad:
Ah.  Thanks!

PR #1245, opened 2018-10-13
Add two new APIs related to name table

### LINKED ISSUE #432: Consider exposing access to 'name' table on hb_face_t or hb_font_t

ebraminio:
WIP

Fixes #432 (tries to, eventually)

### DIFF
--- a/src/hb-ot-name-table.hh
+++ b/src/hb-ot-name-table.hh
@@ -79,12 +79,12 @@ struct name
 {
   static const hb_tag_t tableTag	= HB_OT_TAG_name;
 
-  inline unsigned int get_name (unsigned int platform_id,
-				unsigned int encoding_id,
-				unsigned int language_id,
-				unsigned int name_id,
-				void *buffer,
-				unsigned int buffer_length) const
+  inline unsigned int get_name (unsigned int  platform_id,
+				unsigned int  encoding_id,
+				unsigned int  language_id,
+				unsigned int  name_id,
+				void         *buffer,
+				unsigned int  buffer_length) const
   {
     NameRecord key;
     key.platformID.set (platform_id);
@@ -104,13 +104,27 @@ struct name
   inline unsigned int get_size (void) const
   { return min_size + count * nameRecordZ[0].min_size; }
 
-  inline bool sanitize_records (hb_sanitize_context_t *c) const {
+  inline bool sanitize_records (hb_sanitize_context_t *c) const
+  {
     TRACE_SANITIZE (this);
     char *string_pool = (char *) this + stringOffset;
     unsigned int _count = count;
     for (unsigned int i = 0; i < _count; i++)
       if (!nameRecordZ[i].sanitize (c, string_pool)) return_trace (false);
     return_trace (true);
+  }
+
+  inline unsigned int get_count (void) const
+  { return count; }
+
+  inline const NameRecord& get_record(unsigned int index) const
+  { return hb_array_t<NameRecord> (nameRecordZ.arrayZ, count)[index]; }
+
+  inline hb_bytes_t get_record_string(unsigned int index) const
+  {
+    const NameRecord& record = get_record (index);
+    if (!record.length) return hb_bytes_t ();
+    return hb_bytes_t ((char *) this + stringOffset + record.offset, record.length);
   }
 
   inline bool sanitize (hb_sanitize_context_t *c) const
--- a/src/hb-ot-name.cc
+++ b/src/hb-ot-name.cc
@@ -1,0 +1,95 @@
+/*
+ * Copyright © 2018  Ebrahim Byagowi
+ *
+ *  This is part of HarfBuzz, a text shaping library.
+ *
+ * Permission is hereby granted, without written agreement and without
+ * license or royalty fees, to use, copy, modify, and distribute this
+ * software and its documentation for any purpose, provided that the
+ * above copyright notice and the following two paragraphs appear in
+ * all copies of this software.
+ *
+ * IN NO EVENT SHALL THE COPYRIGHT HOLDER BE LIABLE TO ANY PARTY FOR
+ * DIRECT, INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES
+ * ARISING OUT OF THE USE OF THIS SOFTWARE AND ITS DOCUMENTATION, EVEN
+ * IF THE COPYRIGHT HOLDER HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ *
+ * THE COPYRIGHT HOLDER SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ * FITNESS FOR A PARTICULAR PURPOSE.  THE SOFTWARE PROVIDED HEREUNDER IS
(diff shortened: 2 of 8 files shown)


ebraminio:
Similar to support freetype has https://www.freetype.org/freetype2/docs/reference/ft2-sfnt_names.html AFAIK

behdad:
I thought more about this.  I'm leaning towards NOT exposing the platform/encoding/language madness, and provide a higher-level API that works with hb_language_t only.

src/hb-ot-name-table.hh
```diff
@@ -113,6 +114,32 @@ struct name
     return_trace (true);
   }
 
+  inline unsigned int get_count (void) const
+  { return count; }
+
+  inline const NameRecord& get_record(unsigned int index) const
+  { return index < count ? nameRecordZ[index] : Null (NameRecord); }
```
behdad: We actually have a new idiom for these.  We bind them to an `hb_array_t` that has size, then index:
`{ return hb_array_t<NameRecord> (nameRecordZ, count)[index]; }`

ebraminio:
> I thought more about this. I'm leaning towards NOT exposing the platform/encoding/language madness, and provide a higher-level API that works with hb_language_t only.

Understandable and I was thinking also. We have a search already that can be adopted with it also. For starting that, can you describe its final interface?

ebraminio:
Also @khaledhosny or other dear GitHub subscribers, have you seen other implementation for similar feature (making `name` table accessible) on other libraries and can you link to them here so we can learn from them? Thanks.

behdad:
Ref https://github.com/harfbuzz/harfbuzz/issues/432

twardoch:
Practically, the name table uses U.S. English similarly to the GSUB/GPOS DefaultLangSys while other languages, if they're used, are used like non-default languagesystems. 

The "name" table platforms (3.1, 0.3, 1.0 etc.) are akin to GSUB/GPOS scripts of sorts, or like "generations" of a particular script (like dev3, dev2 and deva). 

The actual name IDs are a bit like features, but that's a very far-fetched comparison. Most of them are OK to be queried by their number or (for the registered IDs < 256) some named constants. 

The only significant divergence is for (21, 22), (16, 17), (1,2), and the PostScript names (6, 20 and 25), which should be implemented as functions of sorts, because there is a recommended precedence (so if 17 does not exist, you check 2). 

And of course there is the STAT table where the style words are supposed to be used to create style names dynamically, and fvar that points to names for predefined instances. So the actual process of arriving at some family and especially style name is a bit complex. 

And then there's names for stylistic sets etc.

behdad:
We expose name-IDs in other API.  So this API need just provide way to fetch by name-ID.  Yes, enum for standard name-IDs might be nice.  I don't think we should do fallback re 21,22/16,17/1,2 or anything.

Please continue discussion in https://github.com/harfbuzz/harfbuzz/issues/432

ebraminio:
> stylistic sets

is done, any other similar thing?

src/hb-ot-name.cc
```diff
@@ -0,0 +1,102 @@
...
+hb_ot_name_get_languages (hb_face_t    *face,
+			  unsigned int  name_index,
+			  unsigned int  start_offset,
+			  unsigned int *languages_length,
+			  unsigned int *languages)
+{
+  const OT::name& table = _get_name (face);
+
+  unsigned int matched = 0;
+  for (unsigned int i = 0; i < table.count; i++)
```
ebraminio: Sad about linear search but other than computers are fast nowadays, not sure how to fill the languages buffer correctly.

ebraminio: I thought your agreement on the other thread was because of agreeing with the API Behdad was suggesting :P

src/hb-ot-name.cc
```diff
@@ -0,0 +1,102 @@
...
+		       unsigned int  start_offset,
+		       unsigned int *buffer_length /* IN/OUT.  May be NULL */,
+		       char         *buffer        /* OUT.     May be NULL */)
+{
+  const OT::name& table = _get_name (face);
+  for (unsigned int i = 0; i < table.count; i++)
+  {
+    const OT::NameRecord& record = table.get_record (i);
+    if (record.languageID == language_id && record.nameID == name_id)
+      return table.get_record_string (i, start_offset, buffer_length, buffer);
```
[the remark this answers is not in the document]
ebraminio: I know, far from prefect obviously. Maybe I should close this. I thought that discussion is reached to some conclusion after your agreement so probably this should be closed by Behdad again :) Also feel free to take over if you like to, I just want to make it happen.

How to implement best match also? And what should be destined encoding, hb_codepoint_t? Maybe I should find your Scribus implementation :)

behdad:
Best-match is similar to what we do in hb-ot-font for finding best cmap subtable.  Or see fontconfig.  Basically, platform 3, then 0, then 1.  Within each platform, kinda higher encoding is better.  All this amongst the is_unicode ones.

src/hb-ot-name-table.hh
```diff
@@ -113,6 +114,32 @@ struct name
     return_trace (true);
   }
 
+  inline unsigned int get_count (void) const
+  { return count; }
+
+  inline const NameRecord& get_record(unsigned int index) const
+  { return hb_array_t<NameRecord> (nameRecordZ.arrayZ, count)[index]; }
+
+  inline unsigned int get_record_string(unsigned int  index,
```
behdad: I think from this function we want to return a hb_bytes_t.  The API entrypoint can decide whether to return it as is, or to copy, or to transcribe it to UTF-8 or UTF-16.

src/hb-ot-name.cc
behdad: Yeah thinking about it more, since this is going to be really slow, let's leave it out now.

behdad: and instead provide the lower-level items so advanced clients (eg. fontconfig) can figure it out themselves.

src/hb-ot-name.cc
```diff
@@ -0,0 +1,102 @@
...
+ * @buffer: (out)
+ *
+ * Return value: The string buffer size
+ *
+ * Since: REPLACEME
+ **/
+unsigned int
+hb_ot_name_get_string (hb_face_t    *face,
+		       unsigned int  name_id,
+		       unsigned int  language_id,
```
behdad: We need mapping from `hb_language_t`.

Really, there's quite some infrastructure to build before this API can be implemented.

src/hb-ot-name.h
```diff
@@ -48,6 +48,16 @@ typedef unsigned int hb_name_id_t;
...
 #define HB_NAME_ID_INVALID 0xFFFF
 
+HB_EXTERN unsigned int
+hb_ot_name_get_string (hb_face_t     *face,
+		       unsigned int   name_id,
+		       hb_language_t *languages,
+		       unsigned int   languages_count,
+		       unsigned int   start_offset,
+		       unsigned int  *buffer_length /* IN/OUT.  May be NULL */,
+		       char          *buffer        /* OUT.     May be NULL */);
```
behdad: Too complicated.  Need one language.  Also, get_utf8 and get_utf16 variants.  See my API details in other issue.  Sketch it there first, not in a patch.  Thanks.  Also, hb_name_id_t.

behdad: Also, don't think we need start_offset, since encoding has to walk through the whole thing...

ebraminio:
Closing at the final version will be so different and other may like to participate

PR #1254, opened 2018-10-16
[WIP] Name table

behdad:
Work in progress.

Fixes https://github.com/harfbuzz/harfbuzz/issues/432

### DIFF
--- a/src/Makefile.sources
+++ b/src/Makefile.sources
@@ -34,6 +34,7 @@ HB_BASE_sources = \
 	hb-ot-kern-table.hh \
 	hb-ot-maxp-table.hh \
 	hb-ot-name-table.hh \
+	hb-ot-name.cc \
 	hb-ot-os2-table.hh \
 	hb-ot-os2-unicode-ranges.hh \
 	hb-ot-post-macroman.hh \
--- a/src/hb-ot-face.cc
+++ b/src/hb-ot-face.cc
@@ -30,6 +30,7 @@
 #include "hb-ot-glyf-table.hh"
 #include "hb-ot-hmtx-table.hh"
 #include "hb-ot-kern-table.hh"
+#include "hb-ot-name-table.hh"
 #include "hb-ot-post-table.hh"
 #include "hb-ot-color-cbdt-table.hh"
 #include "hb-ot-layout-gdef-table.hh"
--- a/src/hb-ot-face.hh
+++ b/src/hb-ot-face.hh
@@ -45,6 +45,9 @@
  * This is as good as any place. */
 #define HB_OT_TABLES \
     /* OpenType shaping. */ \
+    HB_OT_ACCELERATOR(OT, GDEF) \
+    HB_OT_ACCELERATOR(OT, GSUB) \
+    HB_OT_ACCELERATOR(OT, GPOS) \
     HB_OT_TABLE(OT, JSTF) \
     HB_OT_TABLE(OT, BASE) \
     /* AAT shaping. */ \
@@ -59,15 +62,14 @@
     /* OpenType math. */ \
     HB_OT_TABLE(OT, MATH) \
     /* OpenType fundamentals. */ \
-    HB_OT_ACCELERATOR(OT, GDEF) \
-    HB_OT_ACCELERATOR(OT, GSUB) \
-    HB_OT_ACCELERATOR(OT, GPOS) \
     HB_OT_ACCELERATOR(OT, cmap) \
     HB_OT_ACCELERATOR(OT, hmtx) \
     HB_OT_ACCELERATOR(OT, vmtx) \
     HB_OT_ACCELERATOR(OT, post) \
+    HB_OT_ACCELERATOR(OT, name) \
     HB_OT_ACCELERATOR(OT, kern) \
     HB_OT_ACCELERATOR(OT, glyf) \
+    /* OpenType color fonts. */ \
     HB_OT_ACCELERATOR(OT, CBDT) \
     /* */
 
--- a/src/hb-ot-name-table.hh
+++ b/src/hb-ot-name-table.hh
@@ -58,6 +58,13 @@ struct NameRecord
     return 0;
   }
 
+  inline bool supported (void) const
+  {
+    return platformID == 0 ||
+           platformID == 3;
+    /* TODO Add Apple MacRoman (0:1). */
+  }
+
   inline bool sanitize (hb_sanitize_context_t *c, const void *base) const
   {
     TRACE_SANITIZE (this);
@@ -74,6 +81,18 @@ struct NameRecord
   public:
   DEFINE_SIZE_STATIC (12);
 };
+
+static int
+_hb_ot_name_entry_cmp (const void *pa, const void *pb)
+{
+  const hb_ot_name_entry_t *a = (const hb_ot_name_entry_t *) pa;
+  const hb_ot_name_entry_t *b = (const hb_ot_name_entry_t *) pb;
+  if (a->name_id != b->name_id)
+    return a->name_id < b->name_id ? -1 : +1;
+  if (a->index != b->index)
+    return a->index < b->index ? -1 : +1;
+  return 0;
+}
 
 struct name
 {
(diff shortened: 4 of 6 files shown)


behdad:
I fleshed most of this out.

What's left is 1. language handling and 2. encoding handling.  The UTF conversion routines can use some review.  @jfkthame Thanks in advance!

behdad:
I'm done with the implementation.  FULLY untested though...  Going to write a tool to list names in a font...

behdad:
I'm done here.  Tested.  Do "make src/test-name-table" and run that binary on font files to see all names with languages.

behdad:
Can hook up ltag as well, but who cares...

behdad:
Implemented ltag as well.  Not tested.

jfkthame:
Trying the test-name-table tool, I'm seeing "(null)" for Korean names in fonts. Checking the files, these are correctly tagged as 3/1/0x412, which is the langID for Korean according to https://docs.microsoft.com/en-us/windows/desktop/intl/language-identifier-constants-and-strings.

Looking in hb-ot-name-language.cc, I see that there's a commented-out entry for 0x412; what gives?

(FWIW, there's a later entry for 0x812 "ko", which doesn't seem to appear in the MS documentation... at least not on that page.)

jfkthame:
I'm also seeing names that use 3/1/0xC0A, which according to MSDN is Spain (Modern Sort); again, the corresponding entry in the HB table is commented out, so we get (null). Can we fill in these blanks, as far as possible?

jfkthame:
Others seen in the wild, but currently missing in HB, include

    {0x0403,	"ca"},	/* Catalan (Catalan) */
    {0x0424,	"sl"},	/* Slovenian (Slovenia) */

jfkthame:
Also

    {0x042D,	"eu"},	/* Basque (Basque) */

Another note is that if there are multiple entries with the same nameID but different unrecognized lang codes, the test-name-table tool currently exposes only one of them (with its lang listed as "(null)"); the others are silently ignored.

src/hb-utf.hh
```diff
@@ -237,32 +318,53 @@ struct hb_utf32_t
...
+    if (validate && unlikely (unicode >= 0xD800u && (unicode <= 0xDFFFu || unicode > 0x10FFFFu)))
+      unicode = 0xFFFDu;
+    *text++ = unicode;
+    return text;
+  }
 };
 
+typedef hb_utf32_xe_t<uint32_t> hb_utf32_t;
+typedef hb_utf32_xe_t<uint32_t, false> hb_utf32_novalidate_t;
+typedef hb_utf32_xe_t<OT::HBUINT32> hb_utf32_be_t;
```
jfkthame: This typedef looks redundant; I can't recall anywhere in OpenType that would use it?

src/hb-utf.hh
```diff
@@ -271,12 +373,84 @@ struct hb_latin1_t
...
+  typedef uint8_t codepoint_t;
+
+  static inline const codepoint_t *
+  next (const codepoint_t *text,
+	const codepoint_t *end HB_UNUSED,
+	hb_codepoint_t *unicode,
+	hb_codepoint_t replacement HB_UNUSED)
+  {
+    *unicode = *text++;
+    if (*unicode >= 0x100)
```
jfkthame: Shouldn't this be comparing against 0x0080?

behdad:
Thanks @jfkthame.  I fix the code.

Re languages, the lines that are commented out means fontconfig didn't have an entry for them.  Please review and push a commit with fixups.

Note that it's not just fontconfig or freetype data that was incomplete.  The data on OpenType site is incomplete as well.  For example, it doesn't have "Farsi" entry whereas 0x0429 has been in common use...

So the only question left is, what to do with null languages.  I'm leaning on not enumerating them at all.  Makes sense?

behdad:
Also, API question, if user passed null language, should we return "en"?

behdad:
Okay.  Fixed the bugs and skipping null languages now.  Just needs touchup to the language table.

@dscorbett maybe something you can help with as well?

behdad:
I'm filling in the missing entries now.

behdad:
Another thing I'm ignoring right now is the script/territory differences other than for Chinese.  Eg, Mongolian in Mongolian, Chinese, or Cyrillic script.  I need help getting the best-practice language codes for those, as fontconfig is proving to be quite unreliable.

behdad:
Maybe also add an enum of the predefined name_id's?

behdad:
Humm.  I'm also not happy that we have `hb_name_id_t` but the rest of the API is under `hb_ot_name_...`.

behdad:
Added the enum.  Should I change hb_ot_name_t to be the enum type, or stay "unsigned int"?

behdad:
> Added the enum. Should I change hb_ot_name_t to be the enum type, or stay "unsigned int"?

We already have API that returns nameIDs using "hb_name_id_t *out".  So, this is not the safest change.  But that API is also unused by everyone right now.

We already define hb_script_t as an enum where we expect non-enumerated values to be used as well.  That's legitimate in C.  Not sure about C++.  nameIDs are slightly different in that non-enumerated ones are more common?

behdad:
I think I like to rename `hb_name_id_t` to `hb_ot_name_id_t`.  That's a API-breakage, but since no one uses that API so far and it's very recent, I don't think it's an issue.

behdad:
Maybe we don't care about subtags.  It's not clear what the desired behavior is...  Is good if we add subtags that distinguish script. But for regions, say, if user asks for fa-af and there's only fa available we should return it.

behdad:
> if user asks for fa-af and there's only fa available we should return it.

This is similar to what we do for ot-tags already.  Do we need a couple functions for match-level of two language tags?  Or just implement internally here?

brawer:
Regarding `ltag`, note that the table may contain a mix of language and script codes (according to the spec, and also in practice in Apple’s shipping fonts). Thus, if an `ltag` entry consists of four letters, it needs to be prefixed with `und-` to form a valid BCP47 code. For example, if `ltag` contains `Hans`, it should be converted to `und-Hans`, wherease `sr-Latn` needs of course no prefix. Not sure if that’s already handled, but I couldn’t see it in the current code.

behdad:
> Regarding `ltag`, note that the table may contain a mix of language and script codes (according to the spec, and also in practice in Apple’s shipping fonts). Thus, if an `ltag` entry consists of four letters, it needs to be prefixed with `und-` to form a valid BCP47 code. For example, if `ltag` contains `Hans`, it should be converted to `und-Hans`, whereas `sr-Latn` needs of course no prefix. Not sure if that’s already handled, but I couldn’t see it in the current code.

It doesn't.  This is where it should be handled: https://github.com/harfbuzz/harfbuzz/pull/1254/files#diff-fc989e6dfa921e399954f43b75b091c0R68

behdad:
> Regarding `ltag`, note that the table may contain a mix of language and script codes (according to the spec, and also in practice in Apple’s shipping fonts). Thus, if an `ltag` entry consists of four letters, it needs to be prefixed with `und-` to form a valid BCP47 code. For example, if `ltag` contains `Hans`, it should be converted to `und-Hans`, whereas `sr-Latn` needs of course no prefix. Not sure if that’s already handled, but I couldn’t see it in the current code.

It doesn't.  This is where it should be handled: https://github.com/harfbuzz/harfbuzz/pull/1254/files#diff-fc989e6dfa921e399954f43b75b091c0R68

> If HarfBuzz _really_ wants to do language matching (“language subtag handling”), have a look at the Go implementation which is quite cleanly written, and works on small data tables. See [blog post](https://blog.golang.org/matchlang) and [source code](https://github.com/golang/text/tree/master/language). ICU has an implementation too, which is more involved than Go. But my personal recommendation would be to stay away from this; it seems to go beyond the scope of HarfBuzz, and getting it correct is not trivial.

Thanks for the links.  Okay, I'm convinced we don't want to be in this business.  I know Akira TAGOH wrote a library for this, can't find it right now.

khaledhosny:
> I know Akira TAGOH wrote a library for this, can't find it right now.

https://bitbucket.org/tagoh/liblangtag/overview

brawer:
> the non-standard omission of the language subtag is described in the `Meta` spec.

Just checked the system fonts on macOS 10.14. They contain the following language tags in `ltag` tables: `en`, `fa`, `pl`, and `zh-Hant`. All in standard BCP-47 syntax; sorry about my confusion of `ltag` with `Meta`.

khaledhosny:
O think treating null language as “en” makes sense as the English not the localized names is what most applications want by default so makes the common case slightly simpler.

khaledhosny:
We can make `hb-name` tool that takes name id (or even human readable aliases) and language and returns the name, just like the API do. It would also have an option to dumb all the names like what `test-name-table` is doing now.

behdad:
> O think treating null language as “en” makes sense as the English not the localized names is what most applications want by default so makes the common case slightly simpler.

Agreed.  Done.

Also need docs for new API and hooking them up in the docs.

behdad:
> We can make `hb-name` tool that takes name id (or even human readable aliases) and language and returns the name, just like the API do. It would also have an option to dumb all the names like what `test-name-table` is doing now.

Right.  Any idea what to do with newlines in the name?

behdad:
One last question before this goes into a release.  Should I rename `hb_name_id_t` to `hb_ot_name_id_t`?  It affects two APIs that went out in 2.0.0, but I'm sure they are unused.

khaledhosny:
Is there anything OT specific about the name ids (not sure I fully grasp the OT/non-OT split in the API naming).

behdad:
> Is there anything OT specific about the name ids (not sure I fully grasp the OT/non-OT split in the API naming).

Do those name-id numbers make sense for Type1, BDF, PCF, ...? No.

khaledhosny:
Right, in my mind non-OT meant AAT/Graphite etc. I have no strong opinion either way, but probably `hb_ot_name_id_t` makes more sense.

behdad:
> Right, in my mind non-OT meant AAT/Graphite etc.

We refer to GSUB/GPOS as ot-layout.  To the script logic as ot-shape.  Use OT as a encompassing of OpenType and extensions...

jfkthame:
Agree that hb_ot_name.... seems appropriate.

PR #1267, opened 2018-10-18
[test/shape-fuzzer] fail on timeout and ubsan errors

ebraminio:


behdad:
Explain what this is please?

ebraminio:
the script wasn't failing appropriately when a test caused timeout or printed ubsan error, I fixed it by adding a timeout and checking for stdout result.

PR #1273, opened 2018-10-18
Prohibit visually ambiguous vowel sequences in the USE shaper

dscorbett:


behdad:
Oh cool, I didn't know about that table in USE....

I'm uneasy about relying on that document's formatting.  Can we through the table in a file in src/ProhibitedVowelSequences.txt or something, with a dated header instead?  Note that that would be consistent with how we want to move the UISC/UIPC overrides into files as well.

Also, I think the function generated should be called from trivial preprocess_text_indic and preprocess_text_use intead of being ues as those directly.  Or maybe not.

behdad:
Also, please include the generated source file into git.

PR #1277, opened 2018-10-19
Fix hb_map_is_empty logic and write test for hb_map_t helpers

ebraminio:


src/hb-map.hh
```diff
@@ -172,7 +172,7 @@ struct hb_map_t
 
   inline bool is_empty (void) const
   {
-    return population != 0;
+    return population == 0;
```
ebraminio: Or that was intended?

ebraminio:
As Behdad

Issue #844, opened 2018-02-27
Code style change?


ebraminio:
I guess talking about code style on established/grown up projects like harfbuzz is more like a joke but I still try. Maybe we can discuss about it to make the development a bit easier? I know such change will ruin `git blame` but I like to have some place to discuss that so if we want to apply such thing, at least we do it once.

* [vscode](https://github.com/Microsoft/vscode/issues/44438) and Atom editors don't support configuring tab sizes (without change indention style) and Rod and Garret also don't use them always AFAICS. Maybe we can remove their use? Have a look at this https://youtu.be/99Zacm7SsWQ?t=329 also
* Removing space before function call.
* Not putting starting curly braces on new line, to resemble more Java code style! Have a look at the interesting case of JS https://youtu.be/hQVTIJBZook?t=1837 :)

khaledhosny:
Alternatively, we can look into using `clang-format` and making a format file that matches the existing style, so people can just use it to format the code before committing.

rsheeter:
I'm skeptical it makes sense to _change_ the style of a mature project but making it easier to follow would be great. +1 for a clang-format rig.

behdad:
Don't think will happen.

The tab vs space thing is weird not supported by those editors. This is the common style across many common Open Source C libraries.

ebraminio:
No need for other crazy changes, I can go without them, but just no tabs, what @waywardmonkeys said on #1279

ebraminio:
> I can go without them

Not a personal opinion but the mentioned editors requirement, you can see also that every new contributor have the issue (in addition to space before the calls). Khaled said we can alternatively build `clang-format` config but I couldn't find a way to explain mixed space-tabs on that formatter, maybe I didn't investigate enough however.

behdad:
Stupid request: can you please draft a vim tagline or whatever can be added to project to make it do what you propose?

behdad:
> Stupid request: can you please draft a vim tagline or whatever can be added to project to make it do what you propose?

https://andrew.stwrt.ca/posts/project-specific-vimrc/

@anthrotype can we add this for fonttools as well?

behdad:
What's the proposed indentation?  Just expand all tabs?

ebraminio:
[...]

PR #1279, opened 2018-10-19
Replace tabs with spaces to avoid misleading indentation.

waywardmonkeys:
clang-tidy's misleading indentation check triggers on these lines
due to mixed tabs and spaces, so fix that so that the check can
be clean.


ebraminio:
It is #844, I don't like it also but not going to be changed soon :(

src/hb-aat-layout-kerx-table.hh
```diff
@@ -149,9 +149,9 @@ struct KerxSubTableFormat1
       if (flags & Push)
       {
         if (likely (depth < ARRAY_LENGTH (stack)))
```
ebraminio: This should use tab however.

PR #1283, opened 2018-10-19
Cygwin fixes and CI build

khaledhosny:
DirectWrite support failed to build, so I disabled it, not sure we care about building it on Cygwin.

Gobject-Introspection failed on x86 build only, hard to debug without and actual Windows machine, so also disabled for now.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,9 +28,19 @@ environment:
       MINGW_CHOST: i686-w64-mingw32
       MSYS2_ARCH: i686
 
+    - compiler: cygwin
+      CYGWIN_PREFIX: C:\Cygwin64
+      CYGWIN_ARCH: x86_64
+   # Lots of test failures here!
+   #- compiler: cygwin
+   #  CYGWIN_PREFIX: C:\Cygwin
+   #  CYGWIN_ARCH: x86
+
+
 install:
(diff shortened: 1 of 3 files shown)


ebraminio:
> DirectWrite support failed to build, so I disabled it, not sure we care about building it on Cygwin.

Can you look again why? It works on msys2.

khaledhosny:
> Can you look again why? It works on msys2.

See https://ci.appveyor.com/project/khaledhosny/harfbuzz/builds/19649313/job/6i6x0liton5epcdr#L1335

ebraminio:
Aha, the trick is here https://github.com/harfbuzz/harfbuzz/blob/9be8062b4d241f02bb96df436e98f10f08bda4f8/appveyor.yml#L49 please apply it to the bot. Thanks

ebraminio:
Of course if you see worth the hassle (not sure if curl available by default)

khaledhosny:
Lets see if the current build will succeed or not first, it is taking a while already (why running tests on AppVeyor is so slow?)

ebraminio:
Thank you Khaled. Not only it is useful as a CI bot, it is also useful as a how to compile and develop harfbuzz on Cygwin, I like this aspect of CI bot scripts also. Thanks

PR #1284, opened 2018-10-20
Make more 'coords' params const.

waywardmonkeys:


behdad:
How did you notice these ones?

waywardmonkeys:
I ran `clang-tidy` with the `readability-non-const-parameter` check which pointed out a couple of them ... I then manually found the others based on that. That check isn't perfect and sometimes has false positives (which I haven't reported), but it is useful at times!

behdad:
Nice.  Thanks.

ebraminio:
Great. Can you give us exact command you use? I like to integrate it to some bot (just for information). Do you have experience on making `.clangformat` configuration files also?

ebraminio:
There is https://clangformat.com/ also but don't know how to describe the format harfbuzz has, indention with 2 spaces but converting 8 spaces to tabs, are you aware of some project with similar code style having `.clang-format` Behdad?

behdad:
No automatic formatting please.

PR #1016, opened 2018-05-01
Minimal API for COLR/CPAL

khaledhosny:
WIP, the tests do not pass (table in cpal-v1.ttf fails sanitization). Here is a LibreOffice using it (WIP as well):
image

PR #1287, opened 2018-10-20
Expose API for COLR/CPAL

ebraminio:
Concludes @brawer #244 and @khaledhosny #1016 and my work


ebraminio:
https://github.com/harfbuzz/harfbuzz/blob/e42cd58c997adafca5517faa9aacf651b90520f5/src/hb-buffer.h#L102

So the convention is to keep them lowercase, I still like to know however is the preference is `hb_ot_color_t` or `hb_color_t`.

ebraminio:
PTAL

ebraminio:
Or let me see if anything can be improced on the tests first.

behdad:
Rename to `hb_color_t` and move to hb-common.h.  Also, we don't seem to expose table tags any more.  There' not much use to them.  The has_data() is enough.

behdad:
I guess finish it off and we merge and then I'll touch up.

ebraminio:
PTAL

ebraminio:
Exposed ` hb_ot_color_get_palette_flags` and a new `hb_ot_color_get_palette_entry_name_id`, feel free to disable them again but with those the API will become complete.

ebraminio:
Please note that the test has full coverage of the exposed APIs and dump-emoji is able to dump all the emojis available on `Segoe UI Emoji` as a manual testing.

ebraminio:
Please note that palette index and color index are effectively the same thing but conceptually can be thought a little different, palette index is an index to a palette and the term should be used around `cpal` table and its API but a color index is an index stored in a `colr` record which effectively will indicate to a palette index at the index, so we have `hb_ot_color_get_palette_entry_name_id` but `hb_ot_color_get_color_layers` returns `color_indices`, which basically will refer to a specific palette color.

khaledhosny:
I don’t think palette index and color index are the same thing; a font can have more than one palette and the palette index will be used to access a specific palette. Each palette has (the same) number of colors, and a color index will be used to access a specific color inside the palette.

ebraminio:
I mean something similar also, thanks for the cleanup

ebraminio:
(wrong tap on mobile...)

khaledhosny:
They show now, but my previous comment does not always show. GitHub seems drunk.

ebraminio:
CI bots couldn't find the new commits so I guess Behdad and maybe ML's
commits digest sender are using the old address, behdad/harfbuzz and that
somehow confused GitHub.

khaledhosny:
GitHub seems to have been suffering  some major service outage today: https://blog.github.com/2018-10-21-october21-incident-report. Also I seem to have accidentally merged this PR, so future changes should go to master. Oh, and automatic triggering of CI builds seems to still be disabled.

behdad:
I finally finished reviewing and touching this up.  A few notes on things that could be done to make landing this easier.

- We should ALWAYS design API from the user's point of view first, then implement the API to match the table data.  NEVER should the API be designed from the inside out based on what the font format provides.  A pre-requisite of this is that one needs to first FULLY understand the functionality that the table provides, with all its limitations and corner cases, then try to discard the implementation details and capture the core of it in the API.  Same discussion as the name table API...  We saved on exposing a lot of details there.  Same here.

Things like "cpal" or "colr" don't belong in our API.  User doesn't know what those are.  And *shouldn't* need to know to use HarfBuzz.  Same thing about terms like "palette_entry" when one can simply say "color".

- Other thing I did in renaming the API was grouping.  Now we have this:  There's `hb_ot_color_has_palettes`, and if yes, then all the palette API is under `hb_ot_color_palette_...`.   The COLR API is under `hb_ot_color_glyph_get_layers`, and other tables will add more `hb_ot_color_glyph_get_*` functions.  Where these are implemented in the font is irrelevant to user and fully hidden.

- No shortcuts.  If something feels like you are being lazy, don't do it.  Do it properly when you are not feeling lazy.  PARTICULARLY so if it leaks in the API.  Do It Right.

- I had to do multiple hours and more than 20 commits of work on this.  Would have taken shorter to write from the scratch probably.  The tests were INVALUABLE.  But the implementation got more in the way, given that most of it needed review and tweak.  It's the responsibility of whoever writes the a piece of code in the first place to write the best they can.  Half-done code is more work to finish than starting from scratch.

Anyway.  Glad this is in.  Thanks.  Next maybe either CBLC or SVG.  SVG is less work but more API design.  Let's design the SVG API.  Read the spec fully to understand the complexities (multiple glyphs share the same SVG doc...).

Cheers

behdad:
One thing I forgot is effective use of null objects.  I know that feels like magic.  But when done right, a lot of "if (...) return ..." codepaths are not needed, and that simplifies the code.

Other point, the hb-ot-color.cc methods should all be short.  The actual work should almost always happen in tables themselves.  That is exactly the case now if you check.

ebraminio:
Thanks for the fixes and notes. Looks great now

PR #1291, opened 2018-10-21
[use] Fix reordering

### LINKED ISSUE #1235: USE glyph reordering is broken for some scripts
`_hb_ot_complex_shaper_use` does not reorder glyphs correctly in all cases. This hasn’t come up before because, by chance, these cases are not relevant to most of the scripts passed to the USE.

It reorders pre-base vowel signs to just before the last base of the cluster, or to just after the last halant of the cluster, whichever comes last. For example, the cluster ⟨half-form base pre-base-vowel⟩ will reorder to ⟨half-form pre-base-vowel base⟩. It should instead reorder pre-base vowel signs to the beginning of the cluster, or to just after the last halant of the cluster, whichever comes last. I am fairly sure that is what [the USE spec’s section on reordering](https://docs.microsoft.com/en-us/typography/script-development/use#glyph-reordering) means: when it says “base consonant” it probably means “maximal sequence of non-reordering non-halant glyphs within a cluster”, not “single glyph with class B or GB”. See MicrosoftDocs/typography-issues#104.

A specific example is the sequence 𑒞𑓂𑒏𑒱 <i>tki</i> <1149E 114C2 1148F 114B1> in [Noto Sans Tirhuta](https://github.com/googlei18n/noto-fonts/blob/b20c9c07139b22eb38013b1f80f9a5222c6058c6/phaseIII_only/hinted/ttf/NotoSansTirhuta/NotoSansTirhuta-Regular.ttf). Before reordering, the glyphs are ⟨<i>ta</i><sub>half</sub> <i>ka</i> halant <i>i</i>⟩. (That there is a halant there is an idiosyncrasy of Tirhuta; `is_halant` returns `false` for it and HarfBuzz treats it as a base.) The vowel sign gets reordered to before the preceding base, i.e. the halant, which looks wrong.

𑒞𑓂𑒏𑒱

Contrast this with HarfBuzz’s Indic shaper. Here are क्खि in Noto Sans Devanagari with the current HarfBuzz and with HarfBuzz modified to map 'dev2' to the USE. I don’t think moving to the USE is supposed to break fonts this much or require so much refactoring to fix.

क्खि क्खि

This is all true for the reph too, except for the direction of the reordering.

dscorbett:
Fixes #1235.


KrasnayaPloshchad:
CC @eroux

src/hb-ot-shape-complex-use.cc
```diff
@@ -451,19 +452,37 @@ reorder_syllable (hb_buffer_t *buffer, unsigned int start, unsigned int end)
...
   if (info[start].use_category() == USE_R && end - start > 1)
   {
-    /* Got a repha.  Reorder it to after first base, before first halant. */
+    /* Got a repha.  Reorder it towards the end, but before the first post-base
+     * glyph. */
     for (unsigned int i = start + 1; i < end; i++)
-      if ((FLAG_UNSAFE (info[i].use_category()) & (BASE_FLAGS)) || is_halant (info[i]))
+    {
+      bool is_post_base_glyph = (FLAG_UNSAFE (info[i].use_category()) & POST_BASE_FLAGS) ||
+				info[i].use_category() == USE_VBlw ||
```
behdad: Not using flags for these because they are over 32?  Maybe we can add 64bit flags?

dscorbett: Yes, that’s why. A wider `FLAG_UNSAFE` would be convenient. Do you want me to add one for this PR?

behdad: Yes please.  Call it FLAG64 or whatever needs to be.

PR #1295, opened 2018-10-22
Build tests conditionally

donny-dont:
Adds an option in CMake for building harfbuzz tests. Turns off the subset library which only seems to be used within the tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,7 @@ if (WIN32)
   option(HB_HAVE_DIRECTWRITE "Enable DirectWrite shaper backend on Windows" OFF)
 endif ()
 option(HB_BUILD_UTILS "Build harfbuzz utils, needs cairo, freetype, and glib properly be installed" OFF)
+option(HB_BUILD_TESTS "Build harbuzz tests" ON)
 if (HB_BUILD_UTILS)
   set (HB_HAVE_GLIB ON)
   set (HB_HAVE_FREETYPE ON)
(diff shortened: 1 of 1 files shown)


ebraminio:
Oh subset is not for inside use only, this is totally my fault on making iOS build possible, don't spread flags used on HB_IOS as they should be removed when our iOS build bot is ready. Also I believe better to not go for disabling test on build, I intentionally didn't expose it.

What is your usecase for using CMake?

donny-dont:
My mistake I didn't see the subset library being installed anywhere within the CMake code and it was only referenced by the tests.

We use harfbuzz on the [PlayStation](https://doc.dl.playstation.net/doc/ps4-oss/) and I was just trying to land some fixes we've had to make locally to build WebKit. The subset library isn't used by WebKit, the ICU one is hence my earlier issue, so I didn't want to build it. Tests would be nice to turn off because CTest has no concept on how to run on our platform.

I'm fine with keeping our patches local if this is a non starter. Either way thanks for the quick replies.

CMakeLists.txt
```diff
@@ -47,6 +47,7 @@ if (WIN32)
   option(HB_HAVE_DIRECTWRITE "Enable DirectWrite shaper backend on Windows" OFF)
 endif ()
 option(HB_BUILD_UTILS "Build harfbuzz utils, needs cairo, freetype, and glib properly be installed" OFF)
+option(HB_BUILD_TESTS "Build harbuzz tests" ON)
```
ebraminio: "harfbuzz" :) Also feel free to apply naming more relevant to your internal use :)

donny-dont:
The vita hasn't gotten any work in awhile but it would be nice to get the PS4 on the latest and greatest.

So should I be making a `HB_BUILD_SUBSET` as well or is just the `HB_BUILD_TESTS` fine? It seems like I made the wrong assumption here with subset. Happy to do any cleanup while I'm here.

ebraminio:
Exposing `HB_BUILD_SUBSET` is also OK :) Do whatever cleanup/renaming needed and makes more sense :)

ebraminio:
Fixed. Let me know about other needed things :)

PR #1296, opened 2018-10-22
Make tracking optout possible using 'trak' ot feature

ebraminio:
So '--features=-trak' or 'font-feature-settings: 'trak' 0;' can prevent
applying of tracking if used.

### DIFF
--- a/src/hb-ot-shape.cc
+++ b/src/hb-ot-shape.cc
@@ -116,7 +116,8 @@ hb_ot_shape_planner_t::compile (hb_ot_shape_plan_t &plan,
     plan.fallback_mark_positioning = true;
 
   /* Currently we always apply trak. */
-  plan.apply_trak = hb_aat_layout_has_tracking (face);
+  plan.apply_trak = hb_aat_layout_has_tracking (face) &&
+		    plan.map.needs_fallback (HB_TAG ('t','r','a','k'));
 }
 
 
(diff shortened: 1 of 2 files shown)


src/hb-ot-shape.cc
```diff
@@ -177,12 +178,15 @@ hb_ot_shape_collect_features (hb_ot_shape_planner_t          *planner,
   /* Random! */
   map->enable_feature (HB_TAG ('r','a','n','d'), F_RANDOM, HB_OT_MAP_MAX_VALUE);
 
-  map->enable_feature (HB_TAG('H','A','R','F'));
+  /* Tracking. */
+  map->enable_feature (HB_TAG ('t','r','a','k'), F_HAS_FALLBACK);
```
behdad: Need some comment why we are adding this.

src/hb-ot-shape.cc
```diff
@@ -116,7 +116,8 @@ hb_ot_shape_planner_t::compile (hb_ot_shape_plan_t &plan,
     plan.fallback_mark_positioning = true;
 
   /* Currently we always apply trak. */
-  plan.apply_trak = hb_aat_layout_has_tracking (face);
+  plan.apply_trak = hb_aat_layout_has_tracking (face) &&
+		    plan.map.needs_fallback (HB_TAG ('t','r','a','k'));
```
behdad: This change is not good.  It will disable trak table if GSUB or GPOS has trak.  Not sure which one.

behdad:
Merged manually and touched up.  Thanks!

PR #1306, opened 2018-10-23
[svg] Add public API

ebraminio:
* hb_ot_color_has_svg
* hb_ot_color_glyph_reference_blob_svg

### DIFF
--- a/src/hb-open-type.hh
+++ b/src/hb-open-type.hh
@@ -523,7 +523,6 @@ struct ArrayOf
     ::qsort (arrayZ, len, sizeof (Type), Type::cmp);
   }
 
-  private:
   inline bool sanitize_shallow (hb_sanitize_context_t *c) const
   {
     TRACE_SANITIZE (this);
--- a/src/hb-ot-color-svg-table.hh
+++ b/src/hb-ot-color-svg-table.hh
@@ -41,12 +41,22 @@ namespace OT {
 struct SVGDocumentIndexEntry
 {
   friend struct SVG;
+  friend struct SVGDocumentIndex;
 
-  inline bool sanitize (hb_sanitize_context_t *c, const void* base) const
+  inline int cmp (hb_codepoint_t g) const
+  { return g < startGlyphID ? -1 : g > endGlyphID ? 1 : 0; }
+
+  static int cmp (const void *pa, const void *pb)
+  {
+    const hb_codepoint_t *a = (const hb_codepoint_t *) pa;
+    const SVGDocumentIndexEntry *b = (const SVGDocumentIndexEntry *) pb;
+    return b->cmp (*a);
+  }
+
+  inline bool sanitize (hb_sanitize_context_t *c) const
   {
     TRACE_SANITIZE (this);
-    return_trace (c->check_struct (this) &&
-		  (base+svgDoc).sanitize (c, svgDocLength));
+    return_trace (c->check_struct (this));
   }
 
   protected:
@@ -57,7 +67,7 @@ struct SVGDocumentIndexEntry
   LOffsetTo<UnsizedArrayOf<HBUINT8>, false>
 		svgDoc;		/* Offset from the beginning of the SVG Document Index
 				 * to an SVG document. Must be non-zero. */
-  HBUINT32 svgDocLength;	/* Length of the SVG document.
+  HBUINT32	svgDocLength;	/* Length of the SVG document.
 				 * Must be non-zero. */
   public:
   DEFINE_SIZE_STATIC (12);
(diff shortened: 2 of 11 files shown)


ebraminio:
Now its hard part, deciding on what API should look like :)

ebraminio:
The reasoning behind current design, if a client really cares about performance it can ask for a glyph SVG source but caches supporting range of the blob, if not, passing NULL.

Self note: Move to runtime sanitization.

ebraminio:
Hey folks, may I request your participation?

What the final API should look like? Do we like to have a unified access to all the three tables (CBDT, sbix, SVG) or separated ones?

If we want former, we should prioritize which tables we should pick first which I guess not that common case as who ships two/three tables of color font standard in one font. This is a possiblity if 

    

    hb_blob_t hb_ot_color_create_blob_for_glyph
      (hb_font_t *font, /* so we use its ppem, as below [1] */
       hb_codepoint_t glyph,
       hb_tag_t *preferred_types, /* IN, 'svg ', 'png ', or 'jpg '/'png '/'tiff' */ see [2] */
               /* maybe an enum like hb_ot_color_graphic_type instead */
       unsigned int preferred_types_length, /* size of above array */
       hb_tab_t *result_type /* OUT, result time */

[1]: https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ot-color-cbdt-table.hh#L414 and https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ot-color-sbix-table.hh#L75
[2]: https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ot-color-sbix-table.hh#L51

And what about ebdt? Shouldn't be better to consider something useful for monochrome images in fonts (ebdt) now that we are supporting cbdt?

Or we want to go for separated one, how about the current patch, how about renaming `hb_ot_color_has_svg` to `hb_ot_color_has_scalable_graphics` and `hb_ot_color_scalable_create_blob_for_glyph`? How about dropping start_glyph and end_glyph from the current signature, I didn't see a use for it in SVG fonts I had access.

ebraminio:
What if we still use hb_face_t but passing ppem also? What about dropping `preferred_types` and returning blob from `SVG`, `sbix`, `cbdt` and maybe `ebdt`, in this order? I've insisted on using _create_ so client know should release the blob at some point, I saw such on ObjectiveC API, is that unnecessary?

src/hb-open-type.hh
```diff
@@ -523,7 +523,6 @@ struct ArrayOf
     ::qsort (arrayZ, len, sizeof (Type), Type::cmp);
   }
 
-  private:
   inline bool sanitize_shallow (hb_sanitize_context_t *c) const
```
ebraminio: To use `sanitize_shallow` in `SVG`

src/hb-ot-color-svg-table.hh
```diff
@@ -67,11 +77,22 @@ struct SVGDocumentIndex
...
+					     SVGDocumentIndexEntry::cmp);
+    return likely (rec && glyph_id <= rec->endGlyphID) ? *rec : Null(SVGDocumentIndexEntry);
+  }
+
   inline bool sanitize (hb_sanitize_context_t *c) const
   {
     TRACE_SANITIZE (this);
     return_trace (c->check_struct (this) &&
-		  entries.sanitize (c, this));
+		  entries.sanitize_shallow (c));
```
ebraminio: To indicate we are just shallow sanitizing it.

src/hb-ot-color-svg-table.hh
```diff
@@ -40,13 +40,27 @@ namespace OT {
...
+    const hb_codepoint_t *a = (const hb_codepoint_t *) pa;
+    const SVGDocumentIndexEntry *b = (const SVGDocumentIndexEntry *) pb;
+    return b->cmp (*a);
+  }
+
+  inline hb_blob_t* create_blob (hb_blob_t *svg_blob, unsigned int index_offset) const
+  {
+    if (svgDocLength == 0) return hb_blob_get_empty ();
+    return hb_blob_create_sub_blob (svg_blob, (unsigned int) svgDoc + index_offset,
+				    svgDocLength);
```
ebraminio: No more check needed as hb_blob_create_sub_blob implementation

behdad:
Regarding returning start_glyph/end_glyph, I kinda like it, but maybe it's okay to initially go without it.

behdad:
Okay let's continue in https://github.com/harfbuzz/harfbuzz/issues/1316

ebraminio:
> Regarding returning start_glyph/end_glyph, I kinda like it, but maybe it's okay to initially go without it.

I went initially as can be seen, but as `dupe` on `sbix`, I see it now just an implementation detail.

ebraminio:
I think it is better to review and merge as not lots of things can go wrong with it specificly.

behdad:
> I think it is better to review and merge as not lots of things can go wrong with it specificly.

Ugh. Didn't we agree to discuss the API in https://github.com/harfbuzz/harfbuzz/issues/1316 first?!  There's plenty to decide there.  I don't want to rush anything in anymore.  Even the colr/cpal should have had gone in after we fleshed out the entire color API.  So, get to https://github.com/harfbuzz/harfbuzz/issues/1316, let's draft two simple APIs for SVG and PNG, then you can go implement it.  I don't like rushing code in.  It's horrible software engineering practice.

ebraminio:
> I don't like rushing code in.

Sure, it was a misunderstanding as I guessed the conclusion there was to SVG is separate thing.

behdad:
Yes, there's agreement to have SVG in it's own API. But let's conclude that discussion before getting any of it in. Thanks.

PR #1309, opened 2018-10-24
Appveyor speedup

khaledhosny:
Cuts a few minutes here and there, should make things a bit faster.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -57,10 +57,10 @@ build_script:
   - 'if "%compiler%"=="msvc" if not "%platform%"=="ARM" ctest --output-on-failure -C %configuration%'
 
   - 'if "%compiler%"=="msys2" C:\msys64\usr\bin\bash -lc "curl https://raw.githubusercontent.com/mirror/mingw-w64/023eb04c396d4e8d8fcf604cfababc53dae13398/mingw-w64-headers/include/dwrite_1.h > %MINGW_PREFIX%/%MINGW_CHOST%/include/dwrite_1.h"'
-  - 'if "%compiler%"=="msys2" C:\msys64\usr\bin\bash -lc "cd $APPVEYOR_BUILD_FOLDER; PATH=$PATH:/mingw64/bin:/mingw32/bin; ./autogen.sh --with-uniscribe --with-freetype --with-glib --with-gobject --with-cairo --with-icu --with-graphite2 --with-directwrite --build=%MINGW_CHOST% --host=%MINGW_CHOST% --prefix=%MINGW_PREFIX%; make; make check || .ci/fail.sh"'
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -38,30 +38,29 @@ environment:
 
 
 install:
-  - 'if "%compiler%"=="msys2" C:\msys64\usr\bin\bash -lc "pacman --force --noconfirm -Sy && pacman --noconfirm --force -S pacman-mirrors && pacman --force -Syu --noconfirm"'
+# - 'if "%compiler%"=="msys2" C:\msys64\usr\bin\bash -lc "pacman --noconfirm --force -Sy && pacman --noconfirm --force -S pacman-mirrors && pacman --force -Syu --noconfirm"'
+  - 'if "%compiler%"=="msys2" C:\msys64\usr\bin\bash -lc "pacman --noconfirm --force -S --needed mingw-w64-$MSYS2_ARCH-{gcc,freetype,cairo,icu,gettext,gobject-introspection,gcc,gcc-libs,glib2,graphite2,pkg-config,python2}"'
   - C:\msys64\usr\bin\bash -lc "pacman --noconfirm -S mingw-w64-x86_64-ragel"
```
ebraminio: HarfBuzz theoretically can be built without ragel now I guess specially on CMake.

appveyor.yml
```diff
@@ -75,5 +74,8 @@ notifications:
     on_build_failure: true
     on_build_status_changed: true
 
+# Do not build feature branch with open Pull Requests
+skip_branch_with_pr: true
```
ebraminio: Nice. Thanks

appveyor.yml
khaledhosny: Last I checked, this was intermittently failing with autotools, might be file timestamp issue.

ebraminio: Oh ok. VS/CMake ones will be OK I guess, sorry for overloading your change. Thanks

behdad:
Thanks!  Is -j okay or will worse things? Fine with me.

khaledhosny:
I’m not sure either, builds seem to vary by several minutes on each job between runs, so it is hard to measure if it is making things better or worse.

behdad:
Maybe just `-j4` or something.  Not sure how easy it is to overwhelm the Windows scheduler...

khaledhosny:
This [thread](https://help.appveyor.com/discussions/questions/11179-how-many-cores-and-threads-can-be-used-in-free-appveyor-build) suggests that there are two cores in the VM, so I guess `-j2` then.

behdad:
Even with one core, -j2 runs faster than single-process.  That's because IO and CPU tasks will be interleaved, instead of stalled for each other.

behdad:
-j3 maybe then.

khaledhosny:
Trying that now.

khaledhosny:
`-j3` seems to be three minutes faster per each CI job than `-j2`.

PR #1311, opened 2018-10-24
[appveyor] Limit make to three jobs

khaledhosny:
https://github.com/harfbuzz/harfbuzz/pull/1309#issuecomment-432778270

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -57,10 +57,10 @@ build_script:
   - 'if "%compiler%"=="msvc" if not "%platform%"=="ARM" ctest --output-on-failure -C %configuration%'
 
   - 'if "%compiler%"=="msys2" C:\msys64\usr\bin\bash -lc "curl https://raw.githubusercontent.com/mirror/mingw-w64/023eb04c396d4e8d8fcf604cfababc53dae13398/mingw-w64-headers/include/dwrite_1.h > %MINGW_PREFIX%/%MINGW_CHOST%/include/dwrite_1.h"'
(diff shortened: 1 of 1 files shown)


khaledhosny:
All look fine now, except Cygwin; it takes 10 minutes to just install the packages! I’m OK with disabling it again, one failure every ~5 years is an acceptable rate IMO.

behdad:
Agreed.

khaledhosny:
Next would to move MingW builds to another service and limit AppVeyor to MSVC builds.

ebraminio:
> Next would to move MingW builds to another service and limit AppVeyor to MSVC builds.

I've enabled Azure Pipeline CI again https://dev.azure.com/harfbuzz/harfbuzz/_build?definitionId=3 and invited you to it.
It config is here https://github.com/harfbuzz/harfbuzz/blob/master/azure-pipelines.yml
But it is only compiling vcpkg itself https://dev.azure.com/harfbuzz/harfbuzz/_build/results?buildId=9&view=logs
Its script is in PowerShell, don't if possible to run cmd instead and I have no preference on whether msys2 builds should be moved there or msvc or both, so apply any one of them if you like and had time for.

Thanks

khaledhosny:
I was thinking of cross-compiling with MingW from Linux, that should be faster I guess.

ebraminio:
Hmmm, Fedora packages can be used there I guess, CircleCI can provide that. I will try finish Azure one for MSVC

ebraminio:
> Fedora packages can be used there I guess

Or a custom Docker image, reducing compile time of third party libraries.

PR #1313, opened 2018-10-25
Remove some use of Crap in hb-set.

bungeman:
When hb_set_t::page_for_insert needs to insert at the end of the page_map
it ends up evaluating '&page_map[i + 1]' which has hb_vector return an
lvalue of a Crap so that nothing can be moved to its address. This turns
into issues with ThreadSanitizer on Crap when two threads modify different
hb_set_t instances. This can be avoided by using '&page_map[i] + 1'
instead.

### DIFF
--- a/src/hb-set.hh
+++ b/src/hb-set.hh
@@ -669,7 +669,7 @@ struct hb_set_t
 	return nullptr;
 
       pages[map.index].init0 ();
-      memmove (&page_map[i + 1], &page_map[i], (page_map.len - 1 - i) * sizeof (page_map[0]));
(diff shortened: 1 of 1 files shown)


bungeman:
Noticed at https://chromium-review.googlesource.com/c/chromium/src/+/1297256/4 with ThreadSanitizer complaining about this in https://logs.chromium.org/logs/chromium/buildbucket/cr-buildbucket.appspot.com/8931830262547183456/+/steps/webkit_unit_tests__with_patch_/0/stdout (search for ThreadSanitizer).

behdad:
Thanks!

PR #1314, opened 2018-10-25
Avoid overflow in bfind.

bungeman:
https://ai.googleblog.com/2006/06/extra-extra-read-all-about-it-nearly.html


bungeman:
Happened to notice this while investigating https://github.com/harfbuzz/harfbuzz/pull/1313 .

behdad:
Ha!  Thanks. I think we have a few more bsearch impls.

PR #1317, opened 2018-10-26
Limit total operations during feature closure.

garretrieger:


behdad:
While legit fonts don't use the same bytes for a script and a langsys, your code works.  But I dislike making that assumption.  I think we should have separate sets for script and langsys.  wdyt?

Also, taking offset from start of table is misleading, in that we might also receive the nul object, which is outside of the table.  I suggest you handle that by first checking if number of children in object is zero and return in that case, *then* check in the set (the way you have it with offsets is fine).

Nit, I'd rename "get_gsubgpos_table" to "_get_gsubgpos".

Thanks.

src/hb-ot-layout.cc
```diff
@@ -833,13 +849,12 @@ hb_ot_layout_collect_features (hb_face_t      *face,
       if (hb_ot_layout_table_find_script (face,
 					  table_tag,
 					  *scripts,
-					  &script_index))
-        _hb_ot_layout_collect_features_languages (face,
-                                                  table_tag,
-                                                  script_index,
+					  &script_index)
+          && context.visit (&context.gsubgpos.get_script (script_index)))
```
behdad: I don't like that we are once calling get_script to check, and then call a function with script_index.  Please use the function chain to take the script object and next the langsys object, instead of indices, so we look them up once.  And I would move the "visit" check into the next function instead of before calling it.

Also, "visit" name is misleading since it suggests it does the actual visitation.  "may_visit" or something is better.  Or invert it as "visited"...

behdad:
I'm working on this.

PR #1318, opened 2018-10-27
Implement ot-color-png

ebraminio:
<s>Only sbix part for now,</s> see if the direction looks good. (on top of the other PR)


ebraminio:
CBDT is also hooked up, now only cleanup and API decisions remains.

PR #1319, opened 2018-10-27
[ot-color] Add public API for CBDT/sbix/SVG color Emojis

ebraminio:


docs/harfbuzz-sections.txt
```diff
@@ -459,8 +459,12 @@ hb_color_get_blue
 hb_color_get_green
 hb_color_get_red
 hb_ot_color_glyph_get_layers
+hb_ot_color_glyph_reference_blob_png
```
behdad: No "blob" in the name.

src/hb-ot-color-cbdt-table.hh
```diff
@@ -506,9 +506,7 @@ struct CBDT
 
     inline hb_blob_t* reference_blob_for_glyph (hb_codepoint_t  glyph_id,
 						unsigned int    requested_x_ppem,
-						unsigned int    requested_y_ppem,
-						unsigned int   *strike_x_ppem,
```
behdad: I think we still want both x/y so we choose best.  Or at least, we need the max of two.  Don't call it "requested_y_ppem".

src/hb-ot-color-cbdt-table.hh
```diff
@@ -529,26 +527,26 @@ struct CBDT
       switch (image_format)
       {
       case 17: {
-	if (strike_x_ppem) *strike_x_ppem = x_ppem;
-	if (strike_x_ppem) *strike_y_ppem = y_ppem;
+	// if (strike_x_ppem) *strike_x_ppem = x_ppem;
+	// if (strike_x_ppem) *strike_y_ppem = y_ppem;
```
behdad: Why keep these?

docs/harfbuzz-sections.txt
ebraminio: Am sorry

src/hb-ot-color-cbdt-table.hh
ebraminio: cbdt logic is what was before me on get_extents.

src/hb-ot-color-cbdt-table.hh
ebraminio: Sorry, done

src/hb-ot-color-cbdt-table.hh
behdad: > cbdt logic table matching is what was before me on get_extents.

I know.  We should improve it.

src/hb-ot-font.cc
```diff
@@ -180,7 +181,11 @@ hb_ot_get_glyph_extents (hb_font_t *font,
 			 void *user_data HB_UNUSED)
 {
   const hb_ot_face_data_t *ot_face = (const hb_ot_face_data_t *) font_data;
-  bool ret = ot_face->glyf->get_extents (glyph, extents);
+  bool ret = false;
+  if (ot_face->sbix->has_data ())
```
behdad: No has_data check needed!

You're not learning to embrace the nul objects.

ebraminio: Hehehe, right

src/hb-ot-color-sbix-table.hh
```diff
@@ -198,7 +198,7 @@ struct sbix
 
     struct PNGHeader
     {
-      HBUINT8 header[9];
+      HBUINT8 header[16];
```
behdad: Where does 16 comes from!?  8.  8 is the number!

behdad: https://github.com/behdad/color-emoji/blob/master/png.py#L81

src/hb-ot-color-sbix-table.hh
```diff
@@ -215,12 +215,13 @@ struct sbix
       hb_blob_t *blob = reference_blob_for_glyph (glyph, x_ppem, y_ppem,
 						  HB_TAG ('p','n','g',' '),
 						  &x_offset, &y_offset);
-      if (hb_blob_get_length (blob) < sizeof (PNGHeader))
+
+      const PNGHeader &header = *blob->as<PNGHeader>();
+      if (header.width == 0 && header.width == 0)
```
behdad: No!  I meant, put the return at the end of function.  The whole point is to remove error-checking "if" statements.

src/hb-ot-color-sbix-table.hh
behdad: I'm sorry.

behdad: There's 8 byte PNG header, then 8 byte IHDR chunk header:
https://github.com/behdad/color-emoji/blob/master/png.py#L69

Having 16 is fine I guess. Sorry about that.  Thanks.  Or make it signature[8] and chunkHeaderIHDR[8].

behdad: For my OCD, maybe add all the IHDR fields:
```		#  Width:              4 bytes
		#  Height:             4 bytes
		#  Bit depth:          1 byte
		#  Color type:         1 byte
		#  Compression method: 1 byte
		#  Filter method:      1 byte
		#  Interlace method:   1 byte
```

src/hb-ot-color-sbix-table.hh
ebraminio: Hehehe, I see

ebraminio: It will cause uninitialized memory read of x_offset and y_offset (fortunately tests caught it)

src/hb-ot-color-sbix-table.hh
ebraminio: Looks nice! Done. Thanks

src/hb-ot-color-sbix-table.hh
behdad: Humm. need to understand why.

Oh, get_length < sizeof.... yes, you are right.  You still need it.  The as<> only checks for zero-length.  Otherwise assumes it's sane data.  Yeah, put the size check back please.

behdad: I'm sorry. :)

ebraminio:
Merged, yay

PR #1333, opened 2018-10-30
[subset] Limit the number of scripts and langsys' in feature collection.

garretrieger:


src/hb-ot-layout-common.hh
```diff
@@ -51,6 +51,14 @@
 #define HB_CLOSURE_MAX_STAGES	32
 #endif
 
+#ifndef HB_MAX_SCRIPTS
+#define HB_MAX_SCRIPTS	1000
```
behdad: 200 here should be plenty.

src/hb-ot-layout.cc
```diff
@@ -678,6 +679,10 @@ struct hb_collect_features_context_t
 		  !s.get_lang_sys_count ()))
       return true;
 
+    script_count++;
+    if (script_count > HB_MAX_SCRIPTS)
```
behdad: `if (script_count++` please.

src/hb-ot-layout.cc
```diff
@@ -688,6 +693,10 @@ struct hb_collect_features_context_t
 		  !l.get_feature_count ()))
       return true;
 
+    langsys_count++;
+    if (langsys_count > HB_MAX_LANGSYS)
```
behdad: Ditto.

src/hb-ot-layout.cc
```diff
@@ -710,6 +719,8 @@ struct hb_collect_features_context_t
   private:
   hb_set_t visited_script;
   hb_set_t visited_langsys;
+  unsigned int langsys_count;
+  unsigned int script_count;
```
behdad: Nit. Reorder the two so follows previous lines.

src/hb-ot-layout-common.hh
behdad: Okay maybe 500 here.

behdad: And 2000 for language.

garretrieger: Done.

src/hb-ot-layout.cc
garretrieger: Done.

src/hb-ot-layout.cc
garretrieger: Done.

PR #1334, opened 2018-10-30
fix various GCC 8.1.X function pointer warnings

stooke:
Hi, and thanks for all the work on HarfBuzz.
I've been working to reduce warnings from the latest GCC, which is fairly strict.
This patch fixes some function parameter / template and other issues.
It'd be great if you consider accepting it.

### DIFF
--- a/src/hb-debug.hh
+++ b/src/hb-debug.hh
@@ -173,7 +173,7 @@ _hb_debug_msg_va (const char *what,
 
   fprintf (stderr, "\n");
 }
-template <> inline void
+template <> inline void HB_PRINTF_FUNC(7, 0)
 _hb_debug_msg_va<0> (const char *what HB_UNUSED,
 		     const void *obj HB_UNUSED,
 		     const char *func HB_UNUSED,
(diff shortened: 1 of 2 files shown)


behdad:
Not sure about the reinterpret_cast.  Why did you need it?  memset takes void*, and that should be automatic?  Even then, why cast to char*  instead of void* then?  We typically just use C-style casts.

stooke:
Thanks for merging my PR.  You're absolutely right, while my cast works, casting to void* using a C-style cast will work too.  

The warning is not that the type of the parameter is wrong, but that it's modifying a private member.
I use reinterpret_cast to show the code is doing something potentially quite dangerous, but I should have used void* instead of char* (the end result is the same, but yours is more correct).  

I'm more than happy to submit a simple PR to change this to C-style cast to void*, and I'm sorry I didn't match your style to begin with - just let me know.

ebraminio:
Yes. Feel free to do so :)

behdad:
> The warning is not that the type of the parameter is wrong, but that it's modifying a private member.

Ah, interesting.  Maybe I rewrite the code to call a method instead.  That makes sense.

PR #1344, opened 2018-10-31
Change from C++-style to C-style cast

stooke:
As discussed, I've redone my previous PR with C-style casts, and now cast to void*. 
My apologies for the trouble.


ebraminio:
>  My apologies for the trouble.

Oh no problem at all, btw what flags do you use for GCC?

src/hb-set.hh
```diff
@@ -53,8 +53,8 @@ struct hb_set_t
 
   struct page_t
   {
-    inline void init0 (void) { memset (reinterpret_cast<char*>(&v), 0, sizeof (v)); }
-    inline void init1 (void) { memset (reinterpret_cast<char*>(&v), 0xff, sizeof (v)); }
+    inline void init0 (void) { memset ((void *)(&v), 0, sizeof (v)); }
+    inline void init1 (void) { memset ((void *)(&v), 0xff, sizeof (v)); }
```
ebraminio: HarfBuzz code style apparently prefers space after casting operator, just saying for future works :)

behdad:
Not needed anymore.

stooke:
> > My apologies for the trouble.
> 
> Oh no problem at all, btw what flags do you use for GCC?

-Wall will give this warning; in your current codebase, with GCC 8 you'll get it in other places, too.
Unfortunately, GCC doesn't think some of your other structures are simple enough so it flags the memset(this...) calls.  Since I'm working with an older version of your library I haven't hit this yet.

We actually compile with -Wall -Wextra -Werror and then disable quite a few warnings in our global codebase.
I'm working to disable less warnings as I go.

ebraminio:
We are maintaining a similar list for clang https://github.com/harfbuzz/harfbuzz/blob/master/.circleci/config.yml#L123 feel free to share yours so we can add a GCC based one also :)

PR #1346, opened 2018-11-01
[feat] Expose public API

### LINKED ISSUE #1304: [aat] Expose 'feat' table
I don't think we are ever going to expose it in API.  Suggest remove.

ebraminio:
Closes #1304


src/hb-aat-layout-feat-table.hh
```diff
@@ -95,15 +94,19 @@ struct FeatureName
...
     }
-    if (is_exclusive) *is_exclusive = exclusive;
+    if (default_setting)
+    {
+      if (exclusive)
+      {
+	if (settings_count && !not_default) *default_setting = settings[0].setting;
+	else if (not_default) *default_setting = featureFlags & IndexMask;
+      }
+      else *default_setting = HB_AAT_FEATURE_NO_DEFAULT_INDEX;
```
jfkthame: This sounds like an index rather than a setting, too.

I do wonder if returning an index will be simpler? Otherwise, I'm not sure what the result for non-exclusive should be here.

src/hb-aat-layout-feat-table.hh
```diff
@@ -95,15 +94,19 @@ struct FeatureName
...
-      if (exclusive && !not_default && start_offset == 0 && len != 0)
-        records_buffer[0].is_default = true;
     }
-    if (is_exclusive) *is_exclusive = exclusive;
+    if (default_setting)
+    {
+      if (exclusive)
+      {
+	if (settings_count && !not_default) *default_setting = settings[0].setting;
+	else if (not_default) *default_setting = featureFlags & IndexMask;
```
jfkthame: IIUC, this would be the _index of the setting in the setting name array_, not the setting itself. So I think what you want here is more like

    *default_setting = settings[featureFlags & IndexMask].setting;

(hmm, that looks like it may want a range-check).

src/hb-aat-layout.h
```diff
@@ -0,0 +1,79 @@
...
+typedef uint16_t hb_aat_feature_setting_t;
+
+/**
+ * hb_aat_feature_option_record_t:
+ *
+ * Feature type record
+ *
+ * Since: REPLACEME
+ **/
+typedef struct hb_aat_feature_option_record_t
```
drott: Would it make sense to call this 
```suggestion
typedef struct hb_aat_feature_type_selector_t
```
to be consistent with what CoreText calls these?

src/hb-aat-layout.cc
```diff
@@ -301,3 +309,29 @@ _hb_aat_language_get (hb_face_t *face,
 {
   return _get_ltag (face).get_language (i);
 }
+
+/**
+ * hb_aat_get_feature_settings:
+ * @face:            a font face.
+ * @identifier:      aat feature id you are querying.
```
drott:
```suggestion
 * @identifier:      AAT feature id you are querying, for example 1 for "Ligatures" feature, 37 for the "Lower Case" feature, 38 for the "Upper Case" feature, etc.
```

src/hb-aat-layout.cc
```diff
@@ -301,3 +309,29 @@ _hb_aat_language_get (hb_face_t *face,
...
+ * @default_setting: (out): default value for the type. If it is HB_AAT_FEATURE_NO_DEFAULT_INDEX
+ *                          means non is default and it is not exclusive also.
```
drott:
```suggestion
 *                          means none is selected as default and the feature is not exclusive.
```

src/hb-aat-layout.cc
```diff
@@ -301,3 +309,29 @@ _hb_aat_language_get (hb_face_t *face,
...
+ * @start_offset:    start offset, if you are iterating
+ * @records_count:   (inout): gets input buffer size, puts number of filled one
+ * @records_buffer:  (out):  buffer of records
+ *
+ * Returns: Total number of records available for the feature.
```
drott:
```suggestion
 * Returns: Total number of feature selector records available for the feature.
```

src/hb-aat-layout.cc
drott: How would this API allow querying all features, instead of a specific known one / feature id?

src/hb-aat-layout.cc
```diff
@@ -301,3 +309,29 @@ _hb_aat_language_get (hb_face_t *face,
...
   return _get_ltag (face).get_language (i);
 }
+
+/**
+ * hb_aat_get_feature_settings:
+ * @face:            a font face.
+ * @identifier:      aat feature id you are querying.
+ * @default_setting: (out): default value for the type. If it is HB_AAT_FEATURE_NO_DEFAULT_INDEX
+ *                          means non is default and it is not exclusive also.
+ * @start_offset:    start offset, if you are iterating
```
drott: I don't fully understand what start_offset points at? What would the client iterate over?

src/hb-aat-layout-feat-table.hh
ebraminio: Oh, you are right, added with a range check. Thanks.

src/hb-aat-layout-feat-table.hh
ebraminio: > This sounds like an index rather than a setting, too.

Right.

> I do wonder if returning an index will be simpler?

You are right. I was just thinking actual value may mean better on iteration (so it won't be unclear if index 3 is for this iteration or globally). Lets see what Behdad thinks.

> Otherwise, I'm not sure what the result for non-exclusive should be here.

Non-exclusive means not any selected also AFAICS.

src/hb-aat-layout.cc
ebraminio: Applied. Thanks

ebraminio: Revealing index of `feat` sound like exposing something client shouldn't be aware, I am not sure. Maybe a client that wants all the options can scan the table using different numbers (is that the usual case?). It is already sorted by feature type so that won't make much difference anyway.

src/hb-aat-layout.cc
ebraminio: Applied. Thanks

src/hb-aat-layout.cc
ebraminio: Somehow for API consistency and correctness, we need to consider if input is shorter than settings available so giving the client to ask from a different position with their limited buffer. See the test to see how it can be used. Given the low of number of options available here, 0 is what will clients will use most of the time but just like the similar other situation we need to do that consistently I believe.

src/hb-aat-layout.cc
ebraminio: Thanks. Applied

src/hb-aat-layout.h
ebraminio: Applied. Thanks

src/hb-aat-layout-feat-table.hh
```diff
@@ -45,20 +45,23 @@ struct SettingName
     return_trace (likely (c->check_struct (this)));
   }
 
-  protected:
```
behdad: Put a explicit public instead.

src/hb-aat-layout.h
```diff
@@ -0,0 +1,79 @@
...
+ * DAMAGE.
+ *
+ * THE COPYRIGHT HOLDER SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ * FITNESS FOR A PARTICULAR PURPOSE.  THE SOFTWARE PROVIDED HEREUNDER IS
+ * ON AN "AS IS" BASIS, AND THE COPYRIGHT HOLDER HAS NO OBLIGATION TO
+ * PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
+ */
+
+#ifndef HB_AAT_LAYOUT_H
```
behdad: Rename to hb-aat.h.  Since AAT API is tiny, no need to break it into multiple files.  Also, #ifndef etc from hb-ot.h as well.

src/hb-aat-layout.h
```diff
@@ -0,0 +1,79 @@
...
+#define HB_AAT_LAYOUT_H
+
+#include "hb.h"
+#include "hb-ot.h"
```
behdad: What do you need hb-ot.h for?

src/hb-aat-layout.h
```diff
@@ -0,0 +1,79 @@
...
+  hb_ot_name_id_t name_id;
+} hb_aat_feature_type_selector_t;
+
+/*
+ * Since: REPLACEME
+ */
+#define HB_AAT_FEATURE_NO_DEFAULT_SETTING ((hb_aat_feature_setting_t) -1)
+
+HB_EXTERN unsigned int
+hb_aat_get_feature_settings (hb_face_t                      *face,
```
behdad: `hb_aat_layout` prefix.  Not sure about the types.  Those can stay without.

src/hb-aat-layout.h
behdad: Okay I see, name_id.

ebraminio: hb_ot_name_id

src/hb-aat-layout.h
ebraminio: So including automatically with hb-ot.h?

ebraminio: So including automatically with hb-ot.h?

src/hb-aat-layout.h
```diff
@@ -0,0 +1,79 @@
...
+HB_BEGIN_DECLS
+
+/**
+ * hb_aat_type_t:
+ *
+ * Feature identifier
+ *
+ * Since: REPLACEME
+ */
+typedef uint16_t hb_aat_feature_type_t;
```
behdad: In most of the API we use unsigned int instead of smaller types.  I suggest we do that here as well.

src/hb-aat-layout.h
behdad: Maybe the types too.  Features are tightly tied to layout.

src/hb-aat-layout.h
behdad: No.  Becomes hb-aat.h which is a toplevel.  No extra defines needed at this point.  My bad.

test/api/test-aat-layout.c
```diff
@@ -24,9 +24,9 @@
 
 #include "hb-test.h"
 
-#include <hb-aat-layout.h>
+#include <hb-ot.h>
```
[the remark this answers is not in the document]
ebraminio: Just to be sure, do you hb-aat.h be included automatically with hb-ot.h or not?

behdad:
A couple side-issues:

* I currently merge multiple feature settings for the same setting when applying features.  Looks like some features are non-exclusive and hence should not be merged?  Or I can just remove the merge logic...

* This API is not quite useful without mapping from AAT features to OpenType tags.  Do we expose that?

src/hb-aat-layout.h
ebraminio: I know. The problem is I replaced https://github.com/harfbuzz/harfbuzz/blob/79b2fa62ca7eb5a875778d3a2ecae63350083ba3/src/hb-aat-layout.hh#L38-L40 with the types for more consistency and I guess you used uint16_t to optimize feature_mappings size.

ebraminio:
> This API is not quite useful without mapping from AAT features to OpenType tags. Do we expose that?

Maybe a separate issue, you said Chrome doesn't need that right now.

src/hb-aat-layout.cc
drott: I would be in favour of having an enum, or at least clearly specifying a source for where these values should come from / what the reference would be for them.

src/hb-aat-layout.h
behdad: Yeah the size change there doesn't matter.  If anything, int-sized would be faster.

behdad:
> > This API is not quite useful without mapping from AAT features to OpenType tags. Do we expose that?
> 
> Maybe a separate issue, you said Chrome doesn't need that right now.

Right.  But I want to have the full picture laid out before adding anything.  I don't want to find later that we should have done this differently.

behdad:
It occurs to me that it's also possible to get user set AAT features using feature codes directly, by, eg, passing `HB_TAG("   1")` to mean feature type 1, and the value will be the selector.  I think we should hook these up in addition to the current OpenType to AAT feature mapping.

behdad:
> `HB_TAG(" 1")`

Of just `(hb_tag_t) 1`.

behdad:
> Of just `(hb_tag_t) 1`.

Yeah I like this one better.

behdad:
Read 'feat' table.  So, the proposed API completely misses the part about enumerating features themselves.  I don't like that.  I think there should be a first layer API to list all features and get their name_id's and look up feature's index, then the second layer takes a feature_index and lists the settings.

src/hb-aat-layout.cc
ebraminio: Turned into enum.

src/hb-aat-layout-feat-table.hh
ebraminio: Done

src/hb-aat-layout.h
ebraminio: Done

src/hb-aat-layout.h
ebraminio: Done

src/hb-aat-layout.h
ebraminio: Done

ebraminio:
>  I think there should be a first layer API to list all features and get their name_id's and look up feature's index, then the second layer takes a feature_index and lists the settings.

Done. But I liked the first layer just was fetching features and fetching name_id on second layer.

> It occurs to me that it's also possible to get user set AAT features using feature codes directly, by, eg, passing HB_TAG(" 1") to mean feature type 1, and the value will be the selector.

So reverting enum thing?

ebraminio:
Related part of CT:

```
CT_EXPORT const CFStringRef kCTFontOpenTypeFeatureTag CT_AVAILABLE(macos(10.10), ios(8.0), watchos(2.0), tvos(9.0));
CT_EXPORT const CFStringRef kCTFontOpenTypeFeatureValue CT_AVAILABLE(macos(10.10), ios(8.0), watchos(2.0), tvos(9.0));
CT_EXPORT const CFStringRef kCTFontFeatureTypeIdentifierKey CT_AVAILABLE(macos(10.5), ios(3.2), watchos(2.0), tvos(9.0));
CT_EXPORT const CFStringRef kCTFontFeatureTypeNameKey CT_AVAILABLE(macos(10.5), ios(3.2), watchos(2.0), tvos(9.0));
CT_EXPORT const CFStringRef kCTFontFeatureTypeExclusiveKey CT_AVAILABLE(macos(10.5), ios(3.2), watchos(2.0), tvos(9.0));
CT_EXPORT const CFStringRef kCTFontFeatureTypeSelectorsKey CT_AVAILABLE(macos(10.5), ios(3.2), watchos(2.0), tvos(9.0));
CT_EXPORT const CFStringRef kCTFontFeatureSelectorIdentifierKey CT_AVAILABLE(macos(10.5), ios(3.2), watchos(2.0), tvos(9.0));
CT_EXPORT const CFStringRef kCTFontFeatureSelectorNameKey CT_AVAILABLE(macos(10.5), ios(3.2), watchos(2.0), tvos(9.0));
CT_EXPORT const CFStringRef kCTFontFeatureSelectorDefaultKey CT_AVAILABLE(macos(10.5), ios(3.2), watchos(2.0), tvos(9.0));
CT_EXPORT const CFStringRef kCTFontFeatureSelectorSettingKey CT_AVAILABLE(macos(10.5), ios(3.2), watchos(2.0), tvos(9.0));

CFArrayRef __nullable CTFontCopyFeatures( CTFontRef font ) CT_AVAILABLE(macos(10.5), ios(3.2), watchos(2.0), tvos(9.0));

CFArrayRef __nullable CTFontCopyFeatureSettings( CTFontRef font ) CT_AVAILABLE(macos(10.5), ios(3.2), watchos(2.0), tvos(9.0));
```

ebraminio:
Just completed feature types list, I know it including all the opentype features is avoided and we wanted the same here, I think however given the low number of types I've just completed that can be bearable. I have the settings also around but that shouldn't be needed here I guess.

behdad:
Types without their settings... a bit useless?

behdad:
I'm actually fine adding OpenType features.  Maybe with comments for which are default-enabled, which are auto, which are discretionary, ..

ebraminio:
> Types without their settings... a bit useless?

Just added all the selectors...

ebraminio:
There is a copy of selectors here on ICU https://github.com/sencha/icu/blob/master/source/layout/MorphTables.h#L131

src/hb-aat.h
```diff
@@ -0,0 +1,765 @@
...
+HB_BEGIN_DECLS
+
+/*
+ * AAT available features
+ *
+ * Since: REPLACEME
+ */
+typedef enum
+{
+  HB_AAT_LAYOUT_FEATURE_TYPE_ALL_TYPOGRAPHIC                   = 0,
```
behdad: All these alignments should use tab, not spaces.

src/hb-aat.h
```diff
@@ -0,0 +1,765 @@
...
+
+
+/*
+ * Selectors for HB_AAT_LAYOUT_FEATURE_TYPE_ALL_TYPOGRAPHIC
+ *
+ * Since: REPLACEME
+ */
+enum
+{
+  HB_AAT_LAYOUT_SELECTOR_ALL_TYPE_FEATURES_ON  = 0,
```
behdad: I don't like that the name of these selectors doesn't reflect what feature type they apply on.  But guess that's life.

src/hb-aat.h
```diff
@@ -0,0 +1,765 @@
...
+  HB_AAT_LAYOUT_FEATURE_TYPE_LOWER_CASE                        = 37,
+  HB_AAT_LAYOUT_FEATURE_TYPE_UPPER_CASE                        = 38,
+  HB_AAT_LAYOUT_FEATURE_TYPE_LANGUAGE_TAG_TYPE                 = 39,
+  HB_AAT_LAYOUT_FEATURE_TYPE_CJK_ROMAN_SPACING_TYPE            = 103,
+  HB_AAT_LAYOUT_FEATURE_TYPE_UNDEFINED                         = 0xFFFF
+} hb_aat_layout_feature_type_t;
+
+
+/*
+ * Selectors for HB_AAT_LAYOUT_FEATURE_TYPE_ALL_TYPOGRAPHIC
```
behdad: I think you should use `#HB_AAT_LAYOUT_FEATURE_TYPE_ALL_TYPOGRAPHIC` instead.

src/hb-aat.h
```diff
@@ -0,0 +1,765 @@
...
+#include "hb-ot.h"
+
+HB_BEGIN_DECLS
+
+/*
+ * AAT available features
+ *
+ * Since: REPLACEME
+ */
+typedef enum
```
behdad: Given that these expanded into hundreds of lines of enums, maybe we should move to `hb-aat-feature.h` or somehting.  Or not, no need to change right now.

src/Makefile.sources
```diff
@@ -188,6 +188,7 @@ HB_OT_headers = \
 	hb-ot-name.h \
 	hb-ot-shape.h \
 	hb-ot-var.h \
+	hb-aat.h \
```
behdad: WHATS THIS!??!?!?!!!

src/hb-aat.h
```diff
@@ -0,0 +1,765 @@
...
+			    hb_aat_layout_feature_record_t *record_buffer /* OUT.     May be NULL. */);
+
+/**
+ * hb_aat_feature_t:
+ *
+ * Feature value
+ *
+ * Since: REPLACEME
+ */
+typedef unsigned int hb_aat_layout_feature_setting_t;
```
behdad: Maybe we put all the selectors in one enum and call it `hb_aat_layout_feature_setting_t`?  Guess I like that better.  The multiple anonymous enums currently don't add any benefit other than looking nice.

src/hb-aat.h
```diff
@@ -0,0 +1,765 @@
...
+ *
+ * Feature type selector
+ *
+ * Since: REPLACEME
+ **/
+typedef struct hb_aat_layout_feature_type_selector_t
+{
+  hb_aat_layout_feature_setting_t setting;
+  hb_ot_name_id_t name_id;
+} hb_aat_layout_feature_type_selector_t;
```
behdad: "feature type selector" as the name here, where does it come from?  What does it mean?  I would have gone with "feature entry" or something.

behdad: Or better yet.  Remove it.  Add API that just returns feature_setting_t, and separate API to get name_id of a feature, if that makes sense / not too slow / can bsearch?

behdad:
Reviewed.

behdad:
One idea Dominik and I just had is that if `feat` table is not available, we can probably synthesize some of this data from morx chains...

behdad:
Also next step would be, in the `hb-ot-layout` if GSUB / GPOS is not available / used, use these and convert to OpenType-style features.

src/Makefile.sources
ebraminio: Hmm?

behdad: Err. Sorry not this. You included it in hb-ot.h? That I meant to object to.

ebraminio:
Have some stub but have a look if general ideas now looks OK.

src/hb-aat-layout-feat-table.hh
```diff
@@ -45,15 +45,25 @@ struct SettingName
...
+struct feat;
+
 struct FeatureName
 {
+  static int cmp (const void *key_, const void *entry_)
+  {
+    hb_aat_feature_type_t key = * (hb_aat_feature_type_t *) key_;
+    const FeatureName * entry = (const FeatureName *) entry_;
+    return key < entry->feature ? -1 :
+	   key > entry->feature ? 1 :
```
behdad: Nit, but I think we use `+1` in these cases.

src/hb-aat.h
```diff
@@ -0,0 +1,540 @@
...
+
+/*
+ * AAT available settings
+ *
+ * Since: REPLACEME
+ */
+typedef enum
+{
+  HB_AAT_LAYOUT_SELECTOR_UNDEFINED			= 0xFFFF,
+  /* Selectors for #HB_AAT_LAYOUT_FEATURE_TYPE_ALL_TYPOGRAPHIC */
```
behdad: Empty line before each group.

src/hb-aat.h
```diff
@@ -0,0 +1,540 @@
...
+  HB_AAT_LAYOUT_FEATURE_TYPE_ITALIC_CJK_ROMAN                  = 32,
+  HB_AAT_LAYOUT_FEATURE_TYPE_CASE_SENSITIVE_LAYOUT             = 33,
+  HB_AAT_LAYOUT_FEATURE_TYPE_ALTERNATE_KANA                    = 34,
+  HB_AAT_LAYOUT_FEATURE_TYPE_STYLISTIC_ALTERNATIVES            = 35,
+  HB_AAT_LAYOUT_FEATURE_TYPE_CONTEXTUAL_ALTERNATIVES           = 36,
+  HB_AAT_LAYOUT_FEATURE_TYPE_LOWER_CASE                        = 37,
+  HB_AAT_LAYOUT_FEATURE_TYPE_UPPER_CASE                        = 38,
+  HB_AAT_LAYOUT_FEATURE_TYPE_LANGUAGE_TAG_TYPE                 = 39,
+  HB_AAT_LAYOUT_FEATURE_TYPE_CJK_ROMAN_SPACING_TYPE            = 103,
+  HB_AAT_LAYOUT_FEATURE_TYPE_UNDEFINED                         = 0xFFFF
```
behdad: We use `INVALID` in other places.

behdad: Also, a newline before it.

src/hb-aat.h
```diff
@@ -0,0 +1,540 @@
...
+ * Selectors for #HB_AAT_LAYOUT_FEATURE_TYPE_DESIGN_COMPLEXITY_TYPE
+ *
+ * Since: REPLACEME
+ */
+  HB_AAT_LAYOUT_SELECTOR_DESIGN_LEVEL1 = 0,
+  HB_AAT_LAYOUT_SELECTOR_DESIGN_LEVEL2 = 1,
+  HB_AAT_LAYOUT_SELECTOR_DESIGN_LEVEL3 = 2,
+  HB_AAT_LAYOUT_SELECTOR_DESIGN_LEVEL4 = 3,
+  HB_AAT_LAYOUT_SELECTOR_DESIGN_LEVEL5 = 4,
+/*
```
behdad: Remove the empty line from beginning of comment, add real empty line before.

src/hb-aat.h
```diff
@@ -0,0 +1,540 @@
...
+} hb_aat_layout_feature_type_t;
+
+/*
+ * AAT available settings
+ *
+ * Since: REPLACEME
+ */
+typedef enum
+{
+  HB_AAT_LAYOUT_SELECTOR_UNDEFINED			= 0xFFFF,
```
behdad: Same.  INVALID.

src/hb-aat.h
behdad: Maybe move to beginning as well?

src/hb-aat.h
```diff
@@ -0,0 +1,540 @@
...
+ * Since: REPLACEME
+ */
+  HB_AAT_LAYOUT_SELECTOR_FULL_WIDTH_IDEOGRAPHS   = 0,
+  HB_AAT_LAYOUT_SELECTOR_PROPORTIONAL_IDEOGRAPHS = 1,
+  HB_AAT_LAYOUT_SELECTOR_HALF_WIDTH_IDEOGRAPHS   = 2,
+/*
+ * Selectors for #HB_AAT_LAYOUT_FEATURE_TYPE_UNICODE_DECOMPOSITION_TYPE
+ *
+ * Since: REPLACEME
+ */
```
behdad: All the since tags should just go in the gtk-doc block for the enum.

ebraminio: It was just a draft, now fixed.

src/hb-aat.h
ebraminio: Was just a draft, fixed

src/hb-aat.h
ebraminio: Fixed

src/hb-aat.h
ebraminio: Fixed

src/hb-aat-layout-feat-table.hh
ebraminio: Fixed

src/hb-aat.h
ebraminio: Now both are using bsearch but to be honest I can't see what is guaranteed and what now.

`The setting name arrays must follow the feature name arrays. Each setting name array is sorted by setting. However, the setting name arrays can be arranged in any order.

The feature name array follows the feature name header. The feature name array is sorted by feature type. A featureName record specifies the feature ID, the name table index for the feature's name, and the offset in bytes to the feature's name in the name table. The format of the feature name array is as follows:`

src/hb-aat.h
ebraminio: Done

src/Makefile.sources
ebraminio: Done

src/hb-aat.h
ebraminio: Done

src/hb-aat.h
ebraminio: Done

ebraminio:
PTAL

src/hb.hh
```diff
@@ -48,6 +48,8 @@
 #include "hb-ot.h"
 #define HB_OT_H_IN
 
+#include "hb-aat.h"
```
ebraminio: To fix hb-gobject complains

src/hb-aat.h
```diff
@@ -0,0 +1,445 @@
...
+  HB_AAT_LAYOUT_SELECTOR_DEFAULT_UPPER_CASE		= 0,
+  HB_AAT_LAYOUT_SELECTOR_UPPER_CASE_SMALL_CAPS		= 1,
+  HB_AAT_LAYOUT_SELECTOR_UPPER_CASE_PETITE_CAPS		= 2,
+
+  /* Selectors for #HB_AAT_LAYOUT_FEATURE_TYPE_CJK_ROMAN_SPACING_TYPE */
+  HB_AAT_LAYOUT_SELECTOR_HALF_WIDTH_CJK_ROMAN		= 0,
+  HB_AAT_LAYOUT_SELECTOR_PROPORTIONAL_CJK_ROMAN		= 1,
+  HB_AAT_LAYOUT_SELECTOR_DEFAULT_CJK_ROMAN		= 2,
+  HB_AAT_LAYOUT_SELECTOR_FULL_WIDTH_CJK_ROMAN		= 3
+} hb_aat_layout_feature_setting_t;
```
ebraminio: I guess "setting" or "selector" should only be used

src/hb-aat-layout-feat-table.hh
```diff
@@ -113,17 +121,17 @@ struct FeatureName
...
   {
-    const UnsizedArrayOf<SettingName>& settings_table = feat+settingTableZ;
-    for (unsigned int i = 0; i < nSettings; i++)
-      if (settings_table[i].get_setting () == setting)
-        return settings_table[i].get_name_id ();
-    return HB_OT_NAME_ID_INVALID;
+    const SettingName* setting = (SettingName*) hb_bsearch (&key, feat+settingTableZ,
+							    nSettings,
+							    SettingName::static_size,
+							    SettingName::cmp);
```
ebraminio: Probably good idea to have an UnsizedSortedArrayOf (or SortedUnsizedArrayOf?) and move the logic there. COLR/CPAL needed that IIRC.

ebraminio:
There is also some sort of OT-AAT mapping on Zapf AAT table, I am not sure
if they are useful for dynamic mapping or extract and having them somewhere
(and complete ours) is enough.

ebraminio:
You can have access to CTFont even after HB AAT, you should just not pass coretext-aat after using hb_shape_full but better to not go there, Behdad concerns are understandable, maybe we can have this behind some compiler flag enabled only for Chromium and only on macOS for now (so no gentoo/chromeos packaging issue I guess) if just merging it is not a good idea.

drott:
> You can have access to CTFont even after HB AAT, you should just not pass coretext-aat after using hb_shape_full but better to not go there [...]

I will not be passing a CTFont based hb_font to HarfBuzz anymore after switching to HarfBuzz AAT, so I don't think there is an underlying CTFont anymore.

behdad:
> maybe we can have this behind some compiler flag enabled only for Chromium and only on macOS for now

Why not finish it?

ebraminio:
I don't know what to do :)

behdad:
> I don't know what to do :)

I listed:

>     * If `feat` feature is missing, dig into the `morx` table and extracted list of features?

This is easy.  But if you don't want to get into it, that's fine for the first round, since doesn't have API implications.

>     * API to convert AAT-style features into OT-style.  And back?

Propose API for this.  It's close to trivial.  We have an internal API that converts one to another.  Model two APIs based on that (just type here, DON'T send a patch) and we discuss it from there.

>     * If `GSUB` is missing, convert back `feat` table features to OT style and return from hb-ot-layout API.

If you do the conversion API, this part is also easy, but also not required initially since no API implications.

>     * Allow user specifying AAT features directly (without going through OT equivalents) in `hb_feature_t`, by using tag names like `'  12'` for feature number 12.

Also easy.  Also not required initially since no API implications.

behdad:
> > ```
> > * API to convert AAT-style features into OT-style.  And back?
> > ```
> 
> Propose API for this. It's close to trivial. We have an internal API that converts one to another. Model two APIs based on that (just type here, DON'T send a patch) and we discuss it from there.

Even this doesn't have to be implemented before merging this work.  I just want to know we have a plan and discuss the API, to avoid any surprises and blind spots later.

behdad:
One more item to add to TODO list: enum for OpenType features.

behdad:
I studied this more and I'm comfortable merging it.  We can expose something along the lines of `hb_aat_layout_find_feature_mapping()` and inverse later.

src/hb-aat-layout-feat-table.hh
```diff
@@ -68,18 +93,59 @@ struct FeatureName
...
+				    hb_aat_layout_feature_setting_t *default_setting,
+				    unsigned int                     start_offset,
+				    unsigned int                    *count,
+				    hb_aat_layout_feature_setting_t *settings) const
+  {
+    const UnsizedArrayOf<SettingName>& settings_table = feat+settingTableZ;
+    unsigned int settings_count = nSettings;
+    if (count && *count)
+    {
+      unsigned int len = MIN (settings_count - start_offset, *count);
```
behdad: This should be ported to use `hb_array_t` and `sub_array()`.

src/hb-aat-layout-feat-table.hh
behdad: Yes please.

src/hb-aat-layout-feat-table.hh
```diff
@@ -95,11 +161,53 @@ struct feat
...
   static const hb_tag_t tableTag = HB_AAT_TAG_feat;
 
+  inline unsigned int get_features (unsigned int                  start_offset,
+				    unsigned int                 *count,
+				    hb_aat_layout_feature_type_t *features) const
+  {
+    unsigned int feature_count = featureNameCount;
+    if (count && *count)
+    {
+      unsigned int len = MIN (feature_count - start_offset, *count);
```
behdad: Same re using `hb_array_t` and `.sub_array()`.

src/hb-aat.h
behdad: Right.  Pick one and only use that.

behdad: Either "feature_setting" or "feature_selector".  Selector sounds less confusing.

src/hb-aat.h
```diff
@@ -0,0 +1,445 @@
...
+
+HB_EXTERN unsigned int
+hb_aat_layout_get_features (hb_face_t                    *face,
```
behdad: Let's fix namespacing.  Again, you have all under "hb_aat_layout" namespace, ie. they all start with "hb_aat_layout_get".  We want more fine-grained namspacing.

This one is fine, `get_features`.

src/hb-aat.h
```diff
@@ -0,0 +1,445 @@
...
+			    unsigned int                  start_offset,
+			    unsigned int                 *count,   /* IN/OUT.  May be NULL. */
+			    hb_aat_layout_feature_type_t *features /* OUT.     May be NULL. */);
+
+HB_EXTERN hb_ot_name_id_t
+hb_aat_layout_get_feature_name_id (hb_face_t                    *face,
```
behdad: This should be `hb_aat_layout_feature_get_name_id`.

src/hb-aat.h
```diff
@@ -0,0 +1,445 @@
...
+				   hb_aat_layout_feature_type_t  feature);
+
+
+HB_EXTERN unsigned int
+hb_aat_layout_get_feature_settings (hb_face_t                       *face,
```
behdad: `hb_aat_layout_feature_get_settings` or `get_selectors` as discussed.

src/hb-aat.h
```diff
@@ -0,0 +1,445 @@
...
+				    hb_aat_layout_feature_type_t     feature,
+				    hb_aat_layout_feature_setting_t *default_setting, /* OUT.     May be NULL. */
+				    unsigned int                     start_offset,
+				    unsigned int                    *count,           /* IN/OUT.  May be NULL. */
+				    hb_aat_layout_feature_setting_t *settings         /* OUT.     May be NULL. */);
+
+HB_EXTERN hb_ot_name_id_t
+hb_aat_layout_get_feature_setting_name_id (hb_face_t                       *face,
```
behdad: `hb_aat_layout_feature_setting_get_name_id` or, `feature_selector`

src/hb-aat-layout-feat-table.hh
```diff
@@ -68,18 +93,59 @@ struct FeatureName
...
+    }
+    return settings_count;
+  }
+
+  inline hb_aat_layout_feature_type_t get_feature_type () const
+  { return (hb_aat_layout_feature_type_t) (unsigned int) feature; }
+
+  inline hb_ot_name_id_t get_feature_name_id () const { return nameIndex; }
+
+  inline hb_ot_name_id_t get_feature_setting_name_id (const feat                      *feat,
```
behdad: These functions taking feat table look weird.  I think things can be done differently, but I need to pull down the code and take a deeper look than just reading the patch.

behdad:
This part of the spec:
```
For feature types that don't have exclusive settings, there will always be a pair of values. One value turns a selector on and a second value turns the selector off. The on setting must be even and the off setting must be one greater than the corresponding on setting. The off setting is therefore always odd. As a result, only the on setting should have an entry in the setting name array.
```
needs at least to go into documentation...

src/hb-aat.h
behdad: The way we did this in ot-layout, and in general do these things, is to have `find_feature` or something, take feature type and return a feature index.  The subsequent API entrypoints then take that index, to avoid having to look up the feature entry again and again.

src/hb-aat-layout-feat-table.hh
ebraminio: Thought about this before, no sub_array on UnsizedArrayOf, and doesn't worth to IMO.

src/hb-aat.h
ebraminio: Done.

src/hb-aat.h
ebraminio: Done

src/hb-aat.h
ebraminio: Done.

Understandable as did once for `hb_ot_layout_feature_get_characters`

src/hb-aat-layout-feat-table.hh
behdad: UnsizedArrayOf should not be used directly.  Maybe I remove operator[] from it.  Idea is you bind it with a length to get a hb_array_t and use that, which has a sub_array().

behdad: UnsizedArrayOf::as_array().

PR #1347, opened 2018-11-01
[BASE] Improvements

ebraminio:
<s>WIP, but not familiar with variation part so I will need some help I guess.</s>

Only simulation remains, which I am not sure how we can achieve it.


src/hb-ot-layout-base-table.hh
```diff
@@ -185,37 +190,30 @@ struct FeatMinMaxRecord
...
   {
-    const FeatMinMaxRecord *minMaxCoord = find_record (baseline);
-    if (!minMaxCoord) return (this+maxCoord).get_coord ();
-    return minMaxCoord->get_max_value ();
+    const FeatMinMaxRecord *minMaxCoord = (const FeatMinMaxRecord *)
+					  hb_bsearch (&feature_tag, featMinMaxRecords.arrayZ,
+						      featMinMaxRecords.len,
+						      FeatMinMaxRecord::static_size,
+						      FeatMinMaxRecord::cmp);
+    if (minMaxCoord) minMaxCoord->get_min_max (min, max);
```
behdad: New line.

behdad:
Thanks for the refactoring @ebraminio.  Looks amazing!

ebraminio:
Thank you! :) I think it would be good idea to merge its refactoring part first without any API then deciding on finalizing the API (and simulation) if you don't disagree, that case the patch won't get outdated when commons types are refactored or something. (I should be first look however what is wrong with `-Werror -Wundefined-func-template` first of course)

behdad:
Yes, I agree with improvements going in without public API.  Makes reviewing API much easier as well.

PR #1352, opened 2018-11-03
Typename outside template

behdad:
Let's see if this fixes gcc 4.2


behdad:
Nah.  Doesn't look like this is gonna work.

ebraminio:
Configure-time feature testing, defaulting to availability?

behdad:
> Configure-time feature testing, defaulting to availability?

Meh. If someone writes the test, I'll review it.

PR #1355, opened 2018-11-04
[debug] Add a memory hex dump

ebraminio:
See if makes sense and is in right, it was useful for me not sure whether for others.

### DIFF
--- a/src/hb-debug.hh
+++ b/src/hb-debug.hh
@@ -487,5 +487,67 @@ struct hb_no_trace_t {
 #define TRACE_DISPATCH(this, format) hb_no_trace_t<typename context_t::return_t> trace
 #endif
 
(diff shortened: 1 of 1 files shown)


behdad:
I would have liked it more if it had something integrated with HarfBuzz machinery or sanitizer or something.  As it is, not much.

Maybe put it in a tool in src/ that takes table name and dumps it?

ebraminio:
I used its `_hex_dump_font_table` to see first what is going on and then to see if sanitizer has altered the table so thought be useful like this as it was for me sometimes. OK, lets decide about this later.

behdad:
I like having it as a tool as I said.

ebraminio:
That should be integrated with #950 as the difficulty of getting each table's sanitizer.

PR #1358, opened 2018-11-05
Hook AAT's lcar to _get_ligature_carets

### LINKED ISSUE #907: Implement lcar table parsing

ebraminio:
Fixes #907 and #946

<s>The issue is however Lookup is not ready to take Lookup in its first template parameter and that is needed for lcar and Zapf</s>


ebraminio:
PTAL

behdad:
> Lookup is not ready to take Lookup in its first template

You sure you need Lookup to take Lookup?

behdad:
What does your patch do then, if that doesn't work?

ebraminio:
> You sure you need Lookup to take Lookup?

Lookup to array, makes sennse to me

> What does your patch do then, if that doesn't work?

Updated the comment. Now just format 4 as there is a base already on format 4 not sure how to route another base correctly been but will look if the rest looks good

behdad:
> Updated the comment. Now just format 4 as there is a base already on format 4 not sure how to route another base correctly been but will look if the rest looks good

Fine. I'll study the table.

ebraminio:
Now supports format4 (theoretically at least) also.

ebraminio:
About the test font, it is from a font featuring `lcar` but most of its tables is gone and falls somehow in fair use I believe.

ebraminio:
PTAL

src/Makefile.sources
```diff
@@ -104,6 +104,7 @@ HB_OT_sources = \
 	hb-aat-layout-morx-table.hh \
 	hb-aat-layout-trak-table.hh \
 	hb-aat-layout.hh \
+	hb-aat-lcar-table.hh \
```
behdad: aat-layout-lcar please.

ebraminio: Done. Have a look

src/hb-ot-layout.cc
```diff
@@ -275,12 +277,15 @@ hb_ot_layout_get_ligature_carets (hb_font_t      *font,
 				  unsigned int   *caret_count /* IN/OUT */,
 				  hb_position_t  *caret_array /* OUT */)
 {
-  return font->face->table.GDEF->table->get_lig_carets (font,
-							direction,
-							glyph,
-							start_offset,
-							caret_count,
-							caret_array);
+  unsigned int result_caret_count = 0;
```
behdad: You don't need this?  Just pass caret_count to both APIs.

src/hb-aat-layout-lcar-table.hh
```diff
@@ -0,0 +1,93 @@
...
+  inline bool sanitize (hb_sanitize_context_t *c) const
+  {
+    TRACE_SANITIZE (this);
+    return_trace (c->check_struct (this) && lookup.sanitize (c, this));
+  }
+
+  protected:
+  FixedVersion<>version;	/* Version number of the ligature caret table */
+  HBUINT16	format;		/* Format of the ligature caret table. */
+  Lookup<OffsetTo<ArrayOf<HBINT16> > >
```
behdad: `typdef ArrayOf<HBINT16> LigCaretClassEntry;`

src/hb-aat-layout-lcar-table.hh
```diff
@@ -0,0 +1,93 @@
...
+  static const hb_tag_t tableTag = HB_AAT_TAG_lcar;
+
+  inline unsigned int get_lig_carets (hb_font_t      *font,
+				      hb_direction_t  direction,
+				      hb_codepoint_t  glyph,
+				      unsigned int    start_offset,
+				      unsigned int   *caret_count /* IN/OUT */,
+				      hb_position_t  *caret_array /* OUT */) const
+  {
+    const ArrayOf<HBINT16>& array = this+(*lookup.get_value (glyph, font->face->num_glyphs));
```
behdad: `get_value()` can return nullptr.  Use `get_value_or_null()`.

behdad: Typically when there's a pointer, it means it can be null.  When there's a reference, not.

src/hb-aat-layout-lcar-table.hh
```diff
@@ -0,0 +1,93 @@
...
+    unsigned int len = 0;
+    bool is_horizontal = HB_DIRECTION_IS_HORIZONTAL (direction);
+    if (caret_count && caret_array)
```
behdad: Typically the desired check is `if (caret_count && *caret_count)`.

src/hb-aat-layout-lcar-table.hh
```diff
@@ -0,0 +1,93 @@
...
+    {
+      len = MIN (array.len - start_offset, *caret_count);
```
behdad: Use ArrayOf.sub_array().

src/hb-aat-layout-lcar-table.hh
```diff
@@ -0,0 +1,93 @@
...
+      for (unsigned int i = 0; i < len; ++i)
+      {
+	unsigned int coordinate = 0;
+	if (format == 0)
```
behdad: `switch`?

src/hb-aat-layout-lcar-table.hh
```diff
@@ -0,0 +1,93 @@
...
+	  coordinate = array[i + start_offset];
+	else if (format == 1)
+	{
+	  unsigned int point_index = (unsigned int) array[i + start_offset];
+	  hb_position_t x = 0, y = 0;
```
behdad: You don't need to zero.  font does it for you.

src/hb-aat-layout-lcar-table.hh
```diff
@@ -0,0 +1,93 @@
...
+	  font->get_glyph_contour_point (glyph, point_index, &x, &y);
+	  coordinate = is_horizontal ? x : y;
+	}
+	caret_array[i] = is_horizontal ? font->em_scale_x (coordinate) : font->em_scale_y (coordinate);
+      }
+    }
+    if (caret_count) *caret_count = len;
```
behdad: With the change I suggested above, this can go into the previous `if` body as well and unconditional.

src/hb-aat-layout-lcar-table.hh
```diff
@@ -0,0 +1,93 @@
...
+      for (unsigned int i = 0; i < len; ++i)
+      {
+	unsigned int coordinate = 0;
+	if (format == 0)
+	  coordinate = array[i + start_offset];
+	else if (format == 1)
+	{
+	  unsigned int point_index = (unsigned int) array[i + start_offset];
+	  hb_position_t x = 0, y = 0;
+	  font->get_glyph_contour_point (glyph, point_index, &x, &y);
```
behdad: Umm.  I'm not sure if we want `get_glyph_contour_point_for_origin` here or not.

Humm. Also, if returns false, not sure what to do.

behdad: That's what GDEF does.  Let's do that.

src/hb-aat-layout-lcar-table.hh
```diff
@@ -0,0 +1,93 @@
...
+	  coordinate = is_horizontal ? x : y;
+	}
+	caret_array[i] = is_horizontal ? font->em_scale_x (coordinate) : font->em_scale_y (coordinate);
```
behdad: You should only scale format 0 result.  Format 1 is already scaled by the font.

behdad:
Reviewed.  Looks good otherwise.

src/hb-ot-layout.cc
ebraminio: Then `GDEF` makes it zero and `lcar` can't fill it.

src/hb-aat-layout-lcar-table.hh
ebraminio: Thanks.

`this+lookup.get_value_or_null (glyph, font->face->num_glyphs);` fails with

```
hb-aat-layout-common.hh: In instantiation of ‘const typename T::type AAT::LookupFormat10<T>::get_value_or_null(hb_codepoint_t) const [with T = OT::OffsetTo<OT::ArrayOf<OT::IntType<short int, 2> > >; typename T::type =OT::IntType<short unsigned int, 2>; hb_codepoint_t = unsigned int]’:
hb-aat-layout-common.hh:353:61:   required from ‘const typename T::type AAT::Lookup<T>::get_value_or_null(hb_codepoint_t, unsigned int) const [with T = OT::OffsetTo<OT::ArrayOf<OT::IntType<short int, 2> > >; typename T::type = OT::IntType<short unsigned int, 2>; hb_codepoint_t = unsigned int]’
hb-aat-layout-lcar-table.hh:50:97:   required from here
hb-aat-layout-common.hh:310:12: error: could not convert ‘v’ from ‘unsigned int’ to ‘const type {aka const OT::IntType<short unsigned int, 2>}’
     return v;
            ^
```

Manually doing the same.

src/hb-ot-layout.cc
behdad: Got it.  You're right.

src/hb-aat-layout-lcar-table.hh
ebraminio: Done

src/hb-aat-layout-lcar-table.hh
ebraminio: Done

src/hb-aat-layout-lcar-table.hh
ebraminio: Done

src/hb-aat-layout-lcar-table.hh
behdad: Oh right.  Something to eventually fix, such that format10 can work as well.  But adding cast constructors to Offset type currently fails on pre-C++11 I think.  Maybe in a separate PR add an explicit operator cast to numeric types and see if it passes bots.

src/hb-aat-layout-lcar-table.hh
ebraminio: Done

src/hb-aat-layout-lcar-table.hh
ebraminio: Done

src/hb-aat-layout-lcar-table.hh
```diff
@@ -0,0 +1,93 @@
...
+      }
+    }
+    if (caret_count) *caret_count = len;
+    return array.len;
+  }
+
+  inline bool sanitize (hb_sanitize_context_t *c) const
+  {
+    TRACE_SANITIZE (this);
+    return_trace (c->check_struct (this) && lookup.sanitize (c, this));
```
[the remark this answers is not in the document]
ebraminio: Done

src/hb-aat-layout-lcar-table.hh
behdad: https://github.com/harfbuzz/harfbuzz/pull/1370

src/hb-aat-layout-lcar-table.hh
ebraminio: Done

src/hb-aat-layout-lcar-table.hh
```diff
@@ -0,0 +1,96 @@
...
+#define HB_AAT_TAG_lcar HB_TAG('l','c','a','r')
+
+
+namespace AAT {
+
+struct lcar
+{
+  static const hb_tag_t tableTag = HB_AAT_TAG_lcar;
+
+  typedef ArrayOf<HBINT16> LigCaretClassEntry;
```
behdad: Typically this should go outside...

We've been rather inconsistent between typedef or empty inheritance (like you did in your `just` table.)  Maybe clean up one way or another.  Maybe empty struct is cleaner.

src/hb-aat-layout-lcar-table.hh
behdad: This ^

src/hb-aat-layout-lcar-table.hh
ebraminio: Done.

src/hb-aat-layout-lcar-table.hh
ebraminio: Oops, very sorry about the last edit :)

src/hb-aat-layout-lcar-table.hh
ebraminio: Done

src/hb-aat-layout-lcar-table.hh
ebraminio: Done :)

ebraminio:
PTAL (but no rush, any time you want to)

ebraminio:
Maybe typedef is better if someone enables RTTI, no experience with it of
course

behdad:
We don't care about RTTI.  We don't have dynamic inheritance.

behdad:
Plus, we don't expose any C++ types.

src/hb-aat-layout-lcar-table.hh
```diff
@@ -0,0 +1,94 @@
...
+      for (unsigned int i = 0; i < count; ++i)
+	switch (format)
+	{
+	case 0: caret_array[i] = font->em_scale_dir (arr[i], direction); break;
+	case 1:
+	  hb_position_t x, y;
+	  if (font->get_glyph_contour_point_for_origin (glyph, arr[i], direction, &x, &y))
+	    caret_array[i] = HB_DIRECTION_IS_HORIZONTAL (direction) ? x : y;
+	  else
+	    caret_array[i] = 0;
```
behdad: Again, I think the function call zeroes, even if it failed.  If GDEF is doing this fix that too.  No biggy though.

behdad:
LGTM.  Thanks.

src/hb-aat-layout-lcar-table.hh
ebraminio: Trying to reduce and avoid unnecessary `if`s is a good practice. Applied the same to `GDEF`.

ebraminio:
> LGTM. Thanks.

Have a look at the GDEF change then we can merge I guess

behdad:
Looks good.

PR #1364, opened 2018-11-07
[just] Initial table implementation

### LINKED ISSUE #917: Implement just table parsing
This is used to to accommodate AAT specific justification functions.
See: https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6just.html

(not to be confused with OT's JSTF https://docs.microsoft.com/en-us/typography/opentype/spec/jstf )

ebraminio:
Closes #917

### DIFF
--- a/src/hb-aat-layout-just-table.hh
+++ b/src/hb-aat-layout-just-table.hh
@@ -1,0 +1,416 @@
+/*
+ * Copyright © 2018  Ebrahim Byagowi
+ *
+ *  This is part of HarfBuzz, a text shaping library.
+ *
+ * Permission is hereby granted, without written agreement and without
+ * license or royalty fees, to use, copy, modify, and distribute this
+ * software and its documentation for any purpose, provided that the
+ * above copyright notice and the following two paragraphs appear in
+ * all copies of this software.
+ *
+ * IN NO EVENT SHALL THE COPYRIGHT HOLDER BE LIABLE TO ANY PARTY FOR
+ * DIRECT, INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES
+ * ARISING OUT OF THE USE OF THIS SOFTWARE AND ITS DOCUMENTATION, EVEN
+ * IF THE COPYRIGHT HOLDER HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ *
+ * THE COPYRIGHT HOLDER SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ * FITNESS FOR A PARTICULAR PURPOSE.  THE SOFTWARE PROVIDED HEREUNDER IS
+ * ON AN "AS IS" BASIS, AND THE COPYRIGHT HOLDER HAS NO OBLIGATION TO
+ * PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
+ */
+
+#ifndef HB_AAT_LAYOUT_JUST_TABLE_HH
+#define HB_AAT_LAYOUT_JUST_TABLE_HH
+
+#include "hb-aat-layout-common.hh"
+#include "hb-ot-layout.hh"
+#include "hb-open-type.hh"
+
+#include "hb-aat-layout-morx-table.hh"
+
+/*
+ * just -- Justification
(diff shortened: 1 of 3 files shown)


src/hb-aat-layout-just-table.hh
```diff
@@ -0,0 +1,416 @@
...
+ */
+
+#ifndef HB_AAT_LAYOUT_JUST_TABLE_HH
+#define HB_AAT_LAYOUT_JUST_TABLE_HH
+
+#include "hb-aat-layout-common.hh"
+#include "hb-ot-layout.hh"
+#include "hb-open-type.hh"
+
+#include "hb-aat-layout-morx-table.hh"
```
behdad: What do you need this for?

src/hb-aat-layout-just-table.hh
```diff
@@ -0,0 +1,416 @@
...
+    case 1:  return_trace (u.unconditionalAddGlyphAction.sanitize (c));
+    case 2:  return_trace (u.conditionalAddGlyphAction.sanitize (c));
+    // case 3: return_trace (u.stretchGlyphAction.sanitize (c));
+    case 4:  return_trace (u.decompositionAction.sanitize (c));
+    case 5:  return_trace (u.decompositionAction.sanitize (c));
+    default: return_trace (true);
+    }
+  }
+
+  inline unsigned int get_length () const { return u.header.actionLength; }
```
behdad: Move to the top.  Nit: `(void)` instead of `()`.  sanitize always just before the fields.

src/hb-aat-layout-just-table.hh
```diff
@@ -0,0 +1,416 @@
...
+				 * glyphs that only participate in justification
+				 * after the above priorities. Normally all glyphs
+				 * have one of the previous three values. If you
+				 * don't want a glyph to participate in justification,
+				 * and you don't want to set its factors to zero,
+				 * you may instead assign it to the null priority. */
+  };
+
+  protected:
+  Fixed		beforeGrowLimit;/* The ratio by which the advance width of the
```
behdad: Why so much space between type and fields when not needed?  Reduce, and align comments nicely.

src/hb-aat-layout-just-table.hh
```diff
@@ -0,0 +1,416 @@
...
+    return_trace (likely (c->check_struct (this)));
+  }
+
+  protected:
+  HBUINT32	justClass;	/* The justification category associated
+				 * with the wdRecord field. Only 7 bits of
+				 * this field are used. (The other bits are
+				 * used as padding to guarantee longword
+				 * alignment of the following record). */
+  JustWidthDeltaEntry
```
behdad: Here you can use another tab to avoid line break.

src/hb-aat-layout-just-table.hh
```diff
@@ -0,0 +1,416 @@
...
+
+struct just
+{
+  static const hb_tag_t tableTag = HB_AAT_TAG_just;
+
+  inline bool sanitize (hb_sanitize_context_t *c) const
+  {
+    TRACE_SANITIZE (this);
+
+    return_trace (unlikely (c->check_struct (this) &&
```
behdad: Check version major.

src/hb-aat-layout-just-table.hh
ebraminio: It says `ChainSubtable<ObsoleteTypes> morphHeader;	/* Metamorphosis-style subtable header. */`, maybe is a different or maybe should be moved to common but it was very morph specific so I don't know.

src/hb-aat-layout-just-table.hh
ebraminio: Hmmm, this is what I see locally.

image

src/hb-aat-layout-just-table.hh
ebraminio: Applying that no alignment for me,
image

src/hb-aat-layout-just-table.hh
ebraminio: Done https://github.com/harfbuzz/harfbuzz/pull/1369

src/hb-aat-layout-just-table.hh
ebraminio: Done https://github.com/harfbuzz/harfbuzz/pull/1369

behdad:
Don't have time to study the table right now.  Just make sure this is included somewhere so compiler sees it.

ebraminio:
> Don't have time to study the table right now. Just make sure this is included somewhere so compiler sees it.

No need to for now I guess but seems nice. Included on `src/hb-aat-layout.cc`.

PR #1368, opened 2018-11-08
[AAT] Support Lookup<OffsetTo<>>, needed by lcar and just

ebraminio:


behdad:
Thanks.  A couple comments:
1. No newline between several sanitize() functions of the same struct.
2. Don't rename base.
3. The new argument should be called template <typename T> T user_data.  See ArrayOf<>.

ebraminio:
Applied.

behdad:
user_data needs to be templatized as I said, such that it can pass any type (int for example).  See ArrayOf.

ebraminio:
My bad. Applied

PR #1370, opened 2018-11-08
[DONTMERGE] Inttype constructors

behdad:
Testing bots.  See commit.


behdad:
As expected, from the first failing bot:
```
/Users/distiller/project/src/hb-aat-layout-common.hh:391:13: error: union member 'format' has a non-trivial constructor
  HBUINT16              format;         /* Format identifier */
                        ^
```

PR #1376, opened 2018-11-10
[colr/feat/trak] minor

ebraminio:


src/hb-aat-layout-trak-table.hh
```diff
@@ -221,6 +212,16 @@ struct trak
     return_trace (true);
   }
 
+  inline bool sanitize (hb_sanitize_context_t *c) const
+  {
+    TRACE_SANITIZE (this);
+
+    return_trace (likely (c->check_struct (this) &&
+			  version.major == 1 &&
```
ebraminio: This is the only added thing on the patch and needs to be reviewed I guess, the rest is moving sanitization to the right place.

PR #1380, opened 2018-11-13
Don't use Win32 API on Cygwin

kbrow1i:
Cygwin is a Posix platform to the extent possible.  It should use the
Posix API except in special circumstances.

After this change, harfbuzz-2.1.1 builds and passes all tests on both 32-bit and 64-bit Cygwin.  I can't check the git HEAD at the moment because I don't have ragel.

### DIFF
--- a/src/hb.hh
+++ b/src/hb.hh
@@ -246,7 +246,7 @@ struct _hb_alignof
 #endif
 
 
-#if defined(_WIN32) || defined(__CYGWIN__)
+#ifdef _WIN32
    /* We need Windows Vista for both Uniscribe backend and for
     * MemoryBarrier.  We don't support compiling on Windows XP,
     * though we run on it fine. */
--- a/src/hb-atomic.hh
+++ b/src/hb-atomic.hh
@@ -100,7 +100,7 @@ _hb_atomic_ptr_impl_cmplexch (const void **P, const void *O_, const void *N)
 #define hb_atomic_ptr_impl_cmpexch(P,O,N)	_hb_atomic_ptr_impl_cmplexch ((const void **) (P), (O), (N))
 
 
-#elif !defined(HB_NO_MT) && (defined(_WIN32) || defined(__CYGWIN__))
+#elif !defined(HB_NO_MT) && defined(_WIN32)
 
 #include <windows.h>
 
--- a/src/hb-blob.cc
+++ b/src/hb-blob.cc
@@ -481,7 +481,7 @@ hb_blob_t::try_make_writable (void)
 # include <fcntl.h>
 #endif
 
-#if defined(_WIN32) || defined(__CYGWIN__)
+#ifdef _WIN32
 # include <windows.h>
 #else
 # ifndef O_BINARY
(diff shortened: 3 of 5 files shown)


behdad:
> I don't have ragel.

You don't need ragel to build master anymore.  We include generated ragel files in git.

src/hb.hh
```diff
@@ -246,7 +246,7 @@ struct _hb_alignof
 #endif
 
 
-#if defined(_WIN32) || defined(__CYGWIN__)
+#ifdef _WIN32
```
behdad: I'm not sure about this one.  This one should be harmless, right?  What happens if one is building the uniscribe backend?  is that supported?

kbrow1i:
> > I don't have ragel.
> 
> You don't need ragel to build master anymore. We include generated ragel files in git.

I saw that, but for some reason it didn't work for me.  I kept getting messages that I was missing ragel.  Anyway, I've now built and installed ragel and can build master HEAD.  Again, it builds fine and passes all tests with my patch.

src/hb.hh
kbrow1i: I actually have no idea what uniscribe is or whether it's supported.  But I agree that it's harmless, so I don't care if you leave it in.  Should I add a commit my branch to do that, or will you take care of it?  I don't have much experience with PRs, so I'm not sure which way the workflow for you is easiest.

ebraminio:
> Cygwin is a Posix platform to the extent possible.

Does cygwin's POSIX's `mmap` works actually? I remember that it wasn't working but if so why you care about its win32 path as it already prefers `mmap`? I feel the change is a little unjustified as it degrades file reading quality theoretically.

behdad:
Good questions.  I'm also curious as to why this change is needed in the first place.

kbrow1i:
> > Cygwin is a Posix platform to the extent possible.
> 
> Does cygwin's POSIX's `mmap` works actually? I remember that it wasn't working but if so why you care about its win32 path as it already prefers `mmap`?

Cygwin's mmap works fine.

> I feel the change is a little unjustified as it degrades file reading quality theoretically.

The main reason I proposed the change is that it reflects the way harfbuzz (like most Cygwin packages) is actually built for the Cygwin distribution.  Cygwin's harfbuzz maintainer, Yaakov Selkowitz, has been using a patch like this for many years.  And I use a patch like this when I build the Cygwin binaries for TeX Live (which includes harfbuzz in its source tree).

It's tedious to have to keep updating the patch as harfbuzz evolves.  It would be much easier for us  Cygwin package maintainers if you would view Cygwin as a POSIX platform, as we do.  Thanks.

behdad:
Okay, so you want atomic and mutex to be taken from POSIX not Windows?  If that's the case, I think it makes more sense to prioritize those after POSIX alternatives, and then no patch is needed.  Or you are write, we might as well remove the CYGWIN checks.

Can you try compiling `--with-uniscribe` and report if it still works?  Thanks.

kbrow1i:
> Okay, so you want atomic and mutex to be taken from POSIX not Windows? If that's the case, I think it makes more sense to prioritize those after POSIX alternatives, and then no patch is needed. Or you are write, we might as well remove the CYGWIN checks.
> 
> Can you try compiling `--with-uniscribe` and report if it still works? Thanks.

No, it doesn't.  I'll take a look and see what's going on.  (First I'll have to find out what uniscribe is and why we might want to support it in the Cygwin build.)

ebraminio:
Uniscribe is HarfBuzz of Microsoft, simply :) HarfBuzz uses it to get
Microsoft result of what HarfBuzz library do, but in HarfBuzz format and
API, we use it internally to verify with Microsoft results but some clients
may use it to get Uniscribe compatible results, may useful when you measure
with something but render the text with something else and such limited
uses usually AFAIK.

behdad:
If you're happy, I'm happy..

ebraminio:
msys2 doesn't support mmap per these [1](https://sourceforge.net/p/msys2/mailman/message/35411480/) [2](https://sourceforge.net/p/msys2/mailman/message/35413413/) AFAICS, should we target those using __MINGW32__ or __MSYS__?

PR #1387, opened 2018-11-15
Adjust spaces on math and subset related files

ebraminio:


behdad:
Thanks.  Doing such changes in newly added code is fine, but please resist the temptation to do that across the entire codebase.  When touching a piece of code, update it to latest standards, but don't go fishing for changes.

That said, `Null()`, `Crap()`, and a few other exceptions don't get space before `()`.  That's just historically how it has been.

behdad:
I checked the number of occurences.  I'm fine if you go ahead and add space in between all remaining `Null()` and `Crap`.

ebraminio:
The motivation was to not feel that subset and math is written in a different style so tried to limit the changes to the related code.

PR #1388, opened 2018-11-16
[subset] Dump invalid glyph ids after computing the set of glyph ids …

garretrieger:
…to retain.

### DIFF
--- a/src/hb-subset-plan.cc
+++ b/src/hb-subset-plan.cc
@@ -67,6 +67,16 @@ _gsub_closure (hb_face_t *face, hb_set_t *gids_to_retain)
                                            gids_to_retain);
 }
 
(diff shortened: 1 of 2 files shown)


src/hb-subset-plan.cc
```diff
@@ -112,6 +122,7 @@ _populate_gids_to_retain (hb_face_t *face,
   }
   hb_set_destroy (initial_gids_to_retain);
 
+  _remove_invalid_gids (all_gids_to_retain, glyf.max_glyph_id ());
```
behdad: Humm.  We should get num glyph ids from `face->get_num_glyphs ()`.

behdad:
Committed manually.

PR #1397, opened 2018-11-20
Shrink the emoji table by merging adjacent ranges

dscorbett:


behdad:
Same should be done to Pango... I'm too lazy to do right now.

behdad:
Okay, applied to Pango as well.  Thanks!

PR #1399, opened 2018-11-21
Fix USE categories for U+111C9 SHARADA SANDHI MARK

dscorbett:
According to [L2/12-322](https://www.unicode.org/L2/L2012/12322-n4330-sharada-sandhi-mark.pdf), a Sharada cluster may have two sandhi marks, and a sandhi mark may follow U+111C8 SHARADA SEPARATOR.


behdad:
Thanks.  Please add a test if you can.

PR #1401, opened 2018-11-21
[subset] Add an op counter for gsub/gpos closure.

garretrieger:


behdad:
Fixed another way.  Thanks!

PR #1402, opened 2018-11-22
[DONT-MERGE] Decltype

behdad:
Just checking how bots react.


behdad:
Looks like I don't need this after all.

PR #1404, opened 2018-11-23
[aat] Add macOS specific tests

### LINKED ISSUE #1403: Testing against system fonts
Given that we don't have complex-enough mort/morx/kerx test fonts, one way to test those is to modify test suite to allow testing against system fonts.  In the .tests file format, if the font file path is absolute (starts with /), then that test should be skipped if font file does not exist.

If test fails, show sha1 of the font file, possibly font version as well.  A follow up would be to allow only testing if font file is of certain versions.  Then we can add mac-specific tests.

ebraminio:
Closes #1403


test/shaping/run-tests.py
```diff
@@ -63,6 +80,12 @@ def cmd(command):
 			print ("%s %s %s %s --unicodes %s" %
 					 (hb_shape, fontfile, ' '.join(extra_options), options, unicodes))
 
+		# hack to support fonts with space on run-tests.py, after several other tries...
+		if ' ' in fontfile:
+			new_fontfile = os.path.join (tempfile.gettempdir (), 'tmpfile')
+			shutil.copyfile(fontfile, new_fontfile)
+			fontfile = new_fontfile
```
ebraminio: Yeah, such a hack...

test/shaping/run-tests.py
```diff
@@ -63,6 +80,12 @@ def cmd(command):
 			print ("%s %s %s %s --unicodes %s" %
 					 (hb_shape, fontfile, ' '.join(extra_options), options, unicodes))
 
+		# hack to support fonts with space on run-tests.py, after several other tries...
+		if ' ' in fontfile:
```
behdad: Ouch!

behdad: This hack needs to be fixed...

Actually, Codacy warns about using Subprocess without shell=true.  Is that relevant? https://app.codacy.com/project/behdad/harfbuzz/dashboard

PR #1429, opened 2018-11-29
Fix typename issue by compiler feature testing

### LINKED ISSUE #1341: warning: 'typename' occurs outside of a template [-Wc++11-extensions]
Some of the older compilers bots are full of these
```
../src/hb-buffer.hh:229:50: warning: 'typename' occurs outside of a template [-Wc++11-extensions]
    if (unlikely (!make_room_for (0, 1))) return Crap(hb_glyph_info_t);
                                                 ^
../src/hb-null.hh:93:50: note: expanded from macro 'Crap'
#define Crap(Type) Crap<typename hb_remove_const<typename hb_remove_reference<Type>::value>::value>()
                                                 ^
In file included from ../src/hb-aat-map.cc:31:
In file included from ../src/hb-aat-layout.hh:32:
In file included from ../src/hb-ot-shape.hh:32:
In file included from ../src/hb-ot-map.hh:32:
../src/hb-buffer.hh:229:50: warning: 'typename' occurs outside of a template [-Wc++11-extensions]
../src/hb-null.hh:93:25: note: expanded from macro 'Crap'
#define Crap(Type) Crap<typename hb_remove_const<typename hb_remove_reference<Type>::value>::value>()
                        ^
```

And we disabled llvm-gccc because of it, https://github.com/harfbuzz/harfbuzz/commit/e1b8e933f0

But we need to do something about and reviving llvm-gcc bots afterward (till we require C++11 officially).

### LINKED ISSUE #1419: hb-blob.cc:207: error: using 'typename' outside of template
during build process for harfbuzz-2.1.3 getting following error.
is there any patch to get rid of this build error?

hb-blob.cc: In function 'hb_blob_t* hb_blob_get_empty()': 
hb-blob.cc:207: error: using 'typename' outside of template

ebraminio:
Fixes #1341 
Fixes #1419


src/hb.hh
```diff
@@ -149,6 +149,17 @@ struct _hb_alignof
 
 #endif /* __cplusplus < 201103L */
 
+/* https://github.com/harfbuzz/harfbuzz/issues/1341 */
+#ifdef HB_NO_TYPENAME_OUTSIDE_TEMPLATE
+# ifndef typename_outside_template
+#  define typename_outside_template typename
```
behdad: The logic seems reversed?  You want no typename here, and typename in the next?

behdad: Also, the block is a long one to do something simple.  Basically, you are just doing:
```
#ifndef typename_outside_template
#  define typename_outside_template typename
#endif
#ifdef HB_NO_TYPENAME_OUTSIDE_TEMPLATE
#  undef typename_outside_template
#  define typename_outside_template
#endif 
```

ebraminio:
It won't work without typedef? https://circleci.com/gh/harfbuzz/harfbuzz/59876

behdad:
> It won't work without typedef? https://circleci.com/gh/harfbuzz/harfbuzz/59876

¯\_(ツ)_/¯

behdad:
> It won't work without typedef? https://circleci.com/gh/harfbuzz/harfbuzz/59876

¯\_(ツ)_/¯

behdad:
Actually I have an idea.  Let me try.

behdad:
Pushed a fix for https://github.com/harfbuzz/harfbuzz/issues/1419
Re-enable your fave compiler and see...

ebraminio:
>  Re-enable your fave compiler and see...

Hehehe, never ignore a bot ;)

behdad:
No Bot Left Behind...

ebraminio:
@ryandesign Hey there. You can enable gcc-llvm-4.2 now again :) (after the release)

PR #1396, opened 2018-11-20
Implement a simple API for fetching opentype metrics

ebraminio:
hb_ot_metrics_get

Related to #1337

Issue #1337, opened 2018-10-30
Exposing OS/2 table?


khaledhosny:
I’m working on a small project using HarfBuzz and I need to access a couple of fields from `OS/2` table. It is a simple table and I can just read it in my code, but HarfBuzz already reads it for the font extents so I thought to ask if exposing the rest of the table is something HarfBuzz might want to do?

behdad:
If we can model them in a nice way, sure. Which fields?

khaledhosny:
I currently only want sxHeight and sCapHeight, but I guess the API would expose anything. How that API would look is really a tough question, I have been thinking about it and I don’t have an answer. The only similar API I know is [FT_Get_Sfnt_Table](https://www.freetype.org/freetype2/docs/reference/ft2-truetype_tables.html#FT_Get_Sfnt_Table), I like how it can handle newer table versions with additional fields, but I don’t really know if it is a good API design-wise or if it is a good fit for HarfBuzz.

brawer:
Perhaps an API could be based on value tags from the [MVAR table](https://docs.microsoft.com/en-us/typography/opentype/spec/mvar#value-tags)? For example, `xhgt` for OS/2.sxHeight, `cpht` for OS/2.sCapHeight, and `undo` for post.underlinePosition. For a variable font, the implementation would interpolate according to MVAR; for a non-variable font, it would retrieve the value from the respective tables in the font.

```c
HB_EXTERN hb_bool_t
hb_ot_font_get_metrics (hb_font_t *font,
                        hb_tag_t tag,  // eg. hasc, hlgp, undo, xhgt, cpht
                        hb_position_t *position);
```

behdad:
Thanks Sascha.  Yes, that's indeed what I liked before.  A couple changes:
1. The tag should be it's own enum. `hb_ot_metric_t`,
2. API name should say "metric" instead of "metrics".  Not sure if `hb_ot_font_get_metric` or `hb_ot_metric_get`.  Like neither of them.  Anyway, this goes into `hb-ot-metric.h`.

khaledhosny:
`OS/2` has more than metrics, they might not all be useful but why not just expose them all. I guess the enum would then be `hb_ot_os2_field_t` and the function would be `hb_ot_font_get_os2_fields()` or something like that. This can then be a model for other simple tables.

behdad:
I prefer not to expose table by table.

[...]

PR #1432, opened 2018-11-30
Implement a simple API for fetching opentype metrics

ebraminio:
Continuing #1396
Related to #1337

### DIFF
--- a/src/hb-ot-hmtx-table.hh
+++ b/src/hb-ot-hmtx-table.hh
@@ -31,6 +31,7 @@
 #include "hb-ot-hhea-table.hh"
 #include "hb-ot-os2-table.hh"
 #include "hb-ot-var-hvar-table.hh"
+#include "hb-ot-metrics.hh"
 
 /*
  * hmtx -- Horizontal Metrics
@@ -88,22 +89,22 @@ struct hmtxvmtx
 
   template<typename Iterator,
            hb_requires (hb_is_iterator (Iterator))>
-  void serialize (hb_serialize_context_t *c, 
-                  Iterator it, 
+  void serialize (hb_serialize_context_t *c,
+                  Iterator it,
                   unsigned num_advances)
   {
     unsigned idx = 0;
     + it
     | hb_apply ([c, &idx, num_advances] (const hb_item_type<Iterator>& _)
                 {
-                  if (idx < num_advances) 
+                  if (idx < num_advances)
                   {
                     LongMetric lm;
                     lm.advance = _.first;
                     lm.sb = _.second;
                     if (unlikely (!c->embed<LongMetric> (&lm))) return;
-                  } 
-                  else 
+                  }
+                  else
                   {
                     FWORD *sb = c->allocate_size<FWORD> (FWORD::static_size);
                     if (unlikely (!sb)) return;
@@ -120,12 +121,12 @@ struct hmtxvmtx
 
     T *table_prime = c->serializer->start_embed <T> ();
     if (unlikely (!table_prime)) return_trace (false);
-    
+
     accelerator_t _mtx;
     _mtx.init (c->plan->source);
     unsigned num_advances = _mtx.num_advances_for_subset (c->plan);
-    
-    auto it = 
+
+    auto it =
     + hb_range (c->plan->num_output_glyphs ())
     | hb_map ([c, &_mtx] (unsigned _)
 	{
@@ -161,29 +162,16 @@ struct hmtxvmtx
                unsigned int default_advance_ = 0)
     {
       default_advance = default_advance_ ? default_advance_ : hb_face_get_upem (face);
+      ascender = 0; descender = 0; line_gap = 0;
 
-      bool got_font_extents = false;
-      if (T::os2Tag != HB_TAG_NONE && face->table.OS2->is_typo_metrics ())
-      {
-	ascender = abs (face->table.OS2->sTypoAscender);
-	descender = -abs (face->table.OS2->sTypoDescender);
-	line_gap = face->table.OS2->sTypoLineGap;
-	got_font_extents = (ascender | descender) != 0;
-      }
+      hb_ot_metrics_get_position_internal (face, (hb_ot_metrics_t) T::ascenderTag, &ascender);
+      hb_ot_metrics_get_position_internal (face, (hb_ot_metrics_t) T::descenderTag, &descender);
+      hb_ot_metrics_get_position_internal (face, (hb_ot_metrics_t) T::lineGapTag, &line_gap);
+      ascender = fabs (ascender);
+      descender = -fabs (descender);
+      has_font_extents = ascender != 0 || descender != 0;
 
-      hb_blob_t *_hea_blob = hb_sanitize_context_t().reference_table<H> (face);
-      const H *_hea_table = _hea_blob->as<H> ();
-      num_advances = _hea_table->numberOfLongMetrics;
-      if (!got_font_extents)
-      {
-	ascender = abs (_hea_table->ascender);
-	descender = -abs (_hea_table->descender);
-	line_gap = _hea_table->lineGap;
-	got_font_extents = (ascender | descender) != 0;
-      }
-      hb_blob_destroy (_hea_blob);
-
-      has_font_extents = got_font_extents;
+      num_advances = T::is_horizontal ? face->table.hhea->numberOfLongMetrics : face->table.vhea->numberOfLongMetrics;
 
       table = hb_sanitize_context_t().reference_table<hmtxvmtx> (face, T::tableTag);
 
@@ -279,9 +267,9 @@ struct hmtxvmtx
 
     public:
     bool has_font_extents;
-    int ascender;
-    int descender;
-    int line_gap;
+    float ascender;
+    float descender;
+    float line_gap;
 
     protected:
     unsigned int num_metrics;
@@ -322,12 +310,18 @@ struct hmtxvmtx
 struct hmtx : hmtxvmtx<hmtx, hhea> {
   static constexpr hb_tag_t tableTag = HB_OT_TAG_hmtx;
   static constexpr hb_tag_t variationsTag = HB_OT_TAG_HVAR;
-  static constexpr hb_tag_t os2Tag = HB_OT_TAG_OS2;
+  static constexpr hb_tag_t ascenderTag = HB_OT_METRICS_HORIZONTAL_ASCENDER;
+  static constexpr hb_tag_t descenderTag = HB_OT_METRICS_HORIZONTAL_DESCENDER;
+  static constexpr hb_tag_t lineGapTag = HB_OT_METRICS_HORIZONTAL_LINE_GAP;
+  static constexpr bool is_horizontal = true;
 };
 struct vmtx : hmtxvmtx<vmtx, vhea> {
   static constexpr hb_tag_t tableTag = HB_OT_TAG_vmtx;
   static constexpr hb_tag_t variationsTag = HB_OT_TAG_VVAR;
-  static constexpr hb_tag_t os2Tag = HB_TAG_NONE;
+  static constexpr hb_tag_t ascenderTag = HB_OT_METRICS_VERTICAL_ASCENDER;
+  static constexpr hb_tag_t descenderTag = HB_OT_METRICS_VERTICAL_DESCENDER;
+  static constexpr hb_tag_t lineGapTag = HB_OT_METRICS_VERTICAL_LINE_GAP;
+  static constexpr bool is_horizontal = false;
 };
 
 struct hmtx_accelerator_t : hmtx::accelerator_t {};
--- a/src/hb-ot-metrics.cc
+++ b/src/hb-ot-metrics.cc
@@ -1,0 +1,135 @@
+/*
+ * Copyright © 2018  Ebrahim Byagowi
+ *
+ *  This is part of HarfBuzz, a text shaping library.
+ *
+ * Permission is hereby granted, without written agreement and without
+ * license or royalty fees, to use, copy, modify, and distribute this
+ * software and its documentation for any purpose, provided that the
+ * above copyright notice and the following two paragraphs appear in
+ * all copies of this software.
+ *
+ * IN NO EVENT SHALL THE COPYRIGHT HOLDER BE LIABLE TO ANY PARTY FOR
+ * DIRECT, INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES
+ * ARISING OUT OF THE USE OF THIS SOFTWARE AND ITS DOCUMENTATION, EVEN
+ * IF THE COPYRIGHT HOLDER HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ *
+ * THE COPYRIGHT HOLDER SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING,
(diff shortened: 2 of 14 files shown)


src/hb-ot-metrics.h
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * hb_ot_metrics_t:
+ *
+ * From https://docs.microsoft.com/en-us/typography/opentype/spec/mvar#value-tags
+ *
+ * Since: REPLACEME
+ **/
+typedef enum {
+  HB_OT_METRICS_HORIZONTAL_ASCENDER		= HB_TAG ('h','a','s','c'), /* OS/2.sTypoAscender */
```
ebraminio: Maybe should be mentioned this tried to fetch hhea.ascender first then fallback to os2 if we decide that eventually

src/hb-ot-var-mvar-table.hh
```diff
@@ -76,12 +76,13 @@ struct MVAR
   inline float get_var (hb_tag_t tag,
 			const int *coords, unsigned int coord_count) const
   {
+    if (!version.major) return 0.;
```
ebraminio: Not needed optimization maybe?

behdad: Not needed.

src/hb-ot-metrics.h
behdad: No.  I'm leaning towards removing the implementation-details comments in fact.  Not helpful to the casual user.

src/hb-ot-metrics.cc
```diff
@@ -0,0 +1,117 @@
...
+  case HB_OT_METRICS_VERTICAL_DESCENDER:          return GET_X_METRIC (vmtx, descender);
+  case HB_OT_METRICS_VERTICAL_LINE_GAP:           return GET_X_METRIC (vmtx, line_gap);
+  case HB_OT_METRICS_HORIZONTAL_CARET_RISE:       return GET_Y_METRIC (hhea, caretSlopeRise);
+  case HB_OT_METRICS_HORIZONTAL_CARET_RUN:        return GET_X_METRIC (hhea, caretSlopeRun);
+  case HB_OT_METRICS_HORIZONTAL_CARET_OFFSET:     return GET_X_METRIC (hhea, caretOffset);
+  case HB_OT_METRICS_VERTICAL_CARET_RISE:         return GET_X_METRIC (vhea, caretSlopeRise);
+  case HB_OT_METRICS_VERTICAL_CARET_RUN:          return GET_Y_METRIC (vhea, caretSlopeRun);
+  case HB_OT_METRICS_VERTICAL_CARET_OFFSET:       return GET_Y_METRIC (vhea, caretOffset);
+  case HB_OT_METRICS_X_HEIGHT:                    return GET_Y_METRIC (OS2->v2 (), sxHeight);
+  case HB_OT_METRICS_CAP_HEIGHT:                  return GET_Y_METRIC (OS2->v2 (), sCapHeight);
```
behdad: I'm confused.  So, `->v2 ()` works now, even though that's a reference and you expect a pointer here?

src/hb-ot-metrics.cc
```diff
@@ -0,0 +1,117 @@
...
+hb_ot_metrics_get (hb_font_t       *font,
+		   hb_ot_metrics_t  metrics_tag,
+		   hb_position_t   *position     /* OUT.  May be NULL. */)
+{
+  switch (metrics_tag)
+  {
+#define GET_METRIC(TABLE, ATTR, DIR) \
+  (font->face->table.TABLE->has_data () && \
+    (position && (*position = font->em_scale_##DIR (font->face->table.TABLE->ATTR + \
+						    font->face->table.MVAR->get_var (metrics_tag, nullptr, 0))), true))
```
behdad: The more I think about it, the more I'm convinced that we want sane fallback metrics in this API even if the respective tables are not available.  So, I'm not sure if a simple macro will do it...

src/hb-ot-metrics.cc
```diff
@@ -0,0 +1,117 @@
...
+ * @position:    (out) (optional):
+ *
+ * Returns: Whether found the requested metrics
+ *
+ * Since: REPLACEME
+ **/
+hb_bool_t
+hb_ot_metrics_get (hb_font_t       *font,
+		   hb_ot_metrics_t  metrics_tag,
+		   hb_position_t   *position     /* OUT.  May be NULL. */)
```
behdad: Ok.  Thought about more of the API.  I like to have this, in this file:

- `hb_ot_metrics_get_position(font, metrics_tag, *position)` for current API,
- `hb_position_t hb_ot_metrics_get_x_variation(font, metrics_tag)`
- `hb_position_t hb_ot_metrics_get_y_variation(font, metrics_tag)`

The current API (get_position) needs to fill in position with sane fallback, such that clients can continue even if it returned false.  We need to pick fallback values one by one.

Re hmtx/vmtx, there's a chicken-and-egg issue.  I'll comment on that separately.

behdad:
So previously I said we should get ascender/descender/linegap form hmtx/vmtx accelerators.  Now I think it should be the other way around: those should get it from this API.  Which means, this API should implement the logic there.  The logic, for horizontal ascender/descender is that if `face->table.OS2->fsSelection & USE_TYPO_METRICS` then OS/2 values are used, otherwise hhea values.

Now, MVAR specifically refers to OS2 values, because there's wording somewhere else in OpenType 1.8 that says variable fonts should always have USE_TYPO_METRICS bit one.  We can ignore that wording and continue the above logic.  So, that logic needs to be implemented here if both tables are present.

src/hb-ot-metrics.cc
ebraminio: Hmm, a font without OS2 is not that imaginable I believe.

src/hb-ot-metrics.cc
behdad: Oh!  I wasn't suggesting you overload it.  Which, honestly, doesn't make sense.  I was literally meaning if you write `v2 ().operator ->`.  Don't know if that works even if you haven't overloaded it.  Fine for now I guess.  I just don't like a particular use of a class dictating the design of the class.

src/hb-ot-metrics.cc
behdad: > Hmm, a font without OS2 is not that imaginable I believe.

We don't work with "that imaginable".  We must handle all cases that happen.

src/hb-ot-var-mvar-table.hh
ebraminio: Done

src/hb-ot-metrics.cc
```diff
@@ -115,3 +126,11 @@ hb_ot_metrics_get (hb_font_t       *font,
   default:                                        return false;
   }
 }
+
+hb_position_t
+hb_ot_metrics_get_x_variation(hb_font_t *font, hb_ot_metrics_t metrics_tag)
```
behdad: Space before '('.

src/hb-ot-metrics.cc
```diff
@@ -69,9 +69,20 @@ hb_ot_metrics_get (hb_font_t       *font,
 #define GET_X_METRIC(TABLE, ATTR) GET_METRIC (TABLE, ATTR, x)
 #define GET_Y_METRIC(TABLE, ATTR) GET_METRIC (TABLE, ATTR, y)
 
-  case HB_OT_METRICS_HORIZONTAL_ASCENDER:         return GET_Y_METRIC (hmtx, ascender);
-  case HB_OT_METRICS_HORIZONTAL_DESCENDER:        return GET_Y_METRIC (hmtx, descender);
-  case HB_OT_METRICS_HORIZONTAL_LINE_GAP:         return GET_Y_METRIC (hmtx, line_gap);
+#define USE_TYPO_METRICS (1u<<7)
+  case HB_OT_METRICS_HORIZONTAL_ASCENDER:
+    return ((font->face->table.OS2->fsSelection & USE_TYPO_METRICS == 0) &&
+	    GET_Y_METRIC (hhea, ascender)) ||
```
behdad: It would read more natural to test for `^ USE_TYPO_METRICS) && OS2...) || hhead...`.

src/hb-ot-hmtx-table.hh
```diff
@@ -288,9 +274,9 @@ struct hmtxvmtx
 
     public:
     bool has_font_extents;
-    int ascender;
-    int descender;
-    int line_gap;
+    hb_position_t ascender;
+    hb_position_t descender;
+    hb_position_t line_gap;
```
behdad: No, these are not hb_position_t.  Keep int.

src/hb-ot-hmtx-table.hh
```diff
@@ -190,33 +190,19 @@ struct hmtxvmtx
...
-	  got_font_extents = (ascender | descender) != 0;
-	}
-      }
+      /* XXX Shouldn't create a font */
+      hb_font_t *font = hb_font_create (face);
+      hb_ot_metrics_get_position (font, (hb_ot_metrics_t) T::ascenderTag, &ascender);
+      hb_ot_metrics_get_position (font, (hb_ot_metrics_t) T::descenderTag, &descender);
+      hb_ot_metrics_get_position (font, (hb_ot_metrics_t) T::lineGapTag, &line_gap);
+      ascender = abs (ascender);
+      descender = -abs (descender);
```
behdad: Umm.  I wonder if this abs business should happen in the metrics_get_position()...  We want that API and the `font_extents` API to produce same values, no?

src/hb-ot-hmtx-table.hh
```diff
@@ -190,33 +190,19 @@ struct hmtxvmtx
...
-	if (0 != (face->table.OS2->fsSelection & USE_TYPO_METRICS))
-	{
-	  ascender = abs (face->table.OS2->sTypoAscender);
-	  descender = -abs (face->table.OS2->sTypoDescender);
-	  line_gap = face->table.OS2->sTypoLineGap;
-	  got_font_extents = (ascender | descender) != 0;
-	}
-      }
+      /* XXX Shouldn't create a font */
+      hb_font_t *font = hb_font_create (face);
```
behdad: Right.  Then, add an internal API that takes a face, and doesn't scale.  It should return a float.  Oh, yeah, sine we support variations now, we should store float, not int.

jfkthame:
On 30/11/2018 20:37, Ebrahim Byagowi wrote:

> Hmm, a font without OS2 is not that imaginable I believe.

Not sure if it's still true, but I think I remember that Apple used to 
ship some fonts that had no OS/2 table. (FWIW, it's documented as an 
optional table in their TrueType spec.)

JK

ebraminio:
Big majority of them have OS/2 but right there are apparently some that don't (comparing number of maxp and os/2 occurrence)

10.12.6 https://gist.github.com/ebraminio/7627d0e18989a25b16c55701118fc069

10.13.6 https://gist.github.com/ebraminio/d432e831b3f7ebe30245dde5775e1c7e

src/hb-ot-metrics.cc
khaledhosny: I’m pretty sure I have seen fonts in the wild (most likely on macOS) without OS/2 table.

src/hb-ot-metrics.cc
ebraminio: Done

src/hb-ot-hmtx-table.hh
ebraminio: Done

src/hb-ot-metrics.cc
ebraminio: Done

src/hb-ot-metrics.h
ebraminio: Gone.

src/hb-ot-metrics.cc
ebraminio: The API part is done. Don't know about the fallback part yet...

src/hb-ot-hmtx-table.hh
ebraminio: Done

src/hb-ot-metrics.cc
```diff
@@ -0,0 +1,136 @@
...
+  switch (metrics_tag)
+  {
+#define GET_METRIC(TABLE, ATTR, DIR) \
+  (font->face->table.TABLE->has_data () && \
+    (position && (*position = font->em_scale_##DIR (font->face->table.TABLE->ATTR + \
+						    font->face->table.MVAR->get_var (metrics_tag, nullptr, 0))), true))
+#define GET_X_METRIC(TABLE, ATTR) GET_METRIC (TABLE, ATTR, x)
+#define GET_Y_METRIC(TABLE, ATTR) GET_METRIC (TABLE, ATTR, y)
+
+#define USE_TYPO_METRICS (1u<<7)
```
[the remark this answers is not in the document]
ebraminio: Done

src/hb-ot-metrics.cc
ebraminio: OK. Let me know what should happen case by case.

src/hb-ot-metrics.cc
```diff
@@ -0,0 +1,165 @@
...
+ * @font:
+ * @metrics_tag:
+ * @position:    (out) (optional):
+ *
+ * Returns: Whether found the requested metrics
+ *
+ * Since: REPLACEME
+ **/
+hb_bool_t
+hb_ot_metrics_get_position (hb_font_t       *font,
```
ebraminio: I wonder why we should `font` when this can use `face` only.

khaledhosny: I think all other HarfBuzz APIs that return positions return scaled ones, so it would be a departure to return unscaled positions here.

ebraminio: Oh. Ok. Thanks

ebraminio:
And cache/lazy load whole values in the accelerator, hmm, apparently little by little I am digesting what you said before :)

behdad:
What's the status here?  Let's finish and get this in.

behdad:
Can you rework this?  We like to use it in Pango.

@cc matthiasclasen

ebraminio:
Sure, I'd love to

ebraminio:
> Can you rework this? We like to use it in Pango.

Ok. Still not in ideal state for working on codes but this feels OK I guess? I mean what feels bad about this?

behdad:
This looks very good.  A couple things:
* What's the justification for separate -internal.cc file and the regular .cc file?
* Let's add `hb_ot_metrics_get_variation` that doesn't do any scaling.  I have a feeling that we will be encoding non-geometric values in there at some point, if it's not already...  Anyway, that one should probably return a float.
* I see that `hb_ot_metrics_get_x/y_variation` also return float.  Probably a good idea, though technically should be `hb_position_t`.  Not sure :(.
* I have to think through the ascent/descent etc changes.  I'll do that when merging.

ebraminio:
> What's the justification for separate -internal.cc file and the regular .cc file?

You wanted to reuse [abs logic](https://github.com/harfbuzz/harfbuzz/pull/1432#discussion_r238016656) but we don't want to use API in hb-subset, which uses the table, so I had to separate it and use in the subsetter.

> Let's add hb_ot_metrics_get_variation that doesn't do any scaling. I have a feeling that we will be encoding non-geometric values in there at some point, if it's not already... Anyway, that one should probably return a float.

Done, I think

> I see that hb_ot_metrics_get_x/y_variation also return float. Probably a good idea, though technically should be hb_position_t. Not sure :(.

I turned it into hb_position_t as the scale already removed its precision but your concern is understandable.

PR #1442, opened 2018-12-04
Use aat_layout_has_ in coretext-aat

ebraminio:
Maybe both are wrong...


src/hb-coretext.h
```diff
@@ -40,11 +40,6 @@
 HB_BEGIN_DECLS
 
 
-#define HB_CORETEXT_TAG_MORT HB_TAG('m','o','r','t')
-#define HB_CORETEXT_TAG_MORX HB_TAG('m','o','r','x')
-#define HB_CORETEXT_TAG_KERX HB_TAG('k','e','r','x')
```
khaledhosny: This is a public header, so I don’t think we can just drop these macros (unless they are defined elsewhere).

ebraminio: Correct, just was wondering if such is possible as it is not being a part of ABI

behdad: Right.  Revert this change.  Can be moved to deprecated, but I don't see any reason to do that right now, unless we also do for GSUB/... tags.  Or just leave them as is, my preference right now.  The rest looks good.  Thanks.

ebraminio: `hb-coretext.h` was only available on macOS so I guessed this particular removal was not that trouble making... reverted.

behdad: > `hb-coretext.h` was only available on macOS so I guessed this particular removal was not that trouble making...

I think Chromium was actually using these?

behdad: Anyway.  Don't fix it if it ain't broken ;).

ebraminio: They and Qt aren't using it, checked before.

Issue #1443, opened 2018-12-04
[cff] Some CFF files fail to build with warnings in Chrome


drott:
Compare Chromium build output from:

https://logs.chromium.org/logs/chromium/buildbucket/cr-buildbucket.appspot.com/8928054094769963200/+/steps/compile__with_patch_/0/stdout

via the experimental HB roll CL in:
https://chromium-review.googlesource.com/c/chromium/src/+/1361067

```
FAILED: obj/third_party/harfbuzz-ng/harfbuzz_source/hb-ot-cff1-table.o 
[...]

PR #1444, opened 2018-12-04
removed assert

blueshade7:
get rid of assert for issue #1443 and handle default case silently

### DIFF
--- a/src/hb-ot-cff1-table.hh
+++ b/src/hb-ot-cff1-table.hh
@@ -654,6 +654,7 @@ struct NameDictValues
   static inline enum NameDictValIndex name_op_to_index (OpCode op)
   {
     switch (op) {
+      default:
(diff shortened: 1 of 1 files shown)


behdad:
Silently ignoring would be more like "return 0" or something to me.  This is defaulting to opCode_version, which I find weird.

behdad:
Or, it definitely cannot happen, even for bad input?

blueshade7:
When double-checking the code to say "can't happen", I realized that `OpCode_BaseFontName` is not handled when it must be. I will open an issue for it.

blueshade7:
At first I had OpCodes as enum actually, then some compiler complained about value duplicates. I could have split it in two, OpCode_Dict and OpCode_CS and cast between two as necessary, but simply just used macro constants.

PR #1454, opened 2018-12-05
[CFF] fix oss-fuzz issue 11674: parse_bcd

blueshade7:
To address oss-fuzz issue 11674 (integer overflow), added overflow checks for the fractional part and the exponent part against double precision limits. The integral part is accumulated to double, so should automatically saturate.

Also fixed a bug with a fractional part in a negative number, discovered during testing above code change.

Minor code cleanups.


ebraminio:
I guess you've left the testcase.

blueshade7:
Hmmm... seems codecov flagged it unsuccessful as I made too many changes.
How shall we proceed? I think I can minimize the changes to let it pass, but kind of cheating.

PR #1457, opened 2018-12-06
[CFF] oss-fuzz issue 11713 (CFF2VariationStore::serialize)

blueshade7:
Variation Store sub-table in CFF2 forgot to sanitize against the sub-table size. If the size is bad, it will end up in a bad access when the table is later serialized for subset.


behdad:
I don't understand this change.  Is size the size of the cff table?

behdad:
Humm.  Okay, I see.  I think it's an ugly design to keep the size there, but oh well.

I suggest you use the sanitizer set_object() instead, to make sure all of VarStore is within the subtable.  Would be: `hb_sanitize_with_object_t with (c, this);`

behdad:
@blueshade7

PR #1458, opened 2018-12-06
[CFF] oss-fuzz issue 11714: set_blends

blueshade7:
If there aren't enough number of values on the argument stack for CFF2 blend operator, `start` variable in function `process_blend` wraps around to become a very large unsigned value. This causes a bad access in function `set_blends` called immediately after.

Detect this condition and set the error in InterpEnv so let it fail.


behdad:
> This causes a bad access in function `set_blends` called immediately after.

Why aren't we using a safe datastructure for the stack, like hb_vector_t?  What *are* we using then?

blueshade7:
hb_vector_t is used, but passing the address to vector elements to `set_blends` is defeating the bounds check I believe. Sure we could pass vector itself along with the index would allow automatic bounds check.

behdad:
I see.  Try `as_array()` and then `sub_array()` or something?  `hb_array_t` is our light-weight bounded array type that can be passed by-value.

behdad:
`sub_array()` when given only one argument, returns an `hb_array_t` starting at that element and containing  the rest of the elements.

blueshade7:
I am about to make the suggested change but is above true? I don't see hb_array_t::sub_array() variant taking one argument.

behdad:
> I am about to make the suggested change but is above true? I don't see hb_array_t::sub_array() variant taking one argument.

You're right, apparently I didn't do that.  Doing now.

behdad:
Done now.

behdad:
Added `sub_array()` to vector as well.

PR #1459, opened 2018-12-06
[style] Initial implementation

ebraminio:
Related to #1337 (fixes it together with #1432 AFAICT)


src/Makefile.sources
```diff
@@ -118,6 +118,8 @@ HB_BASE_sources = \
 	hb-ot-shape.cc \
 	hb-ot-shape.hh \
 	hb-ot-stat-table.hh \
+	hb-ot-style.cc \
+	hb-ot-style.h \
```
ebraminio: .fon files (and I guess Type 1) could provide lots of provided things, so maybe -ot- removal?

src/hb-aat-layout-common.hh
```diff
@@ -28,6 +28,7 @@
 #define HB_AAT_LAYOUT_COMMON_HH
 
 #include "hb-aat-layout.hh"
+#include "hb-open-type.hh"
```
ebraminio: This was needed here.

ebraminio:
Now this is worth to look I think

src/Makefile.sources
ebraminio: Also whether this should have a separated header file or not.

src/hb-ot-style.cc
```diff
@@ -0,0 +1,85 @@
...
+  /* OS/2, head */
+  switch (style)
+  {
+  case HB_OT_STYLE_ITALIC:
+    return face->table.OS2->is_italic () || face->table.head->is_italic () ? 1.f : 0.f;
+  case HB_OT_STYLE_OPTICAL_SIZE:
+    return face->table.OS2->v5 ().get_optical_size (); // returns 12.f if nothing found
+  case HB_OT_STYLE_SLANT:
+    /* What should be put when font is oblique? */
+    return face->table.OS2->is_oblique () ? 6.f : 0.f;
```
ebraminio: Note this please

behdad:
I'm fine if you push the table improvements in and keep this PR for API addition and bits that need discussion.

ebraminio:
Hmm, STAT allows for multiple AxisValue for a particular AxisRecord? What will be process of retrieving a value going through STAT, iterating over "AxisValue"s by the indicated index, checking the AxisRecord tag? And what should be exposed from different formats of "AxisValue", I guess I should mark this WIP for now.

src/hb-ot-style.cc
```diff
@@ -0,0 +1,78 @@
...
+  switch (style)
+  {
+  case HB_OT_STYLE_ITALIC:
+    return face->table.OS2->is_italic () || face->table.head->is_italic () ? 1.f : 0.f;
+  case HB_OT_STYLE_OPTICAL_SIZE:
+    if (face->table.OS2->v5 ().has_optical_size ())
+      return face->table.OS2->v5 ().get_optical_size ();
+    return 12.f;
+  case HB_OT_STYLE_SLANT:
+    return face->table.OS2->is_oblique () ? 1.f : 0.f;
```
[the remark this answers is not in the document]
ebraminio: `Note: Use of format 1, format 2, or format 3 axis value tables is especially recommended for values on 'wght' and 'wdth' axes, and also for “Italic” (value 1.0 on the 'ital' axis) or “Oblique” (a non-zero value on the 'slnt' axis) variants.` from https://docs.microsoft.com/en-us/typography/opentype/spec/stat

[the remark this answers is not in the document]
ebraminio: Here I just asked if slant means oblique, this is the outdated change my question below is what degree it means when font is oblique per os2.

ebraminio:
Maybe ot-var API should handle axis part querying?

behdad:
STAT is similar, and often times a superset, of variations.  But that's the only overlap.  It can be completely non-variations.  Variations can be completely non-stylistic.  Just because it's similar doesn't mean one should handle the other.

ebraminio:
Correct, I think that was a TODO somewhere and only in hb repo STAT having fonts are the ones with variation, we should have a thread for stat API imagination I think however will such API user friendly enough for an imagined font manager.

behdad:
> Hmm, I think I understand something now, an API should take a tag and fill an array of axis value records (mix, max, current/avg, name id?) and if the STAT table doesn't exist or doesn't have the value for the tag, only returning one record filled with whatever value we could find.

I'm skeptical about tag->xxx; IMO should be an axis_index.

Let's not start on this API design right now.  I want to finish metrics API and merge first.

ebraminio:
```
struct hb_ot_style_record_t {
  hb_ot_name_id_t name_id;
  float from;
  float to;
  float default;
  /* what else? */
};

unsigned int
hb_ot_style_get (
  hb_face_t face,
  hb_ot_style_t style, /* just a tag */
  unsigned int *count, /* IN/OUT */
  hb_ot_style_record_t *records);

(returning number of all available records, which will be at least 1 as fallbacks)
```

ebraminio:
> I'm skeptical about tag->xxx; IMO should be an axis_index.

axis_index sounds like something not user-friendly and polyfill-able IMO.

> Let's not start on this API design right now. I want to finish metrics API and merge first.

Sorry, right, didn't see your comment first.

behdad:
> axis_index sounds like something not user-friendly and polyfill-able IMO.

HarfBuzz is low-level library.  User is expected to know how to handle things like indices.  We don't like tags in enumeration API when they add a bsearch() overhead.

I did that in hb-ot-name as an exception because the alternative was ugly.

ebraminio:
Considering no fallback will be possible with exposing axis_index (AFAICS
at least) this will look ugly and useless as well.

Currently lsearch even, maybe going for some accelerator?

Or a separate API for each needed thing? I don't like it as we were this
close to avoid that but that is what CoreText and GDI is doing I guess and
it is just easier for me as I don't care that much the STAT, my Windows
installation I just checked (not used and updated for a while but still)
had no any fonts with STAT

behdad:
> Considering no fallback will be possible with exposing axis_index (AFAICS at least) this will look ugly and useless as well.

Maybe you should stop and describe the usecases first then.

> Currently lsearch even, maybe going for some accelerator? Or a separate API for each needed thing?

What needed things?  Weight, width, other known axes you mean?

> I don't like it as we were this close to avoid that but that is what CoreText and GDI is doing I guess and it is just easier for me as I don't care that much the STAT, my Windows installation I just checked (not used and updated for a while but still) had no any fonts with STAT

Windows requires STAT table in variable fonts.

I suggest we internally synthesize a STAT table from either OS/2/post or fdsc and use that.  The code, then, can only support STAT table with no fallback code.  I think there should be no fallback code involved when a STAT table exists.

ebraminio:
> Maybe you should stop and describe the usecases first then.

Just like what we did in ot-metrics, using the tags to not expose things like this https://developer.apple.com/documentation/coretext/1510018-ctfontgetascent?language=objc or, maybe that is what we like also?

> What needed things? Weight, width, other known axes you mean?

What described here by Khaled https://github.com/harfbuzz/harfbuzz/issues/1337#issuecomment-434840286 and https://github.com/harfbuzz/harfbuzz/issues/1337#issuecomment-434858416

> I think there should be no fallback code involved when a STAT table exists.

`Arabic UI Text` from macOS 10.13, the only font have STAT there, doesn't have values for all axis.

ebraminio:
Maybe returning all the values with resolved tags in an array, appended
with fallback of not available values (so will at least return 5 values, no
guarantee for ordering, a little complicated from our side but doable),
with no mention of axis index (or -1 when is fallback) which means no
bsearch/lsearch from our side (only 5 bool to know which values do not need
fallback), and another API just to know about axis index, tags and name id
STAT really has.

ebraminio:
And cache/lazy load whole values in the STAT accelerator, hmm, apparently little by little I am digesting what you said before :)

PR #1461, opened 2018-12-07
[CFF] oss-fuzz issue 11690 ASSERT: substr.offset >= opStart

blueshade7:
With --desuborutinize option unspecified, CFF subsetter parses out charstrings & subroutines into temporary data structure `ParsedCStr` while collecting subroutine references for later subroutine subsetting and hint removal. This charstring parser doesn't expect a subroutine may be called recursively; if it happens,  an inconsistency triggers an assert in `ParsedValues::add_op`.

Fixed the problem by modified `SubrSubsetParam::set_current_str` to detect the condition, set an error so the interpreter fails. Disabled assert in `add_op`, and other code nearby (more assert removal to come).


src/hb-cff-interp-common.hh
```diff
@@ -628,7 +630,9 @@ struct ParsedValues
   {
     VAL *val = values.push ();
     val->op = op;
+#ifndef HB_NDEBUG
```
behdad: `HB_NDEBUG` is not to be used around asserts.  Just use plain assert.  From `HB_NDEBUG` description:
```
/* HB_NDEBUG disables some sanity checks that are very safe to disable and 
 * should be disabled in production systems.  If NDEBUG is defined, enable 
 * HB_NDEBUG; but if it's desirable that normal assert()s (which are very 
 * light-weight) to be enabled, then HB_DEBUG can be defined to disable 
 * the costlier checks. */ 
```
I might completely remove it.

blueshade7: I am going to remove asserts entirely.

PR #1464, opened 2018-12-08
Apply non-controversial parts of ot-style impl

ebraminio:
Things to be used in https://github.com/harfbuzz/harfbuzz/pull/1459

### DIFF
--- a/src/hb-ot-os2-table.hh
+++ b/src/hb-ot-os2-table.hh
@@ -93,9 +93,55 @@ struct OS2
 {
   enum { tableTag = HB_OT_TAG_OS2 };
 
+  inline bool has_data () const { return this != &Null (OS2); }
+
   inline const OS2V1Tail &v1 (void) const { return version >= 1 ? v1X : Null (OS2V1Tail); }
   inline const OS2V2Tail &v2 (void) const { return version >= 2 ? v2X : Null (OS2V2Tail); }
   inline const OS2V5Tail &v5 (void) const { return version >= 5 ? v5X : Null (OS2V5Tail); }
+
+  enum fs_selection_flag_t {
+    ITALIC		= 1u<<0,
+    UNDERSCORE		= 1u<<1,
+    NEGATIVE		= 1u<<2,
+    OUTLINED		= 1u<<3,
+    STRIKEOUT		= 1u<<4,
+    BOLD		= 1u<<5,
+    REGULAR		= 1u<<6,
+    USE_TYPO_METRICS	= 1u<<7,
+    WWS			= 1u<<8,
+    OBLIQUE		= 1u<<9
+  };
+
+  inline bool is_italic (void) const  { return fsSelection & ITALIC; }
+  inline bool is_oblique (void) const { return fsSelection & OBLIQUE; }
+
+  enum us_width_class_t {
+    FWIDTH_ULTRA_CONDENSED	= 1, /* 50% */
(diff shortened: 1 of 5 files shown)


test/shaping/run-tests.py
```diff
@@ -138,4 +138,5 @@ def cmd(command):
 elif passes:
 	sys.exit (0)
 else:
-	sys.exit (77)
+	# it should raise 77 on skips but that makes logs unusable
+	sys.exit (0)
```
behdad: What??  This is wrong.

src/hb-ot-os2-table.hh
```diff
@@ -93,10 +93,56 @@ struct OS2
 {
   enum { tableTag = HB_OT_TAG_OS2 };
 
+  inline bool has_data () const { return this != &Null (OS2); }
```
behdad: We typically don't like comapring to Null as a way to detect emptyness.  This has legitimate reasons (there might be different null polls for different .so objects, or even files...).  If there's a better way, use it.  Why not the version number, like we do in all other tables?!

src/hb-ot-stat-table.hh
```diff
@@ -223,10 +204,44 @@ struct AxisValue
...
+    return_trace (likely (c->check_struct (this)));
+  }
+
+  protected:
+  Tag		axisTag;	/* A tag identifying the axis of design variation. */
+  NameID	axisNameID;	/* The name ID for entries in the 'name' table that
+				 * provide a display string for this axis. */
+  HBUINT16	axisOrdering;	/* A value that applications can use to determine
+				 * primary sorting of face names, or for ordering
+				 * of descriptors when composing family or face names. */
```
behdad: Typically, members of struct `StatAxisRecord` don't need to have `axis` prefix.  Makes more readable.

test/shaping/run-tests.py
behdad: Also, doesn't belong in this PR.

src/hb-ot-os2-table.hh
ebraminio: "The version number for the OS/2 table: 0x0000 to 0x0005.", something more reliable please?

behdad: Then we need to add a custom null table for it.   Umm. Keep it as is.  I'll clean up at some point when I'll go over all such checks to Null..

src/hb-ot-stat-table.hh
```diff
@@ -223,10 +204,44 @@ struct AxisValue
...
 {
   enum { tableTag = HB_OT_TAG_STAT };
 
+  inline const AxisValue &get_axis_value (hb_tag_t style) const
+  {
+    const StatAxisRecord &record = (this+designAxesOffset).lsearch (designAxisCount, style);
+    if (!record.has_data () || record.get_ordering () >= axisValueCount)
+      return Null (AxisValue);
+
+    return this+(this+offsetToAxisValueOffsets)[record.get_ordering ()];
```
ebraminio: Plain wrong logic, let me fix...

test/shaping/run-tests.py
ebraminio: Gone

src/hb-ot-stat-table.hh
ebraminio: Thanks. Done

src/hb-ot-stat-table.hh
ebraminio: Removing whole logic for now.

PR #1467, opened 2018-12-11
[CFF] minimized assert usage

blueshade7:
Minimized usage of `assert()` in CFF code, either by:

(1) handle a failure condition as an error if it may ever happen with bad/unexpected input
(2) removed, if a failure shouldn't happen and cannot be easily handled as an error
(3) leave assert, if a failure shouldn't happen


blueshade7:
Fixes oss-fuzz issue 11804

PR #1468, opened 2018-12-11
[style] Fully expose STAT table with fallback

### LINKED ISSUE #1337: Exposing OS/2 table?

ebraminio:
Closes #1337
Not working yet (apparently accelerator won't work if table doesn't exist?) uploaded separately to see if the approach seems better.

Maybe the API should memcpy the infos? (iteration)


src/hb-ot-style.h
```diff
@@ -0,0 +1,87 @@
...
+ * Since: REPLACEME
+ */
+typedef struct
+{
+  hb_tag_t		tag;
+  hb_ot_name_id_t	axis_name_id;
+  unsigned int		axis_ordering;
+
+  unsigned int		axis_index;
+  unsigned int		value_index;
```
ebraminio: specifically for format4.

ebraminio:
>  Not working yet (apparently accelerator won't work if table doesn't exist?) uploaded separately to see if the approach seems better.

Working, but not on empty faces, which I don't know is important or not.

ebraminio:
Two things to consider if general approach looks good,
* Expose elidedFallbackNameID to make this a "full expose"
* Separate API just for axes, deciding if StatAxisRecord.{nameID,ordering} should go there instead.
* Considering designAxisSize also (which is not clear to me how)

src/hb-machinery.hh
```diff
@@ -114,6 +114,7 @@ static inline Type& StructAfter(TObject &X)
 
 #define DEFINE_SIZE_UNBOUNDED(size) \
   DEFINE_INSTANCE_ASSERTION (sizeof (*this) >= (size)) \
+  enum { null_size = (size) }; \
```
behdad: Dude!  There's a reason this line is not there!

behdad:
Not going to review this for a while, until other work settles.  Please don't merge any API changes, or machinery or other part changes.  Just pushing tables in without changing common machineries is fine.

src/hb-machinery.hh
ebraminio: I know... the intention was just to have something to start with as you can see from other added comments also

ebraminio:
Superseded by #1890

Issue #1374, opened 2018-11-09
'hb-set.hh' build errors with MSVC


johne53:
I'm seeing lots of errors when trying to build harfbuzz (from git master) with MSVC.  They relate to some recent changes in 'src/hb-set.hh' (approx line 370) where 'sizeof (pages.arrayZ()[0])' got changed to 'sizeof (pages[0])'.

A similar error occurs on the following line - where 'sizeof (page_map.arrayZ()[0])' got changed to 'sizeof (page_map[0])'.  This is the general form for each error:-

harfbuzz\src\hb-set.hh(378) : error C2666: 'hb_vector_t<Type,StaticSize>::operator []' : 4 overloads have similar conversions
      : could be 'const hb_set_t::page_t &hb_vector_t<Type,StaticSize>::operator [](unsigned int) const'
      : or 'hb_set_t::page_t &hb_vector_t<Type,StaticSize>::operator [](unsigned int)'
      : or 'built-in C++ operator[(T *, int)'
      : or 'built-in C++ operator[(const T *, int)'

I'm also seeing "multiple overload" errors in other places - e.g. for DEFINE_SIZE_ARRAY and its variants.  As it happens, this project needs to get built with MSVC2008 - although these particular errors look like they'd occur, even with the more recent versions.

khaledhosny:
Builds Visual Studio 14 (which is 3 years old already) [don’t have such errors](https://ci.appveyor.com/project/harfbuzz/harfbuzz/builds/20171131/job/yqkc4g7dyoys49qe), so it is either a limitation of this old compiler or a some issue with your setup.

johne53:
Hi Behdad - I haven't loaded your new commit yet (if it only deals with warnings) but here are the errors I see from just compiling 'hb-aat-layout.cc' (the remaining source files seem to produce the same errors):-

(line 378 - hb-set.hh) - memcpy (pages, other->pages, count * sizeof (pages[0]));
error C2666: 'hb_vector_t<Type,StaticSize>::operator []' : 4 overloads have similar conversions
    : could be 'const hb_set_t::page_t &hb_vector_t<Type,StaticSize>::operator [](unsigned int) const'
    : or 'hb_set_t::page_t &hb_vector_t<Type,StaticSize>::operator [](unsigned int)'
    : or 'built-in C++ operator[(T *, int)'
    : or 'built-in C++ operator[(const T *, int)'


[...]

PR #1474, opened 2018-12-12
Few fixes for Visual Studio builds

fanc999:
Hi,

This updates the flags passed to g-ir-scanner in the CMake build files so that it won't complain about not including the headers correctly, and updates the test programs (in C) to declare variables at the top of the block, so that things will continue to build properly.

With blessings, thank you!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -696,6 +696,8 @@ if (HB_HAVE_INTROSPECTION)
       -DHB_H_IN
       -DHB_OT_H
       -DHB_OT_H_IN
+      -DHB_AAT_H
+      -DHB_AAT_H_IN
       -DHB_GOBJECT_H
       -DHB_GOBJECT_H_IN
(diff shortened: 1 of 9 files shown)


ebraminio:
Oh, thanks. I was considering these while me changes at least. Interesting that you can still compile harfbuzz for MSVC2003 I guess(?) but we have complains on MSVC2008 version #1374 (which we were able to reduce them fortunately)

ebraminio:
Oh, thanks. I was considering these while me changes at least. Interesting that you can still compile harfbuzz for MSVC2003 I guess(?) but we have complains on MSVC2008 version #1374 (which we were able to reduce them fortunately)

fanc999:
Hi @ebraminio,

I can't say for 2003 but this certainly will fix things from Visual Studio 2008 to 2012, plus introspection builds using CMake.  Thanks a bunch though.

With blessings, thank you!

PR #1476, opened 2018-12-12
[CFF] bad offset in Index

blueshade7:
This change fixes oss-fuzz issues 11805, 11806.

If a pair of consecutive offsets in an Index table is not in the increasing order, the data length calculated from the pair may become a very large number exceeding the table end. If uncaught, `ByteStr` struct created from the index entry will end up in a crash down the road.

One way to catch this is for `CFFIndex::sanitize` to check every offset pair which takes O(n) time. Another downside is that it would reject a font containing a bad offset pair for an unused entry otherwise working fine.

Instead here the condition is detected when the length for an entry is requested, then the length is reset to zero so that no out of range access occurs.


behdad:
Yep, that's also how I had in mind: in sanitize we check the last offset to make sure that's in range.  And when accessing every other offset, we compare it to the last one.  Thanks.

PR #1477, opened 2018-12-12
[CFF] check out of range FD index

blueshade7:
This change fixes oss-fuzz issue 11807: Null-dereference READ in hb_set_t::add

A bad font dict index may be returned from `FDSelect:: get_fd` then the subroutine subsetter uses it to look up `closures.local_closures[fd]`. A null is returned ending up in a crash when used as a `hb_set_t` pointer.

Added range checks after `get_fd` calls. Also added `SubrSubsetParam::is_valid` function to check validity of both global_closure and local_closure, which may become null if hb_set_create failed.


blueshade7:
Undid SubrSubsetParam::is_valid, since all closures are already validated by SubrClosures::is_valid (if FD is in range)

PR #1479, opened 2018-12-13
hb-unicode.cc: Workaround a Visual Studio 2010 x64 optimization bug

fanc999:
Hi,

This PR attempts to workaround an optimization bug in the Visual Studio 2010 x64 compiler, where the resulting binaries will crash unless compiler optimizations are disabled for the build (i.e. changing "/O2 /Ob1" to "/Od").

This does so by inserting a pointless line that is used only on Visual Studio 2010 x64 builds in hb_unicode_funcs_make_immutable() (which calls an inline function hb_object_is_immutable() in hb-object.hh), so that the optimizer will not do something weird here.

Call stack (before the updates here) for references, where this issue is exhibited when running the test-unicode test program (among other test programs as well):

harfbuzz.dll!hb_unicode_funcs_make_immutable(hb_unicode_funcs_t * ufuncs) line 305 C++
harfbuzz.dll!hb_unicode_funcs_create(hb_unicode_funcs_t * parent) line 172 C++
harfbuzz.dll!hb_ucdn_unicode_funcs_lazy_loader_t::create() line 236 C++
harfbuzz.dll!hb_ucdn_get_unicode_funcs() line 268 C++
test-unicode.exe!main(int argc, char * * argv) line 799 C
test-unicode.exe!__tmainCRTStartup() line 555 C

With blessings, thank you!

### DIFF
--- a/src/hb-unicode.cc
+++ b/src/hb-unicode.cc
@@ -32,6 +32,13 @@
 
 #include "hb-unicode.hh"
 
+/* Workaround a Visual Studio 2010 x64 optimization bug */
+#if (defined (_M_X64) || defined (_M_AMD64)) && (defined (_MSC_VER) && (_MSC_VER >= 1600) && (_MSC_VER < 1700))
+#define HB_WORKAROUND_VS10_X64_OPT_BUG strtod ("10", nullptr);
+#else
+#define HB_WORKAROUND_VS10_X64_OPT_BUG
+#endif
+
 
 /**
  * SECTION: hb-unicode
@@ -302,6 +309,7 @@ hb_unicode_funcs_get_user_data (hb_unicode_funcs_t *ufuncs,
 void
 hb_unicode_funcs_make_immutable (hb_unicode_funcs_t *ufuncs)
 {
+  HB_WORKAROUND_VS10_X64_OPT_BUG
   if (hb_object_is_immutable (ufuncs))
     return;
 


behdad:
> (Please note that this is as of 2.2.0 with the patches to fix ambiguous overloads for older compilers, for references)

Does master build?  Which patches?

behdad:
As for the main proposal here, I'm *very* hesitant to take such an arbitrary "fix" into the codebase.

behdad:
Is there a bug report against the compiler somewhere?  Is the bug known?  Why this line?

fanc999:
Hi @behdad,

I understand your concerns... Yes, unfortunately, this bug still occurs in the master checkout I did today.

As this compiler version is quite old (and MS deleted the MS Connect stuff, unfortunately), the best one I can come up with is this is this StackOverflow item, which I think follows the situation in this bug somewhat:

https://stackoverflow.com/questions/12778471/is-this-a-compiler-optimisation-bug-or-an-undefined-behaviour

Judging from the contents of the StackOverflow article is that the bug was reported and was fixed for Visual Studio 2012 (but not 2010, as it seems from there--the MS Connect links in that article are no longer working at this point :| )  For the "why this line?" question, I think it is more or less like what the OP of the StackOverflow article is doing to try to work around this bug (please see the section "Bonus Info")

The other workaround that I managed to come up with is to remove the inline keyword in hb-object.hh for the declaration of hb_object_is_immutable().  Any advice on the preferences is appreciated.

---
Some side notes:
(these obviously belongs to another PR), building on 2008 is broken again I am getting:

d:\git_repos\harfbuzz\src\hb-ot-color-cbdt-table.hh(56) : error C2440: '=' : cannot convert from 'const OT::HBINT8' to 'hb_position_t'
        No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
d:\git_repos\harfbuzz\src\hb-ot-color-cbdt-table.hh(57) : error C2440: '=' : cannot convert from 'const OT::HBINT8' to 'hb_position_t'
        No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called

But this works for 2010 builds, except that I also needed to change the llu suffix to ull for line 118 (or so):
const uint64_t MAX_FRACT = 0xFFFFFFFFFFFFFllu; /* 1^52-1 */

There are also some C tests that need to go back to C89-style variable declarations for them to build as well.
---
With blessings,
Thank you!

behdad:
> There are also some C tests that need to go back to C89-style variable declarations for them to build as well:

Please send that as a separate PR.  I have the warning enabled now, so will keep them in good shape in the future.

behdad:
> d:\git_repos\harfbuzz\src\hb-ot-color-cbdt-table.hh(56) : error C2440: '=' : cannot convert from 'const OT::HBINT8' to 'hb_position_t'
> No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
> d:\git_repos\harfbuzz\src\hb-ot-color-cbdt-table.hh(57) : error C2440: '=' : cannot convert from 'const OT::HBINT8' to 'hb_position_t'
> No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called

So bizarre.  Just add a `(int) ` cast please.

behdad:
I don't like accepting this hack.  Sorry about that.  Thanks.

fanc999:
Hi @behdad,

No worries, and it also seems to me that this patch is becoming irrelevant as it looks like HarfBuzz is really requiring a C++11 compiler (so 2.40 seems like the last release that would build under 2010 or earlier without big changes)

Thanks for the time and patience.

With blessings, thank you!

Issue #1478, opened 2018-12-13
Remove coretext_aat shaper


behdad:
Right now that is still picking up AAT fonts if coretext is enabled.  Time to remove I think.  Dominik, are we safe from Chrome's point of view?

behdad:
@drott

drott:
If you don'd mind, I'd prefer for it to pass one stable release, according to
https://storage.googleapis.com/chromium-find-releases-static/c67.html#c67f53b0b70f33c47159d37f7e59bb44399b0d09 Chrome enabled HarfBuzz-native AAT shaping in 72. And according to https://www.chromium.org/developers/calendar the estimated stable release date for 72 is mid January, and for 73 is mid-March 2019. 

So since 72 is not in stable yet, I would like to see if anything breaks when it hits stable, and during that time be able to switch back to CoreText-AAT if needed without having to roll back to an old HarfBuzz as a whole.

So, I'd say - if there is no rush, I'd say regarding removal, we can revisit this in March 2019 then.

No objections for deprioritizing hb-coretext shaper choice, if there is a well-tested way to  invoke the old coretext-aat-preferred behavior until March 2019.

behdad:
Can we lower the priority?  That would be trivially-revertible.

behdad:
Upon thinking more, I don't want to remove it.  Just change priority.

khaledhosny:
If we moved it after default shaper, it will be effectively unused (except for clients that select shapers explicitly), so why keep it (we might want to keep the `coretext` shaper, similar to `uniscribe` and `directwrite` ones, but `coretext_aat` would be of little use).

behdad:
> If we moved it after default shaper, it will be effectively unused (except for clients that select shapers explicitly), so why keep it (we might want to keep the `coretext` shaper, similar to `uniscribe` and `directwrite` ones, but `coretext_aat` would be of little use).

`coretext_aat` can still be used by clients to route only AAT fonts to CoreText.

khaledhosny:
Of course, but why? To be future-proof in case AAT grew new stuff HarfBuzz does not support?

behdad:
> Of course, but why? To be future-proof in case AAT grew new stuff HarfBuzz does not support?

No, more like to emulate current behavior.

ebraminio:
[...]

PR #1484, opened 2018-12-17
Remove assumption about Core Text working in 96 DPI

torarnv:
Core Text doesn't actually have a concept of DPI internally, as it
doesn't rasterize anything by itself, it just generates vector paths
that get passed along to Core Graphics.

In practice this means Core Text operates in the classical macOS
logical DPI of 72, with one typographic point corresponding to one
point in the Core Graphics coordinate system, which for a normal
bitmap context then corresponds to one pixel -- or two pixels for
a "retina" context with a 2x scale transform.

Scaling the font point sizes given to HarfBuzz to an assumed DPI
of 96 is problematic with this in mind, as fonts with optical
features such as 'trak' tables for tracking, or color glyphs,
will then base the metrics off of the wrong point size compared
to what the client asked for.

This in turn causes mismatches between the metrics of the shaped
text and the actual rasterization, which doesn't include the 72
to 96 DPI scaling.

If a 96 DPI is needed, such as on the Web, the scaling should be
done outside of HarfBuzz, allowing the client to keep the DPI of
the shaping in sync with the rasterization.

The recommended way to do that is by scaling the font point size,
not by applying a transform to the target Core Graphics context,
to let Core Text choose the right optical features of the target
point size, as described in WWDC 2015 session 804:

  https://developer.apple.com/videos/play/wwdc2015/804/


behdad:
Can you please step back and describe what problem you are facing, without suggesting a solution right away?

torarnv:
> Basically you're saying CoreText points do not represent any particular DPI, so let's assume 72.

Exactly. That's based on the information i have from a Core Text engineer.

>  I'm saying, let's assume 96, because that's what WebKit implies.

Where does it assume that? Curious to look at the code :smiley: 

> Can you please step back and describe what problem you are facing, without suggesting a solution right away?

That would be the introductory e-mail I sent you a few days ago, describing our Qt use-case :smiley: 

In Qt we treat macOS as having a logical DPI of 72, and fill the glyph cache by rendering fonts created via CTFontCreateWithFontDescriptor, where we pass the fonts size as typographic points, without any scaling like Harbuzz does today, and the result looks size-wise identical to e.g. TextEdit.app. (72pt Helvetica).

image

Same when creating a simple Cocoa app with a label set to Helvetica 72pt:

image

Doesn’t this indicate the logic in `coretext_font_size_from_ptem` and `coretext_font_size_to_ptem` is unnecessary, and results in CoreText potentially giving us metrics etc based on the wrong font size?

I think the important part is that the DPI used when shaping matches the one used when drawing it. With HarfBuzz applying a 96DPI for the shaping, but Qt keeping the font size as is for the drawing, the two are no longer in sync. Scaling Qt's drawing up to match HarfBuzz results in larger text than native controls of the same point size.

torarnv:
> https://developer.apple.com/library/archive/documentation/GraphicsAnimation/Conceptual/HighResolutionOSX/Explained/Explained.html

Hmm, it talks about retina, where 1 point isn't a pixel, but I can't see anything in those docs saying Core Text is 96 DPI?

behdad:
> ...created via CTFontCreateWithFontDescriptor, where we pass the fonts size as typographic points...

You are implying here, that `CTFontCreateWithFontDescriptor` takes "typographic points".  I was tripped by that before as well.  My current understanding is that, that is not the case, and CoreText takes CoreGraphics points.  I'm also sure some CoreText engineers have mixed and match the two over time...

behdad:
> Hmm, it talks about retina, where 1 point isn't a pixel, but I can't see anything in those docs saying Core Text is 96 DPI?

It's saying that it's not the same as typographic points.  It's some undefined thing apparently.  It's physical size differs on different systems, but it's roughly the same size...

behdad:
Okay, I'm starting to see what you are claiming.  Let me go check my archives some more.

torarnv:
> It's saying that it's not the same as typographic points. It's some undefined thing apparently. It's physical size differs on different systems, but it's roughly the same size...

I read that whole section as teaching users that what they previously though of as pixel sizes (even though they were named points), can no longer be relied on to be pixel sizes, due to the introduction of retina resolutions and device-pixel-ratios/scale factors.

The mentioning of the typographic point is just to clarify that although the term "point" is also used for typography, it doesn't have any relation to the point used in positions/sizes in e.g. AppKit, and one should not assume a _physical_ DPI of 72.

behdad:
> > I'm saying, let's assume 96, because that's what WebKit implies.
> 
> Where does it assume that? Curious to look at the code

@drott and I arrived at this the following way: looking at Chrome vs Safari's shaping San Francisco font spacing, both through CoreText, we found that if we passed the CSS "point" size (= typographic points = 72dpi) to CoreText, then the spacing will be different from what Safari renders, but if we scale it to CSS "pixel" size (= 96dpi) then the spacing matches.

Now, this might have been unintended on the WebKit side.  Let's see what @litherum and @nedley have to say.  That CoreText doesn't specify what points it operates on is very confusing definitely.

torarnv:
> You are implying here, that CTFontCreateWithFontDescriptor takes "typographic points". I was tripped by that before as well. My current understanding is that, that is not the case, and CoreText takes CoreGraphics points. I'm also sure some CoreText engineers have mixed and match the two over time...

If Core Text isn't assuming anything about the DPI, and just passing things on to Core Graphics, then that means the point size argument to  CTFontCreateWithFontDescriptor or any other CT functions is _both_ typographic points and Core Graphics points. 

Which in practice then makes Core Text have a logical DPI of 72, since that's the definition of a typographic point?

behdad:
> > @drott and I arrived at this the following way: looking at Chrome vs Safari's shaping San Francisco font spacing, both through CoreText, we found that if we passed the CSS "point" size (= typographic points = 72dpi) to CoreText, then the spacing will be different from what Safari renders, but if we scale it to CSS "pixel" size (= 96dpi) then the spacing matches.
> 
> Right, this makes sense.

No it doesn't.

> If you pass the CSS font point size directly to Core Text, and shape and render the text, it will be too small since the CSS pixel is 96 DPI.

In the previous sentence, you once used "CSS font point size" and second time used "CSS pixel".  You are distinguishing between "CSS point" and "CSS pixel", right?  One is 72dpi, other is 96dpi.

torarnv:
Let me phrase it another way. What point size does this:

```html
<div style="font-size: 72pt">foo</div>
```

result in giving to:

1. **Shaping** via HarfBuzz:
    1. `hb_font_set_ptem`
        1. CoreText (`CTFontCreateUIFontForLanguage`/`CTFontCreateWithGraphicsFont`)
1. **Drawing** via CoreText
    1. `CTFontCreateUIFontForLanguage`/`CTFontCreateWithGraphicsFont`
    1. Scaling the CoreGraphics matrix

behdad:
> Let me phrase it another way. What point size does this:
> 
> ```
> <div style="font-size: 72pt">foo</div>
> ```

This is, in my understanding, 72 typographic points, 72 CSS points, and 96 CSS pixels.


> result in giving to:
> 
> 1. **Shaping** via HarfBuzz:
>    
>    1. `hb_font_set_ptem`

72.

>       1. CoreText (`CTFontCreateUIFontForLanguage`/`CTFontCreateWithGraphicsFont`)

96.  This is the value that if passed to CoreText, the spacing matches Safari's.  If we pass 72, then spacing will be off.


> 2. **Drawing** via CoreText
>    
>    1. `CTFontCreateUIFontForLanguage`/`CTFontCreateWithGraphicsFont`
>    2. Scaling the CoreGraphics matrix

I don't know what is passed for drawing.  I'm not involved in those parts.

torarnv:
> > 1. **Shaping** via HarfBuzz:
> >    
> >    1. `hb_font_set_ptem`
>  72

Right

> > ```
> >   1. CoreText (`CTFontCreateUIFontForLanguage`/`CTFontCreateWithGraphicsFont`)
> > ```
> 96
> This is the value that if passed to CoreText, the spacing matches Safari's.  If we pass 72, then spacing will be off.

Right.

> > 1. **Drawing** via CoreText
> >    
> >    1. `CTFontCreateUIFontForLanguage`/`CTFontCreateWithGraphicsFont`
> >    2. Scaling the CoreGraphics matrix
> 
> I don't know what is passed for drawing. I'm not involved in those parts.

This is key. To make the resulting text not just have the 96 DPI scaled _metrics_, _positions_, etc, but also render/rasterize visually at 96 DPI, the text needs to be scaled up also during the drawing, either by scaling up the font size, or by applying a scale transform when drawing each glyph.

And in Blink, and WebKit, this is in sync with the scale done in HarfBuzz, so all is good. But in Qt we don't scale during drawing since we're not targeting the web's 96 DPI pixels, so the scaling in HarfBuzz is a mismatch. 

#### Blink

```C++
// Shaping
CTFontCreateWithGraphicsFont(.., 96,..);
CTRunGetPositionsPtr();

// Drawing
CTFontCreateWithGraphicsFont(.., 96,..);
CTFontDrawGlyphs();
```

#### Qt

```C++
// Shaping
CTFontCreateWithGraphicsFont(.., 96,..);
CTRunGetPositionsPtr();

// Drawing
CTFontCreateWithGraphicsFont(.., 72,..);
CTFontDrawGlyphs();
```

torarnv:
Please note, I'm not arguing against Blink ending up with 96pt font size passed on to Core Text during shaping, it needs to do that to match WebKit. But it should do that by scaling up the font size before passing it into HarfBuzz, leaving HarfBuzz DPI-agnostic.

behdad:
We cannot satisfy the Qt and the Webkit requirements simultaneously...

Why can't Qt scale what it passes to `hb_face_set_ptem`?

> Please note, I'm not arguing against Blink ending up with 96pt font size passed on to Core Text during shaping, it needs to do that to match WebKit. But it should do that by scaling up the font size before passing it into HarfBuzz, leaving HarfBuzz DPI-agnostic.

But then the number passed to HarfBuzz would not be typographic points anymore.  Basically you are saying that webkit is doing the wrong thing passing CSS pixel size to CoreText and it should be passing CSS point size to CoreText instead.  I agree that would make some things simpler.  Let's wait to see what @litherum has to say.

torarnv:
> > Please note, I'm not arguing against Blink ending up with 96pt font size passed on to Core Text during shaping, it needs to do that to match WebKit. But it should do that by scaling up the font size before passing it into HarfBuzz, leaving HarfBuzz DPI-agnostic.
> 
> But then the number passed to HarfBuzz would not be typographic points anymore.

Sure it would. 72pt x 1.33 = 96pt is still a typographic point size. It's not the _same_ point size as expressed in the CSS, but that's the whole point (no pun intended). The point size is scaled up to account for the target DPI. This is the approach described by https://developer.apple.com/videos/play/wwdc2015/804/?time=1835

image

> _"Lastly, with optical sizes, we are really challenging some old assumption as to how fonts work, so for example, if you typeset a word at 15 points and later on you apply a scale to your matrix and you reuse the same font object, well, you are getting a font that is designed for 15 points at, I don't know, 120 points. What you really want to be doing there is actually reinstantiate the font again for that specific point size, and you will get the right behavior. So we really want you to treat fonts as opaque objects if possible, so use the system APIs to get to them."_

Besides, scaling 72pt  to 96pt is already what you are doing in HarfBuzz before passing it on to Core Text, which doesn't have any concept of DPI, and hence operates in 72 DPI too, same as HarfBuzz. So doing the scaling there is just as "wrong" if you are arguing that what's written in the CSS should somehow persist pristine through the pipeline all the way.

> We cannot satisfy the Qt and the Webkit requirements simultaneously...

Well, you can, by not embedding assumptions about the target DPI into the library. HarfBuzz is used by many projects, not only limited to Web engines targeting 96 DPI. 

> Why can't Qt scale what it passes to `hb_face_set_ptem`?

To undo the assumption HarfBuzz is making? That would be one workaround if HarfBuzz sticks to the current behavior yes. A workaround that breaks the "no long a typographic point" rule you argued above btw.

> Basically you are saying that webkit is doing the wrong thing passing CSS pixel size to CoreText and it should be passing CSS point size to CoreText instead.

No, that's not what I'm saying.

I'm saying that the observation that Safari/WebKit rendered `font-size: 72pt` with a tracking as if Core Text was dealing with a 96pt font was correct. But that the conclusion that Core Text "operates" in 96 DPI was wrong, and that embedding this assumption into HarfBuzz makes things harder for clients of HarfBuzz that are not browser engines.

I'm saying that Chromium/Blink should account for the observation made about Safari/WebKit by scaling the font point sizes to 96 DPI before passing them into HarfBuzz, which will result in the exact same point size to Core Text in the end, but scaled in a _different place in the pipeline_.

ebraminio:
Whatever conclusion here will be, please apply it to https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-aat-layout-trak-table.hh#L101 also

> a different place in the pipeline.

Makes sense to me at least

@drott

eskilblomfeldt:
Coming in a bit late, so apologies if I am missing something.

Just speaking from the perspective of Qt, where we want to match the look of native components like NSLabel and the TextEdit app, the assumption has always been that Apple's logical DPI is 72. This is the term we use for the DPI that converts point size to pixel size, since there is hardly any output except for printers that actually use typographically correct points anymore. So 1 pt = 1 px on macOS, and this is a definition that dates back to ancient times, when the monitors were actually designed to be 72 dpi and programmers got accustomed to pixel size and point size being the same. They have never diverged from this (except that they have redefined what a pixel is recently to do retina-scaling, overloading the word 'point' for that, which makes it a bit confusing.) In fact, I think it is even the current definition on iOS. Anyway, point size on macOS has not been typographically correct since the screen DPI was actually 72. 

As I understand it, Harfbuzz has made this change to match the behavior of Safari, which operates on 96 logical DPI because of a web standard (which makes sense, since you want web browsers to look the same everywhere). Since Harfbuzz has other use cases than just web browsers, wouldn't it make sense to handle the conversion from the CSS definition to the platform definition in the browser code itself, instead of in the general shaping library? That way, the web-specific code stays with Chrome.

We *could* work around this in Qt by doing the opposite, i.e. preparing the values we send to Harfbuzz, knowing that they will later be scaled to match the CSS definition. 

But without some proper definition in place, it sounds a bit hacky and vulnerable to similar problems later on, if another bug fix in Chrome causes the behavior to change. If this is the case, would it be possible to define the DPI used in Harfbuzz as 96 for all platforms, and convert all input to match this, regardless of the actual logical DPI used on the platform.

torarnv:
> Whatever conclusion here will be, please apply it to https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-aat-layout-trak-table.hh#L101 also

Done. I've updated `aat-trak.tests`, but we might want to update the trak table in `TRAK.ttf` instead perhaps, if the test is testing specific cutoff-points, and interpolation between table entries?

ebraminio:
> Done. I've updated aat-trak.tests, but we might want to update the trak table in TRAK.ttf instead perhaps, if the test is testing specific cutoff-points, and interpolation between table entries?

No need to worry AFAIK. https://github.com/harfbuzz/harfbuzz/blob/master/test/shaping/data/in-house/tests/macos.tests should be rebased also. It is easy, just look at https://circleci.com/gh/harfbuzz/harfbuzz/67205 and https://circleci.com/gh/harfbuzz/harfbuzz/67203 log, but please check if the change make sense visually also. Thanks

src/hb-coretext.cc
```diff
@@ -341,7 +324,7 @@ hb_coretext_font_data_sync (hb_font_t *font)
   const hb_coretext_font_data_t *data = font->data.coretext;
   if (unlikely (!data)) return nullptr;
 
-  if (fabs (CTFontGetSize((CTFontRef) data) - coretext_font_size_from_ptem (font->ptem)) > .5)
+  if (fabs (CTFontGetSize((CTFontRef) data) - font->ptem) > .5)
```
ebraminio: Already inconsistent, but now that you are touching it, please add a space after two `CTFontGetSize` call to make it match with our style. Thanks

torarnv: Done. Also for the call in `hb_coretext_font_create`.

torarnv:
> https://github.com/harfbuzz/harfbuzz/blob/master/test/shaping/data/in-house/tests/macos.tests should be rebased also.

Strange, `tests/macos.tests` passes here on 10.14.2 without any rebase. 

How do I verify the test cases visually? Is there a diff-mode? Thanks!

ebraminio:
> I'm guessing I end up in the 10.13.6 https://gist.github.com/ebraminio/d432e831b3f7ebe30245dde5775e1c7e test path?

That URL is only about font hashes, just useful to add more tests, nothing to worry about :)

torarnv:
> Easy :) Look at the end of https://circleci.com/gh/harfbuzz/harfbuzz/67205 and https://circleci.com/gh/harfbuzz/harfbuzz/67203 it says it expect what to be become what, which I am pasting here also :)

Ah, right, thanks! `.ci/fail.sh` was also useful to get some more verbose test output.

> Wonder why we ended up doing that? Well, interestingly synthesizing AAT fonts revealed to harder to implementing it itself, so we have macOS/bot specific tests, at least. We have however good in-house tests for `trak` and `morx` that's why we are going to deprecate and coretext-aat (not whole hb-coretext, just the part the part that redirects requests to CoreText when a font used to work better with it)

Interesting! So the `CTFontCreateUIFontForLanguage` code-path in `create_ct_font` won't be needed anymore?

ebraminio:
> Interesting! So the CTFontCreateUIFontForLanguage code-path in create_ct_font won't be needed anymore?

Yes. Chrome 72 works like that currently and they've gained a performance boost from it, they've disabled the whole the hb-coretext, even the fact they could just disable coretext_aat using hb_shape_full. We like to suggest Qt to do so also. If you are making a hb_fact_t from CTFont, you will still need hb-coretext but once we remove coretext_aat (#1478), or lower its priority (which effectively make it to not be called normally), we won't ask CoreText about shaping an AAT font.

ebraminio:
And no bugs like these https://www.theverge.com/2015/3/20/8267443/crash-google-chrome-mac-os-x-text and https://www.macobserver.com/tmo/article/apple-core-text-rendering-bug-crashes-apps-in-os-x-and-ios as we can update HarfBuzz just on spotting such issues quickly but one may can't update their OS that quickly so we strongly recommend not using hb-coretext or at least coretext_aat once stabilized, which we can say it is now also as Chrome (not its stable version however) ships it.

torarnv:
> Yes. Chrome 72 works like that currently and they've gained a performance boost from it, they've disabled the whole hb-coretext, even the fact they could just disable coretext_aat using hb_shape_full. We like to suggest Qt to do that also. If you are making a hb_fact_t from CTFont, you will still need hb-coretext but once we remove coretext_aat (#1478), or lower its priority (which effectively make it to not be called normally), we won't ask CoreText about shaping an AAT font.

Cool! Do we lose anything by not using hb-coretext (or coretext_aat), or is it "pixel/metric-perfect" compared to what Core Text would give?

ebraminio:
> Cool! Do we lose anything by not using hb-coretext (or coretext_aat), or is it "pixel/metric-perfect" compared to what Core Text would give?

We can't promise pixel perfect as either with coretext_aat or not, if you are that much interested in need of such thing you can use whole coretext shaper (possible through harfbuzz, just to pass "CoreText" to hb_shape_full) but AFAICS, as I investigated Chrome test needed update on the transition myself and reported issues which was actually there, I can say it is highly compatible and sometimes better and with your change, it will become even better :)

ebraminio:
> hb-coretext know how to create a hb_face_t from CTFont also

So if you are not using that bit, just disable hb-coretext, otherwise a HarfBuzz update will retire coretext_aat anyway.

ebraminio:
> hb-coretext know how to create a hb_face_t from CTFont also

So if you are not using that bit, just disable hb-coretext, otherwise a HarfBuzz update will retire coretext_aat anyway.

behdad:
I definitely suggest Qt build without hb-coretext these days!  As to this bug, clearly I don't mean hb-coretext to be Chrome-specific.  I'm just waiting to get all sides of the picture (ideally from Webkit) and talk to @drott about Chrome side more before accepting the change.  Thanks.

torarnv:
Thanks @behdad! Yeah, moving away from `hb-coretext` definitely sounds like the way to go for us too!

ebraminio:
No update for a week but of course understandable as the new is coming... I think however we should do the right thing anyway, lets merge this Behdad as it is an assumption proven to be not needed for everyone.

torarnv:
No rush guys, happy holidays! 🎉

behdad:
No merge. Nothing here that cannot wait another month.

torarnv:
Happy new year guys! Who else than @drott do we need to get involved to move this forward?

torarnv:
Ping 🛎 @ebraminio @behdad 

Who do we need to sync up with on the Blink team to move this forward?

ebraminio:
@drott @eaenet @kojiishi 

We should however just merge. They will easily notice the difference on a rebase attempt and apply their local workaround.

ebraminio:
Correct and understandable.

> I would strongly suggest not to merge before that

Sure!

torarnv:
> I will meet Behdad next week in a workshop. I would strongly suggest not to merge before that, as it would break easily rolling HarfBuzz into Chrome without code changes in Blink. An approach like: "Let's change API behavior, clients will just have to adjust." is not a sound release process. 

Agreed. That's why we've been trying to sync up with you, so that we can align the merge with a Blink roll 😃 

Presumably you scale the font size to match 96DPI somewhere in Blink before rendering the resulting glyphs using Core Text. You would have to do a similar scaling before passing on to HarfBuzz. The 96 DPI requirement is then fully contained in the client that needs it, Blink.

> From my point of view, the discussion above has not been resolved. We also still want to solicit feedback from Apple on this.

What part of the discussion is unresolved? Apple's take is that the 96 DPI assumption doesn't make sense.

drott:
I sent an email to contacts at Apple to seek clarification for the mapping between https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6trak.html trak table units, typographic pt, and css pixels. 

The original intention for this API was to get tracking right, while at the same time having a sane input unit for the value to be set, for which we decided on ptem.  Perhaps it's better to focus on that rather than having a somewhat sidetracked discussion on what CoreText is assuming or not assuming.

In terms of tracking results: The mapping between CSS pixels, CoreText points, and the trak table size thresholds in Safari is 1:1:1 - which is slightly odd, as we would assume that values in the trak table are typographic units, not CoreText points, so the mapping should rather be 4:4:3.

In Blink, we convert from CSS pixels to ptem (based on 72dpi) and HarfBuzz scales it back to 96px, so that we end up with the same 1:1:1 mapping, resulting in correct tracking.

For optical sizing, Safari converts CSS pixels to typographic points, compare 
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/platform/graphics/cocoa/FontCacheCoreText.cpp#L621 so here the mapping is 4:3 as the opsz thresholds are in typographic points.

drott:
> > In terms of tracking results: The mapping between CSS pixels, CoreText points, and the trak table size thresholds in Safari is 1:1:1 - which is slightly odd, as we would assume that values in the trak table are typographic units, not CoreText points, so the mapping should rather be 4:4:3.
> 
> The Safari behavior makes sense to me. I create a `<div style="font-size: 72pt">foo</div>`, it's converted to 96 CSS pixels using the 96 DPI CSS definition. Once the conversion to CSS pixels is done, it's all 1:1:1 into CoreText. 1 CoreText point == CoreGraphics point. (which, when rasterized on a 2x retina display will be 2 actual display pixels, but we should leave retina considerations out of this).
> 
> The values in the trak table _are_ typographic units. It's just that Safari has _pre scaled_ the typographic point **size** to match the 96DPI CSS pixel definition, as described in https://developer.apple.com/videos/play/wwdc2015/804/?time=1835.

No, they are definitely not used as typographic units. They are mapped 1:1:1 from CSS Pixels, via CoreText points, to trak-table value thresholds:

The trak table for SF NS Display is [here](https://gist.github.com/drott/7a86ce493418fe4d04e7c29db9ac3fd5).

As you can see, the trak values are stable after whatever-unit-size 80.

I encourage you to take a look at the test case we made:
https://cs.chromium.org/chromium/src/third_party/blink/web_tests/fast/text/mac-system-ui-trak.html

Please look at the 3 expected values at the end, these correspond to CSS *px* sizes 80, see line 28 of that test, where we assign a "px" unit to putting those values in the document. So, after 80, the ratios become stable, exactly as in the trak table.

So, we match what Safari does, which is applying CSS "px" units 1:1 to trak.

> > In Blink, we convert from CSS pixels to ptem (based on 72dpi) and HarfBuzz scales it back to 96px, so that we end up with the same 1:1:1 mapping, resulting in correct tracking.
> 
> Those CSS pixels are pre-scaled, so a `<div style="font-size: 72pt">foo</div>` would be 96 CSS pixels I assume?

It's unclear what you mean by pre-scaled. Internally we store CSS pixel values for font size, so if you'd have a CSS pt unit somewhere, that would be converted to internal CSS pixels, yes, using 96dpi.

drott:
> In Blink, we convert from CSS pixels to ptem (based on 72dpi) and HarfBuzz scales it back to 96px, so that we end up with the same 1:1:1 mapping, resulting in correct tracking.

Amending to that: What I meant to say here: The HB interface set_ptem expects typographic points (with a 72dpi assumption), which are then in HarfBuzz internally scaled back to the same value that they were in CSS px in order to keep the 1:1:1 mapping from CSS px to trak table values.

torarnv:
> It's unclear what you mean by pre-scaled. Internally we store CSS pixel values for font size, so if you'd have a CSS pt unit somewhere, that would be converted to internal CSS pixels, yes, using 96dpi.

Okey, so we have:

 `font-size: 72pt` -> converted to css pixels = `96px`

 1. What do you pass into HB's `set_ptem`? 72 or 96? 
 2. What do you pass into CT when creating the font you end up rasterising the glyhps with in the end? 72 or 96?

drott:
> > It's unclear what you mean by pre-scaled. Internally we store CSS pixel values for font size, so if you'd have a CSS pt unit somewhere, that would be converted to internal CSS pixels, yes, using 96dpi.
> 
> Okey, so we have:
> 
> `font-size: 72pt` -> converted to css pixels = `96px`

For the sake of your example, we can go through it - however, in the test we do not use 72pt as font size, but just 72px. 

> 1. What do you pass into HB's `set_ptem`? 72 or 96?

72, see https://cs.chromium.org/chromium/src/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_face.cc?l=417

> 2. What do you pass into CT when creating the font you end up rasterising the glyhps with in the end? 72 or 96?

We pass the internal CSS pixel value when creating the font, for example using `CTFontCreateWithFontDescriptor` or `NSFont fontWithName: size:`. I am currently not sure how screen device pixel ration and zoom factor are handled - they might be added as a matrix to Skia.

torarnv:
> > Okey, so we have:
> > `font-size: 72pt` -> converted to css pixels = `96px`
> 
> > 1. What do you pass into HB's `set_ptem`? 72 or 96?
> 
> 72, see https://cs.chromium.org/chromium/src/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_face.cc?l=417

And this is the problem. You should pass 96.  As you said, the HB interface set_ptem expects typographic points (with a 72dpi assumption). You've converted the CSS points (96dpi definition) into pixels already, and since 1 px == 1pt in 72dpi, you should pass that on directly.

Right now you are scaling 96 to 72 and then relying on the code in HB to scale it back to 96 before passing it on to CT. Tying Blink and HB together like this cause issues for clients of HB that are not working with a CSS DPI of 96 such as Qt.

> > 1. What do you pass into CT when creating the font you end up rasterising the glyhps with in the end? 72 or 96?
> 
> We pass the internal CSS pixel value when creating the font, for example using `CTFontCreateWithFontDescriptor` or `NSFont fontWithName: size:`.

Exactly, so here it's CSS px 1:1 to CT point. This is what the shaping code should do too.

> For the sake of your example, we can go through it - however, in the test we do not use 72pt as font size, but just 72px.

Just to clarify/confirm, this would then result in passing 54 to `set_ptem` right?

torarnv:
Let me try to summarize my understanding of the situation.

Different OSes use different factors when converting font points to pixels. In Qt we refer to this as the _logical DPI of the OS_, as opposed to the _physical DPI_ of the screen (resolution / physical size), but I'm sure there's a better name for it. macOS has traditionally had a logical DPI of 72, while Windows has had a logical DPI of 96. The Web standardized on a logical DPI of 96.

This effectively means that when web content is rendered on macOS, a CSS `12pt` font will be visually larger than its native application `12pt` counterpart. The former is converted to pixels with a 4:3 ratio, the latter 1:1. Visually the former will render as if it was `16pt` in a native application.

Following the advice in [WWDC 2015 - Session 804](https://developer.apple.com/videos/play/wwdc2015/804/?time=1835), one should not do this 4:3 conversion by scaling the matrix of a `12pt` font, one should create a `16pt` font:

> "Lastly, with optical sizes, we are really challenging some old assumption as to how fonts work, so for example, if you typeset a word at 15 points and later on you apply a scale to your matrix and you reuse the same font object, well, you are getting a font that is designed for 15 points at, I don't know, 120 points. What you really want to be doing there is actually reinstantiate the font again for that specific point size, and you will get the right behavior. So we really want you to treat fonts as opaque objects if possible, so use the system APIs to get to them."

Chromium follows this when rendering the resulting glyphs. The glyphs are drawn with a font that's instantiated based on the CSS pixel size. The CSS pixel size is in turn the result of applying the CSS 96 DPI to a CSS point size, so a CSS `12pt` font will be CSS `16px`, resulting in creating Core Text fonts of 16 points.

Chromium also follows this when shaping the text, but via some extra steps. First, the CSS pixel size is converted _back_ to CSS points, in [harfbuzz_face.cc](https://cs.chromium.org/chromium/src/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_face.cc?l=417), then passed on to `hb_font_set_ptem` as if they were typographic points (72DPI), but they are not, they are CSS points (96 DPI). Then HarfBuzz undoes this: https://github.com/harfbuzz/harfbuzz/blob/2a33ab0560a13cb8fef5983bfe0f6d68dfd47b14/src/hb-coretext.cc#L57

So the end, we end up passing a 16 point value to Core Text in HarfBuzz too, as we should. And which makes Chromium match Safari's tracking ratios in the [mac-system-ui-trak.html](https://cs.chromium.org/chromium/src/third_party/blink/web_tests/fast/text/mac-system-ui-trak.html) test. Because the tracking is computed based on the final _visual_ size of the text on macOS: `16pt`.

The conversion in HarfBuzz cancels out the conversion in Blink. This works for Chromium, since it contains both pieces. But is a problem for other clients of the HarfBuzz library, which do not have the first conversion.

By removing the conversion in HarfBuzz (this patch), and the conversion in Blink ([diff](https://gist.github.com/torarnv/6e19e0463e980e022815d589f4b1d893)), Chromium will have the exact same behaviour, passing on `16pt` to Core Text in the end, but in a 1:1 path from Blink, not going via the extra conversions. It also means Qt and other non-Web clients can use HarfBuzz without patches on top that emulate what Blink is doing just to cancel the two operations out.

For the record, I built Chromium with the two patches applied, and the `mac-system-ui-trak.html` test passes both without the two patches, and with both of them (but of course not with just the HB one)<sup>*</sup>.

---

<sup>*</sup> Technically, `[1/1] fast/text/mac-system-ui-trak.html passed unexpectedly`. Why is this test marked as `WontFix` in `NeverFixTests`?

I also tried running all the `fast/text`tests, but get a lot of image diff failures, even without any patches applied. Do I need to configure my system in a particular way to match the `retina-mac` baselines? I tried both with and without font smoothing enabled.

Or do I need to run the test with some image diff fuzz factor? I remember in WebKit days we had a `--tolerance` argument when running the layout tests.

torarnv:
> I will meet Behdad next week in a workshop

Did you get a chance to discuss this?

behdad:
> Did you get a chance to discuss this?

We did.

I agree with your analysis for the most part.  However, changing it now will be a API behavior change I'm not comfortable with.  So have been postponing figuring out what to do about it.

torarnv:
> I agree with your analysis for the most part. 

Good to hear! 🎉

> However, changing it now will be a API behavior change I'm not comfortable with. 

I'm not sure I agree that changing the behavior now is problematic. It's shouldn't be any more problematic than  the original change that introduced the 96 DPI assumption, should it? (which changed behavior for clients such as Qt).

The only client of HarfBuzz realistically affected by this behavior change is Chromium, where we'll make sure to update Blink in unison with the HarfBuzz roll. All other clients are just getting a bug fix, and as a result improved tracking. 

Philosophically any bug fix is a behavior change for those that relied on the buggy behavior isn't it? I don't think that should prevent us from moving forward on this.

ebraminio:
Hey there, this should be rebased again as this new addition https://github.com/harfbuzz/harfbuzz/blob/master/test/shaping/data/in-house/tests/macos.tests#L54 Sorry and thanks

torarnv:
Will do. Is there a plan for how and when to merge this?

ebraminio:
Sometimes I wish I could override Behdad opinion by some other bureaucratic process like voting or something! :)

behdad:
Hahahahaha.  I promise to reply before Monday.

ebraminio:
ping

ebraminio:
ping

ebraminio:
Considering the time has passed since having this now needs more thinking of course. I believe we still can have this considering limited clients rely on it (only Chrome AFAIK, that we can collaborate with), limited fonts that use it (only few macOS fonts having `trak`) and the amount of different that they will get, 72/96, so may marginally still possible to do the right, like what we did in hb_ot_name_t.

torarnv:
I’d like to see this merged. I’m ready to rebase the patch whenever a plan is outlined on how to merge it.

torarnv:
How can we move this forward? Please let me know if there's anything I can do to help.

behdad:
I'm not sure.  Either adding new API.  Or making trak look for 'opsz' tag in font variations and deprecate current `set_ptem` API.

ebraminio:
😢

> Or making trak look for 'opsz' tag in font variations 

Interesting, does CoreText understand that also?

torarnv:
What's the problem with merging this and updating Blink in sync?

ebraminio:
So WebKit has also removed the assumption https://bugs.webkit.org/attachment.cgi?id=368815&action=diff from https://twitter.com/behdadesfahbod/status/1127292851245568000 can we have this also? Not even browsers need it and the Chrome tests relying on this should be just rebased which easily can be done with an update

jjgod:
I agree that we should get rid of this.

behdad:
> So WebKit has also removed the assumption https://bugs.webkit.org/attachment.cgi?id=368815&action=diff from https://twitter.com/behdadesfahbod/status/1127292851245568000 can we have this also? Not even browsers need it and the Chrome tests relying on this should be just rebased which easily can be done with an update

Is it just me or that WebKit commit just makes things more confusing?  To quote: "It turns out that it means "typographic point size" which is equal to CSS pixels, not CSS points."

Where does *that* come from?  CSS pixels are definitely defined as 96dpi, which is clearly not what a typographic point size is.

It's most frustrating that @litherum refuses to leave a clear reply here.  I'll try to bring it up with him next week.

torarnv:
> Where does that come from? CSS pixels are definitely defined as 96dpi, which is clearly not what a typographic point size is.

I may be wrong, but I think this comes down to `px` being the [canonical unit](https://www.w3.org/TR/css3-values/#canonical-unit) of the absolute lengths, the recommendation of the spec that the pixel unit is anchored to the [reference pixel ](https://www.w3.org/TR/css3-values/#reference-pixel) on screen media, and the recommendation that _"the pixel unit refer to the whole number of device pixels that best approximates the reference pixel."_.

On a non-retina display on macOS, that makes 1 CSS pixel equal to 1 device pixel, since that's the best approximation of the reference pixel. And with the logical DPI of macOS being 72, that means a CSS pixel effectively equals one typographic point.

The CSS then defines one inch as 96 CSS pixels, so yes the "dots per inch" of a CSS pixel is 96, but that's anchored in the CSS pixel, not the other way around. It would be more accurate to say that the CSS _inch_ is 96 DPI.

davelab6:
> On a non-retina display on macOS, that makes 1 CSS pixel equal to 1 device pixel, since that's the best approximation of the reference pixel. And with the logical DPI of macOS being 72, that means a CSS pixel effectively equals one typographic point.
>
> The CSS then defines one inch as 96 CSS pixels, so yes the "dots per inch" of a CSS pixel is 96, but that's anchored in the CSS pixel, not the other way around. It would be more accurate to say that the CSS inch is 96 DPI.

This seems contradictory to me. 

72 typographic points is 1 inch.

72 Logical DPI in macOS.

1 CSS pixel equals 1 pre-retina device pixel on Apple computers.

Therefore, 72 CSS pixels = 1 inch. 

But then... If CSS defines one inch as 96 CSS pixels, that doesn't make sense.

torarnv:
I'm at a loss here. There seems to be agreement that the patch is correct, both from the Chromium/Blink side and from the WebKit side. What's keeping this patch in limbo?

behdad:
I talked about this to Ned and Myles.  It's basically everyone shrugging it off to others.  There's clearly discrepancies.  There's clearly no will for anyone to own it...

@drott can we remove the multiplication in HarfBuzz and coordinate a compensating change in Chrome?

drott:
> @drott can we remove the multiplication in HarfBuzz and coordinate a compensating change in Chrome?

Ok. Please ping me on the commit, and I can remove the calculation on the Chrome side when rolling.

ebraminio:
Great that we can finally have this to serve brilliant Qt community, @torarnv please recheck if this is doing what is expected. Thanks

torarnv:
Thanks guys, happy to see this finally landed! 🎉

drott:
Chrome side change landed https://chromium-review.googlesource.com/c/chromium/src/+/1761282

behdad:
Thanks Ebrahim.  I was going to wait till 2.7 release.  Maybe I do name next release 2.7 even though there's nothing else (other than one bugfix).  Any preference between 2.6.1 vs 2.7.0?

behdad:
> I talked about this to Ned and Myles. It's basically everyone shrugging it off to others. There's clearly discrepancies. There's clearly no will for anyone to own it...

Just to clarify:

* Ned's position is that CoreText points are considered by CoreText as being typographic points for the purposes of optical-sizing.

* Myles's position is that for the same font size set in Safari & Firefox & Chrome, vs native Mac apps, the browsers render things 33% larger, suggesting that a 96/72. conversion is necessary.  The fact that Firefox & Chrome have simply matched Safari and as such it's Safari that invented this discrepancy does not change much.

Anyway.  Happy to say not our problem anymore. ¯\_(ヅ)_/¯

torarnv:
> [...] for the same font size set in Safari & Firefox & Chrome, vs native Mac apps, the browsers render things 33% larger, suggesting that a 96/72. conversion is necessary.

Agreed, and now that conversion is contained within the web engine — not bleeding into an underlying framework used by many other clients 👍🏻

> Anyway. Happy to say not our problem anymore. ¯\\_(ヅ)_/¯

😂

giphy-downsized

Thanks again guys! 🎉

PR #1486, opened 2018-12-17
Remove redundant void from C++ sources

ebraminio:
(as the discussions with Behdad)


ebraminio:
Just to note asan bot fail is not related to this change but `master` also has it.

PR #1494, opened 2018-12-17
Solaris crash: reference counting bug leads to use after free

### LINKED ISSUE #1493: harfbuzz 1.8.5 and later crashes on Solaris
In hb-atomic.hh - when it was hb-private-atomic.hh,
the Solaris section had this macro

#define hb_atomic_int_impl_add(AI, V)           ( ({__machine_rw_barrier ();}), atomic_add_int_nv (&(AI), (V)) - (V))

Today that looks like this :-
---
static inline int _hb_fetch_and_add (int *AI, int V)
{
  _hb_memory_w_barrier ();
  int result = atomic_add_int_nv ((uint_t *) AI, V);
  _hb_memory_r_barrier ();
  return result;
}
#define hb_atomic_int_impl_add(AI, V)           _hb_fetch_and_add ((AI), (V))
---

There's a critical bug in here.
The old macro subtracted the "V" addend so that the macro returned the PREVIOUS value.
The new code returns the NEW value.

Looks to me like the equivalent for the Linux returns the old value and the
subtraction in the old Solaris macro was to mimic that but it was overlooked in the new
code.

It took a while to dawn on me that the calling code looking for "1" to decide that this was
the signal to do a free was because it was expecting the value before decrementing.

So the current code leads to a crash doing almost anything since this is used in
reference counting of use of blobs and we get what looks like a use after free
crash ... pull request for the simple fix coming soon.

prrace:
Fix #1493

### DIFF
--- a/src/hb-atomic.hh
+++ b/src/hb-atomic.hh
@@ -143,7 +143,7 @@ static_assert ((sizeof (LONG) == sizeof (int)), "");
 static inline int _hb_fetch_and_add (int *AI, int V)
 {
   _hb_memory_w_barrier ();
-  int result = atomic_add_int_nv ((uint_t *) AI, V);
+  int result = atomic_add_int_nv ((uint_t *) AI, V) -V;
   _hb_memory_r_barrier ();
   return result;
 }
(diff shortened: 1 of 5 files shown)


src/hb-atomic.hh
```diff
@@ -143,7 +143,7 @@ static_assert ((sizeof (LONG) == sizeof (int)), "");
 static inline int _hb_fetch_and_add (int *AI, int V)
 {
   _hb_memory_w_barrier ();
-  int result = atomic_add_int_nv ((uint_t *) AI, V);
+  int result = atomic_add_int_nv ((uint_t *) AI, V) -V;
```
behdad: Please `- V` instead of `-V`.

behdad:
Please create a branch from master just with the one commit.  This currently includes your other two fixes in it.

Issue #1503, opened 2018-12-21
Logic or public API to check if a font supports adding Kashida


ebraminio:
Related to #586, even the fact we are not that clear on how Kashida API can or should happen, either by adding Kashida to input text, before shaping (which loses the chance to add alternative glyphs available to Kashida a font can support, like a half Kashida, provided by `JSTF`), or by adding the glyph after the shaping (we know DirectWrite [can do that](https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-directwrite.cc#L686-L766) after the shaping but using the original information which was available before the shaping), one things certainly we need anyway I guess is a logic to check if a font supports adding Kashida.

Simply the issue is about detecting this case, regular Arabic fonts, used in displaying, can have Kashida everywhere the script is connected, but this is not the case with stylistic ones.

For example IranNastaliq v1 downloadable from [1](https://www.yjc.ir/fa/news/4555441/%D8%A2%D8%AE%D8%B1%D9%8A%D9%86-%D9%86%D8%B3%D8%AE%D9%87-%D9%81%D9%88%D9%86%D8%AA-%D9%86%D8%B3%D8%AA%D8%B9%D9%84%D9%8A%D9%82-%D8%AF%D8%A7%D9%86%D9%84%D9%88%D8%AF) > [2](https://cdn.yjc.ir/files/fa/news/1392/6/25/1538674_809.rar), this is the result with "متن مـتن متـن",

![image](https://user-images.githubusercontent.com/833473/50335098-841b0880-051f-11e9-8adf-082ffa24647b.png)

Which the two left ones, having Kashida, are broken.

IranNastaliq [v2](https://cdn.yjc.ir/files/fa/news/1392/6/25/1538677_674.rar) however can do this better as the point of update was to officially supports Kashida

![image](https://user-images.githubusercontent.com/833473/50335242-e83dcc80-051f-11e9-9df6-d102b606f5d8.png)

(not the best example, it works on different places, but I don't know how to detect that)

However for the sake of simplicity for such API (don't know if it should hb_face_t aware or hb_buffer_t) I guess the version v2 also should be marked as unsafe for adding Kashida.

Or, we should just somehow consider a font can support adding Kashida on the context, <s>then the final API feels like needs to be something like `hb_shape_justification_width` (like our hb_directwrite_shape_experimental_width) considering everything if there is usable entry for such elongation in GSUB while shaping I guess.</s> (not necessarily now I think) 

[...]

PR #1505, opened 2018-12-21
[WIP] Add APIs to get scripts properties

### LINKED ISSUE #1500: API to get justification character of a script, or, an API for querying info about hb_script_t?
As #586 and Ben comment [here](https://github.com/harfbuzz/harfbuzz/issues/1487#issuecomment-447977498), we like to provide primitives needed for justification, and I guess the best place to learn what APIs can we provide is our [hb-directwrite](https://github.com/harfbuzz/harfbuzz/blob/244a8627f7b0ed651d6010c7ee7e54fa8fcca400/src/hb-directwrite.cc#L688-L765) implementation which uses that DirectWrite primitives in-action...

So one bit that I was thinking about, which is a good start point also, and needed either we want to do all of the justification by our own or letting client to handle it, is an API equal to [GetScriptProperties](https://github.com/harfbuzz/harfbuzz/blob/244a8627f7b0ed651d6010c7ee7e54fa8fcca400/src/hb-directwrite.cc#L706-L709). The call provides different things about script by [DWRITE_SCRIPT_PROPERTIES](https://docs.microsoft.com/en-us/windows/desktop/api/dwrite_1/ns-dwrite_1-dwrite_script_properties) which to me its only justificationCharacter field is useful for justification.

I wanted to just add that thing in an API maybe named `hb_script_get_justifciation_character` (not necessarily public and may be hold till we get other needed things anyway) but thought if we are going for having that, one may ask what about its other infos that dwrite call can provide, don't we need those? This is a sample of info it provides,

```
i	isoScriptCode	clusterLookahead	isBlockWriting	isConnectedWriting	isCursiveWriting	isDiscreteWriting	isDistributedWithinCluster	isoScriptNumber	justificationCharacter	reserved	restrictCaretToClusters	usesWordDividers
0	Zzzz	15	0	0	0	0	0	999	32	0	0	0
1	Zyyy	1	0	0	0	1	0	998	32	0	0	1
2	Zinh	15	0	0	0	0	0	994	32	0	1	0
3	Arab	8	0	1	1	0	0	160	1600	0	0	1
4	Armn	1	0	0	0	1	0	230	32	0	0	1
5	Avst	8	0	0	0	1	0	134	32	0	0	1
6	Bali	15	0	0	0	1	0	360	32	0	1	0
(...)
```

What are interesting columns we can have or provide, if any? As said, the interesting part to me was `justificationCharacter`, which is 1600 (Kashida) for `Arab`, `Mand`, `Syrc`, `Mani`, `Phlp`, 5760 for `Ogam`, and 32 (space) for others (which means that script doesn't have a Kashida-like character concept).

As said this may doesn't worth to have an API even, however I like to ask is Kashida of a script can be extract from unicode files directly? Is there any other library more suitable to provide such info?

Maybe better to have something like `hb-justification.h`, instead thinking about hb_script_get_* for this, and provide things like these there, populate it with an algorithm to detect good places for inserting Kashida?

ebraminio:
Fixes #1500 and #64

### DIFF
--- a/docs/harfbuzz-sections.txt
+++ b/docs/harfbuzz-sections.txt
@@ -121,6 +121,8 @@ hb_script_from_iso15924_tag
 hb_script_from_string
 hb_script_to_iso15924_tag
 hb_script_get_horizontal_direction
+hb_script_get_justification_character
+hb_script_get_properties
 hb_language_from_string
 hb_language_to_string
 hb_language_get_default
@@ -133,6 +135,7 @@ hb_mask_t
 hb_position_t
 hb_tag_t
 hb_script_t
+hb_script_properties_flags_t
 hb_user_data_key_t
 hb_var_int_t
 HB_TAG
--- a/src/hb-common.cc
+++ b/src/hb-common.cc
@@ -590,6 +590,212 @@ hb_script_get_horizontal_direction (hb_script_t script)
 }
 
 
+/**
+ * hb_script_get_justification_character:
+ * @script:
+ *
+ *
+ *
+ * Return value:
+ *
+ * Since: REPLACEME
+ **/
+hb_codepoint_t
+hb_script_get_justification_character (hb_script_t script)
+{
+  switch (script)
+  {
+  case HB_SCRIPT_ARABIC:
+  case HB_SCRIPT_MANDAIC:
+  case HB_SCRIPT_SYRIAC:
+  case HB_SCRIPT_MANICHAEAN:
+  case HB_SCRIPT_PSALTER_PAHLAVI:
+    return 0x0640; /* Tatweel */
+  case HB_SCRIPT_OGHAM:
+    return 0x1680; /* Ogham Space Mark */
+  default:
+    return 0x0020; /* Space */
+  }
+}
+
+struct hb_script_properties_record {
+  hb_tag_t	tag;
+  unsigned int	flags;
+};
+
+/* Meaning of the bits are in #hb_script_properties_flags_t
+ *
+ * Extracted with https://gist.github.com/ebraminio/727689919dbc485e023fe054bdd7c19b
+ * Sort the result (needs a little) after an update, please
+ */
+static hb_script_properties_record script_properties[] =
+{
+  {HB_TAG ('A','d','l','m'), 0x06},
+  {HB_TAG ('A','g','h','b'), 0x06},
+  {HB_TAG ('A','h','o','m'), 0x06},
+  {HB_TAG ('A','r','a','b'), 0x62},
+  {HB_TAG ('A','r','m','i'), 0x06},
+  {HB_TAG ('A','r','m','n'), 0x06},
+  {HB_TAG ('A','v','s','t'), 0x06},
+  {HB_TAG ('B','a','l','i'), 0x05},
+  {HB_TAG ('B','a','m','u'), 0x07},
+  {HB_TAG ('B','a','s','s'), 0x06},
+  {HB_TAG ('B','a','t','k'), 0x06},
+  {HB_TAG ('B','e','n','g'), 0x23},
+  {HB_TAG ('B','h','k','s'), 0x06},
+  {HB_TAG ('B','o','p','o'), 0x0C},
+  {HB_TAG ('B','r','a','h'), 0x06},
+  {HB_TAG ('B','r','a','i'), 0x06},
+  {HB_TAG ('B','u','g','i'), 0x06},
+  {HB_TAG ('B','u','h','d'), 0x04},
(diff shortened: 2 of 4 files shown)


ebraminio:
Sure, how about separating get_justification_character into a separate PR, does that any chance separately?

behdad:
> Sure, how about separating get_justification_character into a separate PR, does that have any chance separately?

Not without wider discussion of how such API is supposed to be used.

khaledhosny:
The `get_justification_character` does not seem that useful as the application really needs to know more about the script to be able to use it. For example, LibreOffice uses kashida justification in Arabic only even though there are other scripts that can have kashida because it does not know where to insert kashida in these scripts.

ebraminio:
It is equivalent to https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-directwrite.cc#L709 LO can add that to Syriac also as harfbuzz-old was doing so https://github.com/harfbuzz/harfbuzz-old/blob/master/src/harfbuzz-arabic.c#L51 we can ask about all these from Microsoft also maybe.

khaledhosny:
I think this really needs a high level design for how this is going to be used before adding any API. All the issues being opened about this really confuses me and I’m a bit lost. Doing proper Arabic justification is still really an open question and a bit of research needs to be don here.

What Microsoft does it not really the best model and it has many known limitations (it does not distinguish between different style of Arabic, kashida is inserted after the text is shaped and no reshaping is done, fonts has zero control over it).

If you really want to work on this, then may be start with sketching the sort of font control that needs to be provided, fix `JSTF` table, and write a demo application that uses the APIs you want to added, so that you can actually test who useful it is.

ebraminio:
Sure, makes sense. You know, I tried to find the most noncontroversial
parts, different point of view obviously :)

jfkthame:
FWIW, I agree with @khaledhosny that a lot of background work is needed here, before thinking about adding API. Just adding something like `get_justification_character` isn't really useful, and reflects an inadequate model of how justification can/should be done.

ebraminio:
OK, now I have the time to defend myself a bit here :)

> I don't want to copy DirectWrite API directly. Definitely not without discussing each point first. 

You said [here](https://github.com/harfbuzz/harfbuzz/issues/1500#issuecomment-4488498480) it can be useful, and I didn't mean to add whole thing straightly otherwise I was adding directly to the repo, so why the complain when reducing this is easier than adding more things...

> I think this really needs a high level design for how this is going to be used before adding any API.

Now that it has taken ten years, I can imagine it can take another ten years :) And no doubt that is not that important thing and that is also why it is not there yet.

> All the issues being opened about this really confuses me and I’m a bit lost.

Each is about different aspect of things, please focus more :P

> Doing proper Arabic justification is still really an open question and a bit of research needs to be don here.

Right. There is also proposal of using font variation for doing flexible Kashida elongation. I more think about something useful for today but well this patch in particular is not related

> What Microsoft does it not really the best model and it has many known limitations 

I wasn't talking about Microsoft API but how they are applying Kashida justification to non-Arabic scripts, something you said you don't know on "because it does not know where to insert kashida in these scripts" and we can ask them.

> it does not distinguish between different style of Arabic

At least it understands when to forget about Kashida, which is something I am discussing in #1503

> kashida is inserted after the text is shaped and no reshaping is done, fonts has zero control over it

Just speculating here, GetJustifiedGlyphs modifies all the things, and uses information generated by GetJustificationOpportunities which uses textString, so couldn't it theoretically reshape also?

> If you really want to work on this, then may be start with sketching the sort of font control that needs to be provided, fix JSTF table.

What is the issue? You said also it is not in that use so I didn't focus it.

> and write a demo application that uses the APIs you want to added, so that you can actually test who useful it is.

Right.

> Just adding something like get_justification_character isn't really useful, 

This PR was not only about that API of course but to defending it, for handling fallback justification anyway in fonts without `JSTF` or `just` (or futuristic plans like [using font variation](https://twitter.com/sahafshar/status/896389627065577472)), I can't imagine a case without need of this bit of information, Kashida justification, if not done while shaping, should be done either before or after the shaping, and both need such thing, so I wonder.

ebraminio:
No interest apparently so let's close and think about it later. Thanks!

behdad:
Yeah we'll add this properly when needed. The current PR doesn't make sense.

ebraminio:
> Yeah we'll add this properly when needed. 

Guess this is most demanded one, https://github.com/flutter/engine/blob/728e473f1a4bce3f4a034474d81cc95b691c6cf0/third_party/txt/src/minikin/Layout.cpp#L398-L406

behdad:
Right. I noted that in https://github.com/harfbuzz/harfbuzz/issues/64

PR #1506, opened 2018-12-21
[WIP] Some Kashida related draft

ebraminio:
It needs just lot of works but I like to share the change somewhere at least, so


ebraminio:
no much value on this one, let's close

Issue #1496, opened 2018-12-18
[CFF] ByteStr issues...


behdad:
This was brought to my attention thanks to ubsan on my local build:
```
../../../test/fuzzing/../subset/data/fonts/SourceHanSans-Regular.otf
../../src/hb-ot-cff-common.hh:152:6: runtime error: member call on misaligned address 0x0000022b17e1 for type 'CFF::ByteStr', which requires 8 byte alignment
0x0000022b17e1: note: pointer points here
 02 24 02  4c 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00
              ^ 
running subset fuzzer against ../../../test/fuzzing/../subset/data/fonts/SourceSansPro-Regular.otf
../../../test/fuzzing/../subset/data/fonts/SourceSansPro-Regular.otf
../../src/hb-ot-cff-common.hh:152:6: runtime error: member call on misaligned address 0x000002b4d363 for type 'CFF::ByteStr', which requires 8 byte alignment
0x000002b4d363: note: pointer points here
 02  5d 02 63 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00
              ^ 
running subset fuzzer against ../../../test/fuzzing/../subset/data/fonts/Mplus1p-Regular.ttf
../../../test/fuzzing/../subset/data/fonts/Mplus1p-Regular.ttf
```

That points to this piece of code:
```
      /* serialize data */ 
      for (unsigned int i = 0; i < byteArray.len; i++) 
      { 
↦       ByteStr  *dest = c->start_embed<ByteStr> (); 
↦       if (unlikely (dest == nullptr || 
↦       ↦             !dest->serialize (c, byteArray[i]))) 
↦         return_trace (false); 
      } 
```
which suggests that `ByteStr` is a font struct.  However, the implementation is NOT.  It has a pointer, constructors, etc.  I don't understand how you can `start_embed` that.  That's just plain wrong.

The convention is that `CamelCase` are font structs, whereas the rest are `hb_name_spaced_t`.

The ByteStr looks awfully like `hb_bytes_t`.  I'm not sure what it's used for.  Can you please explain your needs, so we can figure out what's the best design?

blueshade7:
[...]

PR #1507, opened 2018-12-22
CFF renaming

blueshade7:
Fixes issue #1496.

Reimplemented struct `ByteStr` in CFF code as `byte_str_t` based on `hb_ubytes_t`.

Uncamelized all non-table type names in CFF namespace, e.g., `CFF1TopDict_OpSerializer` to `cff1_top_dict_op_serializer_t`.

### DIFF
--- a/src/hb-cff-interp-common.hh
+++ b/src/hb-cff-interp-common.hh
@@ -522,18 +522,18 @@ struct ArgStack : Stack<ARG, 513>
     return (unsigned)i;
   }
 
-  void push_longint_from_substr (byte_str_ref_t& substr)
+  void push_longint_from_substr (byte_str_ref_t& str_ref)
   {
-    push_int ((substr[0] << 24) | (substr[1] << 16) | (substr[2] << 8) | (substr[3]));
-    substr.inc (4);
+    push_int ((str_ref[0] << 24) | (str_ref[1] << 16) | (str_ref[2] << 8) | (str_ref[3]));
+    str_ref.inc (4);
   }
 
-  bool push_fixed_from_substr (byte_str_ref_t& substr)
+  bool push_fixed_from_substr (byte_str_ref_t& str_ref)
   {
-    if (unlikely (!substr.avail (4)))
+    if (unlikely (!str_ref.avail (4)))
       return false;
-    push_fixed ((int32_t)*(const HBUINT32*)&substr[0]);
-    substr.inc (4);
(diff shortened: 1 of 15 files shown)


ebraminio:
Now it rhymes better with the rest of the project, thanks :) Wanted to suggest appending `hb_cff_` to ones that are in headers, but that feels being too cumbersome now I think.

src/hb-subset-cff-common.hh
```diff
@@ -473,13 +473,13 @@ struct ParsedCStrs : hb_vector_t<ParsedCStr>
...
 
-struct SubrSubsetParam
+struct subr_subset_param_t
 {
-  void init (ParsedCStr *parsed_charstring_,
-		    ParsedCStrs *parsed_global_subrs_, ParsedCStrs *parsed_local_subrs_,
+  void init (parsed_cs_str_t *parsed_charstring_,
+		    parsed_cs_str_vec_t *parsed_global_subrs_, parsed_cs_str_vec_t *parsed_local_subrs_,
 		    hb_set_t *global_closure_, hb_set_t *local_closure_,
 		    bool drop_hints_)
```
ebraminio: Spacing issue here, my fault most likely.

behdad:
> Wanted to suggest prepending `hb_cff_` to ones that are included and are in headers

As long as they are in namespace CFF::, it's fine.

src/hb-cff-interp-common.hh
```diff
@@ -290,125 +290,106 @@ struct UnsizedByteStr : UnsizedArrayOf <HBUINT8>
...
-  ByteStr ()
-    : str (&Null(UnsizedByteStr)), len (0) {}
-  ByteStr (const UnsizedByteStr& s, unsigned int l)
-    : str (&s), len (l) {}
-  ByteStr (const char *s, unsigned int l=0)
-    : str ((const UnsizedByteStr *)s), len (l) {}
+  byte_str_t ()
+    : hb_ubytes_t () {}
+  byte_str_t (const UnsizedByteStr& s, unsigned int l)
+    : hb_ubytes_t ((const unsigned char*)&s, l) {}
```
behdad: This constructor is supposed to create a sub-bytes?  It doesn't check bounds.

blueshade7: It is a constructor from an unsized byte string `s` and its length `l`. Only we can do is to trust the length is correct.
Hopefully, it is pre-checked when extracted from `CFFIndex`.

Issue #1509, opened 2018-12-23
C++11: Modernize codebase


behdad:
I'm designing the iterator structs that I need badly for subsetting GSUB/GPOS, and I'm hitting walls.  I really need `decltype` or `auto`, and not just for convenience.  I suggest we switch to C++11 now, which means older compiler support ends at the end of current year.

Also, suggest we make all our headers declare `namespace HB`.  And we remove a lot of `hb_` prefixes from internal functions.

Since C++11 will be required, we can remove a lot of hacks.  It was becoming really taxing to support older compilers, specially the ambiguous-operator ones.

behdad:
(Re iterators, I'm reinventing ranges it seems..)

behdad:
Ebrahim and I also discussed elsewhere, about switching from `unsigned int` to `unsigned`.  I started doing that in hb-iter...  That, and dropping redundant `inline` and `void`, and using `auto`, make code a lot more compact and pleasant.  So I suggest we go ahead with `unsigned`.  Am undecided whether I like to use `signed` instead of `int` though.

lemzwerg:
I vote against C++11.  Projects like TeXLive, which are intended to run on a very broad range of operating systems, have already nightmares because of poppler (not only because it uses C++11 but also because of ABI incompatibilities).  Luatex, for example, is in the process of abandoning poppler for exactly these reasons.  It would be a pity if HarfBuzz would cause similar issues.

I'll write an e-mail to the TeXLive list, pointing to this tracker issue.  Maybe other people will chime in.

behdad:
> I'll write an e-mail to the TeXLive list, pointing to this tracker issue. Maybe other people will chime in.

Thanks.

Not requiring C++11 has become a major headache, and with the iterators I need to design, pretty much impossible to continue.  I doubt C++11 itself is a problem.  Let's see what TeXLive list thinks.

It's not like we don't need to support very weird systems.  OpenJDK needs to compile HarfBuzz after all.  @prrace

behdad:
For the record, what I'm proposing will not introduce any ABI incompatibility issues.  We still won't be using libstdc++, and won't expose any C++ API / ABI.

ebraminio:
[...]

PR #1510, opened 2018-12-24
[DONTMERGE] Add operator= on IntType

behdad:
Before C++11 types with custom operator= were not allowed in unions.
I cannot replicate this error by passing -std=c++03 to gcc.
I know if I add a constructor, some of the bots will be unhappy.
Testing for operator= to see if this passes through.  I know way
back when, I removed operator= and added set() method because
someone complained.  But that was ten years ago I think.

Okay.  I see Oracle Studio compiler is failing:
https://circleci.com/gh/harfbuzz/harfbuzz/69225
Do we have to support that, or is Java switching to a newer version of that compiler that understands C++11? @prrace

### DIFF
--- a/src/hb-open-type.hh
+++ b/src/hb-open-type.hh
@@ -64,6 +64,8 @@ struct IntType
   typedef typename hb_signedness_int<hb_is_signed<Type>::value>::value wide_type;
 
   void set (wide_type i) { v.set (i); }
+  template <typename T>
+  IntType<Type, Size>& operator = (const T& v) { set (v); return *this; }
   operator wide_type () const { return v; }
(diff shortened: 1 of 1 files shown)


ebraminio:
Please rebase this as I think now it passes in ODS also as I added -std=c++11 to cmake script used to run ODS.

behdad:
> Please rebase this as I think now it passes in ODS also as I added -std=c++11 to cmake script used to run ODS.

Right.  But https://github.com/harfbuzz/harfbuzz/issues/1509#issuecomment-450581453 is not very encouraging.  Something for future.  Leave this open till I add a commented-out line in the source for it.

ebraminio:
What is the blocking part for us on their comment?

PR #1513, opened 2018-12-24
Add color channels getters ABI

ebraminio:
So can be used with language wrappers

What I understood while reading `hb_glyph_info_get_glyph_flags`... just wondered if is needed here also

### DIFF
--- a/src/hb-common.cc
+++ b/src/hb-common.cc
@@ -1121,6 +1121,59 @@ hb_variation_to_string (hb_variation_t *variation,
   buf[len] = '\0';
 }
 
+/**
+ * hb_color_get_alpha:
+ *
+ *
+ *
+ * Since: REPLACEME
+ */
+uint8_t
+(hb_color_get_alpha) (hb_color_t color)
+{
+  return hb_color_get_alpha (color);
+}
+
+/**
+ * hb_color_get_red:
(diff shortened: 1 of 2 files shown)


ebraminio:
Hmm. Having this one may ask why may ask why not having something like `hb_color_create` instead the macro and then removing the macro and naming it `hb_color_create` also.

behdad:
> Hmm. Having this one may ask why may ask why not having something like `hb_color_create` instead the macro and then removing the macro and naming it `hb_color_create` also.

Because there's nothing being created here.

Also, because there's nowhere in the API that takes a color.  Also, this just doesn't matter.  Spend your time on things that do.

ebraminio:
Right

behdad:
Should we merge?  I can go either way.  Either merge or close.

behdad:
Thanks.  Can you please add documemtation as well?  Let's make that rule for all new API.  Also needs listing in docs/harfbuzz-sections.txt

ebraminio:
> Can you please add documemtation as well?

It was there before, don't know why didn't add it back then.

> Let's make that rule for all new API.

Sure, I did that in other APIs.

> Also needs listing in docs/harfbuzz-sections.txt

Already there.

PR #1522, opened 2018-12-30
CircleCI: Test for Python 3 syntax errors and undefined names

cclauss:
Catch missing imports and errors like #1520 and #1521

__E901,E999,F821,F822,F823__ are the "_showstopper_" [flake8](http://flake8.pycqa.org) issues that can halt the runtime with a SyntaxError, NameError, etc. Most other flake8 issues are merely "style violations" -- useful for readability but they do not effect runtime safety.
* F821: undefined name `name`
* F822: undefined name `name` in `__all__`
* F823: local variable name referenced before assignment
* E901: SyntaxError or IndentationError
* E999: SyntaxError -- failed to compile a file into an Abstract Syntax Tree

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -87,7 +87,8 @@ jobs:
     steps:
       - checkout
       - run: pacman --noconfirm -Syu freetype2 cairo icu gettext gobject-introspection gcc gcc-libs glib2 graphite pkg-config ragel python python-pip
-      - run: pip install fonttools
+      - run: pip install flake8 fonttools
+      - run: flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
       # C??FLAGS are not needed for a regular build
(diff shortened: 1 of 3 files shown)


ebraminio:
Merged your change and will try to see what is happening

ebraminio:
The script uses globals in order to have more clarity. What do you suggest for it, removing such use, or, disabling F821 for that specific file?

ebraminio:
This can be merged if #1524 goes in.

behdad:
Our Python scripts are just fine without CI checking...  They need to be run manually once a year...

ebraminio:
We can merge this with our ArchLinux defined on https://github.com/harfbuzz/harfbuzz/blob/master/.circleci/config.yml bot instead introducing a new not.

cclauss:
That sounds fine to me.

cclauss:
@ebraminio Moved to CircleCI Arch Linux

ebraminio:
Thanks for the great work :) I agree also having a CI check for that is somehow over engineering but as it is done inside another bot and is very quick thing to check and we don't expect to have such issues in future anyway I feel it like a nice check with a very little harm also for the project so merged it.

PR #1524, opened 2018-12-30
Don't use globals in gen-use-table.py to make it lint compatible

ebraminio:


cclauss:
The other technique which is probably more readable is to just write:
```python
def my_func():
    global Consonant, Number
    ...
```

behdad:
I don't care what lint says on this script, and I'm not sure if the change is an improvement or a distraction honestly.  I'm fine if you want to merge this time, but I reserve the right to outright say no next time you make a change just to appease a tool.

ebraminio:
Just don't say it angrily! It is just a proposed fix, don't see it like an already merge change. The lint is useful also to find global variables used to be in python2 but not 3. We may instead migrate all the scirpts to python 3 (changing shebangs I guess?) so everybody understands we are supporting py3, or not.

behdad:
I'm not opposed to porting scripts to python3.  By all means do.

PR #1525, opened 2018-12-30
Declare globals in gen-use-table.py

cclauss:
One proposed alternative to #1524


khaledhosny:
I suggest to just disable that check, I don’t see what is wrong with the existing code. It is clearly a false positive.

cclauss:
Not a false positive.  It is just enforcing the Zen of Python — Explicit is better than implicit.  In a compiled language, the compiler would force these declarations.

PR #1552, opened 2019-01-22
Grammar fix

alfiedotwtf:


docs/usermanual-clusters.xml
```diff
@@ -683,8 +683,8 @@
       <title>Other considerations in level 2</title>
       <para>
 	There may be other problems encountered with ligatures under
-	level 2, such as if the direction of the text is forced to
-	opposite of its natural direction (for example, Arabic text
```
khaledhosny: It should be `forced to the opposite of`, `forced opposite of` is not clear to me.

alfiedotwtf: No problem. Fixed in the following commit :+1:

Issue #1540, opened 2019-01-15
TARGET_OS_OSX is not defined on older OSX versions...


kencu:
So this beautiful workaround in `hb-coretext.cc` fails to function:
```
#if TARGET_OS_OSX && MAC_OS_X_VERSION_MIN_REQUIRED < 1060
      atsFont = CTFontGetPlatformFont (new_ct_font, NULL);
      status = ATSFontGetFileReference (atsFont, &fsref);
      if (status == noErr)
        new_url = CFURLCreateFromFSRef (NULL, &fsref);
#else
      new_url = (CFURLRef) CTFontCopyAttribute (new_ct_font, kCTFontURLAttribute);
#endif
```
Once that is fixed, the software builds perfectly on < 10.6, and thank you for that effort.

Would it be OK to just use?
```
#if MAC_OS_X_VERSION_MIN_REQUIRED < 1060
```

ryandesign:
> It was [suggested](https://github.com/harfbuzz/harfbuzz/pull/1179) that this should be fixed by checking:
> 
> ```
> #if !TARGET_OS_IPHONE && MAC_OS_X_VERSION_MIN_REQUIRED < 1060
> ```
> I think that would have worked.

With that change it builds successfully on 10.4.

Another option could be:

```
#if defined(MAC_OS_X_VERSION_MIN_REQUIRED) && MAC_OS_X_VERSION_MIN_REQUIRED < 1060
```

Apple probably has some recommendations on what to do, but I'm not sure how to find them at the moment.

kencu:
Ryan is the maintainer of this software in MacPorts, has patched it already in MacPorts, and originally asked me to open this issue. I will leave the suggested fix to Ryan.

kencu:
@ryandesign -- do you want to finish this off, or shall I take it?

kencu:
We have been using `!TARGET_OS_IPHONE` on MacPorts for the past weeks, and that works on all MacOS systems from 10.4 PPC on up to the current system. It will also exclude (obviously) IPHONE OS. 

I notice there are several other places in this source file where `MAC_OS_X_VERSION_MIN_REQUIRED` is used without any `OS` guards, but presumably these are still somehow working out OK and you want to leave those alone. I will make a PR for this one change, with a short note above it about why.

PR #1571, opened 2019-02-06
hb-coretext.cc: clean up macosx test

kencu:
TARGET_OS_OSX was introduced only in late OS versions
so always returns as "0" on older systems.

Just testing for  !TARGET_OS_IPHONE can work, as it returns as !0 on older
systems where TARGET_OS_IPHONE is not defined, but is not
specific

if TARGET_OS_MAC && !(defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE)
is both specific and accurate on all systems.

This fix is tested on 10.4, 10.5, 10.6, 10.7, 10.13, and 10.14 so far.

See <https://github.com/harfbuzz/harfbuzz/issues/1540>

### DIFF
--- a/src/hb-coretext.cc
+++ b/src/hb-coretext.cc
@@ -171,7 +171,7 @@ create_ct_font (CGFontRef cg_font, CGFloat font_size)
   if (CFStringHasPrefix (cg_postscript_name, CFSTR (".SFNSText")) ||
       CFStringHasPrefix (cg_postscript_name, CFSTR (".SFNSDisplay")))
   {
-#if MAC_OS_X_VERSION_MIN_REQUIRED < 1080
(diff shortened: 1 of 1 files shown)


behdad:
Is the `TARGET_OS_MAC` check doing anything?

behdad:
Isn't that always true?

kencu:
TARGET_OS_MAC would be true building on any macOS system. If this file is never built on any other system, that test is indeed redundant. I put it in because it is typically used in the Apple headers I notice, and for symmetry with what was there before to make it not too different than before and as clear as possible what was being done.

I am happy to change it in any way you prefer, of course.

behdad:
hb-coretext is only built on Apple systems, yes.

kencu:
<https://github.com/harfbuzz/harfbuzz/pull/1578>

PR #1574, opened 2019-02-09
Add a flag to hb_buffer_t to prevent the insertion of dotted circles …

emuller-amazon:
…on incorrect character sequences.

Current behavior unchanged if this flag is not set (and it isn't by default).


src/hb-ot-shape-complex-hangul.cc
```diff
@@ -214,7 +214,8 @@ preprocess_text_hangul (const hb_ot_shape_plan_t *plan HB_UNUSED,
       else
       {
 	/* No valid syllable as base for tone mark; try to insert dotted circle. */
-	if (font->has_glyph (0x25CCu))
+      if (   !(buffer->flags & HB_BUFFER_FLAG_DO_NOT_INSERT_DOTTED_CIRCLE)
+          && font->has_glyph (0x25CCu))
```
ebraminio: Very minor, please use one tab instead 8 character as the current project style and remove extra spaces.

Some testing will be nice also. Thanks

emuller-amazon: Can you please point me to some other test I could use as a model? I have not found a test for the other HB_BUFFER_FLAG.

ebraminio: You can create an API test for it and write bare minimum check code to test with/without the flag result but I guess providing someway to have that from hb-view/hb-shape is the thing we may had.

src/hb-ot-shape-complex-vowel-constraints.cc
```diff
@@ -34,6 +34,9 @@ _hb_preprocess_text_vowel_constraints (const hb_ot_shape_plan_t *plan HB_UNUSED,
 				       hb_buffer_t              *buffer,
 				       hb_font_t                *font HB_UNUSED)
 {
+  if (unlikely (buffer->flags & HB_BUFFER_FLAG_DO_NOT_INSERT_DOTTED_CIRCLE))
```
behdad: This file is generated.  Please update the generator as well.

Also, the use of unlikely is not suggested in this case since it's not an exceptional case but a configuration.  Not a big deal.

PR #1579, opened 2019-02-13
Fix djgpp complains by tweaking templates

ebraminio:
For some reasons djgpp doesn't understand "unsigned int" can be same
with one of uint*_t anyway so lets do that for it explicitly.

Just to note, our CI's djgpp is based GCC 7.2.0 and isn't old.

### DIFF
--- a/src/hb-ot-layout.hh
+++ b/src/hb-ot-layout.hh
@@ -232,10 +232,10 @@ _hb_glyph_info_set_unicode_props (hb_glyph_info_t *info, hb_buffer_t *buffer)
        * FVSes are GC=Mn, we have use a separate bit to remember them.
        * Fixes:
        * https://github.com/harfbuzz/harfbuzz/issues/234 */
-      else if (unlikely (hb_in_range (u, 0x180Bu, 0x180Du))) props |= UPROPS_MASK_HIDDEN;
+      else if (unlikely (hb_in_range<uint32_t> (u, 0x180Bu, 0x180Du))) props |= UPROPS_MASK_HIDDEN;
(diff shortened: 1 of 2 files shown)


src/hb-ot-layout.hh
```diff
@@ -232,10 +232,10 @@ _hb_glyph_info_set_unicode_props (hb_glyph_info_t *info, hb_buffer_t *buffer)
        * FVSes are GC=Mn, we have use a separate bit to remember them.
        * Fixes:
        * https://github.com/harfbuzz/harfbuzz/issues/234 */
-      else if (unlikely (hb_in_range (u, 0x180Bu, 0x180Du))) props |= UPROPS_MASK_HIDDEN;
+      else if (unlikely (hb_in_range<uint32_t> (u, 0x180Bu, 0x180Du))) props |= UPROPS_MASK_HIDDEN;
       /* TAG characters need similar treatment. Fixes:
        * https://github.com/harfbuzz/harfbuzz/issues/463 */
-      else if (unlikely (hb_in_range (u, 0xE0020u, 0xE007Fu))) props |= UPROPS_MASK_HIDDEN;
+      else if (unlikely (hb_in_range<uint32_t> (u, 0xE0020u, 0xE007Fu))) props |= UPROPS_MASK_HIDDEN;
```
behdad: At least use hb_codepoint_t for these two?

ebraminio: Right, occurred to me after the merge also but thought doesn't matter much, lets go for that instead.

ebraminio: https://github.com/harfbuzz/harfbuzz/pull/1582

Issue #1070, opened 2018-06-25
AdobeDevanagari-Regular : Matra positioning


Amritasingh0512:
Text: ष्ट्य ष्ट्‍य ष्ठ्य ष‍्य ट्ञ ष्ट्यि ष्ट्‍यि ष्ठ्यि ष‍्यि ट्ञि
[AdobeDevanagari-Regular.zip](https://github.com/harfbuzz/harfbuzz/files/2132599/AdobeDevanagari-Regular.zip)
<img width="594" alt="devanagari_git_1" src="https://user-images.githubusercontent.com/38288804/41843440-2aa06a5a-788b-11e8-9a17-6b1a83a75b17.PNG">

behdad:
What's expected?

behdad:
When I compare to Windows 7 rendering, it's only the ष्ट्‍यि produces different results.  What are you testing against?

behdad:
I fixed the one case that we differed from Windows.

PR #1580, opened 2019-02-13
[indic] Remove superfluous ZWNJ check in final reorder of pre-base matras

adrianwong:
This is a minor change to a recent fix for #1070.

From what I can tell, it doesn't appear possible for a `Matra` to occur before a `Halant,ZWNJ` sequence, due to `Halant,ZWNJ` being a "terminating" sequence for the Indic state machine.

Using the Devanagari test case in the comments as an example, what we get from the sequence

```
U+091F,U+094D,U+200C,U+092F,U+093F
```

is ultimately two consonant syllables `U+091F,U+094D,U+200C` and `U+092F,U+093F`. `U+093F` gets reordered in front of the second syllable's base `U+092F`, so we effectively get this check for free.

### DIFF
--- a/src/hb-ot-shape-complex-indic.cc
+++ b/src/hb-ot-shape-complex-indic.cc
@@ -1199,9 +1199,14 @@ final_reordering_syllable (const hb_ot_shape_plan_t *plan,
 	      goto search;
 	    }
 	  }
-	  /* -> If ZWNJ follows this halant, position is moved after it. */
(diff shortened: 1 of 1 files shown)


behdad:
Thanks.   Just curious.  How did you come across this?

adrianwong:
Thanks for reviewing the changes, @behdad!

I'm currently working on an Indic shaper at [YesLogic](https://www.princexml.com/). The HarfBuzz code has been a very handy reference, and a huge help.

PR #1581, opened 2019-02-13
Remove coretext_aat shaper

ebraminio:
coretext_aat was a temporary shaper to redirect shaping of AAT fonts
to CoreText and leaving the rest for HarfBuzz.  As HarfBuzz now supports
AAT, and Chrome now actually ships that on a stable version on macOS
we no longer care about such use-case.  If a client really wants 100%
metrics compatibility with CoreText better to use it directly or through
our API.  Replicating the same behavior still is possible using
hb_shape_full, something we don't care or like to offer anymore.

Fixes https://github.com/harfbuzz/harfbuzz/issues/1478


behdad:
Lgtm. cc @drott

drott:
Please see my comment in https://github.com/harfbuzz/harfbuzz/issues/1478#issuecomment-446938844 - I'd prefer if we can wait until E/March with removing this completely in case we need to switch the native AAT support off in stable. Otherwise we would need to roll back stable to an old HarfBuzz.

ebraminio:
Oh, I didn't know we should wait till the release after 72. You definitely can replicate coretext_aat in your code just anytime, just something like this pseudoscope, 

```C
if (hb_aat_layout_has_substitution(face) || hb_aat_layout_has_positioning(face))
  hb_shape_full(font, buffer, features, num_features, {"coretext", "ot", "fallback"});
else
  hb_shape(font, buffer, features, num_features);
```

And that is exactly the same with `coretext_aat`. Considering the possibility, is there any chance we leave that now? Your request is definitely understandable but something not very realistic to be needed IMHO.

behdad:
Just leave this PR open till Dominik is happy with removing.  No harm.

ebraminio:
That is the idea :)

ebraminio:
Chrome 73 is also released but apparently we should wait till the end of the month as Dominik.

drott:
Yes, let's wait for a bit whether there is more feedback on 73. I think the end of the month should be good. Note that we do have a regression from the HarfBuzz roll that included the native AAT (but unrelated): https://github.com/harfbuzz/harfbuzz/issues/1607 - this would be important to fix.

ebraminio:
#1607 gone ;)

ebraminio:
Let's merge this tomorrow if you are not opposed. Thanks

drott:
Removed from Chromium ToT: 
https://chromium-review.googlesource.com/c/chromium/src/+/1546103

PR #1584, opened 2019-02-15
Support xlclang++ on AIX.

TheRealMDoerr:
I'd like to fix build with xlclang++ on AIX which defines `__ibmxl__` but not `__IBMCPP__`.


ebraminio:
Makes sense. Thanks!

TheRealMDoerr:
Excellent. Thank you very much.

PR #1589, opened 2019-02-19
[doc] Move feature_t and its function to hb-common

khaledhosny:
It is rather confusing to have script, language etc, in hb-common section while feature is in hb-shape section. I keep looking for it in hb-common section then using the API index because I can’t find it there.


behdad:
Agree.  Same with hb_variation_t perhaps?

khaledhosny:
> Agree. Same with hb_variation_t perhaps?

I guess so.

Issue #1558, opened 2019-01-24
[subset] TrueType/CFF2 variable font


blueshade7:
- [x] Add HVAR & VVAR subsetting for TT/CFF2
- [x] Add gvar struct for TT (sanitize, look-up, subsetting)
- [x] Calculate TT VF glyph widths (if HVAR/VVAR absent)
- [x] Calculate TT VF glyph extents from glyf & gvar (no hint support)
- [x] Add tests (subset: gvar, HVAR, VVAR)
- [x] Add tests (TT VF widths & extents)
- [x] ItemVariationStore subsetting in GPOS, GDEF
- [ ] Add tests (subset: VF GPOS, GDEF)

Not to subset: MVAR, BASE, cvar
Not in scope: axes subsetting, instantiation

blueshade7:
Removed task "Calculate TT/CFF2 VF glyph widths from HVAR/VVAR" because it is already in place

blueshade7:
This is my plan for HVAR & VVAR subsetting including VariationStore.

`HVAR::subset` to be called by `_subset2`; seems more convenient than `_subset` which requires to know size upfront (maybe rewrite CFF/CFF2 subsetter using `_subset2` later?)

To subset inner indices for VariationStore, I am going to repurpose `CFF::remap_t` (one for each outer index), rewriting it based on `hb_map_t` & `hb_set_t`. Inner indices referenced by advMap/lsbMap/rsbMap are collected in `hb_set_t`, then old to new mapping is stored in `hb_map_t`.

I am not going to bother subsetting & remapping outer indices; for an unused outer index, its corresponding item variation data subtable is emptied. This wastes 6 bytes each but I don’t expect to see many outer indices in wild after all.

blueshade7:
Actually an empty item variation data subtable wastes 10 bytes counting 4-byte offset. It would be nice if an item variation data offset may be set to null for a missing table, but OTF spec doesn't seem to say so. Darn.

blueshade7:
Created branch var-subset for this task

blueshade7:
Decided to subset outer indices & variation data subtables after all, as fontTools appears to support it.
Pushed HVAR & VVAR subset to var-subset branch.

blueshade7:
Pushed advance width calculation from gvar & glyf

blueshade7:
Pushed glyph extents calculation using gvar & glyf
Added api test `test-ot-metrics-tt-var`, subset full font test with a TT variable font

[...]

PR #1594, opened 2019-02-22
Issue 1558: [subset] TrueType/CFF2 variable font

blueshade7:
#1558: [subset] TrueType/CFF2 variable font


behdad:
Okay, I finally merged the 'iter' branch.  Can you rebase your branch?  I'm going to review it and aim to land next week.  Thanks.

src/hb-bimap.hh
```diff
@@ -0,0 +1,109 @@
...
+  /* Optional: after finished adding all mappings in a random order,
+   * reassign new ids to old ids so that both are in the same order. */
+  void reorder ()
+  {
+    new_to_old.qsort (cmp_id);
+    for (hb_codepoint_t _new = 0; _new < count; _new++)
+      old_to_new.set (to_old (_new), _new);
+  }
+
+  unsigned int get_count () const { return count; }
```
behdad: `get_population`. Also don't need to keep count separately.  But map_t and vector_t have it.

src/hb-bimap.hh
```diff
@@ -0,0 +1,109 @@
...
+#ifndef HB_BIMAP_HH
+#define HB_BIMAP_HH
+
+#include "hb.hh"
+
+/* Bi-directional map.
+ * new ids are assigned incrementally & contiguously to old ids
+ * which may be added randomly & sparsely
+ * all mappings are 1-to-1 in both directions */
+struct hb_bimap_t
```
behdad: Okay so when you told me you need this initially I think you said it's a pair of inverse maps.  Hence the name suggestion.  But I see one way it's a vector.  Not sure if we should keep it that way.

src/hb-bimap.hh
```diff
@@ -0,0 +1,109 @@
...
+    count = i;
+    return old_to_new.successful && !new_to_old.in_error ();
+  }
+
+  static int cmp_id (const void* a, const void* b)
+  { return (int)*(const hb_codepoint_t *)a - (int)*(const hb_codepoint_t *)b; }
+
+  /* Optional: after finished adding all mappings in a random order,
+   * reassign new ids to old ids so that both are in the same order. */
+  void reorder ()
```
behdad: This sounds like a weirdly specific method to live here.

src/hb-bimap.hh
```diff
@@ -0,0 +1,109 @@
...
+struct hb_bimap_t
+{
+  hb_bimap_t () { init (); }
+  ~hb_bimap_t () { fini (); }
+
+  void init ()
+  {
+    count = 0;
+    old_to_new.init ();
+    new_to_old.init ();
```
behdad: Also the names old/new don't make much sense.  Forward/backward probably better.  No strong opinion though.

blueshade7: I think I will rename them to forw_map & back_map.

src/hb-bimap.hh
blueshade7: I think I will move this facility to a sub-class `hb_reordered_bimap_t` in the same .hh, leaving essentials in `hb_bimap_t` as pure and simple.

src/hb-bimap.hh
blueshade7: I can use`hb_map_t` for both directions, allowing both mappings to be sparse in case we find such uses later.

src/hb-bimap.hh
blueshade7: true.

src/hb-ot-glyf-table.hh
```diff
@@ -1072,18 +1072,8 @@ struct glyf
...
-
-      /* Undocumented rasterizer behavior: shift glyph to the left by (lsb - xMin), i.e., xMin = lsb */
-      /* extents->x_bearing = hb_min (glyph_header.xMin, glyph_header.xMax); */
-      extents->x_bearing = hmtx_accel.get_side_bearing (glyph);
-      extents->y_bearing = hb_max (glyph_header.yMin, glyph_header.yMax);
-      extents->width     = hb_max (glyph_header.xMin, glyph_header.xMax) - hb_min (glyph_header.xMin, glyph_header.xMax);
-      extents->height    = hb_min (glyph_header.yMin, glyph_header.yMax) - extents->y_bearing;
-
+      hb_bytes_t ((const char *) glyf_table + start_offset,
+		  end_offset - start_offset).as<GlyphHeader> ()->get_extents (extents);
```
ebraminio: https://github.com/harfbuzz/harfbuzz/blob/ae2c4c9d2db196bf1d0b6e8b58f6cbe52f0240c8/src/hb-ot-glyf-table.hh#L296 isn't doing the same, will submit I PR to the branch

ebraminio:
@blueshade7 merged your branch with the latest changes and not failure of checks this time. Sorry for the trouble, wish you feel this ready to merge soon, thanks.

ebraminio:
Please note that I've not used your merge commit so you should `git fetch` and `git reset --hard 8106ba7`, the easiest way I guess, you local branch

ebraminio:
Merged shaping related changes of hvar/vvar, and glyf and gvar changes of this PR in #1999 and merge the branch with the changes

ebraminio:
Hmm actually with the mentioned merges, this has become easy to review also.

ebraminio:
Got permission to merge this, will propose fixes upon it. Thank you for the great and nice work.

PR #1598, opened 2019-02-26
Fix private use langtag parsing

mhosken:
This fixes a bug whereby language tags of the form x-hbot-OTLG are not being parsed.


mhosken:
Ooh, no hyphen expected after x-hbot hmm!

PR #1600, opened 2019-02-27
[dwrite] A new API, hb_directwrite_face_get_font_face

ebraminio:
Can be useful when using HarfBuzz for font loading and shaping
but using DirectWrite for rendering.

Code I literally used for testing,
```C
﻿#include <DWrite_1.h>
#include <stdio.h>
#include <stdint.h>
#include <hb.h>
#include <hb-directwrite.h>

int main (int argc, char **argv)
{
  hb_blob_t *blob = hb_blob_create_from_file ("C:\\Windows\\Fonts\\tahoma.ttf");
  hb_face_t *hb_face = hb_face_create (blob, 0);
  IDWriteFontFace *fontFace = hb_directwrite_face_get_font_face (hb_face);
  hb_face_t *face = hb_directwrite_face_create (fontFace);

  printf ("%d == %d == %d", hb_face_get_glyph_count (hb_face), fontFace->GetGlyphCount(), hb_face_get_glyph_count (face));

  hb_face_destroy (face);
  scanf ("%d", new int);

  return 0;
}
```

### DIFF
--- a/src/hb-directwrite.cc
+++ b/src/hb-directwrite.cc
@@ -846,10 +846,23 @@ _hb_directwrite_shape (hb_shape_plan_t    *shape_plan,
 				     features, num_features, 0);
 }
 
-/*
- * Public [experimental] API
- */
-
+/**
+ * hb_directwrite_shape_experimental_width:
+ * Experimental API to test DirectWrite's justification algorithm.
+ *
+ * It inserts Kashida at wrong order so don't use the API ever.
+ *
+ * It doesn't work with cygwin/msys due to header bugs so one
+ * should use MSVC toolchain in order to use it for now.
+ *
+ * @font:
+ * @buffer:
+ * @features:
+ * @num_features:
+ * @width:
+ *
+ * Since: 1.4.2
+ **/
 hb_bool_t
 hb_directwrite_shape_experimental_width (hb_font_t          *font,
 					 hb_buffer_t        *buffer,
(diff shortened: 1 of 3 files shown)


src/hb-directwrite.cc
```diff
@@ -928,3 +942,15 @@ hb_directwrite_face_create (IDWriteFontFace *font_face)
...
+/**
+* hb_directwrite_face_get_font_face:
+* @face:
+*
+* Since: REPLACEME
+**/
+IDWriteFontFace *
+hb_directwrite_face_get_font_face (hb_face_t *face)
+{
+  return face->data.directwrite->fontFace;
```
khaledhosny: What if `data` is null or it isn’t a DirectWrite-backed face?

ebraminio: IIUC it will always be initiated as can be seen on the test code. I can do null check but will be nice to apply such to `hb_coretext_face_get_cg_font` also.

behdad: Shouldn't need check.

ebraminio: Hmm, I guess we should think about the naming also

ebraminio:
Naming, I guess

behdad:
So, re naming.  We have been inconsistent in FT vs Coretext....  We have `hb_ft_font_get_face` vs `hb_coretext_font_get_ct_font` / `hb_coretext_face_get_cg_font`.  That said, I think I'm fine with `hb_directwrite_face_get_font_face`.

Issue #1381, opened 2018-11-13
Completely move Chromium off sfntly


ebraminio:
"Completely move Chromium off sfntly" was on roadmap for mid 2019 but I was curious what is missing so am turning it into an issue.

Apparently this is the only use of sfntly https://cs.chromium.org/chromium/src/third_party/skia/src/pdf/SkPDFFont.cpp?l=423-440 and https://cs.chromium.org/chromium/src/third_party/skia/src/pdf/SkPDFFont.cpp?l=333-396 and sfntly can not provide `CFF` subsetting which we can provide eventually and GPOS/GSUB subsetting should be done before the port also I guess. Other than this, I wonder if is there any need for sfntly on Chromium?

behdad:
In theory current subsetter should be enough.  GSUB/GPOS is NOT required (and is not implemented by sfntly either) since this is for PDF rendering, which has text already layed out.

bungeman:
@HalCanary is interested in this as well.

behdad:
Hal, can you list requirements / sfntly API you need to replace?

HalCanary:
Here's what Skia's PDF generator backend (SkPDF) needs:

Reproduce the functionality of `SfntlyWrapper::SubsetFont`.

Given raw font data (as returned by `SkTypeface::openStream`), the ttcIndex, and a set of glyph indices, I would like minimal subset font data suitable for embedding in a PDF.

I don't need shaping information, since Skia outputs directly positioned glyphs.

ebraminio:
Sure. It API is easy to use and test-subset-*.c are good starting, let me know anything needed

ebraminio:
Simplest snippet for subsetting,

```C
/* Create some font */
hb_blob_create (font_data, font_data_length, HB_MEMORY_MODE_READONLY, nullptr, nullptr);
/* Or if you like to read an actual file: hb_blob_t *blob = hb_blob_create_from_file (path); */
hb_face_t *face = hb_face_create (blob, 0); // this is ttcIndex
hb_blob_destory (blob);

/* Add your codepoints here */
hb_set_t *codepoints = hb_set_create();
hb_set_add (codepoints, 97);
hb_set_add (codepoints, 99);
hb_face_t *subset = hb_subset (face, hb_subset_test_create_input (codepoints));
hb_set_destroy (codepoints);

/* Result blob */
hb_blob_t *result = hb_face_reference_blob (subset);
unsigned int length;
[...]

PR #1601, opened 2019-03-01
Make hb_subset_input_glyph_set () actually do something.

garretrieger:
Fix needed for #1381

### DIFF
--- a/test/api/test-subset-glyf.c
+++ b/test/api/test-subset-glyf.c
@@ -80,6 +80,29 @@ test_subset_glyf (void)
 }
 
 static void
+test_subset_glyf_with_input_glyphs (void)
+{
+  hb_face_t *face_abc = hb_test_open_font_file ("fonts/Roboto-Regular.abc.ttf");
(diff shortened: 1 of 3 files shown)


test/api/test-subset-glyf.c
```diff
@@ -79,6 +79,29 @@ test_subset_glyf (void)
...
+  hb_set_t *glyphs = hb_set_create();
+  hb_face_t *face_abc_subset;
+  hb_set_add (glyphs, 1);
+  hb_set_add (glyphs, 3);
+  face_abc_subset =
+      hb_subset_test_create_subset (face_abc, hb_subset_test_create_input_from_glyphs (glyphs));
+  hb_set_destroy (glyphs);
+
+  hb_subset_test_check (face_ac, face_abc_subset, HB_TAG ('g','l','y','f'));
+  hb_subset_test_check (face_ac, face_abc_subset, HB_TAG ('l','o','c', 'a'));
```
rsheeter: the tables define their tags, suggest using those (HB_OT_TAG_glyf, HB_OT_TAG_loca)

garretrieger: Unfortunately we don't have access to private headers in this context.

PR #1608, opened 2019-03-07
Add a macOS 10.14.3 fonts tests

ebraminio:


ebraminio:
Only Khmer MN had significant difference from 10.13 results which I check it visually against CoreText

PR #1615, opened 2019-03-13
Remove X from hb-ot-shape-complex-arabic-table.hh

dscorbett:
Code points that have entries in hb-ot-shape-complex-arabic-table.hh but don’t appear in ArabicShaping.txt currently get `X` in the table. This PR resolves `X` to `T` or `U`. It takes the same amount of space and may save some time.

### DIFF
--- a/src/gen-arabic-table.py
+++ b/src/gen-arabic-table.py
@@ -40,7 +40,7 @@ def read_blocks(f):
 		for u in range (start, end + 1):
 			blocks[u] = t
 
-def print_joining_table(f):
+def print_joining_table(f, categories):
 
 	values = {}
 	for line in f:
@@ -71,7 +71,6 @@ def print_joining_table(f):
 		print ("#define %s	%s" % (short, value))
 
 	uu = sorted(values.keys())
-	num = len(values)
 	all_blocks = set([blocks[u] for u in uu])
 
 	last = -100000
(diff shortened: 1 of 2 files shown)


behdad:
While I see the benefit, this does have the effect of hardcoding those chars' behavior at code generation time, rather than using user-provided unicode funcs at runtime.  Given that the runtime cost is negligible, I prefer to keep current behavior.  WDYT?

dscorbett:
It’s already hard-coding the Joining_Type for all the code points in ArabicShaping.txt instead of using a user-provided function. Is there a use case for making Joining_Type user-configurable for Joining_Type=X code points? I suppose if a user wanted to override the General_Category of one of these characters from mark to non-mark or vice versa, thus changing its Joining_Type, this pull request would thwart them, but I don’t see how it is useful to override that.

behdad:
> It’s already hard-coding the Joining_Type for all the code points in ArabicShaping.txt instead of using a user-provided function.

Right.

> Is there a use case for making Joining_Type user-configurable for Joining_Type=X code points? I suppose if a user wanted to override the General_Category of one of these characters from mark to non-mark or vice versa, thus changing its Joining_Type, this pull request would thwart them, but I don’t see how it is useful to override that.

I'm just uncomfortable with this because it applies that (unimportant) limitation to characters that happen to be neighboring characters with Arabic Joining property.  That's the part that is arbitrary and bothers me...

dscorbett:
It is a little arbitrary, but not completely. Most of these jt=X characters are used in joining scripts where it wouldn’t make sense to override them to anything else. I agree it doesn’t really matter; it’s a matter of opinion. Let’s merge it or not, and not keep this PR open indefinitely.

behdad:
I prefer to keep it as is.  Thanks for understanding.

PR #1616, opened 2019-03-14
[fvar] Accept the table on 0 named instance

### LINKED ISSUE #1607: [Regression] HarfBuzz rejects fvar table with 0 named instances after 587d49fc657c10
_(Thanks to http://underware.nl/ for reporting this bug.)_

https://github.com/harfbuzz/harfbuzz/commit/587d49fc657c10c8a20f2409a04d72bf80bb361e line 112 changed the sanitisation rules for the trail end of `fvar`. 

Further down in that change:
`get_first_instance ()` was changed to `get_instance (unsigned int i)`

in which  

```
  inline const InstanceRecord * get_first_instance (void) const	  inline const InstanceRecord &get_instance (unsigned int i) const
  { return &StructAfter<InstanceRecord> (get_axes ()); }
```

was changed to

```
  inline const InstanceRecord &get_instance (unsigned int i) const
  { return &StructAfter<InstanceRecord> (get_axes ()); }	  {
    if (unlikely (i >= instanceCount)) return Null (InstanceRecord);

    return StructAtOffset<InstanceRecord> (&StructAfter<InstanceRecord> (get_axes ()),
					  i * instanceSize);
  }
```

I am assuming that for a font with no named instances the latter does return a `Null (InstanceRecord)` which makes the 
` c->check_range (&get_instance (0), instanceCount, instanceSize));` sanitisation step fail.

If I remove `if (unlikely (i >= instanceCount)) return Null (InstanceRecord);` it starts accepting the test font's fvar again.

My interpretation of https://docs.microsoft.com/en-us/typography/opentype/spec/fvar is that no named instances are acceptable, so I consider this a regression which over-eagerly rejects fonts with no named instances.

ebraminio:
Fixes #1607 

Speculative, a test font also should be added.

### DIFF
--- a/src/hb-ot-var-fvar-table.hh
+++ b/src/hb-ot-var-fvar-table.hh
@@ -109,7 +109,7 @@ struct fvar
 		  axisSize == 20 && /* Assumed in our code. */
 		  instanceSize >= axisCount * 4 + 4 &&
 		  get_axes ().sanitize (c) &&
-		  c->check_range (get_instance (0), instanceCount, instanceSize));
+		  (instanceCount == 0 || c->check_range (get_instance (0), instanceCount, instanceSize)));
   }
 
   unsigned int get_axis_count () const { return axisCount; }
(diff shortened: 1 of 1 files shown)


behdad:
I don't like this change.  I argue that instead, `check_range` should change to return true if count is zero.

ebraminio:
Right, I was thinking about that also, I'll go for that.

ebraminio:
In favor of  #1617

PR #1617, opened 2019-03-14
Allow zero length ranges in sanitization

### LINKED ISSUE #1607: [Regression] HarfBuzz rejects fvar table with 0 named instances after 587d49fc657c10

ebraminio:
Fixes #1607

Closes #1616 also

And sorry for the white space noise. Let me add a test case also.


src/hb-machinery.hh
```diff
@@ -326,37 +326,39 @@ struct hb_sanitize_context_t :
...
   bool check_range (const void *base,
-			   unsigned int len) const
+		    unsigned int len) const
   {
     const char *p = (const char *) base;
-    bool ok = this->start <= p &&
-	      p <= this->end &&
-	      (unsigned int) (this->end - p) >= len &&
-	      this->max_ops-- > 0;
+    bool ok = len == 0 ||
```
behdad: I personally prefer `!len` to `len == 0`, but am fine with this.

ebraminio:
Now has a test also which didn't work before the test.

test/api/test-ot-face.c
```diff
@@ -110,12 +110,21 @@ test_ot_face_empty (void)
   test_face (hb_face_get_empty (), 0);
 }
 
+static void
+test_ot_var_axis_on_zero_named_instance ()
+{
+  hb_face_t *face = hb_test_open_font_file ("fonts/Zycon.ttf");
```
ebraminio: Zycon was available somewhere else on the repo and as duplicating file costs nothing on git I am copying and using it for the test.

ebraminio:
Bots are happy, may I merge it?

behdad:
Yep. Thanks!

drott:
Thanks for the fix, great!

PR #1619, opened 2019-03-15
cff2 subset fuzzer issues

blueshade7:
oss-fuzz 13732, 13733, 13736


ebraminio:
Thanks for the quick fix!

@drott please make sure this change will be used in the same branch as https://github.com/harfbuzz/harfbuzz/commit/8aaab78efcac81a05ec919be13792c98741ea1b5

PR #1621, opened 2019-03-16
[Docs] Add inline gtk-doc comments for hb-ot-layout functions.

n8willis:
This adds GTK-Doc function documentation for the majority of the public API in hb-ot-layout. There are some small gaps remaining, but not much.


n8willis:
Ugh. A little too early.

PR #1622, opened 2019-03-16
[Docs] Add inline gtk-doc comments for hb-ot-layout

n8willis:
This adds GTK-Doc documentation comments for most of hb-ot-layout.cc's public API, and adds documentation comments for the hb_ot_layout_get_glyph_class enum in hb-ot-layout.h. Several odd bits and pieces remain, but this is a big chunk.

### DIFF
--- a/src/hb-ot-layout.cc
+++ b/src/hb-ot-layout.cc
@@ -497,8 +497,8 @@ hb_ot_layout_table_find_script (hb_face_t    *face,
  * @face: #hb_face_t to work upon
  * @table_tag: HB_OT_TAG_GSUB or HB_OT_TAG_GPOS
  * @script_tags: Array of #hb_tag_t script tags
- * @script_index: The index of the requested script tag
- * @chosen_script: #hb_tag_t of the script tag requested
+ * @script_index: (out): The index of the requested script tag
+ * @chosen_script: (out): #hb_tag_t of the script tag requested
  *
  * Deprecated
  **/
@@ -506,8 +506,8 @@ hb_bool_t
 hb_ot_layout_table_choose_script (hb_face_t      *face,
 				  hb_tag_t        table_tag,
 				  const hb_tag_t *script_tags,
-				  unsigned int   *script_index,
-				  hb_tag_t       *chosen_script)
+				  unsigned int   *script_index  /* OUT */,
+				  hb_tag_t       *chosen_script /* OUT */)
 {
   const hb_tag_t *t;
   for (t = script_tags; *t; t++);
--- a/src/hb-ot-layout.cc
+++ b/src/hb-ot-layout.cc
@@ -843,7 +843,7 @@ hb_ot_layout_language_get_feature_indexes (hb_face_t    *face,
  * @start_offset: The offset into the feature-tag array to start returning
  * @feature_count: (inout): The number of feature tags to return as input;
  *                 The number of feature tags returned in the array as output
- * @feature_tags: (out ) The array of #hb_tag_t feature tags found for the query
+ * @feature_tags: (out): The array of #hb_tag_t feature tags found for the query
  *
  * Fetches a list of all features in the specified face's GSUB table
  * or GPOS table, underneath the specified script and language. The list
@@ -1589,15 +1589,15 @@ hb_ot_layout_get_size_params (hb_face_t       *face,
  * @face: #hb_face_t to work upon
  * @table_tag: table tag to query, "GSUB" or "GPOS".
  * @feature_index: index of feature to query.
- * @label_id: (out): (allow-none): The ‘name’ table name ID that specifies a string
+ * @label_id: (out) (allow-none): The ‘name’ table name ID that specifies a string
  *            for a user-interface label for this feature. (May be NULL.)
- * @tooltip_id: (out): (allow-none): The ‘name’ table name ID that specifies a string
+ * @tooltip_id: (out) (allow-none): The ‘name’ table name ID that specifies a string
  *              that an application can use for tooltip text for this
  *              feature. (May be NULL.)
- * @sample_id: (out): (allow-none): The ‘name’ table name ID that specifies sample text
+ * @sample_id: (out) (allow-none): The ‘name’ table name ID that specifies sample text
  *             that illustrates the effect of this feature. (May be NULL.)
- * @num_named_parameters: (out): (allow-none):  Number of named parameters. (May be zero.)
- * @first_param_id: (out): (allow-none): The first ‘name’ table name ID used to specify
+ * @num_named_parameters: (out) (allow-none):  Number of named parameters. (May be zero.)
+ * @first_param_id: (out) (allow-none): The first ‘name’ table name ID used to specify
  *                  strings for user-interface labels for the feature
  *                  parameters. (Must be zero if numParameters is zero.)
  *
--- a/src/hb-ot-layout.cc
+++ b/src/hb-ot-layout.cc
@@ -314,10 +314,10 @@ hb_ot_layout_get_glyphs_in_class (hb_face_t                  *face,
  * hb_ot_layout_get_attach_points:
  * @face: The #hb_face_t to work on
  * @glyph: The #hb_codepoint_t code point to query
- * @start_offset: The offset into the attachment-point array to start returning
- * @point_count: (inout): The number of attachment points to return as input;
- *               The number of attachment points returned in the array as output
- * @point_array: (out): The array of attachment points found for the query
+ * @start_offset: offset of the first attachment point to retrieve
+ * @point_count: (inout) (allow-none): Input = the maximum number of attachment points to return;
+ *               Output = the actual number of attachment points returned (may be zero)
+ * @point_array: (out) (array length=point_count): The array of attachment points found for the query
  *
  * Fetches a list of all attachment points for the specified glyph in the GDEF
  * table of the face. The list returned will begin at the offset provided. 
@@ -344,10 +344,10 @@ hb_ot_layout_get_attach_points (hb_face_t      *face,
  * @font: The #hb_font_t to work on
  * @direction: The #hb_direction_t text direction to use
  * @glyph: The #hb_codepoint_t code point to query
- * @start_offset: The offset into the caret array to start returning
- * @caret_count: (inout): The number of caret positions to return as input;
- *               The number of caret positions returned in the array as output
- * @caret_array: (out): The array of caret positions found for the query
+ * @start_offset: offset of the first caret position to retrieve
+ * @caret_count: (inout) (allow-none): Input = the maximum number of caret positions to return;
+ *               Output = the actual number of caret positions returned (may be zero)
+ * @caret_array: (out) (array length=caret_count): The array of caret positions found for the query
  *
  * Fetches a list of the caret positions defined for a ligature glyph in the GDEF
  * table of the font. The list returned will begin at the offset provided.
@@ -425,10 +425,10 @@ get_gsubgpos_table (hb_face_t *face,
  * hb_ot_layout_table_get_script_tags:
  * @face: #hb_face_t to work upon
  * @table_tag: HB_OT_TAG_GSUB or HB_OT_TAG_GPOS
- * @start_offset: The offset into the script-tag array to start returning
- * @script_count: (inout): The number of script tags to return as input;
- *                The number of script tags returned in the array as output
- * @script_tags: (out): The array of #hb_tag_t script tags found for the query
+ * @start_offset: offset of the first script tag to retrieve
+ * @script_count: (inout) (allow-none): Input = the maximum number of script tags to return;
+ *                Output = the actual number of script tags returned (may be zero)
+ * @script_tags: (out) (array length=script_count): The array of #hb_tag_t script tags found for the query
  *
  * Fetches a list of all scripts enumerated in the specified face's GSUB table
  * or GPOS table. The list returned will begin at the offset provided.
@@ -579,10 +579,10 @@ hb_ot_layout_table_select_script (hb_face_t      *face,
  * hb_ot_layout_table_get_feature_tags:
  * @face: #hb_face_t to work upon
  * @table_tag: HB_OT_TAG_GSUB or HB_OT_TAG_GPOS
- * @start_offset: The offset into the feature-tag array to start returning
- * @feature_count: (inout): The number of feature tags to return as input;
- *                 The number of feature tags returned in the array as output
- * @feature_tags: (out): Array of feature tags found in the table
+ * @start_offset: offset of the first feature tag to retrieve
+ * @feature_count: (inout) (allow-none): Input = the maximum number of feature tags to return;
+ *                 Output = the actual number of feature tags returned (may be zero)
+ * @feature_tags: (out) (array length=feature_count): Array of feature tags found in the table
  *
  * Fetches a list of all feature tags in the given face's GSUB or GPOS table.
  *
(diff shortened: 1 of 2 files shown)


src/hb-ot-layout.cc
```diff
@@ -62,18 +62,50 @@
  * kern
  */
 
+/**
+ * hb_ot_layout_has_kerning:
+ * @face: The #hb_face_t to work on
+ *
+ * Tests whether a face includes any kerning data. Examines the 'kern' table.
```
khaledhosny: I don’t think “any” is appropriate here as it does not check `GPOS` kerning. I think something like “includes kerning data in the `kern` table” would be less confusing.

src/hb-ot-layout.cc
```diff
@@ -62,18 +62,50 @@
...
 {
   return face->table.kern->has_data ();
 }
 
+
+/**
+ * hb_ot_layout_has_machine_kerning:
+ * @face: The #hb_face_t to work on
+ *
+ * Tests whether a face includes any state-machine kerning.
```
khaledhosny: Should mention `kern` table here as well I guess (though it should be obvious, but not everyone reading this knows the intrinsics of font formats).

src/hb-ot-layout.cc
```diff
@@ -62,18 +62,50 @@
...
 {
   return face->table.kern->has_state_machine ();
 }
 
+
+/**
+ * hb_ot_layout_has_cross_kerning:
+ * @face: The #hb_face_t to work on
+ *
+ * Tests whether a face has any cross-stream kerning (i.e., kerns
```
khaledhosny: Same as above.

src/hb-ot-layout.cc
```diff
@@ -250,6 +306,22 @@ hb_ot_layout_get_glyphs_in_class (hb_face_t                  *face,
...
 }
 
+
+/**
+ * hb_ot_layout_get_attach_points:
+ * @face: The #hb_face_t to work on
+ * @glyph: The #hb_codepoint_t code point to query
+ * @start_offset: The offset into the attachment-point array to start returning
+ * @point_count: (in) The number of attachment points to return
+ *               (out) The number of attachment points returned in the array
```
khaledhosny: The gobject-introspection syntax should be `@point_count: (inout):`, the text would be something like “size of point array as input, filled size as output”.

src/hb-ot-layout.cc
```diff
@@ -250,6 +306,22 @@ hb_ot_layout_get_glyphs_in_class (hb_face_t                  *face,
...
+ * @point_array: (out) The array of attachment points found for the query
```
khaledhosny: Missing a colon after `(out)`.

src/hb-ot-layout.cc
```diff
@@ -263,6 +335,21 @@ hb_ot_layout_get_attach_points (hb_face_t      *face,
...
+
+/**
+ * hb_ot_layout_get_ligature_carets:
+ * @font: The #hb_font_t to work on
+ * @direction: The #hb_direction_t text direction to use
+ * @glyph: The #hb_codepoint_t code point to query
+ * @start_offset: The offset into the caret array to start returning
+ * @caret_count: (in) The number of caret positions to return
+ *               (out) The number of caret positions returned in the array
+ * @caret_array: (out) The array of caret positions found for the query
```
khaledhosny: Same as above.

src/hb-ot-layout.cc
```diff
@@ -331,6 +418,19 @@ get_gsubgpos_table (hb_face_t *face,
...
 
 
+/**
+ * hb_ot_layout_table_get_script_tags:
+ * @face: #hb_face_t to work upon
+ * @table_tag: HB_OT_TAG_GSUB or HB_OT_TAG_GPOS
+ * @start_offset: The offset into the script-tag array to start returning
+ * @script_count: (in) The number of script tags to return
+ *                (out) The number of script tags returned in the array
+ * @script_tags: (out) The array of #hb_tag_t script tags found for the query
```
khaledhosny: Same as above.

src/hb-ot-layout.cc
```diff
@@ -345,11 +445,24 @@ hb_ot_layout_table_get_script_tags (hb_face_t    *face,
 
 #define HB_OT_TAG_LATIN_SCRIPT		HB_TAG ('l', 'a', 't', 'n')
 
+/**
+ * hb_ot_layout_table_find_script:
+ * @face: #hb_face_t to work upon
+ * @table_tag: HB_OT_TAG_GSUB or HB_OT_TAG_GPOS
+ * @script_tag: #hb_tag_t of the script tag requested
+ * @script_index: (out) The index of the requested script tag
```
khaledhosny: Missing colon after `(out)`.

src/hb-ot-layout.cc
```diff
@@ -375,6 +488,16 @@ hb_ot_layout_table_find_script (hb_face_t    *face,
   return false;
 }
 
+/**
+ * hb_ot_layout_table_choose_script:
+ * @face: #hb_face_t to work upon
+ * @table_tag: HB_OT_TAG_GSUB or HB_OT_TAG_GPOS
+ * @script_tags: Array of #hb_tag_t script tags
+ * @script_index: The index of the requested script tag
+ * @chosen_script: #hb_tag_t of the script tag requested
```
khaledhosny: The last two are missing `(out)` tag.

src/hb-ot-layout.cc
```diff
@@ -389,6 +512,12 @@ hb_ot_layout_table_choose_script (hb_face_t      *face,
 
 /**
  * hb_ot_layout_table_select_script:
+ * @face: #hb_face_t to work upon
+ * @table_tag: HB_OT_TAG_GSUB or HB_OT_TAG_GPOS
+ * @script_count: Number of script tags in the array
+ * @script_tags: Array of #hb_tag_t script tags
+ * @script_index: (out) The index of the requested script
+ * @chosen_script: (out) #hb_tag_t of the requested script
```
khaledhosny: Missing colon after `(out)`.

src/hb-ot-layout.cc
```diff
@@ -442,6 +571,19 @@ hb_ot_layout_table_select_script (hb_face_t      *face,
...
 
+
+/**
+ * hb_ot_layout_table_get_feature_tags:
+ * @face: #hb_face_t to work upon
+ * @table_tag: HB_OT_TAG_GSUB or HB_OT_TAG_GPOS
+ * @start_offset: The offset into the feature-tag array to start returning
+ * @feature_count: (in) The number of feature tags to return
+ *                 (out) The number of feature tags returned in the array
+ * @feature_tags: (out) Array of feature tags found in the table
```
khaledhosny: Same as above.

src/hb-ot-layout.cc
```diff
@@ -454,11 +596,24 @@ hb_ot_layout_table_get_feature_tags (hb_face_t    *face,
   return g.get_feature_tags (start_offset, feature_count, feature_tags);
 }
 
+
+/**
+ * hb_ot_layout_table_find_feature:
+ * @face: #hb_face_t to work upon
+ * @table_tag: HB_OT_TAG_GSUB or HB_OT_TAG_GPOS
+ * @feature_tag: The #hb_tag_t og the requested feature tag
+ * @feature_index: The index of the requested feature
```
khaledhosny: Missing `(out)`.

src/hb-ot-layout.cc
n8willis: OK; I had wondered about that. Will do.

src/hb-ot-layout.cc
n8willis: No problem, but (for my understanding) is that a style issue or does it affect building gtk-doc?

khaledhosny: It breaks gobject-introspection, and I *think* the HTML generated by GTK-Doc will also be different (these annotations take different styling).

khaledhosny: See how `[out]` is styled here https://harfbuzz.github.io/harfbuzz-hb-buffer.html#hb-buffer-get-glyph-infos.

n8willis: Are there any other annotations worth adding?

n8willis: Well, if there are, that's for later.

khaledhosny: I think `(array length=...)` and `(transfer none)` (as appropriate, search for them in existing files).

src/hb-ot-layout.cc
```diff
@@ -589,6 +833,23 @@ hb_ot_layout_language_get_feature_indexes (hb_face_t    *face,
...
+/**
+ * hb_ot_layout_language_get_feature_tags:
+ * @face: #hb_face_t to work upon
+ * @table_tag: HB_OT_TAG_GSUB or HB_OT_TAG_GPOS
+ * @script_index: The index of the requested script tag
+ * @language_index: The index of the requested language tag
+ * @start_offset: The offset into the feature-tag array to start returning
+ * @feature_count: (inout): The number of feature tags to return as input;
+ *                 The number of feature tags returned in the array as output
+ * @feature_tags: (out ) The array of #hb_tag_t feature tags found for the query
```
khaledhosny: There is an extra space in `(out )` and no colon after it.

src/hb-ot-layout.cc
```diff
@@ -1151,20 +1583,21 @@ hb_ot_layout_get_size_params (hb_face_t       *face,
...
 }
 
+
 /**
  * hb_ot_layout_feature_get_name_ids:
  * @face: #hb_face_t to work upon
  * @table_tag: table tag to query, "GSUB" or "GPOS".
  * @feature_index: index of feature to query.
- * @label_id: (out) (allow-none): The ‘name’ table name ID that specifies a string
+ * @label_id: (out): (allow-none): The ‘name’ table name ID that specifies a string
```
khaledhosny: The colon should be after all the annotations, I think.

src/hb-ot-layout.cc
khaledhosny: Also `(allow-none)`.

src/hb-ot-layout.cc
```diff
@@ -966,6 +1348,15 @@ hb_ot_layout_lookup_would_substitute (hb_face_t            *face,
   return l.would_apply (&c, &face->table.GSUB->accels[lookup_index]);
 }
 
+
+/**
+ * hb_ot_layout_substitute_start:
+ * @font: #hb_font_t to use
+ * @buffer: #hb_buffer_t buffer to work upon
+ *
+ * Begins
```
khaledhosny: Begins what?

src/hb-ot-layout.cc
```diff
@@ -227,6 +272,14 @@ hb_ot_layout_has_glyph_classes (hb_face_t *face)
...
 /**
  * hb_ot_layout_get_glyph_class:
+ * @face: The #hb_face_t to work on
+ * @glyph: The #hb_codepoint_t code point to query
+ *
+ * Fetches the GDEF class of the requested glyph in the specified face.
+ *
+ * Return value: Glyph class (UnclassifiedGlyph, BaseGlyph, LigatureGlyph, MarkGlyph,
+ *               ComponentGlyph) of the given code point in the GDEF table of the
+ *               face.
```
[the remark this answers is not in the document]
n8willis: OK.

src/hb-ot-layout.cc
```diff
@@ -442,6 +575,19 @@ hb_ot_layout_table_select_script (hb_face_t      *face,
   return false;
 }
 
+
+/**
+ * hb_ot_layout_table_get_feature_tags:
+ * @face: #hb_face_t to work upon
+ * @table_tag: HB_OT_TAG_GSUB or HB_OT_TAG_GPOS
+ * @start_offset: The offset into the feature-tag array to start returning
```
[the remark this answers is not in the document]
n8willis: I used similar wording for other start_offsets; happy to edit if there's a proposed change, but would prefer to keep them consistent ... so if the others are off-base too that is important.

src/hb-ot-layout.cc
```diff
@@ -375,12 +492,22 @@ hb_ot_layout_table_find_script (hb_face_t    *face,
...
 
+/**
+ * hb_ot_layout_table_choose_script:
+ * @face: #hb_face_t to work upon
+ * @table_tag: HB_OT_TAG_GSUB or HB_OT_TAG_GPOS
+ * @script_tags: Array of #hb_tag_t script tags
+ * @script_index: (out): The index of the requested script tag
+ * @chosen_script: (out): #hb_tag_t of the script tag requested
+ *
+ * Deprecated
```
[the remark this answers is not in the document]
n8willis: OK, looks like 2.0.0.

src/hb-ot-layout.cc
```diff
@@ -1507,6 +1941,22 @@ typedef enum {
   HB_OT_LAYOUT_BASELINE_ROMN = HB_TAG('r','o','m','n')
 } hb_ot_layout_baseline_t;
 
+
+/**
+ * hb_ot_layout_get_baseline:
+ * @font: The #hb_font_t to work upon
+ * @baseline: The #hb_ot_layout_baseline_t to query
+ * @direction: The #hb_direction_t text direction to use
```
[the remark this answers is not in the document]
n8willis: OK; added a note to that effect (if it passes muster). If it's not getting built, I would suggest not letting this block rolling out the rest; the whole comment could be removed without hurting anything IMO.

src/hb-ot-layout.cc
```diff
@@ -1105,8 +1526,19 @@ hb_ot_layout_position_finish_offsets (hb_font_t *font, hb_buffer_t *buffer)
   OT::GPOS::position_finish_offsets (font, buffer);
 }
 
+
 /**
  * hb_ot_layout_get_size_params:
+ * @face: #hb_face_t to work upon
+ * @design_size: (out): The design size of the face
+ * @subfamily_id: (out): The identifier of the face within the font subfamily
+ * @subfamily_name_id: (out): The  ‘name’ table name ID of the face within the font subfamily
```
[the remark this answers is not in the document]
n8willis: I'm pretty sure that's how the spec defines it, and that it's just a weird usage of 'subfamily' here (each subfamily member being different only in size) that didn't catch on.

Tried to elucidate that more clearly in the comment, and added a link to the doc.

src/hb-ot-layout.cc
n8willis: Good question :) -- I had meant to leave this one blank (as with a couple of other descriptors I need to tackle separately). Cleared it out for now.

src/hb-ot-layout.cc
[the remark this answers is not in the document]
n8willis: @khaledhosny Any comment on whether the new version is better?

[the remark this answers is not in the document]
khaledhosny: I guess so.

src/hb-ot-layout.cc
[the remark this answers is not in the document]
khaledhosny: My understanding is that, the `start_offset` is the first item to start fetching. So say the `GSUB` table has 23 features, but the client code does not know the count and to avoid allocating in the heap, it uses a stack allocated array of length, say 5, and in the first call it sets `start_offset` to zero and `feature_count` to 5, in the next call `start_offset` is 5, and so on until it fetches all the features.

We need also to document the return value.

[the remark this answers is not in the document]
n8willis: Okay; makes sense. So, like here: https://github.com/harfbuzz/harfbuzz/blob/fa3ebf845fcffa827600fc4ca9fdde3aaa99f4c5/src/hb-ot-math.cc#L168 (looks like the most detailed occurrence, based on my searching)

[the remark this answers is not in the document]
khaledhosny: Yes.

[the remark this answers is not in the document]
n8willis: OK; I propagated that wording style and also inserted (allow-none) and (array length=) annotations to the getter functions that use the same _offset mechanism.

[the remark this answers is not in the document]
n8willis: Latest commit also relocates the comment in hb_ot_layout_feature_get_characters regarding worrying about char_count's output and input being equal from the parameter block to a proper note, with a bit of rewording. I'm undecided if that ought to stay (or even be replicated elsewhere); I'm happy with whatever approach others prefer.

PR #1628, opened 2019-03-26
remove deprecated variable glyphs_deprecated

qxliu76:
from hb-subset-plan.hh/.cc files


garretrieger:
Failing bots are due to a 404 trying to get packages, not related to this change.

Issue #1631, opened 2019-03-28
USE overrides for Grantha and Tirhuta visargas


punchcutter:
Currently both Grantha and Tirhuta are listed in Unicode Script Extensions to accept udatta and anudatta marks, but in both cases the visargas are classified as VMPst so they block any following udatta and anudatta marks. This is the same situation as Grantha anusvara in https://github.com/harfbuzz/harfbuzz/issues/1035 
Both 0x11303 and 0x114C1 should be classified as VMAbv to allow for this.

PR #1632, opened 2019-03-28
Override USE category for Grantha and Tirhuta visargas to allow udatta and anudatta

punchcutter:
https://github.com/harfbuzz/harfbuzz/issues/1631


behdad:
Thanks.  Would be nice to add tests as well.

punchcutter:
Yes, I've been meaning to add a few other tests as well. I'll try to get something in here soon.

behdad:
Thanks!

Richard57:
What was the justification for this change?  It seems that Devanagari (dev2, deva?) has the wrong order.  See https://unicode.org/mail-arch/unicode-ml/y2019-m12/0046.html et seq for a discussion on the Unicode list that extends to this month.

PR #1636, opened 2019-03-28
 Add harfbuzzjs build configuration 

### LINKED ISSUE #1633: A WebAssembly version of HarfBuzz
Hi guys, I am developing a free web-based photo editor www.Photopea.com , which is used by around 100 000 people a day. It lets people do image editing, including inserting text into a picture.

As there is no sufficient OpenType parser and layout engine in Javascript, I made my own called [Typr.js](https://github.com/photopea/Typr.js). It is quite advanced and can handle e.g. Arabic text. I also use [this JS implementation of BIDI algorithm](https://github.com/bbc/unicode-bidirectional). 

As more and more people use Photopea, I have to extend Typr.js . Currently, I am adding the support for Urdu and Khmer layout. I am often staring at OpenType specification for 5 - 10 hours, without writing a single line of code, only trying to understand what they mean. I would be more than happy to drop Typr.js and use an alternative, if there was any.

Would you be able to provide a WebAssembly version of your library to the public, while documenting it and maintaining it? I am ready to pay 5k - 10k USD for it. It is also important, that the library is not too large (e.g. 150-200kB zipped), as every person has to download it when starting Photopea.

ebraminio:
Fixes #1633

The demo uses virtually all the important logic of HarfBuzz and the .wasm binary size is 710kb and 215kb zipped.

Lets see how we can have it cleanly here.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,6 +85,10 @@ if (HB_CHECK)
   endif ()
 endif ()
 
+# https://github.com/harfbuzz/harfbuzzjs/
+# emcmake cmake -Bwasmbuild -H. -GNinja -DHB_HARFBUZZJS=1 && ninja -Cwasmbuild
+option(HB_HARFBUZZJS "Enable our opinionated harfbuzzjs related build optimization" OFF)
+
 include_directories(AFTER
   ${PROJECT_SOURCE_DIR}/src
   ${PROJECT_BINARY_DIR}/src
@@ -293,6 +297,16 @@ if (HB_HAVE_GRAPHITE2)
   list(APPEND THIRD_PARTY_LIBS ${GRAPHITE2_LIBRARY})
 
   mark_as_advanced(GRAPHITE2_INCLUDE_DIR GRAPHITE2_LIBRARY)
+endif ()
+
+if (HB_HARFBUZZJS)
+  # we won't need multithread support in harfbuzzjs
+  add_definitions(-DHB_NO_MT)
+
+  add_definitions(-DHB_NO_OPTIONS -DHB_NO_OT_FONT_CFF -DHB_NO_NAME_TABLE_AAT
+    -DHB_NO_OT_FONT_BITMAP -DHB_NO_OT_SHAPE_COMPLEX_HEBREW_FALLBACK
+    -DHB_NO_OT_SHAPE_FALLBACK -DHB_NO_OT_SHAPE_COMPLEX_THAI_FALLBACK
+    -DHB_NO_OT_LAYOUT_BLACKLIST -DHB_NO_OT_SHAPE_COMPLEX_VOWEL_CONSTRAINTS)
 endif ()
 
 if (HB_BUILTIN_UCDN)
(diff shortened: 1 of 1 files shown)


test-wasm.html
```diff
@@ -0,0 +1,102 @@
...
+    var blob = module._hb_blob_create(buffer, font.byteLength, 0/*FIXME*/, 0, 0);
+    // module._hb_blob_get_length(blob), module._hb_face_count(blob)
+    var face = module._hb_face_create(blob, 0);
+    var font = module._hb_font_create(face);
+
+    var buffer = module._hb_buffer_create();
+    [...'Your text to shape!'].forEach((x, i) => {
+      module._hb_buffer_add(buffer, x.codePointAt(0), i);
+    });
+    module._hb_buffer_set_content_type(buffer, 1/*HB_BUFFER_CONTENT_TYPE_GLYPHS*/);
```
behdad: This line is not needed.

test-wasm.html
```diff
@@ -0,0 +1,102 @@
...
+
+    var blob = module._hb_blob_create(buffer, font.byteLength, 0/*FIXME*/, 0, 0);
+    // module._hb_blob_get_length(blob), module._hb_face_count(blob)
+    var face = module._hb_face_create(blob, 0);
+    var font = module._hb_font_create(face);
+
+    var buffer = module._hb_buffer_create();
+    [...'Your text to shape!'].forEach((x, i) => {
+      module._hb_buffer_add(buffer, x.codePointAt(0), i);
+    });
```
khaledhosny: How does this handle UTF-16 surrogate pairs (JavaScript used to do funny things with these). Why not use one of `hb_buffer_add_utfXX` functions?

test-wasm.html
ebraminio: Fails with `uncaught exception: abort("Assertion failed: content_type == HB_BUFFER_CONTENT_TYPE_UNICODE || (!len && content_type == HB_BUFFER_CONTENT_TYPE_INVALID), at: ../src/hb-buffer.cc,642,guess_segment_properties"). Build with -s ASSERTIONS=1 for more info.` if I don't.

behdad: Actually.  Humm.  You setting it to CONTENT_TYPE_GLYPHS?!!!  Your comment is wrong.  1 is CONTENT_TYPE_UNICODE.  The question is, why did your buffer have CONTENT_TYPE_GLYPHS in it otherwise?  Did you not reset the buffer after calling shape on it once?

ebraminio: Hmm, I don't call it more than once, let me use hb_buffer_add_utf8 to resolve this :)

test-wasm.html
ebraminio: Oh I will

ebraminio: Well I just wanted to have a working example

test-wasm.html
behdad: Oh you're right.  It's the `add_utfX` that set the content type to Unicode.  Nvm.

test-wasm.html
ebraminio: Gone. [...str] syntax is aware of UTF-16 thing issue.

test-wasm.html
```diff
@@ -0,0 +1,107 @@
...
+    module.HEAPU8.set(new Uint8Array(font), buffer);
+
+    var blob = module._hb_blob_create(buffer, font.byteLength, 0/*FIXME*/, 0, 0);
+    // module._hb_blob_get_length(blob), module._hb_face_count(blob)
+    var face = module._hb_face_create(blob, 0);
+    var font = module._hb_font_create(face);
+
+    var buffer = module._hb_buffer_create();
+    {
+      var text = utf8Encoder.encode('Your text to shape!');
```
khaledhosny: Since JS uses UTF-16 internally, why not use that instead of re-encoding the text.

ebraminio: Just tried and didn't work for some reason, I don't see it as a performance bottleneck as JS-WASM is more of bottleneck itself but you are right it was nice to have.

ebraminio: `Chrome 54 (ticket) and Opera 41, no other encoding types are available other than utf-8, in order to match the spec.` https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder so JavaScript doesn't provide utf-16 even the fact it may use it itself.

test-wasm.html
```diff
@@ -0,0 +1,109 @@
...
+    var blob = module._hb_blob_create(fontBuffer, fontBlob.byteLength, 2/*HB_MEMORY_MODE_WRITABLE*/, 0, 0);
+    var face = module._hb_face_create(blob, 0);
+    var font = module._hb_font_create(face);
+
+    var buffer = module._hb_buffer_create();
+    {
+      var text = utf8Encoder.encode("سلام!");
+      var text_ptr = module._malloc(text.byteLength);
+      module.HEAPU8.set(text, text_ptr);
+      module._hb_buffer_add_utf8(buffer, text_ptr, text.byteLength, 0, -1);
```
khaledhosny: The remaining `-1` need to be `text.byteLength` as well.

ebraminio: Done. Thanks

ebraminio:
Maybe we should move the build configuration over https://github.com/harfbuzz/harfbuzzjs but as it wasn't so much of disturbance here I merged the configuration for now.

PR #1637, opened 2019-03-28
[ci] Use only CircleCI for macOS

### LINKED ISSUE #1609: Fixing or getting rid of Travis macOS bot
Travis's macOS bot fails and apparently because it can't find glib-2.0 even the fact it is installed. I say let's get rid of the bot now that we have a number of CircleCI based macOS bots. The advantage to CircleCI is we can log into it every time we face an issue but Travis doesn't provide feature AFAIK. It will also simplify our Travis's Linux CI.

ebraminio:
Fixes #1609 

I think this makes macos-10.14.3-aat-fonts on par with our current Travis CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,8 +27,9 @@ jobs:
       xcode: "10.2.0"
     steps:
       - checkout
-      - run: HOMEBREW_NO_AUTO_UPDATE=1 brew install wget autoconf automake libtool pkg-config ragel freetype glib cairo
-      - run: ./autogen.sh --with-freetype --with-glib --with-gobject --with-cairo
+      - run: HOMEBREW_NO_AUTO_UPDATE=1 brew install wget autoconf automake libtool pkg-config ragel freetype glib cairo icu4c
+      - run: brew link --force icu4c
+      - run: export PKG_CONFIG_PATH="/usr/local/opt/icu4c/lib/pkgconfig" && ./autogen.sh --with-freetype --with-glib --with-gobject --with-cairo --with-icu --with-coretext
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -27,8 +27,9 @@ jobs:
       xcode: "10.2.0"
     steps:
       - checkout
-      - run: HOMEBREW_NO_AUTO_UPDATE=1 brew install wget autoconf automake libtool pkg-config ragel freetype glib cairo
-      - run: ./autogen.sh --with-freetype --with-glib --with-gobject --with-cairo
+      - run: HOMEBREW_NO_AUTO_UPDATE=1 brew install wget autoconf automake libtool pkg-config ragel freetype glib cairo icu4c
+      - run: brew link --force icu4c
+      - run: export PKG_CONFIG_PATH="/usr/local/opt/icu4c/lib/pkgconfig" && ./autogen.sh --with-freetype --with-glib --with-gobject --with-cairo --with-icu --with-coretext
```
khaledhosny: `PKG_CONFIG_PATH="/usr/local/opt/icu4c/lib/pkgconfig" ./autogen.sh …` should work as well.

khaledhosny: Actually you shouldn’t need it with `brew link --force icu4c`. Any way Travis macOS was working, so I don’t see what is the benefit of removing it.

khaledhosny: It was also using `--with-graphite2`, we don’t seem to have any macOS CI builds with Graphite now.

ebraminio: #1638

PR #1638, opened 2019-03-28
[ci] Tweak macos and psvita bots

ebraminio:
* Add --with-graphite2 to macOS
* Create a dummy ragel binary


behdad:
I think we should remove the check for ragel from pkgconfig now that we include the output in git.

ebraminio:
It should work even now without ragel as the bot was ok without but timestamps just sometimes confuses the bot that's why I added the dummy script anyway.

Issue #1602, opened 2019-03-02
Harfbuzz's usage of Standard C bsearch may be a potential performance bottleneck


jwtowner:
I was optimizing my custom Unicode Bidirectional layout engine and I happened to come across a major performance bottleneck.

I'm using the UCDN library that's bundled with Harfbuzz, that way I don't bloat up my binaries with an extra set of Unicode tables. The layout engine of course calls ucdn_paired_bracket and ucdn_paired_bracket_type for paired bracket detection. It turns out these two function calls were taking up over half of the time slice of an entire bidi layout run, and the reason being is that they're implemented in terms of the libc bsearch. Since I'm working in C++, I was able to switch UCDN over to use std::lower_bound behind the scenes and as a result more than doubled the performance of my entire bidi layout engine!

Harfbuzz doesn't use these two functions from UCDN, but I think it does use ucdn_compose and ucdn_mirror, both of which also use libc bsearch. While most of Harfbuzz uses its own custom binary search functions, there is at least one spot in hb_ot_tag_from_language that does use the libc bsearch.

Anyway, it would be nice if both Harfbuzz and the UCDN library packaged with Harfbuzz were fixed up to always use custom binary searches that inline the inner loop and comparisons, instead of relying on the libc bsearch with all of its call overhead. Some low-hanging fruit for anyone who wants it.

khaledhosny:
UCDN is maintained here https://github.com/grigorig/ucdn, the copy in HarfBuzz source tree is verbatim with no changes.

behdad:
@dscorbett can you switch to hb_array_t's bsearch in hb-ot-tag.cc?

dscorbett:
Do you mean reimplementing `ot_languages` as a `hb_sorted_array_t`? Sure, I can do that.

behdad:
You don't need to reimplement `ot_languages`.  Just when you need to search, wrap it into one: `hb_sorted_array_t (ot_languages).bsearch...` should work.

behdad:
Fixed the instance in HarfBuzz itself.  Would be interesting to measure the effect of the possible UCDN speedup.

ebraminio:
There is no longer libc's bsearch in harfbuzz https://github.com/harfbuzz/harfbuzz/commit/33d38e793e2e4882337e5f42fbbae7d00d343940 yay

PR #1639, opened 2019-03-29
Search ot_languages with hb_sorted_array_t.bsearch

dscorbett:
As @behdad requested in #1602, this pull request changes `hb_ot_tags_from_language` to search `ot_languages` using the custom `hb_sorted_array_t.bsearch` instead of the standard C `bsearch`. This deserves a careful review because I haven’t had much experience with C++ templates.


behdad:
Thanks.  I ended up doing it without passing cmp down to bsearch.

PR #1645, opened 2019-03-31
[iter] Add hb_reduce

ebraminio:
These will be nice to have also:
* <s>Being able to have accumulator in a different type</s> (done)
* <s>hb_reduce_right (or hb_reduce_end?)</s> (not needed, we should have reverse instead)
* <s>Make it work somehow without initial value also</s> (possible but a respective default type also is needed) 

This alternatively can be named fold, accumulate, aggregate, compress, or inject [also](https://en.wikipedia.org/wiki/Fold_(higher-order_function)) and things like sum/min/max/average can be implemented upon it.

### DIFF
--- a/src/hb-iter.hh
+++ b/src/hb-iter.hh
@@ -344,6 +344,34 @@ static const struct
   { return hb_filter_iter_factory_t<Pred, Proj> (p, f); }
 } hb_filter HB_UNUSED;
 
+template <typename Redu, typename TValue>
+struct hb_reduce_t
+{
+  hb_reduce_t (Redu r, TValue init_value) : r (r), init_value (init_value) {}
+
+  template <typename Iter,
+	    hb_enable_if (hb_is_iterator (Iter))>
+  TValue
+  operator () (Iter it) const
+  {
+    TValue value = init_value;
+    for (; it; ++it)
+      value = r (*it, value);
+    return value;
+  }
+
+  private:
+  Redu r;
+  TValue init_value;
+};
+static const struct
+{
+  template <typename Redu, typename TValue> hb_reduce_t<Redu, TValue>
+  operator () (Redu&& r, TValue init_value) const
+  { return hb_reduce_t<Redu, TValue> (r, init_value); }
+} hb_reduce HB_UNUSED;
+
+
 /* hb_zip() */
 
 template <typename A, typename B>
--- a/src/hb-iter.hh
+++ b/src/hb-iter.hh
@@ -344,31 +344,33 @@ static const struct
   { return hb_filter_iter_factory_t<Pred, Proj> (p, f); }
 } hb_filter HB_UNUSED;
 
-template <typename Redu, typename TValue>
+template <typename Redu, typename InitT>
 struct hb_reduce_t
 {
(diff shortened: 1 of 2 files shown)


behdad:
Thanks.  I have a long list of things to do with the iter stuff.  You can follow in https://github.com/harfbuzz/harfbuzz/issues/1648

behdad:
Thanks.  Easy enough to do without initial value.  let's do that as well?  Just an overload.

ebraminio:
> Easy enough to do without initial value.

Can't see how to do it without specifying a default type however.

src/hb-iter.hh
```diff
@@ -344,6 +344,38 @@ static const struct
...
+};
+static const struct
+{
+  template <typename Redu, typename TValue = int> hb_reduce_t<Redu, TValue>
+  operator () (Redu&& r, TValue init_value = 0) const
+  { return hb_reduce_t<Redu, TValue> (r, init_value); }
+
+  template <typename Redu, typename TValue = int> hb_reduce_t<Redu&, TValue>
+  operator () (Redu *r, TValue init_value = 0) const
+  { return hb_reduce_t<Redu&, TValue> (*r, init_value); }
```
behdad: I don't think we need a separate overload for pointer.  For now remove.  When we address that issue across the board, I think we just need to add a `hb_deref_pointer` to the existing implementation.

behdad:
Humm, yeah no this is not good.  I think we might need to use two different types for the two cases.  I don't mind reverting back to the simple case you had.

Also, `reduce_right` isn't quite in scope for us.  That should be done with something like a hb_reverse() in between.  Or maybe only if iter is random-access.  Anyway, no need right now.

behdad:
Thanks.  One more round and we'd be done: the accumulation type is not necess the same as initial type.  Accumulation type should be `decltype(hb_declval (Redu) (hb_declval (InitType), Iter:__item_type__))`.  Something like that.  Do you agree?

behdad:
Eg. init might be integer, but acummulation type should be float if the reducer returns float.

behdad:
> So what happens when the array is empty. Accumulation type and initial type should be in the same type but the iterator values can be in a different type, what the second test demonstrates.

You assign to accumulation type the initial value and let the type system figure out the conversion.

ebraminio:
> You assign to accumulation type the initial value and let the type system figure out the conversion.

That won't do the thing I'm seeking in the second test IIUC.

behdad:
> That won't do the thing I'm seeking in the second test IIUC.

How does it not?  If iterator is empty, you end up just returning the initial type.  I'm not sure what you are talking about.

ebraminio:
Ok. Definitely makes sense. I am struggling on how to do it :)

behdad:
I wrote it down for you ;).

I suggest this:
```
template <AccT = decltype (...)>
operator () (...)
```

behdad:
Ie. accumulation type is not known until the operator(), since it depends on the iterator item type.

ebraminio:
Done

behdad:
Looks great.  Thanks :).

Issue #1504, opened 2018-12-21
Wrong glyph for Initial and Medial form of Arabic Letter (ئ) YEH WITH HAMZA  above


mowais786:
I am using harfbuzz 2.2.0 wich icu63, icu layout along with icu-le-hb bridge. Initial and medial form of 0x626 are not returned correctly. Base glyph is returned.
Here is the sample input and wrong output
المكتب الرئيسي
المكتب الرئ يسي (Intentionally added space to show the wrong output)

This was working fine with icu56 and layoutengine

mowais786:
This happens with all fonts, even with Arial Unicode. I will try to produce some short code to reproduce. I can give you some hint. I debugged code, it generates fonts fall back map, which contains all four glyph but u.format2.get_coverage returns NOT_COVERED for initial and medial glyph and works fine for End glyph. While building the range for 0x626 it seems that RangeRecord is set start =  0x626 and end = 0.
Hope this will help to identify the issue. In the meanwhile I will try to produce some small code example to reproduce the issue.

gilbahat:
we are also having issues with the same glyph (0x0626) but it appears to be a shaper issue for us AND it doesn't work with icu52 or icu63. it works with coretext but fails for ot or fallback shapers:
gilbahat@pasture:~/magisto$ cat 2.txt
تعبئة
gilbahat@pasture:~/magisto$ hb-shape --verbose --trace --font-file=unifont.ttf --shapers=ot --text-file 2.txt
1: (تعبئة)
1: <U+062A,U+0639,U+0628,U+0626,U+0629>
1: [uniFE94=4+512|uni0626=3+512|uniFE92=2+512|uniFECC=1+512|uniFE97=0+512]
gilbahat@pasture:~/magisto$ hb-shape --verbose --trace --font-file=unifont.ttf --shapers=fallback --text-file 2.txt
1: (تعبئة)
1: <U+062A,U+0639,U+0628,U+0626,U+0629>
1: [uni0629=4+512|uni0626=3+512|uni0628=2+512|uni0639=1+512|uni062A=0+512]
gilbahat@pasture:~/magisto$ hb-shape --verbose --trace --font-file=unifont.ttf --shapers=coretext --text-file 2.txt
1: (تعبئة)
1: <U+062A,U+0639,U+0628,U+0626,U+0629>
1: [uniFE94=4+512|uniFE8C=3+512|uniFE92=2+512|uniFECC=1+512|uniFE97=0+512]

[...]

PR #1646, opened 2019-03-31
Don't skip setting the .end field of the first range

jfkthame:
Fixes a bug in CoverageFormat2::serialize whereby the first range
was not serialized correctly if it consists of only a single glyph ID.
This broke shaping of U+0626 in the Arabic fallback shaper, because it
is not found in the coverage table of the 'init' and 'medi' lookups.

Also fix similar bug in ClassDefFormat2::serialize, noted during code
inspection (I haven't observed a case that was actually affected by
this, but it looks broken).

Fixes https://github.com/harfbuzz/harfbuzz/issues/1504

### DIFF
--- a/src/hb-ot-layout-common.hh
+++ b/src/hb-ot-layout-common.hh
@@ -921,12 +921,13 @@ struct CoverageFormat2
     {
       if (glyphs[i - 1] + 1 != glyphs[i])
       {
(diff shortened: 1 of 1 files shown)


behdad:
Thanks.  Would be nice to add test as well.

PR #1647, opened 2019-03-31
Fix lookup subtables timeout

ebraminio:
Don't know if is appropriate way to fix the issue, wanted to learn how to fix such issues anyway.


behdad:
Can you describe what the timeout is?  Is it just 64k subtables that are all tried?

behdad:
I think right fix is to check and bail out of loop if we hit the max-op limit already.  A couple of such checks in loops should do.

ebraminio:
Have a look if this sounds better, there is no access to buffer on all the uses so another max_ops should be defined anyway and we should name it something I guess.

khaledhosny:
Please add a full URL in the commit messages, so readers don’t need to spend several minutes finding the relevant issue.

ebraminio:
Done

ebraminio:
This is however not enough, a malicious font can easily hit this with some other trick

behdad:
I guess limiting number of subtables to something like 64 is fine.  I'm happy with that.

We can fix apply() and sanitize() etc using their max-ops.  But I don't like that that would leave other users of dispatch vulnerable.

Alternatively, we can leave dispatch() as is and put the max-ops check in the apply_context_t / sanitize_context_t's callbacks.

PR #1656, opened 2019-04-05
Fix MSVC C4068 warning

maksqwe:
Fix "warning C4068: unknown pragma"


ebraminio:
So it sees `#ifndef` as an unknown pragma? That's weird :/

maksqwe:
No, "#pragma GCC" is unknown for MSVC compiler.

ebraminio:
Aha now I see. So lets have this. Thank you :)

PR #1661, opened 2019-04-08
[Docs] Add gtk-doc comments for hb-aat-layout

n8willis:
This adds gtk-doc inline comments documenting public AAT layout API functions and makes light additions to enum documentation (like denoted deprecations).


src/hb-aat-layout.cc
```diff
@@ -46,11 +46,16 @@
  * @short_description: Apple Advanced Typography Layout
  * @include: hb-aat.h
  *
- * Functions for querying OpenType Layout features in the font face.
+ * Functions for querying OpenType Layout features in the font face. 
+ *
+ * HarfBuzz supports all of the AAT tables used to implement shaping. Other
+ * AAT tables and their associated features are not supported.
```
ebraminio: HarfBuzz doesn't support `just` which is related to shaping but supports `lcar`, `feat` and `sbix`. There are of course tables that are not hooked yet like `fdsc` (which Apple itself don't care about, we may remove it also I guess) and `bsln` (`BASE` also isn't yet)

ebraminio: `just` isn't in that use in Apple's fonts also of course.

n8willis: I get that wording from Behdad.... `lcar` seems to fall under the same purview as I see it; reading it provides compatibility with what we say about OpenType shaping being important for caret placement.

Also, I think "supported" is important here -- AIUI, we're just wanting to declare that the purpose is to support shaping; anything else is not be considered a guarantee. (sbix being the domain of hb-ot-color, etc). 

Do you (or @behdad ) have some other wording in mind you would recommend for that sentence?

ebraminio: Given your explanation I feel I understand the intention of course, this corner case didn't make sense just at first so may change of wording is needed or not.

ebraminio:
> We don't support justification, so not supporting it in AAT does not need to be pointed out IMO.

It was more about the wording, but not to mention `just` or `bsln`.

src/hb-aat-layout.cc
ebraminio: > anything else is not be considered a guarantee

I think we like to work on any features AAT has and we don't as I was involved in `lcar` and `sbix` and like to make it on par with OT. May it isn't guaranteed, like the rest of the project, but I think we like to get feedback on not supported things anytime, like feature setting using AAT's convention, that may fall into shaping scope actually. This is anyway some kind of bike-shedding at best so sorry for taking your time.

n8willis: No problem at all. I do think we ought to be clear, but perhaps this section reference is not the ideal spot. Perhaps we can add more detail about what is accessible in the 'user manual' page on AAT.

src/hb-aat-layout.cc
```diff
@@ -46,11 +46,16 @@
  * @short_description: Apple Advanced Typography Layout
  * @include: hb-aat.h
  *
- * Functions for querying OpenType Layout features in the font face.
+ * Functions for querying OpenType Layout features in the font face.
```
khaledhosny: This line looks like it was copied from somewhere else, since this file has AAT not OpenType functions.

src/hb-aat-layout.cc
```diff
@@ -201,9 +217,15 @@ hb_aat_layout_compile_map (const hb_aat_map_builder_t *mapper,
 
 /*
  * hb_aat_layout_has_substitution:
- * @face:
+ * @face: #hb_face_t to work upon
+ *
+ * Tests whether the specified face includes any substitutions in the
+ * `morx` table.
```
khaledhosny:
```suggestion
 * `morx` or `mort` tables.
```

src/hb-aat-layout.cc
```diff
@@ -262,9 +284,15 @@ hb_aat_layout_remove_deleted_glyphs (hb_buffer_t *buffer)
...
+ * in the `kerx` table.
+ *
+ * <note>Note: does not examine the `GPOS` table.</note>
+ *
+ * Return value: true if data found, false otherwise
  *
- * Returns:
  * Since: 2.3.0
  */
 hb_bool_t
```
khaledhosny: While at it, may be document the `hb_aat_layout_position()` function below?

src/hb-aat-layout.cc
```diff
@@ -289,9 +317,13 @@ hb_aat_layout_position (const hb_ot_shape_plan_t *plan,
...
+ *
+ * Tests whether the specified face includes any tracking information
+ * in the `trak` table.
+ *
+ * Return value: true if data found, false otherwise
  *
- * Returns:
  * Since: 2.3.0
  */
 hb_bool_t
```
khaledhosny: Document the `hb_aat_layout_track()` function below as well?

src/hb-aat-layout.cc
```diff
@@ -341,10 +376,12 @@ hb_aat_layout_get_feature_types (hb_face_t                    *face,
 
 /**
  * hb_aat_layout_feature_type_get_name_id:
- * @face: a face object
- * @feature_type: feature id
+ * @face: #hb_face_t to work upon
+ * @feature_type: The #hb_aat_layout_feature_type_t of the requested feature type
  *
- * Return value: Name ID index
+ * Fetches the name ID of the specified feature type in the face's 'name' table.
```
khaledhosny:
```suggestion
 * Fetches the name ID of the specified feature type in the face's `name` table.
```
For consistency with other occurrences of table names.

src/hb-aat-layout.cc
```diff
@@ -341,10 +376,12 @@ hb_aat_layout_get_feature_types (hb_face_t                    *face,
...
+ *
+ * Return value: Name ID index of the requested feature type
```
khaledhosny: I think “index” is redundant, given that it is what “ID” stands for, no?

src/hb-aat-layout.h
```diff
@@ -37,7 +37,49 @@ HB_BEGIN_DECLS
...
+ * @HB_AAT_LAYOUT_FEATURE_TYPE_ITALIC_CJK_ROMAN
+ * @HB_AAT_LAYOUT_FEATURE_TYPE_CASE_SENSITIVE_LAYOUT
+ * @HB_AAT_LAYOUT_FEATURE_TYPE_ALTERNATE_KANA
+ * @HB_AAT_LAYOUT_FEATURE_TYPE_STYLISTIC_ALTERNATIVES
+ * @HB_AAT_LAYOUT_FEATURE_TYPE_CONTEXTUAL_ALTERNATIVES
+ * @HB_AAT_LAYOUT_FEATURE_TYPE_LOWER_CASE
+ * @HB_AAT_LAYOUT_FEATURE_TYPE_UPPER_CASE
+ * @HB_AAT_LAYOUT_FEATURE_TYPE_LANGUAGE_TAG_TYPE
+ * @HB_AAT_LAYOUT_FEATURE_TYPE_CJK_ROMAN_SPACING_TYPE
+ * @_HB_AAT_LAYOUT_FEATURE_TYPE_MAX_VALUE: 0x7FFFFFFFu
```
khaledhosny: I don’t think you need to duplicate them here, they show up already in the generated docs https://harfbuzz.github.io/harfbuzz-hb-aat-layout.html#hb-aat-layout-feature-type-t

src/hb-aat-layout.h
```diff
@@ -90,7 +132,337 @@ typedef enum
...
+ * @HB_AAT_LAYOUT_FEATURE_SELECTOR_UPPER_CASE_SMALL_CAPS
+ * @HB_AAT_LAYOUT_FEATURE_SELECTOR_UPPER_CASE_PETITE_CAPS
+ *
+ * Selectors for #HB_AAT_LAYOUT_FEATURE_TYPE_CJK_ROMAN_SPACING_TYPE
+ * @HB_AAT_LAYOUT_FEATURE_SELECTOR_HALF_WIDTH_CJK_ROMAN
+ * @HB_AAT_LAYOUT_FEATURE_SELECTOR_PROPORTIONAL_CJK_ROMAN
+ * @HB_AAT_LAYOUT_FEATURE_SELECTOR_DEFAULT_CJK_ROMAN
+ * @HB_AAT_LAYOUT_FEATURE_SELECTOR_FULL_WIDTH_CJK_ROMAN
+ *
+ * @_HB_AAT_LAYOUT_FEATURE_SELECTOR_MAX_VALUE: 0x7FFFFFFFu
```
khaledhosny: If you are going to list them explicitly, then the ones starting with underscore are “private” and should not be listed.

src/hb-aat-layout.h
```diff
@@ -90,7 +132,337 @@ typedef enum
 
 /**
  * hb_aat_layout_feature_selector_t:
+ * @HB_AAT_LAYOUT_FEATURE_SELECTOR_INVALID: 0xFFFF
```
khaledhosny: I don’t think we need to explicitly document the value of the entry here.

src/hb-aat-layout.h
```diff
@@ -446,6 +818,13 @@ typedef struct hb_aat_layout_feature_selector_info_t
   unsigned int				reserved;
 } hb_aat_layout_feature_selector_info_t;
 
+/**
+ * HB_AAT_LAYOUT_NO_SELECTOR_INDEX
```
khaledhosny: Should there be a colon at the end of this line?

src/hb-aat-layout.cc
n8willis: I think ID stands for identifier (here). At https://docs.microsoft.com/en-us/typography/opentype/spec/name they always say "name ID" -- as with "script ID"/"language ID"/etc -- so I was just trying for consistency.

khaledhosny: Anyway, index here as at best confusing. I’d understand this to mean the index of a name table entry (i.e. the *n*th entry), but here it is just returning name IDs and there can be multiple entries in the name table for the same ID.

behdad: ID is for identifier.  This is NOT an index.

khaledhosny:
@behdad I already reviewed this :)

behdad:
@khaledhosny Thanks.  Sorry about that.  I just am deferring all docs to you to approve. :)

src/hb-aat-layout.cc
```diff
@@ -79,13 +79,17 @@ AAT::hb_aat_apply_context_t::set_ankr_table (const AAT::ankr *ankr_table_)
  * @short_description: Apple Advanced Typography Layout
  * @include: hb-aat.h
  *
- * Functions for querying OpenType Layout features in the font face.
+ * Functions for querying OpenType Layout features in the font face.
```
khaledhosny, line 82:
```suggestion
 * Functions for querying AAT Layout features in the font face. 
```

src/hb-aat-layout.cc
```diff
@@ -336,10 +370,12 @@ hb_aat_layout_get_feature_types (hb_face_t                    *face,
...
  * hb_aat_layout_feature_type_get_name_id:
- * @face: a face object
- * @feature_type: feature id
+ * @face: #hb_face_t to work upon
+ * @feature_type: The #hb_aat_layout_feature_type_t of the requested feature type
  *
- * Return value: Name ID index
+ * Fetches the name ID of the specified feature type in the face's `name` table.
+ *
+ * Return value: Name ID index of the requested feature type
```
khaledhosny, line 378:
```suggestion
 * Return value: Name ID of the requested feature type
```

khaledhosny:
I resolve the merge conflict and fixed the syntax a bit.

khaledhosny (approved):

alerque (approved):

PR #1668, opened 2019-04-16
[valgrind] Use libtool to not catch bash related issues

ebraminio:


behdad:
Don't run `which libtool`. Use the one from our topbuilddir.

behdad:
Would also help not finding OS X's unrelated libtool.

test/fuzzing/run-shape-fuzzer-tests.py
```diff
@@ -68,11 +68,17 @@ def timeout(p, is_killed):
...
 valgrind = None
+libtool = None
 if os.environ.get('RUN_VALGRIND', ''):
 	valgrind = which ('valgrind')
 	if valgrind is None:
 		print ("""Valgrind requested but not found.""")
 		sys.exit (1)
+	LIBTOOL = os.environ.get('LIBTOOL')
+	if LIBTOOL:
+		libtool = LIBTOOL.split(' ')[-1]
```
ebraminio: Hack, as it was returning something '/bin/sh ../libtool' and I wanted the later part.

test/fuzzing/run-subset-fuzzer-tests.py
```diff
@@ -20,13 +39,32 @@
...
+		print ("""Valgrind requested but not found.""")
+		sys.exit (1)
+	LIBTOOL = os.environ.get('LIBTOOL')
+	if LIBTOOL:
+		libtool = LIBTOOL.split(' ')[-1]
+	else:
+		libtool = which ('libtool')
+
+# Disable it for now
+valgrind = None
```
ebraminio: Disabled as it fails with this
image

test/fuzzing/run-shape-fuzzer-tests.py
behdad: Why?  Just leave it as is.

behdad: I mean, split it and prepend that to the args list...

ebraminio:
Let's merge this for now I think

behdad:
Yeah, whatever.  If it works, sure.

behdad:
Open issue re removing python test runner?

behdad:
I guess if we want to catch timeouts, we would need some kind of test runner.

PR #1669, opened 2019-04-16
Comment out entries in ot_languages where fallback suffices

dscorbett:
Note that this changes the behavior of `hb_ot_tag_to_language` for unknown tags.

### DIFF
--- a/src/gen-tag-table.py
+++ b/src/gen-tag-table.py
@@ -895,11 +895,18 @@ def language_name_intersection (a, b):
 def get_matching_language_name (intersection, candidates):
 	return next (iter (c for c in candidates if not intersection.isdisjoint (get_variant_set (c))))
 
+def same_tag (bcp_47_tag, ot_tags):
+	return len (bcp_47_tag) == 3 and len (ot_tags) == 1 and bcp_47_tag == ot_tags[0].lower ()
+
 for language, tags in sorted (ot.from_bcp_47.items ()):
 	if language == '' or '-' in language:
 		continue
+	commented_out = same_tag (language, tags)
 	for i, tag in enumerate (tags, start=1):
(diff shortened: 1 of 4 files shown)


behdad:
> Note that this changes the behavior of `hb_ot_tag_to_language` for unknown tags.

Right...  

I see what you did (assuming it's ISO 639-3 and encoding it as is).  I'm fine with that.  However, wouldn't that fail if there's an existing ISO 639-3 with that code, that maps to some other tag?  You can just try mapping it forward and only do what you're doing if result is consistent.

dscorbett:
I’ll fix it.

dscorbett:
I think it’s fixed now.

behdad:
Thanks.

Humm. I don't see how your new code doesn't have the same problem.  I don't see what you changed actually.

Let's say, `hb_ot_tag_to_language()` called on `ACQ `.  We should NOT return `acq`, since that maps to `ARA `.  How is that handled?

dscorbett:
What I fixed was 'ALT ' and 'HND '. They correspond to the BCP 47 tags "alt" and "hnd", so they are commented out of `ot_languages`. "atv" and "hno" also correspond to 'ALT ' and 'HND ', so 'ALT ' and 'HND ' were therefore converted to "atv" and "hno". The fix was to add such tags to `hb_ot_ambiguous_tag_to_language`.

'ACQ ' will indeed not round-trip properly. I think that is okay since it is not a registered language system tag.

behdad:
> 'ACQ ' will indeed not round-trip properly. I think that is okay since it is not a registered language system tag.

Right.  But you just need to one `if (hb_ot_tag_from_languge (...) == ...) return` to fix that.  Right?

behdad:
In the end no one cares, so maybe just add a comment and we're done here.  Thanks!

dscorbett:
I’ve added a comment.

behdad:
Thank you.

PR #1674, opened 2019-04-18
fix oss-fuzz issue 14345

blueshade7:
If a bad Encoding has a single-byte code exceeding 0xFF, treat glyph as unencoded


src/hb-ot-cff1-table.hh
```diff
@@ -110,7 +110,8 @@ struct Encoding1 {
     {
       if (glyph <= ranges[i].nLeft)
       {
-	return (hb_codepoint_t)ranges[i].first + glyph;
+	hb_codepoint_t code = (hb_codepoint_t)ranges[i].first + glyph;
+	return (likely (code < 0x100)? code: CFF_UNDEF_CODE);
```
behdad: In both lines, space after the first `)`.

PR #1676, opened 2019-04-21
[Docs] Add gtkdoc comments for low-level core APIs

n8willis:
This fills in and adds GTK-Doc inline documentation for several lower-level APIs, namely hb-set, hb-map, hb-common, hb-blob, and hb-buffer. It also propagates some small changes in capitalization and punctuation to existing GTK-Doc comments to look better in the final generated HTML output.

### DIFF
--- a/src/hb-common.h
+++ b/src/hb-common.h
@@ -136,8 +136,8 @@ typedef union _hb_var_int_t {
 /**
  * hb_tag_t:
  * 
- * Data type for tag identifiers. Tags are arrays of four
- * bytes, each of which holds a character. 
+ * Data type for tag identifiers. Tags are four
+ * byte integers, each byte representing a character.
  *
  * Tags are used to identify tables, design-variation axes,
  * scripts, languages, font features, and baselines with
@@ -236,8 +236,8 @@ hb_direction_to_string (hb_direction_t direction);
  * HB_DIRECTION_IS_FORWARD:
  * @dir: #hb_direction_t to test
  *
- * Tests whether a text direction is forward. Requires
- * that the direction be valid.
+ * Tests whether a text direction moves forward (from left to right, or from
+ * top to bottom). Requires that the direction be valid.
  *
  **/
 #define HB_DIRECTION_IS_FORWARD(dir)	((((unsigned int) (dir)) & ~2U) == 4)
@@ -245,8 +245,8 @@ hb_direction_to_string (hb_direction_t direction);
  * HB_DIRECTION_IS_BACKWARD:
  * @dir: #hb_direction_t to test
  *
- * Tests whether a text direction is backward. Requires
- * that the direction be valid.
+ * Tests whether a text direction moves backward (from right to left, or from
+ * bottom to top). Requires that the direction be valid.
  *
  **/
 #define HB_DIRECTION_IS_BACKWARD(dir)	((((unsigned int) (dir)) & ~2U) == 5)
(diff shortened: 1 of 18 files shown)


src/hb-face.cc
```diff
@@ -278,13 +283,13 @@ hb_face_destroy (hb_face_t *face)
...
- * @face: a face.
- * @key:
- * @data:
- * @destroy:
- * @replace:
- *
+ * @face: A face object
+ * @key: The user-data key
+ * @data: A pointer to the user data
+ * @destroy: A callback to call when @data is not needed anymore
```
behdad: @destroy can be null.  Need both g-i annotations and maybe doc?  Donno.

n8willis: Can do!

n8willis: Are all destroy's nullable, no exceptions?

Also, looking at the GI annotations, it sounds like (optional) might be a better annotation for this than (nullable), but IDK.

behdad: Yes, all are optional. I never understood the difference between optional and nullable. What's your understanding?

n8willis: It seems to me that they want 'optional' to mean the client/user chooses not to use the parameter, but 'nullable' might be something that's required but sometimes has NULL as a valid value?

n8willis: (And that may be a bigger difference on output params than on input, not really mattering here)

src/hb-font.cc
```diff
@@ -1713,37 +1861,37 @@ hb_font_set_scale (hb_font_t *font,
...
  * hb_font_set_ppem:
- * @font: a font.
- * @x_ppem: 
- * @y_ppem: 
+ * @font: #hb_font_t to work upon
+ * @x_ppem: Horizontal ppem value to assign
+ * @y_ppem: Vertical ppem value to assign
  *
- * 
+ * Sets the horizontal and vertical points-per-em (ppem) of a font.
```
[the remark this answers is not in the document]
n8willis: Yeah, it's pixels. That's on me.

src/hb-common.h
```diff
@@ -105,9 +131,33 @@ typedef union _hb_var_int_t {
 
 /* hb_tag_t */
 
+/**
+ * hb_tag_t:
+ * 
+ * Data type for tag identifiers. Tags are arrays of four
+ * bytes, each of which holds a character.
```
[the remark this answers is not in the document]
n8willis: I was stuck for wording on that, too. I didn't want to call them strings.

src/hb-common.h
```diff
@@ -171,7 +275,171 @@ HB_EXTERN hb_language_t
 hb_language_get_default (void);
 
 
-/* hb_script_t */
+/**
+ * hb_script_t:
+ * @HB_SCRIPT_COMMON: HB_TAG ('Z','y','y','y')
+ * @HB_SCRIPT_INHERITED: HB_TAG ('Z','i','n','h')
+ * @HB_SCRIPT_UNKNOWN: HB_TAG ('Z','z','z','z')
+ * @HB_SCRIPT_ARABIC
```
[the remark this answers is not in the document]
n8willis: Perhaps not. I thought that the non-obvious ones might be wanting an explanation, but that if we didn't include the whole list, it would be more confusing. I also considered copying in the HB_TAGs as commentary for each entry, but decided it was not urgent.

khaledhosny:
@behdad I did review this one as well, it is waiting for @n8willis to address my review comments.

khaledhosny (approved):
I rebased, resolved conflicts, and amended the changes with my unresolved review comments.

PR #1679, opened 2019-04-23
[Docs] Add GTK-Doc comments for external APIs

n8willis:
This adds gtk-doc inline comments to mid- and higher-level shaping API funcs & enums, namely in hb-face, hb-unicode, and several OpenType-specific sections of hb-ot-*. Also adds smaller fixes to existing GTK-Doc comment blocks for formatting and consistency.

### DIFF
--- a/src/hb-face.cc
+++ b/src/hb-face.cc
@@ -41,8 +41,10 @@
  * @short_description: Font face objects
  * @include: hb.h
  *
- * Font face is objects represent a single face in a font family.
- * More exactly, a font face represents a single face in a binary font file.
+ * A font face is an object that represents a single face from within a
+ * font family.
+ *
+ * More precisely, a font face represents a single face in a binary font file.
  * Font faces are typically built from a binary blob and a face index.
  * Font faces are used to create fonts.
  **/
@@ -52,7 +54,7 @@
  * hb_face_count:
  * @blob: a blob.
  *
- * Get number of faces in a blob.
+ * Fetches the number of faces in a blob.
  *
  * Return value: Number of faces in @blob
  *
@@ -96,11 +98,16 @@ DEFINE_NULL_INSTANCE (hb_face_t) =
 
 /**
  * hb_face_create_for_tables:
- * @reference_table_func: (closure user_data) (destroy destroy) (scope notified):
- * @user_data:
- * @destroy:
+ * @reference_table_func: (closure user_data) (destroy destroy) (scope notified): Table-referencing function
+ * @user_data: A pointer to the user data
+ * @destroy: (optional): A callback to call when @data is not needed anymore
  *
- *
+ * Variant of hb_face_create(), built for those cases where
+ * client programs are primarily concerned with querying the
+ * font tables, rather than with full shaping.
+ * 
+ * Creates a new face object from the specified @user_data and @reference_table_func,
+ * with the @destroy callback. 
  *
  * Return value: (transfer full)
  *
@@ -182,12 +189,13 @@ _hb_face_for_data_reference_table (hb_face_t *face HB_UNUSED, hb_tag_t tag, void
 
 /**
  * hb_face_create: (Xconstructor)
- * @blob:
- * @index:
+ * @blob: #hb_blob_t to work upon
+ * @index: The index of the face within @blob
  *
+ * Constructs a new face object from the specified blob and
+ * a face index into that blob.
  *
- *
- * Return value: (transfer full):
+ * Return value: (transfer full): The new face object
  *
  * Since: 0.9.2
  **/
@@ -222,9 +230,9 @@ hb_face_create (hb_blob_t    *blob,
 /**
  * hb_face_get_empty:
  *
+ * Fetches the singleton empty face object.
  *
- *
- * Return value: (transfer full)
+ * Return value: (transfer full) The empty face object
  *
  * Since: 0.9.2
  **/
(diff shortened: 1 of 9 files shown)


src/hb-face.cc
```diff
@@ -96,11 +98,16 @@ DEFINE_NULL_INSTANCE (hb_face_t) =
...
- * @user_data:
- * @destroy:
- *
+ * @reference_table_func: (closure user_data) (destroy destroy) (scope notified): Table-referencing function
+ * @user_data: A pointer to the user data
+ * @destroy: (optional): A callback to call when @data is not needed anymore
  *
+ * Variant of hb_face_create(), built for those cases where
+ * client programs are primarily concerned with querying the
+ * font tables, rather than with full shaping.
```
khaledhosny: I don’t think this is true. The faces created with `hb_face_create_for_tables()` can be used for any use (just like faces created with `hb_face_create()`). The difference is that instead of having a blob representing the full font data, a callback function is used to fetch the table data on demand. This can be used when interfacing with APIs that don’t give access to full font data (e.g. Core Text and Core Graphics APIs on macOS) or for whatever a reason the client finds this more appropriate than `hb_face_create()`.

src/hb-face.cc
```diff
@@ -96,11 +98,16 @@ DEFINE_NULL_INSTANCE (hb_face_t) =
...
+ * 
+ * Creates a new face object from the specified @user_data and @reference_table_func,
+ * with the @destroy callback. 
  *
  * Return value: (transfer full)
```
khaledhosny: I think a colon is missing here, and may be some text about the return value.

src/hb-face.cc
```diff
@@ -182,12 +189,13 @@ _hb_face_for_data_reference_table (hb_face_t *face HB_UNUSED, hb_tag_t tag, void
...
 /**
  * hb_face_create: (Xconstructor)
- * @blob:
- * @index:
- *
+ * @blob: #hb_blob_t to work upon
+ * @index: The index of the face within @blob
  *
+ * Constructs a new face object from the specified blob and
+ * a face index into that blob.
```
khaledhosny: May be document that index is for font collections (like TTC, or Dfont).

src/hb-face.cc
```diff
@@ -354,12 +365,13 @@ hb_face_is_immutable (const hb_face_t *face)
 
 /**
  * hb_face_reference_table:
- * @face: a face.
- * @tag:
- *
+ * @face: A face object
+ * @tag: The #hb_tag_t of the table to query
```
khaledhosny: I think we need to document that `HB_TAG_NONE` can be used to fetch the entire font data (but this might not always be possible if the face was created with `hb_face_create_for_tables()` and the reference function does not support this).

src/hb-face.cc
```diff
@@ -372,11 +384,13 @@ hb_face_reference_table (const hb_face_t *face,
 
 /**
  * hb_face_reference_blob:
- * @face: a face.
- *
+ * @face: A face object
  *
+ * Fetches a pointer to the binary blob that contains the
+ * specified face. Can only be used on faces created with
+ * hb_face_builder_create().
```
khaledhosny: I’m not sure the last sentence is correct. This function is basically `hb_face_reference_table(face, HB_TAG_NONE)`, so it should work similar to my comment above.

src/hb-face.h
```diff
@@ -46,6 +46,14 @@ hb_face_count (hb_blob_t *blob);
  * hb_face_t
  */
 
+/**
+ * hb_face_t:
+ * @blob: A pointer to the #hb_blob_t that contains the binary blob of a font file
+ * @index: The index (zero-based) into the font file that points to the face
+ *
+ * Data type for holding font faces.
```
khaledhosny: I don’t think `hb_face_t` has any public members.

src/hb-ot-layout.cc
```diff
@@ -69,7 +69,7 @@
  * Tests whether a face includes any kerning data in the 'kern' table.
  * Does NOT test for kerning lookups in the GPOS table.
  *
- * Return value: true if data found, false otherwise
+ * Return value: %true if data found, false otherwise
```
khaledhosny: Why not also `%false`?

src/hb-unicode.cc
```diff
@@ -127,6 +132,11 @@ extern "C" hb_unicode_funcs_t *hb_glib_get_unicode_funcs ();
 extern "C" hb_unicode_funcs_t *hb_icu_get_unicode_funcs ();
 extern "C" hb_unicode_funcs_t *hb_ucdn_get_unicode_funcs ();
 
+/**
+ *
+ *
+ *
+ **/
```
khaledhosny: Hmm.

src/hb-face.cc
n8willis: OK. Follow-up question for that case: is the callback function capable of providing access to everything that the blob could, or does it mean there are some restrictions?

behdad: Generally there's no restriction in those method. Definitely not for shaping.

Currently our API for enumerating fine tables doesn't work with faces created this way.

src/hb-ot-layout.cc
n8willis: That's a good point (only real reason is that prior inline comments elsewhere used %true). On the other hand, there are a lot of those, sprinkled around a lot of files, so handling them all in a separate PR after the big merges are done might be nicer -- get more docs out faster, so to speak. I'm not sure how useful all the decorations are in reality anyway.

khaledhosny: Sounds good.

khaledhosny:
Rebased and updated.

khaledhosny (approved):

alerque (approved):

PR #1680, opened 2019-04-23
Usermanual: object-model chapter

n8willis:
This adds a usermanual chapter covering the general object model, user_data facility, lifecycle stuff, and the basics of blobs.


docs/usermanual-object-model.xml
```diff
@@ -0,0 +1,249 @@
...
+      In HarfBuzz, all object types provide certain
+      lifecycle-management APIs.  Objects are reference-counted, and
+      constructed with various <function>create()</function> methods, referenced via
+      <function>reference()</function> and dereferenced using
+      <function>destroy()</function>.
+    </para>
+    <para>
+      For example,
+      the <literal>hb_buffer_t</literal> object has
+      <function>hb_buffer_create()</function> as its constructor,
```
behdad: Sounds obvious, but do we need to mention that objects are created with reference count 1?

docs/usermanual-object-model.xml
```diff
@@ -0,0 +1,249 @@
...
+      <listitem>
+	<para>
+	  <emphasis>faces</emphasis>, which represent typefaces from a
+	  font file, but without specific parameters (such as size) set.
+	</para>
+      </listitem>
+      <listitem>
+	<para>
+	  <emphasis>fonts</emphasis>, which represent instances of a
+	  face will all of their parameters specified.
```
behdad: "will" -> "with".

docs/usermanual-object-model.xml
```diff
@@ -0,0 +1,249 @@
...
+	<para>
+	  <emphasis>buffers</emphasis>, which hold Unicode code points
+	  for characters (before shaping) and the shaped glyph output
+	  (after shaping).
+	</para>
+      </listitem>
+      <listitem>
+	<para>
+	  <emphasis>shape plans</emphasis>, which store the settings
+	  that HarfBuzz will use when shaping a particular text segment.
```
behdad: Maybe add a note in brackets that this one is rarely needed by clients?

docs/usermanual-object-model.xml
```diff
@@ -0,0 +1,249 @@
...
+      <function>set_*()</function> calls to set up the
+      object, and then use it without further modification.
+    </para>
+    <para>
+      To ensure that such an object is not modified, client programs
+      can explicitly mark an object as immutable. HarfBuzz provides
+      <function>make_immutable()</function> methods to mark an object
+      as immutable and <function>is_immutable()</function> methods to
+      test whether or not an object is immutable. Attempts to use
+      setter functions on immutable objects will fail; see the API
```
behdad: "fail" -> "fail silently"?

docs/usermanual-object-model.xml
```diff
@@ -0,0 +1,249 @@
...
+    <para>
+      Note also that there are no "make mutable" methods. If client
+      programs need to alter an object previously marked as immutable,
+      they will need to make a duplicate of the original.
+    </para>
+    <para>
+      Finally, object constructors (and, indeed, as much of the API as
+      possible) will never return <literal>NULL</literal>.  Instead,
+      if there is an allocation error, each constructor will
+      return an “empty” object singleton.
```
behdad: We're adding new idiom in the subsetter that is called "create_or_fail()" which can return NULL.  Maybe document here.

docs/usermanual-object-model.xml
```diff
@@ -0,0 +1,249 @@
...
+      The <function>set_user_data()</function> methods also support an
+      optional <function>destroy</function> callback. Client programs
+      can set the <function>destroy</function> callback and receive
+      notification from HarfBuzz whenever the object is destructed.
+    </para>
+    <para>
+      Finally, each <function>set_user_data()</function> method allows
+      the client program to set a <literal>replace</literal> Boolean
+      indicating whether or not the <literal>user_data</literal>
+      associated with a key ought to be replaceable.
```
behdad: The "ought to be replaceable" is misleading or wrong.  "replace" defines whether to override a pre-existing value for the same key or not.  It is necessary for thread-safety, whereas two threads might try to set user-data for the same key.  If the second one unconditionally overrides the first one, we are in trouble.

behdad:
A few random points:
- When talking about threadsafety, maybe the first time, say "(unless library was compiled with `HB_NO_MT`.)"?
- There's also hb_set_t and hb_map_t.  Not relevant to shaping though, so maybe no need to list here.
- Here or somewhere else, document that hb_face_t is rather heavy object (even though still very light) and preferably should be cached / reused; whereas hb_font_t is lightweight and can either be cached for a particular size and reused (possibly from multiple threads), or change size every time needed, or even to create one every time needed, though the latter is not ideal.
- Since you asked me in email: ffuncs can be thought of the "vtable" for hb_font_t, in C++ speak.
- It sounds obvious, but I've seen people doing it: mention that a hb_buffer_t can only be meaningfully used by one thread at a time.

Thanks.

docs/usermanual-object-model.xml
n8willis: Probably wouldn't confuse too many people, but since I'm making other edits, NP to add.

docs/usermanual-object-model.xml
n8willis: Oof.

docs/usermanual-object-model.xml
n8willis: Sure. I put it in there just because the still-WIP chapter about OpenType features mentions shape plans as a way to tie in the higher-level discussions to the what-HB-actually-does stuff.

docs/usermanual-object-model.xml
n8willis: Gotcha. Will fix.

behdad:
@n8willis Want to finish this?  I'm fine if you want to leave as is.  Lemme also give you commit access.

n8willis:
> @n8willis Want to finish this? I'm fine if you want to leave as is. Lemme also give you commit access.

Sure; I'm just getting a couple of other ducks in a row on manual page updates to push.

docs/usermanual-object-model.xml
n8willis: I think for the moment I'd prefer to hold off on mentioning that here. I did insert 'shaping' to "as much of the ___ API as possible" so that when the subsetter material goes in, it'll be a smoother change.

docs/usermanual-object-model.xml
n8willis: :heavy_check_mark:

PR #1681, opened 2019-04-25
Nametable subset

qxliu76:


src/Makefile.sources
```diff
@@ -84,7 +84,6 @@ HB_BASE_sources = \
 	hb-ot-math-table.hh \
 	hb-ot-math.cc \
 	hb-ot-maxp-table.hh \
-	hb-ot-name-language.cc \
```
behdad: Please include the `hb-ot-name-language-static.hh` here.

PR #1685, opened 2019-04-29
Allow some Balinese punctuation and symbols as bases for aksara modre

dscorbett:
According to [the Balinese Unicode proposal](https://www.unicode.org/L2/L2005/05008-n2908-balinese.pdf#page=4), U+1B01 BALINESE SIGN ULU CANDRA can appear on some punctuation and symbols.


behdad:
Thanks.  We should move all the overrides to a file...

I'm also thinking of moving to consuming ucdxml.  So if overrides becoming a file, would be easier if they were ucdxml format.

behdad:
Feel free to merge after bots happy.

PR #1688, opened 2019-04-30
Add hb-render to view serialized buffers

### LINKED ISSUE #1480: hb-render, a hb-view able to input a serialized buffer instead text
Something useful to have around, using hb_buffer_deserialize_glyphs API. (the name was suggested by Behdad), one draft was uploaded at #1688

ebraminio:
Fixes #1480

Occurred to me again while reading --trace output yesterday as I wanted something to see each level change so may something others may need also but the issue is serialized output doesn't contain font size so need some advice I guess.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -184,6 +184,8 @@ extract_make_variable(HB_BASE_RAGEL_GENERATED_sources ${SRCSOURCES})
 
 extract_make_variable(HB_VIEW_sources ${UTILSOURCES})
 add_prefix_to_list(HB_VIEW_sources "${PROJECT_SOURCE_DIR}/util/")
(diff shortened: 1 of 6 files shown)


behdad:
I think font-size needs to be provided to hb-render. Make it default to whatever hb-shape uses.  Though i think we need to allow control of input scale and output font size separately.  I'll comment more from laptop.

behdad:
I haven't reviewed fully yet, but ideally this should involve very little code indeed.

I think we should have a `--scale` option that is used for loading part, and the usual `--font-size` to control output size.

ebraminio:
Needs some more updates and things, let me close it but the file exists #1480 so can be revisited later easily.

PR #1689, opened 2019-04-30
Remove space between right angle brackets now that we have C++11

ebraminio:


behdad:
Thanks. Have been meaning to do this.  Make sure doesn't interfere with @blueshade7's branch much?

ebraminio:
Doesn't conflict with https://github.com/harfbuzz/harfbuzz/pull/1594 AFAICS so lets merge it

PR #1691, opened 2019-04-30
Usermanual: shaping+

n8willis:
This fills in the usermanual chapters covering the shaping API, command-line utilities, and miscellaneous lower-level internal bits. Does not address Platform-Integration (intentionally), as that is coming separately.


docs/usermanual-buffers-language-script-and-direction.xml
```diff
@@ -39,46 +47,350 @@
...
+      <literal>Script</literal> (sc) properties) that is useful
+      for shaping, as well as some useful operations like composing and
+      decomposing code points.
+    </para>
+    <para>
+      At build time, HarfBuzz looks first for the GLib library. If
+      it is found, HarfBuzz will use GLib's Unicode functions by
+      default. If there is no GLib, HarfBuzz will fall back to its own
+      internal, lightweight set of Unicode functions instead.
+    </para>
```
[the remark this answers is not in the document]
n8willis: Are they in the fallback chain below GLib, or orthogonal to that?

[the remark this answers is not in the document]
khaledhosny: Yes, below GLib (when HarfBuzz is built with `--with-icu=builtin`, when ICU support is enable otherwise it is built as a separate harfbuzz-icu library).

docs/usermanual-buffers-language-script-and-direction.xml
```diff
@@ -302,7 +302,9 @@
     <para>
       At build time, HarfBuzz looks first for the GLib library. If
       it is found, HarfBuzz will use GLib's Unicode functions by
-      default. If there is no GLib, HarfBuzz will fall back to its own
+      default. If there is no GLib, HarfBuzz will look for the
+      International Components for Unicode (ICU) library. If neither
+      GLib nor ICU is found, HarfBuzz will fall back to its own
```
behdad: Actually, it's been a while since we prefer our own implementation to that of GLib or ICU.  And as of yesterday, our is always compiled in using the existing build systems.

n8willis:
Happy to merge this if it looks ready enough for public consumption.

docs/usermanual-buffers-language-script-and-direction.xml
n8willis: Ah, timing....

n8willis: So: if you enable GLib & ICU config switches, how does that affect the ufuncs selection scheme?

behdad: Still internal preferred.  Unless turned off.  Don't document how to turn it off.  That's reserved for systems that custom-integrate / build.  Tell them to refer to hb-unicode.cc if needed.  Here's the exact logic:

https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-unicode.cc#L135

behdad: Over time the glib/icu callbacks have become less relevant for sure.

docs/usermanual-buffers-language-script-and-direction.xml
```diff
@@ -39,46 +47,363 @@
...
+    </para>
+    <para>
+      HarfBuzz recognizes four text directions in
+      <type>hb_direction_t</type>: left-to-right
+      (<literal>HB_DIRECTION_LTR</literal>), right-to-left (<literal>HB_DIRECTION_RTL</literal>),
+      top-to-bottom (<literal>HB_DIRECTION_TTB</literal>), and
+      bottom-to-top (<literal>HB_DIRECTION_BTT</literal>).  For the
+      script property, HarfBuzz uses identifiers based on the
+      <ulink
+      url="https://unicode.org/iso15924/">ISO-15924
```
dscorbett: “ISO 15924”, without the hyphen, is how they style it.

docs/usermanual-opentype-features.xml
```diff
@@ -6,14 +6,296 @@
...
+      used for complex scripts can be quite involved; HarfBuzz tries
+      to be 100% compatible with the OpenType Layout specification
+      and, wherever there is any ambiguity, HarfBuzz attempts to replicate the
+      output of Microsoft's Uniscribe engine. See the <ulink
+      url="https://docs.microsoft.com/en-us/typography/script-development/standard">Microsoft
+      Typeography pages</ulink> for more detail.
+    </para>
+    <para>
+      In general, though, all that you need to know if that
+      <function>hb-shape()</function> returns the results of shaping
```
dscorbett: The function is `hb_shape`, with an underscore.

docs/usermanual-buffers-language-script-and-direction.xml
n8willis: OK. Now just mentioning they exist, and trusting that any interested parties will enjoy exploring it.

docs/usermanual-fonts-and-faces.xml
```diff
@@ -5,20 +5,449 @@
...
+      fonts, the latest update to the OpenType font format, and at
+      some other recent additions to OpenType.
+    </para>
+
+  <section id="fonts-and-faces-objects">
+    <title>Font and face objects</title>
+    <para>
+      The outcome of shaping a run of text depends on the contents of
+      a specific font file (such as the substitutions and positioning
+      moves in the 'GSUB' and 'GPOS' tables), so HarfBuzz makes
```
dscorbett: The user manual is inconsistent about how to style tags: `abcd`, `'abcd'`, `"abcd"`, or `<literal>abcd</literal>`?

docs/usermanual-opentype-features.xml
```diff
@@ -6,14 +6,296 @@
...
+    <para>
+      With a well-behaved OpenType font, you expect GDEF, GSUB, and
+      GPOS tables to all be applied. HarfBuzz implements the
+      script-specific shaping models in internal functions, rather
+      than in the public API.
+    </para>
+    <para>
+      The algorithms
+      used for complex scripts can be quite involved; HarfBuzz tries
+      to be 100% compatible with the OpenType Layout specification
```
dscorbett: There are intentional differences, so the goal is not 100% compatibility.

docs/usermanual-opentype-features.xml
```diff
@@ -6,14 +6,296 @@
...
+      and, wherever there is any ambiguity, HarfBuzz attempts to replicate the
+      output of Microsoft's Uniscribe engine. See the <ulink
+      url="https://docs.microsoft.com/en-us/typography/script-development/standard">Microsoft
+      Typeography pages</ulink> for more detail.
```
dscorbett: An ironic typo.

docs/usermanual-opentype-features.xml
```diff
@@ -6,14 +6,296 @@
...
+      will get performed:
+    </para>
+    <programlisting language="C">
+      hb_shape(font, buf, userfeatures, num_features);
+    </programlisting>
+    <para>
+      Just like we enabled the <literal>dlig</literal> feature by
+      setting its <parameter>value</parameter> to
+      <literal>1</literal>, you would disable a feature by setting its
+      <parameter>value</parameter> to <literal>0</literal>.
```
dscorbett: Would it be worth explaining that there are other valid values than 0 and 1?

docs/usermanual-opentype-features.xml
n8willis: Well, there's also not exactly a "specification" for OpenType Layout. Hence https://github.com/n8willis/opentype-shaping-documents ....

docs/usermanual-fonts-and-faces.xml
n8willis: My personal belief is that table & feature names match <token> best (literal is too generic; it's what I've used when I can't find a good match elsewhere in docbook); the docbook <term> kind of just doesn't match specification requirements, since it's so code-focused. But that's something that I need to do in a full pass on all the docs. The a-priori material was also inconsistent, and in the interest of time I have not been trying to rectify those while also getting the new text written. Results will be better when markup-consistency is its own pass, IMO.

docs/usermanual-opentype-features.xml
n8willis: Added a brief note. My position is that explaining all the valid ranges/options is tangential in user-manual body text; it ought to be spelled out in the reference (HB or OpenType).

PR #1695, opened 2019-05-04
[use] Skip CGJ, ZWNJ, and most other default ignorables when clustering

### LINKED ISSUE #483: CGJ and ZWNJ in USE
There should be no problem with using CGJ and ZWNJ in USE away from stackers and viramas.  However, I get dotted circles in the display from hb-view (and Firefox 53.0) for the Tai Tham sequence <U+1A3E, U+1A66, U+1A63, U+0020, U+1A3E, U+1A66, U+034F, U+1A63, U+0020, U+1A36, U+1A63 U+0020, U+1A36, U+200C, U+1A63>:
cgj
The first pair is without and with CGJ between the vowels, and the second pair is without and with ZWNJ in the middle.  The first pair should look identical, whereas the first word of the second pair should have a ligature and the latter should not.  The font is Da Lekh Seri, available at http://wrdingham.co.uk/lanna/dalekh_seri.ttf (and also '.woff').

The first pair is the start of the real word ᨾᩦᩣ᩠ᨠ and the second is an actual word which can appear unligated in teaching material.

dscorbett:
This pull request refactors the Ragel machine index into a new struct, `machine_index_t`, with an overloaded `operator ++` that skips most default ignorables. ZWNJ is only skipped if it precedes a mark (ignoring CGJ per [MicrosoftDocs/typography-issues#216](https://github.com/MicrosoftDocs/typography-issues/issues/216#issuecomment-489719320)). ZWJ and WJ have their own USE categories and are not skipped. This fixes #483, whose example is now rendered thus:
ᨾᩦᩣ ᨾᩦ͏ᩣ ᨶᩣ ᨶ‌ᩣ

There is a new iterator, `machine_index_t`, that knows its current index, by assuming that its inner iterator was created using `hb_enumerate`. Assigning an `unsigned` to it advances or rewinds the iterator till it gets to that index. It compares for equality by comparing the index, which only makes sense if you only compare iterators backed by the same array, which this code makes sure to do.

As an extension to the USE spec, the machine skips not only CGJ but other default ignorable code points not relevant to Indic shaping, like U+1D173 MUSICAL SYMBOL BEGIN BEAM. This includes the variation selectors, which USE allows only after non-marks. That restriction is wrong because variation selectors, being default ignorable, should be ignored when they are in invalid positions, and because not all variation sequence bases are non-marks.

Another extension is that ZWNJ is skipped before all marks, including gc=Me, not just gc=Mc and gc=Mn.

This pull request includes a bug fix in `_hb_ot_shape_normalize`. A CGJ that precedes a starter (a code point with ccc=0) cannot have blocked reordering, so it should be ignored, not hidden.


src/gen-use-table.py
```diff
@@ -246,8 +239,6 @@ def is_SYM(U, UISC, UGC):
 	return UGC in [So, Sc] and U not in [0x1B62, 0x1B68]
 def is_SYM_MOD(U, UISC, UGC):
 	return U in [0x1B6B, 0x1B6C, 0x1B6D, 0x1B6E, 0x1B6F, 0x1B70, 0x1B71, 0x1B72, 0x1B73]
-def is_VARIATION_SELECTOR(U, UISC, UGC):
-	return 0xFE00 <= U <= 0xFE0F
```
behdad: Do we need a comment somewhere, that we handle CGJ and VS separately?

src/hb-ot-shape-complex-machine-index.hh
```diff
@@ -0,0 +1,75 @@
...
+  hb_buffer_t *buffer;
+  unsigned int p;
+
+  public:
+  constexpr machine_index_t (hb_buffer_t *const buffer)
+    : buffer (buffer), p (0u) {}
+
+  unsigned int operator + (const int i) const
+  {
+    return p + i;
```
behdad: But now we are skipping.  This should loop over `++` operator or it will be wrong.

src/hb-ot-shape-complex-machine-index.hh
```diff
@@ -0,0 +1,75 @@
...
+  {
+    p = i;
+  }
+
+  bool operator == (const unsigned int i) const
+  {
+    return p == i;
+  }
+
+  unsigned int operator ++ ()
```
behdad: I think both `++` operators should be implemented for consistency.

src/hb-ot-shape-complex-machine-index.hh
```diff
@@ -0,0 +1,75 @@
...
+  {
+    do {
+      ++p;
+    } while (p < buffer->len
+	     && (buffer->info[p].use_category() == USE_O || buffer->info[p].use_category() == USE_Rsv)
+	     && _hb_glyph_info_is_default_ignorable (&buffer->info[p]));
+    return p;
+  }
+
+  void operator -- (int)
```
behdad: Same. Walking back you should skip the same things or it will be inconsistent and there will be dragons...

src/hb-ot-shape-complex-machine-index.hh
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  unsigned int operator ++ ()
+  {
+    do {
+      ++p;
+    } while (p < buffer->len
+	     && (buffer->info[p].use_category() == USE_O || buffer->info[p].use_category() == USE_Rsv)
+	     && _hb_glyph_info_is_default_ignorable (&buffer->info[p]));
+    return p;
```
behdad: Please make this class a template that takes a function-object to call to skip things, so we can use it from other shapers as well.

behdad: Might even use it for AAT state machines...  Although I know @jfkthame had concerns there.

src/hb-ot-shape-complex-use-machine.rl
```diff
@@ -157,15 +156,15 @@ main := |*
 static void
 find_syllables (hb_buffer_t *buffer)
 {
-  unsigned int p, pe, eof, ts, te, act;
+  machine_index_t p (buffer);
+  unsigned int pe, eof, ts, te, act;
```
behdad: Are we should the other vars are never arithmatically-operated on?  I think that's correct.  Don't know how to ensure it though.

behdad: Maybe use a type that has int cast operators but no arithmetic?

dscorbett:
It seems like the cleanest implementation is a wrapper around an iterator. It’ll take me a while to become familiar with hb-iter.hh.

behdad:
> It seems like the cleanest implementation is a wrapper around an iterator. It’ll take me a while to become familiar with hb-iter.hh.

Humm. Possibly.  Though implementing the operators is not that much work.

Anyway, I think you should reuse the `matcher_t` inside, at least to get the bulk of the work, and just do USE-specific part yourself, if any.

dscorbett:
I am not sure what you mean about `matcher_t`. This PR is just about clustering; I don’t think there is a problem with matching.

I know that avoiding hb-iter.hh would be easier, but I think using an iterator is better design, so that there aren’t two essentially identical implementations of these sets of operators. My plan is to enumerate `buffer`, filter out CGJ etc., and map to the enumerated index, all of which will be wrapped in a struct that can be implicitly cast to `int`.

behdad:
I was suggesting that you take your iterator template take a type, and `matcher_t` can be passed to it for normal skipping.  But forget about that, unnecessary.

Upon further thinking, I like the iterator design AND think it's very simple.  First, you don't need the cast operator.  Override `getkey` as needed.  All we need is to 1. implement `buffer::iter()`, then use filtering.  Ie. sketch:
```
  auto p =
  + hb_iter (buffer)
  | hb_filter ([&] (hb_glyph_info_t &_) { return whether to match, you can access buffer-> info here... }
  ;
  auto pe = decltype (p) (p) + buffer->len;
  ...
```

dscorbett:
Ragel generates statements like `ts = p;` and `ts = 0;`, so there needs to be either a cast overload for `p` or an assignment overload for `ts`.

behdad:
> Ragel generates statements like `ts = p;` and `ts = 0;`, so there needs to be either a cast overload for `p` or an assignment overload for `ts`.

`ts` etc all have the same type as `p`.  So it's just the integer assignment that needs to be implemented.  Let me think about it.

behdad:
Implementing ts as a special type sounds easier.

behdad:
> ```
>   auto p =
>   + hb_iter (buffer)
>   | hb_filter ([&] (hb_glyph_info_t &_) { return whether to match, you can access buffer-> info here... })
>   ;
>   auto pe = p + p.len ()
>   ...
> ```

This should work now.  Previously filter iterators were not assignable.  I made them so.

dscorbett:
Because of `te = 0;`, `te = p + 1;`, and `p = ((te))-1;`, I conclude that `p` and `te` need to have the same type and that that type needs an assignment overload for `unsigned`. I see three problems: an iterator doesn’t know the current index in the buffer, but maybe I can use `hb_enumerate` to keep track of the index; `hb_enumerate` doesn’t take an array and a length like `hb_iter` does; and I am way out of my depth with all these fancy C++ features. Here is what I’m trying in `find_syllables`:
```cpp
  auto itx =
    + hb_iter (buffer, buffer->len)
    | hb_filter ([&] (hb_glyph_info_t &_)
		 { return !((_.use_category() == USE_O || _.use_category() == USE_Rsv) && _hb_glyph_info_is_default_ignorable (&_)); });
  struct machine_index_t : decltype (itx)
  {
    HB_ITER_USING (decltype (itx));
    void operator = (unsigned i)
    { /* TODO */ }
  };
  machine_index_t p = itx;
```
And here is the error message:
```
In file included from hb-ot-shape-complex-use.cc:257:
./hb-ot-shape-complex-use-machine.hh:397:18: error: no viable conversion from 'hb_filter_iter_t<hb_filter_iter_t<hb_array_t<hb_buffer_t>, (lambda at
      ./hb-ot-shape-complex-use-machine.hh:386:15), const (anonymous struct at ./hb-algs.hh:37:1) &, nullptr>, (lambda at
      ./hb-ot-shape-complex-use-machine.hh:388:15), const (anonymous struct at ./hb-algs.hh:37:1) &, nullptr>' to 'machine_index_t'
        machine_index_t p = itx;
                        ^   ~~~
./hb-ot-shape-complex-use-machine.hh:391:9: note: candidate constructor (the implicit copy constructor) not viable: cannot bind base class object of type
      'hb_filter_iter_t<hb_filter_iter_t<hb_array_t<hb_buffer_t>, (lambda at ./hb-ot-shape-complex-use-machine.hh:386:15), const
      (anonymous struct at ./hb-algs.hh:37:1) &, nullptr>, (lambda at ./hb-ot-shape-complex-use-machine.hh:388:15), const
      (anonymous struct at ./hb-algs.hh:37:1) &, nullptr>' to derived class reference 'const machine_index_t &' for 1st argument
        struct machine_index_t : decltype (itx)
               ^
./hb-ot-shape-complex-use-machine.hh:391:9: note: candidate constructor (the implicit move constructor) not viable: cannot bind base class object of type
      'hb_filter_iter_t<hb_filter_iter_t<hb_array_t<hb_buffer_t>, (lambda at ./hb-ot-shape-complex-use-machine.hh:386:15), const
      (anonymous struct at ./hb-algs.hh:37:1) &, nullptr>, (lambda at ./hb-ot-shape-complex-use-machine.hh:388:15), const
      (anonymous struct at ./hb-algs.hh:37:1) &, nullptr>' to derived class reference 'machine_index_t &&' for 1st argument
        struct machine_index_t : decltype (itx)
               ^
```
Hopefully you can see what I’m trying to do, even if the compiler can’t.

---

I’ve been locally rebasing to get all the new iterator changes. For future reference, because it was not obvious, here is what I had to do to compile Ragel 7.0.0.11.
```bash
brew install gcc
wget http://www.colm.net/files/colm/colm-0.13.0.6.tar.gz
tar zxf colm-0.13.0.6.tar.gz
cd colm-0.13.0.6
CC=/usr/local/bin/gcc-8 CXX=/usr/local/bin/g++-8 ./configure --prefix="$PWD/target"
make
make install
cd ..
wget http://www.colm.net/files/ragel/ragel-7.0.0.11.tar.gz
tar zxf ragel-7.0.0.11.tar.gz
cd ragel-7.0.0.11
CC=/usr/local/bin/gcc-8 CXX=/usr/local/bin/g++-8 ./configure --prefix="$PWD/target" --with-colm="$(dirname "$PWD")/colm-0.13.0.6/target"
make
make install
export PATH="$PWD/target/bin:$PATH"
```

behdad:
Re ragel, I might actually go back to the previous version.  It's just that my Fedora 29 ships with ragel 7.  I see weird warnings with that code though.

Regarding iterator knowing its location, you can use subtraction to find that.  I haven't read your comment in full yet.  Will do. Thanks.

behdad:
> `hb_enumerate` doesn’t take an array and a length like `hb_iter` does

You can do `hb_enumerate (hb_iter (..., ...))`.

behdad:
> ```c++
>     + hb_iter (buffer, buffer->len)
> ```

You want `buffer->info`, not `buffer`. Right?

behdad:
We should implement `.iter()` on buffer.  See for example: https://github.com/harfbuzz/harfbuzz/pull/1704/files

dscorbett:
I’ll make those changes, but what I really need help with is the assignment overload.

behdad:
Yeah let me think about it.

dscorbett:
Never mind, I figured it out. I’ll push once I’ve cleaned up the code.

behdad:
> Never mind, I figured it out. I’ll push once I’ve cleaned up the code.

OOOOOOOOOOOoooooooooooohhhhh..   Now I'm curious to see!

behdad:
Curious to see.

dscorbett:
This is ready for review again. I’ve updated `machine_index_t` to be an iterator. The USE-specific skipping logic is all in the Ragel source file.

behdad:
I see.  Thanks.  Like generally.  Should/can we apply to other shapers as well?

Wonder if we should add minus and assignment operators to some of the iterators natively.  For example, the enumerate can do that and then you wouldn't need to.

dscorbett:
The other Ragel shapers should use this system eventually. They all have subtly different requirements for default ignorables, so I’d like to finish with USE first.

Generalizing `machine_index_t`’s new behaviors to the other iterators sounds good to me. Assignment from an index requires a notion of current index, like what I added to `hb_array_t`. `==` would be similar to the existing `!=` implementations; what was the reason for implementing `!=` but not `==`?

behdad:
Just because != is what the range-based for loop calls.

dscorbett:
I don’t like how the first filter’s predicate is duplicated in the second. Is there a way to avoid that? Basically, I want a predicate to receive a copy of the iterator, so the result can depend on adjacent elements.

behdad:
> I don’t like how the first filter’s predicate is duplicated in the second. Is there a way to avoid that? Basically, I want a predicate to receive a copy of the iterator, so the result can depend on adjacent elements.

The typical way to do that is to zip the iter with ++iter so you get pairs...  That's becoming way too complex already..

A completely different approach, BTW, would be to combine two ragel machines together.  Not suggesting that you change approach.  Just something I thought about the other day.

dscorbett:
> The typical way to do that is to zip the iter with ++iter so you get pairs...

`hb_zip (p0, p0 + 1 | hb_map (hb_second))`, where `p0` is the iterator up to and not including the ZWNJ filter, doesn’t work: the second of the two zipped iterators is one shorter than the other, so the original iterator’s last element is skipped. Is there a way to add a dummy value to the end of an iterator?

behdad:
I will have more time to help finish this in a couple of weeks.

src/hb-ot-shape-complex-use-machine.rl
```diff
@@ -157,17 +156,35 @@ main := |*
...
-  unsigned int p, pe, eof, ts, te, act;
-  int cs;
   hb_glyph_info_t *info = buffer->info;
+  auto p =
+    + hb_iter (info, buffer->len)
+    | hb_enumerate
+    | hb_filter ([&] (hb_glyph_info_t &i)
+		 { return !((i.use_category() == USE_O || i.use_category() == USE_Rsv) && _hb_glyph_info_is_default_ignorable (&i)); },
+		 hb_second)
+    | hb_filter ([&] (const hb_pair_t<unsigned, hb_glyph_info_t &> p)
```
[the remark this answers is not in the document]
dscorbett: The second one captures `buffer` and `info`.

dscorbett:
Would any of your recent changes to `hb_zip` and other iterators help with this?

behdad:
> Would any of your recent changes to `hb_zip` and other iterators help with this?

Still need this: https://github.com/harfbuzz/harfbuzz/issues/1943#issuecomment-529209855
which would enable zip_longest...

khaledhosny:
What is the status of this PR, is their anything still blocking it?

dscorbett:
As far as I remember, this works. There is a tiny bit of duplication in `find_syllables_use` that I had hoped could be cleaned up. Because that is surprisingly difficult, and because there’s been no progress in a year, I think it should not block merging.

behdad:
Thanks. I'll check it out again and merge.

behdad:
I will fix the duplication after merging. Looks great. Thanks again.

src/hb-ot-shape-complex-machine-index.hh
```diff
@@ -0,0 +1,66 @@
...
+  static constexpr bool is_sorted_iterator = Iter::is_sorted_iterator;
+
+  typename Iter::item_t __item__ () const { return *it; }
+  typename Iter::item_t __item_at__ (unsigned i) const { return it[i]; }
+  unsigned __len__ () const { return it.len (); }
+  void __next__ () { ++it; }
+  void __forward__ (unsigned n) { it += n; }
+  void __prev__ () { --it; }
+  void __rewind__ (unsigned n) { it -= n; }
+  void operator = (unsigned n) { it -= (*it).first; it += n; }
```
behdad, line 47: This is not good since it's O(n). Fortunately, is easy to fix:
```
  void operator = (unsigned n) { it += n - (*it).first; }
```

behdad: Add casts to `int` as necessary.

src/hb-ot-shape-complex-use-machine.rl
```diff
@@ -173,17 +172,36 @@ main := |*
...
 find_syllables_use (hb_buffer_t *buffer)
 {
-  unsigned int p, pe, eof, ts, te, act;
-  int cs;
   hb_glyph_info_t *info = buffer->info;
+  auto p =
+    + hb_iter (info, buffer->len)
+    | hb_enumerate
+    | hb_filter ([] (const hb_glyph_info_t &i)
+		 { return !((i.use_category() == USE_O || i.use_category() == USE_Rsv) && _hb_glyph_info_is_default_ignorable (&i)); },
```
behdad, line 180: Just move this to a one-line function.

behdad:
Left a couple of comments. Please merge after addressing.  Thanks again, this is important.

src/hb-ot-shape-complex-machine-index.hh
```diff
@@ -0,0 +1,66 @@
...
+  static constexpr bool is_sorted_iterator = Iter::is_sorted_iterator;
+
+  typename Iter::item_t __item__ () const { return *it; }
+  typename Iter::item_t __item_at__ (unsigned i) const { return it[i]; }
+  unsigned __len__ () const { return it.len (); }
+  void __next__ () { ++it; }
+  void __forward__ (unsigned n) { it += n; }
+  void __prev__ () { --it; }
+  void __rewind__ (unsigned n) { it -= n; }
+  void operator = (unsigned n) { if ((*it).first < n) it += n - (*it).first; else if ((*it).first > n) it -= (*it).first - n; }
```
behdad, line 47: Thanks. Just because `it` can be arbitrarily complex, access `(*it).first` once only maybe?

This iterator will probably be more useful that just here, so can be named and moved to `hb-iter.hh`. But is fine here for now.

And please merge after. Thanks!

src/hb-ot-shape-complex-machine-index.hh
```diff
@@ -0,0 +1,67 @@
...
+ * DIRECT, INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES
+ * ARISING OUT OF THE USE OF THIS SOFTWARE AND ITS DOCUMENTATION, EVEN
+ * IF THE COPYRIGHT HOLDER HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ *
+ * THE COPYRIGHT HOLDER SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ * FITNESS FOR A PARTICULAR PURPOSE.  THE SOFTWARE PROVIDED HEREUNDER IS
+ * ON AN "AS IS" BASIS, AND THE COPYRIGHT HOLDER HAS NO OBLIGATION TO
+ * PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
```
behdad, line 20: Care adding your copyright line? Is definitely not necessary, but I find it useful to quickly see who has written a file generally.

behdad:
Thanks again!

PR #1700, opened 2019-05-07
[subset] fix for name table serializing

qxliu76:


src/hb-null.hh
```diff
@@ -67,7 +67,7 @@ template <typename T, typename B>
 struct _hb_static_size
 { enum { value = sizeof (T) }; };
 template <typename T>
-struct _hb_static_size<T, hb_bool_tt<true || sizeof (T::min_size)>>
+struct _hb_static_size<T, hb_bool_tt<true || sizeof (T::min_size)> >
```
behdad: These are unrelated and undesired changes.  Please remove.

src/hb-ot-name-table.hh
```diff
@@ -100,7 +100,9 @@ struct NameRecord
 		    const void *dst_base) const
   {
     auto *out = c->embed (this);
-    out->offset = 0;
+    if (!out)
+      return nullptr;
+
```
behdad: Instead of the three lines, just have one line:
```
if (unlikely (!out)) return_trace (false);
```
and yes, add tracing to the function please.

src/hb-ot-name-table.hh
```diff
@@ -292,6 +294,9 @@ struct name
     + hb_iter (name_record_idx_to_retain)
     | hb_apply ([&] (unsigned _) { c->copy (src_array[_], src_base, dst_base); })
     ;
+ 
+    if (c->ran_out_of_room)
+      return_trace (false);
```
behdad: Add `unlikely`, and return can go on the same line.

qxliu76:
> Thanks. Do the tests pass with this, or that's a separate issue?

this is fix for failed tests.

behdad:
Ah, so the assert was failing?

qxliu76:
the tests are failing in NameRecord->copy (), in which c->embed() returns a nullptr due to ran_out_of_room. Then test crashesdue to serialize_copy with a nullptr.

PR #1704, opened 2019-05-08
[map] add ability to iterate map

rsheeter:
As per whiteboard :)


behdad:
Thanks.  Fixed and merged manually.

PR #1706, opened 2019-05-11
Fix rewinding

dscorbett:
`__forward__` decreases the length of an array iterator. `__rewind__` also decreases it! That means that any code of the form `p = p - 1` where `p` is backed by an array iterator will get the wrong item from the array and eventually drain the iterator. This has the potential to cause much confusion and frustration. My fix is to keep track of the length of the array before the current index.

`__next__` advances to the next item of a filter iterator that satisfied the predicate. `__prev__` rewinds to the previous item without checking the predicate. My fix is to loop until the predicate is satisfied.

### DIFF
--- a/src/hb-array.hh
+++ b/src/hb-array.hh
@@ -42,20 +42,20 @@ struct hb_array_t : hb_iter_with_fallback_t<hb_array_t<Type>, Type&>
   /*
    * Constructors.
    */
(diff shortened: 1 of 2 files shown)


behdad:
Thanks.  Originally I didn't like to burden the array with keeping an extra field, and didn't like an unsafe behavior either, hence implementing an arbitrary behavior.  But I'm happy to do it this way.  Thanks.

Issue #1687, opened 2019-04-30
[subset] name IDs referenced by fvar are dropped


blueshade7:
Name IDs referenced by the fvar table shouldn’t be dropped from the name table when subsetting a variable font. 
In the example below, name IDs 257 thru 268 are dropped.
[SourceSerifVariable-Roman.ttf.zip](https://github.com/harfbuzz/harfbuzz/files/3129628/SourceSerifVariable-Roman.ttf.zip)
```
% hb-subset SourceSerifVariable-Roman.ttf --text=A --output-file=subset.ttf
% ttx -tfvar -tname subset.ttf
Dumping "subset.ttf" to "subset.ttx"...
Dumping 'fvar' table...
Dumping 'name' table...
% cat subset.ttx
<?xml version="1.0" encoding="UTF-8"?>
<ttFont sfntVersion="\x00\x01\x00\x00" ttLibVersion="3.41">

  <fvar>
    <Axis>
      <AxisTag>wght</AxisTag>
      <Flags>0x0</Flags>
      <MinValue>200.0</MinValue>
      <DefaultValue>389.34425</DefaultValue>
      <MaxValue>900.0</MaxValue>
      <AxisNameID>256</AxisNameID>
    </Axis>
    <NamedInstance flags="0x0" postscriptNameID="258" subfamilyNameID="257">
      <coord axis="wght" value="200.0"/>
    </NamedInstance>
    <NamedInstance flags="0x0" postscriptNameID="260" subfamilyNameID="259">
      <coord axis="wght" value="300.0"/>
    </NamedInstance>
    <NamedInstance flags="0x0" postscriptNameID="262" subfamilyNameID="261">
      <coord axis="wght" value="400.0"/>
    </NamedInstance>
    <NamedInstance flags="0x0" postscriptNameID="264" subfamilyNameID="263">
      <coord axis="wght" value="600.0"/>
    </NamedInstance>
    <NamedInstance flags="0x0" postscriptNameID="266" subfamilyNameID="265">
      <coord axis="wght" value="700.0"/>
    </NamedInstance>
    <NamedInstance flags="0x0" postscriptNameID="268" subfamilyNameID="267">
      <coord axis="wght" value="900.0"/>
    </NamedInstance>
  </fvar>

  <name>
    <namerecord nameID="1" platformID="1" platEncID="0" langID="0x0" unicode="True">
      Source Serif Variable
    </namerecord>
[...]

PR #1711, opened 2019-05-13
[subset] retian nameids from STAT and fvar tables in name table subsetting

qxliu76:


src/hb-ot-stat-table.hh
```diff
@@ -65,6 +65,8 @@ struct AxisValueFormat1
     return_trace (likely (c->check_struct (this)));
   }
 
+  unsigned get_value_name_id () const { return valueNameID; }
```
behdad: We have `hb_ot_name_id_t` for these value types.

behdad: Also in several other places.

src/hb-ot-stat-table.hh
```diff
@@ -212,6 +232,8 @@ struct StatAxisRecord
     return_trace (likely (c->check_struct (this)));
   }
 
+  NameID get_name_id () const { return nameID; }
```
behdad: Including here.

src/hb-ot-stat-table.hh
```diff
@@ -230,18 +252,58 @@ struct STAT
...
+			  version.major == 1 &&
 			  designAxesOffset.sanitize (c, this, designAxisCount) &&
 			  offsetToAxisValueOffsets.sanitize (c, this, axisValueCount, &(this+offsetToAxisValueOffsets))));
   }
+  
+  bool has_data () const { return version.to_int (); }
+
+  unsigned get_design_axis_count () const { return designAxisCount; }
+
+  unsigned get_axis_record_name_id (unsigned int axis_record_index) const
```
behdad: `hb_ot_name_id_t`.  Also, be consistent.  Just `unsigned` instead of `unsigned int` now that we're doing that.

src/hb-ot-stat-table.hh
```diff
@@ -230,18 +252,58 @@ struct STAT
...
+    + get_axis_value_offsets ()
+    | hb_apply ([&] (const OffsetTo<AxisValue>& _) { hb_set_add (nameids_to_retain, (this + _).get_value_name_id ()); })
+    ;
+  }
+
+  protected:
+  hb_array_t<const StatAxisRecord> const get_design_axes () const
+  { return (this+designAxesOffset).as_array (designAxisCount); }
+
+  hb_array_t<const OffsetTo<AxisValue> > const get_axis_value_offsets () const
```
behdad: The space between `> >` is not needed anymore.  Just write `>>`.

src/hb-ot-var-fvar-table.hh
```diff
@@ -280,6 +280,32 @@ struct fvar
     return axisCount;
   }
 
+  void collect_name_ids (hb_set_t *nameids) const
+  {
+    if (!has_data ()) return;
+
+    get_axes ()
```
behdad: Add a `+ ` before all of these to line them up.

src/hb-ot-stat-table.hh
```diff
@@ -230,18 +252,58 @@ struct STAT
...
+  {
+    if (!has_data ()) return;
+
+    + get_design_axes ()
+    | hb_map (&StatAxisRecord::get_name_id)
+    | hb_sink (nameids_to_retain)
+    ;
+
+    + get_axis_value_offsets ()
+    | hb_apply ([&] (const OffsetTo<AxisValue>& _) { hb_set_add (nameids_to_retain, (this + _).get_value_name_id ()); })
```
behdad: Move the `hb_set_add` to be a `hb_sink` still.  The `get_value_name_id` also can be a `hb_map` by itself.  You just need a lambda to add `this` to the offset.

I'll probably add a method to do that without having to define a lambda.

src/hb-ot-stat-table.hh
```diff
@@ -230,18 +252,58 @@ struct STAT
   bool sanitize (hb_sanitize_context_t *c) const
   {
     TRACE_SANITIZE (this);
-    return_trace (likely (c->check_struct (this) &&
-			  majorVersion == 1 &&
-			  minorVersion > 0 &&
+    return_trace (likely (version.sanitize (c) &&
+                          c->check_struct (this) &&
+			  version.major == 1 &&
```
behdad: This change is very wrong.  Please rewrite it to be equivalent to existing check, or otherwise make it match the spec.  As it is, it's just wrong.

src/hb-ot-var-fvar-table.hh
```diff
@@ -280,6 +280,32 @@ struct fvar
     return axisCount;
   }
 
+  void collect_name_ids (hb_set_t *nameids) const
+  {
+    if (!has_data ()) return;
+
+    + get_axes ()
```
garretrieger: Here I believe you should be able to do
| hb_map (&AxisRecord::axisNameId)
| hb_sink (nameids)
;

src/hb-ot-var-fvar-table.hh
```diff
@@ -280,6 +280,32 @@ struct fvar
...
+    | hb_apply ([&] (const AxisRecord& _) { hb_set_add (nameids, (unsigned) _.axisNameID); })
+    ;
+
+    + hb_range ((unsigned) instanceCount)
+    | hb_apply ([&] (const unsigned _)
```
garretrieger: change apply to be hb_map, and keep the lambda then add an hb_sink after to add to the set, likewise below.

blueshade7:
fixes #1687

PR #1714, opened 2019-05-15
Convert Consonant_Initial_Postfixed to CONS_MED

dscorbett:
According to [L2/17-120](https://www.unicode.org/L2/L2017/17120-isc-corrections.pdf#page=3), U+1A5A TAI THAM CONSONANT SIGN LOW PA can precede a vowel sign, so it should not be CONS_FINAL. U+1A5A is the only character with Indic_Syllabic_Category = Consonant_Initial_Postfixed.


behdad:
Thanks.

PR #1717, opened 2019-05-17
[subset] Add one ttf file with fvar/STAT tables to integration test, and also some fixes for tests

qxliu76:
fixes including:
-Ignore gvar/MVAR/HVAR table
-add support for --nameIDs=* option, make default behavior consistent with fonttools
-update cmap subsetting to only include the format 12 if the source file has a 12


util/options-subset.cc
```diff
@@ -29,15 +29,24 @@
...
 
-  hb_set_clear (name_ids);
+  char last_name_char = name[strlen (name) - 1];
+
+  if (last_name_char != '+' && last_name_char != '-')
+    hb_set_clear (name_ids);
+
+  if (0 == strcmp (arg, "*"))
+  {
+    hb_set_add_range (name_ids, 0, 0x7FFF);
```
garretrieger: You'll need to check if your in removal mode and call del_range if so.

PR #1718, opened 2019-05-17
[subset] Add drop tables flag to the subsetter.

garretrieger:


util/options.cc
```diff
@@ -1013,6 +1015,43 @@ parse_nameids (const char *name G_GNUC_UNUSED,
...
+                   GError    **error G_GNUC_UNUSED)
+{
+  subset_options_t *subset_opts = (subset_options_t *) data;
+
+  hb_set_clear (subset_opts->drop_tables);
+  char *s = (char *) arg;
+
+  while (s && *s)
+  {
+    while (*s && strchr (", ", *s))
```
behdad: This loop (and the one below) can be replaced with calls to `strpbrk`.

garretrieger: Switched to strtok

ebraminio:
Was removing the API intended? I think @drott and @HalCanary should made be aware of the change, and `morx`, `mort`, `kern` and `kerx` should be dropped as well.

src/hb-subset-input.cc
```diff
@@ -45,11 +45,39 @@ hb_subset_input_create_or_fail ()
...
+    HB_TAG ('P', 'C', 'L', 'T'),
+    HB_TAG ('L', 'T', 'S', 'H'),
+    // Graphite tables
+    HB_TAG ('F', 'e', 'a', 't'),
+    HB_TAG ('G', 'l', 'a', 't'),
+    HB_TAG ('G', 'l', 'o', 'c'),
+    HB_TAG ('S', 'i', 'l', 'f'),
+    HB_TAG ('S', 'i', 'l', 'l'),
+    // Colour
+    HB_TAG ('s', 'b', 'i', 'x')
```
ebraminio: I think `CBDT`, `CLBC`, `COLR`, `CPAL` and `svg` should be dropped as well

src/hb-subset-input.cc
```diff
@@ -45,11 +45,39 @@ hb_subset_input_create_or_fail ()
...
   input->drop_hints = false;
-  input->drop_layout = true;
   input->desubroutinize = false;
   input->retain_gids = false;
 
+  hb_tag_t default_drop_tables[] = {
+    // Layout disabled by default
+    HB_TAG ('G', 'S', 'U', 'B'),
+    HB_TAG ('G', 'P', 'O', 'S'),
+    HB_TAG ('G', 'D', 'E', 'F'),
```
ebraminio: And `kern`, `kerx`, `mort` and `morx`

garretrieger: Thanks, I'll add those.

src/hb-subset-input.cc
garretrieger: I'd like this set to mirror the default drop set found in fontTools. Do you think it would make sense to update fontTools to also drop those by default?

garretrieger:
> Was removing the API intended? I think @drott and @HalCanary should made be aware of the change, and `morx`, `mort`, `kern` and `kerx` should be dropped as well.

Yes, trying to mirror fontTools. It does not have a drop layout type option instead you just drop the associated tables. Will add those tables to the drop list.

src/hb-subset-input.cc
ebraminio: I don't know the context and requirement of both but if sbix is dropped somewhere for being a color related those also should be dropped.

What worries me more is to embed a full color font into a saved PDF just because accidental use of Emoji in the page, for the reason I wished we had whitelisting instead blacklisting or dropping usually huge color font by default somehow.

Issue #525, opened 2017-08-22
Dotted circle between instances of U+1A7F


dscorbett:
U+1A7F TAI THAM COMBINING CRYPTOGRAMMIC DOT may appear multiple times on the same syllable. ⟨ᨠ᩿᩿᩿᩿᩿⟩ has five cryptogrammic dots, and dotted circles are inserted before all but the first. No dotted circles should be inserted.

jfkthame:
U+1A7F has Indic Syllabic Category = Syllable_Modifier, which is the same category as 1A7B..1A7C, which look as though they should probably be restricted to a single occurrence. Sure enough, the USE cluster regex only accepts a single (optional) trailing FM.

(Also, the USE model appears to only recognize Consonant Final Modifier (FM) as appearing at the top of the cluster, whereas U+1A7F appears below.)

A workaround for the time being may be to override the category of 1A7F to treat it as a Final Consonant Below instead of Modifier; AFAICS this won't clash with anything else Tai Tham is using, and will allow it to be repeated.

PR #1720, opened 2019-05-19
[use] Allow multiple FMs in a cluster

dscorbett:
Instead of overriding certain characters to F (#525, #1399), this PR simply allows multiple FMs in a cluster. FM is split into FMAbv, FMBlw, and FMPst. Only one FMPst is allowed, because all FMPsts are non-marks, so the USE spec never inserts a dotted circle before one: if another FM precedes an FMPst, the FMPst just becomes an independent cluster. I didn’t want to deviate from the spec in a case where the current rendering would not get a dotted circle.


behdad:
> Only one FMPst is allowed, because all FMPsts are non-marks, so the USE spec never inserts a dotted circle before one: if another FM precedes an FMPst, the FMPst just becomes an independent cluster. I didn’t want to deviate from the spec in a case where the current rendering would not get a dotted circle.

But then why include FMPst at all?  I don't like the "|" in there.  Feels unjustified.

dscorbett:
Currently, using my new categorization but the old behavior, C FMPst FMPst is two clusters: C FMPst and FMPst. If the machine used FMPst\*, there would be one cluster. If I got rid of FMPst, there would be three clusters. I prefer following the spec, even when it’s a little weird, as long as it isn’t breaking anything, i.e. inserting dotted circles in valid text.

Unlike other \*Pst categories, FMPst contains no marks, which is why it behaves differently than FMAbv and FMBlw. I had considered calling the category FMEnd to make it more obviously different. Would that help?

behdad:
Thanks.

PR #1722, opened 2019-05-21
[subset] Use iterators in glyf/loca subsetting

rsheeter:
Start moving glyf/loca over the shiny new toys


src/hb-ot-glyf-table.hh
```diff
@@ -81,26 +80,170 @@ struct glyf
...
 
-  bool subset (hb_subset_plan_t *plan) const
+  template<typename Iterator,
+	   hb_requires (hb_is_source_of (Iterator, unsigned int))>
+  static bool
+  _add_loca_and_head (hb_subset_plan_t * plan, Iterator padded_offsets)
+  {
+    unsigned int max_offset =
+    + padded_offsets
+    | hb_reduce (hb_max, 0);
```
behdad: Either make this statement a one-liner, or semi-colon on next line.

src/hb-ot-glyf-table.hh
```diff
@@ -81,26 +80,170 @@ struct glyf
...
+    bool use_short_loca = max_offset <= 131070;
+    unsigned int loca_prime_size = padded_offsets.len () * (use_short_loca ? 2 : 4);
+    char *loca_prime_data = (char *) calloc(1, loca_prime_size);
```
behdad: So, padded_offsets includes the offset as well?

Instead of doing the multiply yourself and calling calloc with 1, pass two separate numbers to calloc and let it do the mul.

src/hb-ot-glyf-table.hh
```diff
@@ -81,26 +80,170 @@ struct glyf
...
+
+    if (unlikely (!loca_prime_data)) return false;
+
+    if (use_short_loca)
+      _write_loca <decltype (padded_offsets), HBUINT16> (padded_offsets, 1, hb_array_t<HBUINT16> ((HBUINT16*) loca_prime_data, padded_offsets.len ()));
```
behdad: Function can deduce template args? If not, reorder the function template parameters such that the one you need to provide is first, and then provide just that one.

behdad: Calling .len() on an iterator is not free.  So, cache it in a `unsigned count`.

src/hb-ot-glyf-table.hh
```diff
@@ -81,26 +80,170 @@ struct glyf
...
+    else
+      _write_loca <decltype (padded_offsets), HBUINT32> (padded_offsets, 0, hb_array_t<HBUINT32> ((HBUINT32*) loca_prime_data, padded_offsets.len ()));
+
+    hb_blob_t * loca_blob = hb_blob_create (loca_prime_data,
+					    loca_prime_size,
+					    HB_MEMORY_MODE_READONLY,
```
behdad: Huh.  READONLY?

behdad: Ah.  Doens't matter, sure, but WRITABLE is more appropriate.

src/hb-ot-glyf-table.hh
```diff
@@ -81,26 +80,170 @@ struct glyf
...
+    plan->add_table (HB_OT_TAG_loca, loca_blob);
+    _add_head_and_set_loca_version(plan, use_short_loca);
+
+    hb_blob_destroy (loca_blob);
+    return true;
+  }
+
+  template<typename IteratorIn, typename EntryType, typename IteratorOut,
+	   hb_requires (hb_is_source_of (IteratorIn, unsigned int)),
+	   hb_requires (hb_is_source_of (IteratorOut, EntryType))>
```
behdad: `is_sink_of` for second one.

src/hb-ot-glyf-table.hh
```diff
@@ -81,26 +80,170 @@ struct glyf
...
+  static void
+  _write_loca (IteratorIn it, unsigned right_shift, IteratorOut dest)
+  {
+    unsigned int offset = 0;
+    + it
+    | hb_map ([&] (unsigned int padded_size) {
+      unsigned int result = offset >> right_shift;
```
behdad: We have started using just `unsigned` instead of `unsigned int`.

src/hb-ot-glyf-table.hh
```diff
@@ -81,26 +80,170 @@ struct glyf
...
+
+  template<typename IteratorIn, typename EntryType, typename IteratorOut,
+	   hb_requires (hb_is_source_of (IteratorIn, unsigned int)),
+	   hb_requires (hb_is_source_of (IteratorOut, EntryType))>
+  static void
+  _write_loca (IteratorIn it, unsigned right_shift, IteratorOut dest)
+  {
+    unsigned int offset = 0;
+    + it
+    | hb_map ([&] (unsigned int padded_size) {
```
behdad: `&offset` instead of `&`?

src/hb-ot-glyf-table.hh
```diff
@@ -81,26 +80,170 @@ struct glyf
...
+      unsigned int result = offset >> right_shift;
+      DEBUG_MSG(SUBSET, nullptr, "loca entry offset %d shifted %d", offset, result);
+      offset += padded_size;
+      return result;
+    })
+    | hb_sink (dest);
```
behdad: Semi-colon next line.

src/hb-ot-glyf-table.hh
```diff
@@ -81,26 +80,170 @@ struct glyf
...
-    if (hb_subset_glyf_and_loca (plan, &use_short_loca, &glyf_prime, &loca_prime)) {
-      success = success && plan->add_table (HB_OT_TAG_glyf, glyf_prime);
-      success = success && plan->add_table (HB_OT_TAG_loca, loca_prime);
-      success = success && _add_head_and_set_loca_version (plan, use_short_loca);
-    } else {
-      success = false;
+    TRACE_SERIALIZE (this);
+
+    + it
+    | hb_apply ( [&] (const SubsetGlyph& _) { _.serialize (c, plan); });
```
behdad: Semi-colon next line.

behdad: Capture `=` instead of `&`.

src/hb-ot-glyf-table.hh
```diff
@@ -81,26 +80,170 @@ struct glyf
...
+
+    return_trace (true);
+  }
+
+  bool subset (hb_subset_context_t *c) const
+  {
+    TRACE_SUBSET (this);
+
+    glyf *glyf_prime = c->serializer->start_embed <glyf> ();
+    if (unlikely (!glyf_prime)) return_trace (false);
```
behdad: No.  Remember subset return value is not enough to record error.  `if (unlikely (!c->serializer->check_succuess (glyf_prime))) return_trace (false);`.  In other places as well I think.

src/hb-ot-glyf-table.hh
```diff
@@ -81,26 +80,170 @@ struct glyf
...
+    // unpadded, hints removed if so requested
+    // If we fail to process a glyph we produce an empty (0-length) glyph
+    hb_vector_t<SubsetGlyph> glyphs;
+    _populate_subset_glyphs (c->plan, &glyphs);
+
+    glyf_prime->serialize (c->serializer, hb_iter (glyphs), c->plan);
+
+    hb_vector_t<unsigned int> padded_offsets;
+    + hb_iter (glyphs)
+    | hb_map ([&] (const SubsetGlyph& _) { return _.padded_size(); })
```
behdad: This can be just: `| hb_map(&SubsetGlyph::padded_size)`

src/hb-ot-glyf-table.hh
```diff
@@ -81,26 +80,170 @@ struct glyf
...
+    | hb_sink (padded_offsets);
```
behdad: Semi-colon next line.  Maybe I give up on  that. :)

src/hb-ot-glyf-table.hh
```diff
@@ -81,26 +80,170 @@ struct glyf
...
+
+    // loca ends with a final entry == last offset (+0)
+    padded_offsets << 0;
```
behdad: actually.  I think you don't need to write these to a vector.  Skip the 0 here.  Pass the iterator as is to the funciton.  In the function, write out a 0 first, then add the post-sum values.  Makes sense?

src/hb-ot-glyf-table.hh
garretrieger: Should be able to use hb_map (&SubsetGlyph::padded_size) here.

src/hb-ot-glyf-table.hh
```diff
@@ -81,26 +80,170 @@ struct glyf
...
-  bool subset (hb_subset_plan_t *plan) const
+  template<typename Iterator,
+	   hb_requires (hb_is_source_of (Iterator, unsigned int))>
+  static bool
+  _add_loca_and_head (hb_subset_plan_t * plan, Iterator padded_offsets)
+  {
+    unsigned int max_offset =
+    + padded_offsets
+    | hb_reduce (hb_max, 0);
+    bool use_short_loca = max_offset <= 131070;
```
khaledhosny: Nitpick: IMHO, hex is more readable/obvious here. I can easily guess what `<= 0x1FFFE` (or even better `< 0x1FFFF`) is about, but not the decimal number used.

src/hb-ot-glyf-table.hh
rsheeter: 1-lined

src/hb-ot-glyf-table.hh
rsheeter: done

src/hb-ot-glyf-table.hh
rsheeter: reduced template args, had trouble getting rid of all of them

src/hb-ot-glyf-table.hh
rsheeter: I had in mind it isn't meant to be written to anymore / this hb_blob isn't meant for writing. Adjusted to writable.

src/hb-ot-glyf-table.hh
rsheeter: done, ty

src/hb-ot-glyf-table.hh
rsheeter: Modified, curious why?

src/hb-ot-glyf-table.hh
rsheeter: added by-val auto + &offset

src/hb-ot-glyf-table.hh
rsheeter: fancy!

src/hb-ot-glyf-table.hh
rsheeter: It looks cooler, I like it lolz

src/hb-ot-glyf-table.hh
rsheeter: I like passing the iterator. Writing a 0 first I'm not quite following. I re-wrote the last value instead.

src/hb-ot-glyf-table.hh
rsheeter: Done

src/hb-ot-glyf-table.hh
rsheeter: Ty, I like `< 0x1FFFF`

src/hb-ot-glyf-table.hh
rsheeter: Fixed, ty. Will take a skim through for other places.

src/hb-ot-glyf-table.hh
```diff
@@ -81,26 +80,171 @@ struct glyf
...
+    bool result = plan->add_table (HB_OT_TAG_loca, loca_blob)
+		  && _add_head_and_set_loca_version(plan, use_short_loca);
+
+    hb_blob_destroy (loca_blob);
+    return result;
+  }
+
+  template<typename EntryType, typename IteratorIn, typename IteratorOut,
+	   hb_requires (hb_is_source_of (IteratorIn, unsigned int)),
+	   hb_requires (hb_is_sink_of (IteratorOut, EntryType))>
```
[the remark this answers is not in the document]
rsheeter: Much nicer, ty. Works locally fingers crossed for bots.

src/hb-ot-glyf-table.hh
```diff
@@ -81,26 +80,169 @@ struct glyf
...
+    unsigned int max_offset = + padded_offsets | hb_reduce (hb_max, 0);
+    unsigned num_offsets = padded_offsets.len () + 1;
+    bool use_short_loca = max_offset < 0x1FFFF;
+    unsigned entry_size = use_short_loca ? 2 : 4;
+    char *loca_prime_data = (char *) calloc (entry_size, num_offsets);
+
+    if (unlikely (!loca_prime_data)) return false;
+
+    if (use_short_loca)
+      _write_loca (padded_offsets, 1, hb_array_t<HBUINT16> ((HBUINT16*) loca_prime_data, num_offsets));
```
behdad: One last nit: using `hb_array` function you can avoid duplicating the type name.

PR #1725, opened 2019-05-24
Fix for fuzzer crash testcase

qxliu76:
Add a check for stringOffSet(uint16) overflow, return early if overflow happens


src/hb-ot-name-table.hh
```diff
@@ -230,6 +228,7 @@ struct name
     TRACE_SANITIZE (this);
     return_trace (c->check_struct (this) &&
 		  likely (format == 0 || format == 1) &&
+		  likely (count <= 16382) &&
```
garretrieger: Can you add a comment explaining how this number is derived and the reason for the check.

src/hb-ot-name-table.hh
```diff
@@ -197,8 +197,6 @@ struct name
 
     if (unlikely (c->ran_out_of_room)) return_trace (false);
 
-    assert (this->stringOffset == c->length ());
```
behdad: Why remove?  If overflow happens, we should detect it earlier and return.

behdad: The line a few lines earlier where we have this:
```
   this->stringOffset = c->length ();
```
It should become something like:
```
  if (unlikely (!c->check_assign (this->stringOffset, c->length ())) return_trace (false);
```

src/hb-ot-name-table.hh
```diff
@@ -230,6 +228,7 @@ struct name
     TRACE_SANITIZE (this);
     return_trace (c->check_struct (this) &&
 		  likely (format == 0 || format == 1) &&
+		  likely (count <= 16382) &&           /*make sure stringOffSet(uint16) won't overflow, 6 bytes (header) + 4 bytes (name record) * count <= 65535*/
```
behdad: We shouldn't need this.

src/hb-ot-name-table.hh
qxliu76: oh, so you don't check in sanitize() for invalid tables but rather do checks in serializer.

qxliu76: sorry, I have a question here: 
we have a task to remove return type of bool from all serialize calls. Is serializer be able to capture this error state in this case?

behdad: Yes, the `check_assign` thing records the failure.

behdad: > oh, so you don't check in sanitize() for invalid tables but rather do checks in serializer.

The job of `sanitize` is only to make sure the rest of the methods to *access* the table don't cause invalid memory access.  We are not in the font law enforcement business :D.

qxliu76: got it. Thanks! :)

Issue #170, opened 2015-11-10
Tai Tham uses subjoined consonants after right-side vowels


roozbehp:
Basically, the regular expression used in Universal Shaping engine doesn't allow subjoined consonants after right-side vowels (visually hanging under them), while Tai Tham seems to be using them.

Specific examples should be available from Martin and Patrick, CC-ed.

/cc @mhosken @pchew-change

behdad:
Shouldn't this be solved at the USE level first?

roozbehp:
Not necessarily. We cannot wait for USE. We know we need to allow the sequences.

behdad:
But I rather not diverge from it.  We don't need to wait for USE spec to be updated, but at least you and Andrew to decide how you want to address it.

pchew-change:
some examples include:

ᩈᩣ᩠ᨾ (ᩈ | ᩣ | ᩠ | ᨾ) U+1A48 U+1A63 U+1A60 U+1A3E
(‘three’)

ᩃ᩶ᩣ᩠ᨶ (ᩃ | ᩣ | ᩠ | ᨶ) U+1A43 U+1A76 U+1A63 U+1A60 U+1A36
(‘myriad, ten-thousand’)

ᨷ᩶ᩤ᩠ᨶ (ᨷ | ᩶ | ᩤ | ᩠ | ᨶ) U+1A37 U+1A76 U+1A64 U+1A60 U+1A36
(‘home, village’)

ᨩ᩶ᩣ᩠ᨦ (ᨩ | ᩶ | ᩣ |  ᩠ | ᨦ) U+1A29 U+1A76 U+1A63 U+1A60 U+1A26
(‘elephant’)

ᩃᩖᩣ᩠ᨿ (ᩃ | ᩖ | ᩣ |  ᩠ | ᨿ) U+1A43 U+1A56 U+1A63 U+1A60 U+1A3F
(‘many’)

ᨠᩣ᩠ᩁ (ᨠ | ᩣ |  ᩠ | ᩁ) U+1A20 U+1A63 U+1A60 U+1A41
(‘the process of…, the activity of…, an enactment of…, the state of…’ )

ᩈᩣ᩠ᩅ (ᩈ | ᩣ |  ᩠ | ᩅ) U+1A48 U+1A63 U+1A60 U+1A45
(‘young woman’)

Please let me know if you need more.

pchew-change:
Rendered versions in the attached png.
[...]

PR #1731, opened 2019-05-26
[use] Allow U+1A60 TAI THAM SIGN SAKOT after a vowel sign

dscorbett:
This adds a new class, Sk, as described in #170. Its only member is U+1A60 TAI THAM SIGN SAKOT because other scripts’ invisible stackers are not used after vowel signs. I did not allow CM after Sk B because Andrew Glass’s proposal is ambiguous on that subject; it doesn’t make any practical difference because Tai Tham doesn’t currently have any CMs.


src/hb-ot-shape-complex-use-machine.rl
```diff
@@ -92,7 +93,7 @@ FMAbv	= 45; # CONS_FINAL_MOD	UIPC = Top
 FMBlw	= 46; # CONS_FINAL_MOD	UIPC = Bottom
 FMPst	= 47; # CONS_FINAL_MOD	UIPC = Not_Applicable
 
-h = H | HVM; # https://github.com/harfbuzz/harfbuzz/issues/1102
+h = H | HVM | Sk;
```
behdad: Maybe keep the URL, and add the one for this issue as well?

behdad:
Merged.  Thanks.

PR #1735, opened 2019-05-30
[subset] Hmtx/vmtx tables to use subset2 and new iterator frameworks

qxliu76:


src/hb-ot-hmtx-table.hh
```diff
@@ -86,74 +92,68 @@ struct hmtxvmtx
...
+    _mtx.init (c->plan->source);
+    unsigned num_advances = _mtx.num_advances_for_subset (c->plan);
+    
+    auto it = 
+    + hb_range (c->plan->num_output_glyphs ())
+    | hb_map ([c, &_mtx] (unsigned _)
+	{
+	  hb_codepoint_t old_gid;
+	  if (c->plan->old_gid_for_new_gid (_, &old_gid))
+            return hb_pair<unsigned, unsigned> (_mtx.get_advance (old_gid), _mtx.get_side_bearing (old_gid));
```
behdad: Do you actually need to specify the template types here?  I'd assume not.

src/hb-ot-hmtx-table.hh
```diff
@@ -86,74 +92,68 @@ struct hmtxvmtx
...
+          else
+	    return hb_pair<unsigned, unsigned> (0, 0);
```
behdad: Same here.  Just pass `0u` instead of `0`.

src/hb-subset.cc
```diff
@@ -67,7 +67,7 @@ template<typename TableType>
 static bool
 _subset2 (hb_subset_plan_t *plan)
 {
-  bool result = true;
+  bool result = false;
```
behdad: Why this?

src/hb-ot-hmtx-table.hh
```diff
@@ -51,6 +51,12 @@ struct LongMetric
   FWORD		sb; /* Leading (left/top) side bearing. */
   public:
   DEFINE_SIZE_STATIC (4);
+  LongMetric (unsigned ad,
+              unsigned side_bearing)
+  {
+    advance = ad;
+    sb = side_bearing;
+  }
```
behdad: We actually don't currently have constructors in any of our font structs.  Previously that was a requirement of C++<11.  I still want to keep it that way.  So just assign to the members outside for now.  Thanks.

src/hb-ot-hmtx-table.hh
```diff
@@ -86,74 +92,68 @@ struct hmtxvmtx
...
+    unsigned idx = 0;
+    + it
+    | hb_apply ([c, &idx, num_advances] (const hb_item_type<Iterator>& _)
+                {
+                  if (idx < num_advances) 
+                  {
+                    LongMetric lm (_.first, _.second);
+                    if (unlikely (!c->embed<LongMetric> (&lm)))
+                      return;
+                  } else {
```
behdad: Braces on new line.

src/hb-ot-hmtx-table.hh
```diff
@@ -86,74 +92,68 @@ struct hmtxvmtx
...
-    char * dest_pos = (char *) dest;
+    unsigned idx = 0;
+    + it
+    | hb_apply ([c, &idx, num_advances] (const hb_item_type<Iterator>& _)
+                {
+                  if (idx < num_advances) 
+                  {
+                    LongMetric lm (_.first, _.second);
+                    if (unlikely (!c->embed<LongMetric> (&lm)))
+                      return;
```
behdad: Return on same line.

src/hb-subset.cc
qxliu76: this is to be consistent with _subset(), otherwise there's a unit test failure.

PR #1736, opened 2019-05-31
[subset] Correct calculation of max offset for loca

rsheeter:
Fixes bug where we wouldn't use long loca when we should.

Don't write head/loca if we know we are bailing out.


behdad:
Lgtm if bots happy.  WOuld be nice to have test for but maybe too large.

PR #1739, opened 2019-06-01
Update qsort_r and use it everywhere instead libc's qsort

### LINKED ISSUE #593: hb-dsalgs.hh
Just filing this to dump what I like to do.  I don't have time to do it right now, just logging for now.

We have hb_sort_r() and hb_bsearch_r() now.  Rename hb_sort_r() to hb_qsort_r(), and then add native implementation of hb_qsort() and hb_bsearch() as well as use them instead of system ones.  Since qsort is recursive, it won't inline well.  For bsearch, try templatizing it and see if it can be vectorized for certain types.

Add in-place mergesort (I have acquired a relicensed version of https://github.com/liuxinyu95/AlgoXY/blob/algoxy/sorting/merge-sort/src/mergesort.c from its author). Make it a `_r` version.  Use it instead of our current insertion-sort.

For both qsort and mergesort, add a version that takes a swap() function, such that we can sort two parallel arrays, like hb_buffer_t::sort().  Also, use bsearch_r() for CmapSubtableFormat4::accelerator_t::get_glyph_func().

In general, consolidate all search and sort functions.

Then add hb_prealloced_array_t and hb_lockable_set_t from hb-private.hh to hb-dsalgs.hh.  At some point implement a hashmap. Maybe public? Integer->integer only?

ebraminio:
Let's see, fixes #593

### DIFF
--- a/src/hb-aat-map.hh
+++ b/src/hb-aat-map.hh
@@ -66,7 +66,7 @@ struct hb_aat_map_builder_t
     hb_aat_layout_feature_selector_t  setting;
     unsigned  seq; /* For stable sorting only. */
 
-    HB_INTERNAL static int cmp (const void *pa, const void *pb)
+    HB_INTERNAL static int cmp (const void *pa, const void *pb, void *arg HB_UNUSED)
     {
       const feature_info_t *a = (const feature_info_t *) pa;
       const feature_info_t *b = (const feature_info_t *) pb;
--- a/src/hb-algs.hh
+++ b/src/hb-algs.hh
@@ -638,7 +638,7 @@ hb_bsearch (const void *key, const void *base,
 
 /*
 
-hb_sort_r function to be exported.
+hb_qsort function to be exported.
 
 Parameters:
   base is the array to be sorted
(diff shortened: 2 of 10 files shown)


behdad:
I wonder what the size implications are.

ebraminio:
> I wonder what the size implications are.

Fair. What compile flags you use for comparison?

ebraminio:
Ok merged the two as your suggestion I guess so no size implication anymore

ebraminio:
From https://github.com/harfbuzz/harfbuzz/pull/1741#issuecomment-497963227

> I'm fine w doing this but also for both bsearch and qsort, if HB_OPTIMIZE_SIZE is enabled we should introduce overload without extra args to call the C lib one. Curious what the size diff will be.

Specially for qsort that there is a qsort_r and it works if I replace sort_r_simple with it (so I guess there is some bug sort_r_simple already)

> Maybe combine the two?

Sure

ebraminio:
PTAL

Sad that test issues gone with sort_r update but even more sad that [apparently] we have a broken sort_r in use in the wild so should publish this quickly.

Please note that our `inline` is not equal to sort_r's [inline](https://github.com/noporpoise/sort_r/blob/master/sort_r.h#L25-L43)s.

behdad:
> Sad that test issues gone with sort_r update but even more sad that [apparently] we have a broken sort_r in use so should publish this quickly I think (will do a bisect to recheck this claim)

What are you talking about?  What's broken?

> Please note that our `inline` is not equal to sort_r's [inline](https://github.com/noporpoise/sort_r/blob/master/sort_r.h#L25-L43)s.

How is that relevant?

behdad:
I don't like this.  It's not even clear to me what you are doing with the actual sort changes.  Extra, unused, args to cmp functions are not acceptable.

Sorry but I'm closing this.  Please reopen small PRs.  First, one to modify hb_qsort to prepare it for varargs, and a later one to port existing uses to it.

behdad:
> Fair. What compile flags you use for comparison?

`-Os -DHB_TINY` with gcc.

ebraminio:
> Sorry but I'm closing this. Please reopen small PRs. First, one to modify hb_qsort to prepare it for varargs, and a later one to port existing uses to it.

But that's done to not impact harbuzz binary size.

> I don't like this. It's not even clear to me what you are doing with the actual sort changes. Extra, unused, args to cmp functions are not acceptable.

Well, making possible not go for the template duplication by unifying them, I think that was what you meant from "merge" them! :)

behdad:
That's why I always like discussing and agreeing on a plan before making changes :).

(My bad for not communicating better.  By merging two I actually meant merging this PR and the bsearch PR.  I'm sorry for confusing you.  But then again, I feel like if I don't reply immediately, even when not convenient for me, you keep going farther in directions that I might not like and will end up wasting more of your time.)

So, I see three ways this can be compiled:

  - Default should be qsort / bsearch inlined and fastest, with little regard to size,

  - On systems that provide the functions we need, the size-optimized build should defer to system qsort / qsort_r.  The latter might not be worth detecting because of the confusing history of that function.  For bsearch I'm hesitant to move to system because how hot bsearch it is in our profiles.  But then again, maybe we can keep one or two bsearch'es inlined and call system on the rest.  For making these decisions I like to see the binary size overhead of both qsort variations and bsearch variants.

  - On systems that don't provide libc, for size reasons we might want to compile internally but without template bloating / inlining.

It should be doable to support all three options.  But I like to fully understand which ones we are aiming for first.

ebraminio:
>>Please note that our inline is not equal to sort_r's inlines.

>How is that relevant?

glibc and sort_r both are using real inline to prefer speed over size, if you want completely match them we should have some always inline macro instead our ineffective `inline`.

>> Sad that test issues gone with sort_r update but even more sad that [apparently] we have a broken sort_r in use so should publish this quickly I think (will do a bisect to recheck this claim)

>What are you talking about? What's broken?

Roughly sure our current sort_r copy is broken, I think I can show that somewhere else

behdad:
> Roughly sure our current sort_r copy is broken, I think I can show that somewhere else

I probably broke it in this:
commit 3a9fa8c026bf28bf87e20ec95327f74fd7070b74
Author: Behdad Esfahbod <behdad@behdad.org>
Date:   Sat Nov 10 01:56:37 2018 -0500

    [qsort] Fix O(N^2) behavior if all array elements are the same
    
    Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=11327
    
    Reported as https://github.com/noporpoise/sort_r/issues/7

ebraminio:
> It should be doable to support all three options. But I like to fully understand which ones we are aiming for first.

Different platform sort_r are not portable and reliable I believe, see [1](https://github.com/noporpoise/sort_r/blob/master/sort_r.h#L246) and [2](https://github.com/noporpoise/sort_r/blob/master/sort_r.h#L292-L304), what know what other libc's is doing, glibc way or bsd... We will need sort_r anyway and now that we have `qsort_r ` lets have that as the canonical sort implementation (to also not have bugs like #1751), just like what `qsort_s` means in MSVC CRT, it is the security enhanced version of `qsort` and it provides context whether needed or not.

PR #1741, opened 2019-06-01
Use a unified and internal bsearch

ebraminio:
Unintentionally but part of #593, I didn't know its essence at the time! :)

### DIFF
--- a/src/hb-aat-layout.cc
+++ b/src/hb-aat-layout.cc
@@ -139,11 +139,11 @@ hb_aat_layout_find_feature_mapping (hb_tag_t tag)
   return nullptr;
 #endif
 
-  return (const hb_aat_feature_mapping_t *) bsearch (&tag,
-						     feature_mappings,
-						     ARRAY_LENGTH (feature_mappings),
-						     sizeof (feature_mappings[0]),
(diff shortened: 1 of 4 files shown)


behdad:
I'm fine w doing this but also for both bsearch and qsort, if HB_OPTIMIZE_SIZE is enabled we should introduce overload without extra args to call the C lib one.  Curious what the size diff will be.

behdad:
Maybe combine the two?

ebraminio:
> I'm fine w doing this but also for both bsearch and qsort, if HB_OPTIMIZE_SIZE is enabled we should introduce overload without extra args to call the C lib one. Curious what the size diff will be.

Actually this brought to my attention as I wanted reduce harfbuzzjs size, so I see that as unnecessary complication as I have to do that only when there is no extra variable.

> Maybe combine the two?

Which two?

ebraminio:
> Actually this brought to my attention as I wanted reduce harfbuzzjs size,

And doing so undermines my wish on writing some slim libc to reduce harfbuzzjs wasm size even more.

ebraminio:
Merged this as it didn't have size implication and works correctly apparently, now #1739 remains which I don't why fails even

PR #1744, opened 2019-06-03
test: Use nullptr in C++ code.

waywardmonkeys:


ebraminio:
Correct, thanks. OK now that we have C++11 as we used to polyfill that before that.

PR #1748, opened 2019-06-04
[hb-ot-post-table]: Fixed memory leak in get_glyph_from_name()

raminfp:
that on line 163, if the block is not entered, the gids block is not freed and that is indeed a memory leak.

Sign-off-by: Ramin Farajpour Cami <ramin.blackhat@gmail.com>

### DIFF
--- a/src/hb-ot-post-table.hh
+++ b/src/hb-ot-post-table.hh
@@ -170,6 +170,8 @@ struct post
       hb_bytes_t st (name, len);
       const uint16_t *gid = (const uint16_t *) hb_bsearch (hb_addressof (st), gids, count,
 							   sizeof (gids[0]), cmp_key, (void *) this);
+      free(gids);
+
       if (gid)
       {
 	*glyph = *gid;


src/hb-ot-post-table.hh
```diff
@@ -170,6 +170,8 @@ struct post
       hb_bytes_t st (name, len);
       const uint16_t *gid = (const uint16_t *) hb_bsearch (hb_addressof (st), gids, count,
 							   sizeof (gids[0]), cmp_key, (void *) this);
+      free(gids);
```
ebraminio: This is intentional as the allocated array is actually stored in `gids_sorted_by_name.cmpexch (nullptr, gids)` line for future uses, AFAIK

raminfp: We see use `gids` allocated in the `hb_bsearch()` function, CC: @behdad

behdad: What @ebraminio said.  This is freed when the accelerator is fini()ed eventually.

PR #1750, opened 2019-06-04
Always use C++11 atomics.

waywardmonkeys:
Now that we require C++11, use C++11 atomics when threading is
enabled.


waywardmonkeys:
I left in the case for `#elif !defined(HB_NO_MT)` ... which can probably be removed since we require C++11 and that block shouldn't be able to be hit.

I also left in the configure check for the Intel primitives because they're still used in the mutex code. That also can probably be simplified in the future.

behdad:
I prefer to keep them in for a while longer.  Maybe we can switch the default?

Also, I remember liking the compiler intrinsics over the c++11 ones since we don't link to libc++.

waywardmonkeys:
I'll revise this and force push something soon.

behdad:
As I said, I prefer to wait a few months before doing this.

ebraminio:
> I'll revise this and force push something soon.

That's ok, I applied some changes myself instead leaving comment, it is possible in GitHub UI and that's feels good also sometimes, feel free to squash those minor changes into yours and fix msvc bot in the meanwhile if you like to.

See https://github.com/harfbuzz/harfbuzz/commit/1bada656a86e9cb27d4a6b9fcc50748f0bd9c1d9 also, the !HB_NO_MT is broken for some release and nobody has noticed it, signalling it may can go away.

Also please be careful to not break our tsan tester if want to change the default as it relies to pthread and doesn't work with C++11's locks yet IIRC.

behdad:
> See [1bada65](https://github.com/harfbuzz/harfbuzz/commit/1bada656a86e9cb27d4a6b9fcc50748f0bd9c1d9) also, the !HB_NO_MT is broken for some release and nobody has noticed it, signalling it may can go away.

Sure we don't expect that code path to be used.  But the commit you point out to cannot be used to justify that nobody is using it.

khaledhosny:
TeX Live builds build with `-DHB_NO_MT`, so it certainly wasn’t broken.

ebraminio:
> TeX Live builds with -DHB_NO_MT, so it certainly wasn’t broken.

Oh, it is not that one which Firefox found broken https://github.com/harfbuzz/harfbuzz/commit/179570d4cae9aefcd91d540b8865a81d569b3358 but doesn't differ that much also as all I meant was having less CI tested build configuration lead to such issues, which, aren't that important also.

behdad:
I'm doing this now.

PR #1757, opened 2019-06-06
[subset] Move OS/2 to subset2.

garretrieger:


src/hb-ot-os2-table.hh
```diff
@@ -218,6 +209,15 @@ struct OS2
   font_page_t get_font_page () const
   { return (font_page_t) (version == 0 ? fsSelection & 0xFF00 : 0); }
 
+  unsigned get_size () const
+  {
+    unsigned result = min_size;
+    if (unlikely (version >= 1)) result += v1X.get_size ();
```
behdad: These unlikely's are not justified.  Unlikely should be used for exceptional paths.

garretrieger: Fixed.

behdad:
lgtm.

PR #1760, opened 2019-06-07
Update and use internal qsort everywhere

### LINKED ISSUE #1751: Our hb_sort_r is br0ken
```patch
diff --git a/src/hb-ot-name-table.hh b/src/hb-ot-name-table.hh
index 68d7b3a6..81620f6c 100644
--- a/src/hb-ot-name-table.hh
+++ b/src/hb-ot-name-table.hh
@@ -144,7 +144,7 @@ _hb_ot_name_entry_cmp_key (const void *pa, const void *pb)
 }
 
 static int
-_hb_ot_name_entry_cmp (const void *pa, const void *pb)
+_hb_ot_name_entry_cmp (const void *pa, const void *pb, void *arg HB_UNUSED)
 {
   /* Compare by name_id, then language, then score, then index. */
 
@@ -259,7 +259,9 @@ struct name
        entry->entry_index = i;
       }
 
-      this->names.qsort (_hb_ot_name_entry_cmp);
+      hb_sort_r (this->names.arrayZ, this->names.length,
+                this->names.item_size,
+                _hb_ot_name_entry_cmp, nullptr);
       /* Walk and pick best only for each name_id,language pair,
        * while dropping unsupported encodings. */
       unsigned int j = 0;

```

It breaks `test-ot-name` test and will be fixed by applying this

```patch
diff --git a/src/hb-algs.hh b/src/hb-algs.hh
index 0b34f710..e5ccb3c6 100644
--- a/src/hb-algs.hh
+++ b/src/hb-algs.hh
@@ -675,7 +675,7 @@ static inline void sort_r_simple(void *base, size_t nel, size_t w,
                                 void *arg)
 {
   char *b = (char *)base, *end = b + nel*w;
-  if(nel < 7) {
+  if(true || nel < 7) {
     /* Insertion sort for arbitrarily small inputs */
     char *pi, *pj;
     for(pi = b+w; pi < end; pi += w) {
```
Or using qsort_r of glibc

Personal conclusion: That's why we need a unified and duplication free qsort! :)

ebraminio:
#593 related, closes #1739 and fixes #1751

Using internal implementation for both:
`make -Cbuild clean && make -Cbuild CXXFLAGS="-Os" -j9` with the sort_r_simple for both
`size build/src/.libs/libharfbuzz.so.0.20501.0`
`732670	  12784	    512	 745966	  b61ee	build/src/.libs/libharfbuzz.so.0.20501.0`

Using system qsort but internal for qsort_r:
`730030	  12792	    512	 743334	  b57a6	build/src/.libs/libharfbuzz.so.0.20501.0`

Using both system's qsort_*s, no internal implementation:
`729434	  12800	    512	 742746	  b555a	build/src/.libs/libharfbuzz.so.0.20501.0`

I still like #1739 more personally as it will use an internal and well tested `qsort` implementation and can be more useful for my case and ever more a code become hot chances are PGO tunes it even better but adds bloat to comparison functions so up to you to decide.

Size of #1739:
`732702	  12784	    512	 745998	  b620e	build/src/.libs/libharfbuzz.so.0.20501.0`

### DIFF
--- a/src/hb-algs.hh
+++ b/src/hb-algs.hh
@@ -628,115 +628,213 @@ hb_bsearch (const void *key, const void *base,
 
 
 /* From https://github.com/noporpoise/sort_r
- * With following modifications:
- *
- * 10 November 2018:
- * https://github.com/noporpoise/sort_r/issues/7
- */
+   Feb 5, 2019 (c8c65c1e)
+   Modified to support optional argument using templates */
 
 /* Isaac Turner 29 April 2014 Public Domain */
 
 /*
-
-hb_sort_r function to be exported.
-
+hb_qsort function to be exported.
 Parameters:
   base is the array to be sorted
   nel is the number of elements in the array
   width is the size in bytes of each element of the array
   compar is the comparison function
-  arg is a pointer to be passed to the comparison function
+  arg (optional) is a pointer to be passed to the comparison function
 
-void hb_sort_r(void *base, size_t nel, size_t width,
-               int (*compar)(const void *_a, const void *_b, void *_arg),
-               void *arg);
+void hb_sort(void *base, size_t nel, size_t width,
+             int (*compar)(const void *_a, const void *_b, [void *_arg]),
+             [void *arg]);
 */
 
+#define SORT_R_SWAP(a,b,tmp) ((tmp) = (a), (a) = (b), (b) = (tmp))
 
-/* swap a, b iff a>b */
-/* __restrict is same as restrict but better support on old machines */
-static int sort_r_cmpswap(char *__restrict a, char *__restrict b, size_t w,
-			  int (*compar)(const void *_a, const void *_b,
-					void *_arg),
-			  void *arg)
+/* swap a and b */
+/* a and b must not be equal! */
+static inline void sort_r_swap(char *__restrict a, char *__restrict b,
+                               size_t w)
 {
   char tmp, *end = a+w;
-  if(compar(a, b, arg) > 0) {
-    for(; a < end; a++, b++) { tmp = *a; *a = *b; *b = tmp; }
+  for(; a < end; a++, b++) { SORT_R_SWAP(*a, *b, tmp); }
+}
+
+/* swap a, b iff a>b */
+/* a and b must not be equal! */
+/* __restrict is same as restrict but better support on old machines */
+template <typename ...Ts>
+static inline int sort_r_cmpswap(char *__restrict a,
+                                 char *__restrict b, size_t w,
+                                 int (*compar)(const void *_a,
+                                               const void *_b,
+                                               Ts... _ds),
+                                 Ts... ds)
+{
+  if(compar(a, b, ds...) > 0) {
+    sort_r_swap(a, b, w);
     return 1;
   }
   return 0;
 }
 
+/*
+Swap consecutive blocks of bytes of size na and nb starting at memory addr ptr,
+with the smallest swap so that the blocks are in the opposite order. Blocks may
+be internally re-ordered e.g.
+  12345ab  ->   ab34512
+  123abc   ->   abc123
+  12abcde  ->   deabc12
+*/
+static inline void sort_r_swap_blocks(char *ptr, size_t na, size_t nb)
+{
+  if(na > 0 && nb > 0) {
+    if(na > nb) { sort_r_swap(ptr, ptr+na, nb); }
+    else { sort_r_swap(ptr, ptr+nb, na); }
+  }
+}
+
+/* Implement recursive quicksort ourselves */
 /* Note: quicksort is not stable, equivalent values may be swapped */
+template <typename ...Ts>
 static inline void sort_r_simple(void *base, size_t nel, size_t w,
-				 int (*compar)(const void *_a, const void *_b,
-					       void *_arg),
-				 void *arg)
+                                 int (*compar)(const void *_a,
+                                               const void *_b,
+                                               Ts... _ds),
+                                 Ts... ds)
 {
   char *b = (char *)base, *end = b + nel*w;
-  if(nel < 7) {
+
+  /* for(size_t i=0; i<nel; i++) {printf("%4i", *(int*)(b + i*sizeof(int)));}
+  printf("\n"); */
+
+  if(nel < 10) {
     /* Insertion sort for arbitrarily small inputs */
     char *pi, *pj;
     for(pi = b+w; pi < end; pi += w) {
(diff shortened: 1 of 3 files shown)


behdad:
Please use the `size` command on binaries, not `wc`.

src/hb-algs.hh
```diff
@@ -628,115 +628,211 @@ hb_bsearch (const void *key, const void *base,
...
 
 static inline void
-hb_sort_r (void *base, size_t nel, size_t width,
-	   int (*compar)(const void *_a, const void *_b, void *_arg),
-	   void *arg)
+hb_qsort (void *base, size_t nel, size_t width,
+	  int (*compar)(const void *_a, const void *_b))
 {
-    sort_r_simple(base, nel, width, compar, arg);
+#if defined(HAVE_QSORT) || defined(__OPTIMIZE_SIZE__)
```
behdad: Did you mean `&&` here?

src/hb-algs.hh
```diff
@@ -628,115 +628,211 @@ hb_bsearch (const void *key, const void *base,
...
+    sort_r_simple(end-(pre-pr), (pre-pr)/w, w, compar, ds...);
   }
 }
 
 static inline void
-hb_sort_r (void *base, size_t nel, size_t width,
-	   int (*compar)(const void *_a, const void *_b, void *_arg),
-	   void *arg)
+hb_qsort (void *base, size_t nel, size_t width,
+	  int (*compar)(const void *_a, const void *_b))
```
behdad: Another option that I was thinking was to make this `hb_enable_if()`-gated.  But I'm fine with they way you did it.

src/hb-algs.hh
```diff
@@ -628,115 +628,211 @@ hb_bsearch (const void *key, const void *base,
...
+}
+
+static inline void
+hb_qsort (void *base, size_t nel, size_t width,
+	  int (*compar)(const void *_a, const void *_b, void *_arg),
+	  void *arg)
+{
+#if defined(HAVE_GNU_QSORT_R) || \
+    (defined(__OPTIMIZE_SIZE__) && defined(HAVE_GLIB) && \
+     defined(__GLIBC__) && ((__GLIBC__ < 2) || (__GLIBC__ == 2 && __GLIBC_MINOR__ < 8)))
```
behdad: Huh.  What's does GLIB have to do here?!!!

behdad:
Thanks.  Left some comments.  Please address and report `size` sizes. :)  Also re `qsort_r`, either make it automatically pick it up (autotools, etc), or don't bother.

src/hb-algs.hh
ebraminio: Ops, ok

src/hb-algs.hh
ebraminio: Lets remove it, all the sane libc's will have qsort anyway, instead having HB_DONT_USE_SYSTEM_QSORT for harfbuzzjs I guess

behdad: Do the measurement first.  If has no effect on tiny build size, I don't care not calling system one.

ebraminio: updated the numbers

behdad:
> I still like #1739 more personally as it will use an internal and well tested `qsort` implementation and can be more useful for my case and ever more a code become hot chances are PGO tunes it even better but adds bloat to comparison functions so up to you to decide.

I don't understand.  Clarify what's the difference please.

Anyway, let's clean this up and get it in.  I'm also curious to see the size impact on `HB_TINY` builds.

ebraminio:
> I don't understand. Clarify what's the difference please.

Well, for harfbuzzjs use it would be just one canonical implementation that I don't have to implement in a libc and, on a completely different clients that can do PGO, compilers can do all its tricks if found it hot, but I am not sure why I'm getting the worse size there. Nothing to worry about, I can continue that later or even can put a bubble sort in harfbuzzjs libc, that doesn't matter really and sorry bringing it up again.

For now, I will just need some macro to disable use of system qsort, if it makes sense.

> Anyway, let's clean this up and get it in. I'm also curious to see the size impact on HB_TINY builds.

Sure. I've updated the top post for the number whether internal qsort is used or not.

Before the change on HB_TINY but only sort_r update is applied:

`446181	   8928	    512	 455621	  6f3c5	build/src/.libs/libharfbuzz.so.0.20501.0`

After the change and use of templates on qsort_r, but using system's qsort:

`446181	   8928	    512	 455621	  6f3c5	build/src/.libs/libharfbuzz.so.0.20501.0`

ebraminio:
And here is parent build without sort_r update,

``` 446029	   8928	    512	 455469	  6f32d	build/src/.libs/libharfbuzz.so.0.20501.0```

**Comment by ebraminio** [comment]
So let's merge this?

File: src/hb-algs.hh [hunk 4]
```diff
@@ -628,115 +628,213 @@ hb_bsearch (const void *key, const void *base,
...
 
 static inline void
-hb_sort_r (void *base, size_t nel, size_t width,
-	   int (*compar)(const void *_a, const void *_b, void *_arg),
-	   void *arg)
+hb_qsort (void *base, size_t nel, size_t width,
+	  int (*compar)(const void *_a, const void *_b))
 {
-    sort_r_simple(base, nel, width, compar, arg);
+#ifdef __OPTIMIZE_SIZE__
```
ebraminio: How about making this `defined(__OPTIMIZE__) || defined(__OPTIMIZE_SIZE__)`

behdad: I don't understand.  Our version is supposedly faster because it can inline.  What's the point otherwise?  This is becoming annoying quickly since you seemed focused on simply solving the harfbuzzjs case which doesn't have a system qsort :).

ebraminio: Hehe, not that only, once I have this in I can simply have one line patch harfbuzz in harfbuzzjs for what I need, that is not my concern now, what I like more is to not have a broken qsort_r by using the same code in more places, I don't see the need for Chrome and Firefox to not use this qsort for size reasons but like to have this qsort in our local development and CI tested.

ebraminio: At the end whatever you decide, I see this in a no hurting state at least and fixes qsort_r logic and tests it.

behdad:
Merge it and we touch if needed.

For future, and I said this multiple times, I would have wanted you to do a first PR to just update the qsort to new upstream to fix the bug, then all the rest.  But am fine with this now.

ebraminio:
> I would have wanted you to do a first PR to just update the qsort to new upstream to fix the bug, then all the rest.

Correct. And it made sense also as in case of revert we at least had the fixes :/ Thanks

Issue #1764, opened 2019-06-09
Suggestions for HB_TINY additions


ebraminio:
Would be nice if we had some definition for removing use of `setlocale` and disable `hb_blob_create_from_file` use of fopen/fclose/...

Also not sure why we have a round [polyfill](https://github.com/harfbuzz/harfbuzz/blob/3c240bd3dc0aaca38154da555d0aef350da62ee6/src/hb.hh#L518) but isn't used anywhere or instead roundf, probably should be deleted altogether.

ebraminio:
Also would be nice to have `#define roundf __builtin_roundf` somewhere inside the project, probably guarded with `__has_builtin(__builtin_roundf)` or something

ebraminio:
Some `#if defined(__OPTIMIZE_SIZE__) || defined(HB_USE_INTERNAL_QSORT)` https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-algs.hh#L846 is needed

ebraminio:
801d93f

behdad:
Right.  Let's rename NO_FILE_READ to NO_OPEN and then I add these to TINY.

PR #1765, opened 2019-06-10
Remove round polyfill

ebraminio:
Added in 01dff1e and 19256bef, this was targeted at older
msvc versions that don't support C99 but now as we require
C++11 we don't target places those envs thus removing this.

(which apparently was unused anyway as we use roundf not round?)
(#1764 related)


src/hb.hh
```diff
@@ -515,20 +515,6 @@ typedef uint64_t hb_vector_size_impl_t;
 #define VAR 1
```
ebraminio: Not related to this, would be nice to get rid of this also sometime and replace all the uses with UnsizedArrayOf

behdad: Most uses of this should be replaced with UnsizedArrayOf, yes.  But it will remain in implementation of UnsizedArrayOf, ArrayOf, etc.  There's very few instances.

ebraminio: Or using UnsizedArrayOf inside ArrayOf and others and having only one of them, you said that once before IIRC

PR #1766, opened 2019-06-10
Add HB_NO_SETLOCALE

ebraminio:
#1764 related

### DIFF
--- a/src/hb-config.hh
+++ b/src/hb-config.hh
@@ -61,6 +61,7 @@
 #define HB_NO_LAYOUT_UNUSED
 #define HB_NO_MATH
 #define HB_NO_NAME
+#define HB_NO_SETLOCALE
 #define HB_NO_SUBSET_LAYOUT
 #endif
 
--- a/src/hb.hh
+++ b/src/hb.hh
@@ -359,7 +359,7 @@ extern "C" int hb_memalign_impl(void **memptr, size_t alignment, size_t size);
 #      define HB_NO_GETENV
 #    endif
 #    if _WIN32_WCE < 0x800
-#      define setlocale(Category, Locale) "C"
+#      define HB_NO_SETLOCALE
 static int errno = 0; /* Use something better? */
 #    endif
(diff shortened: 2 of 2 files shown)


behdad:
I like this.

Wider question: should we switch to using HAVE_SETLOCALE, etc, for these?  That would change the default for people who build custom.

Alternatively, maybe have these HB_NO_SETLOCALE etc as is, and have a HB_NO_LIBC that disables all libc functions?

ebraminio:
> Alternatively, maybe have these HB_NO_SETLOCALE etc as is, and have a HB_NO_LIBC that disables all libc functions?

An HB_LEAN_LIBC that defines HB_NO_SETLOCALE and HB_USE_INTERNAL_QSORT and defining some round fallback, and maybe enabling our baked mem* str* of libc, sad that malloc and friends involve more magic, I wish we had a zero allocation mode like https://www.sqlite.org/malloc.html#memsys5

behdad:
> An HB_LEAN_LIBC that defines HB_NO_SETLOCALE and HB_USE_INTERNAL_QSORT and defining some round fallback, and maybe enabling our baked mem* str* of libc, sad that malloc and friends involve more magic, I wish we had a zero allocation mode like https://www.sqlite.org/malloc.html#memsys5

`HB_NO_QSORT` please. :)

Also, no internal memory allocator.  Imagine if every library did that...

ebraminio:
> HB_NO_QSORT please. :)

Already added in https://github.com/harfbuzz/harfbuzz/commit/9407ef8d4bb96346b1f8b07757d79d3f8cc61cf7 :/ This `HB_NO_QSORT` feel to be not that nice naming but I am OK with rename

behdad:
> Already added in [9407ef8](https://github.com/harfbuzz/harfbuzz/commit/9407ef8d4bb96346b1f8b07757d79d3f8cc61cf7) :/ This `HB_NO_QSORT` feel to be not that nice naming but I am OK with rename

Fine, let's keep your naming.  I just don't like adding all kinds of arbitrary macros.

PR #1767, opened 2019-06-10
Add HB_NO_OEPN

### LINKED ISSUE #1764: Suggestions for HB_TINY additions

ebraminio:
But don't see the need to add it to lean or tiny personally, what motivated me for PR is to ask about the naming

(fixes #1764)

### DIFF
--- a/src/hb-config.hh
+++ b/src/hb-config.hh
@@ -60,7 +60,9 @@
 #define HB_NO_GETENV
 #define HB_NO_LAYOUT_UNUSED
 #define HB_NO_MATH
+#define HB_NO_MMAP
 #define HB_NO_NAME
+#define HB_NO_OEPN
 #define HB_NO_SETLOCALE
 #define HB_NO_SUBSET_LAYOUT
 #endif
--- a/src/hb-blob.cc
+++ b/src/hb-blob.cc
@@ -543,6 +543,9 @@ _hb_mapped_file_destroy (void *file_)
 hb_blob_t *
 hb_blob_create_from_file (const char *file_name)
 {
+#ifdef HB_NO_OEPN
+  return hb_blob_get_empty ();
+#endif
   /* Adopted from glib's gmappedfile.c with Matthias Clasen and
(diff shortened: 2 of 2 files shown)


behdad:
Rename to HB_NO_OPEN?  We also have HB_NO_MMAP.   Let's add both to HB_TINY.

Also, I had a change of mind: I think in HB_TINY, or in general, API that gets disabled should be removed instead of returning empty/error/false.  That way we know if a client links it will work.  And if it doesn't link it will point to which feature needs to be re-enabled.  I'll go ahead and do that.

ebraminio:
> Also, I had a change of mind: I think in HB_TINY, or in general, API that gets disabled should be removed instead of returning empty/error/false. That way we know if a client links it will work. And if it doesn't link it will point to which feature needs to be re-enabled. I'll go ahead and do that.

Yep, makes lots of sense, I just fear from compile warnings :/

behdad:
> Yep, makes lots of sense, I just fear from compile warnings :/

Right :(.  Let me see what I can come up with.

src/hb-config.hh
```diff
@@ -60,6 +61,7 @@
 #define HB_NO_GETENV
 #define HB_NO_LAYOUT_UNUSED
 #define HB_NO_MATH
+#define HB_NO_MMAP
```
ebraminio: Or having both in lean? My reasoning was it can work without mmap so it is lean but disabling the whole API in tiny

behdad: Right. TINY itself doesn't do these; LEAN does.  IMplement it by disabling the feature now.  Then we switch all to remove the API.  I don't think we need both.

PR #1770, opened 2019-06-11
Alternative approach to unify qsort implementation

ebraminio:
So no longer local modification of the original qsort_r implementation
and having of some random definitions around.

Just occurred to me this can be possible but with very little cost and wondering why it didn't happened to me initially. And let's remove HAVE_GNU_QSORT_R also, I liked to define that in our build after the migration to meson but now that we have one canonical implementation without templates (if accepted) I don't see much point to it considering how it can be optimized by compiler.


behdad:
I don't like this.

ebraminio:
Sure, understandable.

Issue #1719, opened 2019-05-18
[subset] CFF integration tests (test/subset) not being run.


garretrieger:
cff-full-font.tests and cff-japanese.tests are missing from ./test/subset/data/Makefile.sources and as a result are not currently being run. I tried enabling them, but they currently fail. Did not get a chance to look into the specific cause of the failures.

Issue #1167, opened 2018-09-21
[subset] OS/2 Unicode range bits shouldn't be set if unset in original


blueshade7:
**Problem:**
hb-subset overwrites the Unicode range bits in the OS/2 table with the range bits of Unicode characters selected for the subset font, mismatching the fonttools’ behavior where range bits are set if they are also set in the original font.

**Actual** (hb-subset result):
```
% hb-subset --output-file=hb-os2.otf --unicodes=41,100,1FC,FEFF SourceSansPro-Regular.otf
% spot -tOS/2 hb-os2.otf
### [OS/2] (0001be90)
```
_SNIP_
```
unicodeRange1     =0000000d   (Basic Latin
                               Latin Extended-A
                               Latin Extended-B)
unicodeRange2     =00000000
unicodeRange3     =00000008   (Arabic Presentation Forms-B)
unicodeRange4     =00000000
```
_SNIP_

**Expected** (fonttools result):
```
% pyftsubset --output-file=ft-os2.otf --unicodes=41,100,1FC,FEFF SourceSansPro-Regular.otf
% spot -tOS/2 ft-os2.otf
### [OS/2] (000003bc)
```
_SNIP_
```
unicodeRange1     =00000005   (Basic Latin
                               Latin Extended-A)
unicodeRange2     =00000000
unicodeRange3     =00000000
unicodeRange4     =00000000
```
_SNIP_

**Original Font:**
```
spot -tOS/2 SourceSansPro-Regular.otf
### [OS/2] (00000150)
```
_SNIP_
```
unicodeRange1     =600002f7   (Basic Latin
                               Latin-1 Supplement
                               Latin Extended-A
                               IPA Extensions
                               Spacing Modifier Letters
                               Combining Diacritical Marks
                               Basic Greek
                               Cyrillic
                               Latin Extend Additional
                               Greek Extended)
unicodeRange2     =02000001   (Superscripts and Subscripts
                               Surrogates)
unicodeRange3     =00000000
unicodeRange4     =00000000
```
_SNIP_

behdad:
@garretrieger Can you look into this?

blueshade7:
I am going to fix this at the same time as #1719 since manually patching fontTools results is a pain every time I regenerate subset baselines (there's still cmap diffs though).

blueshade7:
fixed by https://github.com/harfbuzz/harfbuzz/commit/82d4bfb8f3ac30fecce41f8abe59f58ad64ba98a

PR #1771, opened 2019-06-11
Enabled CFF subset tests #1719

blueshade7:
Also fixes OS/2 subset issue #1167.
Additionally, modified cmap subsetter to include Unicode/Microsoft cmap sub-tables iff their corresponding sub-tables available in the source font.

These changes are to better align with fontTools subset results for comparison.
Regenerated expected subset test results using fontTools (previously, they were hand-modified to match our results).


src/hb-ot-cmap-table.hh
```diff
@@ -918,8 +957,14 @@ struct cmap
     // Write out format 12 sub table.
     if (cmap_subset_plan.format12_groups)
     {
-      EncodingRecord &format12_rec = table->encodingRecord[2];
+      if (unlikely (!cmap_subset_plan.has_unicode_ucs4 && !cmap_subset_plan.has_ms_ucs4)) return false;
```
garretrieger: Not having a cmap 12 shouldn't be an error as long as you have at least one cmap sub-table.

blueshade7: True, but note the conditional `if (cmap_subset_plan.format12_groups)` two lines above. If `format12_groups` exists but no ucs4 cmap tables exist, something must be wrong. I speculate it could still happen say if non-Unicode cmap has format 12 contents unexpectedly. Thus this `if (unlikely` conditional.

garretrieger: Ah ok, that makes sense.

PR #1772, opened 2019-06-11
fix oss-fuzz 15211

blueshade7:


src/hb-subset-cff-common.cc
```diff
@@ -226,7 +226,7 @@ hb_serialize_cff_fdselect (hb_serialize_context_t *c,
 						fdselect_ranges);
 
     default:
-      assert(false);
+      return_trace (false);
   }
 
   return_trace (true);
```
ebraminio: Hmm, this isn't possible

ebraminio: Oh, I am wrong but will do some minor format around

PR #1775, opened 2019-06-13
[util] Add --reverse and --reverse-clusters options

khaledhosny:
Calling the corresponding hb_buffer_* functions, can be handy for testing.

### DIFF
--- a/util/options.cc
+++ b/util/options.cc
@@ -427,6 +427,8 @@ shape_options_t::add_options (option_parser_t *parser)
     {"utf8-clusters",	0, 0, G_OPTION_ARG_NONE,	&this->utf8_clusters,		"Use UTF8 byte indices, not char indices",	nullptr},
     {"cluster-level",	0, 0, G_OPTION_ARG_INT,		&this->cluster_level,		"Cluster merging level (default: 0)",	"0/1/2"},
     {"normalize-glyphs",0, 0, G_OPTION_ARG_NONE,	&this->normalize_glyphs,	"Rearrange glyph clusters in nominal order",	nullptr},
+    {"reverse",		0, 0, G_OPTION_ARG_NONE,	&this->reverse,			"Reverse output glyphs",	nullptr},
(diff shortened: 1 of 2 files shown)


khaledhosny:
`hb-shape` is fine, `hb-view` asserts with `--cluster-level=1` and `--reverse-clusters`:
```
hb-view NotoSerifHebrew-Regular.ttf --unicodes="U+05D1,U+05BC,U+05B0" --cluster-level=1 --reverse-clusters
**
ERROR:../../util/helper-cairo.cc:532:void helper_cairo_line_from_buffer(helper_cairo_line_t*, hb_buffer_t*, const char*, unsigned int, int, hb_bool_t):
assertion failed: (hb_glyph[i].cluster > hb_glyph[i+1].cluster)
Aborted (core dumped)
```

util/options.cc
```diff
@@ -419,6 +419,8 @@ shape_options_t::add_options (option_parser_t *parser)
     {"utf8-clusters",	0, 0, G_OPTION_ARG_NONE,	&this->utf8_clusters,		"Use UTF8 byte indices, not char indices",	nullptr},
     {"cluster-level",	0, 0, G_OPTION_ARG_INT,		&this->cluster_level,		"Cluster merging level (default: 0)",	"0/1/2"},
     {"normalize-glyphs",0, 0, G_OPTION_ARG_NONE,	&this->normalize_glyphs,	"Rearrange glyph clusters in nominal order",	nullptr},
+    {"reverse",		0, 0, G_OPTION_ARG_NONE,	&this->reverse,			"Rearrange glyph clusters in nominal order",	nullptr},
+    {"reverse-clusters",0, 0, G_OPTION_ARG_NONE,	&this->reverse_clusters,	"Rearrange glyph clusters in nominal order",	nullptr},
```
dscorbett: The new options’ documentation strings are the same as `normalize-glyphs`’s.

behdad:
lgtm w David's comment.  Donno what to do about the hb-view.   These only make sense to hb-shape... Maybe we should separate such options?

PR #1777, opened 2019-06-16
Canonically reorder U+0C55 and U+0C56

dscorbett:
Setting ccc=0 for U+0C55 and U+0C56 prevents reordering, which means among a set of canonically equivalent strings like {<0C15, 0C4D, 0C56>, <0C15, 0C56, 0C4D>} some get dotted circles and some don’t. The original reason to override their classes was to make sure U+0C4D TELUGU SIGN VIRAMA doesn’t get reordered before them, to effect which it is sufficient to change their classes to anything less than 9.


behdad:
Thanks.  Would be good to add tests as well.

PR #1778, opened 2019-06-16
[myanmar] Allow U+1037 and U+103A in either order

dscorbett:
U+1037 MYANMAR SIGN DOT BELOW has ccc=7 and U+103A MYANMAR SIGN ASAT has ccc=9. `modified_combining_class` modifies U+1037 to have ccc=9, so if the input string contains <U+103A, U+1037>, it won’t be canonicalized to <U+1037, U+1039>. The Ragel machine, however, expects canonicalized strings, so it sometimes adds a dotted circle to only one of a pair of canonically equivalent strings. The solution is to allow U+1039 to precede U+1037 in various regexes.

There is one case that solution doesn’t handle: a broken cluster with a kinzi and a dot below: <U+101B, U+103A, U+1039, U+1037> is one cluster but its NFC form <U+101B, U+1037, U+103A, U+1039> is two. It’s broken either way so it doesn’t matter.

Another solution would be to remove the override from `modified_combining_class`. What were the test strings that 1c8654ead41ca746d577549c92d2a41c594ab639 fixed?


behdad:
I suggest you remove the hack I introduced, and see how to best solve the resulting situation.
You can probably find the sequences if you shape the entire Myanmar wikipedia corpus before/after: https://github.com/harfbuzz/harfbuzz-testing-wikipedia

That's what I was testing.

Thanks.

dscorbett:
I tested a few fonts on that corpus with and without that override. The only differences were in 8 tokens like ⟨သည့်⟩ containing the sequence ⟨base, asat, dot below⟩, where the mark glyphs are in a different order but the final rendering is the same. Although there could be a font that relies on the non-canonical order, none of my fonts seem to: each is either too simplistic to care about mark order or sophisticated enough to use mark classes. So the solution might just be to remove the override and nothing more.

behdad:
Sounds great! Thanks for testing.

dscorbett:
googlefonts/noto-fonts#1530 has an example of a string where the order does matter and the non-canonical order is preferred. On the other hand, no such string appears in that Wikipedia extract, so maybe it doesn’t matter enough for HarfBuzz to do anything about it. What do you think?

behdad:
I don't have a strong opinion.  @jfkthame @mhosken

mhosken:
Kinzi followed by a lower dot does not make any sense either orthographically or technically. The kinzi is a diacritic as is the 1037. The two would never follow each other like that since they both need a base between them. Yes, I suppose someone could enter that and it would look all wrong. But that is a case of GIGO.

More generally, I think it would be very remiss if normalizing text broke its rendering.

PR #1780, opened 2019-06-18
add hb_bimap_t

blueshade7:
Extracted `hb_bimap_t` changes from var-subset branch.
Rewrote its implementation to use a pair of `hb_map_t` replacing a pair of `hb_map_t` & `hb_vector_t`, leaving it simple & generic.
Moved `add()` and `sort()` (renamed from `reorder`) methods to its new subclass `hb_inc_bimap_t` for use by CFF1 and CFF2 subsetters replacing `CFF::remap_t`.
Miscellaneous renaming & typo fixes.

### DIFF
--- a/src/Makefile.sources
+++ b/src/Makefile.sources
@@ -44,6 +44,7 @@ HB_BASE_sources = \
 	hb-machinery.hh \
 	hb-map.cc \
 	hb-map.hh \
+	hb-bimap.hh \
 	hb-meta.hh \
 	hb-mutex.hh \
 	hb-null.hh \
--- a/src/hb-bimap.hh
+++ b/src/hb-bimap.hh
@@ -1,0 +1,129 @@
+/*
+ * Copyright © 2019 Adobe Inc.
+ *
+ *  This is part of HarfBuzz, a text shaping library.
+ *
+ * Permission is hereby granted, without written agreement and without
+ * license or royalty fees, to use, copy, modify, and distribute this
+ * software and its documentation for any purpose, provided that the
+ * above copyright notice and the following two paragraphs appear in
+ * all copies of this software.
+ *
+ * IN NO EVENT SHALL THE COPYRIGHT HOLDER BE LIABLE TO ANY PARTY FOR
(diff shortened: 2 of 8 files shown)


behdad:
Thanks.  I'm fine with this right now.  If we end up using it more often, we should make it more like map.  For example, the method `get` and `operator[]` should implement the forward mapping.

blueshade7:
I am happy to add [] op actually.

behdad:
Keep it similar to hb_map_t API please then.

blueshade7:
If we want `hb_bimap_t` to have the same operators & functions as `hb_map_t` and behave as the forward map, how about subclass it and add a backward map as its member? That way we could skip defining some of them.

blueshade7:
Also, shall we make it a template based on `hb_hashmap_t` instead of `hb_map_t`, in case someone wants to instantiate it differently?

behdad:
> Also, shall we make it a template based on `hb_hashmap_t` instead of `hb_map_t`, in case someone wants to instantiate it differently?

Not originally.  Can always change in the future, same way we added hashmap after map was added.

behdad:
> If we want `hb_bimap_t` to have the same operators & functions as `hb_map_t` and behave as the forward map, how about subclass it and add a backward map as its member? That way we could skip defining some of ops & funcs.

Not a good idea, as it will allow it to automatically cast to map_t, and using it that way is wrong because changes will not be reflected in backward map.

You can maybe subclass from it privately (or protected) and use `using` to pull in the couple of members that you want.

We can add a cast operator to `const hb_map_t&` to provide the forward map, but only as a const.

blueshade7:
Added []op, get() etc.
Didn't bother adding iterator interface though.
Some cleanup.

behdad:
Thanks.  One last issue: if key or value is INVALID, do we need to better handle that?  If value is INVALID that should act as a delete.  If key is INVALID, we should just ignore request.  Would be good to add tests at least.  Can you add test-bimap.cc?

behdad:
The test file would go straight in src/.

behdad:
I'm merging this.  Fixes can go in after.

PR #1781, opened 2019-06-18
Depend setlocale use to build system provided check

ebraminio:


src/hb-common.cc
```diff
@@ -35,10 +35,6 @@
 #include <xlocale.h>
 #endif
 
-#ifdef HB_NO_SETLOCALE
-#define setlocale(Category, Locale) "C"
-#endif
```
ebraminio: Apparently redefining this way is not enough, that motivated to look for another solution and thought lets turn it into a check anyway.

behdad:
I thought a lot about what we should check in configure (ie. default disabled) vs define `HB_NO_*` for (ie. default enable).  I prefer that things widely available as part of C standard library to be default-enabled, such that they are picked up automatically by alternative builders.  So, why not keep it that way just change the implementation?

behdad:
Ie. anything that is C89 should be default enabled :).  Maybe more than just that.

behdad:
Probably add `HB_NO_LIBC` or something as well.

ebraminio:
I had to create mock headers also anyway and am no longer seeing this, so :)

PR #1782, opened 2019-06-18
[subset] VORG table to use _subset2 method and new iterator frameworks

qxliu76:


src/hb-ot-vorg-table.hh
```diff
@@ -69,102 +69,52 @@ struct VORG
...
-                                        free);
-    bool success = plan->add_table (HB_OT_TAG_VORG, result);
-    hb_blob_destroy (result);
-    subset_metrics.fini ();
-    hb_blob_destroy (vorg_blob);
-    return success;
+    vorg_prime->serialize (c->serializer, it, defaultVertOriginY);
+
+    if (unlikely (c->serializer->ran_out_of_room || c->serializer->in_error ())) return_trace (false);
+    return_trace (true);
```
behdad: These two lines can be joined.

However, it's not even correct.  Subset should return false if table is not needed in the subset font.  Ie. should return false if no glyphs survived.  In the case of a table like VORG that covers all glyphs in a font, just return true always.  The error is taken care of by the serializer.

PR #1783, opened 2019-06-18
hb-coretext: remove unnecessary semicolons

mistydemeo:
Recent versions of clang feature a new warning, `-Wextra-semi-stmt`, which identifies empty statements. The CoreText code contains several macros which themselves end in semicolons, and the places they're called also terminated them with semicolons. The second semicolon triggered the new clang warning. Simply removing the duplicate semicolons silences it.

Sample error:
```
hb-coretext.cc:682:36: error: empty expression statement has no effect; remove
      unnecessary ';' to silence this warning [-Werror,-Wextra-semi-stmt]
      FAIL ("Buffer resize failed");
```

### DIFF
--- a/src/hb-coretext.cc
+++ b/src/hb-coretext.cc
@@ -679,7 +679,7 @@ resize_and_retry:
     old_scratch_used = scratch - old_scratch;
 
     if (unlikely (!buffer->ensure (buffer->allocated * 2)))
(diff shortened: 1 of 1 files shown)


behdad:
Huh.  Weird that we didn't catch this before.  At any rate, the fix is to remove the semicolon in the macro.  I'll fix locally.

PR #1784, opened 2019-06-18
coretext: remove trailing macro from SCRATCH_RESTORE

mistydemeo:
Follows up on #1783. While 10bac21bb5b25cf20c2198934e99e444625dfd97 fixed this for `FAIL`, it didn't touch `SCRATCH_RESTORE`, a coretext.cc-only macro which was also affected by the same issue.


behdad:
Ah thanks.  Sorry about that!

PR #1788, opened 2019-06-23
Add missing cairo-svg dependency to test-ot-color

jameshilliard:
Should fix:
```
test-ot-color.cc:40:10: fatal error: cairo-svg.h: No such file or directory
   40 | #include <cairo-svg.h>
      |          ^~~~~~~~~~~~~
```

### DIFF
--- a/src/test-ot-color.cc
+++ b/src/test-ot-color.cc
@@ -25,7 +25,9 @@
 
 #include "hb.hh"
 
-#ifndef HB_NO_COLOR
+#include <cairo.h>
+
+#if !defined(HB_NO_COLOR) && defined(CAIRO_HAS_SVG_SURFACE)
 
 #include "hb-ot.h"
 
@@ -35,7 +37,6 @@
 #include FT_FREETYPE_H
 #include FT_GLYPH_H
 
-#include <cairo.h>
 #include <cairo-ft.h>
 #include <cairo-svg.h>
 


behdad:
I don't like configure check for this.  @ebraminio can you change to just use cairo PNG backend instead? Or use #ifdefs.  That's what hb-view does.

jameshilliard:
cairo PNG would still require a dependency check for this right?

behdad:
The cairo PNG backend is always buillt in, so we don't check extra for hb-view and never had any complaints.

These are macros to check for all:
```
helper-cairo.cc:  #ifdef CAIRO_HAS_PS_SURFACE
helper-cairo.cc:#ifdef CAIRO_HAS_PNG_FUNCTIONS
helper-cairo.cc:  #ifdef CAIRO_HAS_PNG_FUNCTIONS
helper-cairo.cc:  #ifdef CAIRO_HAS_SVG_SURFACE
helper-cairo.cc:  #ifdef CAIRO_HAS_PDF_SURFACE
helper-cairo.cc:  #ifdef CAIRO_HAS_PS_SURFACE
```

jameshilliard:
> The cairo PNG backend is always buillt in

Are you sure? What about when cairo is built with `--disable-png`?

behdad:
You're right.  The image backend is always built in, but the PNG functions can be disabled.  At any rate, I don't like adding extra checks just for `test-ot-colors`.  The test file can just not save to anything if the necessary cairo features are not available.

jameshilliard:
Updated to use the `CAIRO_HAS_SVG_SURFACE` macro instead.

ebraminio:
Hey there, shouldn't cairo header itself should be included before checking its definitions?

jameshilliard:
updated, hopefully that works

ebraminio:
Merged, thanks for your contribution :)

PR #1790, opened 2019-06-24
Cast long->size_t to ensure comparison of similar types

GaryQian:
When compiling this for Flutter, we are seeing a failure with comparing `unsigned long` with `long`.

### DIFF
--- a/src/hb-common.cc
+++ b/src/hb-common.cc
@@ -70,7 +70,7 @@ _hb_options_init ()
         p = c + strlen (c);
 
 #define OPTION(name, symbol) \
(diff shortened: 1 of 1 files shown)


ebraminio:
We should fix this more systematically, what is the compiler warning you are seeing?

behdad:
I agree I'm surprised none of our bots hit this, since we clearly turn this on as an error...

I remember believing that clang and gcc both have some smartness when it comes to warning about sign-compare if they had to up-cast any of the operands.  So yes, I'm surprised this was an issue for flutter.  But maybe with an older / unusual compiler?

PR #1791, opened 2019-06-24
Replace create_from_file with empty blob if HB_NO_OPEN is defined

ebraminio:
In order to compile the library however HB_NO_OPEN is also needed in CFLAGS... let me revert the change so we decide which to pick

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -38,6 +38,10 @@
 
 using namespace OT;
 
+#ifdef HB_NO_OPEN
+#define hb_blob_create_from_file(x)  hb_blob_get_empty ()
+#endif
+
 int
 main (int argc, char **argv)
 {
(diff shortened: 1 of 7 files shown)


behdad:
Not in util/ please.  I don't want hb-shape / hb-view that compile but not work.  Just want that for test programs.  Thanks!

ebraminio:
> Not in util/ please. I don't want hb-shape / hb-view that compile but not work. Just want that for test programs. Thanks!

I would say why caring about the buildability with HB_TINY when all we need is `gcc -DHB_TINY src/hb-*.cc` now :) Done but one should only build `src/` from now on

behdad:
> I would say why caring about the buildability with HB_TINY when all we need is `gcc -DHB_TINY src/hb-*.cc` now :)

That takes 30s here!

PR #1792, opened 2019-06-24
Turn create_from_file into an empty call if HB_NO_OPEN is defined

ebraminio:
Let's decide which to pick, this or #1791


behdad:
Def not this one.  The new (current) mode of operation is, we remove API, not make them disfunctional.  Such that if app links, it works.

PR #1796, opened 2019-06-26
[ucd] Include scripts added in Unicode 10 or later

dscorbett:


behdad:
Ouch!  Thanks for this....  There is a TODO item in the generator code:
```
# TODO Currently if gc_order or sc_order do not capture all values, we get in 
# trouble because they silently add new values.  We should be able to "freeze" 
# them, or just do the mapping ourselves. 
```

I will go ahead and fix that as well, so this wouldn't happen again.

Test suite probably should have a test as well.

behdad:
In test-unicode.c we "soft-fail" script tests for Unicode > 5.2.  While times have changed, I don't think we can switch to testing for Unicode 10 scripts there...

Still, would be nice to add.  Maybe should NOT soft-fail for internal UCD implementation, but possibly do for glib / ICU.

arzanilorestan:
آدرس،خرم آباد،اسدآبادی(کوروش)خیابان۱۴متری،پایینترازچهارراه۱۴متری نبش کوچه نیایش۳

behdad:
> In test-unicode.c we "soft-fail" script tests for Unicode > 5.2. While times have changed, I don't think we can switch to testing for Unicode 10 scripts there...

Let's add at least one character per newer Unicode version there.  @dscorbett can you do that please?

PR #1797, opened 2019-06-26
[CFF] Fix dict values so can be addressed from a null pool

ebraminio:
fixes https://crbug.com/971933

### DIFF
--- a/src/hb-ot-cff1-table.hh
+++ b/src/hb-ot-cff1-table.hh
@@ -55,7 +55,8 @@ typedef Subrs<HBUINT16>    CFF1Subrs;
 struct CFF1FDSelect : FDSelect {};
 
 /* Encoding */
-struct Encoding0 {
+struct Encoding0
+{
   bool sanitize (hb_sanitize_context_t *c) const
   {
     TRACE_SANITIZE (this);
@@ -67,9 +68,7 @@ struct Encoding0 {
     assert (glyph > 0);
     glyph--;
     if (glyph < nCodes ())
-    {
-      return (hb_codepoint_t)codes[glyph];
-    }
+      return (hb_codepoint_t) codes[glyph];
     else
       return CFF_UNDEF_CODE;
   }
@@ -77,7 +76,7 @@ struct Encoding0 {
   HBUINT8 &nCodes () { return codes.len; }
   HBUINT8 nCodes () const { return codes.len; }
 
-  ArrayOf<HBUINT8, HBUINT8> codes;
+  PString codes;
 
   DEFINE_SIZE_ARRAY_SIZED (1, codes);
 };
@@ -126,7 +125,8 @@ struct Encoding1 {
   DEFINE_SIZE_ARRAY_SIZED (1, ranges);
 };
 
-struct SuppEncoding {
+struct SuppEncoding
+{
   bool sanitize (hb_sanitize_context_t *c) const
   {
     TRACE_SANITIZE (this);
@@ -880,7 +880,7 @@ struct cff1_private_dict_values_base_t : dict_values_t<VAL>
   {
     dict_values_t<VAL>::init ();
     subrsOffset = 0;
-    localSubrs = &Null(CFF1Subrs);
+    localSubrs = nullptr;
   }
   void fini () { dict_values_t<VAL>::fini (); }
 
@@ -895,8 +895,9 @@ struct cff1_private_dict_values_base_t : dict_values_t<VAL>
     return size;
   }
 
-  unsigned int      subrsOffset;
-  const CFF1Subrs    *localSubrs;
+  unsigned int	subrsOffset;
+  hb_nonnull_ptr_t<const CFF1Subrs>
+		localSubrs;
 };
 
 typedef cff1_private_dict_values_base_t<op_str_t> cff1_private_dict_values_subset_t;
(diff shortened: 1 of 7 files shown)


src/hb-ot-cff1-table.hh
```diff
@@ -895,7 +895,14 @@ struct cff1_private_dict_values_base_t : dict_values_t<VAL>
     return size;
   }
 
+  const CFF1Subrs* get_subrs () const
+  { return localSubrs ? localSubrs : &Null (CFF1Subrs); }
+
+  void set_subrs (const CFF1Subrs* _localSubrs)
+  { localSubrs = _localSubrs; }
+
   unsigned int      subrsOffset;
```
ebraminio: This also should become private but wanted to do that in separate change

behdad:
Use `hb_nonnull_ptr_t`?  I don't know the CFF code well, so cannot really review.

ebraminio:
> Use hb_nonnull_ptr_t? I don't know the CFF code well, so cannot really review.

Done

src/hb-subset-cff-common.hh
```diff
@@ -658,9 +658,8 @@ struct subr_subsetter_t
     parsed_global_subrs.init (acc.globalSubrs->count);
     parsed_local_subrs.resize (acc.fdCount);
     for (unsigned int i = 0; i < acc.fdCount; i++)
-    {
-      parsed_local_subrs[i].init (acc.privateDicts[i].localSubrs->count);
-    }
+      parsed_local_subrs[i].init (acc.privateDicts[i].localSubrs.get ()->count);
```
behdad: Don't actually need `.get()` before dereferencing...

src/hb-ot-cff2-table.hh
```diff
@@ -491,8 +492,8 @@ struct cff2
 	if (unlikely (!priv_interp.interpret (privateDicts[i]))) { fini (); return; }
 
 	privateDicts[i].localSubrs = &StructAtOffsetOrNull<CFF2Subrs> (&privDictStr[0], privateDicts[i].subrsOffset);
-	if (privateDicts[i].localSubrs != &Null(CFF2Subrs) &&
-	  unlikely (!privateDicts[i].localSubrs->sanitize (&sc)))
+	if (privateDicts[i].localSubrs.get () != &Null (CFF2Subrs) &&
```
behdad: Maybe add is_null() to hb_nonnull_ptr_t?

The way these things were intended to be used, one should never compare to null object...

blueshade7:
The suggested changes look ok, but they don’t seem to make the original problem go away. I still get the error when I run shape fuzzer with the test case.

Complaint appears to come from the fact that a null Subrs pointer is derefed (*) then passed as a ref param (&) to `cs_interp_env_t::init` then to `biased_subrs_t::init`, even though they are actually used as pointers after refed (&). My bad for this inconsistency. So my suggestion is to change these init functions to take Subrs params as pointers instead of references.

Shall I push this change to cffnull branch?

src/hb-ot-cff2-table.hh
ebraminio: Yep, I don't want put comparison to null pool there so we should compare to nullptr itself and that means removing the above line, which, has unintended effect so lets skip this for now.

ebraminio:
> The suggested changes look ok, but they don’t seem to make the original problem go away. I still get the error when I run shape fuzzer with the test case.

Are you sure testing the branch? I don't see the issue both locally and in the bots and they were complaining when I first uploaded the testcase without the fix.

> Complaint appears to come from the fact that a null Subrs pointer is derefed (*) then passed as a ref param (&) to cs_interp_env_t::init then to biased_subrs_t::init, even though they are actually used as pointers after refed (&). My bad for this inconsistency. So my suggestion is to change these init functions to take Subrs params as pointers instead of references.

I believe complaint is coming from `*acc.privateDicts[fd].localSubrs` that fd is over the `privateDicts` `hb_vector_t` size which means the returned values from indexing will be from null pool and this fixes it by making it not need to explicitly `init ()`ing needed so the returned value from null pool can work also.

> Shall I push this change to cffnull branch?

Feel free to merge this first, and/or, create a separate PR based on these changes or not, I am pretty sure this is the fix the testcase, I should've explained why.

blueshade7:
> Are you sure testing the branch?

Ah I failed to grab branch from your depot. As the result I accidentally tested on master and came up with another fix as outlined above. Anyways I created a branch for it as `harfbuzz/cffnull_fix2`

ebraminio:
Great, please upload also as a PR and put the testcase on the patch. Thanks

src/hb-ot-cff1-table.hh
```diff
@@ -1138,8 +1137,7 @@ struct cff1
 	if (unlikely (!priv_interp.interpret (*priv))) { fini (); return; }
 
 	priv->localSubrs = &StructAtOffsetOrNull<CFF1Subrs> (&privDictStr, priv->subrsOffset);
-	if (priv->localSubrs != &Null(CFF1Subrs) &&
-	    unlikely (!priv->localSubrs->sanitize (&sc)))
+	if (priv->subrsOffset && !priv->localSubrs->sanitize (&sc))
```
ebraminio: Apparently CFF1Subrs::sanitize (and CFF2Subrs) isn't returning true on NullPool objects which isn't good.

ebraminio: `Number of object data. Note there are (count+1) offsets` hmm, so we can't do anything about

ebraminio:
Lets deal with this later

behdad:
Just commit whitespace changes separately and push to master... or not make whitespace changes.

PR #1800, opened 2019-06-27
[CFF] fix for Subrs null ptr access

blueshade7:
Alternate fix for PR #1797
https://crbug.com/971933


ebraminio:
Ok, this makes sense to me, specially why it was happening in some compiler not in others, lets have this in and go for other improvements later. Thanks

behdad:
I don't quite understand how this fixes anything.

ebraminio:
This passes nullptr used in 'localSubrs.init (localSubrs_);' without derefing it, straightly to the function which is apparently able to deal with nullptr, I did like my solution more still but that also not enough, we need to dedicate some real time with no hurry to make the code similar to other parts of the base, not passing nullptrs around and not comparing with nullpool addr and that's why I accepted this for now

behdad:
Got it.  Thanks.

PR #1804, opened 2019-06-28
[CFF] Move Null comparison to sanitizers

ebraminio:
This, hmm, actually this change doesn't look good but comparing to &Null after each StructAtOffsetOrNull doesn't look either


behdad:
This is not how it's all supposed to work yeah...

ebraminio:
Yep, this isn't any better.

PR #1806, opened 2019-07-01
Make harfbuzz working on all existing versions of Windows

carlo-bramini:
I discovered that harfbuzz is blocking the execution of KICAD EDA (http://kicad-pcb.org/) on Windows XP.
This happens because the library is linked for loading DWRITE.DLL implicitely, which works fine on Windows 7 but this blocks the execution on older versions, since this library does not exists.
I would like to suggest a minimal change for checking at runtime the presence of DirectWrite support on the running platform. Together with this fix, I also included few little improvements (check return value of the creator and fix case of included PSDK file names), which are descripted in the commits.
I hope that you will find this useful.

Sincerely.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -376,7 +376,7 @@ if test "x$with_directwrite" = "xyes" -a "x$have_directwrite" != "xtrue"; then
 fi
 if $have_directwrite; then
 	DIRECTWRITE_CXXFLAGS=
-	DIRECTWRITE_LIBS="-ldwrite"
+	DIRECTWRITE_LIBS=
 	AC_SUBST(DIRECTWRITE_CXXFLAGS)
 	AC_SUBST(DIRECTWRITE_LIBS)
 	AC_DEFINE(HAVE_DIRECTWRITE, 1, [Have DirectWrite library])
--- a/src/hb-directwrite.cc
+++ b/src/hb-directwrite.cc
@@ -167,8 +167,8 @@ _hb_directwrite_shaper_face_data_create (hb_face_t *face)
     return nullptr; \
   } HB_STMT_END
 
-  data->dwrite_dll = LoadLibrary(TEXT("DWRITE"));
-  if (data->dwrite_dll == NULL)
+  data->dwrite_dll = LoadLibrary (TEXT ("DWRITE"));
+  if (unlikely (!data->dwrite_dll))
     FAIL ("Cannot find DWrite.DLL");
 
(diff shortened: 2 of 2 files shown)


ebraminio:
Hey there, how harfbuzz is used there? Couldn't you or distributor disable hb-directwrite altogether? You won't need the whole thing I'm pretty sure.

I'm fine with the fix but apparently it breaks 64bit builds on msys2 https://ci.appveyor.com/project/harfbuzz/harfbuzz/builds/25660591/job/wyxdtb3vkrbi2oec#L424

ebraminio:
Have developed the thing, the whole things is for internal testing for now so feel free to disable it altogether wherever it is built to save some [k]bytes.

carlo-bramini:
Thank you very much for the quick reply.

I implemented the same fix you did inside `src/hb-uniscribe.cc`, by temporally ignoring `-Wcast-function-type`. In this way, the error rised by GCC8 should be fixed. I tested it with GCC 9.1.0 and now it works fine. Previously I compiled 64bit version with GCC 7.2.0 so I did not noticed it, sorry.

Unfortunately, I think that disabling the feature won't be accepted by developers, because I already faced this kind of problem in the past:
https://bugs.launchpad.net/kicad/+bug/1828527
in few words: they want to use the binary packages as they are made available by MSYS2 team and so, the only solution was to improve the code of original library.
Since MSYS2 team provides a build script with `--with-directwrite=yes` activated:
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-harfbuzz/PKGBUILD
the solution here is to check if support for DirectWrite is available.
Actually, this type of solution is already used for supporting Uniscribe into `src/hb-uniscribe.cc`, so in my opinion it won't be a bad idea to do the same here.

ebraminio:
Applied some minor style normalization, lets see if bots are happy and merge if you don't see any issue, thanks :)

ebraminio:
Also would you like to give us a hand on some other Windows XP issue? :) We like to have our tools to work in XP and this is the artifact of one of our bots https://98027-5050525-gh.circle-artifacts.com/0/harfbuzz-win32.zip is being built by this line in our CI https://github.com/harfbuzz/harfbuzz/blob/master/.circleci/config.yml#L209 but it fails to work even the fact we tried to fix it by preferring pthread mutex over ours https://github.com/harfbuzz/harfbuzz/commit/210f93c14

PR #1815, opened 2019-07-04
hb_set_previous_range docs: fix presumed copy/paste error

SimonSapin:


behdad:
Thanks.

Curious: what are you onto? :)

SimonSapin:
I’m reading through the manual, occasionally jumping to the API reference and back, to understand what Harfbuzz does or doesn’t do and how to use it. When I had last looked at the manual several years ago I was disappointed to find multiple chapters completely empty except for their title. It’s really good that this is no longer the case!

Next, I will look into [Skribo](https://github.com/linebender/skribo) (which uses Harfbuzz and adds font selection and fallback) for eventually using it in Servo.

behdad:
> When I had last looked at the manual several years ago I was disappointed to find multiple chapters completely empty except for their title. It’s really good that this is no longer the case!

Indeed!  I'm glad we finally have something readable... Thanks to @n8willis and Mozilla sponsorship.

I'd be happy to discuss anything, as you know.  Cheers

PR #1826, opened 2019-07-09
Fix hb-subset unintialized memory read of cmap table

### LINKED ISSUE #1823: msan failure in roboto-black.ttf
One linear repro:
`git clone https://github.com/harfbuzz/harfbuzzjs && cd harfbuzzjs && git clone https://github.com/harfbuzz/harfbuzz --depth=1 && cd subset && clang test.cc -fsanitize=memory -o test -I ../harfbuzz/src/ -fno-rtti -fno-exceptions -lm -fno-omit-frame-pointer && ./test`

More verbose one:
1. cd src
2. wget https://github.com/harfbuzz/harfbuzzjs/raw/master/subset/roboto-black.ttf
3. Save this as a.cc
```C
#include <stdio.h>
#include <stdlib.h>

#include "harfbuzz.cc"
#include "hb-subset.cc"
#include "hb-subset-cff1.cc"
#include "hb-subset-cff2.cc"
#include "hb-subset-cff-common.cc"
#include "hb-subset-input.cc"
#include "hb-subset-plan.cc"

int main() {
    hb_blob_t *blob = hb_blob_create_from_file ("roboto-black.ttf");
    hb_face_t *face = hb_face_create (blob, 0/*this is ttcIndex*/);
    hb_blob_destroy (blob); /* face keeps a reference of to it so you can destroy it here */

    /* Add your glyph indices here and subset */
    hb_set_t *glyphs = hb_set_create ();
    hb_set_add (glyphs, 'a');
    hb_set_add (glyphs, 'b');
    hb_set_add (glyphs, 'c');

    hb_subset_input_t *input = hb_subset_input_create_or_fail ();
    hb_set_t *input_glyphs = hb_subset_input_unicode_set (input);
    hb_set_union (input_glyphs, glyphs);
    hb_set_destroy (glyphs);
    //hb_subset_input_set_drop_hints (input, true);
    hb_face_t *subset = hb_subset (face, input);

    /* Clean up */
    hb_subset_input_destroy (input);

    /* Get result blob */
    hb_blob_t *result = hb_face_reference_blob (subset);
    unsigned int length = hb_blob_get_length (result);
    const char *data = hb_blob_get_data (result, 0);

    /* Write. If you like! */
    FILE *f = fopen ("hgfhgf.ttf", "wb");
    fwrite (data, 1, length, f);
    fclose (f);

    /* Clean up */
    hb_blob_destroy (result);
    hb_face_destroy (subset);
    hb_face_destroy (face);

    return 0;
}
```
4. clang a.cc -fsanitize=memory -I ../harfbuzz/src/ -fno-rtti -fno-exceptions -lm -fno-omit-frame-pointer && ./a.out

What I see
```
Uninitialized bytes in __interceptor_fwrite at offset 48 inside [0x71e000000000, 2716)
==12878==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x5604f6e2ee19 in main  subset/harfbuzz/src/a.out+0x169e19)
    #1 0x7f8ebc26cee2 in __libc_start_main (/usr/lib/libc.so.6+0x26ee2)
    #2 0x5604f6ce511d in _start  subset/harfbuzz/src/a.out+0x2011d)

SUMMARY: MemorySanitizer: use-of-uninitialized-value (harfbuzzjs/subset/harfbuzz/src/a.out+0x169e19) in main
```

Not sure what is going on, can you give me a hand @garretrieger @rsheeter maybe

ebraminio:
Related to #1823 and https://github.com/harfbuzz/harfbuzzjs/issues/9#issuecomment-509143536
Fixes what https://circleci.com/gh/harfbuzz/harfbuzz/99080 now sees

### DIFF
--- a/src/hb-ot-cmap-table.hh
+++ b/src/hb-ot-cmap-table.hh
@@ -878,7 +878,7 @@ struct cmap
     cmap_plan->has_ms_bmp = find_subtable (3, 1);
     cmap_plan->has_ms_ucs4 = find_subtable (3, 10);
     cmap_plan->num_enc_records = cmap_plan->has_unicode_bmp + cmap_plan->has_unicode_ucs4 + cmap_plan->has_ms_bmp + cmap_plan->has_ms_ucs4;
-  
+
     if (unlikely (!CmapSubtableFormat4::create_sub_table_plan (plan, &cmap_plan->format4_segments)))
       return false;
 
@@ -978,6 +978,14 @@ struct cmap
       CmapSubtableFormat12 &format12 = subtable.u.format12;
       if (unlikely (!format12.serialize (&c, cmap_subset_plan.format12_groups)))
 	return false;
+    }
+    else
+    {
+      // FIXME: Merge this with above or, remove and tweak #final_size
+      // and rebase all the tests expectations
+      HBUINT32 empty;
+      empty = 0;
+      for (unsigned int i = 0; i < 4; ++i) c.copy (empty);
     }
 
     c.end_serialize ();


behdad:
How does this fix anything? :)

behdad:
I also don't like leaving around a temp file...

Maybe you can just checksum it or something.  Why is the face builder not checksuming the tables?

behdad:
Thanks.  But even the second commit is just hiding the problem.  No?  Why is an unitialized byte getting in in the first place?  Suspect.

Didn't we port cmap to use subset2 facilities?

cc @garretrieger

ebraminio:
> How does this fix anything? :)

Hehe, be more patient, I was testing and logging the bots before the change, 
ubsan https://circleci.com/gh/harfbuzz/harfbuzz/98920 and valgrind https://circleci.com/gh/harfbuzz/harfbuzz/98936

> Maybe you can just checksum it or something. 

Didn't work, interestingly.

> Thanks. But even the second commit is just hiding the problem. No? Why is an unitialized byte getting in in the first place? Suspect.

I know, see https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ot-cmap-table.hh#L535 apparently it adds 16 bytes for an unknown reason, if I change that I guess I should to update many different tests, that's why I testing with this for now

behdad:
> > Thanks. But even the second commit is just hiding the problem. No? Why is an unitialized byte getting in in the first place? Suspect.
> 
> I know, see https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ot-cmap-table.hh#L535 apparently it adds 16 bytes for an unknown reason, if I change that I guess I should to update many different tests, that's why I testing with this for now

That looks like the subtable header size.  Anyway, suspect.  Needs more look.

ebraminio:
> That looks like the subtable header size. Anyway, suspect. Needs more look.

Yes as that won't executed when there is none https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ot-cmap-table.hh#L966-L981

behdad:
I let you and garret figure out.

behdad:
Okay thanks.  I'm fine you pushing this in now while we work on a real fix.  DON'T CLOSE ANY ISSUES PLEASE :D.

behdad:
cc @drott

PR #1827, opened 2019-07-10
Force blob generation and memory check in hb-subset-fuzzer

ebraminio:


ebraminio:
Yeah, rabbit hole of issues, ubsan bot complain is completely different from valgrind one. My suggestion is that I go for silencing and work around both issues with the following patch temporarily then you go for an actual and revert the work around fix whenever want to.

```diff
diff --git a/src/hb-open-file.hh b/src/hb-open-file.hh
index 5318c7fc..5ea7ffbe 100644
--- a/src/hb-open-file.hh
+++ b/src/hb-open-file.hh
@@ -146,9 +146,10 @@ typedef struct OffsetTable
 
       /* Allocate room for the table and copy it. */
       char *start = (char *) c->allocate_size<void> (rec.length);
-      if (unlikely (!start)) {return false;}
+      if (unlikely (!start)) return false;
+      if (unlikely (!blob->data)) continue;
 
-      memcpy (start, hb_blob_get_data (blob, nullptr), rec.length);
+      memcpy (start, blob->data, rec.length);
 
       /* 4-byte alignment. */
       c->align (4);
diff --git a/src/hb-ot-cmap-table.hh b/src/hb-ot-cmap-table.hh
index ac75bd96..c7e819ea 100644
--- a/src/hb-ot-cmap-table.hh
+++ b/src/hb-ot-cmap-table.hh
@@ -878,7 +878,7 @@ struct cmap
     cmap_plan->has_ms_bmp = find_subtable (3, 1);
     cmap_plan->has_ms_ucs4 = find_subtable (3, 10);
     cmap_plan->num_enc_records = cmap_plan->has_unicode_bmp + cmap_plan->has_unicode_ucs4 + cmap_plan->has_ms_bmp + cmap_plan->has_ms_ucs4;
-  
+
     if (unlikely (!CmapSubtableFormat4::create_sub_table_plan (plan, &cmap_plan->format4_segments)))
       return false;
 
@@ -997,7 +997,7 @@ struct cmap
 
     // We now know how big our blob needs to be
     size_t dest_sz = cmap_subset_plan.final_size ();
-    void *dest = malloc (dest_sz);
+    void *dest = calloc (1, dest_sz);
     if (unlikely (!dest)) {
       DEBUG_MSG(SUBSET, nullptr, "Unable to alloc %lu for cmap subset output", (unsigned long) dest_sz);
       return false;
```

ebraminio:
Or merge this bot failing patch just now

ebraminio:
Fixed ubsan issue, why it was happening? Subsetter will result an empty glyf when no codepoint is matched with the one exists in font so this is not hiding things AFAICT.

behdad:
Can you explain it to me like I'm a 4-year old? :)

behdad:
Push the trivially-correct / harmless parts to master.  Re the cmap issue, SHOULD BE FIXED PROPERLY.  It really rubs me the wrong way when you go down "let's just push this in to make bots green for now". :)

behdad:
@garretrieger can you look into this?  Ebrahim is losing sleep over it. :D

ebraminio:
> Can you explain it to me like I'm a 4-year old? :)

Hehehe, it initializes 16 bytes that is assumed in final_size without actually writing to it.

> SHOULD BE FIXED PROPERLY.

I know, I just don't like to rebase all the tests for a quick fix, giving time for an actual non rushed fix.

> "let's just push this in to make bots green for now"

The code has uninit memory read, is in use in Chrome behind a flag and causes harfbuzzjs to act differently, for such issues I believe an prompt fix is needed.

ebraminio:
> Push the trivially-correct / harmless parts to master. 

Which one, the three changes here, one makes bots to fail, the others fix the issues caused by check.

ebraminio:
Okay let's merge those harmless ones deciding about the fix in #1826

PR #1828, opened 2019-07-11
[fuzz] Don't fail when blob is empty

ebraminio:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=15735 not a real issue but am enabling more tests in our local subset fuzz runner


ebraminio:
Just logging, Timeout in tsan bot with
```
./fonts/clusterfuzz-testcase-minimized-hb-subset-fuzzer-6543700493598720
error: timeout, 
./fonts/clusterfuzz-testcase-minimized-hb-subset-fuzzer-5716947896893440
error: timeout, 
```
and in valgrind
```
./fonts/clusterfuzz-testcase-minimized-hb-subset-fuzzer-6543700493598720
error: timeout, ==33698== Memcheck, a memory error detector
==33698== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==33698== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info
==33698== Command: /root/project/test/fuzzing/.libs/hb-subset-fuzzer ./fonts/clusterfuzz-testcase-minimized-hb-subset-fuzzer-6543700493598720
==33698==```

behdad:
Why have such arbitrary conditions at all?

ebraminio:
If some local variable is write only, it can be optimized away I guess? Reached to this experimentally. I was thinking about remove whole check, apparently valgrind won't be trigged if not some jump is actually done based on an uninitialized memory.

behdad:
As I said, `hb_face_reference_blob` will checksum all tables and should catch any uninitialized-memory access.  Please verify.

ebraminio:
Rechecked, won't work without the part and the assert

behdad:
> Rechecked, won't work without the part and the assert

Please debug?!  Maybe there's something else going on.  Rather we understand than sweep under the rug.

ebraminio:
Valgrind says `Conditional jump or move depends on uninitialised value(s)` and it just makes sense to me!

ebraminio:
The part that I don't understand is however why msan doesn't see that in modes in -O0, valgrind is consistent here

behdad:
> Valgrind says `Conditional jump or move depends on uninitialised value(s)` and it just makes sense to me!

humm.  You're right.  Even checksum wouldn't catch it if it's not acted upon.

Maybe write the results to /dev/null?  Easiest.

behdad:
> The part that I don't understand is however why msan doesn't see that in modes other than -O0, valgrind is consistent here

Valgrind tracks every bit of memory.  msan doesn't quite do that.

ebraminio:
> Maybe write the results to /dev/null? Easiest.

Thought about that, specific

> Valgrind tracks every bit of memory. msan doesn't quite do that.

Still optimization level dependence mystifies me

ebraminio:
But it was failing (bad behavior) consistently in webassembly as its custom allocator

behdad:
> Still optimization level dependence mystifies me

Doesn't surprise me.

ebraminio:
> Doesn't surprise me.

Then feel free to share :)

PR #1832, opened 2019-07-14
Add hb_gdi_hfont_face_create API

### LINKED ISSUE #1831: Add hb_gdi_font_create
We want to provide hb_font_t objects for fonts in pango, across platforms. The win32 font backend in pango is written in terms of gdi and LOGFONTW.  Currently, harfbuzz has
hb_directwrite_face_create(), and it seems that going from gdi objects to directwrite ones requires going through C++, so harfbuzz might be the best place for such a helper.

ebraminio:
Based on Konstantin Ritt's work [here](https://www.mail-archive.com/harfbuzz@lists.freedesktop.org/msg06538.html).

Related or may fixes #1831, maybe not yet.

Not the final API nor tested yet, just compiles, feedbacks are welcomed.

<details>
<summary>WIP</summary>
```
/**
 * hb_gdi_face_create:
 * @hdc: a HFONT object.
 *
 * Return value: #hb_font_t object corresponding to the given input
 *
 * Since: REPLACEME
 **/
hb_font_t *
hb_gdi_font_create (HFONT hfont)
{
  hb_face_t *face = hb_gdi_face_create (hfont);
  hb_face_destroy (face);

  LOGFONT log_font;
  GetObject (hFont, sizeof (log_font), &log_font);
  logFont.lfHeight;

  hb_font_set_scale (font,
		     hfont->lfHeight,
		     hfont->lfHeight);
  return font;
}
```
</details>


ebraminio:
Yep, works OK with the following test code
```C
#include <stdio.h>

#include <hb.h>
#include <hb-uniscribe.h>

int main() {
    HDC hdc = GetDC (NULL);
    HFONT hf = CreateFont (20, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, "Tahoma");
    SelectObject (hdc, hf);
    hb_face_t *face = hb_uniscribe_hdc_face_create (hdc);
    printf ("%d", hb_face_get_glyph_count (face));
    return 0;
}
```
With no build tools, `gcc main.cc src/harfbuzz.cc -Isrc -DHAVE_UNISCRIBE -lusp10 -lgdi32 -lrpcrt4 -fno-exceptions -fno-rtti && ./a.exe` (great that I don't have to use any build tools as autotools was taking just to much time with no hope in msys2 of this machine)

Now the question is what we should get as parameter and return as a result, apparently getting hdc is the easiest but not a match to hb_face_t anyway and maybe we should go for hb_font_t, and, where the code should reside and the API name, etc.

src/hb-uniscribe.cc
```diff
@@ -603,6 +605,56 @@ hb_uniscribe_font_get_hfont (hb_font_t *font)
...
+ * Return value: #hb_face_t object corresponding to the given input
+ *
+ * Since: REPLACEME
+ **/
+hb_face_t *
+hb_uniscribe_hfont_face_create (HFONT hfont)
+{
+  // XXX: Release things...
+  HDC hdc = GetDC(nullptr);
+  SelectObject(hdc, hfont);
```
khaledhosny: What about moving these two lines to `_hb_uniscribe_hdc_reference_table` (and rename it as well)?

src/hb-uniscribe.cc
```diff
@@ -603,6 +605,56 @@ hb_uniscribe_font_get_hfont (hb_font_t *font)
...
+
+  char *buffer = NULL;
+  DWORD length = 0;
+
+  length = GetFontData (hdc, hb_uint32_swap (tag), 0, buffer, length);
+  if (unlikely (length == GDI_ERROR)) return hb_blob_get_empty ();
+
+  buffer = (char *) malloc (length);
+  length = GetFontData (hdc, hb_uint32_swap (tag), 0, buffer, length);
+  if (unlikely (length == GDI_ERROR)) length = 0;
```
khaledhosny: Should it return here as well, like it does in the check above?

src/hb-uniscribe.cc
ebraminio: Done

src/hb-uniscribe.cc
ebraminio: Done

src/hb-uniscribe.cc
```diff
@@ -603,6 +605,46 @@ hb_uniscribe_font_get_hfont (hb_font_t *font)
...
+
+  length = GetFontData (hdc, hb_uint32_swap (tag), 0, buffer, length);
+  if (unlikely (length == GDI_ERROR)) goto fail_with_releasedc;
+
+  buffer = (char *) malloc (length);
+  if (unlikely (!buffer)) goto fail_with_releasedc;
+  length = GetFontData (hdc, hb_uint32_swap (tag), 0, buffer, length);
+  if (unlikely (length == GDI_ERROR)) goto fail_with_releasedc_and_free;
+
+  return hb_blob_create ((const char *) buffer, length, HB_MEMORY_MODE_READONLY, buffer, free);
```
khaledhosny: You are not releasing the DC if there are no errors.

src/hb-uniscribe.h
```diff
@@ -40,6 +40,8 @@ hb_uniscribe_font_get_logfontw (hb_font_t *font);
 HB_EXTERN HFONT
 hb_uniscribe_font_get_hfont (hb_font_t *font);
 
+HB_EXTERN hb_face_t *
+hb_uniscribe_hfont_face_create (HFONT hfont);
```
khaledhosny: It should be `hb_gdi_*` AFAIK, Uniscribe is not involved in this API at all.

src/hb-uniscribe.cc
ebraminio: Oh, thanks, done

src/hb-uniscribe.h
ebraminio: Done

ebraminio:
Work locally also on a Windows machine so worth a PTAL I guess

src/hb.hh
```diff
@@ -476,6 +476,11 @@ static_assert ((sizeof (hb_var_int_t) == 4), "");
 /* Size signifying variable-sized array */
 #define VAR 1
 
+/* Endian swap, used in Windows related backends */
+static inline uint16_t hb_uint16_swap (const uint16_t v)
+{ return (v >> 8) | (v << 8); }
+static inline uint32_t hb_uint32_swap (const uint32_t v)
+{ return (hb_uint16_swap (v) << 16) | hb_uint16_swap (v >> 16); }
```
ebraminio: It is used in all the three hb-directwrite, hb-uniscribe and now hb-gdi thus moved it here above the similar functionality so maybe we merge them later or not.

behdad: Good.  Though, hb-algs would be better.  Or even better, move near / merge with BEInt stuff in hb-open-type.

hb.hh is typically for macro-intensive platform-specific stuff and trying to minimize it.

behdad: Ow.  hahahah.  BEInt stuff *is* in hb.hh.  Ignore me.  Thanks.

ebraminio: It is above BEInt already :) Agreed that we should consider a merge.

ebraminio:
So, not pushing for merge but please let me know about any functionality needed changed @khaledhosny @matthiasclasen @behdad 
The current API is a working HFONT -> hb_face_t but maybe we need a LOGFONTW -> hb_font_t or can or should we defer to the client? There is also a LOGFONTA and maybe someone likes to pass a HDC anyway so which one harfbuzz could or should provide, just let me thought and I will try to implement. Thanks!

matthiasclasen:
Chun-Wei Fan has put some adapttion of this work to pangos win32 backend here:

https://gitlab.gnome.org/GNOME/pango/merge_requests/91

It almost works, modulo some issue with font metrics:

https://gitlab.gnome.org/GNOME/pango/-/jobs/358239/artifacts/file/_build/hello.png

behdad:
This looks great.  Thanks.  I'm reviewing the actual code, but happy with the direction (hb-gdi vs hb-uniscribe).  I feel like hb-uniscribe should be using it internally or should be some integration?  I just checked.  I feel like both `hb-uniscribe.h` APIs should also be moved to `hb-gdi.h` and `hb-uniscribe.h` deprecated and removed (no back compat concerns).

src/hb-gdi.h
```diff
@@ -0,0 +1,39 @@
...
+#define HB_GDI_H
+
+#include "hb.h"
+
+#include <windows.h>
+
+HB_BEGIN_DECLS
+
+HB_EXTERN hb_face_t *
+hb_gdi_hfont_face_create (HFONT hfont);
```
behdad: First.  HFONT has a size, so I feel like it corresponds to `hb_font_t`.  If there's no separate equivalent to hb-face in GDI, it's fine to have both from the HFONT.  That's what we do for `FT_Face`.

Name should be `hb_gdi_font_create()` / `hb_gdi_face_create()`.

src/hb-gdi.cc
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+#include "hb.hh"
+
+#ifdef HAVE_GDI
+
+#include "hb-gdi.h"
+
+static hb_blob_t *
+_hb_gdi_hfont_reference_table (hb_face_t *face HB_UNUSED, hb_tag_t tag, void *user_data)
```
behdad: This doesn't seem to implement the null-tag fetching the entire font.  Can we see if that can also be implemented?

src/hb-gdi.cc
```diff
@@ -0,0 +1,73 @@
...
+  length = GetFontData (hdc, hb_uint32_swap (tag), 0, buffer, length);
+  if (unlikely (length == GDI_ERROR)) goto fail_with_releasedc;
+
+  buffer = (char *) malloc (length);
+  if (unlikely (!buffer)) goto fail_with_releasedc;
+  length = GetFontData (hdc, hb_uint32_swap (tag), 0, buffer, length);
+  if (unlikely (length == GDI_ERROR)) goto fail_with_releasedc_and_free;
+  ReleaseDC (nullptr, hdc);
+
+  return hb_blob_create ((const char *) buffer, length, HB_MEMORY_MODE_READONLY, buffer, free);
```
behdad: `MODE_WRITABLE` is correct since we just allocated it ourselves.

src/hb-gdi.cc
```diff
@@ -0,0 +1,73 @@
...
+
+  HDC hdc = GetDC (nullptr);
+  if (unlikely (!SelectObject (hdc, (HFONT) user_data))) goto fail;
+
+  length = GetFontData (hdc, hb_uint32_swap (tag), 0, buffer, length);
+  if (unlikely (length == GDI_ERROR)) goto fail_with_releasedc;
+
+  buffer = (char *) malloc (length);
+  if (unlikely (!buffer)) goto fail_with_releasedc;
+  length = GetFontData (hdc, hb_uint32_swap (tag), 0, buffer, length);
```
behdad: Guess GDI has no way to get table without copying.

src/hb-gdi.cc
ebraminio: It is able to do already as https://docs.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-getfontdata "If embedding is permitted, the application can retrieve the entire font file, specifying zero for the dwTable, dwOffset, and cbData parameters." and the test code that is able to write a complete font, not sure if it uses this however.

src/hb-gdi.cc
ebraminio: Done

src/hb-gdi.h
ebraminio: Reading http://www.winprog.org/tutorial/fonts.html and https://support.microsoft.com/en-ca/help/74299/info-calculating-the-logical-height-and-point-size-of-a-font I am now more familiar with this but am still not sure about how to map lfHeight of LOGFONTW object to our scales, feel free to give me a hand if you like to, or I should learn this by putting some time, that probably will help pango parallel work also, or, lets see what happens there by @fanc999

behdad: I'm fine moving forward with face for now.  And *not* moving the stuff from Uniscribe here either.  But please consider that in the future.

ebraminio: So feel free if it seems fine and you don't have API looking concerns. As I am switching between machines to access mingw for build and test, am considering installing fedora in my main machine.

behdad: Needs at least rename.  Am fine going in with rename and everything else later.

ebraminio: How I missed while I was testing this also :/

ebraminio: OK, you are looking at the outdated version I guess

behdad: Okay.  Comment thread still shows old code.  Np.

behdad:
lgtm.  The moving of integer functions should have been in separate commit. ;)

PR #1835, opened 2019-07-17
Revert "[Myanmar] Prevent reordering between Asat and Dot below"

dscorbett:
This reverts commit 1c8654ead41ca746d577549c92d2a41c594ab639. It is a simpler alternative to #1778. I don’t think fonts should rely on input strings not being normalized.


behdad:
Can we add any test here?

PR #1839, opened 2019-07-19
Add HB_NO_METRICS and fix HB_TINY build

ebraminio:


src/hb-ot-face-table-list.hh
```diff
@@ -53,9 +53,7 @@ HB_OT_ACCELERATOR (OT, cmap)
 HB_OT_TABLE (OT, hhea)
 HB_OT_ACCELERATOR (OT, hmtx)
 HB_OT_TABLE (OT, OS2)
-#ifndef HB_NO_OT_FONT_GLYPH_NAMES
 HB_OT_ACCELERATOR (OT, post)
-#endif
```
behdad: What's with this change?

ebraminio: Addressed in #1842

ebraminio:
> This also change hb-ot-font's get_font_extents. not sure if that's a deal-breaker.

Addressed in #1842 I guess

PR #1842, opened 2019-07-19
Decouple subset from metrics

### LINKED ISSUE #1840: subset use of hmtxvmtx's accelerator should be reconsidered
so we can remove the hb_ot_metrics_get_position_internal need in hb-subset thus merge of `hb-ot-metrics-internal.cc` file with `hb-ot-metrics.cc`

ebraminio:
Fixes #1840 (not exactly but well)


behdad:
Okay I'm going to take a stance here and say NO to big commits.

Please rework as small individual commits.

behdad:
And why do you need the `_common` function?  I don't understand.

src/hb-ot-metrics.cc
```diff
@@ -24,70 +24,130 @@
...
+	   GET_METRIC (hhea, ascender, y);
+  if (metrics_tag == HB_OT_METRICS_HORIZONTAL_DESCENDER)
+    return (face->table.OS2->use_typo_metrics () && GET_METRIC (OS2, sTypoDescender, y)) ||
+	   GET_METRIC (hhea, descender, y);
+  if (metrics_tag == HB_OT_METRICS_HORIZONTAL_LINE_GAP)
+    return (face->table.OS2->use_typo_metrics () && GET_METRIC (OS2, sTypoLineGap, y)) ||
+	   GET_METRIC (hhea, lineGap, y);
+  if (metrics_tag == HB_OT_METRICS_VERTICAL_ASCENDER)  return GET_METRIC (vhea, ascender, x);
+  if (metrics_tag == HB_OT_METRICS_VERTICAL_DESCENDER) return GET_METRIC (vhea, descender, x);
+  if (metrics_tag == HB_OT_METRICS_VERTICAL_LINE_GAP)  return GET_METRIC (vhea, lineGap, x);
```
behdad: What the .... is this block of code pasta?!

ebraminio:
> Okay I'm going to take a stance here and say NO to big commits.

Hehe, oh well, you are right.

> And why do you need the _common function? I don't understand.

This is to not require all the tables on slim builds.

src/hb-ot-metrics.cc
ebraminio: I can't use switch as it won't cover all the labels so a compiler warning happens, probably fixes if casted however.

ebraminio: It is just separated from `hb_ot_metrics_get_position` for size consideration.

behdad:
Just do incremental commits that are easy to agree on, and we restructure the main function after.

ebraminio:
Separated all the changes

src/hb-ot-metrics.cc
```diff
@@ -32,6 +32,19 @@
...
+_fix_ascender_descender (float value, hb_ot_metrics_t metrics_tag)
+{
+  if (metrics_tag == HB_OT_METRICS_HORIZONTAL_ASCENDER ||
+      metrics_tag == HB_OT_METRICS_VERTICAL_ASCENDER)
+    return fabs ((double) value);
+  if (metrics_tag == HB_OT_METRICS_HORIZONTAL_DESCENDER ||
+      metrics_tag == HB_OT_METRICS_VERTICAL_DESCENDER)
+    return -fabs ((double) value);
+  return value;
+}
```
ebraminio: I am not sure if we like to expose this logic to API users but you said this once IIRC, otherwise possible to do it hb-ot-font

src/hb-ot-font.cc
```diff
@@ -230,32 +230,24 @@ hb_ot_get_glyph_from_name (hb_font_t *font HB_UNUSED,
...
-  const hb_ot_face_t *ot_face = (const hb_ot_face_t *) font_data;
-  const OT::vmtx_accelerator_t &vmtx = *ot_face->vmtx;
-  metrics->ascender = font->em_scale_x (vmtx.ascender);
-  metrics->descender = font->em_scale_x (vmtx.descender);
-  metrics->line_gap = font->em_scale_x (vmtx.line_gap);
-  // TODO Hook up variations.
-  return vmtx.has_font_extents;
+  return hb_ot_metrics_get_position_common (font, HB_OT_METRICS_VERTICAL_ASCENDER, &metrics->ascender) &&
+	 hb_ot_metrics_get_position_common (font, HB_OT_METRICS_VERTICAL_DESCENDER, &metrics->descender) &&
+	 hb_ot_metrics_get_position_common (font, HB_OT_METRICS_VERTICAL_LINE_GAP, &metrics->line_gap);
```
ebraminio: doesn't initialize the variable with zero if couldn't find the value I should note.

src/hb-ot-metrics.cc
behdad: Yes, we should.  If we are fixing font crap, we should do it consistently.

src/hb-ot-font.cc
ebraminio: This also doesn't return false when both the ascender and descender are zero like what `has_font_extents = ascender != 0 || descender != 0;` was doing in `hmtx` table.

ebraminio:
Thanks for the merge and for suggesting the solution :) At least it is in a better shape now I feel

PR #1844, opened 2019-07-20
Fallback hhea's ascender/descender to OS2

ebraminio:


behdad:
Why?

ebraminio:
I was false `hhea` is less frequent than `OS2` and apparently it is a wrong assumption comparing search results `hhea` has over `OS/2` on https://gist.github.com/ebraminio/4b731a82f11a662b2164622ebb93086a

ebraminio:
Add the revert in #1842 as it is touching that part of the code anyway.

behdad:
Whatever your assumptions were, they should have been documented in the commit / PR at least.

At any rate, even the previous one I had put in was bogus since it was checking only OS/2 being present, which was the case already or we wouldn't have got `use_typo_metrics` to be true.  I think we should just trust `use_typo_metrics` and use whatever it points to.  I'll review other branch and do after merging it.

ebraminio:
> Whatever your assumptions were, they should have been documented in the commit / PR at least.

Sure, actually this was what I did originally.

> I think we should just trust use_typo_metrics and use whatever it points to.

Given the numbers now I also think that will be better to do.

khaledhosny:
IMHO the `get_metrics` API should not do any fallback and should be a low level API for getting the values from various tables untouched. It also should return an special value for not available metrics, so clients can act upon it, otherwise they can’t tell a missing metrics from one intentionally set to zero.

Some clients need different fallback strategy than what HarfBuzz might do even if HarfBuzz’s is spec-complaint (e.g. LibreOffice wants to be as compatible with MS Office as much as possible).

Higher level API like the `get_font_metrics` API can do the fallback as they want. This is also closer to what FreeType provides.

ebraminio:
> IMHO the hb_ot_metrics_get_position API should not do any fallback and should be a low level API for getting the values from various tables untouched. 

Makes sense, #1846

> It also should return an special value for not available metrics, so clients can act upon it, otherwise they can’t tell a missing metrics from one intentionally set to zero.

It is checking table availability and giving that as the API result, hope that would be enough.

khaledhosny:
> > It also should return an special value for not available metrics, so clients can act upon it, otherwise they can’t tell a missing metrics from one intentionally set to zero.
> 
> It is checking table availability and giving that as the API result, hope that would be enough.

Right, I missed the boolean return part. All good then.

PR #1846, opened 2019-07-21
[metrics] Don't fallback to hhea when OS/2 table is not available

ebraminio:


src/hb-ot-metrics.cc
```diff
@@ -69,14 +69,14 @@ hb_ot_metrics_get_position_common (hb_font_t       *font,
     (position && (*position = font->em_scalef_y (_fix_ascender_descender ( \
       face->table.TABLE->ATTR + GET_VAR, metrics_tag))), true))
   case HB_OT_METRICS_HORIZONTAL_ASCENDER:
-    return (face->table.OS2->use_typo_metrics () && GET_METRIC_Y (OS2, sTypoAscender)) ||
-	   GET_METRIC_Y (hhea, ascender);
+    return face->table.OS2->use_typo_metrics ()
+	   ? GET_METRIC_Y (OS2, sTypoAscender) : GET_METRIC_Y (hhea, ascender);
```
khaledhosny: What if `hhea` tale is missing, shouldn’t it return the `OS/2` metrics in this case even if `USE_TYPO_METRICS` is not set.

Anyway, I think this API should not do fallback at all and instead has different keys for `OS/2` typo metrics, win metrics, and `hhea` metrics. Or keep the current behavior and add something like `HB_OT_METRICS_OS_2_ASCENDER`, `HB_OT_METRICS_HHEA_ASCENDER` etc that does not do fallback.

ebraminio: > What if hhea tale is missing, shouldn’t it return the OS/2 metrics in this case even if USE_TYPO_METRICS is not set.

This is what I was assuming that hhea is less frequent than OS2 but apparently not true in macOS cirpos.

> Anyway, I think this API should not do fallback at all and instead has different keys for OS/2 typo metrics, win metrics, and hhea metrics. Or keep the current behavior and add something like HB_OT_METRICS_OS_2_ASCENDER, HB_OT_METRICS_HHEA_ASCENDER etc that does not do fallback.

This is to remove it and make it according to the spec and I don't think I can add another tag that easily.

behdad: Missing hhea is not valid IMO.

Re Khaled's point.  I agree in general.  However, I think using `use_typo_metrics` and combining the two metrics as one is logically how OpenType operates now.  That's why there's not a separate variation `MVAR` tag for the two.  The fact that a new field was added is just for backwards-compatibility BS.  I suggest we keep combining them as is.

The only other processing we are doing is fixing the sign.  Again, if we fix up, I think we should do it consistently.  Otherwise it's just asking for troubled clients.

ebraminio: OK then

khaledhosny: LibreOffice has its own fallback steps and I can’t change this anymore (every time I touch this I get people screaming about certain fonts having different line height in MS Office), so if I were to use this API (and I’d love too, LibreOffice’s OpenType parsing code is a collections of bugs that sometimes work) I need access to the “raw” table data to replicate its existing behavior.

behdad: I'm fine giving you access to raw ones using undocumented private tags.

ebraminio:
Interesting idea!

src/hb-ot-metrics.cc
ebraminio: Interesting idea! What will be value of those private tags, maybe `HB_TAG (0x7F, 0xFF, 0xFF, 0x0..0xFE)`? And what about the naming?

behdad: 1. Let's open a new issue for this,
2. I suggest 'Hasc' / 'Oasc' for hhead/os2 versions of ascender.  Same for other two.

PR #1847, opened 2019-07-21
Replace "assert (0)"s with "abort ()" to make compilers happy

ebraminio:


behdad:
assert is compiled away in production builds.  abort isn't I believe.

ebraminio:
> assert is compiled away in production builds. abort isn't I believe.

Hmm, but do we like to expose such inconsistent state to our clients rather than finding them, weird to have such distinction in not multiple leveled assertion (DEBUG_ASSERT, RELEASE_ASSERT, etc) project.

behdad:
> but do we like to expose such inconsistent state to our clients rather than finding them

you're assuming there *is* a bug.  there's no reason to believe that. :)

ebraminio:
OK then

PR #1848, opened 2019-07-21
Implement meta table parsing

ebraminio:
As "Determining whether a font is CJK (Chinese, Japanese, or Korean) or not, as in the second-last “Else” clause above, can be done by checking the 'dlng' entry (if present) of the 'meta' table, the CJK-related bits of the OS/2.ulUnicodeRange fields, or the OS/2.ulCodePageRange fields." on https://docs.microsoft.com/en-us/typography/opentype/spec/baselinetags#ideographic-em-box we need to parse and understand this table to check if a font is CJK. Fortunately it is already implemented on fonttools https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/ttLib/tables/_m_e_t_a.py

### DIFF
--- a/src/hb-ot-face-table-list.hh
+++ b/src/hb-ot-face-table-list.hh
@@ -62,6 +62,7 @@ HB_OT_ACCELERATOR (OT, name)
 #ifndef HB_NO_STAT
 HB_OT_TABLE (OT, STAT)
 #endif
+HB_OT_TABLE (OT, meta)
 
 /* Vertical layout. */
 HB_OT_TABLE (OT, vhea)
--- a/src/hb-ot-meta-table.hh
+++ b/src/hb-ot-meta-table.hh
@@ -1,0 +1,89 @@
+/*
+ * Copyright © 2019  Ebrahim Byagowi
+ *
+ *  This is part of HarfBuzz, a text shaping library.
+ *
+ * Permission is hereby granted, without written agreement and without
+ * license or royalty fees, to use, copy, modify, and distribute this
+ * software and its documentation for any purpose, provided that the
+ * above copyright notice and the following two paragraphs appear in
+ * all copies of this software.
+ *
+ * IN NO EVENT SHALL THE COPYRIGHT HOLDER BE LIABLE TO ANY PARTY FOR
+ * DIRECT, INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES
+ * ARISING OUT OF THE USE OF THIS SOFTWARE AND ITS DOCUMENTATION, EVEN
+ * IF THE COPYRIGHT HOLDER HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ *
+ * THE COPYRIGHT HOLDER SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ * FITNESS FOR A PARTICULAR PURPOSE.  THE SOFTWARE PROVIDED HEREUNDER IS
+ * ON AN "AS IS" BASIS, AND THE COPYRIGHT HOLDER HAS NO OBLIGATION TO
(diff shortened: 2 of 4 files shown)


khaledhosny:
I don’t get what do you mean by `check if a font is CJK`, the baseline does not depend on whether the font is a CJK font or not (whatever this means) but whether the text being shaped is CJK or not.

ebraminio:
It is for synthesising baseline per OT.

ebraminio:
> [..] CJK font or not (whatever this means) but whether the text being shaped is CJK or not.

It is what suggested by OT spec mentioned at the quoted sentences.

behdad:
Remove the whitespace changes from this PR.

khaledhosny:
> > [..] CJK font or not (whatever this means) but whether the text being shaped is CJK or not.
> 
> It is what suggested by OT spec mentioned at the quoted sentences.

I see, I missed that part in the link.

src/hb-ot-meta-table.hh
```diff
@@ -0,0 +1,89 @@
...
+			  dataMaps.sanitize (c, this)));
+  }
+
+  protected:
+  HBUINT32	version;	/* Version number of the metadata table — set to 1. */
+  HBUINT32	flags;		/* Flags — currently unused; set to 0. */
+  HBUINT32	dataOffset;	/* Per Apple specification:
+				 * Offset from the beginning of the table to the data.
+				 * Per OT specification:
+				 * Reserved. Not used; should be set to 0. */
```
behdad: Update to OpenType version: https://docs.microsoft.com/en-us/typography/opentype/spec/meta

ebraminio: IIUC why not have both like what it is now since it is effectless anyway, San Fransisco font is actually using the field for the mentioned use, fonttools also mentioning it not like OT spec.

ebraminio:
Considering how trouble this will involve, parsing a comma separated text to find CJK languages, I am rethinking about doing any simulation in baseline API.

src/hb-ot-meta-table.hh
behdad: Meh. Whateva'.

behdad:
> Considering how trouble this will involve, parsing a comma separated text to find CJK languages, I am rethinking about doing any simulation in baseline API.

lol. If you follow the same place in BASE table, the "else" branch after (that is, if font is determined to NOT be CJK) is to say ideographic box is undefined.  So, basically you don't need to care / determine if "font is CJK", whatever that means.

behdad:
I have no problem adding API for `meta` table using tags.  Sounds straightforward.

ebraminio:
> I have no problem adding API for meta table using tags. Sounds straightforward.

Wanted to say am reluctant to add an API for things CoreText and DirectWrite don't expose but just recalled the API CoreText has in Objective C objects [apparently?] to expose everything they got so nice to have API I guess. So let's merge and think about that later. Thanks!

behdad:
> Wanted to say am reluctant to add an API for things CoreText and DirectWrite don't expose but just recalled the API CoreText has in Objective C objects [apparently?] to expose everything they got so nice to have API I guess (or do you know any explicit API? I should recheck also). So let's merge and think about that later. Thanks!

Note that DirectWrite / CoreText are full text stacks.  More like a combo of HarfBuzz, FriBidi, FreeType, FontConfig, and probably then some... So, things we need to expose in public API to call from rest of stack they can just use internally without exposing to API.

src/hb-ot-face-table-list.hh
```diff
@@ -62,6 +62,7 @@ HB_OT_ACCELERATOR (OT, name)
 #ifndef HB_NO_STAT
 HB_OT_TABLE (OT, STAT)
 #endif
+HB_OT_TABLE (OT, meta)
```
behdad: Remove this part until we actually start using...

I think I like to add a separate area for the rarely-used tables.  Maybe doesn't matter.  But I think currently we only expose tables that are actually used.

behdad: I like having every table listed there, but commented out if unused.  That's the case for BASE and JSTF.  Not sure if there's other tables that need to be listed for completeness (and commented out).

PR #1851, opened 2019-07-22
Fix sbix glyph extents

khaledhosny:
* The `height` needs to be negated since the API returns “distance from top to bottom side”.
* Similarly, the `y_offset` needs to be added to the height to get the `y_bearing`, since sbix’s offset is “the point in the glyph relative to its lower-left corner which corresponds to the origin” while `y_bearing` is the “top side of glyph from origin”.

With these changes the sbix glyph metrics return values similar to other tables, as they were otherwise unusable.


behdad:
Lgtm. Thanks!

ebraminio:
Thanks Khaled, great to see ot-color is finally going to be used somewhere and sorry for the encountered bug.

PR #1854, opened 2019-07-22
[metrics] Fix _get_variation API to works with actual coord values

### LINKED ISSUE #1853: Why and how hb_ot_metrics_get_variation has nothing to do with a hb_face_t?
Asking about the just added API https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ot-metrics.cc#L179 @behdad has instructed me to use `face->table.MVAR->get_var (metrics_tag, nullptr, 0);` but I honestly still don't get it and thought this might be worth to ask before the release, isn't the whole point of variation on these values is to have varying e.g. ascender/descender value based on values of variation axes values? How and why this API doesn't reflect that?

ebraminio:
Closes #1853


behdad:
Lgtm.  Thanks!

PR #1856, opened 2019-07-22
[meta] New API, hb_ot_meta_reference_entry for fetching meta entries

ebraminio:


src/hb-ot-meta.cc
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+#include "hb.hh"
+
+#ifndef HB_NO_META
+
+#include "hb-ot-meta-table.hh"
+
+hb_blob_t *
+hb_ot_meta_reference_entry (hb_face_t *face, hb_tag_t tag)
```
ebraminio: It naming, and probably should get a known enum instead this hb_tag_t.

src/hb-ot-meta-table.hh
```diff
@@ -62,6 +67,19 @@ struct meta
...
   static constexpr hb_tag_t tableTag = HB_OT_TAG_meta;
 
+  struct accelerator_t
+  {
+    void init (hb_face_t *face)
+    { table = hb_sanitize_context_t ().reference_table<meta> (face); }
+    void fini () { table.destroy (); }
+
+    hb_blob_t *reference_entry (hb_tag_t tag) const
+    { return table->dataMaps.lsearch (tag, Null (DataMap)).reference_entry (table.get_blob ()); }
```
ebraminio: Maybe will be fine with bsearch

src/hb-ot-meta.cc
ebraminio: maybe this should distinguish an actual empty entry from not available entry by _or_fail like API?

src/hb-ot-meta-table.hh
ebraminio: btw, great to have all these infrastructures...

ebraminio:
I guess we can have some hb_language_t iteration of `dlng` and `slng` comma separated string as a separate API also, probably later.

ebraminio:
And maybe hb-ot-metadata instead hb-ot-meta.

src/hb-ot-meta.h
```diff
@@ -0,0 +1,57 @@
...
+ * Since: REPLACEME
+ **/
+typedef enum {
+/*
+   HB_OT_META_APPL		= HB_TAG ('a','p','p','l'),
+   HB_OT_META_BILD		= HB_TAG ('b','i','l','d'),
+*/
+  HB_OT_META_DESIGN_LANGUAGES	= HB_TAG ('d','l','n','g'),
+  HB_OT_META_SUPPORTED_LANGUAGES= HB_TAG ('s','l','n','g')
+} hb_ot_meta_t;
```
behdad: Add a max value equal to HB_TAG_MAX_SIGNED.

src/hb-ot-meta-table.hh
behdad: Yep, nice.

src/hb-ot-meta.h
ebraminio: Deciding that in #1860

ebraminio:
> Maybe add a tool like test-ot-name.cc to print them as well?

So we need a tag iterating API also I guess.

behdad:
Yes, I think so.

ebraminio:
> Maybe add a tool like test-ot-name.cc to print them as well?

Done

src/hb-ot-meta.h
ebraminio: Done

ebraminio:
Probably we need another API to get index of tag and have to modify this _reference_entry to work with indexes instead tags.

behdad:
No need for that I'd say.

src/hb-ot-meta.cc
behdad: > maybe this should distinguish an actual empty entry from not available entry by _or_fail like API?

Umm.  No let's keep as is.

ebraminio:
Ok, so maybe the patch can be considered as stable now

docs/harfbuzz-sections.txt
```diff
@@ -607,6 +607,12 @@ hb_ot_math_get_min_connector_overlap
 hb_ot_math_get_glyph_assembly
 </SECTION>
 
+<SECTION>
+<FILE>hb-ot-metadata</FILE>
+hb_ot_metadata_reference_entry
+hb_ot_metadata_t
```
behdad: I personally try putting types before their uses.  Not sure if that's done consistently.

src/hb-ot-meta-table.hh
```diff
@@ -62,6 +69,34 @@ struct meta
...
+
+    unsigned int get_entries (unsigned int      start_offset,
+			      unsigned int     *count,
+			      hb_ot_metadata_t *entries) const
+    {
+      unsigned int entries_count = table->dataMaps.len;
+      if (count && *count)
+      {
+	unsigned int len = hb_min (entries_count - start_offset, *count);
```
behdad: Huh. What if entries_count is smaller than start_offset?

I'm fairly sure we do these a bit nicer in other plcaes.  If not, we *should*.  Basically, create an hb_array_t for the whole array first.  Then make a sub_array of it with start_offset and *count, then iterate over that and copy, maybe just using hb_copy() if feasible.  Yes, I think we should change all our uses to that...

ebraminio: Done I guess

ebraminio:
> Also, would you be annoyed if i say i changed my mind re meta table and like the 'ot-meta' name better than 'ot-metadata'?

Done

ebraminio:
Yay!

PR #1857, opened 2019-07-23
[metrics] Add max value to hb_ot_metrics_t enum

ebraminio:


ebraminio:
Hmm, probably not needed

behdad:
We want to add for all enum types that are interop with hb_tag_t.  Do the same way that hb_script_t is done.  Thanks.

ebraminio:
Sure

PR #1858, opened 2019-07-23
[metrics] Add a test that actually practices variation

ebraminio:
The test font from is from https://github.com/fonttools/fonttools/blob/master/Tests/varLib/data/master_ttx_varfont_otf/TestCFF2VF.ttx


test/api/test-ot-metrics.c
```diff
@@ -40,15 +40,39 @@ test_ot_metrics_get (void)
...
+  g_assert_cmpint (hb_ot_metrics_get_x_variation (font, HB_OT_METRICS_HORIZONTAL_ASCENDER), ==, 0);
+  g_assert_cmpint (hb_ot_metrics_get_y_variation (font, HB_OT_METRICS_HORIZONTAL_ASCENDER), ==, 0);
+  g_assert_cmpint (hb_ot_metrics_get_x_variation (font, HB_OT_METRICS_X_HEIGHT), ==, 0);
+  float coords[] = {100.f};
+  hb_font_set_var_coords_design (font, coords, 1);
+  g_assert (hb_ot_metrics_get_position (font, HB_OT_METRICS_X_HEIGHT, &value));
+  g_assert_cmpint (value, ==,  478);
+  g_assert_cmpint (hb_ot_metrics_get_x_variation (font, HB_OT_METRICS_HORIZONTAL_ASCENDER), ==, 0);
+  g_assert_cmpint (hb_ot_metrics_get_y_variation (font, HB_OT_METRICS_HORIZONTAL_ASCENDER), ==, 0);
+  g_assert_cmpint (hb_ot_metrics_get_x_variation (font, HB_OT_METRICS_X_HEIGHT), ==, -8);
```
ebraminio: Given the coords, I am not if is that what really should do or not, probably you (@khaledhosny or @behdad) can check https://github.com/fonttools/fonttools/blob/master/Tests/varLib/data/master_ttx_varfont_otf/TestCFF2VF.ttx as my lack of knowledge. Thanks

behdad: Not sure.  WOn't be able to check until next week at least.

ebraminio:
Ok, lets have this now but we should check if their values are actually correct.

PR #1861, opened 2019-07-23
Add _MAX_VALUE to hb_ot_metrics_t

### LINKED ISSUE #1860: Deprecate or consistently apply _MAX_VALUE in hb_tag_t compatible enums
For sake of consistency [as the off tracker discussions] we should decide either deprecate all the _MAX_VALUE entries in enums having 4 ascii characters with HB_TAG as `_HB_SCRIPT_MAX_VALUE` , or, add those to the rest of enums that don't have, like `hb_buffer_serialize_format_t` and `hb_ot_metrics_t`.

ebraminio:
Fixes #1860

### DIFF
--- a/src/hb-ot-metrics.cc
+++ b/src/hb-ot-metrics.cc
@@ -53,7 +53,7 @@ _hb_ot_metrics_get_position_common (hb_font_t       *font,
 				    hb_position_t   *position     /* OUT.  May be NULL. */)
 {
   hb_face_t *face = font->face;
-  switch ((unsigned int) metrics_tag)
+  switch ((unsigned) metrics_tag)
   {
 #ifndef HB_NO_VAR
 #define GET_VAR face->table.MVAR->get_var (metrics_tag, font->coords, font->num_coords)
@@ -117,7 +117,7 @@ hb_ot_metrics_get_position (hb_font_t       *font,
 			    hb_position_t   *position     /* OUT.  May be NULL. */)
 {
   hb_face_t *face = font->face;
-  switch (metrics_tag)
+  switch ((unsigned) metrics_tag)
   {
   case HB_OT_METRICS_HORIZONTAL_ASCENDER:
   case HB_OT_METRICS_HORIZONTAL_DESCENDER:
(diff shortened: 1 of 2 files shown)


src/hb-buffer-serialize.cc
```diff
@@ -89,12 +89,13 @@ hb_buffer_serialize_format_from_string (const char *str, int len)
...
 {
-  switch ((unsigned) format)
+  switch (format)
   {
     case HB_BUFFER_SERIALIZE_FORMAT_TEXT:	return serialize_formats[0];
     case HB_BUFFER_SERIALIZE_FORMAT_JSON:	return serialize_formats[1];
     default:
-    case HB_BUFFER_SERIALIZE_FORMAT_INVALID:	return nullptr;
+    case HB_BUFFER_SERIALIZE_FORMAT_INVALID:
+    case _HB_BUFFER_SERIALIZE_MAX_VALUE:	return nullptr;
```
behdad: Ugh. No.

src/hb-buffer-serialize.cc
```diff
@@ -89,12 +89,13 @@ hb_buffer_serialize_format_from_string (const char *str, int len)
 const char *
 hb_buffer_serialize_format_to_string (hb_buffer_serialize_format_t format)
 {
-  switch ((unsigned) format)
+  switch (format)
```
behdad: Keep the cast.  We should cast in all cases of using these kinds of enums in switch.  I think otherwise compiler is free to assume that value is only one of the ones declared in the enum.

src/hb-buffer-serialize.cc
```diff
@@ -484,6 +486,7 @@ hb_buffer_deserialize_glyphs (hb_buffer_t *buffer,
 
     default:
     case HB_BUFFER_SERIALIZE_FORMAT_INVALID:
+    case _HB_BUFFER_SERIALIZE_MAX_VALUE:
```
behdad: Yeah I don't like these.  We have default there already.

behdad: I see why you added.  Let's add casts instead.

PR #1863, opened 2019-07-24
[subset] cmap table to use _subset2 and new iterator frameworks

qxliu76:


src/hb-ot-cmap-table.hh
```diff
@@ -851,182 +929,62 @@ struct cmap
...
 
-  struct subset_plan
+  template<typename Iterator,
+           hb_requires (hb_is_iterator (Iterator))>
+  void serialize (hb_serialize_context_t *c,
+                  Iterator it,
+                  const EncodingRecord *has_unicode_bmp,
+                  const EncodingRecord *has_unicode_ucs4,
+                  const EncodingRecord *has_ms_bmp,
+                  const EncodingRecord *has_ms_ucs4)
```
behdad: I find this api confusing.  These names sound like should be boolean, but they are actual records, which can be nullptr I guess.

PR #1867, opened 2019-07-25
[metrics] Expose raw OS2/HHEA asc/dsc values using private tags

### LINKED ISSUE #1866: [ot-metrics] Define undocumented values for untouched hhea/OS2 independently from typometrics flag
It has been requested by suggested by @khaledhosny that we expose untouched values of OS/2 in an API but since that is not covered by https://docs.microsoft.com/en-us/typography/opentype/spec/mvar#value-tags we like to go for undocumented private tags (starting with uppercase) for them instead. The [current suggestion](https://github.com/harfbuzz/harfbuzz/pull/1846#discussion_r307413202) by @behdad is `'Hasc' / 'Oasc' for hhead/os2 versions of ascender. Same for other two.` but maybe we don't actually want `hhea` ones explicitly or need to do that for linegap also, depending what clients would like we to have.

ebraminio:
Fixes #1866 (a proposed fix)

### DIFF
--- a/src/hb-ot-metrics.cc
+++ b/src/hb-ot-metrics.cc
@@ -100,6 +100,14 @@ _get_gasp (hb_face_t *face, float *result, hb_ot_metrics_t metrics_tag)
 }
 #endif
 
+/* Private tags for https://github.com/harfbuzz/harfbuzz/issues/1866 */
+#define _HB_OT_METRICS_HORIZONTAL_ASCENDER_OS2   HB_TAG ('O','a','s','c')
+#define _HB_OT_METRICS_HORIZONTAL_ASCENDER_HHEA  HB_TAG ('H','a','s','c')
+#define _HB_OT_METRICS_HORIZONTAL_DESCENDER_OS2  HB_TAG ('O','d','s','c')
+#define _HB_OT_METRICS_HORIZONTAL_DESCENDER_HHEA HB_TAG ('H','d','s','c')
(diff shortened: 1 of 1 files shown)


behdad:
Rename `_HB_OT_METRICS_OS2_HORIZONTAL_ASCENDER` to `_HB_OT_METRICS_HORIZONTAL_ASCENDER_OS2` for better grouping.

By private I really meant private.  Ie. not exposed in our public headers.  Eg. like:
```
$ git grep ZAWGYI
hb-ot-shape-complex.hh:    case HB_SCRIPT_MYANMAR_ZAWGYI:
hb.hh:#define HB_SCRIPT_MYANMAR_ZAWGYI  ((hb_script_t) HB_TAG ('Q','a','a','g'))
```

ebraminio:
Sure, done. I was thinking about reusing the login in _common but that probably complicates things, let me know anyway.

behdad:
Lgtm. What logic in _common? I don't think we want sign-fix these raw values.

ebraminio:
> Lgtm. What logic in _common? I don't think we want sign-fix these raw values.

Great. Aha, not that, I wanted to first make _common fallback of this and make _common call itself instead having repeating `GET_METRIC_Y (` macros but probably won't worth the hassle as the complication.

ebraminio:
So lets merge @khaledhosny also feels good about this.

khaledhosny:
LGTM

PR #1868, opened 2019-07-25
[colr][feat][meta] Port sub_array iteration to dagger

ebraminio:


src/hb-aat-layout-lcar-table.hh
```diff
@@ -55,8 +55,7 @@ struct lcar
     if (caret_count)
     {
       hb_array_t<const HBINT16> arr = array.sub_array (start_offset, caret_count);
-      unsigned int count = arr.length;
-      for (unsigned int i = 0; i < count; ++i)
+      for (unsigned int i = 0; i < arr.length; ++i)
```
ebraminio: Something inside stack allocated hb_array_t doesn't differ much with an explicitly stack allocated variable from optimizing standpoint I guess.

ebraminio:
> Umm. No. This would write into *count.

That is intended, isn't.

src/hb-ot-meta-table.hh
```diff
@@ -42,7 +42,7 @@ struct DataMap
 {
   int cmp (hb_tag_t a) const { return tag.cmp (a); }
 
-  hb_tag_t get_tag () const { return tag; }
+  hb_ot_meta_t get_tag () const { return (hb_ot_meta_t) (hb_tag_t) tag; }
```
behdad: The way I mentally think about this is, the `meta` table implementation only returns `hb_tag_t`, and `hb_ot_meta_t` is a detail of the public API.  So, the cast belongs to the API implementation to me.

Also, maybe better name for `hb_ot_meta_t`?  We have `hb_ot_name_id_t`.  `hb_ot_meta_tag_t` sounds better to me.

src/hb-ot-color-colr-table.hh
```diff
@@ -39,13 +39,21 @@ namespace OT {
 
 struct LayerRecord
 {
+  hb_ot_color_layer_t get_color_layer () const
```
behdad: Implement as cast operator maybe?

src/hb-aat-layout-lcar-table.hh
```diff
@@ -55,8 +55,7 @@ struct lcar
...
 	switch (format)
```
behdad: This switch is so badly places.  Should have been subformats.  Fix later please.

src/hb-aat-layout-lcar-table.hh
behdad: Right.  But it's a pattern we have all over codebase.  Fine either way.  Just don't like mixing with other changes.

src/hb-aat-layout-feat-table.hh
```diff
@@ -47,17 +47,18 @@ struct SettingName
   hb_aat_layout_feature_selector_t get_selector () const
   { return (hb_aat_layout_feature_selector_t) (unsigned) setting; }
 
-  void get_info (hb_aat_layout_feature_selector_info_t *s,
-			hb_aat_layout_feature_selector_t default_selector) const
+  hb_aat_layout_feature_selector_info_t get_info (hb_aat_layout_feature_selector_t default_selector) const
```
behdad: Again, see if can implement as cast operator?

src/hb-ot-meta-table.hh
ebraminio: Done

src/hb-aat-layout-lcar-table.hh
ebraminio: Done

src/hb-ot-color-colr-table.hh
ebraminio: Nice! Done

src/hb-aat-layout-feat-table.hh
ebraminio: It has an argument so I don't will be feasible unlike the other will worked nicely

src/hb-ot-color-colr-table.hh
```diff
@@ -39,13 +39,21 @@ namespace OT {
 
 struct LayerRecord
 {
+  operator hb_ot_color_layer_t () const
+  {
+    hb_ot_color_layer_t result = {0};
+    result.glyph = glyphId;
+    result.color_index = colorIdx;
```
behdad: Since `hb_ot_color_layer_t` is public and never changes, just initialize in one line?

src/hb-aat-layout-feat-table.hh
```diff
@@ -47,17 +47,18 @@ struct SettingName
...
-    s->disable = default_selector == HB_AAT_LAYOUT_FEATURE_SELECTOR_INVALID ?
-		 (hb_aat_layout_feature_selector_t) (s->enable + 1) :
-		 default_selector;
+    s.enable = (hb_aat_layout_feature_selector_t) (unsigned int) setting;
+    s.disable = default_selector == HB_AAT_LAYOUT_FEATURE_SELECTOR_INVALID ?
+		(hb_aat_layout_feature_selector_t) (s.enable + 1) :
+		default_selector;
 
-    s->reserved = 0;
+    s.reserved = 0;
```
behdad: Same. Just use the initializer since this struct won't change?

Issue #1875, opened 2019-07-27
refactor lcar switch


ebraminio:
The switch here https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-aat-layout-lcar-table.hh#L59a per https://github.com/harfbuzz/harfbuzz/pull/1868#discussion_r307893450 is badly placed, this and its gsub equivalent should be refactored to dagger also, which I couldn't.

ebraminio:
Let's close this

PR #1876, opened 2019-07-27
[lcar] flip for and switch position

ebraminio:
#1875 related


behdad:
This is better.  Thanks.  Ideally we should define different structs for different format... like we do everywhere else.

PR #1877, opened 2019-07-28
[base] Add hb_ot_layout_get_baseline API

### LINKED ISSUE #418: support for 'hang' BASE OpenType feature
Please support the [BASE](https://www.microsoft.com/typography/otspec/baselinetags.htm) OpenType feature. I've gathered all information I was able to find on [a dedicated Github repo](https://github.com/eroux/otf-base-hang). There's a font with a hang BASE table in the [LO bug report](https://bugs.documentfoundation.org/show_bug.cgi?id=104930). Apparently recent MS Word support it, OSX apps don't, but they support the [AAT bsln](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6bsln.html) table, which is more or less equivalent.

ebraminio:
Fixes #418
Closes #1354 (that one goes for `bsln` also which this one isn't, but this one has [a not finalized] simulation per the spec)

### DIFF
--- a/src/hb-ot-layout.cc
+++ b/src/hb-ot-layout.cc
@@ -1940,75 +1940,34 @@ hb_ot_layout_substitute_lookup (OT::hb_ot_apply_context_t *c,
   apply_string<GSUBProxy> (c, lookup, accel);
 }
 
-#if 0
+#ifndef HB_NO_BASE
+/**
+ * hb_ot_layout_get_baseline:
+ * @font: a font
+ * @baseline: a baseline tag
+ * @direction: text direction.
+ * @script_tag:  script tag.
+ * @language_tag: language tag.
+ * @coord: (out): baseline value if found.
+ *
+ * Fetches a baseline value from the face.
+ *
+ * Return value: if found baseline value in the the font.
+ **/
 hb_bool_t
-hb_ot_layout_get_baseline (hb_font_t               *font,
-			   hb_ot_layout_baseline_t  baseline,
-			   hb_direction_t           direction,
-			   hb_tag_t                 script_tag,
-			   hb_tag_t                 language_tag,
-			   hb_position_t           *coord        /* OUT.  May be NULL. */)
+hb_ot_layout_get_baseline (hb_font_t                   *font,
+			   hb_ot_layout_baseline_tag_t  baseline_tag,
+			   hb_direction_t               direction,
+			   hb_tag_t                     script_tag,
+			   hb_tag_t                     language_tag,
+			   hb_position_t               *coord        /* OUT.  May be NULL. */)
 {
-  bool result = font->face->table.BASE->get_baseline (font, baseline, direction, script_tag,
-						      language_tag, coord);
+  bool result = font->face->table.BASE->get_baseline (font, baseline_tag, direction, script_tag, language_tag, coord);
 
-  /* TODO: Simulate https://docs.microsoft.com/en-us/typography/opentype/spec/baselinetags#ideographic-em-box */
-  if (!result && coord) *coord = 0;
-
-  if (coord) *coord = font->em_scale_dir (*coord, direction);
+  if (result && coord)
+    *coord = HB_DIRECTION_IS_HORIZONTAL (direction) ? font->em_scale_y (*coord) : font->em_scale_x (*coord);
 
   return result;
 }
-
-/* To be moved to public header */
-/*
- * BASE
- */
-
-/**
- * hb_ot_layout_baseline_t:
- *
- * https://docs.microsoft.com/en-us/typography/opentype/spec/baselinetags
- *
- * Since: DONTREPLACEME
- */
-typedef enum {
-  HB_OT_LAYOUT_BASELINE_HANG = HB_TAG('h','a','n','g'),
-  HB_OT_LAYOUT_BASELINE_ICFB = HB_TAG('i','c','f','b'),
-  HB_OT_LAYOUT_BASELINE_ICFT = HB_TAG('i','c','f','t'),
-  HB_OT_LAYOUT_BASELINE_IDEO = HB_TAG('i','d','e','o'),
-  HB_OT_LAYOUT_BASELINE_IDTB = HB_TAG('i','d','t','b'),
-  HB_OT_LAYOUT_BASELINE_MATH = HB_TAG('m','a','t','h'),
-  HB_OT_LAYOUT_BASELINE_ROMN = HB_TAG('r','o','m','n'),
-
-  _HB_OT_LAYOUT_BASELINE_MAX_VALUE = HB_TAG_MAX_SIGNED /*< skip >*/
-} hb_ot_layout_baseline_t;
-
-
-/**
- * hb_ot_layout_get_baseline:
- * @font: The #hb_font_t to work upon
- * @baseline: The #hb_ot_layout_baseline_t to query
- * @direction: The #hb_direction_t text direction to use (horizontal or vertical)
- * @script_tag:  #hb_tag_t of the script to use
- * @language_tag: #hb_tag_t of the language to use
- * @coord: (out): The position of the requested baseline
- *
- * Fetches the coordinates of the specified baseline in the face, underneath
- * the specified script and language and in the specified text direction.
- *
- * Return value: true if the baseline is found for the settings queried, false otherwise
- *
- **/
-HB_EXTERN hb_bool_t
-hb_ot_layout_get_baseline (hb_font_t               *font,
-			   hb_ot_layout_baseline_t  baseline,
-			   hb_direction_t           direction,
-			   hb_tag_t                 script_tag,
-			   hb_tag_t                 language_tag,
-			   hb_position_t           *coord        /* OUT.  May be NULL. */);
-
 #endif
-
-
 #endif
--- a/src/hb-ot-layout.h
+++ b/src/hb-ot-layout.h
@@ -402,6 +402,38 @@ hb_ot_layout_feature_get_characters (hb_face_t      *face,
 				     unsigned int   *char_count    /* IN/OUT.  May be NULL */,
 				     hb_codepoint_t *characters    /* OUT.     May be NULL */);
 
+/*
+ * BASE
+ */
+
+/**
+ * hb_ot_layout_baseline_tag_t:
+ *
+ * https://docs.microsoft.com/en-us/typography/opentype/spec/baselinetags
+ *
+ * Since: REPLACEME
+ */
+typedef enum {
+  HB_OT_LAYOUT_BASELINE_TAG_ROMAN			= HB_TAG ('r','o','m','n'),
+  HB_OT_LAYOUT_BASELINE_TAG_HANGING			= HB_TAG ('h','a','n','g'),
+  HB_OT_LAYOUT_BASELINE_TAG_IDEO_FACE_BOTTOM_OR_LEFT	= HB_TAG ('i','c','f','b'),
+  HB_OT_LAYOUT_BASELINE_TAG_IDEO_FACE_TOP_OR_RIGHT	= HB_TAG ('i','c','f','t'),
+  HB_OT_LAYOUT_BASELINE_TAG_IDEO_EMBOX_BOTTOM_OR_LEFT	= HB_TAG ('i','d','e','o'),
+  HB_OT_LAYOUT_BASELINE_TAG_IDEO_EMBOX_TOP_OR_RIGHT	= HB_TAG ('i','d','t','p'),
+  HB_OT_LAYOUT_BASELINE_TAG_MATH			= HB_TAG ('m','a','t','h'),
+
+  _HB_OT_LAYOUT_BASELINE_TAG_MAX_VALUE = HB_TAG_MAX_SIGNED /*< skip >*/
+} hb_ot_layout_baseline_tag_t;
+
+HB_EXTERN hb_bool_t
+hb_ot_layout_get_baseline (hb_font_t                   *font,
+			   hb_ot_layout_baseline_tag_t  baseline_tag,
+			   hb_direction_t               direction,
+			   hb_tag_t                     script_tag,
+			   hb_tag_t                     language_tag,
+			   hb_position_t               *coord        /* OUT.  May be NULL. */);
+
+
 HB_END_DECLS
 
 #endif /* HB_OT_LAYOUT_H */
--- a/docs/harfbuzz-sections.txt
+++ b/docs/harfbuzz-sections.txt
@@ -548,6 +548,7 @@ HB_OT_TAG_GDEF
 HB_OT_TAG_GPOS
 HB_OT_TAG_GSUB
 HB_OT_TAG_JSTF
+hb_ot_layout_baseline_tag_t
 hb_ot_layout_collect_lookups
 hb_ot_layout_collect_features
 hb_ot_layout_feature_get_characters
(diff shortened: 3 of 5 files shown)


src/hb-ot-layout.cc
```diff
@@ -1949,7 +1949,90 @@ hb_ot_layout_get_baseline (hb_font_t               *font,
 			   hb_tag_t                 language_tag,
 			   hb_position_t           *coord        /* OUT.  May be NULL. */)
 {
-  return font->face->table.BASE->get_baseline (font, baseline_tag, direction, script_tag, language_tag, coord);
+  hb_face_t *face = font->face;
+#define INVALID INT_MAX
+  *coord = INVALID;
```
ebraminio: Returning false when couldn't find the value inside the font and INT_MAX in the value when simulation has failed.

src/hb-ot-layout.cc
```diff
@@ -1949,7 +1962,89 @@ hb_ot_layout_get_baseline (hb_font_t               *font,
...
+  bool is_horizontal = HB_DIRECTION_IS_HORIZONTAL (direction);
+  hb_position_t ideoEmboxBottom = INVALID, ideoEmboxLeft = INVALID, ideoEmboxTop = INVALID, ideoEmboxRight = INVALID;
+  hb_position_t icfBottom = INVALID, icfLeft = INVALID, icfTop = INVALID, icfRight = INVALID;
+  struct { hb_position_t ideo = INVALID, idtp = INVALID, icfb = INVALID, icft = INVALID; } HorizAxis, VertAxis;
+  hb_position_t margin = INVALID;
+#define GET_BASE(axis, member) font->face->table.BASE->get_baseline (font, member##_tag, \
+				&axis == &HorizAxis ? HB_DIRECTION_TTB : HB_DIRECTION_LTR, \
+				script_tag, language_tag, &axis.member)
+
+  /* https://docs.microsoft.com/en-us/typography/opentype/spec/baselinetags#ideographic-em-box */
```
ebraminio: Tried to mimic the specification with a DSL, still not sure if I've got it right as lack of indention on the pseudocode...

src/hb-ot-layout.cc
```diff
@@ -1941,6 +1941,19 @@ hb_ot_layout_substitute_lookup (OT::hb_ot_apply_context_t *c,
...
+ * @font: a font
+ * @baseline: a baseline tag
+ * @direction: text direction.
+ * @script_tag:  script tag.
+ * @language_tag: language tag.
+ * @coord: (out): baseline value, INT_MAX when simulation also failed.
+ *
+ * Fetches a baseline value from the face, simulate if is possible.
+ *
+ * Return value: true only if the baseline is found inside the font.
```
ebraminio: Maybe we can return a tristate enum instead, enum { NONE, EXISTS, SIMULATED } or something.

src/hb-ot-layout.cc
```diff
@@ -1949,7 +1962,89 @@ hb_ot_layout_get_baseline (hb_font_t               *font,
...
+  if (GET_BASE (VertAxis, idtp))
+    ideoEmboxRight = VertAxis.idtp;
+  else
+    ideoEmboxRight = upem;
+  if (GET_BASE (VertAxis, ideo) && VertAxis.ideo)
+  { /* Warning: Bad VertAxis.ideo value */ }
+  else /* If this is a CJK font: */ /* But we haven't implemented this check */
+  {
+    ideoEmboxBottom = font->face->table.OS2->sTypoDescender;
+    ideoEmboxTop = font->face->table.OS2->sTypoAscender;
```
ebraminio: I couldn't use the just introduced ot-metrics API as the it is needed to have the value unscaled till the end.

src/hb-ot-layout.cc
dscorbett: [The old version of the page](http://web.archive.org/web/20170714012736/https://www.microsoft.com/typography/otspec/baselinetags.htm) has proper indentation.

ebraminio: Coool, thanks!

ebraminio:
Please note that this API doesn't need language_tag but we are getting as the discussions (its min_max part needs however which is separate from this).

Maybe this should go in hb-ot-baseline instead hb-ot-layout so then we can expose min_max part also.

src/hb-ot-layout.h
```diff
@@ -402,6 +402,38 @@ hb_ot_layout_feature_get_characters (hb_face_t      *face,
...
+ * Since: REPLACEME
+ */
+typedef enum {
+  HB_OT_LAYOUT_BASELINE_TIBETAN_HANGING					= HB_TAG('h','a','n','g'),
+  HB_OT_LAYOUT_BASELINE_IDEOGRAPHIC_CHARACTER_FACE_BOTTOM_OR_LEFT_EDGE	= HB_TAG('i','c','f','b'),
+  HB_OT_LAYOUT_BASELINE_IDEOGRAPHIC_CHARACTER_FACE_TOP_OR_RIGHT_EDGE	= HB_TAG('i','c','f','t'),
+  HB_OT_LAYOUT_BASELINE_IDEOGRAPHIC_EMBOX_BOTTOM_OR_LEFT_EDGE		= HB_TAG('i','d','e','o'),
+  HB_OT_LAYOUT_BASELINE_IDEOGRAPHIC_EMBOX_TOP_OR_RIGHT_EDGE		= HB_TAG('i','d','t','p'),
+  HB_OT_LAYOUT_BASELINE_MATHEMATICAL_CHARACTERS_CENTER			= HB_TAG('m','a','t','h'),
+  HB_OT_LAYOUT_BASELINE_ROMAN						= HB_TAG('r','o','m','n'),
```
ebraminio: Something I've chosen, please review all of them and suggest better ones.

src/hb-ot-layout.cc
```diff
@@ -1940,75 +1940,117 @@ hb_ot_layout_substitute_lookup (OT::hb_ot_apply_context_t *c,
...
-
-  /* TODO: Simulate https://docs.microsoft.com/en-us/typography/opentype/spec/baselinetags#ideographic-em-box */
-  if (!result && coord) *coord = 0;
+#define INVALID INT_MAX
+  *coord = INVALID;
+  bool result = font->face->table.BASE->get_baseline (font, baseline_tag, direction, script_tag, language_tag, coord);
+
+#define SCALE(value) (value != INVALID ? (HB_DIRECTION_IS_HORIZONTAL (direction) \
+					  ? font->em_scale_y (value) \
+					  : font->em_scale_x (value)) : value)
```
ebraminio: Do _y scale if the given script direction is horizontal and do _x scale if is vertical, matches what ot-metrics IIUC.

src/hb-ot-layout.h
```diff
@@ -402,6 +402,38 @@ hb_ot_layout_feature_get_characters (hb_face_t      *face,
...
+ * Since: REPLACEME
+ */
+typedef enum {
+  HB_OT_LAYOUT_BASELINE_TIBETAN_HANGING					= HB_TAG ('h','a','n','g'),
+  HB_OT_LAYOUT_BASELINE_IDEOGRAPHIC_CHARACTER_FACE_BOTTOM_OR_LEFT_EDGE	= HB_TAG ('i','c','f','b'),
+  HB_OT_LAYOUT_BASELINE_IDEOGRAPHIC_CHARACTER_FACE_TOP_OR_RIGHT_EDGE	= HB_TAG ('i','c','f','t'),
+  HB_OT_LAYOUT_BASELINE_IDEOGRAPHIC_EMBOX_BOTTOM_OR_LEFT_EDGE		= HB_TAG ('i','d','e','o'),
+  HB_OT_LAYOUT_BASELINE_IDEOGRAPHIC_EMBOX_TOP_OR_RIGHT_EDGE		= HB_TAG ('i','d','t','p'),
+  HB_OT_LAYOUT_BASELINE_MATHEMATICAL_CHARACTERS_CENTER			= HB_TAG ('m','a','t','h'),
+  HB_OT_LAYOUT_BASELINE_ROMAN						= HB_TAG ('r','o','m','n'),
```
behdad: need `_TAG` in value names.  They must match enum name.

src/hb-ot-layout.h
```diff
@@ -402,6 +402,38 @@ hb_ot_layout_feature_get_characters (hb_face_t      *face,
...
+
+/**
+ * hb_ot_layout_baseline_tag_t:
+ *
+ * https://docs.microsoft.com/en-us/typography/opentype/spec/baselinetags
+ *
+ * Since: REPLACEME
+ */
+typedef enum {
+  HB_OT_LAYOUT_BASELINE_TIBETAN_HANGING					= HB_TAG ('h','a','n','g'),
```
behdad: Remove TIBETAN.

src/hb-ot-layout.h
```diff
@@ -402,6 +402,38 @@ hb_ot_layout_feature_get_characters (hb_face_t      *face,
...
+  HB_OT_LAYOUT_BASELINE_IDEOGRAPHIC_CHARACTER_FACE_BOTTOM_OR_LEFT_EDGE	= HB_TAG ('i','c','f','b'),
+  HB_OT_LAYOUT_BASELINE_IDEOGRAPHIC_CHARACTER_FACE_TOP_OR_RIGHT_EDGE	= HB_TAG ('i','c','f','t'),
+  HB_OT_LAYOUT_BASELINE_IDEOGRAPHIC_EMBOX_BOTTOM_OR_LEFT_EDGE		= HB_TAG ('i','d','e','o'),
+  HB_OT_LAYOUT_BASELINE_IDEOGRAPHIC_EMBOX_TOP_OR_RIGHT_EDGE		= HB_TAG ('i','d','t','p'),
+  HB_OT_LAYOUT_BASELINE_MATHEMATICAL_CHARACTERS_CENTER			= HB_TAG ('m','a','t','h'),
```
behdad: Just `MATH`.

src/hb-ot-layout.h
```diff
@@ -402,6 +402,38 @@ hb_ot_layout_feature_get_characters (hb_face_t      *face,
...
+ * https://docs.microsoft.com/en-us/typography/opentype/spec/baselinetags
+ *
+ * Since: REPLACEME
+ */
+typedef enum {
+  HB_OT_LAYOUT_BASELINE_TIBETAN_HANGING					= HB_TAG ('h','a','n','g'),
+  HB_OT_LAYOUT_BASELINE_IDEOGRAPHIC_CHARACTER_FACE_BOTTOM_OR_LEFT_EDGE	= HB_TAG ('i','c','f','b'),
+  HB_OT_LAYOUT_BASELINE_IDEOGRAPHIC_CHARACTER_FACE_TOP_OR_RIGHT_EDGE	= HB_TAG ('i','c','f','t'),
+  HB_OT_LAYOUT_BASELINE_IDEOGRAPHIC_EMBOX_BOTTOM_OR_LEFT_EDGE		= HB_TAG ('i','d','e','o'),
+  HB_OT_LAYOUT_BASELINE_IDEOGRAPHIC_EMBOX_TOP_OR_RIGHT_EDGE		= HB_TAG ('i','d','t','p'),
```
behdad: `IDEOGRAPHIC` -> `IDEO`.  Remove `_CHARACTER`.  Remove `_EDGE`.

src/hb-ot-layout-base-table.hh
```diff
@@ -376,10 +373,10 @@ struct BaseScriptList
 
 struct Axis
 {
-  bool get_baseline (hb_ot_layout_baseline_t   baseline,
-		     hb_tag_t                  script_tag,
-		     hb_tag_t                  language_tag,
-		     const BaseCoord         **coord) const
+  bool get_baseline (hb_ot_layout_baseline_tag_t   baseline,
```
behdad: Just take `hb_tag_t` here.  The type comes in just in public API.

src/hb-ot-layout.cc
```diff
@@ -1940,75 +1940,117 @@ hb_ot_layout_substitute_lookup (OT::hb_ot_apply_context_t *c,
...
+			   hb_position_t               *coord        /* OUT.  May be NULL. */)
 {
-  bool result = font->face->table.BASE->get_baseline (font, baseline, direction, script_tag,
-						      language_tag, coord);
-
-  /* TODO: Simulate https://docs.microsoft.com/en-us/typography/opentype/spec/baselinetags#ideographic-em-box */
-  if (!result && coord) *coord = 0;
+#define INVALID INT_MAX
+  *coord = INVALID;
+  bool result = font->face->table.BASE->get_baseline (font, baseline_tag, direction, script_tag, language_tag, coord);
```
behdad: You shouldn't write to `coord` until we find.

behdad: Ugh.  I hate this function.

Remove all the simulation.  Let's get a simple fetcher implementation in first.  Then we talk about simulation if needed.

src/hb-ot-layout.cc
behdad: No need for tristate.  If needed, is fine to "return false and output simulated".  We do that in some other API.

src/hb-ot-layout.h
ebraminio: So hb-ot-meta and hb-ot-metrics enums also?

ebraminio:
> Please push the noncontroversial parts into master and create/update PR for the parts that need discussion. I'm highly suspicious of needing to simulate stuff the way spec specifies. So, I like to stare at that separately and discuss it in a separate PR.

Do you see any noncontroversial part in this? I don't :) Was nice if GitHub had Gerrit like dependent changes

src/hb-ot-layout.cc
behdad: Umm. Need to apply the sign conversion.  Plus, variations.  I think you have to use the metrics API.  Again, remove the fallback and we rebuild it separately.

src/hb-ot-layout.h
behdad: Yes.  Enums always have to be consistent like that.

src/hb-ot-layout.cc
ebraminio: Removed simulation and put it here for future reference https://github.com/harfbuzz/harfbuzz/compare/master...ebraminio:baseline-base

ebraminio:
All done, let me know if you like hb-ot-baseline more instead hb-ot-layout.

src/hb-ot-layout.h
```diff
@@ -402,6 +402,38 @@ hb_ot_layout_feature_get_characters (hb_face_t      *face,
...
+ * Since: REPLACEME
+ */
+typedef enum {
+  HB_OT_LAYOUT_BASELINE_TAG_HANGING			= HB_TAG ('h','a','n','g'),
+  HB_OT_LAYOUT_BASELINE_TAG_IDEO_FACE_BOTTOM_OR_LEFT	= HB_TAG ('i','c','f','b'),
+  HB_OT_LAYOUT_BASELINE_TAG_IDEO_FACE_TOP_OR_RIGHT	= HB_TAG ('i','c','f','t'),
+  HB_OT_LAYOUT_BASELINE_TAG_IDEO_EMBOX_BOTTOM_OR_LEFT	= HB_TAG ('i','d','e','o'),
+  HB_OT_LAYOUT_BASELINE_TAG_IDEO_EMBOX_TOP_OR_RIGHT	= HB_TAG ('i','d','t','p'),
+  HB_OT_LAYOUT_BASELINE_TAG_MATH			= HB_TAG ('m','a','t','h'),
+  HB_OT_LAYOUT_BASELINE_TAG_ROMAN			= HB_TAG ('r','o','m','n'),
```
behdad: Move ROMAN to be first, since it's the most common...

behdad:
Looks great.  Don't forget adding to docs.  Merge after that minor issue.

PR #1887, opened 2019-07-30
[cpal] port to dagger

ebraminio:


src/hb-ot-color-cpal-table.hh
```diff
@@ -146,8 +146,9 @@ struct CPAL
       /* Always return numColors colors per palette even if it has out-of-bounds start index. */
       unsigned int count = hb_min ((unsigned) hb_max ((int) (numColors - start_offset), 0), *color_count);
```
ebraminio: Intentionally didn't touch the logic here.

PR #1888, opened 2019-08-02
Expose italic angle as fourth field of hb_font_extents_t

### LINKED ISSUE #1880: Expose post table’s italicAngle?

ebraminio:
As @khaledhosny suggestion, fixes #1880

### DIFF
--- a/src/hb-ot-font.cc
+++ b/src/hb-ot-font.cc
@@ -228,6 +228,32 @@ hb_ot_get_glyph_from_name (hb_font_t *font HB_UNUSED,
 }
 #endif
 
+#define _HB_OT_STYLE_SLANT HB_TAG ('s','l','n','t')
+#define _HB_OT_STYLE_ITALIC HB_TAG ('i','t','a','l')
+
+static float
+hb_ot_get_font_italic_angle (hb_font_t *font)
+{
+#ifndef NO_VAR
+  hb_ot_var_axis_info_t info;
+  /* Check first if is set by variation as
+     https://docs.microsoft.com/en-us/typography/opentype/spec/dvaraxistag_slnt */
+  if (hb_ot_var_find_axis_info (font->face, _HB_OT_STYLE_SLANT, &info) &&
+      info.axis_index < font->num_coords)
+    return (float) font->coords[info.axis_index] / 16384.f;
+
+  if (hb_ot_var_find_axis_info (font->face, _HB_OT_STYLE_ITALIC, &info) &&
+      info.axis_index < font->num_coords)
+    return (float) font->coords[info.axis_index] * -12.f / 16384.f;
+#endif
+  /* TODO: Maybe should be fetched from STAT values first */
+#ifndef HB_NO_METRICS
+  return font->face->table.post->table->italicAngle.to_float ();
+#else
+  return 0;
+#endif
+}
+
 static hb_bool_t
 hb_ot_get_font_h_extents (hb_font_t *font,
 			  void *font_data HB_UNUSED,
@@ -236,7 +262,8 @@ hb_ot_get_font_h_extents (hb_font_t *font,
 {
   return _hb_ot_metrics_get_position_common (font, HB_OT_METRICS_TAG_HORIZONTAL_ASCENDER, &metrics->ascender) &&
 	 _hb_ot_metrics_get_position_common (font, HB_OT_METRICS_TAG_HORIZONTAL_DESCENDER, &metrics->descender) &&
-	 _hb_ot_metrics_get_position_common (font, HB_OT_METRICS_TAG_HORIZONTAL_LINE_GAP, &metrics->line_gap);
+	 _hb_ot_metrics_get_position_common (font, HB_OT_METRICS_TAG_HORIZONTAL_LINE_GAP, &metrics->line_gap) &&
+	 (metrics->italic_angle = hb_ot_get_font_italic_angle (font), true);
 }
 
 static hb_bool_t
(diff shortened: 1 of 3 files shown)


behdad:
We first need to figure out what exactly we are exposing...  Like, the 'ital' variations axis need to be preferred if available.

Maybe just turn it to a float?

src/hb-ot-font.cc
```diff
@@ -228,6 +228,32 @@ hb_ot_get_glyph_from_name (hb_font_t *font HB_UNUSED,
...
+  hb_ot_var_axis_info_t info;
+  /* Check first if is set by variation as
+     https://docs.microsoft.com/en-us/typography/opentype/spec/dvaraxistag_slnt */
+  if (hb_ot_var_find_axis_info (font->face, _HB_OT_STYLE_SLANT, &info) &&
+      info.axis_index < font->num_coords)
+    return (float) font->coords[info.axis_index] / 16384.f;
+
+  if (hb_ot_var_find_axis_info (font->face, _HB_OT_STYLE_ITALIC, &info) &&
+      info.axis_index < font->num_coords)
+    return (float) font->coords[info.axis_index] * -12.f / 16384.f;
```
ebraminio: Italic axis is in 0..1 range and 12° apparently is what italic means in Arial, and it should be negative for "forward" (!) values, https://docs.microsoft.com/en-us/typography/opentype/spec/post

src/hb-ot-font.cc
```diff
@@ -228,6 +228,32 @@ hb_ot_get_glyph_from_name (hb_font_t *font HB_UNUSED,
...
+static float
+hb_ot_get_font_italic_angle (hb_font_t *font)
+{
+#ifndef NO_VAR
+  hb_ot_var_axis_info_t info;
+  /* Check first if is set by variation as
+     https://docs.microsoft.com/en-us/typography/opentype/spec/dvaraxistag_slnt */
+  if (hb_ot_var_find_axis_info (font->face, _HB_OT_STYLE_SLANT, &info) &&
+      info.axis_index < font->num_coords)
+    return (float) font->coords[info.axis_index] / 16384.f;
```
ebraminio: Maybe subject of race condition? I wish coords be represented with hb_array_t internally.

src/hb-ot-font.cc
jfkthame: I don't think we should be trying to convert an 'ital' axis value to an italic angle at all.

Perhaps there should be an MVAR tag to allow the 'post' table's italicAngle field to respond to variations. In the absence of that, the most reasonable heuristic to synthesize an italicAngle for a variation font, if we have some reason to believe it *should* vary, would be based on the caretSlopeRise and caretSlopRun values.

behdad: Actually.... my bad again.  I meant `slnt` axis, not `ital`.  In fact, there's enough words in the spec to tell us exactly what to do.  From MVAR spec:
```
Note: The OS/2.usWeightClass, OS/2.usWidthClass and post.italicAngle values are not supported by variation data in the 'MVAR' table. This is because values for these three fields correspond directly to input axis values for the 'wght', 'wdth' and 'slnt' variation axes. See the discussion of these axes in the OpenType Design-Variation Axis Tag Registry for details on the relationship between these fields and the corresponding design axes.
```
https://docs.microsoft.com/en-us/typography/opentype/otspec182/mvar

From the axis tags description for `slnt`:
```
Note that the scale for the Slant axis is interpreted as the angle of slant in counter-clockwise degrees from upright. This means that a typical, right-leaning oblique design will have a negative slant value. This matches the scale used for the italicAngle field in the 'post' table.

In a variable font that implements 'slnt' variations, the value in the italicAngle field of the 'post' table must match the default 'slnt' value specified in the 'fvar' table. For non-default instances of a variable font, the 'slnt' axis value can be used as the post.italicAngle value for the instance.
```
https://docs.microsoft.com/en-us/typography/opentype/otspec182/dvaraxistag_slnt

The spec for `wght` and `wdth` also have similar wording in reference to their OS/2 counterparts.

So!  I propose we add the three tags `slnt`, `wght`, and `wdth` to access those three values in the ot-metrics API.  These will return the currently-set variation value for the item (in user-space, which we currently don't keep but we want to) definitely clamped to the valid range only, and falling back to the `post` / `OS/2` table values.

If we add separate ot-style API later, we probably will expose the same logic there too...  Or maybe not.  That API may only expose what's encoded in `STAT` only, with very limited fallback if there's no `STAT` at all, in that case maybe exposing the three items.

ebraminio:
>
>
> I think this should return ptem value on opsz also, wish I could just go
without hb_face_t as found unnormalizing coordinates tricky but will try
more.

src/hb-ot-font.cc
behdad: Changing font while using it from another thread is unsupported behavior.

behdad: And no `hb_array_t`, nothing short of a mutex will fix that.

behdad:
> I think this should return ptem value on opsz also, wish I could just go
> without hb_face_t as found unnormalizing coordinates tricky but will try more.

Not sure what you mean.

We should switch to keeping design coordinates in hb_font_t.  Trying to convert back works right now but won't with avar2.

src/hb-ot-font.cc
ebraminio: > Changing font while using it from another thread is unsupported behavior.

Good to know and makes sense also

> And no hb_array_t, nothing short of a mutex will fix that.

Correct, I should've reached to that

ebraminio:
> Not sure what you mean.

Meant I liked to have some hb_face_t oriented API but now hb-ot-style work also works like with hb_font_t and able to return just set coords

ebraminio:
> We should switch to keeping design coordinates in hb_font_t. Trying to convert back works right now but won't with avar2.

Makes sense, I have the unnormalize patch uploaded, we will need that anyway as we have set normalized coords in out API I guess, where is avar2 specification btw, nice to see I think

behdad:
> where is avar2 specification btw, nice to see I think

In... my... head...?

ebraminio:
As a part of PR cleanups I'm doing these days, this feels unlikely be a solution or at least I don't see that way but feel free to pick it up on a separate PR. Thanks

behdad:
I know how to do this one. Was going to do it eventually.

behdad:
The request is legit.

ebraminio:
The request is legit so #1880 is open, let's continue there.

PR #1889, opened 2019-08-02
[subset] GPOS Singlepos

qxliu76:


src/hb-ot-layout-gpos-table.hh
```diff
@@ -583,6 +676,55 @@ struct SinglePosFormat2
...
+  void serialize (hb_serialize_context_t *c,
+		  Iterator glyph_val_iter_pairs,
+                  ValueFormat valFormat)
+  {
+    if (unlikely (!c->extend_min (u.format))) return;
+    unsigned format = 2;
+
+    if (glyph_val_iter_pairs)
+    {
+      auto get_format = [&] (Iterator glyph_val_iter_pairs)
```
garretrieger: Can you move this function definition outside of this method?

PR #1890, opened 2019-08-04
[style] New API, hb_style_get_value

### LINKED ISSUE #1337: Exposing OS/2 table?

### LINKED ISSUE #1880: Expose post table’s italicAngle?

ebraminio:
Searches variation axes of a hb_font_t object for a specific axis first,
if not set, then tries to get default style values from different
tables of the font.

Related patches, https://github.com/harfbuzz/harfbuzz/compare/master...ebraminio:style https://github.com/harfbuzz/harfbuzz/compare/master...ebraminio:style2 https://github.com/harfbuzz/harfbuzz/compare/master...ebraminio:italicangle #1468


behdad:
Haven't looked at what you've done, but related:
https://github.com/harfbuzz/harfbuzz/pull/1888#discussion_r310377314

behdad:
At first glance I like this, but see my comment linked above.  This needs to hook up to font's currently-set variation values as well.  And of course, what you have is the fallback path.  The (missing) main path should be reading `STAT` table.  And the API to expose that table is a lot more involved.  So we shouldn't rush this, but keep evolving it till it's complete.

ebraminio:
> This needs to hook up to font's currently-set variation values as well.

The values here don't scale and are like `hb_ot_var_find_axis_info`.default_value which works with hb_face_t. Say in a font management system we want to know about different aspects of a face, why one should create a hb_font_t just to learn about a face.

> And of course, what you have is the fallback path. The (missing) main path should be reading STAT table. 

Was working on it, done

> And the API to expose that table is a lot more involved.

I'd say lets do that later, fallback won't be needed there even, mixing that with this just complicates things for a simple font crawling application.

ebraminio:
If someday we decide to expose one of these flags, `OS2.fselection.{UNDERSCORE,OUTLINED,STRIKEOUT}` or `post.isFixedPitch` or `head.macStyle.{UNDERLINE,OUTLINE,SHADOW}` (which we don't, just theoretically), for example outlined flag which we can check both OS2 and head for better compatibility, don't we go for a naming like `hb_bool_t hb_ot_style_is_outlined (hb_face_t *face)`, I think the API here is also similar to that, working upon a face, giving best possible result.

test/api/test-ot-style.c
```diff
@@ -0,0 +1,116 @@
...
+  hb_font_set_var_named_instance (font, 6);
+
+  g_assert_cmpfloat_hb (hb_ot_style_get_value (font, HB_OT_STYLE_TAG_ITALIC), 0);
+  g_assert_cmpfloat_hb (hb_ot_style_get_value (font, HB_OT_STYLE_TAG_OPTICAL_SIZE), 12);
+  g_assert_cmpfloat_hb (hb_ot_style_get_value (font, HB_OT_STYLE_TAG_SLANT), 0);
+  g_assert_cmpfloat_hb (hb_ot_style_get_value (font, HB_OT_STYLE_TAG_WIDTH), 100);
+  /* XXX: It should be 900! */
+  g_assert_cmpfloat_hb (hb_ot_style_get_value (font, HB_OT_STYLE_TAG_WEIGHT), 1);\
+  /* XXX: It should be 50! */
+  g_assert_cmpfloat_hb (hb_ot_style_get_value (font, (hb_ot_style_tag_t) HB_TAG ('C','N','T','R')), 0.5);
```
ebraminio: Oh I see why but not sure if we like to do that :/

src/hb-ot-var-avar-table.hh
```diff
@@ -128,7 +128,7 @@ struct avar
     const SegmentMaps *map = &firstAxisSegmentMaps;
     for (unsigned int i = 0; i < count; i++)
     {
-      coords[i] = map->map (coords[i]);
+//       coords[i] = map->map (coords[i]);
```
ebraminio: guess can't reverse this easily.

src/hb-ot-var-avar-table.hh
```diff
@@ -128,11 +162,24 @@ struct avar
...
```
ebraminio: guess can't reversed easily, test-ot-style test is working without this not sure why the unmap doesn't work however.

test/api/test-ot-style.c
```diff
@@ -0,0 +1,175 @@
...
+
+#include <hb-ot.h>
+
+/* Unit tests for hb-ot-style.h */
+
+#ifndef g_assert_cmpfloat_with_epsilon
+#define g_assert_cmpfloat_with_epsilon(n1, n2, epsilon) g_assert_cmpfloat ((int) roundf (n1), ==, (int) roundf (n2))
+#endif
+
+#define EPSILON 0.01
```
ebraminio: This is its precision on retrieving original coords which I believe is pretty good.

ebraminio:
Dropped `_ot_` as most of the tags can be considered for non OT fonts also.

How about renaming `hb_style_get_value` to `hb_font_get_style_value`? Guess `hb_style_tag_t` naming should be remained, or, going for `hb_font_style_tag_t` also?

ebraminio:
@behdad can you review this also? Thanks!

behdad:
> @behdad can you review this also? Specially on `hb_style_get_value` vs `hb_font_get_style_value` and `hb_style_t` part. Thanks!

Not now.  I'm very conflicted on this still :).  Let me recover a bit and be more comfortable with other code that has gone in first.  Like to come to a conclusion on the pen stuff first.  Thanks.

ebraminio:
as previous off tracker agreement we had on having this as an experimental API, just like to experiment with the API on other project to show its usefulness and guess there will be some clients for this just like the draw API clients found on github so they can comment on usefulness and correctness of the API also so let's have this experimental API for now.

ebraminio:
@khaledhosny, this can be used like `hb_style_get_value (font, HB_TAG ('s','l','n','t'))` to retrieve [italic angle](https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-style.cc#L46) of both VF and non-VF fonts, something may useful for your request here #1880, of course no serious app should dynamic link to such API but may useful for ones you control yourself, and, one can have

```C
hb_position_t run, rise;
hb_ot_metrics_get_position (font, HB_OT_METRICS_TAG_HORIZONTAL_CARET_RUN, run);
hb_ot_metrics_get_position (font, HB_OT_METRICS_TAG_HORIZONTAL_CARET_RISE, rise);
float angle = -asinf ((float) run / rise) * 180.f / (float) M_PI;
```

to construct the angle using current public API but guess not much of VF fonts do `mvar` also correctly.

This fills the absence of such APIs https://github.com/RazrFalcon/ttf-parser/blob/master/c-api/ttfparser.h#L253-L314 also which can be used for font managers and font book like apps. An old technology given VF font advances, still is relevant for many fonts on the font.

PR #1893, opened 2019-08-07
Implement fvar/avar inverse mapping logic

ebraminio:


behdad:
I really dislike your abuse of templates here!

ebraminio:
> I really dislike your abuse of templates here!

Because everyone likes only their own abuses 😂

ebraminio:
Done

src/hb-ot-var-avar-table.hh
```diff
@@ -133,14 +168,27 @@ struct avar
     }
   }
 
+  int unmap_coord (unsigned int index, int coord) const
```
behdad: Like `map_coords`, implement `unmap_coords` that does all coords in one go.  Otherwise it will become O(n^2).  It doesn't have to be.

src/hb-ot-var-avar-table.hh
```diff
@@ -93,6 +93,41 @@ struct SegmentMaps : ArrayOf<AxisValueMap>
...
+
+    if (unlikely (arrayZ[i-1].toCoord == arrayZ[i].toCoord))
+      return arrayZ[i-1].fromCoord;
+
+    int denom = arrayZ[i].toCoord - arrayZ[i-1].toCoord;
+    return arrayZ[i-1].fromCoord +
+	   ((arrayZ[i].fromCoord - arrayZ[i-1].fromCoord) *
+	    (value - arrayZ[i-1].toCoord) + denom/2) / denom;
+  }
+
```
behdad: Okay I'm appreciating your template hack a lot more right now...

Maybe moving the common logic into a function that takes pointer to `fromCoords` and `toCoords` with a stride of 4 bytes... so the logic is shared as well as binary code?

behdad: I have wanted to add stride template parameter to `hb_array_t`.  Not sure if it's a good idea / can be easily extended.  But we definitely have a use for it in a couple of places, including this one.

src/hb-ot-var-avar-table.hh
ebraminio: Done

src/hb-ot-var-avar-table.hh
ebraminio: Done

ebraminio:
Merged to #1905 (while keeping the changes separate)

behdad:
Merging separately since easier.

PR #1896, opened 2019-08-07
[os2] Replace null pool addr compare by checking vital fields

ebraminio:
It should have the mentioned fields otherwise I wonder if it useful at all.

### DIFF
--- a/src/hb-ot-os2-table.hh
+++ b/src/hb-ot-os2-table.hh
@@ -59,7 +59,7 @@ struct OS2V1Tail
 
 struct OS2V2Tail
 {
-  bool has_data () const { return this != &Null (OS2V2Tail); }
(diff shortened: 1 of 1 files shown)


behdad:
Fine. :)

src/hb-ot-os2-table.hh
```diff
@@ -59,7 +59,7 @@ struct OS2V1Tail
 
 struct OS2V2Tail
 {
-  bool has_data () const { return this != &Null (OS2V2Tail); }
+  bool has_data () const { return sxHeight || sCapHeight; }
```
behdad: This actually fixes a bug since previously the V2Tail was coming after the OS2 table and as such wouldn't compare equal to Null(OS2VTail).

PR #1899, opened 2019-08-09
[WIP] Replace hb_bool_t with bool

ebraminio:
Fails in test suit and `hb-shape --verify`, interesting at least for educational reasons to see why

### DIFF
--- a/src/hb-unicode.h
+++ b/src/hb-unicode.h
@@ -198,12 +198,12 @@ hb_unicode_funcs_reference (hb_unicode_funcs_t *ufuncs);
 HB_EXTERN void
 hb_unicode_funcs_destroy (hb_unicode_funcs_t *ufuncs);
 
-HB_EXTERN hb_bool_t
+HB_EXTERN bool
 hb_unicode_funcs_set_user_data (hb_unicode_funcs_t *ufuncs,
 			        hb_user_data_key_t *key,
 			        void *              data,
 			        hb_destroy_func_t   destroy,
-				hb_bool_t           replace);
+				bool                replace);
 
 
 HB_EXTERN void *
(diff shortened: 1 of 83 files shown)


src/hb-unicode.h
```diff
@@ -240,16 +240,16 @@ typedef hb_script_t			(*hb_unicode_script_func_t)		(hb_unicode_funcs_t *ufuncs,
...
+typedef bool			(*hb_unicode_compose_func_t)		(hb_unicode_funcs_t *ufuncs,
+									 hb_codepoint_t      a,
+									 hb_codepoint_t      b,
+									 hb_codepoint_t     *ab,
+									 void               *user_data);
+typedef bool			(*hb_unicode_decompose_func_t)		(hb_unicode_funcs_t *ufuncs,
+									 hb_codepoint_t      ab,
+									 hb_codepoint_t     *a,
+									 hb_codepoint_t     *b,
+									 void               *user_data);
```
dscorbett: The second and third columns should stay aligned with the columns of the other typedefs.

behdad:
Why do this when you can instead just do:
```
diff --git a/src/hb-common.h b/src/hb-common.h
index edd9ffb8..b3c5b814 100644
--- a/src/hb-common.h
+++ b/src/hb-common.h
@@ -87,7 +87,9 @@ typedef unsigned __int64 uint64_t;
 HB_BEGIN_DECLS
 
 
-typedef int hb_bool_t;
+#include <stdbool.h>
+
+typedef bool hb_bool_t;
 
 typedef uint32_t hb_codepoint_t;
 typedef int32_t hb_position_t;
```

Anyway.  This has no benefit and just possibility of breaking API/ABI.

The test suite / utils breakage is because they assume hb_bool_t is same size as int.  Something that we cannot change.

src/hb-unicode.h
ebraminio: Done

ebraminio:
Reopened for uploading a working patch, am closing after CI passes,

This has been a fruitful experiment however :) let me document it for myself!
1. Like this https://bugs.freedesktop.org/attachment.cgi?id=65187 casting bool* to int*, sending that pointer means memory corruption.
2. Discovered util/hb-subset is using private API, which it shouldn't, and, we should put some check to disable such private API in util/ sources after fixing it.
3. This _may_ work if we don't have any hb_bool_t* (according to mailing list) or exposing a struct with hb_bool_t on it (so some client might used the pointer of hb_bool_t, like # 2 which caused by using a private struct with hb_bool_t), which we don't! And maybe not in LE machines(?), let's not do risky thing.

So maybe having this in harfbuzz2, if ever. Thanks!

Issue #1891, opened 2019-08-06
getting optical margins out of the font


matthiasclasen:
The OpenType spec describes lfdb and rfbd features that supposedly contain the information I need to implement optical margins: https://docs.microsoft.com/en-us/typography/opentype/spec/features_ko#lfbd

But it is very unclear to me how 

> Application interface: For GIDs found in the 'lfbd' coverage table, the application passes a GID to the table and gets back a new XPlacement and XAdvance value

translates to harfbuzz api.

khaledhosny:
There is no API for this, nor there are any fonts using these features to the best of my knowledge. After thinking about this for a long time, I think using layout features for optical margins was misguided since applications need this information after layout and during line breaking (at least PDFTeX and its descendants, I don’t know any other implementation), and using OpenType features makes for an awkward API.

Apple’s [`opbd`](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6opbd.html) table might be a better bet, though I doubt there are many fonts using it (certainly not on Linux).

jfkthame:
FWIW, a quick scan of fonts on my macOS machine finds just two that have `opbd` tables, and they're both fonts (Apple Chancery and Hoefler Text) with roots back in the TrueType GX days. I suspect the presence of the table is just a historical accident at this point, and am doubtful it is really maintained or used -- I think it was an idea that failed to gain any traction.

khaledhosny:
That is probably two fonts more than fonts with `opbd`/`lfbd`/`rfbd` features. But what I’m trying to say if fonts were to support optical bounds now, the Apple one is the best of the two.

matthiasclasen:
so how do other text rendering stacks to optical margins? heuristics?

khaledhosny:
In PDFTeX/XeTeX you have to set the values manually for each glyph. In Scribus it is some hard-coded value for certain punctuation character (Scribus is ugly). No idea what InDesign does, and I don’t know any other applications that support optical margins.

matthiasclasen:
I fonts don't have the data, i guess that kills it.

jfkthame:
Something of a chicken and egg situation, perhaps. If no software implements this feature, there's no motivation for font designers to include the data (and no way for them to check the result of providing it). And if fonts don't include the data, there's no motivation for software to implement support for it.

matthiasclasen:
True.

Getting back to the original question: How would a harfbuzz api for opbd look like?

[...]

PR #1900, opened 2019-08-10
[opbd] Implement table parsing

ebraminio:
#1891 related

### DIFF
--- a/src/hb-aat-layout-opbd-table.hh
+++ b/src/hb-aat-layout-opbd-table.hh
@@ -1,0 +1,111 @@
+/*
+ * Copyright © 2019  Ebrahim Byagowi
+ *
+ *  This is part of HarfBuzz, a text shaping library.
+ *
+ * Permission is hereby granted, without written agreement and without
+ * license or royalty fees, to use, copy, modify, and distribute this
+ * software and its documentation for any purpose, provided that the
+ * above copyright notice and the following two paragraphs appear in
+ * all copies of this software.
+ *
+ * IN NO EVENT SHALL THE COPYRIGHT HOLDER BE LIABLE TO ANY PARTY FOR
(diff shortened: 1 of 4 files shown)


src/hb-aat-layout-opbd-table.hh
```diff
@@ -0,0 +1,111 @@
...
+
+struct OpticalBounds
+{
+  bool sanitize (hb_sanitize_context_t *c) const
+  {
+    TRACE_SANITIZE (this);
+    return_trace (likely (c->check_struct (this)));
+  }
+
+  HBINT16	leftSide;
```
behdad: FWORD or something?

src/hb-aat-layout-opbd-table.hh
```diff
@@ -0,0 +1,111 @@
...
+  HBINT16	bottomSide;
+  public:
+  DEFINE_SIZE_STATIC (8);
+};
+
+struct opbd
+{
+  static constexpr hb_tag_t tableTag = HB_AAT_TAG_opbd;
+
+  bool get_optical_bounds (hb_font_t *font, hb_codepoint_t glyph_id,
```
behdad: The `optical` in the method name is redundant.  Also maybe make it return a glyph_extents_t?

src/hb-aat-layout-opbd-table.hh
```diff
@@ -0,0 +1,111 @@
...
+    TRACE_SANITIZE (this);
+    return_trace (likely (c->check_struct (this) &&
+			  version.major == 1 &&
+			  lookupTable.sanitize (c, this)));
+  }
+
+  protected:
+  FixedVersion<>version;	/* Version number of the optical bounds
+				 * table (0x00010000 for the current version). */
+  HBUINT16	format;		/* Format of the optical bounds table.
```
behdad: Again, this is not how we handle multi-format tables.  The way we do it is more verbose and requires more code.  But I want you to do that.  Same in `lcar` table.  There's no next time..

PR #1902, opened 2019-08-14
 [opbd] Use multiformat convention on the table

ebraminio:


behdad:
Thank you!

In other places we typically keep the format in all subtables as well (as well as in the union).  I'm not particularly attached to that though.  This is good enough for me.

Issue #1883, opened 2019-07-30
hb_font_set_variations forgets the current coords


matthiasclasen:
It uses 0 for every axis that is not specified in the variations.

This was unexpected to me when I did this:

```
hb_font_set_var_named_instance (font, instance);
hb_font_set_variations (font, just_one_axis, 1);
```

matthiasclasen:
I ended up changing things around and instead of calllng hb_font_set_var_named_instance, do the following:

```
hb_ot_var_named_instance_get_design_coords (...)
/*  override some coords */
hb_font_set_var_coords_design (...)
```

behdad:
Yes this is bothersome to me...  I'm leaning towards making font remember which named instance it's pegged to, and apply variations on top of that when set using `set_variations`.  Not sure what about when set with `set_var_coords_design` with a short vector.  The latter resetting unspecified axes to their default (of the pegged instance) makes most sense to me.

behdad:
Am reviewing API changes for release, and came here to express again how much this bothers me!

I'm making the release but let's figure this out soon.

ebraminio:
I think I like to disagree here, we should have one other API like `hb_font_set_variation` that just touches one axix without having anything to do with named instances, keeping set_variations semantic clear.

behdad:
I'm getting nervous about this mess we have right now.  Let's aim for an understanding and agreement here.  @jfkthame can you please chip in?

behdad:
Basically question is, I think, do we want to have a set of "sticky" variations on the font, and then adjustments automatically applied on top of that?  Or maybe that's the wrong way to think about it.

jfkthame:
What I think makes sense to me from an API point of view is that `hb_font_set_variations` should set *only* the variation axes that are specifically passed to it, leaving any others untouched.

If I have a font that has multiple axes, I should be able to set them one by one in separate calls to `hb_font_set_variations`, or (more efficiently, perhaps) all at once in a single call, but the end result would be the same.

`hb_font_set_var_named_instance` sets all axes, of course; `hb_font_set_variations` could then be used to adjust one (or more) of the axes from the values that the instance specified, so that the font no longer represents that instance, but any axes that aren't mentioned in `hb_font_set_variations` would keep the values that the instance gave them.

[...]

PR #1905, opened 2019-08-14
[var] Implement hb_font_get_var_coords_design

### LINKED ISSUE #1865: Recover design coords from a hb_font_t
I want to create a ui for font variations.

I have a hb_font_t. From that I can get the hb_face_t, and all the axis infos to set up sliders.

But there is currently no way for me to find the coords of the hb_font_t to set the sliders current value. hb_font_get_var_coords_normalized gives me _something_, but it does not obviously relate to the min/max/default values in the axis info.

ebraminio:
Hold design coords and simulate when normalized coords are set directly.

Fixes #1865

### DIFF
--- a/src/hb-font.cc
+++ b/src/hb-font.cc
@@ -33,6 +33,9 @@
 
 #include "hb-ot.h"
 
+#include "hb-ot-var-avar-table.hh"
+#include "hb-ot-var-fvar-table.hh"
+
 
 /**
  * SECTION:hb-font
@@ -1309,6 +1312,7 @@ DEFINE_NULL_INSTANCE (hb_font_t) =
 
   0, /* num_coords */
   nullptr, /* coords */
+  nullptr, /* design_coords */
 
   const_cast<hb_font_funcs_t *> (&_hb_Null_hb_font_funcs_t),
 
@@ -1360,6 +1364,20 @@ hb_font_create (hb_face_t *face)
   return font;
 }
 
+static void
+_hb_font_adopt_var_coords (hb_font_t *font,
+			   int *coords, /* 2.14 normalized */
+			   float *design_coords,
+			   unsigned int coords_length)
+{
+  free (font->coords);
+  free (font->design_coords);
+
+  font->coords = coords;
+  font->design_coords = design_coords;
+  font->num_coords = coords_length;
+}
+
 /**
  * hb_font_create_sub_font:
  * @parent: parent font.
@@ -1389,15 +1407,22 @@ hb_font_create_sub_font (hb_font_t *parent)
   font->y_ppem = parent->y_ppem;
   font->ptem = parent->ptem;
 
-  font->num_coords = parent->num_coords;
-  if (font->num_coords)
+  unsigned int num_coords = parent->num_coords;
+  if (num_coords)
   {
-    unsigned int size = parent->num_coords * sizeof (parent->coords[0]);
-    font->coords = (int *) malloc (size);
-    if (unlikely (!font->coords))
-      font->num_coords = 0;
+    int *coords = (int *) calloc (num_coords, sizeof (parent->coords[0]));
+    float *design_coords = (float *) calloc (num_coords, sizeof (parent->design_coords[0]));
+    if (likely (coords && design_coords))
+    {
+      memcpy (coords, parent->coords, num_coords * sizeof (parent->coords[0]));
+      memcpy (design_coords, parent->design_coords, num_coords * sizeof (parent->design_coords[0]));
+      _hb_font_adopt_var_coords (font, coords, design_coords, num_coords);
+    }
     else
-      memcpy (font->coords, parent->coords, size);
+    {
+      free (coords);
+      free (design_coords);
+    }
   }
 
   return font;
@@ -1457,6 +1482,7 @@ hb_font_destroy (hb_font_t *font)
   hb_font_funcs_destroy (font->klass);
 
   free (font->coords);
+  free (font->design_coords);
 
   free (font);
 }
@@ -1818,17 +1844,6 @@ hb_font_get_ptem (hb_font_t *font)
  * Variations
  */
 
-static void
-_hb_font_adopt_var_coords_normalized (hb_font_t *font,
-				      int *coords, /* 2.14 normalized */
-				      unsigned int coords_length)
-{
-  free (font->coords);
-
-  font->coords = coords;
-  font->num_coords = coords_length;
-}
-
 /**
  * hb_font_set_variations:
  *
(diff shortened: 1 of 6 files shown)


test/api/test-var-coords.c
```diff
@@ -0,0 +1,75 @@
...
+
+#include <hb-ot.h>
+
+/* Unit tests for hb_font_[gs]et_var_coords_ */
+
+#ifndef g_assert_cmpfloat_with_epsilon
+#define g_assert_cmpfloat_with_epsilon(n1, n2, epsilon) g_assert_cmpfloat ((int) roundf (n1), ==, (int) roundf (n2))
+#endif
+
+#define EPSILON 0.05
```
ebraminio: Converting back precision, client can round with such precision for getting a reliable design coord

test/api/test-var-coords.c
```diff
@@ -0,0 +1,75 @@
...
+  g_assert_cmpint ((int) hb_font_get_var_coords_design (font, NULL)[0], ==, 206);
+
+  for (float weight = 200; weight < 901; ++weight)
+  {
+    int normalized;
+    hb_ot_var_normalize_coords (face, 1, &weight, &normalized);
+    hb_font_set_var_coords_normalized (font, &normalized, 1);
+    float converted_back = hb_font_get_var_coords_design (font, NULL)[0];
+    // fprintf (stderr, "%f: %d => %f\n", weight, normalized, converted_back);
+    g_assert_cmpfloat_with_epsilon ((long double) converted_back, (long double) weight, EPSILON);
```
ebraminio: glib uses long double and if we don't cast we will double promotion.

ebraminio:
As it is now test covered also, PTAL. It unblocks my ot-style work also.

ebraminio:
Some feedback here will be nice as it blocks ot-style and font slant angle retrieval, unless, we decide to expose those in hb_face_t not hb_font_t.

behdad:
Will check soon.  Reviewed the avar/fvar changes.  Is this dependent on that or not?

ebraminio:
> Will check soon. Reviewed the avar/fvar changes. Is this dependent on that or not?

Merged the two.

src/hb-font.cc
```diff
@@ -1389,15 +1407,23 @@ hb_font_create_sub_font (hb_font_t *parent)
...
+  if (num_coords)
   {
-    unsigned int size = parent->num_coords * sizeof (parent->coords[0]);
-    font->coords = (int *) malloc (size);
-    if (unlikely (!font->coords))
-      font->num_coords = 0;
-    else
-      memcpy (font->coords, parent->coords, size);
+    unsigned int coords_size = num_coords * sizeof (parent->coords[0]);
+    int *coords = (int *) malloc (coords_size);
```
behdad: calloc.

src/hb-font.cc
```diff
@@ -1389,15 +1407,23 @@ hb_font_create_sub_font (hb_font_t *parent)
...
+    if (likely (coords))
+    {
+      unsigned int design_coords_size = num_coords * sizeof (parent->design_coords[0]);
+      float *design_coords = (float *) malloc (design_coords_size);
```
behdad: calloc.

behdad: Also, do both calloc calls then check both results together?

src/hb-font.cc
```diff
@@ -1854,10 +1870,25 @@ hb_font_set_variations (hb_font_t *font,
   if (unlikely (coords_length && !normalized))
     return;
 
+  float *design_coords = coords_length ? (float *) calloc (coords_length, sizeof (float)) : nullptr;
+  if (unlikely (coords_length && !design_coords))
+  {
+    free (normalized);
+    return;
+  }
```
behdad: Same here.  Combine the allocation-failure checking.

src/hb-font.cc
```diff
@@ -1854,10 +1870,25 @@ hb_font_set_variations (hb_font_t *font,
...
+
+  for (unsigned int i = 0; i < variations_length; i++)
+  {
+    hb_ot_var_axis_info_t info;
+    if (hb_ot_var_find_axis_info (font->face, variations[i].tag, &info) &&
+	info.axis_index < coords_length)
+      design_coords[info.axis_index] = variations[i].value;
+  }
```
behdad: Rewrite this function to populate design coords completely, then normalize them...

src/hb-font.cc
```diff
@@ -1877,8 +1908,18 @@ hb_font_set_var_coords_design (hb_font_t *font,
   if (unlikely (coords_length && !normalized))
     return;
 
+  float *design_coords = coords_length ? (float *) calloc (coords_length, sizeof (float)) : nullptr;
+  if (unlikely (coords_length && !design_coords))
+  {
+    free (design_coords);
+    return;
+  }
```
behdad: Combine allocation/failure checks.

src/hb-font.cc
ebraminio: Done

src/hb-font.cc
ebraminio: Done

src/hb-font.cc
ebraminio: Done

src/hb-font.cc
ebraminio: Done

ebraminio:
We should either have this or make ot-style to get hb_face_t, so

> How are we going to communicate, when avar2 is adopted, that design coords are not available?

There should be some, even non unique, design coord for a normalized coord even after avar2? If so why we should care about uniqueness I wonder

> This is completely lacking in named-instance handling. Might be slightly orthogonal, but I will feel a lot more comfortable if we answer questions in #1883 first.

I see them differently but understand your concern, gave my 2c there.

src/hb-font.cc
ebraminio: don't know what "completely" means here but have merged normalize to the code no extra axis fetching.

ebraminio:
Wished not being able for transforming back values for not yet developed version of a not very essential table wasn't blocking things like returning italic angle and other nice things ot-style returns for non-VF fonts but apparently that is the case.

ebraminio:
My opinion, I think we should just announce [not necessarily now and not necessarily going to be removed] `hb_font_set_var_coords_normalized` (thus other normalized coords) as lightly deprecated, why other libraries should know about such internal details of a font, this will be only useful for hb-ft integration and we could do that without public APIs anyway.

src/hb-font.cc
ebraminio: Too much of duplication as it used to call `hb_ot_var_normalize_variations` which does another `hb_ot_var_find_axis_info` and if I just normalize every axis it makes `tests/GPOS-5.tests` fail, can upload on a separate PR.

ebraminio:
> How are we going to communicate, when avar2 is adopted, that design coords are not available?

I think let's deal with that when `avar2` become available, let's slightly deprecate APIs dealing with normalized coords I would say as clients should only care about design coords.

> This is completely lacking in named-instance handling. Might be slightly orthogonal, but I will feel a lot more comfortable if we answer questions in #1883 first.

I think this has more importance than that, named instances are also themselves stored as design coords, not normalized coords, and I think I can reach to that sometime after this.

ebraminio:
Okay, lets merge this and work on ways to adopt #1883 also

Issue #1907, opened 2019-08-14
Should dist be enabled by dfault for all scripts?


khaledhosny:
The Segoe UI Emoji font has a `dist` feature to fix some spacing issues (see https://github.com/khaledhosny/harftex/issues/12#issuecomment-521450494) and seems to expect it to be applied to emoji text.

HarfBuzz seems to enable it only in USE and a selected set of scripts, but Core Text seems to apply it to the emoji text is well, can’t test Uniscribe/DirectWrite at the moment.

Is `dist` now enabled for all scripts or does emoji gets shaped with USE now?

dscorbett:
See also MicrosoftDocs/typography-issues#106.

behdad:
I'm fine enabling it

behdad:
But really. Segoe UI Emoji, why?

khaledhosny:
> But really. Segoe UI Emoji, why?

My guess would be because Word disables `kern` by default (I’d have used `mark` though, since mark anchors are already involved).

behdad:
Right.. so using dist as required kerning basically. Let's do.

PR #1908, opened 2019-08-15
[ot-shape] Enable dist feature by default

khaledhosny:
Fixes https://github.com/harfbuzz/harfbuzz/issues/1907

Not sure if I did this right.


dscorbett:
Some specific shapers currently enable 'dist'; they should be changed not to explicitly, but instead inherit from the default setting.

khaledhosny:
> Some specific shapers currently enable 'dist'; they should be changed not to explicitly, but instead inherit from the default setting.

Right, but I’m not sure what is the way to do this, just drop `dist` from their feature lists?

dscorbett:
I guess so: USE etc. all enable 'mark' too, and that isn’t explicit in each one’s feature list.

khaledhosny:
Waiting for Behdad, I’m pretty much stabbing in the dark here.

behdad:
1. I think it should go into `horizontal_features`.  The `F_GLOBAL` is correct.

2. Remove from indic, USE, khmer, myanmar.

I wonder if we should also apply `abvm`/`blwm` unconditionally.

khaledhosny:
>     1. I think it should go into `horizontal_features`.  The `F_GLOBAL` is correct.
> 
>     2. Remove from indic, USE, khmer, myanmar.

I just did so.

> I wonder if we should also apply `abvm`/`blwm` unconditionally.

Good question. Core Text seems to apply them to Latin text, but Uniscribe doesn’t.

behdad:
> > I wonder if we should also apply `abvm`/`blwm` unconditionally.
> 
> Good question. Core Text seems to apply them to Latin text, but Uniscribe doesn’t.

I'd say we do it.  Feel free to merge this and do that later though.

I think dist goes into horizontal-features, whereas abvm/blwm go in common-features.

PR #1909, opened 2019-08-15
[ot-shape] Enable abvm/blwm features by default

khaledhosny:
Core Text seems to apply them to Latin text, but Uniscribe doesn’t.

See https://github.com/harfbuzz/harfbuzz/pull/1908#issuecomment-521819343


behdad:
Lgtm. I think the list was kept sorted.

khaledhosny:
> I think the list was kept sorted.

I missed that, both arrays are sorted now.

Issue #566, opened 2017-10-12
Speedup using vectorization


behdad:
It feels like we've hit the limit of making HarfBuzz faster using traditional methods.  Next round of optimizations will be by using generic or architecture-specific SIMD vectorization intrinsics.

I have two ideas that are ripe with potential; one affects heavy fonts, the other light ones.  So, combined, they can probably provide a good speedup all around.

1. For heavy fonts, ie. those having dozens, or over a hundred, lookups (Amiri, IranNastaliq, heaving Indic fonts, etc), the bottleneck is just looping over lookups and for each lookup over the buffer contents and seeing if current lookup applies to current glyph; normally that would be a binary search in the Coverage table.  I made this much faster a few years ago by adding these `hb_set_digest_t` things, which are narrow (3*sizeof(long) bytes) structures that implement approximate matching, similar to Bloom Filters or Quotient Filters. These digests do all bitwise operations, so they can be easily vectorized.  Combined with a gather operation, or just multiple fetches in a row (which should parallelize), this will allow us to skip over 4 or 8 glyphs at a time.

2. For fast fonts, like simple Latin fonts, like Roboto, the majority of time is spent in binary searching in the Coverage table of kern and liga lookups.  For this, any way to speed up the binary search using vectorization will be helpful.  Or possibly even just parallelism, not vectorization.  For example, implementing a quaternary search instead of binary, that fetches the three mid values first and does the comparisons after might make it much faster if accessing memory is the bottleneck.

behdad:
@drott

behdad:
> For heavy fonts, ie. those having dozens, or over a hundred, lookups (Amiri, IranNastaliq, heaving Indic fonts, etc), the bottleneck is just looping over lookups and for each lookup over the buffer contents and seeing if current lookup applies to current glyph; normally that would be a binary search in the Coverage table. I made this much faster a few years ago by adding these hb_set_digest_t things, which are narrow (3*sizeof(long) bytes) structures that implement approximate matching, similar to Bloom Filters or Quotient Filters. These digests do all bitwise operations, so they can be easily vectorized. Combined with a gather operation, or just multiple fetches in a row (which should parallelize), this will allow us to skip over 4 or 8 glyphs at a time.

This happens in `apply_forward()`, and the function to be vectorized is `may_have()`.

behdad:
[...]

PR #1911, opened 2019-08-18
Add initial shaping performance test and profiler

ebraminio:
My personal code related to #566 inspired from https://github.com/harfbuzz/harfbuzz/pull/1398#discussion_r235566351 and https://youtu.be/nXaxk27zwlk


ebraminio:
image

ebraminio:
Arabic text with Amiri
image

behdad:
Can we reuse hb-shape?  It has all parameters for it.  This is how I use it:
```
$ time perf stat ./hb-shape -O '' -o /dev/null Roboto-Regular.ttf --text-file thelittleprince.txt  --num-iterations=100 --font-funcs=ot
```

The only reason to write C code for perf is to integrate with a perf library, which I like us to do.  I remember Google had a library for this.  Looks like https://github.com/gperftools/gperftools ?  Or not?

behdad:
Oh, maybe https://github.com/google/benchmark

behdad:
I also have had good luck w sysprof recently.

behdad:
As for test matrix:

* Long paragraphs.  For Latin thelittleprince is good.  For Persian, in Pango we have a looong single-paragraph monologue.  Use that.  I have permission for it.  https://github.com/GNOME/pango/blob/master/utils/test-long-paragraph.txt

* Words.  To emulate systems that do word-caching.  The Wikipedia word-list for various languages from https://github.com/harfbuzz/harfbuzz-testing-wikipedia or for ASCII, `/usr/share/dict/words`.

As for fonts, we care about testing at least:

* Roboto.
* Noto Nastaliq Urdu.
* Amiri.
* Noto Sans Devanagari.

ebraminio:
> Can we reuse hb-shape?

> Oh, maybe https://github.com/google/benchmark

You should want either this or that and the current is also neither of them :) (wanted to go for bench, first am experiencing if can I for simd first)

behdad:
Yeah I want to try SIMD as well if I find a free afternoon... Just timing hb-shape with the right parameters should be fine for that one.

behdad:
Or, not time, but `perf stat`.  On an otherwise-idle laptop, the instruction-count I get from `perf stat` is super deterministic.  Also pay attention to `insn per cycle`.

behdad:
Can't we use hb-shape instead of adding new source for that?  I use:
```
perf stat ./hb-shape -O '' -o /dev/null NotoNastaliq-Regular.ttf --text-file test-long-paragraph.txt  --num-iterations=100  --font-funcs=ot
```

ebraminio:
Then guess all I want is to have one linear command to compile hb-shape, let's close, will post accordingly

ebraminio:
`clang++ -DPACKAGE_NAME='""' -DPACKAGE_VERSION='""' util/hb-shape.cc util/options.cc src/harfbuzz.cc -Isrc -DHAVE_GLIB $(pkg-config --cflags --libs glib-2.0) -DHB_NO_MT -fno-exceptions -fno-rtti -o hb-shape` does the trick

ebraminio:
Actually the rest can be useful, let's update this with the command

ebraminio:
Have a look Behdad

behdad:
Thanks.  We're thinking along the same lines indeed.  I'll comment with my wishlist here later today.

behdad:
Rough notes / suggestions:

- Add toplevel perf/ dir indeed.

- Add test-long-paragraph.txt and thelittheprince.txt, for Arabic and Latin scripts.  Can rename them to something more useful.  Should check how long paragraphs in thelittleprince.txt are.  Maybe also add a wordlist, to emulate word-cache workflows.

- Add Roboto (whatever latest version Android ships), NotoNastaliqUrdu, Amiri, and maybe NotoSansDevanagari.  Find good Devanagari text to add above as well.

- Add `--perf` to `hb-shape` that sets the defaults.  Ie: `./hb-shape -O '' -o /dev/null --num-iterations=100 --font-funcs=ot`

- Add a perf framework ala what we have in our test framework, that implements measuring time elapsed, repeating test, dropping outliers, and summarizing numbers.  This is to be used to write micro-benchmarks, like reporting bsearch vs ksearch performance for various number of input elements.  I really need this test!  Same for measuring set operations performance and possible vectorization speedup there again.

- By default no special optimization flags (unroll loops, etc) should be specified.  We should be measuring 1. defaults (which should have -O2 already), 2. -O2, 3. -O3 with `-march=native -mtune=native` perhaps.

behdad:
We can then define a perf suite that is simply pairs of font-file and text-file.  Later we will try running and collecting those numbers online on variety of machines over time perhaps maybe.

test/perf/.gitignore
```diff
@@ -0,0 +1,3 @@
+hb-shape
```
behdad: We don't keep .gitignore in tree.

ebraminio: Done

ebraminio:
> Add toplevel perf/ dir indeed.

Done

> Add test-long-paragraph.txt and thelittheprince.txt, for Arabic and Latin scripts. Can rename them to something more useful. Should check how long paragraphs in thelittleprince.txt are. Maybe also add a wordlist, to emulate word-cache workflows.

Already done except wordlist, is this https://github.com/sindresorhus/word-list/blob/master/words.txt ok? Is 2mb…

> Add Roboto (whatever latest version Android ships), NotoNastaliqUrdu, Amiri, and maybe NotoSansDevanagari. Find good Devanagari text to add above as well.

Done

> Add --perf to hb-shape that sets the defaults. Ie: ./hb-shape -O '' -o /dev/null --num-iterations=100 --font-funcs=ot
> Add a perf framework ala what we have in our test framework, that implements measuring time elapsed, repeating test, dropping outliers, and summarizing numbers. This is to be used to write micro-benchmarks, like reporting bsearch vs ksearch performance for various number of input elements. I really need this test! Same for measuring set operations performance and possible vectorization speedup there again.

Maybe later? I mean, let's have something to hack upon it over the time, think that should work also.

Current output is:

```
 Performance counter stats for './hb-shape -o /dev/null fonts/NotoNastaliqUrdu-Regular.ttf --text-file text/test-long-paragraph.txt --num-iterations=100 --font-funcs=ot':

          5,003.70 msec task-clock:u              #    1.000 CPUs utilized          
                 0      context-switches:u        #    0.000 K/sec                  
                 0      cpu-migrations:u          #    0.000 K/sec                  
               541      page-faults:u             #    0.108 K/sec                  
    17,819,606,588      cycles:u                  #    3.561 GHz                    
    50,330,220,517      instructions:u            #    2.82  insn per cycle         
     8,486,482,328      branches:u                # 1696.041 M/sec                  
        58,516,636      branch-misses:u           #    0.69% of all branches        

       5.005475650 seconds time elapsed

       4.983649000 seconds user
       0.000000000 seconds sys
````

> By default no special optimization flags (unroll loops, etc) should be specified. We should be measuring 1. defaults (which should have -O2 already), 2. -O2, 3. -O3 with -march=native -mtune=native perhaps.

Done

behdad:
Looks good for starters.  I like to add the `--perf` to hb-shape as well.  Not a huge fan of `run` script.  Prefer using existing build system.  But yeah, happy to merge this.  Specially since it doesn't go into the tarball yet.  Feel free to merge.  We change from there.

Let's rename `test-long-paragraph.txt` to something else as well, reflecting what's inside maybe.

ebraminio:
> Let's rename test-long-paragraph.txt to something else as well, reflecting what's inside maybe.

Done

PR #1915, opened 2019-08-20
Minor, replace hb_bool_t with bool on local variables

ebraminio:
came to my attention while editing hb-directwrite


behdad:
There used to be warning reasons that I made these that way I think.  Check blame.

ebraminio:
ok, no time for this right now, let's forget about this also

Issue #1849, opened 2019-07-21
[uniscribe/gdi] Remove hb-uniscribe.h and move to hb-gdi.h


behdad:
Uniscribe is a shaper.  Now that we have hb-gdi, move hb-uniscribe.h stuff to hb-gdi and rethink / rename as needed.

ebraminio:
Unrelated to this but I like also to put https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-directwrite.h#L33 behind some compile or even remove it now that I can easily access private API using harfbuzz.cc inclusion.

behdad:
> Unrelated to this but I like also to put https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-directwrite.h#L33 behind some compile flag or even remove it now that I can easily access private API using harfbuzz.cc inclusion as I don't like to see single use of it anywhere in msys2/mingw/cygwin etc communities.

That would be nice, yes.

PR #1916, opened 2019-08-20
Remove hb_directwrite_shape_experimental_width public API

ebraminio:
I've specially exposed the API as I didn't know how to embed harfbuzz
easily elsewhere but now with harfbuzz.cc it has become very easy
and I don't like to see its use anywhere as it has a bad naming and
its Kashida adding is bogus and only useful to check where it should
be added not visually useful.

https://github.com/harfbuzz/harfbuzz/issues/1849#issuecomment-515232748 related


behdad:
Thanks.

PR #1922, opened 2019-08-23
[fuzz] Fix oss-fuzz/16639

ebraminio:
Fixes https://crbug.com/oss-fuzz/16639


ebraminio:
I think we had some idiom to avoid manual freeing of blob but I don't remember where it was and whether works with hb_blob_t

behdad:
lgtm.

ebraminio:
PTAL again, disallowed a subset much bigger than the table itself as to msan failure and https://ci.appveyor.com/project/harfbuzz/harfbuzz/builds/26913998/job/lkp2osx2l5fwwkd8#L1563

behdad:
How do a leak fix and some other fix belong to the same PR?

ebraminio:
Both revealed by the same test case. Ok should investigate more

PR #1930, opened 2019-08-25
[subset] Raise the bar in new vs old table size

ebraminio:
https://crbug.com/oss-fuzz/16740

This is actually an interesting thing that {h,v}mtx allocates as
much as a font pretends to have glyphs but the solution is not
that obvious as regular fonts can have less than actually containing
metrics in their {h,v}mtx. This change raises the bar to consider this
hmtx 4 byte for every glyph case.

Initially we wanted to just find things allocating crazy amount of
memory but having the assert has led to interesting findings also
so let's don't remove the assert and see what we can find elsewhere.


ebraminio:
Done

ebraminio:
> Can you explain why glyf subset code was running out of room repeatedly? Ideally we like to detect such issues if we can formulate a better way. Thanks.

You are right that I should've explained it more and I am sorry.

https://github.com/ebraminio/harfbuzz/blob/2023f0da44ba492740676b5fcd0fb19f4d883979/src/hb-ot-glyf-table.hh#L575 wasn't considering instruction size itself and wasn't failing when there was no space for that. However the logic here `source_glyph.length - dest_start.length - instruction_length` meant if `source_glyph.length` was equal to `dest_start.length` and `instruction_length` was anything but zero (by reading memory out of the glyph bound) it was becoming a negative value.

ebraminio:
> Yeah we should rewrite that bound checking in terms of hb_array_t...

That's why I went for rewriting most of it after painfully understanding of what is going on it.

PR #1932, opened 2019-08-27
[glyf] Move GlyphHeader::from_bytes to hb_bytes_t, introduce .as<T> ()

ebraminio:


src/hb-array.hh
```diff
@@ -181,6 +181,10 @@ struct hb_array_t : hb_iter_with_fallback_t<hb_array_t<Type>, Type&>
   hb_array_t<Type> sub_array (unsigned int start_offset, unsigned int seg_count) const
   { return sub_array (start_offset, &seg_count); }
 
+  template <typename T>
```
ebraminio: Ops, wait, I should add that hb_enable_if

ebraminio:
Can't turn as<T> to return reference because of UBSan https://circleci.com/gh/harfbuzz/harfbuzz/107646

PR #1934, opened 2019-08-27
[fuzz] Add oss-fuzz/16810 testcase

ebraminio:
Fixes https://crbug.com/oss-fuzz/16810

PR #1947, opened 2019-08-30
[subset] Move drop 0 tag table logic

ebraminio:
If I've got https://github.com/harfbuzz/harfbuzz/pull/1934#issuecomment-526616048 correctly.


behdad:
No no no no. :)
Put it inside `hb_face_reference_table`.

ebraminio:
So I don't understand, explain more please :)

PR #1949, opened 2019-09-01
Form an internal number parser

ebraminio:
CFF parsing needs a 4bit digits BCD parser anyway so I've extracted its float parser from it and formed a common parser for the project. As a side effect we won't need to use strtod_l, get_C_locale anymore which is good for the portability and consistency of the library also.


ebraminio:
PTAL

behdad:
What's the size implication of this?  I'm not a fan of removing library calls.

Name the header `hb-float.hh`.  Also, why a type, instead of functions?  What's the internal API you are adding?  The type is messy as is.

ebraminio:
Thanks for the feedback. Wish can provide a better version :)

> I'm not a fan of removing library calls.

Understandable

> Name the header hb-float.hh.

Wanted to go for other number parsers (strtol, etc) also later 😂 but sure the main justification I have here for this is unreliability of system libraries and locale dependency of the strtod and all the hassle we go for having access to `strtod_l` and C locale, other system calls are not like this. And the fact we have this parser already in our code.

> Also, why a type, instead of functions?

I think the best would be writing it in Ragel. The advantage of current implementation is having the parser to work in place with `memcpy`.

> What's the internal API you are adding?

```
float_parser_t parser;
parser.consume ();
parser.is_in_error ();
parser.end ();
```

Guess the naming here isn't that great. I probably should go with Ragel but guess can't make it in place anymore.

> The type is messy as is.

You are right.

> What's the size implication of this? 

Less than 1kb, without considering removal of all these https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-common.cc#L770-L825 

-Oz
```
number-parser:
$ size harfbuzz.o
   text	   data	    bss	    dec	    hex	filename
 665373	   2880	    464	 668717	  a342d	harfbuzz.o

master:
$ size harfbuzz.o
   text	   data	    bss	    dec	    hex	filename
 664970	   2880	    464	 668314	  a329a	harfbuzz.o
```

-O3:
```
number-parser:
$ clang -c src/harfbuzz.cc -O3
$ size harfbuzz.o
   text	   data	    bss	    dec	    hex	filename
 926352	   2864	    456	 929672	  e2f88	harfbuzz.o

master:
$ size harfbuzz.o
   text	   data	    bss	    dec	    hex	filename
 925500	   2864	    456	 928820	  e2c34	harfbuzz.o
```

behdad:
We still need to access locale to get default language.  So that's not fully going away...

behdad:
I suggest u provide this if the existing code cannot find an exisitng solution.

Also, I suggest the CFF code build a string, and just call strtod().

ebraminio:
> We still need to access locale to get default language. So that's not fully going away...

No other use for get_C_locale and xlocale and , please check it.

> I suggest u provide this if the existing code cannot find an exisitng solution.

That is so broad and philosophical. I mean, this is done for some reason.

> Also, I suggest the CFF code build a string, and just call strtod().

I don't trust this locale setting for different systems. Prefer to not do anything than making internal functionality of the library to availability of strtod_l.

ebraminio:
So let's close and maybe think about it later.

PR #1950, opened 2019-09-01
[subset] Run once for a tag

ebraminio:
As https://github.com/harfbuzz/harfbuzz/pull/1934#issuecomment-526927171


behdad:
Thanks. Since subsetter had access to internals, you can just allocate set on stack and let constructor/destructor work for you.

behdad:
Also `set->has`.

ebraminio:
Makes sense, #1951

PR #1953, opened 2019-09-02
Fallback if roundf didn't exist, like in dietlibc

### LINKED ISSUE #1898: polyfill roundf for dietlibc
dietlibc doesn't have `roundf` so we should probably do something like bringing back https://github.com/harfbuzz/harfbuzz/commit/c4cae81a26a816979f3206418c47856b5ed2d8bb for `roundf` this time.

ebraminio:
Closes #1898, c4cae81 revert but for roundf, if you don't like to this let's close both, we should notify HarfBuzz embedders like Chrome to add HAVE_ROUNDF in their build scripts if we like to go this way.


ebraminio:
Thanks. To be clear this wasn't intended for harfbuzzjs's libc as that was using roundf even without `ceil` and `floor` and maybe going to continue that but for the sake of doing something correct for known 'libc's, as @fanc999 once wanted it for older Visual Studio versions.

@drott: Please add a HAVE_ROUNDF in the next check-in, thanks.

PR #1955, opened 2019-09-02
Aggregate and merge number parsing codes

ebraminio:


behdad:
I'm slightly confused as to why there's no `uint` version in the internal API.  But whatever works I'm fine.  Feel free to merge.

ebraminio:
> I'm slightly confused as to why there's no uint version in the internal API. But whatever works I'm fine. Feel free to merge.

Added now, wasn't there because parse_uint and parse_uint32 of in hb-common.cc were using `strtol` so -1 could mean a big number, as the current comments indicates but the one in hb_codepoint_parse needed that have added it, have used template and lambda to remove duplicated parsing code also, will merge once wrote test for these and am sure there is no behavior change if you don't object. Thanks for the support :)

behdad:
I like the end result.  Thank you. :)

behdad:
`hb_codepoint_parse` can move into `hb-number.hh` maybe as well.  No strong opinion.

behdad:
Don't worry.  Leave as is.  Thanks for working on this.

behdad:
Don't worry.  Leave as is.  Thanks for working on this.

ebraminio:
> I like the end result. Thank you. :)

Thanks for expressing the feeling :)

PR #1959, opened 2019-09-06
Replace strtod with a ragel implementation

ebraminio:
Update: Ended up to use the implementation only when strtod_l wasn't there.

----

This probably never will end up in harfbuzz repository itself but I wanted to develop it first here, test it and then move it to harfbuzzjs repository, but though having it here also isn't that bad, let's list pros and cons, at least from what I can see,

Pros:
* Having clear semantic of what `hb_parse_double` does, the use on [`parse_variation_value`](https://github.com/harfbuzz/harfbuzz/blob/1083df8b80b08aa1a4f2dabfe414aaa4a0ec8aa1/src/hb-common.cc#L990) apparently accepts space after the '=' (as `strtod` guarantees) but not before the `=` sign, I don't think that is intended.
* No locale dependency of the library for font parsing matters.
* Removal of dependency to `xlocale`, maybe there are systems that `strtod_l` doesn't exist but the main strtod is still locale dependent, maybe not.
* No need to play with errno around for this, actually if we can remove this and also other `strtoX`s, there will be only one other `errno` use in the library. FYI glibc's strtoul, thus our hb_parse_uint, also are accepting strings having +/- in start AFAICS (but maybe false claim? please recheck), something not expected on the uses, serializer gid/cluster and `hb_codepoint_parse`.
* Being able to do the conversion in place (not the current version, as I want to move the current version to harfbuzzjs's libc so it should be strtod compatible), reducing one `strncpy` call that `_parse_number` has and a `strlen` that a `strtod` may have internally.

Cons:
* Having to maintain one more ragel state machine, which means new code, new bugs and new headaches.
* Additional size (but maybe we can have this only when we couldn't find a `strtod_l`, maybe not)

With all these (which maybe are false/inaccurate claims) however I understand if this isn't acceptable for the project. So bear with me! :)

Thanks!

### DIFF
--- a/src/hb-number-parser.rl
+++ b/src/hb-number-parser.rl
@@ -1,0 +1,122 @@
+/*
+ * Copyright © 2019  Ebrahim Byagowi
+ *
+ *  This is part of HarfBuzz, a text shaping library.
+ *
+ * Permission is hereby granted, without written agreement and without
+ * license or royalty fees, to use, copy, modify, and distribute this
+ * software and its documentation for any purpose, provided that the
+ * above copyright notice and the following two paragraphs appear in
+ * all copies of this software.
+ *
+ * IN NO EVENT SHALL THE COPYRIGHT HOLDER BE LIABLE TO ANY PARTY FOR
+ * DIRECT, INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES
+ * ARISING OUT OF THE USE OF THIS SOFTWARE AND ITS DOCUMENTATION, EVEN
+ * IF THE COPYRIGHT HOLDER HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ *
+ * THE COPYRIGHT HOLDER SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ * FITNESS FOR A PARTICULAR PURPOSE.  THE SOFTWARE PROVIDED HEREUNDER IS
+ * ON AN "AS IS" BASIS, AND THE COPYRIGHT HOLDER HAS NO OBLIGATION TO
+ * PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
+ *
+ */
+
+#ifndef HB_NUMBER_PARSER_HH
+#define HB_NUMBER_PARSER_HH
+
+#include "hb.hh"
+
+%%{
+
+machine double_parser;
+alphtype unsigned char;
+write data;
+
+action see_neg { neg = true; }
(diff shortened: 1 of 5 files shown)


src/hb-number-parser.rl
```diff
@@ -0,0 +1,130 @@
...
+		(('+'|'-'@see_neg)? '.' num @add_mantissa)
+	)
+	(('e'|'E') (('+'|'-'@see_exp_neg)? num @add_exp))?
+);
+
+}%%
+
+constexpr double
+pow2 (double x)
+{
```
behdad: Make function single-line?

src/hb-number-parser.rl
```diff
@@ -0,0 +1,130 @@
...
+  return x * x;
+}
+
+constexpr double
+pow10_of_2i (unsigned int n)
+{
+  return n == 1 ? 10. : pow2 (pow10_of_2i (n >> 1));
+}
```
behdad: Looks wrong to me.

behdad: `return (n & 1 ? 10 : 1) * pow2 (...)`

src/hb-number-parser.rl
```diff
@@ -0,0 +1,130 @@
...
+  pow10_of_2i (0x100),
+  pow10_of_2i (0x80),
+  pow10_of_2i (0x40),
+  pow10_of_2i (0x20),
+  pow10_of_2i (0x10),
+  pow10_of_2i (0x8),
+  pow10_of_2i (0x4),
+  pow10_of_2i (0x2),
+  pow10_of_2i (0x1),
+};
```
behdad: Inverting the order here makes more sense to me.

behdad:
I'm fine using this if our existing codepaths don't find a useful locale-independent strtod, or if `HB_NO_LIBC` is defined.

behdad:
Needs testing please though.

src/hb-number-parser.rl
ebraminio: Done

src/hb-number-parser.rl
ebraminio: Hmm, couldn't make it work.

src/hb-number-parser.rl
ebraminio: The advantage is I can write "for (const double *power = powers_of_10; '''mask'''; ++power, mask >>= 1)"

ebraminio:
> I'm fine using this if our existing codepaths don't find a useful locale-independent strtod, or if HB_NO_LIBC is defined.

Nice. Thanks :)

> Needs testing please though.

Already done with test-number :)

src/hb-number-parser.rl
behdad: You could still do that, just needed to reverse the order.

ebraminio: Oh, ok, let me see

src/hb-number-parser.rl
behdad: THIS IS WRONG.....  How should I say?  You are wrong, the code is wrong, the output is wrong.

Don't merge work under review.

PR #1961, opened 2019-09-07
Facilitate errno uses and restore the value after the use

ebraminio:
As https://stackoverflow.com/a/15242186 thought maybe is something nice to have, or maybe not.

### DIFF
--- a/src/hb-blob.cc
+++ b/src/hb-blob.cc
@@ -48,7 +48,6 @@
 #endif /* HAVE_SYS_MMAN_H */
 
 #include <stdio.h>
-#include <errno.h>
 #include <stdlib.h>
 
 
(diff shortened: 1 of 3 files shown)


behdad:
Not particularly interested in this.

Rename to `hb_errno_t` or something if we decide to keep.

But fundamentally, I don't see what problem you are trying to solve.  If you want to guarantee HarfBuzz not clobber errno, your patch doesn't do that.  You are only guarding all the places where we access errno.  I'm sure we use libc calls that modify errno but we don't check (because we don't need to).

I'm leaning towards closing this.

ebraminio:
> You are only guarding all the places where we access errno. I'm sure we use libc calls that modify errno but we don't check (because we don't need to).

Oh, I see. Thanks

PR #1962, opened 2019-09-07
[subset] Handle blob destroy using blob stack allocated wrapper

ebraminio:
Maybe we should have something like this in a wrapper or hb_blob_t itself.

### DIFF
--- a/src/hb-blob.hh
+++ b/src/hb-blob.hh
@@ -69,6 +69,15 @@ struct hb_blob_t
   hb_destroy_func_t destroy;
 };
 
+struct hb_blob_reference_t
+{
+  hb_blob_t *blob;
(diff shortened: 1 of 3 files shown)


behdad:
First, great idea.  Two things:

1. Need a better name to clarify that it consumes the reference instead of grabbing one.  I'll think about it.

2. This isn't blob-specific. Make it call obj->destroy on whatever object. We had hb_auto_t<> before for calling init/fini on arbitrary structured.

behdad:
I think we can close this.

When https://github.com/harfbuzz/harfbuzz/issues/2152 happens we can use `hb::shared_ptr<hb_blob_t>` and others internally.  We can also store those in our internal C++ structs to get automatic destruction.

ebraminio:
Oh sure, thanks

PR #1973, opened 2019-09-16
Extract an avoid errno compile flag

ebraminio:
The fact this was already in use for old versions of _WIN32_WCE made me more confident that this can reside in main repo itself, useful in harfbuzz.js but more needed in the brand new harfbuzz-kernel.


behdad:
lgtm.

PR #1974, opened 2019-09-16
Avoid stdint.h in Kernel module compiles

ebraminio:
I don't expect you like this :) Other than #1973 it is the only needed change, yet is more symbolic! (otherwise makes more sense to keep the patch locally, but is the only really specific needed thing) Both probably can be avoided if I avoid whole glibc headers, which I found can be avoided.


behdad:
This lgtm actually.  Do what you feel is right though.

PR #1980, opened 2019-09-19
[subset] GPOS Lookup Type 2: PairPos

qxliu76:


src/hb-ot-layout-gpos-table.hh
```diff
@@ -1011,8 +1089,49 @@ struct PairPosFormat2
...
+                  | hb_filter (klass2_set)
+                  | hb_apply ([&] (const unsigned class2_idx)
+                              {
+                                unsigned idx = (class1_idx * (unsigned) class2Count + class2_idx) * record_len;
+                                for (unsigned i = 0; i < record_len; i++)
+                                  c->serializer->copy (values[idx+i]);
+                              })
+                  ;
+                })
+    ;
```
behdad: Let's look into this together tomorrow.

PR #1984, opened 2019-09-21
Turn more of simple dagger chains to foreach

ebraminio:
Less noise, as was agreed before and applied 385741d565de also

### DIFF
--- a/src/hb-ot-cff-common.hh
+++ b/src/hb-ot-cff-common.hh
@@ -184,7 +184,8 @@ struct CFFIndex
     else
     {
       serialize_header(c, + it | hb_map ([] (const byte_str_t &_) { return _.length; }));
-      + it | hb_apply ([&] (const byte_str_t &_) { _.copy (c); });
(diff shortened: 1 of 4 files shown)


behdad:
Would help if you say what problem you face when doing that.

ebraminio:
> Would help if you say what problem you face when doing that.

Uploaded it as a PR just to show CI failures, https://travis-ci.org/harfbuzz/harfbuzz/jobs/587812319#L2265 for example

behdad:
> Uploaded it as a PR just to show CI failures, https://travis-ci.org/harfbuzz/harfbuzz/jobs/587812319#L2265 for example

Right.  Not helpful...

Feel free to assign to @qxliu76 maybe.

ebraminio:
> Right. Not helpful...

I was thinking like 0u/0 thing something else, and, not with the specific code here has went wrong that something with apparently same semantic is resulting to weird difference.

PR #1985, opened 2019-09-21
Minor, turn more of hb_apply to for each

ebraminio:


ebraminio:
Hmm, C++11 enable much more compiler checks, like second change of this PR and #1986, which is good I believe.

src/hb-ot-glyf-table.hh
```diff
@@ -144,9 +144,8 @@ struct glyf
   {
     TRACE_SERIALIZE (this);
 
-    + it
-    | hb_apply ([=] (const SubsetGlyph& _) { _.serialize (c, plan); })
-    ;
+    for (const SubsetGlyph& _ : it)
```
behdad: Can use auto type as well.

src/hb-ot-layout-gpos-table.hh
```diff
@@ -675,22 +669,14 @@ struct SinglePos
...
-		  ;
-		})
-    ;
+    for (hb_array_t<const Value> val_iter : + glyph_val_iter_pairs
+					    | hb_map (hb_second))
+      for (const hb_pair_t<Value, Value> _ : hb_zip (val_iter, first_val_iter))
+	if (_.first != _.second)
+	{
+	  subset_format = 2;
+	  continue;
```
behdad: Why continue?  Should just return immediately.

src/hb-ot-layout-gpos-table.hh
```diff
@@ -601,15 +600,10 @@ struct SinglePosFormat2
...
-		{
-		  + val_iter
-		  | hb_apply ([=] (const Value& _) { c->copy (_); })
-		  ;
-		})
-    ;
+    for (hb_array_t<const Value> val_iter : + it
+					    | hb_map (hb_second))
+      for (const Value& _ : val_iter)
+	c->copy (_);
```
behdad: This part should be improved (I know old code did as is).  I wonder if `c->copy (val_iter)` would work.  We should make it work.

src/hb-ot-vorg-table.hh
```diff
@@ -84,9 +84,8 @@ struct VORG
     this->defaultVertOriginY = defaultVertOriginY;
     this->vertYOrigins.len = it.len ();
 
-    + it
-    | hb_apply ([c] (const VertOriginMetric& _) { c->copy (_); })
-    ;
+    for (const VertOriginMetric _ : it)
```
behdad: Missing reference `&`.  Again, auto& maybe.  In all above ones as well.

behdad: Or `const auto &` if that's a thing.

behdad: Okay I see why you not referencing here.  It sucks that we cannot just use `const auto &` and let compiler handle.  Though, adding the const it might actually work.

src/hb-ot-layout-gpos-table.hh
ebraminio: Done later

src/hb-ot-vorg-table.hh
ebraminio: Compiler warnings related to it are brought https://github.com/harfbuzz/harfbuzz/pull/1985/commits/aa9960971a54bc1ae400d614bc7c63ec53cb8511 don't have the version with the flag to check it locally, it doesn't seem to.

src/hb-ot-layout-gpos-table.hh
ebraminio: Correct. Am looking into it.

src/hb-ot-vorg-table.hh
ebraminio: Happens with 'const auto &' also, https://circleci.com/gh/harfbuzz/harfbuzz/111243

behdad: Okay.  But at least change most / all to `const auto` or `const auto &`?

behdad: I can't get my mind around it, but what about `const auto &&`?

src/hb-ot-glyf-table.hh
ebraminio: Done cbc0646a6cfc0f6a04804d6888636f12130cd51b

src/hb-ot-vorg-table.hh
ebraminio: Done cbc0646a6cfc0f6a04804d6888636f12130cd51b

src/hb-ot-layout-gpos-table.hh
ebraminio: #1987

src/hb-ot-vorg-table.hh
ebraminio: And reverted, clang didn't like that also :/

PR #1989, opened 2019-09-24
[subset] GPOS Lookup Type 3: CursivePos

qxliu76:


src/hb-ot-layout-gpos-table.hh
```diff
@@ -1064,6 +1098,19 @@ struct EntryExitRecord
     return_trace (entryAnchor.sanitize (c, base) && exitAnchor.sanitize (c, base));
   }
 
+  EntryExitRecord* copy (hb_serialize_context_t *c,
+                         const void *src_base,
+                         const void *dst_base) const
```
ebraminio: Your work will become prefect if you replace initial 8 spaces with tab :)

src/hb-ot-layout-common.hh
```diff
@@ -2070,6 +2072,12 @@ struct HintingDevice
     return_trace (c->check_struct (this) && c->check_range (this, this->get_size ()));
   }
 
+  HintingDevice* copy (hb_serialize_context_t *c) const
+  {
+    TRACE_SERIALIZE (this);
+    return_trace (c->embed<HintingDevice> (this));
```
behdad: Do you need to specify the type?  It should be deduces from `this` automatically, no?  I think that's how we do in other places.

src/hb-ot-layout-common.hh
```diff
@@ -2131,6 +2139,12 @@ struct VariationDevice
   hb_position_t get_y_delta (hb_font_t *font, const VariationStore &store) const
   { return font->em_scalef_y (get_delta (font, store)); }
 
+  VariationDevice* copy (hb_serialize_context_t *c) const
+  {
+    TRACE_SERIALIZE (this);
+    return_trace (c->embed<VariationDevice> (this));
```
behdad: Same.  And other places.

src/hb-ot-layout-gpos-table.hh
```diff
@@ -1191,11 +1238,47 @@ struct CursivePosFormat1
...
+    const hb_map_t &glyph_map = *c->plan->glyph_map;
+
+    auto *out = c->serializer->start_embed (*this);
+    if (unlikely (!out)) return_trace (false);
+
+    auto it =
+    + hb_zip (this+coverage, entryExitRecord)
+    | hb_filter (glyphset, hb_first)
+    | hb_map_retains_sorting ([&] (hb_pair_t<hb_codepoint_t, const EntryExitRecord&> p) -> hb_pair_t<hb_codepoint_t, const EntryExitRecord&>
+                              { return hb_pair (glyph_map[p.first], p.second);})
```
behdad: I like to add something that makes it easier to map only first or second component of a pair without writing a full lambda...

PR #1999, opened 2019-10-01
Implement gvar table and variable glyf

ebraminio:
Splitted from Ariza's "[subset] TrueType/CFF2 variable font" work on #1594


ebraminio:
Now it passes all of its tests with the hvar/vvar parts that just merged so if I can have this in I can work toward cleaning it up also, possibly after the merge so I can merge Ariza branch with the it.

src/hb-ot-hmtx-table.cc
```diff
@@ -0,0 +1,54 @@
...
+
+#include "hb-ot-hmtx-table.hh"
+#include "hb-ot-glyf-table.hh"
+
+namespace OT {
+
+int hmtxvmtx_accelerator_base_t::get_side_bearing_var_tt (hb_font_t *font, hb_codepoint_t glyph, bool vertical)
+{
+  glyf::accelerator_t glyf_accel;
+  glyf_accel.init (font->face);
```
ebraminio: This probably is suboptimal, we should see what we can do about it I guess

ebraminio: Done

behdad:
What do these do:
```
HB_INTERNAL int
hb_ot_get_side_bearing_var_tt (hb_font_t *font, hb_codepoint_t glyph, bool is_vertical);

HB_INTERNAL unsigned
hb_ot_get_advance_var_tt (hb_font_t *font, hb_codepoint_t glyph, bool is_vertical);
```
what's the tt for, etc?  I also don't like the blob checks against null...

ebraminio:
> what's the tt for, etc?

loading things for truetype/gvar variable fonts without hvar/vvar, renamed them.

> I also don't like the blob checks against null...

Correct. Fixed

PR #2001, opened 2019-10-02
Limit the number of Scripts and LangSys' that can be processed during subsetting.

garretrieger:
To fix fuzzer reported time outs.


behdad:
Regarding rewriting ==/hash on array... I'm really concerned that rewriting as loop seems to be 10x faster.  I like to understand why.  The iter-based solution should come expand down to same loop.  Any idea how to investigate?

garretrieger:
> Regarding rewriting ==/hash on array... I'm really concerned that rewriting as loop seems to be 10x faster. I like to understand why. The iter-based solution should come expand down to same loop. Any idea how to investigate?

I can send you a profile of the before/after the change that may help illuminate the differences. Also the speedup was measured under the compilation settings for fuzzing so it's possible that the fuzzer asan tracking stuff causes the iter approach to be much slower then the for loop.

garretrieger:
So when I compile without the fuzzing options both versions seem to run in about the same time.

behdad:
> So when I compile without the fuzzing options both versions seem to run in about the same time.

That sounds better. :)

What I do like to see optimized still is hash/== for byte-arrays.  That's what the TODO is about and should help immensely.

PR #2003, opened 2019-10-02
[subset] GSUB Lookup Type 6: ChainContextSubst

qxliu76:
tests are added.


src/hb-ot-layout-gsubgpos.hh
```diff
@@ -2083,6 +2154,39 @@ struct ChainRuleSet
...
+    if (unlikely (!c->serializer->extend_min (out))) return_trace (false);
+
+    for (const OffsetTo<ChainRule>& _ : rule)
+    {
+      if (!_) continue;
+      auto *o = out->rule.serialize_append (c->serializer);
+      if (unlikely (!o)) continue;
+      
+      auto o_snap = c->serializer->snapshot ();
+      bool ret = o->serialize_subset (c, _, this, out, backtrack_klass_map, input_klass_map, lookahead_klass_map);
```
garretrieger: You can merge this line with the if below.

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -2002,6 +2002,77 @@ struct ChainRule
...
   }
 
+  void serialize (hb_serialize_context_t *c,
+                  const hb_map_t *backtrack_map,
+                  const hb_map_t *input_map = nullptr,
+                  const hb_map_t *lookahead_map = nullptr) const
+  {
+    const hb_map_t *mapping = backtrack_map;
+    if (unlikely (!c->allocate_size<ArrayOf<HBUINT16>> (backtrack.get_size ()))) return;
+    c->copy (backtrack.len);
```
garretrieger: Make a helper method for this block of code, since it's identical to the for each of backtrack, input, and lookahead.

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -2002,6 +2002,77 @@ struct ChainRule
...
+			    | hb_map (mapping))
+    {
+      HBUINT16 gid;
+      gid = g;
+      c->copy (gid);
+    }
+
+    if (unlikely (!c->allocate_size<HeadlessArrayOf<HBUINT16>> (inputX.get_size ()))) return;
+    c->copy (inputX.lenP1);
+    if (input_map) mapping = input_map;
```
garretrieger: Is it intentional that backtrack mapping is used if input_map is not specified?

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -2002,6 +2002,77 @@ struct ChainRule
...
+                  const hb_map_t *lookahead_map = nullptr) const
+  {
+    const hb_map_t *mapping = backtrack_map;
+    if (unlikely (!c->allocate_size<ArrayOf<HBUINT16>> (backtrack.get_size ()))) return;
+    c->copy (backtrack.len);
+    for (const unsigned g : + backtrack.iter ()
+			    | hb_map (mapping))
+    {
+      HBUINT16 gid;
+      gid = g;
```
garretrieger: can you do the assignment on the line above?

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -2002,6 +2002,77 @@ struct ChainRule
 					      lookup.arrayZ, lookup_context));
   }
 
+  void serialize (hb_serialize_context_t *c,
```
garretrieger: This isn't really a serialize method since it's being called on the source object, recommend renaming to copy() or copy_and_remap()

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -2002,6 +2002,77 @@ struct ChainRule
...
+                  const hb_map_t *backtrack_map,
+                  const hb_map_t *input_map = nullptr,
+                  const hb_map_t *lookahead_map = nullptr) const
+  {
```
garretrieger: Looking at other methods in this class, looks like you need something like this:

```
const HeadlessArrayOf<HBUINT16> &input = StructAfter<HeadlessArrayOf<HBUINT16>> (backtrack);
const ArrayOf<HBUINT16> &lookahead = StructAfter<ArrayOf<HBUINT16>> (input);
const ArrayOf<LookupRecord> &lookup = StructAfter<ArrayOf<LookupRecord>> (lookahead);
```

Too access the array members, since the members on this struct don't actually point to the real locations.

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -2457,11 +2630,52 @@ struct ChainContextFormat3
...
+
+    return_trace (out->len);
+  }
+
   bool subset (hb_subset_context_t *c) const
   {
     TRACE_SUBSET (this);
-    // TODO(subset)
-    return_trace (false);
+
```
garretrieger: You'll need to use the StructAfter mechanic here as well.

src/hb-ot-layout-common.hh
```diff
@@ -1073,6 +1135,17 @@ struct Coverage
     }
   }
 
+  bool subset (hb_subset_context_t *c) const
+  {
```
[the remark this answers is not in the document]
qxliu76: after my other change that serialize () only takes an iterator of pairs (glyph id, class value), now it's impossible to define subset() in this way.

src/hb-ot-layout-gsubgpos.hh
qxliu76: no. compiler will complaining about: conversion from ‘const unsigned int’ to non-scalar type ‘OT::HBUINT16’

src/hb-ot-layout-gsubgpos.hh
qxliu76: yes.
format2 has different mappings for backtrack, input and lookahead arrays: 
       old class -> new class 
format1, mappings are the same for all arrays: 
        old glyph id -> new glyph id,

src/hb-ot-layout-common.hh
```diff
@@ -1210,53 +1302,70 @@ struct ClassDefFormat1
...
+      if (!klass) continue;
+
+      glyphs.push (glyph_map[gid]);
+      gid_org_klass_map.set (glyph_map[gid], klass);
+      orig_klasses.push (klass);
     }
-    c->serializer->propagate_error (glyphs, klasses);
-    ClassDef_serialize (c->serializer, glyphs, klasses);
+
+    bool has_no_match = glyphset.get_population () > gid_org_klass_map.get_population ();
```
garretrieger: This line and onwards is the same for both ClassDefFormat1 and Format2. Can you pull it out into a helper method to avoid duplicating the code.

src/hb-ot-layout-gsubgpos.hh
garretrieger: sounds good.

src/hb-ot-layout-common.hh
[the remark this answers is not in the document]
qxliu76: Oh, yeah you're right. Somehow I thought you were talking about "struct ClassDef". :)

behdad:
I'm trying to review this.  For one thing, could be committed into multiple smaller commits.  Eg. one adding ClassDef_serialize_and_unmap, etc.

PR #2006, opened 2019-10-03
Implement hmtx/vmtx metrics read from hvar/vvar

ebraminio:
First step toward #1594 per @blueshade7 and needed for #1999 (so going to be tested there)

### DIFF
--- a/src/hb-ot-hmtx-table.hh
+++ b/src/hb-ot-hmtx-table.hh
@@ -172,7 +172,6 @@ struct hmtxvmtx
     void init (hb_face_t *face,
 	       unsigned int default_advance_ = 0)
     {
-      memset (this, 0, sizeof (*this));
       default_advance = default_advance_ ? default_advance_ : hb_face_get_upem (face);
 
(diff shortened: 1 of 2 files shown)


behdad:
I'm also not sure what the purpose of `hmtxvmtx_accelerator_base_t` is.

blueshade7:
If my memory serves me right, it had something to do with a cyclic reference between `glyf::accelerator_t` and `hmtxvmtx::accelerator_t`. Since both can’t own each other, I ended up with the former owning the latter, the latter temporary using the former through hb-ot-hmtx-table.cc. To simplify this code, I created this base class.

ebraminio:
No problem I turned it into a private API afterward.

behdad:
> No problem I turned it into a private API afterward.

In that case, maybe we can remove the base class?  Or use templates to break down the recursive use.

ebraminio:
I tried to keep the original change untouched and had removed that base
class and some other refactoring after it on the same PR or next. There is
a recursive use between hmtx and glyf headers which if it could be resolved
I didn't need that private API also but guess that is fine.

PR #2008, opened 2019-10-06
[ot-font] Apply font scaling before turning to int

### LINKED ISSUE #1940: What is hb_glyph_extents_t for?
What is  `hb_glyph_extents_t` intended for? Internally calculated as floats from outlines (especially in case of VF), returned as integers, they can't be accurate. If they are used for clipping, I believe `floor` & `ceil` would produce better results, but I realized rounding better match FreeType results (though not exactly, probably due to math error).

ebraminio:
Fixes #1940


behdad:
lgtm.

PR #2009, opened 2019-10-07
[glyf] cleanups

ebraminio:


behdad:
lgtm without reviewing much.

PR #2010, opened 2019-10-08
Add subset_offset_array helper function.

garretrieger:
simplifies subsetting offset arrays during stream operations.


behdad:
Thanks for merging.  I'll review and find better home for this function, as part of one of the existing types eventually.

PR #2011, opened 2019-10-08
[WIP][ot-font] Use side-bearing variation before scaling

ebraminio:
Just not to forget and discuss the solution


behdad:
We want to avoid float in the non-variable use-case if possible.  Just stating requirements from my POV.

ebraminio:
maybe some other time

PR #2012, opened 2019-10-10
Reorder U+1C29 LEPCHA VOWEL SIGN OO

### LINKED ISSUE #2000: Difference in behavior between HarfBuzz and DirectWrite
My understanding is that behavior of Harfbuzz and DirectWrite should be identical. I'm seeing a difference when I try to add glyph reordering directly in the OpenType FEA code.

(I am having to implement some reordering in the OpenType directly due to the fact that this font needs to be usable in InDesign which does not support Lepcha reordering.)

My guess about what is happening is that either (1) automatic reordering is happening at a different place in the two engines, which makes the two reordered glyphs end up in a different place relative to each other, or (2) the rules controlling reordering are different.
[ReorderBug.zip](https://github.com/harfbuzz/harfbuzz/files/3678743/ReorderBug.zip)

Also it appears that when I use `languagesystem DFLT dflt` in my FEA code instead of `languagesystem lepc DFLT`, the resulting font does not perform automatic reordering in Harfbuzz (as I would expect) but it still does reordering in DirectWrite. I am testing Harfbuzz in hb-shape on a Linux bionic VM and DirectWrite in Notepad on Windows 10 (v 1903)

dscorbett:
Fixes #2000.

### DIFF
--- a/src/gen-use-table.py
+++ b/src/gen-use-table.py
@@ -383,6 +383,9 @@ def map_to_use(data):
 		# TODO: In USE's override list but not in Unicode 12.0
 		if U == 0x103C: UIPC = Left
 
+		# TODO: https://github.com/harfbuzz/harfbuzz/pull/2012
+		if U == 0x1C29: UIPC = Left
+
 		# TODO: These are not in USE's override list that we have, nor are they in Unicode 12.0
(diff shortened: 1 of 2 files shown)


behdad:
Thanks.  Can we add a test as well?

PR #2013, opened 2019-10-10
Document hb_feature_t.

bungeman:
This documents hb_feature_t. This is motivated mostly by the ambiguity
of the units for 'start' and 'end' (clusters) and whether they are
inclusive or exclusive. This also documents that for lookup type 3 the
value is the one based index into the alternates and that in a list of
features later feature values override previous feature values with the
same tag.

### DIFF
--- a/src/hb-common.h
+++ b/src/hb-common.h
@@ -425,6 +425,21 @@ typedef void (*hb_destroy_func_t) (void *user_data);
  */
 #define HB_FEATURE_GLOBAL_END	((unsigned int) -1)
 
+/**
+ * hb_feature_t:
+ * @tag: a feature tag
+ * @value: 0 disables the feature, non-zero (usually 1) enables the feature.
+ * For features implemented as lookup type 3 (like 'salt') the @value is a one
+ * based index into the alternates.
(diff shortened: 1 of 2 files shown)


ebraminio:
Nice. Thank you

src/hb-common.h
```diff
@@ -425,6 +425,21 @@ typedef void (*hb_destroy_func_t) (void *user_data);
...
+/**
+ * hb_feature_t:
+ * @tag: a feature tag
+ * @value: 0 disables the feature, non-zero (usually 1) enables the feature.
+ * For features implemented as lookup type 3 (like 'salt') the @value is a one
+ * based index into the alternates.
+ * @start: the cluster to start applying this feature setting (inclusive).
+ * @end: the cluster to end applying this feature setting (exclusive).
+ *
+ * The hb_feature_t is the structure that holds information about requested
```
behdad: The `hb_feature_t` part needs some annotation.  Don't remember if it's `#` or `@`.

ebraminio: https://github.com/harfbuzz/harfbuzz/commit/9a68e6b901967da14a5da5acb535c944794627b8

Issue #1686, opened 2019-04-29
Unified API to fetch glyph countors


ebraminio:
Having some API equal to `GetOutlineTextMetricsW`, `CTFontCreatePathForGlyph`, `FT_Outline_Decompose` will be nice now that we have both glyf and CFF parsing for subsetters inside harfbuzz. Personally all I want is to use the API to create a svg path.

* I think I don't like to have an API like `FT_Outline_Decompose` that user should pass callbacks to it but something like `CTFontCreatePathForGlyph` that returns `CGPathRef`, enough information upper layers to do their works not IoC.
* We should consider some flag parameter for the API so user can choose to have either the path in non-hinted (default), hinted or autohinted forms (whenever of whether implemented, we can proxy to freetype for that also)
* `GetOutlineTextMetricsW` and `CTFontCreatePathForGlyph`, both get a matrix in input, I am not sure what is the propose of it, I think we don't need it.
* Should it apply variation itself, thus, getting hb_font_t in input? If so that will be a huge work but that is OK since variable font instantiation is in roadmap anyway.
* We should resolve the subcomponents paths ourselves not exposing the concept to client I guess.
* If we can't offer all the things from the beginning, maybe exposing the API behind some compile flag initially before we get confident what would be best.

I have something related but very subtle on #875 and it doesn't deal with `glyf` glyphs' sub-components and lots of other things

ebraminio:
Useful links:
* Glyphy's FT_Outline_Decompose use https://github.com/behdad/glyphy/blob/73d7fbd96975249c4eb806be94b9889df839335b/src/glyphy-freetype.h#L44-L97
* https://developer.apple.com/documentation/coretext/1510921-ctfontcreatepathforglyph?language=objc
* https://docs.microsoft.com/en-us/windows/desktop/api/wingdi/nf-wingdi-getglyphoutlinew
* https://docs.microsoft.com/en-us/windows/desktop/api/wingdi/nf-wingdi-getoutlinetextmetricsw
* https://github.com/nothings/stb/blob/master/stb_truetype.h features both contours fetch of cff and glyf and rasterization in one single header.
* Typr.js's CFF https://github.com/photopea/Typr.js/blob/877046eeb2c78ffeeb1f34e167308e0b785a1457/src/tabs/cff.js and glyf https://github.com/photopea/Typr.js/blob/877046eeb2c78ffeeb1f34e167308e0b785a1457/src/tabs/glyf.js parsing
[...]

PR #2016, opened 2019-10-13
Implement glyph outline path API

### LINKED ISSUE #2107: No CFF, CFF2 parsing tests?
I could not find any CFF, CFF2 related tests. Afaiu, we need CFF and CFF2 only to get glyph extents during fallback shaping. And we have tests only for `glyf`,`CBDT` and `sbix` tables.

I can work on pull request if needed.

ebraminio:
```
$ make -Csrc test-ot-glyph -j9 && src/test-ot-glyph ./test/subset/data/fonts/Roboto-Regular.ttf
```

Gives

image

Related: #875 #1686

Fixes #2107


ebraminio:
Its `glyf` part is ready for review/

ebraminio:
Its CFF1/CFF2 part is also ok. Is ready to review to get some initial feedback at least I guess, maybe it should be named `hb_ot_path_t` or `hb_path_t` as it can be done for non-opentype fonts also?

behdad:
So, do you want to propose an API or what?

A callback-based API is simplest and preferred IMO.  However, that's trickier to do with composite glyphs in glyf table when component attachment is handled using contour points.  That's the only thing I can think of.

Alternative is API that fills an array with operations.  That doesn't quite solve the composite issue though.

cc @raphlinus

ebraminio:
> A callback-based API is simplest and preferred IMO

Agreed.

ebraminio:
> However, that's trickier to do with composite glyphs in glyf table when component attachment is handled using contour points.

Haven't seen the thing in our glyf implementation, at least yet, but for accents in CFF/CFF2, were you referring to that or were are missing something here?

khaledhosny:
I’m testing this with a CFF font and it works very nicely, except for a few glyphs that seem to be distorted (missing points). Here is a font with a subset of glyphs that are failing for me.

[RanaKufi-Regular.subset.zip](https://github.com/harfbuzz/harfbuzz/files/3854246/RanaKufi-Regular.subset.zip)

khaledhosny:
> I’m testing this with a CFF font and it works very nicely, except for a few glyphs that seem to be distorted

Fixed in #2053

khaledhosny:
(I took the liberty of rebasing this pull request on master and fixing the merge conflict).

ebraminio:
> (I took the liberty of rebasing this pull request on master and fixing the merge conflict).

Thank you Khaled, weren't around for days for some reasons you may know…

> I’m testing this with a CFF font and it works very nicely, except for a few glyphs that seem to be distorted

The fix is very very nice, and very great that having this incomplete work has resulted to a fix in the CFF parsing code, I believe checking some `glyf` having fonts also will result in some interesting findings also.

src/hb-ot-glyph.cc
```diff
@@ -129,6 +129,8 @@ struct _hb_ot_glyph_path_vectors
   {
     coords->fini ();
     commands->fini ();
+    free (coords);
+    free (commands);
```
ebraminio: @khaledhosny, my bad, sadly :/ just wanted to ping you if you testing it somewhere, and there is some other issue on accessing tables that feels isn't limited to this work which currently makes tests fail, will try to file that as a separate but just wanted to mention the work may is very unstable :/

zhuah:
Hi, how's the status of this PR? 
I'm writing a simple layout engine based on harfbuzz and the glyf branch, it works good,  i'm glad to see it get to be merged.

Thanks.

ebraminio:
I can say it works already but it's API should be turned to some callback
based one like freetype, which can be done easily but meaning some more
changes on project internals should happen to make it really zero alloc
underneath.

zhuah:
alright, thanks. 
i also want to support emoji font, maybe i should implement this in my own ,
rather than depends on harfbuzz,
or just use freetype to do these works.

ebraminio:
> i also want to support emoji font, maybe i should implement this in my own ,
> rather than depends on harfbuzz,
> or just use freetype to do these works.

`hb-ot-color.h` can be considered if you are going to handle its bits yourself, freetype provides a higher level API which specially for ot-svg (which not sure if is merged yet but was good looking) can be very useful if you are going to rasterize it eventually.

ebraminio:
Ok, now its API is very similar to what it should be and it zero allocs for CFF1/CFF2, has still some internal temporary allocation in glyf which happens in extents also, otherwide should good to go when we final review the namings and os.

src/hb-ot-var-gvar-table.hh
```diff
@@ -42,12 +42,14 @@ namespace OT {
 
 struct contour_point_t
 {
-  void init (float x_=0.f, float y_=0.f) { flag = 0; x = x_; y = y_; }
+  void init (float x_ = 0.f, float y_ = 0.f, bool is_end_point_ = false)
+  { flag = 0; x = x_; y = y_; is_end_point = is_end_point_; }
```
ebraminio: This could go without addition of `is_end_point` here IIRC, will try to do that before or after the merge.

test/api/test-ot-face.c
```diff
@@ -109,6 +131,14 @@ test_face (hb_face_t *face,
   hb_ot_var_normalize_variations (face, NULL, 0, NULL, 0);
   hb_ot_var_normalize_coords (face, 0, NULL, NULL);
 
+  hb_ot_glyph_decompose_funcs_t funcs;
+  funcs.move_to = (hb_ot_glyph_decompose_move_to_func_t) move_to;
+  funcs.line_to = (hb_ot_glyph_decompose_line_to_func_t) line_to;
+  funcs.conic_to = (hb_ot_glyph_decompose_conic_to_func_t) conic_to;
+  funcs.cubic_to = (hb_ot_glyph_decompose_cubic_to_func_t) cubic_to;
+  for (unsigned gid = 0; gid < 10; ++gid)
```
khaledhosny: I like the new API (I haven’t tested it yet), but for consistency with the rest of HarfBuzz API, this should be modeled after the font functions API (having create and set methods and making the struct opaque).

src/hb-ot-glyph.h
```diff
@@ -0,0 +1,72 @@
...
+ * Glyph decompose callbacks.
+ *
+ * Since: REPLACEME
+ **/
+typedef struct hb_ot_glyph_decompose_funcs_t
+{
+  hb_ot_glyph_decompose_move_to_func_t move_to;
+  hb_ot_glyph_decompose_line_to_func_t line_to;
+  hb_ot_glyph_decompose_conic_to_func_t conic_to;
+  hb_ot_glyph_decompose_cubic_to_func_t cubic_to;
```
khaledhosny: Do we need `close` and `end` functions? or how do clients differentiate between closed and open paths (not sure if open paths are allowed at all).

test/api/test-ot-face.c
ebraminio: Oh, thank you so much, will apply.

src/hb-ot-glyph.h
ebraminio: I have the `close` for CFF and end for `glyf`, they however not exposed by freetype so guess we can also without them.

src/hb-ot-glyf-table.hh
```diff
@@ -1039,6 +1042,62 @@ struct glyf
         add_gid_and_children (item.glyphIndex, gids_to_retain, depth);
     }
 
+    bool
+    get_path (hb_font_t *font, hb_codepoint_t gid,
+	      hb_ot_glyph_decompose_funcs_t *funcs, void *user_data) const
```
ebraminio: Should mention learned from https://github.com/opentypejs/opentype.js/blob/4e0bb99/src/tables/glyf.js#L222 for this

src/hb-ot-glyph.h
khaledhosny: I prefer having an explicit `close` (and `end` if open contours are possible). Makes client code simpler IMO.

ebraminio: Your request is definitely valid, I am trying to see where to insert those :)

We have a start_path and end_path in CFF/CFF2 https://github.com/harfbuzz/harfbuzz/pull/2016/files#diff-7ae1ce6c07f89cffd0c78595d6db8134R359-R360 which perhaps matches `open()` and `end()` or `close()`? am confused more about glyf part, https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L1232-L1242 it wraps each simple glyph with `open()`, `end()`, this https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L1242 however doesn't have begin part weirdly. So I think we can skip those for now if we can draw SVG path without them for now.

fonttools also does a nice thing https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/pens/basePen.py#L187 that has three must override functions and having the rest optional, its `_qCurveToOne` is able to convert quadratic bezier command to cubic, nice thing I guess, perhaps we can provide it (opt-in/default disabled, later?) using some config on the proposed hb_ot_glyph_decompose_funcs_t, wdyt?

test/api/test-ot-face.c
ebraminio: Done, thanks a lot for the suggestion, please have a look.

src/hb-ot-glyph.h
ebraminio: Okay based on opentype.js's closePath have added open_path and close_path callbacks.

ebraminio: And guess turning quadratic bezier to cubic isn't that suitable for us.

ebraminio: Have a look please :)

ebraminio:
Guess we can use @RazrFalcon  https://github.com/RazrFalcon/ttf-parser/blob/0374beca5100b7fbdb308174d0b6f533682d1d9e/tests/tests.rs API and tests for this

RazrFalcon:
`ttf-parser` test suite is garbage, sadly. Not sure how it can help.

ebraminio:
test/api/test-ot-glyph.c of this patch is similar to ehat you have there.

src/hb-ot-glyph.cc
```diff
@@ -0,0 +1,241 @@
...
+hb_ot_glyph_decompose_funcs_set_close_path_func (hb_ot_glyph_decompose_funcs_t           *funcs,
+						 hb_ot_glyph_decompose_close_path_func_t  close_path)
+{
+  if (unlikely (funcs == &Null (hb_ot_glyph_decompose_funcs_t))) return;
+  funcs->close_path = close_path;
+}
+
+static void
+_move_to_noop (hb_position_t to_x HB_UNUSED, hb_position_t to_y HB_UNUSED, void *user_data HB_UNUSED) {}
+
```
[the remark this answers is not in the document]
ebraminio: Done :)

ebraminio:
> The new API looks good, thank you. I tested it briefly and I’m not having any issues so far.

Thank you very much :)

Will adopt harfbuzzjs with the changes and merge afterward if no other concern exist.

ebraminio:
Here is `hbjs_ot_glyph_svg` https://github.com/harfbuzz/harfbuzzjs/commit/15d0c426b89d359d6bd603e7321e68a398faa74a#diff-137c3b7d26662504dcff2b682c482699

ebraminio:
Merged, it's API however is considered unstable (or maybe get disabled) till the next release.

ebraminio:
WDYT about renaming hb_ot_glyph_decompose* to hb_glyph_decompose* and dropping _ot_? Even the fact we mainly focus on OT on hb one can argue we might have non-OT vector graphic glyph sometime.

henrik-lindqvist:
What happened to this API?

PR #2019, opened 2019-10-21
[subset] Fix null-point dereference

ebraminio:
Fixes https://crbug.com/oss-fuzz/18363


ebraminio:
Or maybe doing null check once before writing/dereferencing?

ebraminio:
Closing in favor of #2023

PR #2022, opened 2019-10-21
[gpos] Don't move pointer when match_glyph_data is not set

ebraminio:
Possible fix to revive ubsan bot fail on https://circleci.com/gh/harfbuzz/harfbuzz/114888


ebraminio:
@behdad guess you should have a look at this, I don't like the way this struct works but guess the fix isn't hurting at least

behdad:
Okay so we were increasing null pointers that we never dereference...  Guess is fine merging.  One day we switch to templates for these and will be optimized out.

PR #2023, opened 2019-10-21
fuzzer crash fix: Null-dereference WRITE

qxliu76:
fix for fuzzer crash:   https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=18363


src/hb-ot-layout-common.hh
```diff
@@ -1545,6 +1545,7 @@ struct ClassDefFormat2
     range_rec.value = prev_klass;
 
     RangeRecord *record = c->copy (range_rec);
+    if (unlikely (!record)) return_trace (false);
```
ebraminio: Aren't the other null checks unlikely also?

ebraminio:
Please add the test case also, thanks

qxliu76:
> Please add the test case file also, thanks

added. :)

ebraminio:
Thanks :)

PR #2024, opened 2019-10-21
Speed up intersects_array() and array_t::hash() when running with sanitization.

garretrieger:


src/hb-array.hh
```diff
@@ -99,7 +99,14 @@ struct hb_array_t : hb_iter_with_fallback_t<hb_array_t<Type>, Type&>
   template <typename T> operator T * () const { return arrayZ; }
 
   HB_INTERNAL bool operator == (const hb_array_t &o) const;
-  HB_INTERNAL uint32_t hash () const;
+
+  inline HB_INTERNAL uint32_t hash () const {
```
ebraminio: We put curly on a new line in harfbuzz, you probably shouldn't need need inline, as that is default on methods defined inside a class and HB_INTERNAL shouldn't be needed as we have -fvisibility-inline-hidden in our build.

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -712,11 +712,11 @@ static inline bool intersects_array (const hb_set_t *glyphs,
 				     intersects_func_t intersects_func,
 				     const void *intersects_data)
 {
-  return
-  + hb_iter (values, count)
-  | hb_map ([&] (const HBUINT16 &_) { return intersects_func (glyphs, _, intersects_data); })
-  | hb_any
-  ;
+  for (const HBUINT16 &_ : + hb_iter (values, count))
+  {
```
ebraminio: Feel free to remove curly braces in one line codes, compilers nowadays raise an interesting error on misleading indention.

behdad:
Thanks Garret.  I see you merged this, but can you please help me understand a couple of things:

1. Ideally the iter version should compile to same code.  Can you share your perf measurements?

2. The specialization for char that does the exact same thing should not be necessary/faster?  My plan for specialization was a hashing scheme that would process one int at a time so it would be 4 times faster.  Doing the same thing in a specialization doesn't make sense to me.

ebraminio:
> Ideally the iter version should compile to same code. Can you share your perf measurements?

Right, that specific bot has flags intended to make inlining harder so feel free to ignore it and add -O3 to it and remove `-fno-omit-frame-pointer` from it, but now I see valgrind fails also so maybe we should go for reducing `HB_MAX_LANGSYS` but I don't like to reduce just for specific users.

PR #2026, opened 2019-10-22
Store the hash value of a key in item_t in hb_map

garretrieger:
Reduces the number of hash computations needed.


behdad:
Lgtm!  Haven't checked why bots unhappy.

behdad:
Maybe also clear hash in clear()?

PR #2027, opened 2019-10-23
[subset] closure lookups for GSUB/GPOS

qxliu76:
closure and remap lookup indexes for GSUB/GPOS, save them in the subset_plan object

### DIFF
--- a/src/hb-ot-layout-gsub-table.hh
+++ b/src/hb-ot-layout-gsub-table.hh
@@ -56,6 +56,8 @@ struct SingleSubstFormat1
     ;
   }
 
+  void closure_lookups (hb_closure_lookups_context_t *c) const {}
+
   void collect_glyphs (hb_collect_glyphs_context_t *c) const
   {
     if (unlikely (!(this+coverage).add_coverage (c->input))) return;
@@ -153,6 +155,8 @@ struct SingleSubstFormat2
     | hb_sink (c->output)
     ;
   }
+
+  void closure_lookups (hb_closure_lookups_context_t *c) const {}
 
   void collect_glyphs (hb_collect_glyphs_context_t *c) const
   {
@@ -395,6 +399,8 @@ struct MultipleSubstFormat1
     ;
   }
 
+  void closure_lookups (hb_closure_lookups_context_t *c) const {}
+
   void collect_glyphs (hb_collect_glyphs_context_t *c) const
   {
     if (unlikely (!(this+coverage).add_coverage (c->input))) return;
@@ -604,6 +610,8 @@ struct AlternateSubstFormat1
     | hb_apply ([c] (const AlternateSet &_) { _.closure (c); })
     ;
   }
+
+  void closure_lookups (hb_closure_lookups_context_t *c) const {}
 
   void collect_glyphs (hb_collect_glyphs_context_t *c) const
   {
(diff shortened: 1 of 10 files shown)


src/hb-ot-layout-gsub-table.hh
```diff
@@ -1533,6 +1565,25 @@ template <typename context_t>
   const SubstLookup &l = c->face->table.GSUB.get_relaxed ()->table->get_lookup (lookup_index);
   return l.dispatch (c);
 }
+
+/*static*/ inline hb_closure_lookups_context_t::return_t SubstLookup::dispatch_closure_lookups_recurse_func (hb_closure_lookups_context_t *c, unsigned this_index)
+{
```
garretrieger: Similar to GPOS this method can be simplified by calling closure_lookups

src/hb-ot-layout.cc
```diff
@@ -1212,6 +1212,52 @@ hb_ot_layout_collect_lookups (hb_face_t      *face,
...
+        l.closure_lookups (&c, lookup_index);
+        return;
+      }
+    }
+  }
+
+  hb_set_union (lookup_indexes, &visited_lookups);
+  hb_set_subtract (lookup_indexes, &inactive_lookups);
+
+  hb_set_clear (&visited_lookups);
```
garretrieger: no need to clear these they'll be destructed on return anyways.

ebraminio:
Oh, thanks for this work, as this is going to introduce a new public API, `hb_ot_layout_closure_lookups`, can you add it to https://github.com/harfbuzz/harfbuzz/blob/master/docs/harfbuzz-sections.txt I think it would be fine to go without test here but please add the API here https://github.com/harfbuzz/harfbuzz/blob/master/test/api/test-ot-face.c and optionally in a separate unit in https://github.com/harfbuzz/harfbuzz/tree/master/test/api after these I think it should be fine to merge this if Behdad also has no concern, thanks for the patience btw :)

qxliu76:
> Oh, thanks for this work, as this is going to introduce a new public API, `hb_ot_layout_closure_lookups`, can you add it to https://github.com/harfbuzz/harfbuzz/blob/master/docs/harfbuzz-sections.txt I think it would be fine to go without test here but please add the API here https://github.com/harfbuzz/harfbuzz/blob/master/test/api/test-ot-face.c and optionally in a separate unit in https://github.com/harfbuzz/harfbuzz/tree/master/test/api after these I think it should be fine to merge this if Behdad also has no concern, thanks for the patience btw :)

done. Thanks for your review!

src/hb-ot-layout.cc
```diff
@@ -1191,6 +1191,49 @@ hb_ot_layout_collect_lookups (hb_face_t      *face,
...
+
+/**
+ * hb_ot_layout_closure_lookups:
+ * @face: #hb_face_t to work upon
+ * @table_tag: HB_OT_TAG_GSUB or HB_OT_TAG_GPOS
+ * @lookup_indexes: (inout): lookup_indices collected from feature
+ * list
+ *
+ * Returns all inactive lookups reachable from lookup_indices
+ **/
```
ebraminio: Please add a `* Since: REPLACEME` so we can replace the tag while releasing, after that it ready to merge I believe. Thanks :)

ebraminio:
Great! Thank you :)

PR #2028, opened 2019-10-24
[ci] Fix subset timeout issue of tsan/valgrind bots

### LINKED ISSUE #2020: Make -5717414645334016 subset faster
Had to upload https://crbug.com/oss-fuzz/17898 in tree e76678315 as it was closed and as I didn't want we won't forget it but now it fails tsan bot https://circleci.com/gh/harfbuzz/harfbuzz/114791 @garretrieger @rsheeter @qxliu76 can you make faster please?

ebraminio:
Fixes #2020

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ jobs:
 
   macos-10.12.6-aat-fonts:
     macos:
-      xcode: "9.2.0"
+      xcode: "9.0.1"
     steps:
       - checkout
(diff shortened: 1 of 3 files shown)


behdad:
Why do we run those w -O1?

I prefer we dont change compile parameters for bots but not very keen on it.

Also in commit message subject please say what change does, not just "fix xyz issue".

ebraminio:
> Why do we run those w -O1?

To get better stacktrace on failures.

> Also in commit message subject please say what change does, not just "fix xyz issue".

Correct.

Issue #2031, opened 2019-10-27
cmap table that maps a code point to GID 0


khaledhosny:
The font [`cmunrm.otf`](http://tug.ctan.org/fonts/cm-unicode/fonts/otf/cmunrm.otf) has a format 4 `cmap` subtable that map U+0000 to GID 0, and this seems to cause `hb_face_collect_unicodes()` to include U+0000, but [the spec](https://docs.microsoft.com/en-us/typography/opentype/spec/cmap#overview) says:

> Regardless of the encoding scheme, character codes that do not correspond to any glyph in the font should be mapped to glyph index 0. The glyph at this location must be a special glyph representing a missing character, commonly known as .notdef.

So shouldn’t HarfBuzz skip U+0000 then? Furthermore, `hb_font_get_nominal_glyph()` returns `false` for U+0000 in this font, which is rather inconsistent.

khaledhosny:
For comparison, `FT_Get_First_Char()` gives 0x0001, skipping the character mapped to GID 0.

behdad:
I agree, we should just remove 0 from collected unicodes of OpenType fonts.

behdad:
Simple patch please.

behdad:
That was wrong.  I'm fixing correctly now.

Existing code tried to do this correctly, but apparently doesn't in some places.  Investigating.

behdad:
Humm.  I don't see where in the code this is happening.

Khaled, can you point out to what mapping in the font is causing this?  I'm failing to find it.  Can you please debug and submit a PR?  Ideally with a test?  Thanks.

khaledhosny:
I’ll try to debug later, here is the test code I used to check this:
```c
#include <assert.h>
#include <stdio.h>
#include <hb.h>

int
main(int argc, char** argv)
{
	if (argc < 2)
	{
		fprintf(stderr, "Usage: %s font\n", argv[0]);
		return 1;
	}

	hb_blob_t *blob = hb_blob_create_from_file(argv[1]);
	hb_face_t *face = hb_face_create(blob, 0);
	hb_font_t *font = hb_font_create(face);

	hb_set_t *codes = hb_set_create();
	hb_face_collect_unicodes(face, codes);

	size_t count = hb_set_get_population(codes);
	hb_codepoint_t code = HB_SET_VALUE_INVALID;
	hb_codepoint_t glyph = 0;
	while (hb_set_next(codes, &code))
	{
[...]

PR #2032, opened 2019-10-28
Don't include codepoint 0 in the results of collect_unicodes.

garretrieger:
It is always assumed to be the notdef glyph. Fix for #2031.


behdad:
Huh.  This is wrong.  I utterly confused myself.

We should *not* add unicode characters who *map to* gid 0.

PR #2034, opened 2019-10-28
[cmap] Check GID before adding ranges in format 4 & 12

khaledhosny:
Fixes https://github.com/harfbuzz/harfbuzz/issues/2031


src/hb-ot-cmap-table.hh
```diff
@@ -344,7 +344,17 @@ struct CmapSubtableFormat4
       {
 	unsigned int rangeOffset = this->idRangeOffset[i];
 	if (rangeOffset == 0)
-	  out->add_range (this->startCount[i], this->endCount[i]);
+	{
+	  for (hb_codepoint_t codepoint = this->startCount[i];
+	       codepoint <= this->endCount[i];
```
behdad: Nicer if you also save endCount[i] in a variable.

src/hb-ot-cmap-table.hh
```diff
@@ -344,7 +344,17 @@ struct CmapSubtableFormat4
...
+	       codepoint++)
+	  {
+	    hb_codepoint_t gid = codepoint + this->idDelta[i];
```
behdad: I *think* this needs to be modulo 65536?  Check the get_glyph code for it.

src/hb-ot-cmap-table.hh
khaledhosny: I just copied the loop from the branch below it.

src/hb-ot-cmap-table.hh
khaledhosny: Right. I did check get_glyph but overlooked this.

PR #2041, opened 2019-10-29
Add in_range in hb_bytes_t to merge range_checker_t with it

### LINKED ISSUE #2033: Promote range_checker_t
In `hb-ot-var-gvar-table.hh` there's `range_checker_t`.  We should promote that and use it in places that we currently abuse `hb_sanitize_context_t` for access-time range checking.  Eg. in `hb-aat-layout-common.hh`.  That might be the only one.  But I think there's other places we can use `range_checker_t` instead of adhoc `end` pointers as well.

ebraminio:
Fixes #2033

### DIFF
--- a/src/hb-array.hh
+++ b/src/hb-array.hh
@@ -196,6 +196,15 @@ struct hb_array_t : hb_iter_with_fallback_t<hb_array_t<Type>, Type&>
   const T *as () const
   { return length < hb_null_size (T) ? &Null (T) : reinterpret_cast<const T *> (arrayZ); }
 
+  template <typename T,
+	    unsigned P = sizeof (Type),
+	    hb_enable_if (P == 1)>
+  bool in_range (const T *p, unsigned int size = T::static_size) const
+  {
+    return ((const char *) p) >= arrayZ
+	&& ((const char *) p + size) <= arrayZ + length;
+  }
+
   /* Only call if you allocated the underlying array using malloc() or similar. */
   void free ()
   { ::free ((void *) arrayZ); arrayZ = nullptr; length = 0; }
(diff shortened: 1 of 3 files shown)


src/hb-array.hh
```diff
@@ -196,6 +196,15 @@ struct hb_array_t : hb_iter_with_fallback_t<hb_array_t<Type>, Type&>
...
 
+  template <typename T,
+	    unsigned P = sizeof (Type),
+	    hb_enable_if (P == 1)>
+  bool in_range (const T *p, unsigned int size = T::static_size) const
+  {
+    return ((const char *) p) >= arrayZ
+	&& ((const char *) p + size) <= arrayZ + length;
+  }
+
```
behdad: The addition to size can overflow.  I know not big deal but good to check.  Check how we do in sanitizer.

behdad: Can also maybe enable if byte array or if T is_same as array member type...

ebraminio: > Check how we do in sanitizer.

Oh, worth to have, there is an assert and a sign check there, can you please indicate which line you are referring. Thanks

behdad: Maybe also rename to `check_range`.
```c++
  bool check_range (const void *base,
                    unsigned int len) const
  {
    const char *p = (const char *) base;
    bool ok = !len ||
              (this->start <= p &&
               p <= this->end &&
               (unsigned int) (this->end - p) >= len &&
               this->max_ops-- > 0);

    DEBUG_MSG_LEVEL (SANITIZE, p, this->debug_depth+1, 0,
                     "check_range [%p..%p]"
                     " (%d bytes) in [%p..%p] -> %s",
                     p, p + len, len,
                     this->start, this->end,
                     ok ? "OK" : "OUT-OF-RANGE");

    return likely (ok);
  }
```

PR #1437, opened 2018-12-01
Add support for Meson build system

tp-m:
Probably not perfect, but builds for me with test suite passing on:
 - Linux (64-bit, gcc 8.2.0 / Debian sid)
 - Windows with MSVC (32-bit, vs2015)
 - Windows mingw (32-bit, fontconfig disabled).

PR #2045, opened 2019-11-05
[cmake] Add harfbuzz-icu library

donny-dont:
Currently the CMake files do not create the harfbuzz-icu library. This creates it if `HB_HAVE_ICU` is turned on.


donny-dont:
I know #1437 is a thing but not sure the migration timeline and the library isn't currently being created.

ebraminio:
Looks good! Thanks :)

PR #2046, opened 2019-11-08
fix range-for-loop for hb_set_t

qxliu76:
Wrong end(), which is actually returning begin(). It's not needed in fact.

### DIFF
--- a/src/hb-set.hh
+++ b/src/hb-set.hh
@@ -707,7 +707,6 @@ struct hb_set_t
     void __next__ () { s->next (&v); if (l) l--; }
     void __prev__ () { s->previous (&v); }
     unsigned __len__ () const { return l; }
-    iter_t end () const { return iter_t (*s); }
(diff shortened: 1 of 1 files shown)


src/hb-set.hh
```diff
@@ -707,7 +707,6 @@ struct hb_set_t
     void __next__ () { s->next (&v); if (l) l--; }
     void __prev__ () { s->previous (&v); }
     unsigned __len__ () const { return l; }
-    iter_t end () const { return iter_t (*s); }
```
behdad: Removing it is not a good idea.  Just return `iter_t ()`.

qxliu76: > Removing it is not a good idea. Just return `iter_t ()`.

I tried that, and returning iter_t () seems make "test-iter" test running forever

behdad:
Pushed fix to master.  Can you add a test to test-iter somehow unintrusively?

Issue #1012, opened 2018-04-28
USE data overrides


behdad:
From an opentype-layout thread, from Andrew Glass:
```
Here is the override:

  # ================================================
  # Indic_Matra_Category=Bottom
  0F72        ; Bottom # Mn      TIBETAN VOWEL SIGN I
  0F7A..0F7D  ; Bottom # Mn  [4] TIBETAN VOWEL SIGN E..TIBETAN VOWEL SIGN OO
  0F80        ; Bottom # Mn      TIBETAN VOWEL SIGN REVERSED I
  11127..11129; Bottom # Mn  [3] CHAKMA VOWEL SIGN A..CHAKMA VOWEL SIGN II
  1112D       ; Bottom # Mn      CHAKMA VOWEL SIGN AI
  11130       ; Bottom # Mn      CHAKMA VOWEL SIGN OI
```

behdad:
From another email from Andrew Glass, also:
```
  # Indic_Syllabic_Category=Gemination_Mark
  11134      ; Gemination_Mark # Mc      CHAKMA MAAYYAA
```

behdad:
Actually, those are now listed in USE spec itself:
https://docs.microsoft.com/en-gb/typography/script-development/use#overrides

```
Overrides to Indic_Syllabic_Category

  # Indic_Syllabic_Category=Bindu
  AA29       ; Bindu          # Mn       CHAM VOWEL SIGN AA
  # ================================================
  # Indic_Syllabic_Category=Nukta
  0F71       ; Nukta          # Mn       TIBETAN VOWEL SIGN AA
  # ================================================
  # Indic_Syllabic_Category=Tone_Mark
  A982       ; Tone_Mark      # Mn       JAVANESE SIGN LAYAR
  # ================================================
  # Indic_Syllabic_Category=Consonant_Dead
  0F7F       ; Consonant_Dead # Mc       TIBETAN SIGN RNAM BCAD
  # ================================================
  # Indic_Syllabic_Category=Gemination_Mark 
  11134      ; Gemination_Mark # Mc      CHAKMA MAAYYAA

Overrides to Indic_Positional_Category

  # Indic_Matra_Category=Top
  0F74        ; Top     # Mn      TIBETAN VOWEL SIGN U
  AA35        ; Top     # Mn      CHAM CONSONANT SIGN
  1A18        ; Top     # Mn      BUGINESE VOWEL SIGN U

  # ================================================
[...]

PR #2047, opened 2019-11-09
Insert a dotted circle within <U+0B85, U+0BC2>

dscorbett:
This adds the Tamil vowel constraint mentioned in MicrosoftDocs/typography-issues#205. I removed HBIndicVowelConstraints.txt because the file is now available at https://github.com/microsoft/font-tools/blob/master/USE/IndicShapingInvalidCluster.txt.


behdad:
Oh those data files from MS are amazing!

Can we copy the one we use into our git tree, so it's tracked?  Unlike Unicode, there's no easy versioning for those.

behdad:
Just import the one you are using now.  We process the rest in https://github.com/harfbuzz/harfbuzz/issues/1012

dscorbett:
Where should I put [the license file](https://github.com/microsoft/font-tools/blob/d6514ee7f25cf04cd0ce6427dc9626cab191db1e/LICENSE)?

behdad:
> Where should I put [the license file](https://github.com/microsoft/font-tools/blob/d6514ee7f25cf04cd0ce6427dc9626cab191db1e/LICENSE)?

It's very similar to ours.  Maybe create the directory src/ms-use and put the data files and their license there.

behdad:
Data files like those are not quite covered by copyright per se also.

dscorbett:
Better safe than sorry! I added the file.

behdad:
Thanks!

behdad:
Probably should add a README to that directory pointing to the source repo.  No need to include hash of the commit we have pulled down though.

Also need to make `make dist` distribute them. :)

dscorbett:
I’ll do that for #1012.

PR #2053, opened 2019-11-17
[cff] Fix a typo in processing rlinecurve

khaledhosny:
We want to process all pairs of arguments except the last 6 as lines, so should have been subtracting 6 here, otherwise if the number of arguments happens to be multiples of 6 they will be all treated as curves.

See https://github.com/harfbuzz/harfbuzz/pull/2016#issuecomment-554640098

### DIFF
--- a/src/hb-cff-interp-cs-common.hh
+++ b/src/hb-cff-interp-cs-common.hh
@@ -573,7 +573,7 @@ struct path_procs_t
   static void rlinecurve (ENV &env, PARAM& param)
   {
     unsigned int i = 0;
-    unsigned int line_limit = (env.argStack.get_count () % 6);
+    unsigned int line_limit = (env.argStack.get_count () - 6);
     for (; i + 2 <= line_limit; i += 2)
     {
       point_t pt1 = env.get_pt ();
--- a/src/hb-cff-interp-cs-common.hh
+++ b/src/hb-cff-interp-cs-common.hh
@@ -551,8 +551,13 @@ struct path_procs_t
 
   static void rcurveline (ENV &env, PARAM& param)
   {
+    unsigned int arg_count = env.argStack.get_count ();
+    if (unlikely (arg_count < 8))
+      return;
+
     unsigned int i = 0;
-    for (; i + 6 <= env.argStack.get_count (); i += 6)
+    unsigned int curve_limit = arg_count - 2;
+    for (; i + 6 <= curve_limit; i += 6)
     {
       point_t pt1 = env.get_pt ();
       pt1.move (env.eval_arg (i), env.eval_arg (i+1));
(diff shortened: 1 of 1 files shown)


khaledhosny:
See the font in the linked comment above. Here is the first glyph before and after this change:
before after

The glyph has:
```
322 322 -321 321 -105 107 24 -58 -15 -73 -46 -46 rlinecurve
```

which before this change was converted to 2 curve segments, while after this change it is (correctly) converted to 3 line and 1 curve segments.

Quoting from Technical Note \#5177:
> **rlinecurve** |- {dxa dya}+  dxb dyb dxc dyc dxd dyd **rlinecurve** (25) |-
> is equivalent to one **rlineto** for each pair of arguments beyond the six arguments *dxb...dyd* needed for the one **rrcurveto** command. The number of lines is determined from the count of items on the argument stack.

So there is only one curve segment and as many line segments as fits.

blueshade7:
I see thanks. I believe we need to check if `get_count()` is >= 6 otherwise `- 6` may wrap around. Also the second loop is unnecessary.

khaledhosny:
> Also the second loop is unnecessary.

The same applies for `rcurveline` above it. The loop at least makes sure there are enough arguments on the stack, otherwise the code needs a bit more validity checks for the number of arguments here. What should be done if the argument stack is too short or too long?

blueshade7:
> What should be done if the argument stack is too short or too long?

CFF or CFF2 specs do not appear to say unfortunately.
Obviously we shouldn’t crash or behave unpredictably.

What AFDKO is doing in [`t2cstr.c`](https://github.com/adobe-type-tools/afdko/blob/4030ed294aa4734d807e4b1b2566a4fab093bf54/c/public/lib/source/t2cstr/t2cstr.c#L1513) are:

1. minimum stack depth is checked, if short throw error
2. arguments are taken from stack bottom; extra arguments on top are ignored

For 1, I would suggest just ignore the opcode, instead of failing the font entirely.

Let’s follow 2. Namely, I suggest changing the for-loop in question to if-statement:
```
    for (; i + 6 <= env.argStack.get_count (); i += 6)
```
to:
```
    if (i + 6 <= env.argStack.get_count ())
```

behdad:
lgtm

khaledhosny:
> What AFDKO is doing in [`t2cstr.c`](https://github.com/adobe-type-tools/afdko/blob/4030ed294aa4734d807e4b1b2566a4fab093bf54/c/public/lib/source/t2cstr/t2cstr.c#L1513) are:
>    1. minimum stack depth is checked, if short throw error
>    2. arguments are taken from stack bottom; extra arguments on top are ignored

I did something to this effect before seeing your comment, please check.

PR #2054, opened 2019-11-17
[CFF] get glyph name out of CFF

### LINKED ISSUE #2052: [ot-font] Get glyph names from CFF table
Currently otf-font gets glyph names from `post` table only, but since fonts with `CFF` table use `post` version 3 we don’t get any glyph names for such fonts and need to get them from the `CFF` table.

blueshade7:
Fixes #2052
Added `lookup_standard_string_for_sid()` for standard string lookup.
Implemented `cff1::accelerator_t::get_glyph_name()` and hooked it up with `hb_ot_get_glyph_name()`
For this, repurposed & moved `glyph_to_sid()` from `cff1::accelerator_subset_t`  to `cff1::accelerator_temp_t` along with `glyph_to_code()` etc.
This left `cff1::accelerator_subset_t` gutted as a mere instantiation of `accelerator_temp_t`. 
The above change exposed a fuzz crasher in `get_sid()`. Fixed this by adding `num_glyphs` param to check `glyph` against it.

Added `cff1::accelerator_t::get_glyph_from_name()` to return glyph ID given a glyph name.
Added test program` test-ot-glyphname` to exercise both `get_glyph_name()` and `get_glyph_from_name()` given a CFF font file.

Added api test as well.


khaledhosny:
Thanks, it would be nice to implement `get_glyph_from_name()` as well.

src/hb-ot-cff1-table.cc
```diff
@@ -125,6 +125,67 @@ static const uint8_t standard_encoding_to_sid [] =
     0,   144,   0,    0,    0,  145,    0,    0,  146,  147,  148,  149,    0,    0,    0,    0
 };
 
+/* SID to standard string */
+static const char *sid_to_standard_string [] =
```
behdad: We need better storage for this.  See `hb-string-array.hh`.

src/hb-ot-cff1-table.hh
```diff
@@ -40,6 +40,7 @@ namespace CFF {
 #define HB_OT_TAG_cff1 HB_TAG('C','F','F',' ')
 
 #define CFF_UNDEF_SID   CFF_UNDEF_CODE
+#define HB_CFF_STD_STR_COUNT	391
```
behdad: This also magic.  Should rewrite in terms of the actual array size or something else.

src/hb-ot-cff1-table.cc
blueshade7: ok

src/hb-ot-cff1-table.hh
blueshade7: I see in `hb-ot-post-table.hh`, the value of `NUM_FORMAT1_NAMES` is hard-coded to 258 when it could be calculated in `hb-string-array.h`:
```
#define HB_STRING_ARRAY_LENGTH_NAME	HB_PASTE(HB_STRING_ARRAY_NAME, _length)
static const unsigned HB_STRING_ARRAY_LENGTH_NAME = (ARRAY_LENGTH(HB_STRING_ARRAY_OFFS_NAME)-1);
```
Are you suggesting something like this?

behdad: Sure, that would work.  Please submit that separately first.

blueshade7: created PR 2055

blueshade7:
> Thanks, it would be nice to implement `get_glyph_from_name()` as well.

Doable, but can't find any tools/tests calling it. Need to write a test I guess.

khaledhosny:
> > Thanks, it would be nice to implement `get_glyph_from_name()` as well.
> 
> Doable, but can't find any tools/tests calling it. Need to write a test I guess.

I see some uses in `test/api/test-ot-face.c` and `test/api/test-ot-math.c`.

blueshade7:
> I see some uses in `test/api/test-ot-face.c` and `test/api/test-ot-math.c`.

I tried them, but `test-ot-face` calls `hb_font_get_glyph_from_name` with a null face. `test-ot-math` ends up in ft. Neither hits my code (I have it already).

behdad:
Just add another `src/test-ot-glyphname` or something.

blueshade7:
ok

blueshade7:
Added test program `test-ot-glyphname` to exercise `cff1::accelerator_t::get_glyph_name()` and `cff1::accelerator_t::get_glyph_from_name()` given a CFF font file.

behdad:
Is this ready now?

blueshade7:
> Is this ready now?

Would be nice if `test-ot-glyphname` hooked up to a build test but couldn't figure out how to. No tests under `src` are. Otherwise, I think it's ready.

behdad:
In that case please add a test in tests/api.  There's precedent in those tests to open a font file I think.

blueshade7:
added api test
then one fuzzed file failed for timeout
doesn't make sense & can't repro ☹️

blueshade7:
rebuild invoked by rebase let the failure go away (for now)

blueshade7:
it's ready now

khaledhosny:
Can this be merged now?

src/hb-ot-font.cc
```diff
@@ -210,7 +210,11 @@ hb_ot_get_glyph_name (hb_font_t *font HB_UNUSED,
 		      void *user_data HB_UNUSED)
 {
   const hb_ot_face_t *ot_face = (const hb_ot_face_t *) font_data;
-  return ot_face->post->get_glyph_name (glyph, name, size);
+  hb_bool_t ret = ot_face->post->get_glyph_name (glyph, name, size);
+#ifndef HB_NO_OT_FONT_CFF
+  if (!ret) ret = ot_face->cff1->get_glyph_name (glyph, name, size);
+#endif
+  return ret;
```
behdad: This, and next block, maybe better to rewrite as bunch of `if (...get_glyph_name (...)) return true;`.  Looks good to me otherwise.  Feel free to merge.

blueshade7: made the suggested change

behdad: Thanks.  There's a similar block in the other function (`get_glyph_from_name`) as well.

blueshade7: ok... fixed `hb_ot_get_glyph_from_name` and `hb_ot_get_glyph_extents` as well taking this opportunity

behdad: Thank you!!!!!

PR #2056, opened 2019-11-19
Don’t modify a ccc to 6

dscorbett:
ccc=6 will be used in Unicode 13.0, so I moved these modified values out of its way.


behdad:
Thanks!

PR #2058, opened 2019-11-22
Add meson support

boxerab:
Build the latest version of harfbuzz with meson.


khaledhosny:
We already have an open PR for meson #1437, what is the relation between this and the other one?

boxerab:
This PR adds one new commit to #1437 that brings it up to date with latest Harfbuzz master. If I can simply add this commit to #1437 somehow, then this could be closed.

PR #2065, opened 2019-12-08
[WIP] Use SIMD intrinsics when available to speed up shaping

behdad:
Will fix https://github.com/harfbuzz/harfbuzz/issues/566 when ready.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -416,39 +416,47 @@ dump_use_data_SOURCES = dump-use-data.cc hb-ot-shape-complex-use-table.cc
 dump_use_data_CPPFLAGS = $(HBCFLAGS)
 dump_use_data_LDADD = libharfbuzz.la $(HBLIBS)
 
-COMPILED_TESTS = test-algs test-iter test-meta test-number test-ot-tag test-unicode-ranges test-bimap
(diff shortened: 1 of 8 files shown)


behdad:
@jfkthame @drott @nona-google 
Nothing to test yet, but really curious to explore speedup opportunities with this work and your help.

behdad:
Okay so the numbers were disappointing.  Using vector ops seems to stall the pipeline and instruction-level parallelism suffers.  Twice slower with NotoNastaliqUrdu...

I keep investigating.  But am convinced now that the bsearch cannot be improved.  Still might be a win for the `hb_set_digest_t` part of the plan.

behdad:
So on the status of this experiment. So far I wasn't able to gain any speedup. I think the reason and how to fix is:
- k-ary search to benefit from this is only relevant for largest of tables. Since ours are limited to 64k, we might never get there.
- Since overhead of the SIMD search is high, it means that we should use linear-search for arrays up to a much higher size.

Based on the above, I get a sense that maybe we don't need the k-ary search and should just invest in a SIMD lsearch and make our bsearch use that more often.  So that's what the next step would be. Which would be a lot easier when we have https://github.com/harfbuzz/harfbuzz/pull/2506

behdad:
And this is up for grabs if someone wants to learn and continue.

PR #2069, opened 2019-12-11
Fix unary minus operator applied to unsigned int

Daeren:
Applying unary minus operator to unsigned int causes the following error on MSVS: error C4146
This patch fixes the error.

### DIFF
--- a/src/hb-ot-color-sbix-table.hh
+++ b/src/hb-ot-color-sbix-table.hh
@@ -237,7 +237,7 @@ struct sbix
       extents->x_bearing = x_offset;
       extents->y_bearing = png.IHDR.height + y_offset;
       extents->width     = png.IHDR.width;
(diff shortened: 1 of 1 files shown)


ebraminio:
This change caused this failure ubsan failure,
```
clusterfuzz-testcase-minimized-hb_shape_fuzzer-5633785895911424
src/hb-ot-color-sbix-table.hh:304:28: runtime error: negation of -2147483648 cannot be represented in type 'int'; cast to an unsigned type to negate this value to itself
```
which is reported as https://crbug.com/1050424

ebraminio:
Have applied https://github.com/harfbuzz/harfbuzz/commit/ff984ed wondering if that won't make another ubsan complain, let's see. Someone should add the flag, `/we4146` or `/SDL`, to our cmake build script so we now we don't break it in future.

PR #2072, opened 2019-12-12
Prefer _hb_glyph_info_is_unicode_mark where possible.

RazrFalcon:
We already have a shorter method that does the same.


ebraminio:
Looks good! Thanks :)

PR #2075, opened 2019-12-15
Remove unnecessary check in hb_buffer_t::set_masks.

### LINKED ISSUE #2073: Unnecessary check in hb_buffer_t::set_masks?
In [hb_buffer_t::set_masks](https://github.com/harfbuzz/harfbuzz/blob/34ed8e7218d9147a6ccd99198db594cf8f66f61b/src/hb-buffer.cc#L430) we have this check:

https://github.com/harfbuzz/harfbuzz/blob/34ed8e7218d9147a6ccd99198db594cf8f66f61b/src/hb-buffer.cc#L441

But `hb_buffer_t::set_masks` used only once in [hb_ot_shape_setup_masks](https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ot-shape.cc#L727) and we already filtered this case there:

https://github.com/harfbuzz/harfbuzz/blob/34ed8e7218d9147a6ccd99198db594cf8f66f61b/src/hb-ot-shape.cc#L740

Should we remove it and maybe add an `assert`?

PS: and probably, we should use `UINT_MAX` or `HB_FEATURE_GLOBAL_END` instead of `(unsigned int)-1`.

RazrFalcon:
Bounds are already checked by the caller.

Closes #2073

### DIFF
--- a/src/hb-buffer.cc
+++ b/src/hb-buffer.cc
@@ -438,13 +438,6 @@ hb_buffer_t::set_masks (hb_mask_t    value,
   if (!mask)
     return;
 
-  if (cluster_start == 0 && cluster_end == (unsigned int)-1) {
-    unsigned int count = len;
-    for (unsigned int i = 0; i < count; i++)
-      info[i].mask = (info[i].mask & not_mask) | value;
-    return;
-  }
-
   unsigned int count = len;
   for (unsigned int i = 0; i < count; i++)
     if (cluster_start <= info[i].cluster && info[i].cluster < cluster_end)


RazrFalcon:
Not sure what's wrong with `clang-tsan`. There is a timeout error, but this change should not cause it.

RazrFalcon:
Looks like my other pull requests are failed with the same error, so this probably `clang-tsan` error and not the code related bug.

behdad:
@ebraminio can you look into increasing timeout for tsan, other sans?

ebraminio:
Ok https://github.com/harfbuzz/harfbuzz/commit/80762cc4d0663a9ca465e94e73424e3b352de24c

PR #2076, opened 2019-12-15
Remove non-breaking spaces from comments.

RazrFalcon:
This is a cosmetic change.

`0xC2A0` (afaiu) was used, for some reasons. It's not really a problem, but Qt Creator
constantly trying to replace them with regular spaces, so I have to edit those
files separately.

### DIFF
--- a/src/hb-buffer.cc
+++ b/src/hb-buffer.cc
@@ -958,7 +958,7 @@ hb_buffer_get_direction (hb_buffer_t    *buffer)
  *
  * You can pass one of the predefined #hb_script_t values, or use
  * hb_script_from_string() or hb_script_from_iso15924_tag() to get the
- * corresponding script from an ISO 15924 script tag.
+ * corresponding script from an ISO 15924 script tag.
  *
  * Since: 0.9.2
  **/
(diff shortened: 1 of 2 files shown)


behdad:
Weird.  Wonder if @n8willis put those in.

khaledhosny:
That was me to prevent undesirable line breaks here, not a big deal, but I think QtCreator is wrong in trying to change text in semantically incompatible way like this.

RazrFalcon:
I agree. QtC can keep an original formatting, but I could not find an option to prevent spaces replacing on save. I guess a similar issue can arise in other editors.

Not sure why non-breaking spaces are needed in the source code since word wrapping is usually disabled in code editors (at least in my experience).

khaledhosny:
These are not comments, but online documentation. The HTML on harfbuzz.github.io is generated from them.

RazrFalcon:
Oh! I see. Then it makes sense, I guess.

behdad:
Maybe gtkdoc has syntax to wrap those as acronyms or otherwise prevent breaking inside?

n8willis:
> Maybe gtkdoc has syntax to wrap those as acronyms or otherwise prevent breaking inside?

I'm not aware of a feature like that in gtkdoc, but we could try the HTML -abbr- element which might have the same effect for some browsers when used inline. Prevent line-breaking these days seems to be virtually impossible without jumping through CSS hoops, though.

jfkthame:
I wouldn't expect `<abbr>` to help here. IMO this is a completely legitimate use of U+00A0, and it's unfortunate (a bug?) that Qt Creator wants to mess with it, at least when it's in comments rather than in actual code.

Can you write it as an entity `&nbsp;`, or would that just end up getting escaped?

RazrFalcon:
Looks like this is the related issue: [QTCREATORBUG-22052](https://bugreports.qt.io/browse/QTCREATORBUG-22052).

PR #2077, opened 2019-12-15
Prefer UINT_MAX instead of uint overflow.

RazrFalcon:
Also, prefer HB_FEATURE_GLOBAL_START and HB_FEATURE_GLOBAL_END.

### DIFF
--- a/src/hb-buffer.cc
+++ b/src/hb-buffer.cc
@@ -612,7 +612,7 @@ done:
 void
 hb_buffer_t::unsafe_to_break_impl (unsigned int start, unsigned int end)
 {
-  unsigned int cluster = (unsigned int) -1;
+  unsigned int cluster = UINT_MAX;
   cluster = _unsafe_to_break_find_min_cluster (info, start, end, cluster);
   _unsafe_to_break_set_mask (info, start, end, cluster);
 }
@@ -628,7 +628,7 @@ hb_buffer_t::unsafe_to_break_from_outbuffer (unsigned int start, unsigned int en
   assert (start <= out_len);
   assert (idx <= end);
 
-  unsigned int cluster = (unsigned int) -1;
+  unsigned int cluster = UINT_MAX;
   cluster = _unsafe_to_break_find_min_cluster (out_info, start, out_len, cluster);
   cluster = _unsafe_to_break_find_min_cluster (info, idx, end, cluster);
   _unsafe_to_break_set_mask (out_info, start, out_len, cluster);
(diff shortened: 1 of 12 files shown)


src/hb-buffer.cc
```diff
@@ -1738,7 +1738,7 @@ hb_buffer_add_codepoints (hb_buffer_t          *buffer,
  * @buffer: an #hb_buffer_t.
  * @source: source #hb_buffer_t.
  * @start: start index into source buffer to copy.  Use 0 to copy from start of buffer.
- * @end: end index into source buffer to copy.  Use (unsigned int) -1 to copy to end of buffer.
+ * @end: end index into source buffer to copy.  Use UINT_MAX to copy to end of buffer.
```
behdad: This should say `HB_FEATURE_GLOBAL_END`.

src/hb-common.h
```diff
@@ -423,7 +423,7 @@ typedef void (*hb_destroy_func_t) (void *user_data);
  *
  * Since: 2.0.0
  */
-#define HB_FEATURE_GLOBAL_END	((unsigned int) -1)
+#define HB_FEATURE_GLOBAL_END	UINT_MAX
```
behdad: I prefer not to rely on `UINT_MAX` in the public headers.

RazrFalcon:
Well, I don't mind. I can rollback `UINT_MAX` and keep only `HB_FEATURE_GLOBAL_*`.

Personally, I think that `UINT_MAX` is just easier to read. `(unsigned int)-1` look like a weird magic.

behdad:
Is okay to keep. Thanks. Just the two issues I pointed out in review.

RazrFalcon:
Done.

src/hb-buffer.cc
```diff
@@ -1738,7 +1738,7 @@ hb_buffer_add_codepoints (hb_buffer_t          *buffer,
  * @buffer: an #hb_buffer_t.
  * @source: source #hb_buffer_t.
  * @start: start index into source buffer to copy.  Use 0 to copy from start of buffer.
- * @end: end index into source buffer to copy.  Use (unsigned int) -1 to copy to end of buffer.
+ * @end: end index into source buffer to copy.  Use HB_FEATURE_GLOBAL_END to copy to end of buffer.
```
behdad: One more thing.  The ones in documentation I think should be written as `@HB_FEATURE_GLOBAL_END` for doc linking to work.

RazrFalcon:
Done.

PR #2080, opened 2019-12-17
Clamp max_ops to upper bound in hb-sanitize.hh

zauguin:
Alternative fix for the problem in #2079: Never make `max_ops` bigger than `HB_SANITIVE_MAX_OPS_MAX`, especially keeping the value small enough to avoid overflows.


behdad:
Maybe call `hb_unsigned_mul_overflows` in there as well?

behdad:
It still can overflow...

PR #2082, opened 2019-12-20
[coretext] Use kCTFontOpenTypeFeatureTag

khaledhosny:
Instead of trying to map OpenType features to AAT feature selectors which only works for a small subset of OpenType features, use the simpler `kCTFontOpenTypeFeatureTag` with OpenType feature tags directly.

With this change, features like `cvXX` can be enabled in coretext shaper, while they were previously ignored due to missing mapping.

This seems to work even with AAT fonts that don’t have OpenType layout tables, which suggests that CoreText is doing the mapping itself in this case.

`kCTFontOpenTypeFeatureTag` seems to have been introduced in macOS 10.10 and iOS 8.0, though, so its use is conditional on version check for now. Not sure how to check iOS version, so I left this out.

### DIFF
--- a/src/hb-coretext.cc
+++ b/src/hb-coretext.cc
@@ -475,13 +475,19 @@ _hb_coretext_shape (hb_shape_plan_t    *shape_plan,
     hb_vector_t<feature_event_t> feature_events;
     for (unsigned int i = 0; i < num_features; i++)
     {
+      active_feature_t feature;
+
+#if MAC_OS_X_VERSION_MIN_REQUIRED < 1010
       const hb_aat_feature_mapping_t * mapping = hb_aat_layout_find_feature_mapping (features[i].tag);
       if (!mapping)
 	continue;
 
-      active_feature_t feature;
       feature.rec.feature = mapping->aatFeatureType;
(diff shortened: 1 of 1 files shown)


src/hb-coretext.cc
```diff
@@ -538,6 +545,18 @@ _hb_coretext_shape (hb_shape_plan_t    *shape_plan,
 	      CFNumberCreate (kCFAllocatorDefault, kCFNumberIntType, &active_features[j].rec.feature),
 	      CFNumberCreate (kCFAllocatorDefault, kCFNumberIntType, &active_features[j].rec.setting)
 	    };
+#else
+	    char tag[4];
+	    hb_tag_to_string (active_features[j].rec.feature, tag);
```
behdad: Need to make room for a nul byte in tag and setting to zero no?  `hb_tag_to_string` just writes four bytes.

behdad: You can also do `char tag[5] = {HB_UNTAG (active_features[j].rec.feature)};`.

khaledhosny: Much simpler. Done.

PR #2085, opened 2020-01-06
[subset] Fix null pointer dereference in hvar/vvar subset

ebraminio:
Rest of the code assumes there is at least one subtable, lets return here if not.

* https://crbug.com/oss-fuzz/19827
* https://crbug.com/oss-fuzz/19847


ebraminio:
> looks fine

Thank you :)

PR #2086, opened 2020-01-07
fix for cmap4 subsetting: largest character code allowed is 0xFFFF

qxliu76:


qxliu76:
> Can you add a test which covers this case.

test added. Also, add a fix for OS_2 table: according to spec, "usFirstCharIndex" and "usLastCharIndex" field should be set to 0xFFFF if the index value is a supplementary character.

src/hb-ot-os2-table.hh
```diff
@@ -216,8 +216,10 @@ struct OS2
 					  uint16_t *min_cp, /* OUT */
 					  uint16_t *max_cp  /* OUT */)
   {
-    *min_cp = codepoints->get_min ();
-    *max_cp = codepoints->get_max ();
+    hb_codepoint_t min_codepoint = codepoints->get_min ();
+    hb_codepoint_t max_codepoint = codepoints->get_max ();
+    *min_cp = min_codepoint > 0xFFFF ? 0xFFFF : min_codepoint;
```
garretrieger: could also write this as:
*min_cp = hb_min(0xFFFF, codepoints->get_min());

Likewise for max_cp

Issue #2060, opened 2019-11-25
[subset] Fails to subset GPOS in variable font


blueshade7:
`hb-subset` fails with the attached CFF2 variable font if GPOS table included as:
```
hb-subset --text=A --drop-tables-=GPOS --output-file=out.otf AdobeVFPrototype.otf
```
`out.otf` is not created. If `GPOS` table is not included (default), `out.otf` is created.

[AdobeVFPrototype.otf.zip](https://github.com/harfbuzz/harfbuzz/files/3888753/AdobeVFPrototype.otf.zip)

garretrieger:
Running the subsetter in debug mode. It looks like sanitization of the GPOS table is failing. I'll have to dig in a bit deeper to see if that's a bug, or the table is in fact invalid in some way.

garretrieger:
More specifically sanitization is failing for "struct HintingDevice" which I believe corresponds with the device table (https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#device-and-variationindex-tables).

blueshade7:
I see what's wrong. An offset to a device table within a PairPosFormat1 lookup must be based on the beginning of its "immediate parent", i.e., PairSet table, according to OT spec, but sanitizer uses its grandparent, PairPosFormat1 table as the base. Correcting this generated a bad subset font however. I suspect serialization has the same issue. Looking further.

blueshade7:
Can't find code to serialize device tables. Seems unimplemented.

blueshade7:
created a branch `subset-gpos-var` with GPOS device table fixes

blueshade7:
Now a subset font can be generated with the change in the branch. fontTools `ttx` works fine with the output but doesn't pass `ots-sanitize` seemingly due to null lookup subtable offset:
https://github.com/harfbuzz/harfbuzz/blob/8ed46c3678cd4518efe5c9907bb9d22e6161bbd5/src/hb-ot-layout-common.hh#L855-L856
This prevents adding a test case. I guess I will just create a PR for what I have now.

garretrieger:
We have a few PR's out that will add support for dropping null lookups (ex: https://github.com/harfbuzz/harfbuzz/pull/2043) hoping to get those merged in the next few days.

blueshade7:
Subsetting variation store in GDEF hasn't been implemented yet, but subsetting var GPOS works now, so let me close this.

PR #2087, opened 2020-01-08
[subset] fixed GPOS device table sanitize & serialize

blueshade7:
Fixed the following sub-issues found in #2060 :

1. Offsets to device tables in PairPosFormat1 were based on the lookup table, not on the PairSet table (see `PairSet::sanitize_closure_t` in `PairPosFormat1::sanitize`).
2. Device tables were not serialized when their offsets were listed in value records in SinglePosFormat1, SinglePosFormat2, PairPosFormat1, or PairPosFormat2.

Since variation data in the GPOS table are stored in a variant of a device table, above issues were preventing a GPOS table with variation data in a variable font to be subsetted correctly.

### DIFF
--- a/src/hb-ot-layout-gpos-table.hh
+++ b/src/hb-ot-layout-gpos-table.hh
@@ -155,6 +155,22 @@ struct ValueFormat : HBUINT16
     return ret;
   }
 
+  void serialize_copy (hb_serialize_context_t *c, const void *src_base, const void *dst_base, const Value *values) const
+  {
+    unsigned int format = *this;
+    if (!format) return;
+
+    if (format & xPlacement) c->copy (*values++);
+    if (format & yPlacement) c->copy (*values++);
+    if (format & xAdvance)   c->copy (*values++);
+    if (format & yAdvance)   c->copy (*values++);
+
+    if (format & xPlaDevice) copy_device (c, src_base, dst_base, values++);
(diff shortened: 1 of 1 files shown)


garretrieger:
Looks good to me as well. Though would also like to see tests. If you don't mind waiting for us to get the lookup subsetting PR's merged, those will unblock you on adding tests.

blueshade7:
@garretrieger You mean PR 2042? Yes I can wait for it and then add test.

garretrieger:
Yeah, that's the one.

blueshade7:
Actually, variation store for GPOS are stored in GDEF, currently dropped. For GPOS to function properly with VF, GDEF table needs to serialize variation store. If we are to subset it, outer/inner indices in device tables in GPOS need to be remapped. More work needed for a test to output the correct result.

garretrieger:
Ah good point, I was hoping to start work on implementing GDEF subsetting in the near future.

blueshade7:
In order to subset varStore in GDEF, outer/inner indices need collected from GPOS, JSTF, and GDEF itself. At least, PairPosFormat1 fix in this PR needs merged first. A sort of chicken-egg problem.
Shall we just merge this without adding a test?

garretrieger:
I'm fine with merging this now if tests will be added in a later PR. FYI, we just got all of the script/lookup/feature subsetting PR's merged now.

PR #2090, opened 2020-01-09
[subset/hvar] Don't insert on null sets

ebraminio:
fixes https://crbug.com/oss-fuzz/19978

Don't like its complication thus issues around, guess was needed anyway but feel free to look if its flow can be refactored.

@blueshade7 wdyt about the fix, is this in the right place? If not, feel free to submit your own fix or let me know what you think, thanks :)


ebraminio:
In favor #2091

PR #2091, opened 2020-01-10
fixes oss-fuzz 19978: Null-dereference READ

blueshade7:


ebraminio:
Thank you :)

behdad:
This is wrong.  Why is `add`ing to NULL set causing trouble to begin with?  That's where the problem is and needs to be fixed.

PR #2094, opened 2020-01-11
Fix collect lookups logic

ebraminio:
Fixes https://crbug.com/oss-fuzz/20021 and https://crbug.com/oss-fuzz/20022


ebraminio:
@qxliu76 does the change look good? The fact it is not test covered so my change doesn't change anything makes it worrisome :/

ebraminio:
Ok am confident now that is the fix as "Offset to a feature table substitution table, from beginning of the FeatureVariations table." from https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2

qxliu76:
thank you for the fix!

ebraminio:
No problem, please make sure some test will cover it however. Thanks

PR #2100, opened 2020-01-13
Fix duplicate check in hb_ot_rotate chars

### LINKED ISSUE #2099: Duplicate c->font->has_glyph (codepoint) check in hb_rotate_chars
Looks like there's a duplicate line after the condition was rewritten.

drott:
Fixes #2099. Fix indentation in mirroring section as well.

### DIFF
--- a/src/hb-ot-shape.cc
+++ b/src/hb-ot-shape.cc
@@ -652,7 +652,6 @@ hb_ot_rotate_chars (const hb_ot_shape_context_t *c)
   {
     for (unsigned int i = 0; i < count; i++) {
       hb_codepoint_t codepoint = hb_vert_char_for (info[i].codepoint);
-      if (c->font->has_glyph (codepoint))
       if (unlikely (codepoint != info[i].codepoint && c->font->has_glyph (codepoint)))
 	info[i].codepoint = codepoint;
     }


src/hb-ot-shape.cc
```diff
@@ -642,19 +642,18 @@ hb_ot_rotate_chars (const hb_ot_shape_context_t *c)
     for (unsigned int i = 0; i < count; i++) {
       hb_codepoint_t codepoint = unicode->mirroring (info[i].codepoint);
       if (unlikely (codepoint != info[i].codepoint && c->font->has_glyph (codepoint)))
-	info[i].codepoint = codepoint;
+        info[i].codepoint = codepoint;
       else
-	info[i].mask |= rtlm_mask;
+        info[i].mask |= rtlm_mask;
```
ebraminio: Please revert turning tab to spaces, thanks :)

drott: It seems to me that the rest of the file is space indented. Or was this meant to be ironic?

ebraminio: Rest of the file is mostly consistent on HarfBuzz's current default style that is using two spaces for indention but turning 8 of them to tab, not that fan of the style personally but let preserve it for now :)

drott: 🤯Wow, I've never seen such a convention, hadn't even noticed that in HarfBuzz. What's the origin of that convention or why would it ever make sense? :) - but yes, updated the CL.

ebraminio: @behdad should know :)

PR #2101, opened 2020-01-13
[subset] Add support for sbix

ckitagawa-work:
This PR adds support for subsetting sbix fonts. The testcases are based on those used in fonttools.

[Spec](https://docs.microsoft.com/en-us/typography/opentype/spec/sbix)

I tried to use iterators, and existing subsetting utilities like serialize_subset() and subset_offset_array(), but found issues with sbix since the spec calls for arrays of offsets that need to be ordered. Perhaps there is an opportunity to add something that supports this to hb-serialize? (CBLC/CBDT also requires this).

Any and all feedback is appreciated!

_Context:_ I'm hoping/working to add subset support for all fonts that use images as they tend to be quite large (>> 10 MB). The motivation for this is a Chromium related feature that will result in saving static pages and I'd like to only save the font data for glyphs used in the page to save on disk space.


ebraminio:
Looks very nice! Please remove those hex files from history also. Thanks :)

src/hb-ot-color-sbix-table.hh
```diff
@@ -116,6 +146,38 @@ struct SBIXStrike
...
+      }
+      unsigned int delta = imageOffsetsZ[old_gid + 1] - imageOffsetsZ[old_gid];
+      unsigned int glyph_data_length = delta - SBIXGlyph::min_size;
+      if (!(this + imageOffsetsZ[old_gid]).subset(c, glyph_data_length))
+        return_trace (false);
+      out->imageOffsetsZ[new_gid] = head;
+      head += delta;
+    }
+    out->imageOffsetsZ[num_output_glyphs] = head;
+    return_trace (true);
```
garretrieger: subset() should return false, if no data should be written for this sub object. So for example if due to the subset of glyphs this particular strike contains no images, then this entire strike can be left out so the subset() method should return false.

src/hb-ot-color-sbix-table.hh
```diff
@@ -275,6 +337,70 @@ struct sbix
...
+  }
+
+  bool serialize_strike_offsets (hb_subset_context_t *c,
+                                 const void *dst_base) const
+  {
+    TRACE_SERIALIZE (this);
+
+    auto *out = c->serializer->start_embed<LOffsetLArrayOf<SBIXStrike>> ();
+    if (unlikely (!out)) return_trace (false);
+    if (unlikely (!c->serializer->allocate_size<HBUINT32> (HBUINT32::static_size))) return_trace (false);
```
garretrieger: I believe you should be able to use serializer->extend_min(out) which should extend the buffer enough to fit the length parameter.

test/subset/data/tests/sbix.tests
```diff
@@ -0,0 +1,13 @@
+FONTS:
+sbix.ttf
+
+PROFILES:
+default.txt
+drop-hints.txt
+drop-hints-retain-gids.txt
+retain-gids.txt
+name-ids.txt
```
garretrieger: You can leave out name-ids.txt it should have no effect on sbix subsetting.

test/subset/data/tests/sbix.tests
```diff
@@ -0,0 +1,13 @@
...
+
+SUBSETS:
+X
+Y
```
garretrieger: Can you add XY as well.

src/hb-ot-color-sbix-table.hh
```diff
@@ -40,6 +41,32 @@ namespace OT {
...
+    if (unlikely (!c->extend (*this, data_length))) return_trace (false);
+
+    xOffset = x_offset;
+    yOffset = y_offset;
+    graphicType = graphic_type;
+    memcpy(&data, src_data, data_length);
+    return_trace (true);
+  }
+
+  bool subset (hb_subset_context_t *c, unsigned int data_length) const {
```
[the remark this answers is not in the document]
ckitagawa-work: Done. Not sure if there is a simpler way of doing this copy since it contains an UnsizedArray?

src/hb-ot-color-sbix-table.hh
ckitagawa-work: Done. Should I handle a failure to copy a glyph in that case? My understanding is that the serializer will be in an error state at that point, but should I still try to do cleanup?

src/hb-ot-color-sbix-table.hh
ckitagawa-work: Done.

test/subset/data/tests/sbix.tests
ckitagawa-work: Done.

test/subset/data/tests/sbix.tests
ckitagawa-work: Done.

src/hb-ot-color-sbix-table.hh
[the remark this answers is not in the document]
garretrieger: Yeah, you can use the copy(serialize context, count) method of unsized array instead of the memcpy.

src/hb-ot-color-sbix-table.hh
garretrieger: return false after a failure to copy is reasonable. As you mentioned the serializer will be in an error state so everything will be discarded once you exit the top level subset() call (and typically the operation will be retried with a larger buffer).

garretrieger:
One last thought, can you add the sbix test file into the fuzzer seeds directory (test/fuzzing/fonts/). This will help give the fuzzer a head start on pfuzzing sbix subsetting.

src/hb-ot-color-sbix-table.hh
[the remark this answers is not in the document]
ckitagawa-work: Done.

src/hb-ot-color-sbix-table.hh
ckitagawa-work: Ack.

ebraminio:
Guess using prefixes instead `/sbix/` is enough here otherwise we have to add the new path https://github.com/google/oss-fuzz/blob/master/projects/harfbuzz/build.sh#L41

> Do I need to add any kind of unittests, or are the subset tests sufficient?

If no new API introduced current usual tests are enough usually.

garretrieger:
You can add unit tests for the subsetting of sbix under ./test/api/. There's some examples from other tables in there. It's not strictly required, but they are nice to have. For example here's the cmap unit tests: https://github.com/harfbuzz/harfbuzz/blob/master/test/api/test-subset-cmap.c

The unit tests differ from the integration tests in that they typically check the binary result of the subsetting operation instead of checking that TTX interprets the files the same.

ckitagawa-work:
garretrieger@ can you take a final pass and merge this? I don't have write access to the repo. Thanks!

I've added api tests and improved coverage by modifying sbix.ttf so that some strikes are missing X or Y to ensure that empty strikes get dropped. I also removed sbix from the default drop table list although I can re-add it if that was desiredable?

garretrieger:
Removing sbix from default drop table sgtm.

PR #2114, opened 2020-01-19
[subset] remove empty lookup subtables

blueshade7:
Implemented _TODO Drop empty ones…_ comment in `Lookup::subset` in hb-ot-layout-common.h.
For this, added a variant of `subset_offset_array` taking an extra argument passed to `serialize_subset`.

Added `intersects` to GPOS `PosLookupSubtable` and GSUB `SubstLookupSubTable` to determine which subtables are empty.

Added an API test `test-subset-gpos.c` initially with only one test case to subset GPOS LookupType 2 (Pair) in Roboto-Regular containing two sub-tables PairPos1 and PairPos2 to one sub-table PairPos2.


blueshade7:
This PR should fix the remaining sub-issue of #2060.

PR #2117, opened 2020-01-21
[subset] Fix sbix subsetting fuzzer bug

ckitagawa-work:
Root cause: reading past end of sbix file.

The issue was due to a misunderstanding I had that the table was fully sanitized. However, it is only partly sanitized, therefore a more rigorous check that offsets are within range of the table was required.


ckitagawa-work:
The failing check looks to be msan failing due to a timeout, it is likely a flake. I don't have permission to retry though.

src/hb-ot-color-sbix-table.hh
```diff
@@ -133,7 +133,7 @@ struct SBIXStrike
     return hb_blob_create_sub_blob (sbix_blob, glyph_offset, glyph_length);
   }
 
-  bool subset (hb_subset_context_t *c) const
+  bool subset (hb_subset_context_t *c, unsigned int sbix_len, unsigned int strike_offset) const
```
garretrieger: In this method you only ever use the value sbix_len - strike_offset. So I recommend that you just pass one param which is that value. Could be called something like 'available_len'

src/hb-ot-color-sbix-table.hh
```diff
@@ -334,7 +336,8 @@ struct sbix
   bool add_strike (hb_subset_context_t *c,
                    const void *dst_base,
                    LOffsetTo<SBIXStrike>* o,
```
garretrieger: Just notice that the only thing that happens with the 'o' parameter in this call is that it's set to 0. Recommend that you remove it, and do the set to zero in the serialize_strike_offsets method instead.

src/hb-ot-color-sbix-table.hh
ckitagawa-work: Done.

src/hb-ot-color-sbix-table.hh
ckitagawa-work: Done.

ckitagawa-work: Done.

ckitagawa-work:
PTAL and merge. The failing tests appear to all be timeouts on fuzzers which I think are flakes as they are not the new testcases.

garretrieger:
The fuzzer timeout's a known issue. On my list too look at it. Looks good, will merge.

PR #2118, opened 2020-01-22
[subset] remove empty cmap4 subtable

qxliu76:


qxliu76:
> Could you also add an integration test case that covers the situation where you have a cmap 12 but no cmap 4 because the only characters in the subset are past 0xFFFF? You should be able to add one to the japanese tests.

Done.

PR #2119, opened 2020-01-22
[subset] Add "--name-languages" and "--name-legacy" options

qxliu76:
make name table subsetting consistent with fontTools


src/hb-ot-name-table.hh
```diff
@@ -108,6 +108,17 @@ struct NameRecord
     return_trace (out);
   }
 
+  bool isUnicode () const
+  {
+    unsigned int p = platformID;
+    unsigned int e = encodingID;
+    
+    if (p == 0 ||
+        (p == 3 && (e == 0 || e == 1 || e == 10))) return true;
```
garretrieger: can simplify this to:

return (p == 0 ||
        (p == 3 && (e == 0 || e == 1 || e == 10)));

src/hb-ot-name-table.hh
```diff
@@ -209,7 +220,17 @@ struct name
 
     auto it =
     + nameRecordZ.as_array (count)
-    | hb_filter (c->plan->name_ids, &NameRecord::nameID)
+    | hb_filter ([&] (const NameRecord& namerecord)
```
garretrieger: I think you can do this as three separate filters, without needing to define a lambda:
...
| hb_filter (c->plan->name_ids, &NameRecord::nameID)
| hb_filter (c->plan->name_languages, &NameRecord::languageID);

if (!c->plan->name_legacy) {
  it = it | hb_filter(&NameRecord::isUnicode);
}

src/hb-subset-input.cc
```diff
@@ -204,3 +215,29 @@ hb_subset_input_get_retain_gids (hb_subset_input_t *subset_input)
...
+HB_EXTERN hb_bool_t
+hb_subset_input_get_name_legacy (hb_subset_input_t *subset_input)
+{
+  return subset_input->name_legacy;
+}
+
+HB_EXTERN void
+hb_subset_input_set_obfuscate_names (hb_subset_input_t *subset_input,
+				     hb_bool_t obfuscate_names)
+{
```
garretrieger: I don't see any code in hb-ot-name-table.hh that uses obfuscate_names. If this PR isn't adding the obfuscate names implementation then I'd leave it out of the subset api and hb-subset util options for now.

ebraminio:
Had to revert this on e171beeb5f3f26dee84373b4db223512675f5837 as it was making tests fail, please reapply when bots are OK, thanks :)

qxliu76:
> Had to revert this on [e171bee](https://github.com/harfbuzz/harfbuzz/commit/e171beeb5f3f26dee84373b4db223512675f5837) as it was making tests fail, please reapply when bots are OK, thanks :)

Could you point me which test failed? My local tests are all fine

ebraminio:
Hmm, maybe I've rushed a bit, on https://github.com/harfbuzz/harfbuzz/commit/fd85818b5b85fb0ad2db71c60e94ca0689d38bca multiple bots have failed which the parent isn't, e.g. https://circleci.com/gh/harfbuzz/harfbuzz/122671 search for "ERROR: ttx for expected and actual does not match.". If it isn't because of this change, guess fonttools has changed their result, this particular change is ran when fontTools is installed and `ttx` is available.

ebraminio:
Can you please quickly resubmit this in a separate PR? It will be the best to see if the failures was because this change. Thanks

PR #2120, opened 2020-01-22
[subset] refixed PairPos1 PR #2087

blueshade7:
My PR#2087 supposed to fix GPOS PairPosFormat1 sanitizer was incorrect in that it was using the beginning of the pairSet _offset_ array as the base for device tables (blush). Instead the beginning of each pairSet table containing device table offsets should be used as the base.

This sanitizer bug caused oss-fuzz 20211 and fixed here. Added its test file to fuzzing/fonts.

PairPos1 subsetting was still not working, because of another bug & fixed in `PairSet::subset()`.

Added a test case for subsetting GPOS table containing device tables using a variable font AdobeVFPrototype.otf to `test-subset-gpos`.


blueshade7:
two bot failures are both timeouts with fuzzed fonts

PR #2121, opened 2020-01-22
[subset] Add CBLC/CBDT support

ckitagawa-work:
This PR adds basic CBLC/CBDT subset support for

- index formats 1 and 3
- image formats 17-19

As follow-up work I may expand harfbuzz to support index formats 2, 4 and 5 in addition to other image formats (although I haven't seen many examples of them in use).

Suggestions for additional tests and font sources for which to test are welcome as fonttools only had a single small example with only one index and image combination and one BitmapSizeTable, which is somewhat lacking for test coverage.

See specs at:
[1] [CBLC](https://docs.microsoft.com/en-us/typography/opentype/spec/cblc)
[2] [CBDT](https://docs.microsoft.com/en-us/typography/opentype/spec/cbdt)
[3] [EBLC](https://docs.microsoft.com/en-us/typography/opentype/spec/eblc)
[4] [EBDT](https://docs.microsoft.com/en-us/typography/opentype/spec/ebdt)


ebraminio:
Great! We exposed ot-color related APIs for all the tables in one release, I wonder do you think you would be able to add COLR and SVG support also? They should be easier to do also I think. Thanks

khaledhosny:
`COLR` is easy, `SVG` is trickier https://github.com/fonttools/fonttools/issues/534.

src/hb-ot-color-cbdt-table.hh
```diff
@@ -43,6 +43,15 @@
 
 namespace OT {
 
+namespace CBDT_internal {
+
+// Helper function internal to CBDT.
+bool copy_data_to_cbdt (hb_vector_t<char> *cbdt,
```
ebraminio: Should be marked with HB_INTERNAL otherwise it gets public visiblity in Unix

ebraminio:
> COLR is easy, SVG is trickier fonttools/fonttools#534.

Oh, thank you, so better to preserve glyph ids for such font, filed #2122

src/hb-ot-color-cbdt-table.hh
```diff
@@ -342,6 +670,23 @@ struct CBLC
...
+                                           new_cbdt->arrayZ,
+                                           free);
+    new_cbdt->init ();  // Leak arrayZ to the blob.
+    bool ret = c->plan->add_table (HB_OT_TAG_CBDT, cbdt_blob);
+    hb_blob_destroy (cbdt_blob);
+    return ret;
+  }
+
+  // Implemented in cc file as it depends on definition of CBDT.
+  bool subset (hb_subset_context_t *c) const;
```
ebraminio: Add one HB_INTERNAL here also :)

ckitagawa-work:
> Great! We exposed ot-color related APIs for all the tables in one release, I wonder do you think you would be able to add COLR and SVG support also? They should be easier to do also I think. Thanks

I'm most concerned about subsetting tables that tend to be large (i.e. fonts that become O(MB) in size). COLR and SVG are typically smaller than the other color tables so they are lower priority for me. COLR seems easy enough so it will likely be a quick followup. I'm concerned about the possible complexities of SVG, but I'll get to it at some point hopefully.

src/hb-ot-color-cbdt-table.hh
```diff
@@ -243,6 +435,101 @@ struct IndexSubtableArray
...
+
+    for (unsigned int i = 0; i < lookup.length; i++)
+    {
+      // On a match in formats we can extend the old table.
+      const IndexSubtable *old_subtable = lookup[i].second->get_subtable (this);
+      auto* new_header = curr_subtable->get_header ();
+      auto* old_header = old_subtable->get_header ();
+      if (!success ||
+          new_header->indexFormat != old_header->indexFormat ||
+          new_header->imageFormat != old_header->imageFormat)
```
garretrieger: Should also check if the current gid (new) is adjacent to the last processed gid (new). If it's not, end the sub table since sub tables should not have gaps in them.

src/hb-ot-color-cbdt-table.hh
```diff
@@ -243,6 +435,101 @@ struct IndexSubtableArray
...
+                                                            &curr_subtable)))
+          return_trace (false);
+        if (unlikely (!curr_record || !curr_subtable)) return_trace (false);
+
+        *num_tables += 1;
+        num_glyphs = 0;
+      }
+      hb_codepoint_t new_gid = lookup[i].first;
+      unsigned int num_missing = curr_record->add_glyph_for_subset (new_gid);
+      if (unlikely (!curr_subtable->fill_missing_glyphs (c->serializer, cbdt->length, num_missing, size, &num_glyphs)))
```
garretrieger: If you implement the comment above then there shouldn't be any gaps and this becomes unnecessary.

garretrieger:
For test fonts, you could use Noto Color Emoji (https://www.google.com/get/noto/help/emoji/) I believe it has a CBLC/CBDT table in it (likely want to add a subset of it since the whole font is quite big).

twardoch:
SVG subsetting would indeed be complex. Keep in mind that the relationship between glyphs and SVG documents in the SVG table in kind of n:m. The `g` elements are kind of like elements in FL7, or like CFF subroutines, and they can be reused in a multitude of ways.

behdad:
I just want to note that these two tables are extremely tricky.  Please be super diligent about them. :)  And thanks for the work!

src/hb-ot-color-cbdt-table.hh
garretrieger: Ah good catch, I would do whatever fontTools currently does in this case then. Since we treat that as the gold standard for what the result should be.

ckitagawa-work: If the gap exists before subsetting and the glyph for that gap is kept then a space is kept to respect the font's design. fontTools tries to respect the format of existing tables first and foremost so we still need to support gaps.

ckitagawa-work:
I've added more subsets tests for scenarios with gaps, multiple BitmapSizeTables, and index format 3 and fixed any issues I found associated with this.

The API test is now failing due to the order in which links are resolved causing the offsets to be different than expected. This is due to the order in which links get resolved leading to the IndexSubtables to be added in reverse (due to the stacking order) a similar problem will also occur with the order when there are multiple BitmapSizeTables. If we care about the API tests I believe I will need to find a workaround like I did with the sbix table although this will make the code more confusing and less readable...

Options
1. Add a different order for resolving links? I don't know how feasible this is.
2. Work around like sbix.
3. Ignore the bytewise compatibility as the table is valid according to the spec (I'm not sure if this is acceptable)

Any suggestions are welcome.

Example:

```
Harfbuzz CBLC

Version             00 03 00 00
# BitmapSizeTables  00 00 00 01

BitmapSizeTables
[0]
Offset          00 00 00 38
Subtable Size   00 00 00 3c
# of Subtables  00 00 00 02
ColorRef        00 00 00 00
Hori            65 e5 88 00 00 00 00 00 00 00 00 00
Vert            65 e5 88 00 00 00 00 00 00 00 00 00
StartGid        00 01
EndGid          00 05
PPEM            6d 6d
BitDepth        20
Flags           01

IndexSubtableArray
[0]
FirstGid  00 01
Last      00 03
Offset    00 00 00 24
[1]
FirstGid  00 04
Last      00 05
Offset    00 00 00 10

IndexSubtables
[1]
IndexFormat  00 01
ImageFormat  00 11
ImageOffset  00 00 14 bb
Offsets      00 00 00 00
             00 00 08 52
End          00 00 0e a1
[0]
IndexFormat  00 01
ImageFormat  00 11
ImageOffset  00 00 00 04
Offsets      00 00 00 00
             00 00 03 a6
	     00 00 07 50
End          00 00 14 b7

===========================================================

Fonttools CBLC

Version             00 03 00 00
# BitmapSizeTables  00 00 00 01

BitmapSizeTables
[0]
Offset          00 00 00 38
Subtable Size   00 00 00 3c
# of Subtables  00 00 00 02
ColorRef        00 00 00 00
Hori            65 e5 88 00 00 00 00 00 00 00 00 00
Vert            65 e5 88 00 00 00 00 00 00 00 00 00
StartGid        00 01
EndGid          00 05
PPEM            6d 6d
BitDepth        20
Flags           01

IndexSubtableArray
[0]
FirstGid  00 01
Last      00 03
Offset    00 00 00 10
[1]
FirstGid  00 04
Last      00 05
Offset    00 00 00 28

IndexSubtables
[0]
IndexFormat  00 01
ImageFormat  00 11
ImageOffset  00 00 00 04
Offsets      00 00 00 00
             00 00 03 a6
	     00 00 07 50
End          00 00 14 b7
[1]
IndexFormat  00 01
ImageFormat  00 11
ImageOffset  00 00 14 bb
Offsets      00 00 00 00
             00 00 08 52
End          00 00 0e a1
```

garretrieger:
We don't need to maintain bytewise compatibility with fontTools, only requirement is that the TTX interpretation of the result is the same. We have subsetting for other tables where we don't have bytewise compatibility. In those cases we usually hand craft the binary output for the api test (though those have been much simpler tables then these one).

garretrieger:
It would be fine to use the harfbuzz produced binary in the api test as long as you have verified it's correct manually.

ckitagawa-work:
Thanks Garret! I've fixed up the API tests to pass using harfbuzz produced binary files where necessary and verified they are correct via hexdumps. This was required for the case of multiple `BitmapSizeTable`s and Index Format 3 (due to a FontTools bug https://github.com/fonttools/fonttools/issues/1817)

behdad:
Why does this change have cmap changes in it?  Those need separate discussion and fixing if any.

src/hb-ot-color-cbdt-table.hh
behdad: > I haven't seen a case where gaps occur so maybe we can just drop this though?

That's not a good way to write defensive code. :)

ckitagawa-work:
@behdad NotoColorEmoji.ttf uses cmap Format 14 PlatformID 0 EncodingID 5 and this gets dropped by our current cmap subsetting. I'll move that support to another PR.

ebraminio:
Please adopt the PR with 5b436033851960ea3638fbf882100b8c04d6d5a3, sorry for the trouble, thanks :) Like to have `head` as a separate PR also, https://github.com/harfbuzz/harfbuzz/blob/1896ba2be80180fe7f1c563ef125c9e3ffb0fa18/src/hb-ot-cff1-table.hh#L1383 and https://github.com/harfbuzz/harfbuzz/blob/1896ba2be80180fe7f1c563ef125c9e3ffb0fa18/src/hb-ot-cff2-table.hh#L546 also does something about head, maybe those also be merged with this.

ckitagawa-work:
I'll split the `head` change out to another PR then. I'll adopt the new base commit once I'm done fuzzing locally.

ebraminio:
>  I'll adopt the new base commit once I'm done fuzzing locally.

Guess this will be useful :)
```
clang -fsanitize=address,fuzzer test/fuzzing/hb-shape-fuzzer.cc src/harfbuzz.cc -Isrc/ -DHB_NO_MT -O3
```

ckitagawa-work:
@ebraminio I don't think this should have been merged yet.

ebraminio:
Ran ASan fuzzer locally and merged it manually.

> I don't think this should have been merged yet.

Oh, am sorry, what do you suggest, reverting whole thing?

ckitagawa-work:
> Oh, am sorry, what do you suggest, reverting whole thing?

I'll put up a PR to undo the `cmap` change that shouldn't have been made.

ebraminio:
Oh, now I see, thanks!

ebraminio:
Sorry for the rush, wanted to apppy some cleanups myself that's why rushed
into merging, and thanks for the fixing it :) am thinking was nice if
GitHub had system of dependent patches like Gerrit, having that may could
avoid aome part of this merge error.

About SVG in OT fonts, am thinking we can still have some best effort
subsetting SVG fonts for retain gid cases removing not referenced blobs at
least, or, maybe not.

ckitagawa-work:
> am thinking was nice if GitHub had system of dependent patches like Gerrit, having that may could
avoid aome part of this merge error.

Agreed, change dependencies are very nice.

> About SVG in OT fonts, am thinking we can still have some best effort
subsetting SVG fonts for retain gid cases removing not referenced blobs at
least, or, maybe not.

Sounds potentially reasonable. It is definitely better to be conservative in that case, I'll look into the SVG table more to see if that is reasonable... It will be harder to test though as I don' think FontTools supports subsetting SVG so there will be no "golden" to compare with.

PR #2125, opened 2020-01-23
[subset] Fix UBSAN issue in sbix

ckitagawa-work:
Fuzzer found a UBSAN assert failure if the sbix table had an 0 offset or offset beyond the file for a strike then we wouldn't push() on the serializer. We would then try to pop_discard() and there was no previous state to revert to causing an assert failure. The fix is to simply to push() before doing the check so that the cleanup behavior is the same regardless of if the offset is valid.


src/hb-ot-color-sbix-table.hh
```diff
@@ -337,12 +337,13 @@ struct sbix
                    const void *dst_base,
                    unsigned int i,
                    unsigned int sbix_len) const {
+    // Push first so reverting doesn't fail.
+    c->serializer->push ();
```
ebraminio: Guess better to be called in the caller, serialize_strike_offsets, so number of push and pop there can be balanced?

ckitagawa-work: Good idea. Done.

PR #2128, opened 2020-01-24
[subset] sbix fix missed offset is_null() check

ckitagawa-work:
If data is ill formed it is possible `imageOffsetsZ[]` is null in which case if we try to do a subset `(this+imageOffsetsZ[])` is invalid leading to a possible issue. Fix this by checking if the offsets are null before trusting/using them.


ckitagawa-work:
The japanese subset failures appear to be in master?

ebraminio:
Thanks Calder :)

Garret, please review this as I rushed a little here.

garretrieger:
Looks good to me.

PR #2129, opened 2020-01-24
Only prefer AAT morx over GSUB for horizontal runs

### LINKED ISSUE #2124: Prefering AAT over GSUB breaks upright layout for Hiragino
The [change to always prefer AAT over GSUB](https://github.com/harfbuzz/harfbuzz/commit/0414341ac576c4403231f507e216c7716c8f0828) triggered issues with upright vertical layout for the Hiragino fonts. 

On Mac OS 10.10 through 10.12 Hiragino Mincho & Kaku seem to have both, mort/morx and GSUB tables. Their AAT table does have the vertical substitution feature, but there it behaves rather like `vrt2` than `vert`, which means, it rotates Latin glyphs and digits sideways. This is not what's desired. 

But because AAT does not have a distinction between `vert`-style and `vrt2` style rotations, i.e. desired punctuation rotation and "Latin-sideways" rotation are not split, we would likely be better off using GSUB in these cases.

I would propose to experimentally partially revert 0414341ac576c4403231f507e216c7716c8f0828 in the sense of only preferring AAT over GSUB when the direction is horizontal. 

If that's too broad (what's your opinion @jfkthame ?) - then I'd be glad if we could similar to as before, fingerprint the fonts in question, and keep preferring GSUB for those that have too broad AAT vertical substitution and rotate Latin and digits sideways.

On Mac OS 10.13 and later, it seems that OpenType `vert` is preferred over AAT style vertical substitutions and some of the vertical fonts do not have mort/morx anymore.

More details about the issues after the roll in: https://bugs.chromium.org/p/chromium/issues/detail?id=1040017

drott:
Fixes #2124.

### DIFF
--- a/src/hb-ot-shape.cc
+++ b/src/hb-ot-shape.cc
@@ -70,7 +70,8 @@ hb_ot_shape_planner_t::hb_ot_shape_planner_t (hb_face_t                     *fac
 						map (face, props),
 						aat_map (face, props)
 #ifndef HB_NO_AAT_SHAPE
-						, apply_morx (hb_aat_layout_has_substitution (face))
+                        /* https://github.com/harfbuzz/harfbuzz/issues/2124 */
+						, apply_morx (HB_DIRECTION_IS_HORIZONTAL(props->direction) && hb_aat_layout_has_substitution (face))
 #endif
 {
   shaper = hb_ot_shape_complex_categorize (this);


ebraminio:
Actually this isn't like what we had before as it disables morx for every vertical run but we should try `morx` if really no GSUB exists. WDYT @kojiishi @jfkthame? Will try to upload a fix myself also

PR #2130, opened 2020-01-24
Run morx if run is horizontal or GSUB doesn't exist

ebraminio:
Related to #2129 and #2124

Dominik, can you check if this still works on those fonts?

### DIFF
--- a/src/hb-ot-shape.cc
+++ b/src/hb-ot-shape.cc
@@ -48,6 +48,16 @@
 #include "hb-aat-layout.hh"
 
 
+#ifndef HB_NO_AAT_SHAPE
+static inline bool
+_hb_apply_morx (hb_face_t *face, const hb_segment_properties_t *props)
+{
+  /* https://github.com/harfbuzz/harfbuzz/issues/2124 */
+  return hb_aat_layout_has_substitution (face) &&
+	 (HB_DIRECTION_IS_HORIZONTAL (props->direction) || !hb_ot_layout_has_substitution (face));
+}
+#endif
+
 /**
  * SECTION:hb-ot-shape
  * @title: hb-ot-shape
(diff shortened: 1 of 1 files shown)


src/hb-ot-shape.cc
```diff
@@ -48,6 +48,16 @@
 #include "hb-aat-layout.hh"
 
 
+#ifndef HB_NO_AAT_SHAPE
+static inline bool
+_hb_apply_morx (hb_face_t *face, const hb_segment_properties_t *props)
+{
+  /* https://github.com/harfbuzz/harfbuzz/issues/2124 */
+  return hb_aat_layout_has_substitution (face) &&
+	 (HB_DIRECTION_IS_HORIZONTAL (props->direction) || !hb_ot_layout_has_substitution (face));
```
ebraminio: I think semantically `hb_ot_layout_has_substitution` should itself check `hb_aat_layout_has_substitution` also but it isn't the case for now so we should check this in our tests somehow, Dominik, can you provide a hb-shape arguments to trigger this? We have macOS bots that have access to its font so no issue here.

ebraminio:
Am merging but please have a look. Thanks

src/hb-ot-shape.cc
drott: Test input for which part? I only have an example for a GSUB + Morx font in vertical mode, for example:
```util/hb-view ~/dev/Mac_Fonts/10_10/System/Library/Fonts/ヒラギノ明朝\ ProN\ W3.otf --direction=ttb  "[Hello4]"```
 
I currently do not have an example for a font that has vertical substitution in AAT, but not GSUB. Does that help?

Change LGTM, Chromium bot run completed successfully here:
https://chromium-review.googlesource.com/c/chromium/src/+/2020963

ebraminio: Great, I needed exactly something like the command have given, is there any repro on 10.12 10.13 10.14 10.15 fonts? I couldn't replicate the issue on a similar font on 10.15, should find CJK fonts having both morx/gsub using lists have generated before.

ebraminio: Done https://github.com/harfbuzz/harfbuzz/commit/d809aca8d05c64c0e871923b36bdd3b17a85358d

PR #2132, opened 2020-01-24
[subset] Add COLR support

ckitagawa-work:
This table was pretty straightforward to subset.

Test font is from: https://github.com/mozilla/twemoji-colr please let me know if you prefer a different source.

[Spec](https://docs.microsoft.com/en-us/typography/opentype/spec/colr)

I'm leaving the CPAL table as is for now. In theory we could try to subset it, but since both COLR and SVG use it there is some complexity there. I don't think Fonttools even tries to subset it so might not be worthwhile.


src/hb-ot-color-colr-table.hh
```diff
@@ -112,6 +115,48 @@ struct COLR
...
+    bool is_valid ()
+    {
+      return colr.get_blob ()->length;
+    }
+
+    void get_related_glyphs (hb_codepoint_t glyph,
+                             hb_set_t *related_ids /* OUT */) const
+    {
+      colr->get_related_glyphs (glyph, related_ids);
+    }
```
ebraminio: Feel free to use one-liner method definitions here :)
```
    void fini () { this->colr.destroy (); }

    bool is_valid () { return colr.get_blob ()->length; }

    void get_related_glyphs (hb_codepoint_t glyph,
                             hb_set_t *related_ids /* OUT */) const
    { colr->get_related_glyphs (glyph, related_ids); }
```

ckitagawa-work: Done.

src/hb-ot-color-colr-table.hh
```diff
@@ -120,6 +155,89 @@ struct COLR
 			  (this+layersZ).sanitize (c, numLayers)));
   }
 
+  template<typename Iterator,
+           hb_requires (hb_is_iterator (Iterator))>
+  bool serialize (hb_subset_context_t *c,
```
garretrieger: We prefer to have serialize not take a subset_context where possible (so that serialize can be used outside of subsetting). Since this table is pretty straightforward that should be possible. You could change this function to take a serialize context, the version and two iterators, one that iterates through BaseGlyphRecord's and a second that iterates through the LayerRecord's. The iterators should yield the final record (with gid's and layer indices already remapped) at each iteration instead of relying on the serialize function to fix those. The remapping logic would then be moved up into the subset call.

src/hb-ot-color-colr-table.hh
```diff
@@ -120,6 +155,89 @@ struct COLR
...
+    | hb_map ([&](unsigned int new_gid)
+        {
+          hb_codepoint_t old_gid = 0;
+          if (unlikely (!c->plan->old_gid_for_new_gid (new_gid, &old_gid)))
+            return hb_pair_t<hb_codepoint_t, const BaseGlyphRecord*>(new_gid, nullptr);
+
+          const BaseGlyphRecord* record = &(this+baseGlyphsZ).bsearch (numBaseGlyphs, old_gid);
+          if (record && (hb_codepoint_t) record->glyphId != old_gid)
+            record = nullptr;
+          return hb_pair_t<hb_codepoint_t, const BaseGlyphRecord*> (new_gid, record);
```
garretrieger: As I mentioned above you could modify this iterator to return BaseGlyphRecord instead of BaseGlyphRecord* and then remap the gid here so serialize doesn't need to do it later on.

src/hb-ot-color-colr-table.hh
```diff
@@ -120,6 +155,89 @@ struct COLR
...
+
+    return_trace (true);
+  }
+
+  bool subset (hb_subset_context_t *c) const
+  {
+    TRACE_SUBSET (this);
+
+    auto it =
+    + hb_range (1u, c->plan->num_output_glyphs ())  // Skip notdef.
```
garretrieger: If you start the iterator with c->plan->glyph_map->iter() that will give you pairs of (old gid, new gid) then you don't need to manually call old_gid_for_new_gid and check for null's.

src/hb-ot-color-colr-table.hh
```diff
@@ -120,6 +155,89 @@ struct COLR
...
+    | hb_map ([&](unsigned int new_gid)
+        {
+          hb_codepoint_t old_gid = 0;
+          if (unlikely (!c->plan->old_gid_for_new_gid (new_gid, &old_gid)))
+            return hb_pair_t<hb_codepoint_t, const BaseGlyphRecord*>(new_gid, nullptr);
+
+          const BaseGlyphRecord* record = &(this+baseGlyphsZ).bsearch (numBaseGlyphs, old_gid);
```
garretrieger: Recommend creating a small helper function get_base_glyph_record(gid) that does bsearch() and checks the resulting record for a match.

src/hb-ot-color-colr-table.hh
```diff
@@ -112,6 +115,38 @@ struct COLR
     return glyph_layers.length;
   }
 
+  struct accelerator_t
```
ebraminio: I think you can drop this accelerator_t here.

behdad:
`get_related_glyphs` should be called `closure_glyphs` or whatever it is we call it in other places.

src/hb-ot-color-colr-table.hh
ckitagawa-work: That makes sense. Thanks for the insight/suggestion.

src/hb-ot-color-colr-table.hh
ckitagawa-work: I was using it as a RAII style constructing the table to use `closure_glyphs ()`. I could add a helper instead, but all the other tables that add closure glyphs have an accelerator to do so...

src/hb-ot-color-colr-table.hh
garretrieger: Ah good point about the ordering. You are correct, the map iteration order is random.

src/hb-ot-color-colr-table.hh
```diff
@@ -120,6 +155,113 @@ struct COLR
...
+            out_layers[i].glyphId = new_gid;
+          }
+
+          return hb_pair_t<bool, hb_vector_t<LayerRecord>> (true, out_layers);
+        })
+    | hb_filter (hb_first)
+    | hb_map_retains_sorting (hb_second)
+    ;
+
+    if (unlikely (!base_it.len () ||
```
garretrieger: You can just do !base_it instead of !base_it.len(), likewise for layer_it

src/hb-ot-color-colr-table.hh
```diff
@@ -112,6 +115,38 @@ struct COLR
...
+  void closure_glyphs (hb_codepoint_t glyph,
+                       hb_set_t *related_ids /* OUT */) const
+  {
+    const BaseGlyphRecord &record = (this+baseGlyphsZ).bsearch (numBaseGlyphs, glyph);
+
+    hb_array_t<const LayerRecord> all_layers = (this+layersZ).as_array (numLayers);
+    hb_array_t<const LayerRecord> glyph_layers = all_layers.sub_array (record.firstLayerIdx,
+                                                                       record.numLayers);
+    for (unsigned int i = 0; i < glyph_layers.length; i++)
+      hb_set_add (related_ids, glyph_layers[i].glyphId);
```
behdad: `set.add_array()` can be used here with a custom stride argument.

ckitagawa-work: Done. That's very convenient :)

ckitagawa-work:
I don't have write access, can someone please merge when they have a second? Thanks!

ebraminio:
Sure! Thanks :)

PR #2133, opened 2020-01-25
Rename hb_ot_glyph_* to hb_outline_*

ebraminio:
Let's settle this before the release, @khaledhosny WDYT? Even the fact we don't support non opentype formats and it's unlikely that would be changed in future, I think vector graphic glyphs are also available outside OpenType world and we may also be able to off load our decompose function for such fonts to other libraries in future, that's my reasoning but not sure if that's enough.


khaledhosny:
I have no strong opinion either way, and I don’t totally understand the `ot` distinction in HarfBuzz.

ebraminio:
> I have no strong opinion either way, and I don’t totally understand the ot distinction in HarfBuzz.

Me neither :) Perhaps we should wait for Behdad then

ebraminio:
And is the naming OK to you? I mean maybe hb_outline_ instead hb_glyph_ or something else instead decompose, please let me know about anything you feel about the API. Thank you very much

ebraminio:
Hmm. Am convincing myself more and more that if we support Type1 for example there would be no reason to not provide its outlines also through this very API.

ebraminio:
have got some feedbacks from Behdad finally and am updating accordingly

PR #2135, opened 2020-01-26
Use correct return type in _unsafe_to_break_find_min_cluster.

RazrFalcon:
Cluster should be unsigned. I guess it can also lead to a precision loss on large numbers.


ebraminio:
Thank you :)

PR #2138, opened 2020-01-28
Fix page_map corruption in hb_set_t during process().

garretrieger:
If a process operation results in less pages then the current set has, it will likely corrupt the page_map since it overwrites page_map entries ahead of where it's processing. This fixes that problem by always writing updated entries at the end of the page_map and then compacting and cleaning up orphaned entries at the end of process().

### DIFF
--- a/src/hb-set.hh
+++ b/src/hb-set.hh
@@ -452,21 +452,23 @@ struct hb_set_t
 
   void compact (unsigned int length)
   {
-    hb_hashmap_t<uint32_t, uint32_t> old_index_to_page_map_index;
+    hb_vector_t<uint32_t> old_index_to_page_map_index;
+    old_index_to_page_map_index.resize(pages.length);
+    for (uint32_t i = 0; i < old_index_to_page_map_index.length; i++)
+      old_index_to_page_map_index[i] = 0xFFFFFFFF;
 
-    for (unsigned int i = 0; i < length; i++) {
-      old_index_to_page_map_index.set (page_map[i].index, i);
-    }
+    for (uint32_t i = 0; i < length; i++)
+      old_index_to_page_map_index[page_map[i].index] =  i;
 
     compact_pages (old_index_to_page_map_index);
   }
 
-  void compact_pages (const hb_hashmap_t<uint32_t, uint32_t>& old_index_to_page_map_index)
+  void compact_pages (const hb_vector_t<uint32_t>& old_index_to_page_map_index)
   {
     unsigned int write_index = 0;
     for (unsigned int i = 0; i < pages.length; i++)
     {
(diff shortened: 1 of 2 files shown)


behdad:
I first like to understand the problem.  Will try reading again later.

behdad:
Okay yes I see the problem.  Let me review the code now.

behdad:
So, since you are creating a new vector in compact anyway, why not just collect the process results in a new vector and assign at the end?  The `is_empty()` optimization can be added right in `process()` itself if necessary.  Maybe can take a hint from caller in terms of template constants as to whether processing of two non-empty pages can produce empty page, since `union`, the most common operator we use, doesn't do that.

garretrieger:
Another option, is I've worked out a way to do the compact() in place (but will require a hb_map_t to be allocated). I'll also look into seeing if I can do it as part of collect as you suggest, but I want to avoid changing the code for the unions and other ops that don't reduce page count since that's correct as is.

garretrieger:
Alright, added an inplace version of compact. I prefer this version since it does less memory allocation and copying. If you agree I'll delete the collect and copy version.

behdad:
Before I review yours let me share this other approach idea: in `process` as is, in the first round where we are counting final size, we can remove unneeded items from the `this->page_map`.  That would address the original problem.  Might be simpler.

garretrieger:
Thought about it some more, and you're right we'll actually need to remove page_map entries in the first loop. There's another bug where if you have a op which is only pass through right you can have a situation where a page is removed from the left set, but one is add back from the right set (so total size doesn't change). Currently the code will write out of bounds of the page array since it assumes all added pages start at the original pages.length.

So we'll need to remove uneeded page_map entries in the first loop and then compact down the array and set first_page to the end of the compacted area. Then finally also compact and re-index the pages. Then the second half of the code can be left untouched.

garretrieger:
Ok, updated to remove during the first part. Please take a look.

garretrieger:
Ping. Does this look good now?

ebraminio:
To be honest I am yet to digest the whole thing as wasn't curious to know till now how it works underneath and that's why haven't merged it so far but am worried about the existence such issues and also am thinking about how we should fuzz our hb_map_t and hb_set_t to not have surprises.

garretrieger:
I wrote up a hb-set-fuzzer that I used to test this and check for any additional cases. I've got to clean it up a bit then can send a PR. Should be pretty easy to do a hb-map one as well.

behdad:
Feel free to merge.  I still think can be done more efficiently but do whatever you think is right.  Instead of `hb_map_t` can use a vector to map old to new, more efficient.

ebraminio:
Looks good! Thanks :)

PR #2139, opened 2020-01-29
[subset] Fix namerecord ordering

qxliu76:
(In theory original name records should be in sorted order already, I may need look into more details)
This will fix inconsistency with fontTools.
Also according to the spec, namerecords must be sorted
first by platform ID, then by platform-specific ID,
then by language ID, and then by name ID.


behdad:
Can't we sort them in the output array?

behdad:
Also don't use malloc.  Use calloc.

qxliu76:
> Also don't use malloc. Use calloc.

done

qxliu76:
> Can't we sort them in the output array?

done

src/hb-ot-name-table.hh
```diff
@@ -116,6 +116,29 @@ struct NameRecord
...
             (p == 3 && (e == 0 || e == 1 || e == 10)));
   }
+  
+  static int cmp (const void *pa, const void *pb)
+  {
+    const NameRecord *a = (const NameRecord *)pa;
+    const NameRecord *b = (const NameRecord *)pb;
+
+    if (a->platformID != b->platformID)
+      return a->platformID < b->platformID ? -1 : +1;
```
garretrieger: A slightly more compact way to do this is

```
return a->platformID - b->platformID;
```

src/hb-ot-name-table.hh
```diff
@@ -200,7 +223,19 @@ struct name
 
     const void *dst_string_pool = &(this + this->stringOffset);
 
-    c->copy_all (it, src_string_pool, dst_string_pool);
+    NameRecord *name_records = (NameRecord *) calloc (it.len (), NameRecord::static_size);
+    hb_array_t<NameRecord> records (name_records, it.len ());
```
garretrieger: Use hb_vector_t and call alloc () to size it and push() to add. Then you don't need to do the calloc()/free() and memcpy(). After that you can qsort() and pass .iter() to the copy_all as you're currently doing.

garretrieger: Oh actually, you don't need the temporary array/vector I think you should be able to just qsort() on nameRecordZ after you copy_all too it.

behdad: > Oh actually, you don't need the temporary array/vector I think you should be able to just qsort() on nameRecordZ after you copy_all too it.

Right.  That's what I thought as well.

src/hb-ot-name-table.hh
behdad: Also put the `cmp` in `NameRecord` itself.

src/hb-ot-name-table.hh
qxliu76: the resolve_links () is called after serialize (), if we qsort after copy_all, the offset value might be set to wrong value.

Also the NameRecord doesn't overload operator = function, so the push() in vector doesn't work. According to coding style, we don't usually overload "=" for struct?

garretrieger: Ah okay, makes sense. Then this looks good as is.

src/hb-ot-name-table.hh
ebraminio: > A slightly more compact way to do this is
> return a->platformID - b->platformID;

This isn't the common pattern on the table, will apply it later.

> Also put the cmp in NameRecord itself.

Won't work as is, needs the following also, will provide it as a separate PR.

```diff
diff --git a/src/hb-array.hh b/src/hb-array.hh
index cbd6485a..2c1bb2fd 100644
--- a/src/hb-array.hh
+++ b/src/hb-array.hh
@@ -145,16 +145,22 @@ struct hb_array_t : hb_iter_with_fallback_t<hb_array_t<Type>, Type&>
     return not_found;
   }
 
-  hb_sorted_array_t<Type> qsort (int (*cmp_)(const void*, const void*))
+  HB_INTERNAL static int qsort_cmp (const void *pa, const void *pb)
+  {
+    Type *a = (Type *) pa;
+    Type *b = (Type *) pb;
+    return b->cmp (*a);
+  }
+  hb_sorted_array_t<Type> qsort ()
   {
     if (likely (length))
-      hb_qsort (arrayZ, length, this->get_item_size (), cmp_);
+      hb_qsort (arrayZ, length, this->get_item_size (), qsort_cmp);
     return hb_sorted_array_t<Type> (*this);
   }
-  hb_sorted_array_t<Type> qsort ()
+  hb_sorted_array_t<Type> qsort (int (*cmp_)(const void*, const void*))
   {
     if (likely (length))
-      hb_qsort (arrayZ, length, this->get_item_size (), Type::cmp);
+      hb_qsort (arrayZ, length, this->get_item_size (), cmp_);
     return hb_sorted_array_t<Type> (*this);
   }
   void qsort (unsigned int start, unsigned int end)
diff --git a/src/hb-ot-name-table.hh b/src/hb-ot-name-table.hh
index 0da56718..b3b1a9e3 100644
--- a/src/hb-ot-name-table.hh
+++ b/src/hb-ot-name-table.hh
@@ -112,30 +112,27 @@ struct NameRecord
   {
     unsigned int p = platformID;
     unsigned int e = encodingID;
-    
+
     return (p == 0 ||
             (p == 3 && (e == 0 || e == 1 || e == 10)));
   }
-  
-  static int cmp (const void *pa, const void *pb)
-  {
-    const NameRecord *a = (const NameRecord *)pa;
-    const NameRecord *b = (const NameRecord *)pb;
 
-    if (a->platformID != b->platformID)
-      return a->platformID < b->platformID ? -1 : +1;
+  int cmp (const NameRecord &record)
+  {
+    if (record.platformID != platformID)
+      return record.platformID < platformID ? -1 : +1;
 
-    if (a->encodingID != b->encodingID)
-      return a->encodingID < b->encodingID ? -1 : +1;
+    if (record.encodingID != encodingID)
+      return record.encodingID < encodingID ? -1 : +1;
 
-    if (a->languageID != b->languageID)
-      return a->languageID < b->languageID ? -1 : +1;
+    if (record.languageID != languageID)
+      return record.languageID < languageID ? -1 : +1;
 
-    if (a->nameID != b->nameID)
-      return a->nameID < b->nameID ? -1 : +1;
+    if (record.nameID != nameID)
+      return record.nameID < nameID ? -1 : +1;
 
-    if (a->length != b->length)
-      return a->length < b->length ? -1 : +1;
+    if (record.length != length)
+      return record.length < length ? -1 : +1;
 
     return 0;
   }
```

PR #2142, opened 2020-01-29
[draw] Turn funcs to pen to keep current position

ebraminio:
Just for the record I guess, if we like to have two ways calling on hb_draw_funcs_t at least for testings proposes like quad to cubic translator it would need some internal state so renaming to `hb_draw_pen_t` also.


ebraminio:
Let's close this, better to discuss on #2145 anyway, can be revived anytime.

behdad:
Okay I'm confused now.

What's `draw` and what's `pen`?  Why need both?

Maybe we can set up a video meeting to discuss this some time.  Pens are a big thing.  I don't want to commit to adding all kinds of pens right now, but want whatever API we add to be usable with pen composition down the road.

ebraminio:
We don't need both, _funcs_ naming doesn't imply has an internal state, which is needed, a _pen_ naming can imply so more.

Sure. I am not happy enough about current state and `glyf` part of the API, and should study Windows and macOS APIs and their results.

PR #2144, opened 2020-01-30
[draw] Put current points on callbacks

ebraminio:
Another draft like #2142, this makes callbacks themselves ugly but usable for quadratic to cubic or curve to lines usecases or make it possible to be called from outside (that would be ugly also as client has to pass current point). I don't happy about putting hb_draw_funcs_t on the call but it was needed for translating calls, avoidable if we user_data somehow on funcs object so no need for passing both.


ebraminio:
Let's close this, better to discuss on #2145 anyway, can be revived anytime.

Issue #2145, opened 2020-01-30
Improve glyph draw design


ebraminio:
Made at Behdad request on discussing and finalizing the API in GitHub. See https://github.com/harfbuzz/harfbuzz/issues/2145#issuecomment-589993318 also

Our current hb_font_draw_glyph/hb_draw_funcs_t design can be considered something stable and useful. It is able to draw a multilayer COLR/CPAL Emoji color fonts even[[1]](https://github.com/harfbuzz/harfbuzz/blob/master/src/main.cc#L127-L242)! But we like to make sure it is the best design we can go with. That's why I've upload a one-click away disable #2141 patch for and these #2142 #2144 to may improve its design.

See things like fonttools [BasePen](https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/pens/basePen.py#L187) and its [other awesome pens](https://github.com/fonttools/fonttools/tree/master/Lib/fontTools/pens). I think we like to have a design like that to make us or the clients capable of implementing things like that or even implementing a rasterize on top of our API easily.

One request one may have is being able to call path commands callbacks from client code. The other request is being able to implement quadratic to cubic or, cubic-line to calls translation but those aren't useful if one doesn't have access to current point/state (as one needs to know starting point to translate those call) but our current API, tends to be immutable, doesn't hold the current point.

#2142 tries to tackle the issue by keeping current point and state in the object, #2144 goes the other way by passing state through the callbacks. The first looks more what BaePen does in fonttools but both sounds a bit off to me right now so guess we should create a separate object called pen to support these usecases and keep `hb_draw_funcs_t` intact. Maybe that pen object should return a `hb_draw_funcs_t` object that can be passed to `hb_font_draw_glyph` and having different kind of callbacks with having current point also.

Getting this right would be a nice thing considering if we get an in-house rasterizer someday, one will not only be able to rasterize a glyph of a font but call the rasterizer for their custom paths or implement several utilities like curve to lines translator using an established design, it makes testing of those functionalities also easier.

ebraminio:
I am trying to convince myself more and more this is minimal but acceptable API we can provide and can build other things on top of it but am of course love to hear any feedback about the API before the release.

ebraminio:
Skia uses quadTo instead quadratic, may we should also?

ebraminio:
Useful links, 
* https://github.com/fonttools/fonttools/tree/master/Lib/fontTools/pens (added a pen #2170 should add more things from there)
[...]

PR #2149, opened 2020-02-02
[draw] Add hb_draw_funcs_is_immutable and hb_draw_funcs_make_immutable

ebraminio:
Auxiliary things

Other things to consider,
* parent parameter for create and inheritence, `hb_unicode_funcs_create (hb_unicode_funcs_t *parent);` has it but `hb_font_funcs_create (void)` don't so not sure if we need it also.
* `hb_draw_{move,line,quadratic,cubic}_to ()` and `hb_draw_close_path ()`, I think doing so implies `hb_draw_funcs_t` is useful for two ways callings but without having current points using untouched `hb_draw_funcs_t` object isn't that useful for cases line quadratic to cubic or, curve to lines translations, guess we need another abstraction, maybe called pen, to provide those.

### DIFF
--- a/docs/harfbuzz-sections.txt
+++ b/docs/harfbuzz-sections.txt
@@ -231,6 +231,8 @@ hb_draw_move_to_func_t
 hb_draw_quadratic_to_func_t
 hb_draw_funcs_create
 hb_draw_funcs_destroy
+hb_draw_funcs_is_immutable
+hb_draw_funcs_make_immutable
 hb_draw_funcs_reference
 hb_draw_funcs_set_close_path_func
 hb_draw_funcs_set_cubic_to_func
--- a/src/hb-draw.cc
+++ b/src/hb-draw.cc
@@ -187,6 +187,38 @@ hb_draw_funcs_destroy (hb_draw_funcs_t *funcs)
 }
 
 /**
+ * hb_draw_funcs_make_immutable:
+ * @funcs: draw functions
+ *
+ * Makes funcs object immutable.
+ *
+ * Since: REPLACEME
+ **/
+void
+hb_draw_funcs_make_immutable (hb_draw_funcs_t *funcs)
+{
+  if (hb_object_is_immutable (funcs))
+    return;
+
+  hb_object_make_immutable (funcs);
+}
+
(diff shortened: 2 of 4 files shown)


behdad:
I like to review this work.  I'm really struggling though because there seems to be PRs coming and going like mushrooms.  Why isn't there like one central place to discuss design first implement later? :)

behdad:
> Other things to consider,
> 
> * parent inherit system, `hb_unicode_funcs_create (hb_unicode_funcs_t *parent);` has it but `hb_font_funcs_create (void)` don't so not sure if we need it also.

For fonts, inheritance happens with `hb_font_create_sub_font`.

> * `hb_draw_{move,line,quadratic,cubic}_to ()` and `hb_draw_close_path ()`, I think doing so implies `hb_draw_funcs_t` is useful for two ways callings but without having current points using untouched `hb_draw_funcs_t` object isn't that useful for cases line quadratic to cubic or, curve to lines translations, guess we need another abstraction, maybe called pen, to provide those.

Let's discuss the full design in one place and adapt implementation.

ebraminio:
>  Why isn't there like one central place to discuss design first implement later? :)

#2145 but gets no attention :)

ebraminio:
> For fonts, inheritance happens with hb_font_create_sub_font.

Was referring to `font_funcs`

> Let's discuss the full design in one place and adapt implementation.

The best way to become confident about the design is to implement (not necessarily merge) enough use for the implementation, things like rasterizer, transforms, etc.

behdad:
> > For fonts, inheritance happens with hb_font_create_sub_font.
> 
> Was referring to `font_funcs`

I know.  And I was making a very relevant point.

The reason font_funcs don't have inheritance themselves is that font_funcs don't carry state.  font_t does.  So, font_funcs are to fonts what in your current tentative design draw_funcs are to pens.  The difference is that there's no common functionality to pens so there's nothing we can formalize in harfbuzz itself.

> The best way to become confident about the design is to implement (not necessarily merge) enough use for the implementation, things like rasterizer, transforms, area calc, different pens, etc.

Sure.  I'll take another look.

PR #2151, opened 2020-02-03
[subset] Avoid incorrectly dropping cmap for NotoColorEmoji.ttf

ckitagawa-work:
[NotoColorEmoji.ttf](https://www.google.com/get/noto/help/emoji/) uses two  `cmap` subtables

```
Format 14 | Platform ID 0 | Platform Encoding ID  5
Format 12 | Platform ID 3 | Platform Encoding ID 10
```

This combination results in the `cmap` table being dropped during subsetting despite being valid/required.

As far as I can tell the issue is in this block of code

```cpp
// src/hb-ot-cmap-table.hh Line 1204
const EncodingRecord *unicode_bmp= nullptr, *unicode_ucs4 = nullptr, *ms_bmp = nullptr, *ms_ucs4 = nullptr;
bool has_format12 = false;

for (const EncodingRecord& _ : encodingrec_iter)
 {
    unsigned format = (this + _.subtable).u.format;
    if (format == 12) has_format12 = true; // This is true.

    const EncodingRecord *table = hb_addressof (_);
    if      (_.platformID == 0 && _.encodingID ==  3) unicode_bmp = table;
    else if (_.platformID == 0 && _.encodingID ==  4) unicode_ucs4 = table;
    else if (_.platformID == 3 && _.encodingID ==  1) ms_bmp = table;
    else if (_.platformID == 3 && _.encodingID == 10) ms_ucs4 = table; // This is set.
}

// At this point ms_ucs4 != nullptr and has_format12 == true
if (unlikely (!unicode_bmp && !ms_bmp)) return_trace (false);  // PROBLEM: Evaluates to false so the  table is dropped!
if (unlikely (has_format12 && (!unicode_ucs4 && !ms_ucs4))) return_trace (false);
```

The intended logic appears to try to abort if an invalid combination of tables are provided. However, it filters a working and valid combination for NotoColorEmoji (I've check that the binary format of the resulting `cmap` table is valid when using hb-subset).

I'm not sure what the "correct" fix is as I'm not very familiar with the in-use formats for `cmap` tables. ~~This PR currently provides a working solution, but it may not be the semantically correct one in the general case. I'd appreciate if someone with more experience could take a look and tell me if this is the right fix.~~ EDIT: updated to use Garret's method, although it sounds like most of this might need to be re-written? I've added both API and subset tests for this.

Thanks!

[Spec](https://docs.microsoft.com/en-us/typography/opentype/spec/cmap)

Originally part of: https://github.com/harfbuzz/harfbuzz/pull/2121


garretrieger:
So it appears we're not allowing a font that is only using ucs4 tables (for example something with just a cmap 12), but that feels wrong to me. So another possible fix is:

Changing: 
```
if (unlikely (!unicode_bmp && !ms_bmp)) return_trace (false);
if (unlikely (has_format12 && (!unicode_ucs4 && !ms_ucs4))) return_trace (false);
```
To
```
if (unlikely (!has_format12 && !unicode_bmp && !ms_bmp)) return_trace (false);
if (unlikely (has_format12 && (!unicode_ucs4 && !ms_ucs4))) return_trace (false);
```
But I'll defer to Behdad.

behdad:
Okay I read the existing code and have comments.  But the code can be fully rewritten / overhauled. :)

First, note how we *use* cmap table.  That's mostly in accelerator's `init()`.  We find "best" subtable, as well as possibly a "uvs" (Unicode Variation Selectors) subtable.  For finding the "uvs", we look for plat/enc pair of (0,5) and only accept if its format is 14.

To that end, Garret's comment is correct, and your proposal doesn't make sense to me.

Most of the code in `subset` can go.  I'll take another stab at writing my thoughts down tomorrow.

ebraminio:
Merging but waiting for Behdad comment also

PR #2154, opened 2020-02-04
[subset] Keep head when there is no glyf table

ckitagawa-work:
For CBDT/CBLC subsetting the test font NotoColorEmoji is used. It has no `glyf` table yet it has a `head` table to keep. Similar behavior was added inside the subset routines for `cff1` and `cff2`. 

Rather than special casing this behavior for certain tables, this PR makes this behavior the default for fonts that don't have a `glyf` table (or when the `glyf` table is dropped) by adding direct subset methods for the `head` table and calling it from `hb-subset.cc`.

This behavior is already covered by the tests for the `cff` tables.

### DIFF
--- a/src/hb-ot-cff1-table.hh
+++ b/src/hb-ot-cff1-table.hh
@@ -27,7 +27,6 @@
 #ifndef HB_OT_CFF1_TABLE_HH
 #define HB_OT_CFF1_TABLE_HH
 
-#include "hb-ot-head-table.hh"
 #include "hb-ot-cff-common.hh"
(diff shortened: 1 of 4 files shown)


ckitagawa-work:
Infra failure :(

ebraminio:
Looks good, am merging and try to remove or improve `_is_table_present` :)

ebraminio:
> and try to remove or improve _is_table_present :)

Actually less costly way to do the thing already! Just want to see if I can do something more idiomatic.

PR #2155, opened 2020-02-04
[subset] GPOS4 MarkBase subsetting support

qxliu76:


src/hb-ot-layout-gpos-table.hh
```diff
@@ -491,12 +516,27 @@ struct MarkRecord
 {
   friend struct MarkArray;
 
+  unsigned get_klass () const { return (unsigned) klass; }
```
ebraminio: guess get_class should work here :)

src/hb-ot-layout-gpos-table.hh
```diff
@@ -1514,6 +1568,31 @@ typedef AnchorMatrix BaseArray;		/* base-major--
 					 * mark-minor--
 					 * ordered by class--zero-based. */
 
+static void Markclass_closure_and_remap_indexes (const Coverage  &mark_coverage,
+						 const MarkArray &mark_array,
+						 const hb_set_t  &glyphset,
+						 hb_map_t*        klass_mapping /* INOUT */)
+{
+  hb_vector_t<HBUINT16> orig_klasses;
```
ebraminio: ditto

src/hb-ot-layout-gpos-table.hh
```diff
@@ -1514,6 +1568,31 @@ typedef AnchorMatrix BaseArray;		/* base-major--
...
+  hb_vector_t<HBUINT16> orig_classes;
+
+  + hb_zip (mark_coverage, mark_array)
+  | hb_filter (glyphset, hb_first)
+  | hb_map (hb_second)
+  | hb_map (&MarkRecord::get_class)
+  | hb_sink (orig_classes)
+  ;
+
+  orig_classes.qsort ();
```
garretrieger: Another option is to use an hb_set_t instead of a vector, since those maintain sorted order by default.

PR #2158, opened 2020-02-05
[glyf] Skip empty contours

ebraminio:
As https://savannah.nongnu.org/bugs/index.php?57701

@bungeman Thanks for documenting the need at https://savannah.nongnu.org/bugs/index.php?57701 we like to provide some API akin to FT_Outline_Decompose, what we offer here right is very primitive however, no rendering and no [auto] hinting right now but variations support, hopefully we can get its API right at this point correctly at least and build all the missing things on top of it later so feel free to have a look.

@RazrFalcon The work here is similar to https://github.com/RazrFalcon/ttf-parser/blob/8cda9aa/src/glyf.rs#L235-L238 (not optimized like that code however of course) Am asking however how to have the same logic on CFF/CFF2 as couldn't find respective code there with a quick scan at least, maybe not needed at all also. Thanks!


src/hb-ot-glyf-table.hh
```diff
@@ -1098,6 +1098,14 @@ struct glyf
 	  if (points[i].is_end_point)
 	    break;
 	}
+
+	/* Skip contours with less than 3 points */
+	if (contour_length < 3)
```
bungeman: A glyf contour with two non-coincident points is still an implicitly closed contour with length but no area. This can make a difference when stroking. It's an odd case, but I would hate to preclude it. In particular, it seems GDI/DirectWrite only skip single point contours and report two point contours (even if the points are coincident).

behdad: Agree.

RazrFalcon:
>asking however how to have the same logic on CFF/CFF2 as couldn't find respective code

We don't know the char string length beforehand, so we can't do this. `ttf-parser` checks the final bbox, and if it has a zero size - returns `None`, which indicates an empty contour.

src/hb-ot-glyf-table.hh
ebraminio: Done, added Stroking.ttf also from https://skia-review.googlesource.com/c/skia/+/266945 that tests the difference. Thanks!

ebraminio:
Thanks @RazrFalcon. So according to 

> I suppose that OpenType CFF fonts aren't supposed to use attachment points (or at least afdko doesn't support it) [0], so speculation in the previous comment isn't really relevant. However, the fact that ps_builder_close_contour is dropping the points is still interesting.
>
> [0] http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#2evi-contour-point

We now know we can care less about the case so guess we are good to have this now.

Thanks!

bungeman:
Just wanted to note that there really are two very different kinds of users.

The first and most common is the user who wants the get_path updated here, who just wants the paths to do graphics drawing. In addition to skipping single point contours these users generally want to skip all fully degenerate curves (all points the same as the start point) and skip contours which are fully degenerate (all curves in the contour fully degenerate).

The second set of users are those who really would like a validated and better structured view of the original data for editing or visualization of the glyph itself. Such users would like different structures for different formats, for example the point (x, y, on/off) and contour starts from 'glyf' or the list of points/operators from 'CFF'.

FreeType tries to cater to both of these through FT_Outline_Decompose and FT_Outline, respectively. The first sort of user is fine with abstracting the difference between how different formats encode outlines but the second often is not. If the goal is to provide the 'one true parser' then different tables with outlines (like 'glyf' and 'CFF ') may need different ways to present the validated data. In other words FT_Outline is a nice intermediate format for FreeType, but not actually low level enough for all glyph inspection since it somewhat shoehorns 'CFF ' into a 'glyf'+cubics structure, but it does a good job with 'glyf'.

ebraminio:
Thanks for the explanation. I think HarfBuzz intention is to be in service of first kind of users.

Like to know if our current output format and API is something useful for Skia, here is our tests https://github.com/harfbuzz/harfbuzz/blob/master/test/api/test-draw.c which I am just checking visually against SVG's path which like to check against stricter rules.

Wish someday [which understandably isn't soon] Skia for WebAssembly or Flutter could use HarfBuzz for size reduction matter as it reuse the same facilities as the rest of the library and provides numerous size reduction build flags.

PR #2161, opened 2020-02-06
Add hb-set-fuzzer.

garretrieger:
It fuzzes all of the hb_set process methods (intersection, subtraction, union, and symmetric difference).

### DIFF
--- a/test/fuzzing/Makefile.am
+++ b/test/fuzzing/Makefile.am
@@ -26,6 +26,7 @@ EXTRA_DIST += \
 check_PROGRAMS = \
 	hb-shape-fuzzer \
 	hb-subset-fuzzer \
+	hb-set-fuzzer \
 	$(NULL)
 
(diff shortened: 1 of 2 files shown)


ebraminio:
Do you think we can have hb-map on the same fuzzer, or both on shape or
subset fuzzers, just am thinking about carbon footprint of adding one other
fuzzer ☺️ otherwise honestly have no other thumb rule for when to add other
fuzzer or reuse the older ones, guess this one is the valid one however.

garretrieger:
I think it's better to have each fuzzer specific to the area of code/api it's attempting to fuzz, since the data input will be pretty specific to that api.  Setting up additional fuzz targets on oss-fuzz is pretty light weight so no concern there.

I can add a separate fuzzer for hb_map_t.

ebraminio:
Feel free to hb_map_t fuzzer also :)

Issue #2156, opened 2020-02-04
[subset] add_link/resolve_links for absolute offset


blueshade7:
The current implementation of `add_link` & `resolve_links` appears to assume a parent object containing an offset to a child object is also the base of the offset, maybe with some constant bias.
This is not the case with Private dict in CFF/CFF2, since its offset is stored in FontDict while its offset base is CFF/CFF2 itself (could be said global or absolute).

I am thinking to add a new function `add_link_abs` or a boolean flag to `add_link` to allow its offset to be calculated as absolute.

Better idea?

blueshade7:
Let me interpret no feedback as “ok to proceed”.

I am also going to allow offset to be signed 2-byte or 4-byte, in addition to unsigned.
So in addition to `is_absolute`, going to add `is_signed` flag to `link_t` struct so that `check_assign` does the right thing.
https://github.com/harfbuzz/harfbuzz/blob/c31762e9e89bc440d21a94d8da34608d1649dc85/src/hb-serialize.hh#L70-L76
I am hesitant to steal two additional bits out of `position`, so going to separate flags from it.

`hb_is_signed` does not compile with `OT::IntType` as T. Any workaround?

garretrieger:
One thought, is it possible to do a c->push() at the start of serializing the CFF table so that you have a parent obj and then use the start of the table as the base for calling c->add_link(...)?

blueshade7:
I tried that, and didn't work... `base` arg to `add_link` is a mere char pointer. `link.bias` is calculated as diff between the current obj head from base at`add_link` time, which may change by `resolve_links` time.
A general grand solution may be allowing arg base to be another object, not just a pointer (maybe with an optional constant bias), but this seems an overkill. AFAIK no such table exists positioned at an offset based from another arbitrary table. Either from its immediate parent, or from the very beginning (this problem case).

garretrieger:
Ah got it. In that case, I agree that the absolute offset is the way to go.

behdad:
> I am hesitant to steal two additional bits out of `position`, so going to separate flags from it.

No.  Please steal.  Position doesn't need anything like even 20 bits.

behdad:
If I would to do it, I'd make add_link take one or two additional optional parameters.  Maybe an enum.  Like to have three modes:
- Default: relative to start of current object
- Relative to end of current object; that's useful for name table and other places where there's a pool at the end of an object.
- Absolute.

blueshade7:
[...]

Issue #1981, opened 2019-09-19
Move CFF subsetters to _subset2 and make it the default


ebraminio:
AFAICS CFF tables are the only _subset users and with moving those to _subset2 we can drop the  weird 2 suffix, curious to know anything is blocking that change? @blueshade7

blueshade7:
Just I am tied up by some other task, unlikely to be able to work on it soon myself.
_subset2 would substantially simplify CFF/CFF2 subsetting I believe. I wish I had it when I started working on it, but wasn't, unfortunately.

ebraminio:
Feel free to put some time for it anytime you like, I will also try myself. Thank you very much :)

blueshade7:
I gave it some thought.

Currently CFF/CFF2 subsetters based on `_subset` run in two-phases. During the first phase, sub-tables are generated determining each size & offset, finally the total size.
By using `_subset2`, some of this can be done in one phase.

Table offsets fixed at 2 or 4-bytes in length (thanks to myself) can use the existing mechanism of`hb_serialize_context_t::add_link` & `resolve_links`.
_length_ of Private dict may need an extension to this.

INDEX data, such as CharStrings or Subrs still need to be generated in two-phases, I believe, since the offset size (1 thru 4) cannot be determined until the total size is known. Fixing its size at 4-byte can be prohibitively wasteful. Using the original size may not work with desubroutinized CharStrings (and again can be wasteful).

iter syntax must be handy at various places.

blueshade7:
created `subset2-cff` branch for this work

ebraminio:
Thank you very much, just saw it, looks nice and complete actually, feel free to submit, will merge it if you don't have any concern, thanks!

blueshade7:
The current impl is still written a-la `_subset`, calculating sub-table offsets in first pass, serializing in second pass, not benefitting much from calling `_subset2`. I am planning to rewrite them with `push`/`pop_pack`/`add_link` as much as possible.
There is some awkwardness in doing so though. By the time calling `add_link` for an offset serialized in a Dict, the corresponding sub-table needs to be packed & `objidx` must be known, i.e., tables need to be serialized bottom-up, in reverse order from the way it is currently written.
Another thing is that calculating hash while packing a huge singleton sub-table like CharStrings seems a waste, as it will be unlikely shared.

ebraminio:
[...]

PR #2163, opened 2020-02-07
added add_link_abs()

blueshade7:
As described in issue #2156, I plan to use this new function `add_link_abs()` in hb-serialize.hh for setting up a sub-table link with an absolute offset (i.e., based on the very beginning of the table) for my work of modernizing CFF subsetter with _subset2 #1981.

`ofs` argument to `add_link_abs` may be a signed IntType or an unsigned IntType.

### DIFF
--- a/src/hb-open-type.hh
+++ b/src/hb-open-type.hh
@@ -362,25 +362,11 @@ struct OffsetTo : Offset<OffsetType, has_null>
     return ret;
   }
 
-  template <typename ...Ts>
   bool serialize_copy (hb_serialize_context_t *c,
 		       const OffsetTo& src,
 		       const void *src_base,
-		       const void *dst_base,
-		       Ts&&... ds)
-  {
-    *this = 0;
-    if (src.is_null ())
-      return false;
-
-    c->push ();
-
-    bool ret = c->copy (src_base+src, hb_forward<Ts> (ds)...);
-
-    c->add_link (*this, c->pop_pack (), dst_base);
-
-    return ret;
-  }
+		       const void *dst_base)
+  { return serialize_copy (c, src, src_base, dst_base, hb_serialize_context_t::Head); }
 
   bool sanitize_shallow (hb_sanitize_context_t *c, const void *base) const
   {
--- a/src/hb-open-type.hh
+++ b/src/hb-open-type.hh
@@ -338,6 +338,30 @@ struct OffsetTo : Offset<OffsetType, has_null>
   }
 
   /* TODO: Somehow merge this with previous function into a serialize_dispatch(). */
+  /* Workaround clang bug: https://bugs.llvm.org/show_bug.cgi?id=23029
+   * Can't compile: whence = hb_serialize_context_t::Head followed by Ts&&...
+   */
+  template <typename ...Ts>
+  bool serialize_copy (hb_serialize_context_t *c,
+		       const OffsetTo& src,
+		       const void *src_base,
+		       const void *dst_base,
+		       hb_serialize_context_t::whence_t whence,
+		       Ts&&... ds)
+  {
+    *this = 0;
+    if (src.is_null ())
+      return false;
+
+    c->push ();
+
+    bool ret = c->copy (src_base+src, hb_forward<Ts> (ds)...);
+
+    c->add_link (*this, c->pop_pack (), dst_base, whence);
+
+    return ret;
+  }
+
   template <typename ...Ts>
   bool serialize_copy (hb_serialize_context_t *c,
 		       const OffsetTo& src,
--- a/src/hb-serialize.hh
+++ b/src/hb-serialize.hh
@@ -70,7 +70,9 @@ struct hb_serialize_context_t
     struct link_t
     {
       bool is_wide: 1;
-      unsigned position : 31;
+      bool is_signed: 1;
+      bool is_absolute: 1;
+      unsigned position;
       unsigned bias;
       objidx_t objidx;
     };
@@ -127,7 +129,7 @@ struct hb_serialize_context_t
 
   template <typename T1, typename T2>
   bool check_assign (T1 &v1, T2 &&v2)
-  { return check_equal (v1 = v2, v2); }
+  { return check_equal ((hb_remove_reference<T2>) (v1 = v2), v2); }
 
   template <typename T> bool propagate_error (T &&obj)
   { return check_success (!hb_deref (obj).in_error ()); }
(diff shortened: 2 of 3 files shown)


src/hb-serialize.hh
```diff
@@ -127,7 +129,7 @@ struct hb_serialize_context_t
 
   template <typename T1, typename T2>
   bool check_assign (T1 &v1, T2 &&v2)
-  { return check_equal (v1 = v2, v2); }
+  { return check_equal ((hb_remove_reference<T2>) (v1 = v2), v2); }
```
blueshade7: BUG:
If T1 = int32_t and T2 = unsigned = uint32_t (i.e., on 32-bit platform), this code fails to catch an integer overflow with v2 >= (1<<31).
Need a fix.

blueshade7: Currently T2 is always `unsigned`, needs not a template argument, so casting to `uint64_t` should work as long as sizeof(T1) < sizeof(uint64_t):
```
template <typename T>
bool check_assign (T &v1, unsigned &&v2)
{ return check_equal ((uint64_t) (v1 = v2), v2); }

```
Better idea?

blueshade7: Above code doesn't compile... trying to come up with a right fix.

src/hb-serialize.hh
```diff
@@ -277,11 +279,47 @@ struct hb_serialize_context_t
...
     link.is_wide = sizeof (T) == 4;
+    link.is_signed = false;
+    link.is_absolute = false;
     link.position = (const char *) &ofs - current->head;
     link.bias = (const char *) base - current->head;
     link.objidx = objidx;
   }
 
+  /*
+   * Set a link to an object with an absolute offset.
```
garretrieger: Can you clarify in this comment what an absolute offset is. My understanding is this means from the start of the serialize context.

behdad:
I feel like the implementation is too much changes for what this tries to achieve.  Can you break down into smaller commits?  One for signedness, one for absolute, etc?

src/hb-serialize.hh
behdad: I pushed this, which should do?
https://github.com/harfbuzz/harfbuzz/commit/d663e28af1b84f27bea9730a9a8f18212387bb7f

blueshade7: yes thanks

blueshade7:
Updated PR reflecting feedback & latest master:

Merged `add_link_abs()` into `add_link()`.

`add_link()` now takes another optional enum argument `mode` defaults to `Head`, specifying the current behavior.

When an absolute offset is desired, mode is set to `Absolute`.
Another option is `Tail`.

In either case, the existing optional `base` argument may point at a desired base, distanced from the default base head/tail/start with a constant bias.

Offset type `T` may be signed or unsigned.
Sign of T is tested with `hb_is_signed()` after calling the new meta function`hb_unwrap_type()`.
Bits for two new flags in`link_t` are stolen from `position`.

Minor: private function assign renamed to assign_offset.

behdad:
Thanks.

I would have called new argument `whence`, which is what `seek(3)` uses, probably with same modes: http://man7.org/linux/man-pages/man3/fseek.3.html

The non-null base vs non-default modes need more work.  The base argument was originally added to handle situations were `end` offsetting was desired.  But looks like it's used for other things as well.  At any rate; for mode=start and mode=absolute it's clear what base should do.  For mode=end it's less clear.  We need to figure outl

It also  looks to me like you are "resolving" the offset for mode=end based on current tail of the object; whereas the object might be enlarged after.  We should do the end-math in resolve_links.

src/hb-serialize.hh
```diff
@@ -279,15 +289,34 @@ struct hb_serialize_context_t
...
+    auto& link = *current->links.push ();
+
+    switch (mode)
+    {
+    case Head:
+      dflt_base = current->head;
+      link.is_absolute = false;
+      break;
+    case Tail:
+      dflt_base = current->tail;
```
behdad: This is the problematic part.

blueshade7: >whereas the object might be enlarged after. We should do the end-math in resolve_links.

I can do that. In that case, I believe mode should be stored in `link_t` replacing `is_absolute`.

behdad: > I can do that. In that case, I believe mode should be stored in `link_t` replacing `is_absolute`.

Correct.

We should try to port `name` table to use that.   But feel free to do this without that for now.  Thanks.

blueshade7:
> At any rate; for mode=start and mode=absolute it's clear what base should do. For mode=end it's less clear. We need to figure outl

I see what you mean. I am suspecting currently in many cases base is set to `current->head` so unnecessary in the first place. If there are cases where base is meaningfully used, I am thinking to replace it with a bias to be calculated by the caller. I am looking at each call (10 in total) now.

blueshade7:
Experimentally removed base argument from `add_link()` and rewrote name table to serialize with `whence=Tail` in a branch `absolute-link-test`. All test passed (except one annoying timeout with a fuzzed file).

blueshade7:
I want `OffsetTo::serialize_copy()` to default `whence` to `Head` as:
```
template <typename ...Ts>
bool serialize_copy (hb_serialize_context_t *c,
		     const OffsetTo& src,
		     const void *src_base,
		     hb_serialize_context_t::whence_t dst_base = hb_serialize_context_t::Head,
		     Ts&&... ds)
```
but doesn’t compile with clang due to its bug: https://bugs.llvm.org/show_bug.cgi?id=23029
Need to work around somehow. Sad.

blueshade7:
I have a strong suspicion that the base argument to `add_link()` is meaningfully used for the name table only, in other cases base is always left nullptr or set to `current->head`, but there are too many code paths to get there & not sure if the existing test cases cover them all.

I think I will just leave the existing behavior of base & bias for the case of whence=Head so we can revisit this later, and implement Tail & Absolute to ignore base and set bias = 0.

blueshade7:
Re-implemented Tail mode (whence) to base offsets on the object tail during resolution.
Rewrote name table serialization to use whence = Tail.

Currently base argument is ignored when whence = Tail or Absolute. It is still respected to calculate bias when whence = Tail, but I suspect in all cases base is set to current->head and could be removed altogether. Since this involves substantial code changes, it should be looked at as a separate issue.

blueshade7:
Accidentally forgot to respect bias & broke gpos4 and other subset tests to fail revealing cases where base is meaningfully used.

behdad:
Thanks for getting to the bottom of this Ariza!  I agree base was added to work around missing whence support.  I'm okay if we remove it and add bias argument later if necessary.  Or get this in as is if you think it's ready.  Thanks again!

src/hb-open-type.hh
```diff
@@ -338,6 +338,30 @@ struct OffsetTo : Offset<OffsetType, has_null>
...
+  /* Workaround clang bug: https://bugs.llvm.org/show_bug.cgi?id=23029
+   * Can't compile: whence = hb_serialize_context_t::Head followed by Ts&&...
+   */
+  template <typename ...Ts>
+  bool serialize_copy (hb_serialize_context_t *c,
+		       const OffsetTo& src,
+		       const void *src_base,
+		       const void *dst_base,
+		       hb_serialize_context_t::whence_t whence,
+		       Ts&&... ds)
```
behdad: This is a problem now since next function doesn't have whence but still has parameter pack!  Correct fix is for next function to have only defaulted whence and no parameter pack.

behdad: Oh, and obviously one of them should call the other.  Why would you ever duplicate code like that?!

behdad: Do we actually need this flavor?  It already takes src AND dst base.  Why does it need whence?!

blueshade7: AFAICT, second one is now used without parameter pack so it can be removed. And duplicating code is certainly not clean. I can change second to call first.

blueshade7: Second serialize_copy with defaulting whence without parameter pack chaining to first won't work, because second calls itself recursively.
```
  bool serialize_copy (hb_serialize_context_t *c,
		       const OffsetTo& src,
		       const void *src_base,
		       const void *dst_base,
		       hb_serialize_context_t::whence_t whence = hb_serialize_context_t::Head)
  { return serialize_copy (c, src, src_base, dst_base, whence); }
```
 I think I will remove whence as well, and specify Head as whence when chaining to first.

behdad: Right.  Make second not take whence at all and implicitly default.

behdad: Or whichever way you find.  Am happy as long  as not duplicated code.  Thanks!

blueshade7: Modified second definition of serialize_copy() to call first as I described above.
I believe this PR is ready to merge.

PR #2165, opened 2020-02-09
Add links to files used by python scripts.

### LINKED ISSUE #2150: Where to find files used by python scrips?
`gen-arabic-table.py` requires `ArabicShaping.txt UnicodeData.txt Blocks.txt`.

`gen-tag-table.py` requires `languagetags language-subtag-registry`.

etc.

But I could not find those files in the repo, neither links from where to download them.

Yes, after some googling you can find them, but maybe it would be better if links were included in the scripts itself? They can also auto-download them.

RazrFalcon:
Closes #2150

### DIFF
--- a/src/gen-arabic-table.py
+++ b/src/gen-arabic-table.py
@@ -1,5 +1,10 @@
 #!/usr/bin/env python
 
+# Input files:
+# - https://unicode.org/Public/12.0.0/ucd/ArabicShaping.txt
+# - https://unicode.org/Public/12.0.0/ucd/UnicodeData.txt
+# - https://unicode.org/Public/12.0.0/ucd/Blocks.txt
+
 from __future__ import print_function, division, absolute_import
 
 import io, os.path, sys
--- a/src/gen-emoji-table.py
+++ b/src/gen-emoji-table.py
@@ -1,5 +1,7 @@
 #!/usr/bin/python
 
+# Input file: https://www.unicode.org/Public/emoji/12.0/emoji-data.txt
+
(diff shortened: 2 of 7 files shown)


ebraminio:
Looks great! Do you believe we can somehow indicate also it is the latest version of the files but the address can be change for next release for future readers somehow? Maybe something like "Input files of the script for the latest unicode release:" or anything you see better or not at all.

jfkthame:
> Looks great! Do you believe we can somehow indicate also it is the latest version of the files but the address can be change for the unicode next release for future readers somehow? Maybe something like "Input files of the script for the latest unicode release:" or anything you see better or not at all.

Would it make more sense to point to the https://www.unicode.org/Public/UCD/latest/ location, rather than having to update links each time a new Unicode version is released?

ebraminio:
Didn't know that, makes sense! Sorry for the bikeshedding anyway, with such addresses, just a little indication that it was there as that specific unicode version, something like "Input files as Unicode 12:" maybe would be nice, considering the chance for the file rename in future releases, or maybe not.

RazrFalcon:
Maybe add something like: "Only Unicode 12.0 files are supported." ?

ebraminio:
On that case guessing direct links are better indicative :/ Or, let's merge and think about it more later. Thanks :)

behdad:
I don't understand why this is a good idea. Make it print in the usage output and don't link to specific Unicode version. What's there now is already stale.

ebraminio:
Sure, 6a390df

PR #2166, opened 2020-02-10
[subset] GDEF glyphClassDef subsetting support

qxliu76:
glyphClassDef uses the same ClassDef format. However, glyphClassDef table
uses predefined class values so we do not remap class values.

Also includes a fix for serialize ClassDef format


src/hb-ot-layout-common.hh
```diff
@@ -1574,15 +1574,25 @@ static void ClassDef_remap_and_serialize (hb_serialize_context_t *c,
                                           const hb_set_t &glyphset,
                                           const hb_map_t &gid_klass_map,
                                           hb_sorted_vector_t<HBGlyphID> glyphs,
```
garretrieger: This one can be changed to reference to so the glyphs vector doesn't need to be copied.

src/hb-ot-layout-common.hh
```diff
@@ -1574,15 +1574,25 @@ static void ClassDef_remap_and_serialize (hb_serialize_context_t *c,
...
-  bool has_no_match = glyphset.get_population () > gid_klass_map.get_population ();
+  if (!klass_map)
+  {
+    auto it =
+    + glyphs.iter ()
+    | hb_map_retains_sorting ([&] (const HBGlyphID& gid) -> hb_pair_t<hb_codepoint_t, unsigned>
+                              {
+                                return hb_pair ((hb_codepoint_t)gid, gid_klass_map[gid]);
+                              })
+    ;
```
garretrieger: I think you can also accomplish this with a zip:

hb_zip (+ glyphs.iter (), + glyphs.iter () | hb_map (gid_klass_map))

src/hb-ot-layout-common.hh
```diff
@@ -1574,15 +1574,25 @@ static void ClassDef_remap_and_serialize (hb_serialize_context_t *c,
...
+    ClassDef_serialize (c, it);
+    return;
+  }
 
-  hb_map_t m;
-  if (!klass_map) klass_map = &m;
+  if (glyphset.get_population () > gid_klass_map.get_population ())
```
garretrieger: Would it make more sense to check for the presence of a mapping for key 0 if present in glyphset and then add if it's missing?

qxliu76: There's case that if font file has 4 glyphs which are explicitly assigned to class 1,2,3,4 respectively, then we need a mapping like 1->0, 2->1, 3->2, 4->3.  
In this case 0->0 mapping is not needed.

qxliu76: oh we have a map from gid->class, but not have a class->gid map here

garretrieger: Can you add a comment here explaining the purpose of the check.

qxliu76: I think you're right we could check for the presence of a class 0 mapping. Given we only have gid->class mapping, we need to traverse the whole map or build a reverse map to check this. 

Another reason is to be consistent with fonttools:
https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/subset/__init__.py#L436

garretrieger: Ok, reviewed again and the current approach seems fine. No need to change it but can you add a comment though that explains what the check is for.

qxliu76: comments added

PR #2167, opened 2020-02-10
When serializing cmap14 order the offsets from smallest to largest.

garretrieger:
Current versions of OTS fail fonts with cmap 14's who's last offset does not point to the a block at the end of the table.

### DIFF
--- a/src/hb-array.hh
+++ b/src/hb-array.hh
@@ -171,6 +171,21 @@ struct hb_array_t : hb_iter_with_fallback_t<hb_array_t<Type>, Type&>
 
   unsigned int get_size () const { return length * this->get_item_size (); }
 
+  void reverse ()
+  {
+    int rhs = length - 1;
+    int lhs = 0;
+    while (rhs > lhs)
+    {
+      Type value_rhs = arrayZ[rhs];
+      Type value_lhs = arrayZ[lhs];
+      arrayZ[rhs] = value_lhs;
+      arrayZ[lhs] = value_rhs;
+      rhs--;
+      lhs++;
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


src/hb-ot-cmap-table.hh
```diff
@@ -953,16 +947,73 @@ struct CmapSubtableFormat14
...
+
+    // Correct the incorrect write order by reversing the order of the variation
+    // records array.
+    _reverse_variation_records ();
+
+    // Now that records are in the right order, we can set up the offsets.
+    _add_links_to_variation_records (c, obj_indices);
+  }
+
+  void _reverse_variation_records ()
```
behdad: This sounds like should be a `reverse` operation on `hb_array_t` or something similar.

garretrieger: A general purpose reverse on array_t wouldn't quite work here, since we're only reversing a single field in VariationSelectorRecord and not the whole record.

garretrieger: On second thought, we end up overwriting the other two fields of VariationSelector in the next method, so a general purpose reverse would work.

src/hb-array.hh
```diff
@@ -171,6 +171,21 @@ struct hb_array_t : hb_iter_with_fallback_t<hb_array_t<Type>, Type&>
...
   unsigned int get_size () const { return length * this->get_item_size (); }
 
+  void reverse ()
+  {
+    int rhs = length - 1;
+    int lhs = 0;
+    while (rhs > lhs)
+    {
+      Type value_rhs = arrayZ[rhs];
+      Type value_lhs = arrayZ[lhs];
```
behdad: This uses two temps. Please do with just one. Make buffer's reverse call this.  Add start/end so that works.

PR #2170, opened 2020-02-11
[draw] Implement hb_draw_area_pen_t

ebraminio:
Main idea here is to implement some operators for draw to see if we are happy enough with our abstractions.


ebraminio:
turned out totally out of requirements here, let's close.

Issue #1912, opened 2019-08-19
enabling introspection gives error


pnemade:
Check this scratch build https://koji.fedoraproject.org/koji/taskinfo?taskID=37139119
OR
compile latest release 2.6.0 as
[parag@toolbox harfbuzz-2.6.0]$ ./configure --disable-static --with-graphite2 --with-gobject=yes --enable-introspection=yes --prefix=/tmp
[parag@toolbox harfbuzz-2.6.0]$ make V=1

and this gives....

```
 /tmp/harfbuzz-2.6.0/src/hb-aat-layout.h:36: syntax error, unexpected identifier in 'HB_BEGIN_DECLS' at 'HB_BEGIN_DECLS'

/tmp/harfbuzz-2.6.0/src/hb-aat-layout.h:431: syntax error, unexpected '*', expecting ')' or ',' in 'hb_aat_layout_get_feature_types (hb_face_t *face,' at '*'
/tmp/harfbuzz-2.6.0/src/hb-aat-layout.h:433: syntax error, unexpected UNSIGNED, expecting identifier or '(' or '*' in '     unsigned int *feature_count, /* IN/OUT.  May be NULL. */' at 'unsigned'
/tmp/harfbuzz-2.6.0/src/hb-aat-layout.h:434: syntax error, unexpected typedef-name, expecting identifier or '(' or '*' in '     hb_aat_layout_feature_type_t *features /* OUT.     May be NULL. */);' at 'hb_aat_layout_feature_type_t'
/tmp/harfbuzz-2.6.0/src/hb-aat-layout.h:434: syntax error, unexpected ')', expecting ',' or ';' in '     hb_aat_layout_feature_type_t *features /* OUT.     May be NULL. */);' at ')'
/tmp/harfbuzz-2.6.0/src/hb-aat-layout.h:442: syntax error, unexpected identifier in '  hb_ot_name_id_t name_id;' at 'hb_ot_name_id_t'
/tmp/harfbuzz-2.6.0/src/hb-aat-layout.h:447: syntax error, unexpected '}' in '} hb_aat_layout_feature_selector_info_t;' at '}'
/tmp/harfbuzz-2.6.0/src/hb-aat-layout.h:452: syntax error, unexpected '*', expecting ')' or ',' in 'hb_aat_layout_feature_type_get_selector_infos (hb_face_t *face,' at '*'
[...]

PR #2172, opened 2020-02-11
[gintrospection] Fix g-ir-scanner syntax errors

Phaiax:
... when compiling with --with-gobject=yes --enable-introspection=yes

The file list given to `g-ir-scanner` is somehow position dependent. See first post in issue #1912 for the error output.
The errors disappear if hb-aat-layout.h and hb-ot-shape.h slide down the list.
(Note that this was no fatal failure. Some warnings remain.)

### DIFF
--- a/src/Makefile.sources
+++ b/src/Makefile.sources
@@ -189,7 +189,6 @@ HB_BASE_RAGEL_sources = \
 	$(NULL)
 
 HB_BASE_headers = \
-	hb-aat-layout.h \
 	hb-aat.h \
 	hb-blob.h \
 	hb-buffer.h \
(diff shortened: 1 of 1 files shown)


behdad:
Not a good idea to me.

ebassi:
The issue is not caused by the header position, but by the fact that we need to define symbols like `HB_H` in order to inspect the single headers; this breaks the `#include "hb.h"` directive in the sub-headers, and thus sub-headers end up not depending on the symbols they use.

Ideally, we'd have a `HB_COMPILATION` pre-processor symbol, and the single-header inclusion guard would look something like:

```cpp
#if !defined(HB_H_IN) && !defined(HB_COMPILATION)
#error "Include <hb.h>"
#endif
```

Then we'd be able to define `HB_COMPILATION` instead of `HB_H`

ebassi:
Another option is for every C header to include all the sub-headers it needs to include.

khaledhosny:
Closing in favor of https://github.com/harfbuzz/harfbuzz/pull/2853

PR #2173, opened 2020-02-12
Improve GIR build

ricotz:
This adds the proper c-header and pkg-config references to the GIR.
Also the correct symbol-prefix 'hb' is added instead of 'hb_'.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -628,12 +628,14 @@ if (HB_HAVE_INTROSPECTION)
     POST_BUILD
     COMMAND ${G_IR_SCANNER_CMD}
       --warn-all --no-libtool --verbose
-      -n hb
       --namespace=HarfBuzz
       --nsversion=0.0
+      --symbol-prefix=hb
+      --symbol-prefix=hb_gobject
       --identifier-prefix=hb_
       --include GObject-2.0
-      --pkg-export=harfbuzz
+      --pkg-export=harfbuzz-gobject
+      --c-include=hb-gobject.h
       --cflags-begin
       -I${PROJECT_SOURCE_DIR}/src
       -I${PROJECT_BINARY_DIR}/src
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -452,7 +452,16 @@ if HAVE_INTROSPECTION
 
 -include $(INTROSPECTION_MAKEFILE)
 INTROSPECTION_GIRS = HarfBuzz-0.0.gir # What does the 0 mean anyway?!
-INTROSPECTION_SCANNER_ARGS = -I$(srcdir) -n hb --identifier-prefix=hb_ --warn-all
+INTROSPECTION_SCANNER_ARGS = \
+	-I$(srcdir) \
+	--warn-all --verbose \
+	--namespace=HarfBuzz \
+	--nsversion=0.0 \
+	--symbol-prefix=hb \
+	--symbol-prefix=hb_gobject \
+	--identifier-prefix=hb_ \
+	--pkg-export=harfbuzz-gobject \
+	--c-include=hb-gobject.h
 INTROSPECTION_COMPILER_ARGS = --includedir=$(srcdir)
 INTROSPECTION_SCANNER_ENV = CC="$(CC)"
 


behdad:
lgtm. but can you please also summarize what problems this fixes?

behdad:
lgtm. but can you please also summarize what problems this fixes?

ricotz:
It would make sense to retire one of the existing buildsystems to avoid the duplicated efforts.

behdad:
> It would make sense to retire one of the existing buildsystems to avoid the duplicated efforts.

To me only autotools one is supported :P.

ricotz:
> > It would make sense to retire one of the existing buildsystems to avoid the duplicated efforts.
> 
> To me only autotools one is supported :P.

I see, so better drop cmake, and consider meson in the future.

ebraminio:
CMake support and the part you were tweaking which is added by @fanc999 was needed for Gnome on Windows AFAIK, we had nmake there for it, let's drop CMake when we got Meson.

ricotz:
@behdad I hope you noticed that I added the improved in the first post.

Would you consider adding a proper GObject wrapper?
Like https://github.com/ebassi/graphene/blob/master/src/graphene-gobject.c

behdad:
> Would you consider adding a proper GObject wrapper?

How is our hb-gobject different from that?

ricotz:
> > Would you consider adding a proper GObject wrapper?
> 
> How is our hb-gobject different from that

The name of structures are defined with camel-cased names.

e.g. `buffer_t` would be `Buffer`, `glyph_position_t` would be `GlyphPosition`

sdroege:
@ricotz That sounds like cairo and cairo-gobject, and that can be handled by bindings from what I understand :)

ricotz:
> @ricotz That sounds like cairo and cairo-gobject, and that can be handled by bindings from what I understand :)

Of course that can be handled by bindings. Still having a nicely formatted GIR with some type-and-function-relation is nice too. :-)

behdad:
> e.g. `buffer_t` would be `Buffer`, `glyph_position_t` would be `GlyphPosition`

That would invalidate our existing bindings.  I'm fine with that if you can propose a one-breakage solution.

behdad:
Then again, @ebraminio is the maintainer. :)

ricotz:
@behdad You can consider this merge finished. Actual gi-annotation fixes should be proposed separately.

ebraminio:
Thanks!

ebraminio:
@ricotz @sdroege @mcatanzaro @ebbasi have merged meson port, can you check if GIR support is OK? Thanks!

ricotz:
> @ricotz @sdroege @mcatanzaro @ebbasi have merged meson port, can you check if GIR support is OK? Thanks!

Hi, actually applying the changes of this PR would be a start. Afaics there are some enums and OT related symbols missing.

Simply comparing the built gir files of meson and automake is quite easy. (ignore the <source-position ...> attribute changes)

ebraminio:
> actually applying the changes of this PR would be a start. Afaics there are some enums and OT related symbols missing.

Ok, apparently https://github.com/harfbuzz/harfbuzz/blob/master/src/meson.build#L336-L352 is the equivalent place, have applied

```
diff --git a/src/meson.build b/src/meson.build
index 63bcf182..68e5ad3e 100644
--- a/src/meson.build
+++ b/src/meson.build
@@ -338,7 +338,10 @@ if have_gobject
       namespace: 'HarfBuzz',
       nsversion: '0.0',
       identifier_prefix: 'hb_',
-      symbol_prefix: 'hb_',
+      symbol_prefix: ['hb', 'hb_gobject'],
+      pkg_export: 'harfbuzz-gobject',
+      c_include: 'hb-gobject.h',
+      warn_all: true, verbose: true,
       includes: ['GObject-2.0'],
       install: true,
       extra_args:  ['--cflags-begin',
```

and it says and ran with `meson build -Dgobject=enabled` but it says `src/meson.build:358:2: ERROR: Unknown variable "hb_gen_files_gir".`, which well, happens before the patch also so apparently its gir is already broken, @tp-m

fanc999:
Hi,

Sorry, I didn't catch the message earlier... here comes PR 2255 for this regard...

PR #2174, opened 2020-02-13
add hb_unwrap_type

blueshade7:
to be used in a solution for #2156

### DIFF
--- a/src/hb-meta.hh
+++ b/src/hb-meta.hh
@@ -396,17 +396,16 @@ using hb_is_trivial= hb_bool_constant<
 >;
 #define hb_is_trivial(T) hb_is_trivial<T>::value
 
-/* hb_get_type (T)
- * Returns T::type if T::type is typedef'ed, or T otherwise.
+/* hb_unwrap_type (T)
+ * If T has no T::type, returns T. Otherwise calls itself on T::type recursively.
  */
 
 template <typename T, typename>
-struct _hb_get_type : hb_type_identity_t<T> {};
+struct _hb_unwrap_type : hb_type_identity_t<T> {};
 template <typename T>
-struct _hb_get_type<T, hb_void_t<typename T::type>> : hb_type_identity_t<typename T::type> {};
+struct _hb_unwrap_type<T, hb_void_t<typename T::type>> : _hb_unwrap_type<typename T::type, void> {};
 template <typename T>
-using hb_get_type = _hb_get_type<T, void>;
-#define hb_get_type(T) typename hb_get_type<T>::type
-
+using hb_unwrap_type = _hb_unwrap_type<T, void>;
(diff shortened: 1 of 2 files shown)


behdad:
Thanks.  Two comments:
1. Rename to `hb_unwrap_type()`.  Current name is too generic.
2. Recursively unwrap `T::type`.

blueshade7:
Not sure how I can make this work recursively.

src/hb-meta.hh
```diff
@@ -396,5 +396,17 @@ using hb_is_trivial= hb_bool_constant<
...
 #define hb_is_trivial(T) hb_is_trivial<T>::value
 
+/* hb_get_type (T)
+ * Returns T::type if T::type is typedef'ed, or T otherwise.
+ */
+
+template <typename T, typename>
+struct _hb_get_type : hb_type_identity_t<T> {};
+template <typename T>
+struct _hb_get_type<T, hb_void_t<typename T::type>> : hb_type_identity_t<typename T::type> {};
```
behdad: Recursive like this?
```
struct _hb_get_type<T, hb_void_t<typename T::type>> : hb_get_type<typename T::type>::type {};
```

blueshade7: It doesn't compile since hb_get_type is defined a line below.

behdad: > It doesn't compile since hb_get_type is defined a line below.

Then just expand it to `_hb_get_type<typename T::type, void>`.

behdad: See `hb_priority` for recursive templates example.

blueshade7: >Then just expand it to _hb_get_type<typename T::type, void>.
 
I tried that, but it doesn't compile since it is accessing `T::type::type` before `type` is known to exist at two-level deep.

`hb_priority` is a bit different, since an integer template argument is used as the recursion termination condition by evaluating its value. In this case, the termination condition is on a type argument.

As a workaround, we can define the external macro to call implementation at two (or more) level deep. Still ugly, but hide the ugliness of `hb_unwrap_type(hb_unwrap_type(T))` from its caller.

blueshade7: I think I figured out how to implement recursion.

Define another meta function hb_has_type(T) to return true/false depending on whether T has type. Use it in hb_enable_if whether to call hb_wrap_type implementation recursively.

I can try writing this code, but not sure if it’s worth complexity, when we need it only at a constant level deep, in this case two.

blueshade7: Never mind, this worked just fine:
```
template <typename T>
struct _hb_unwrap_type<T, hb_void_t<typename T::type>> : _hb_unwrap_type<typename T::type, void> {};
```
Updated PR, please re-review.

behdad:
Looks great.  Thank you!

PR #2175, opened 2020-02-15
Implement a rasterizer for HarfBuzz based on font-rs

ebraminio:
The glyph rendered on the test,
image

### DIFF
--- a/docs/harfbuzz-sections.txt
+++ b/docs/harfbuzz-sections.txt
@@ -113,6 +113,20 @@ hb_buffer_message_func_t
 </SECTION>
 
 <SECTION>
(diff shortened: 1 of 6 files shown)


ebraminio:
Should note that it won't support glyphs overlap AFAIK (and color currently) so maybe more indicative naming instead would be naming or ftrying to the issue somehow, something like hb-raster or hb-rasterizer or hb-renderer instead hb-canvas or maybe not.

behdad:
Not going to comment before we finish the outline API :).

ebraminio:
No feedback from Raph and the formula still not adopted for cubic curves, would be nice to have #2200 before this, close till resolve.

Issue #2164, opened 2020-02-08
[subset] pop_pack() option to not hash & share table


blueshade7:
As outlined in #1981, I am working on a task to modernize CFF/CFF2 subsetter using _subset2 without calculating table offsets in first pass, then actually serialize in second pass, as previously done with _subset.

Serializing INDEX table in a single pass is even possible by serializing all data first while calculating the minimum offSize required, then pop_pack() data so they move to tail of the serializer buffer, followed by serializing the INDEX header along with offsets using offSize calculated above.

Problem:

If I serialize sections of a contiguous table by calling pop_pack() piece-wise, some may be found in packed_map and actually not serialized, as the result a corrupted table is generated.

A possible workaround is to add a boolean flag to pop_pack() to not look it up in packed_map, so that it is surely serialized. This option may be also helpful to skip hashing a huge singleton sub-table unlikely shared, as discussed in above issue.

Ideas?

blueshade7:
I came to think I am trying to use a facility (object packing) for something it was not intended for (swapping order of sub-sections of a table), so I withdraw this issue. I will fix my problem with INDEX table locally somehow.

An option to not hash an unlikely shared table may be worth to have though.

blueshade7:
Let me reopen this. Struggling in some failed attempts, beginning to realize "not_to_map" pop_pack() option may be the easiest workaround to serialize FDArray in a single pass.

garretrieger:
Seems reasonable to me.

behdad:
I agree the not-sharing option is useful in a couple other places as well.  I'm happy to have that.  The objects should not be inserted in the sharing pool either in that case.  And yes, that sounds perfect solution for INDEX.

behdad:
I'm reopening this to reconsider and solve differently.  Seems like being able to pack objects that will be packed "immediate after" current object is useful in quite a few places.  Even your INDEX case can be addressed this way.  INDEX is interesting because the offset size is also only known at runtime.  I want to come up with a generic-enough solution to even address that.

So I'm thinking that instead of "dont-share" we want "pack_immediate" or "pack_right"after" mode.  "pack_immediate" name might be a bit misleading.  I'm leaning on "pack_after".  One would be able to add offsets to these or even not and the packer will pack that object right after.  Multiple "pack_after" objects will be packed in the order they are added.

[...]

PR #2176, opened 2020-02-16
[subset] add share option to pop_pack()

blueshade7:
a solution of #2164 [subset] pop_pack() option to not hash & share table

### DIFF
--- a/src/hb-serialize.hh
+++ b/src/hb-serialize.hh
@@ -195,7 +195,12 @@ struct hb_serialize_context_t
     obj->fini ();
     object_pool.free (obj);
   }
-  objidx_t pop_pack ()
+
+  /* Set share to false when an object is unlikely sharable with others
+   * so not worth an attempt, or a contiguous table is serialized as
+   * multiple consecutive objects in the reverse order so can't be shared.
+   */
+  objidx_t pop_pack (bool share=true)
   {
     object_t *obj = current;
     if (unlikely (!obj)) return 0;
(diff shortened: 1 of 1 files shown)


blueshade7:
I am not sure why we need to call `pop_pack()` in `end_serialize()` here, but if it is to stay, I suppose it needs share = `false`.
https://github.com/harfbuzz/harfbuzz/blob/a94d1af193ac1601a1725266dbb0dd51195af98b/src/hb-serialize.hh#L169

behdad:
> I am not sure why we need to call `pop_pack()` in `end_serialize()` here, but if it is to stay, I suppose it needs share = `false`.
> https://github.com/harfbuzz/harfbuzz/blob/a94d1af193ac1601a1725266dbb0dd51195af98b/src/hb-serialize.hh#L169

It doesn't *need* share =  false, but can use it, yes.

We need that there such that users of serialize that don't participate in push/pop at all still work.

LGTM.  Thanks.

blueshade7:
>We need that there such that users of serialize that don't participate in push/pop at all still work.

But this condition above the line appears to imply this pop_pack() is executed when push/pop_pack are used.
https://github.com/harfbuzz/harfbuzz/blob/a94d1af193ac1601a1725266dbb0dd51195af98b/src/hb-serialize.hh#L164-L167
copy_bytes() appears to take care of data never pushed/popped. This brought me to the question: why do we need pop_pack() there?

blueshade7:
Let me merge this PR and open an issue for above discussion.

PR #2179, opened 2020-02-18
[glyf] Don't send phantom points to gvar

ebraminio:
@blueshade7 is it okay to not send phantom points to gvar, which I guess, or it is passing the tests as it is just uncovered?


ebraminio:
Oh how I didn't see the failure locally, it made sense to not work of course, let me close to investigate.

PR #2180, opened 2020-02-18
[subset] add one empty byte to glyf table

qxliu76:
if it's going to end up empty after subsetting
This will make OTS happy and match what fontTools is doing now.

See also fonttools/fonttools#1829


ebraminio:
Thanks!

PR #2183, opened 2020-02-20
[draw] Implement hb_*_font_draw_glyph for platform libraries

ebraminio:
Maybe better to do this transparently without any extra API, maybe not.

Done for freetype / directwrite / gdi / coretext


behdad:
Okay trouble for you.  Looks to me like draw should be part of font_funcs :D.

ebraminio:
> Okay trouble for you. Looks to me like draw should be part of font_funcs :D.

#2185

ebraminio:
> Okay trouble for you. Looks to me like draw should be part of font_funcs :D.

Btw it is a very good idea as DirectWrite can 'draw' a hb_buffer_t like object but its callback has SetFillMode (D2D1_FILL_MODE_ALTERNATE | D2D1_FILL_MODE_WINDING) to circumstance glyph overlaps apparently.

ebraminio:
in favor of #2185

PR #2184, opened 2020-02-20
Fix glyph closure for alternate substitution.

garretrieger:
It was not filtering on the glyphs to be retained.


ebraminio:
Looks nice! Thanks

PR #2185, opened 2020-02-21
[draw] Add FreeType as a draw glyph backend

ebraminio:
#2183 spin off (that can be reconsidered after this), its naming doesn't make it useful for HB_FONT_FUNCS_IMPLEMENT_CALLBACKS however as it isn't a getter unlike the others and the fact we should carry two user_data, one for the specific draw call and as font funcs makes it a bit pain. One of my concern about the API is the decision of keeping or merging user_data of a specific draw call with the draw funcs which the situation here makes me a bit more undecided. WDYT

### DIFF
--- a/src/hb-draw.cc
+++ b/src/hb-draw.cc
@@ -24,13 +24,8 @@
 
 #include "hb.hh"
 
-#ifndef HB_NO_DRAW
-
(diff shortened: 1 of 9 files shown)


ebraminio:
Needs much more changes to become usable

Issue #2186, opened 2020-02-21
[subset] GSUB closure taking too long with a fuzzed font


blueshade7:
subset_fuzzer is taking long time with a fuzzed font: https://github.com/harfbuzz/harfbuzz/blob/master/test/fuzzing/fonts/clusterfuzz-testcase-minimized-hb-subset-fuzzer-5763024094232576

As the result gcc-valgrind bot often times out as seen in: https://circleci.com/gh/harfbuzz/harfbuzz/127382?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Most of time appears to be spent in a while loop in `hb_ot_layout_lookups_substitute_closure` iterating up to 27 times.
https://github.com/harfbuzz/harfbuzz/blob/07504569d5e7ca2f7da987fc016fc4141eeca263/src/hb-ot-layout.cc#L1529

This is especially wasteful with this fuzzed font, because its `face->num_glyphs` is actually zero, i.e., containing no valid glyphs.

ebraminio:
We should apply or lower some of our limits to alleviate the issue. Same goes for https://crbug.com/oss-fuzz/19802 and https://crbug.com/oss-fuzz/18529 I think, can you have a look at them to find some common pattern for limiting?

blueshade7:
They all appear to be different issues, but 19802 may be similar in that GSUB subsetter is trying to process thousands of glyphs when num_glyphs=32. We may throw out gid >= num_glyphs. But We can’t prevent fuzzer to come up with a font containing a very large number as num_glyphs. Actually 18529 has num_glyphs=16384.

blueshade7:
Actually oss-fuzz/19802 is different, `hb_ot_layout_collect_features` is collecting thousands of feature indices from thousands of scripts.

ebraminio:
> They all appear to be different issues [..]
> Actually oss-fuzz/19802 is different, [..]

Correct. That is somehow what I meant as I think limiting operations to sane and usual cases is our only approach here like what happened in shaping also.

blueshade7:
I think each case needs to be taken a look at separately for a different solution.
Created PR#2188 where `hb_closure_context_t::flush()` removes all glyphs above num_glyphs.

ebraminio:
> I think each case needs to be taken a look at separately [..]

[...]

PR #2188, opened 2020-02-23
remove invalid glyphs from closure

blueshade7:
fix issue #2186 [subset] GSUB closure taking too long with a fuzzed font

### DIFF
--- a/src/hb-ot-layout-gsubgpos.hh
+++ b/src/hb-ot-layout-gsubgpos.hh
@@ -114,6 +114,7 @@ struct hb_closure_context_t :
 
   void flush ()
   {
+    hb_set_del_range (output, face->get_num_glyphs (), hb_set_get_max (output));	/* Remove invalid glyphs. */
     hb_set_union (glyphs, output);
     hb_set_clear (output);
   }


src/hb-ot-layout-gsubgpos.hh
```diff
@@ -114,6 +114,7 @@ struct hb_closure_context_t :
 
   void flush ()
   {
+    hb_set_del_range (output, face->get_num_glyphs (), hb_set_get_max (output));	/* Remove invalid glyphs. */
```
ebraminio: What if the same font states higher num glyphs, I think we need this kind of https://github.com/harfbuzz/harfbuzz/commit/baf7779d2d6e4810168a8f036bbf8f9e6493dd1a limitations rather than depending on details fonts still can change.

blueshade7: `hb_ot_layout_lookups_substitute_closure ` already has such a limit `HB_CLOSURE_MAX_STAGES` but the current value 32 does not prevent fuzzer from timing out. Are you suggesting to lower it?
My concern about doing so is that it might start rejecting valid fonts.

Throwing out invalid glyphs has no such concern. It is cheap and has no downside. Yes, if a fuzzed font has a large num_glyphs, this check can't prevent it from timing out again. Probably we should do both.

ebraminio: We should be highly careful about not rejecting valid fonts, if you don't see another limitation possible for now and see this as last resort, it is ok and please merge it. Thank you very much

ebraminio: Meant I am *not* in favor of :D

garretrieger:
I would avoid using hb_set_del_range() as is. It's current implementation is to call delete on every value in the range (a, b). So if we have one invalid glyph, 0xFFFF, it will cause del() to be called ~65,000 times.

Recommend fixing the implementation of hb_set_del_range() or changing the code here to iterate through the set starting from num glyphs and call delete on each value individually.

blueshade7:
Good point. I wonder if it is worth re-implementing`hb_set_del_range()` that way.

garretrieger:
Yeah, sounds good.

behdad:
Fixing the implementation sounds good.  `del_range` is a good candidate for also purging pages that become obviously empty in the middle.

blueshade7:
Opened an issue #2193

PR #2192, opened 2020-02-24
[draw] Add a fuzzer

ebraminio:
Specially checks correctness of the API semantics:
* no move happens when a path is already opened with move-to.
* no path will be left open and close-path will happen at the end of opened paths.
* no path opens with a move-to and will be closed with no length.
* paths start and ending points matches.
* no line/quadratic/cubic command will be issued when no path is started.

As the recent fixes specially the fuzzer <s>locally looks quite good!</s> (update: needs more work) Guess however there will be some failures when ran on clouds, let's see if we can fix them.


ebraminio:
Have found and fixed interesting things along the way (1f5a54c 96b71e8 acc2d47 2f97aa6) and now it passes all the fonts we have in the repo itself (without that gid limit that the current code has), it practices gvar and CFF2 variation also which apparently are missing in shape fuzzer currently.

Issue #2193, opened 2020-02-24
Reimplement hb_set_t::del_range()


blueshade7:
The current implementation of `hb_set_t::del_range()` is inefficient especially when there are few values to delete in the specified range.
Also, when all values are deleted from a page, it is left empty. It should be removed instead.

garretrieger:
After you remove pages the compact() function in hb-set.hh can be used to correct the page map and shrink the pages array. See process() for an example of how to use.

behdad:
> After you remove pages the compact() function in hb-set.hh can be used to correct the page map and shrink the pages array. See process() for an example of how to use.

I was thinking doing without that, just remove pages that fall completely in the range to be deleted.  That way the runtime of the del_range won't be small if the range is small.

garretrieger:
Oh, got it. Agreed, that sounds like a good idea.

blueshade7:
pushed a tentative fix to branch set-del-range then fuzzed font #2186 timed out for me... karma?

PR #2194, opened 2020-02-25
Re-implement hb_set_t::del_range

blueshade7:
Re-implemented `del_range()` as outlined in the issue #2193,  modeled upon the existing function `add_range()`.

- If a range covers entire pages, they are deleted from `pages` vector and also from `page_map` by calling `compact()`.
- Added `page_t::del_range()`modeled upon `page_t::add_range()` for hb_set_t::del_range()` to call to delete a sub-range within the corresponding page.
- Added a function `test_set_delrange()` to API test `test-set.c` to exercise & verify various cases of range deletion.

### DIFF
--- a/src/hb-set.hh
+++ b/src/hb-set.hh
@@ -89,6 +89,23 @@ struct hb_set_t
       }
     }
 
+    void del_range (hb_codepoint_t a, hb_codepoint_t b)
+    {
+      elt_t *la = &elt (a);
+      elt_t *lb = &elt (b);
+      if (la == lb)
+	*la &= ~((mask (b) << 1) - mask(a));
+      else
+      {
+	*la &= mask (a) - 1;
+	la++;
+
+	memset (la, 0, (char *) lb - (char *) la);
+
+	*lb &= ~((mask (b) << 1) - 1);
+      }
+    }
+
     bool is_equal (const page_t *other) const
     {
       return 0 == hb_memcmp (&v, &other->v, sizeof (v));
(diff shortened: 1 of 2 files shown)


src/hb-set.hh
```diff
@@ -366,14 +383,48 @@ struct hb_set_t
...
+      }
+    }
+    if (de < (int)mb && ma != mb)
+    {
+      page_t *page = page_for (b);
+      if (page)
+	page->del_range (major_start (mb), b);
+    }
+    if (ds <= de)
+    {
```
garretrieger: The contents of this if could be extracted into a function, del_pages(ds, de) which would help improve readability of this method.

src/hb-set.hh
```diff
@@ -366,14 +383,48 @@ struct hb_set_t
...
-    /* TODO Optimize, like add_range(). */
     if (unlikely (!successful)) return;
-    for (unsigned int i = a; i < b + 1; i++)
-      del (i);
+    if (unlikely (a > b || a == INVALID || b == INVALID)) return;
+    dirty ();
+    unsigned int ma = get_major (a);
+    unsigned int mb = get_major (b);
+    /* Delete pages from ds through de if ds <= de. */
+    int ds = (a == major_start (ma))? (int)ma: (int)(ma + 1);
```
garretrieger: Spacing: (int)ma should be (int) ma. Here and elsewhere in this function.

test/api/test-set.c
```diff
@@ -470,6 +470,36 @@ test_set_empty (void)
   hb_set_destroy (b);
 }
 
+static void
+test_set_delrange (void)
```
garretrieger: Can you add coverage of a few more interesting scenarios:
- The case where [a, b] covers exactly one page and also where [a, b] covers exactly two or more pages.
- Del range where there is no page at a or b.

blueshade7: Updated impl & added more test cases.

PR #2197, opened 2020-02-26
Set hb_buffer_t to use array_t.reverse().

garretrieger:


src/hb-array.hh
```diff
@@ -173,16 +173,25 @@ struct hb_array_t : hb_iter_with_fallback_t<hb_array_t<Type>, Type&>
...
-    int lhs = 0;
-    while (rhs > lhs)
-    {
-      Type value_rhs = arrayZ[rhs];
-      Type value_lhs = arrayZ[lhs];
-      arrayZ[rhs] = value_lhs;
-      arrayZ[lhs] = value_rhs;
-      rhs--;
-      lhs++;
+    return reverse (0, length);
```
behdad: Make start/end have defaults so you don't need separate override.

src/hb-array.hh
```diff
@@ -173,16 +173,25 @@ struct hb_array_t : hb_iter_with_fallback_t<hb_array_t<Type>, Type&>
...
+
+  /*
+   * Reverse the order of items in this array in the range [start, end).
+   */
+  void reverse (unsigned start, unsigned end)
+  {
+    start = hb_min (start, length);
+    end = hb_min (end, length);
+
+    if (end - start < 2)
```
behdad: Problematic if end < start.

src/hb-buffer.cc
```diff
@@ -448,27 +448,15 @@ void
...
 
-  for (i = start, j = end - 1; i < j; i++, j--) {
-    hb_glyph_info_t t;
-
-    t = info[i];
-    info[i] = info[j];
-    info[j] = t;
-  }
+  hb_array_t<hb_glyph_info_t> info_array(info, len);
+  info_array.reverse (start, end);
```
behdad: Just `hb_array (info, len).reverse (start, end);`. Same below.

src/hb-array.hh
garretrieger: Done.

src/hb-buffer.cc
garretrieger: Done.

src/hb-array.hh
garretrieger: Done.

src/hb-array.hh
```diff
@@ -171,18 +171,21 @@ struct hb_array_t : hb_iter_with_fallback_t<hb_array_t<Type>, Type&>
 
   unsigned int get_size () const { return length * this->get_item_size (); }
 
-  void reverse ()
+  /*
+   * Reverse the order of items in this array in the range [start, end).
+   */
+  void reverse (unsigned start = 0, unsigned end = -1)
```
behdad: Maybe use `UINT_MAX` or `(unsigned) -1`.  I don't know which one we're using these days.

src/hb-buffer.cc
```diff
@@ -448,27 +448,13 @@ void
...
     return;
 
-  for (i = start, j = end - 1; i < j; i++, j--) {
-    hb_glyph_info_t t;
-
-    t = info[i];
-    info[i] = info[j];
-    info[j] = t;
-  }
+  hb_array_t<hb_glyph_info_t> (info, len).reverse (start, end);
```
behdad: Again.  I meant `hb_array (info, len).reverse (start, end);`.  `hb_array` is a function that deduces the template type for you so you don't have to type it.

PR #2200, opened 2020-02-27
[draw] Implement bezier to lines when cubic callback isn't set

ebraminio:
We translate quad call to cubic already so translating cubic calls to lines means translating both to lines, but maybe we can use different approach.

### DIFF
--- a/src/hb-draw.hh
+++ b/src/hb-draw.hh
@@ -36,6 +36,7 @@ struct hb_draw_funcs_t
   hb_draw_quadratic_to_func_t quadratic_to;
   bool is_quadratic_to_set;
   hb_draw_cubic_to_func_t cubic_to;
+  bool is_cubic_to_set;
   hb_draw_close_path_func_t close_path;
 };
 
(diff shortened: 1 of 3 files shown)


src/hb-draw.hh
```diff
@@ -95,7 +100,35 @@ struct draw_helper_t
 	equal_to_current (to_x, to_y))
       return;
     if (!path_open) start_path ();
-    funcs->cubic_to (control1_x, control1_y, control2_x, control2_y, to_x, to_y, user_data);
+    if (funcs->is_quadratic_to_set)
+      funcs->cubic_to (control1_x, control1_y, control2_x, control2_y, to_x, to_y, user_data);
+    else
+    {
+      /* just an inaccurate one based https://github.com/raphlinus/font-rs/blob/master/src/raster.rs#L100 */
```
ebraminio: @raphlinus Have used you quad-to-lines translation code, the formulas below aren't good for it I think, can you instruct me on how to improve and letting me know if I am ok to reuse the code here. Thanks

ebraminio:
No feedback from Raph and the formula still not adopted for cubic curves, close till resolve.

PR #2203, opened 2020-02-27
[subset] Add source_blob as a hb_subset_context_t field

ebraminio:
So no more double sanitizing source table.

Maybe there was a reason that this was left out in the original design, let's find out.


ebraminio:
@garretrieger @blueshade7 any thought? No problem if you don't like it, I will just close on that case.

ebraminio:
tyvm :)

PR #2205, opened 2020-02-27
Limit the number of feature indices processed during feature collection.

### LINKED ISSUE #2191: [subset] feature closure takes long time with a fuzzed font
This issue was originally reported by oss-fuzz as: https://crbug.com/oss-fuzz/19802
Also mentioned in #2186.

subset-fuzzer takes a few minutes to complete when run locally with the fuzzed font: 
[clusterfuzz-testcase-minimized-hb-subset-fuzzer-6276691949518848.zip](https://github.com/harfbuzz/harfbuzz/files/4243304/clusterfuzz-testcase-minimized-hb-subset-fuzzer-6276691949518848.zip)

Time appears to be spent in function:
https://github.com/harfbuzz/harfbuzz/blob/f00eb4ebfa02dea593c4842a7672aa2640a6bbf2/src/hb-ot-layout.cc#L1253
trying to collect all features from all languages, each fuzzed to become a very large number.

Since OT spec defines 600+ languages and 130+ features at this time of writing, probably it would make sense to set a limit on each, with some extra room to grow.

garretrieger:
To fix #2191 .


garretrieger:
I set the limit pretty high, we can tune it down later if needed. Looking at a merged copy of Noto Sans which contains support for everything other than CJK it uses ~500 feature indices.

ebraminio:
Please put the font that shows the issue on test/fuzzing/fonts :)

garretrieger:
Done :)

garretrieger:
Test case is still timing out, I'll try lowering the limit.

garretrieger:
Did some profling, now a lot of time is being spent in closure features (as opposed to collect features), probably needs some limiting applied in there as well.

garretrieger:
Now Script subsetting is the most expensive part, will need to apply feature limiting there too I suspect.

garretrieger:
So with all the fixes, it goes from taking ~5s to run the fuzzer file to ~0.08s.

PR #2206, opened 2020-02-27
[subset] GDEF Variation Store: step 1

qxliu76:
collect device varidxes, create old->new varidx mapping and store them in subset-plan.
Tests are yet to be added.

### DIFF
--- a/src/hb-ot-layout-common.hh
+++ b/src/hb-ot-layout-common.hh
@@ -150,6 +150,28 @@ struct hb_subset_layout_context_t :
   unsigned lookup_index_count;
 };
 
+struct hb_collect_variation_indices_context_t :
+       hb_dispatch_context_t<hb_collect_variation_indices_context_t, hb_empty_t, 0>
+{
+  const char *get_name () { return "CLOSURE_LAYOUT_VARIATION_IDXES"; }
+  template <typename T>
+  return_t dispatch (const T &obj) { obj.collect_variation_indices (this); return hb_empty_t (); }
+  static return_t default_return_value () { return hb_empty_t (); }
+
+  hb_set_t *layout_variation_indices;
+  const hb_set_t *glyph_set;
+  const hb_map_t *gpos_lookups;
+  unsigned int debug_depth;
+
+  hb_collect_variation_indices_context_t (hb_set_t *layout_variation_indices_,
+					  const hb_set_t *glyph_set_,
+					  const hb_map_t *gpos_lookups_) :
+					layout_variation_indices (layout_variation_indices_),
+					glyph_set (glyph_set_),
+					gpos_lookups (gpos_lookups_),
+					debug_depth (0) {}
+};
+
 template<typename OutputArray>
 struct subset_offset_array_t
 {
@@ -2916,6 +2938,12 @@ struct VariationDevice
     return_trace (c->embed<VariationDevice> (this));
   }
 
+  void record_variation_index (hb_set_t *layout_variation_indices) const
+  {
+    unsigned var_idx = (outerIndex << 16) + innerIndex;
+    layout_variation_indices->add (var_idx);
+  }
+
   bool sanitize (hb_sanitize_context_t *c) const
   {
     TRACE_SANITIZE (this);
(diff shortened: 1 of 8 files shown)


ebraminio:
`varidxes` naming feels too cryptic, `variation_indexes` maybe is better, and not sure about _create naming also. As we didn't have APIs similar to ones proposed here, can we avoid having this as a public API and somehow use them just inside subset module I wonder. And is this limited to GDEF and GPOS don't have GSUB equivalent? @behdad please have a look also. Thanks

qxliu76:
> This is collecting up all variation indices referenced from any GPOS lookups or the GDEF lig caret list. However, we'll need to treat this like other subsetting tasks and only collect variation indices from lookups and glyphs that are going to be retained in the final subset.
> 
> So to that end, I think it would be more appropriate to call it closure_varidxes instead of collect_varidxes. The method will probably need to take in a subset plan so var indices referenced from dropped lookups and glyphs can be filtered out and not added to the output set.
> 
> Roughly, you'll want to loop through each lookup id in plan->gpos_lookups and call the closure_varidxes method on each one. Inside each lookup you'll then need to filter to plan->_glyphset before collecting the indices from the device tables.
> 
> Since this will make the closure and mapping specific to subsetting I agree with Ebrahim that we don't need to make these methods part of the public API.

renamed method, added filtering based on gpos_lookups/glyphset. 
Not able to move these methods out of hb-ot-layout.cc yet. Subset-plan.cc can not access accelerator_t of GDEF and GPOS, whose init () calls is_blacklisted () function and that function is declared as HB_INTERNAL in gdef-table.hh, but its definition exists in hb-ot-layout.cc.

ebraminio:
> Subset-plan.cc can not access accelerator_t of GDEF and GPOS, whose init () calls is_blacklisted () function and that function is declared as HB_INTERNAL in gdef-table.hh, but its definition exists in hb-ot-layout.cc.

Oh, is that the only blocker? If so I think we can move the definition of them to some header, perhaps -gsubgpos.hh so it can be used on hb-subset also, other options would be improving and accepting this public API or merging hb-subset into hb and finalizing its API.

qxliu76:
> > Subset-plan.cc can not access accelerator_t of GDEF and GPOS, whose init () calls is_blacklisted () function and that function is declared as HB_INTERNAL in gdef-table.hh, but its definition exists in hb-ot-layout.cc.
> 
> Oh, is that the only blocker? If so I think we can move the definition of them to some header, perhaps -gsubgpos.hh so it can be used on hb-subset also, other options would be improving and accepting this public API or merging hb-subset into hb and finalizing its API.

I tried moving definition to gdef.hh file, still doesn't work. There is circular dependency issue. 
e.g:
in GDEF::is_blacklisted(),  code "face->table.GSUB->table.get_length () " depends on gsub.hh
and gsub.hh depends on gsubgpos.hh, which then depends on gdef.hh again...

qxliu76:
added missing  closure code for SinglePos and PairPos, could you help review?  @ebraminio @garretrieger

behdad:
What's the status of this?  Looks generally okay.  I'll comment on it in another issue I'm writing about layout subsetting reorganization.  The public API needs to be removed.

This doesn't actually subset the VariationStore, right?

behdad:
Again.  Naming.  There's nothing you are "closing over".  You are just collecting.

Add a `collect_variation_indices` to GDEF and GPOS.  (I know Ebrahim sent you that direction but I would have been fine using `collect_varidxes` internally.  We use `varidx` as something like `gid`.  Though internally we also don't use gid in naming and prefer glyphs unless there's a clear distinction necessary.)

After that, the actual remapping I like us to discuss more.  Because the remap logic you have reflects a very specific subset implementation of VarStore.  For example in fonttools we do a full optimization on VarStore, which would generate a different mapping.  So I'm not comfortable with the mapping being distinct from the subsetting.

I'm still making up my mind about how to resolve that.  And will share my thoughts soon on https://github.com/harfbuzz/harfbuzz/issues/2367

In the mean time, is easier to have self-contained chunks of code (like the collect method) to be merged in while we figure out the best way to subset VariationStore.

Thanks for this!

behdad:
> Again. Naming. There's nothing you are "closing over". You are just collecting.

Just so we are on the same page, when we talk about closure we mean this:
https://en.wikipedia.org/wiki/Closure_(mathematics)

Ie. we have a set and an operation, where calling operation with any item already in the set, produces a set of items.  We keep adding those until the set is closed over operation.

If that doesn't happen, then it shouldn't be called closure.  For example, if you start with a set of lookup indices and produce a set of feature indices that intersect with those, that's not a closure operation.

qxliu76:
I just resolved conflicts so it can be merged. I'll work on the naming.

PR #2207, opened 2020-02-28
reimplment serialize_int using check_assign()

blueshade7:
Another step of issue #1981, rewriting CFF/CFF2 subsetter using _subset2.
Rewrote UnsizedByteStr::serialize_int() using hb_serialize_context_t::check_assign()<T1, T2> now accepts a signed type as T1.

serialize_offset4_op() and serialize_offset2_op() will be later reimplemented with add_link() but for now they call into serialize_int().


ebraminio:
Looks good actually. Thanks!

Issue #2204, opened 2020-02-27
A shape fuzzer timeout case


ebraminio:
I have apparently <s>a recent regression</s> (maybe not) of an infinite loop with this font,

[1746cad6bc3fb2b355db50a5af37c9b58d9ad376.zip](https://github.com/harfbuzz/harfbuzz/files/4263528/1746cad6bc3fb2b355db50a5af37c9b58d9ad376.zip)

`clang test/fuzzing/hb-shape-fuzzer.cc test/fuzzing/main.cc -Isrc src/harfbuzz.cc -DHB_NO_MT -lm -fno-exceptions -g -Og && ./a.out 1746cad6bc3fb2b355db50a5af37c9b58d9ad376`

```
Program received signal SIGINT, Interrupt.
0x00005555555b8ee5 in hb_set_t::get_major (this=0x555555710f70, 
    g=<optimized out>) at src/hb-set.hh:876
876	  unsigned int get_major (hb_codepoint_t g) const { return g / page_t::PAGE_BITS; }
(gdb) bt
#0  0x00005555555b8ee5 in hb_set_t::get_major (this=0x555555710f70, 
    g=<optimized out>) at src/hb-set.hh:876
#1  0x00005555555b8f12 in hb_set_t::page_for_insert (this=0x555555710f70, 
    g=361016) at src/hb-set.hh:843
#2  0x0000555555594bd5 in hb_set_t::add (this=0x555555710f70, g=361016)
    at src/hb-set.hh:289
#3  0x00005555555a16f2 in OT::CmapSubtableLongSegmented<OT::CmapSubtableFormat12>::collect_unicodes (this=<optimized out>, out=0x555555710f70)
    at src/hb-ot-cmap-table.hh:554
#4  0x00005555555565d7 in test_font (font=0x55555570fb00, cp=1744896000)
    at test/fuzzing/../api/test-ot-face.c:48
#5  0x0000555555556532 in LLVMFuzzerTestOneInput (data=<optimized out>, 
    size=23293) at test/fuzzing/hb-shape-fuzzer.cc:53
#6  0x0000555555556a2d in main (argc=2, argv=0x7fffffffdc58)
    at test/fuzzing/main.cc:15
```
Pretty sure this is a recent regression, can you @garretrieger @blueshade7 have a quick look at it? Thanks

blueshade7:
[...]

PR #2212, opened 2020-02-28
issue #2204, collect_unicodes() with clamp, calling add_range()

blueshade7:
Another attempt to fix the timeout issue with a fuzzed cmap sub-table 12 by clamping each code range using num_glyphs.
Also replaced multiple `hb_set_t::add()` calls with  a single call to`hb_set_t::add_range()` for each range for a better performance.
cmap sub-table format 13 is uniformly handle by the fix, with a minor special handling (we need a test case for cmap 13).


ebraminio:
Please put the font itself on the change :)

ebraminio:
tyvm :)

PR #2214, opened 2020-02-29
[gvar] Use hb_iterator_t for tuple_iterator_t

ebraminio:
But fails when I use `for (auto &tuple : iterator)` instead `for (; iterator.__more__ (); iterator.__next__ ())`, aren't they supposed to be same? Need help here I think


ebraminio:
nvm

PR #2219, opened 2020-03-02
Limit OT::Lookup subtables

ebraminio:
Fixes https://crbug.com/oss-fuzz/14626

Closes #1647, an alternative fix for it that does the check font wide.


ebraminio:
Thank you Garret. I am pretty sure Behdad will have complains about it however 😂

PR #2221, opened 2020-03-03
[var] Fix hb_ot_var_get_axis_infos's offset semantic

ebraminio:
The API was adding offset to input's infos buffer index also which is
unusual between our APIs and wrong.

### DIFF
--- a/src/hb-ot-var-fvar-table.hh
+++ b/src/hb-ot-var-fvar-table.hh
@@ -150,17 +150,8 @@ struct fvar
   {
     if (axes_count)
     {
-      /* TODO Rewrite as hb_array_t<>::sub-array() */
-      unsigned int count = axisCount;
-      start_offset = hb_min (start_offset, count);
-
-      count -= start_offset;
-      axes_array += start_offset;
-
-      count = hb_min (count, *axes_count);
-      *axes_count = count;
-
-      for (unsigned int i = 0; i < count; i++)
+      hb_array_t<const AxisRecord> arr = hb_array (&(this+firstAxis), axisCount).sub_array (start_offset, axes_count);
+      for (unsigned i = 0; i < arr.length; ++i)
 	get_axis_deprecated (start_offset + i, axes_array + i);
     }
     return axisCount;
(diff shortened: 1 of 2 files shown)


src/hb-ot-var-fvar-table.hh
```diff
@@ -173,17 +164,8 @@ struct fvar
   {
     if (axes_count)
     {
-      /* TODO Rewrite as hb_array_t<>::sub-array() */
-      unsigned int count = axisCount;
-      start_offset = hb_min (start_offset, count);
-
-      count -= start_offset;
-      axes_array += start_offset;
```
ebraminio: This is the part that was wrong, start_offset isn't intended to be added to axes_array index but just used just as a AxisRecord index offset.

test/api/test-var-coords.c
```diff
@@ -67,10 +67,29 @@ test_get_var_coords (void)
...
+
+  unsigned num_coords = hb_ot_var_get_axis_count (face);
+  hb_ot_var_axis_info_t info;
+  unsigned c = 1;
+
+  hb_ot_var_get_axis_infos (face, 0, &c, &info);
+  g_assert (info.tag == HB_TAG ('w','g','h','t'));
+
+  hb_ot_var_get_axis_infos (face, 1, &c, &info);
+  g_assert (info.tag == HB_TAG ('w','d','t','h'));
```
ebraminio: This didn't work…

ebraminio:
@khaledhosny @behdad What do you suggest? Do you know clients that were using offset of the API? This overflows buffer for clients who iterating axis in usual way of HarfBuzz API.

ebraminio:
Pango fortunately doesn't use its offset, guess no client would use as it is broken anyway.

ebraminio:
Any pointers?

behdad:
Yeah I think is safe to fix.

No idea why I didn't just use sub_array to begin with...  But yeah I put TODOs where I insert bugs apparently!

ebraminio:
Thank you :)

PR #2223, opened 2020-03-03
[aat] Consume glyph insertion from buffer's max_ops

ebraminio:
Glyph insertion is an expensive operation and we like to have it limited
based on buffer's input size which is handled by buffer's max_ops.

clusterfuzz-testcase-minimized-harfbuzz_fuzzer-5754958982021120:

Before the change: 0.67s user 0.00s system 99% cpu 0.674 total
 After the change: 0.02s user 0.00s system 98% cpu 0.024 total

Which takes much longer on valgrind and tsan bots.

What makes this expensive is specially the fact multiple insertion making lots of glyphs and then the happened rearrangement becomes much expensive.


ebraminio:
@jfkthame can you have a look please?

jfkthame:
> @jfkthame can you have a look please?

Yeah, seems fine - I see behdad already merged it. :)

ebraminio:
Thank you :)

PR #2226, opened 2020-03-03
CFF/CFF2 subsetters in one-pass

blueshade7:
Updated CFF/CFF2 subsetters for issue #1981 Move CFF subsetters to _subset2 and make it the default.

- Tables are now serialized in one-pass.
- Table offsets are set by calling `add_link()` specifying `whence` as either `Head` by default or `Absolute` as necessary.
- Simplified code by using iterator & C++11 idioms here and there.
- Some API test results needed update due to subtle behavioral changes.

Sorry for the size of the changes. I tried to split CFF and CFF2 in two PRs but gave up as they share lots of code.
Some unused code is left there for later removal.


src/hb-subset-cff1.cc
```diff
@@ -883,186 +701,204 @@ struct cff_subset_plan {
...
     {
-      DEBUG_MSG (SUBSET, nullptr, "failed to serialize CFF name INDEX");
+      c->pop_discard ();
       return false;
     }
   }
 
-  /* top dict INDEX */
+  /* FDArray (FD Index) */
+  if (acc.fdArray != &Null(CFF1FDArray))
```
ebraminio: Can we just check array's count?

src/hb-subset-cff1.cc
```diff
@@ -883,186 +701,204 @@ struct cff_subset_plan {
...
     {
-      DEBUG_MSG (SUBSET, nullptr, "failed to serialize CFF top dict");
+      c->pop_discard ();
       return false;
     }
   }
 
-  /* String INDEX */
+  /* FDSelect */
+  if (acc.fdSelect != &Null(CFF1FDSelect))
```
ebraminio: ditto

src/hb-subset-cff1.cc
blueshade7: It didn't occur to me accessing a null obj for non-error cases. Yes, it should work and I can make that change.

src/hb-subset-cff1.cc
blueshade7: Unlike FDArray, FDSelect is not an array, has no count field.

src/hb-subset-cff1.cc
ebraminio: > It didn't occur to me accessing a null obj for non-error cases.

Yes, we prefer always doing so, handling error just like empty structs.

src/hb-subset-cff1.cc
ebraminio: Meta discussion, seeing places that we have a comparison to Null without no choice making me believe we need something akin to std::optional also maybe.

ebraminio: Or just using nullptr instead null pool.

So guess no other change is needed here, let's merge and improve them later.

Thanks :)

PR #2230, opened 2020-03-05
Removed unused code from CFF/CFF2 subsetters

blueshade7:
Removed code no longer used as the result of one-pass CFF/CFF2 subsetters PR#2226.


ebraminio:
Looks very good! Thanks :)

Issue #2227, opened 2020-03-04
revisit hb_serialize_context_t::add_link() base arg


blueshade7:
While working on #2156, it turned out that currently `base` argument to `add_link()` is in many cases set to be the same as `current->head` when `whence` is the default `Head`. This means that `base` could  be done without, the default value `nullptr` could be assumed instead.

See if there are cases where base is meaningfully specified. If so, consider replacing `base` as a pointer with `bias` as an offset, so that it can be optionally specified also for two other `whence` modes `Tail` and `Absolute`, for which `base` is currently ignored.

blueshade7:
As an experiment, commented out base argument in add_link() calls except the one in OffsetTo::serialize_copy() then all tests still passed. It appears that non-object head base is used for Anchor table offsets in GPOS lookup format 4 and 6.

blueshade7:
Even though in many cases `base` argument to `add_link()` is actually the same as `c->current->head`, this fact may not be so obvious from source where each call is made, making it a bit scary change to actually remove `base` from each call.

As a compromise, I plan to add another `add_link()` prototype taking `whence` and `bias` (defaulting to zero), and modify the existing & old `add_link()` prototype which still takes `base`, but converts `base` to `bias` then calls into the first as its implementation.

behdad:
> It appears that non-head base is used for Anchor table offsets in GPOS lookup format 4 and 6.

Interesting.  I'll check.

But yeah I agree with the general sentiment here.

behdad:
> As a compromise, I plan to add another `add_link()` prototype taking `whence` and `bias` (defaulting to zero), and modify the existing & old `add_link()` prototype which still takes `base`, but converts `base` to `bias` then calls into the first as its implementation.

With plan to eventually remove old one, right?  Otherwise it's just adding more complexity.

blueshade7:
Created another PR#2235, where I removed "dst_base" argument to add_link() and its callers where base appears to be actually current->head, leaving non-head base as bias for Anchor table offsets. This ended up in about 100 lines of code changes including `subset_offset_array`. add_link() taking base removed. 
Simpler, kind of scary :-)  but all subset tests passed.

PR #2232, opened 2020-03-06
add "add_link()" with bias arg

blueshade7:
Addresses issue #2227: revisit hb_serialize_context_t::add_link() base arg

Added a prototype of hb_serialize_context_t::add_link() taking `bias` argument (defaulting to 0) which may be specified for all whence_t enum constants `Tail`, `Absolute`, in addition to `Head`. The existing add_link() with base arg calls into this with `bias = base - c->current->head` if base is non-null.


ebraminio:
am learning a bit more about subset internal with the changes, looks very good :)

PR #2233, opened 2020-03-06
Merge max_subtables of sanitizer with its max_ops

ebraminio:
Most likely it is safe to consume from sanitizer's max_ops while reading from Lookup subtables, added on #2219 to fix #1647


ebraminio:
Happened to me accepting subtable count of (binary size of the table) * 64 is a very relax condition! I think if it consumes 8 times more it will be more acceptable.

ebraminio:
nvm

PR #2235, opened 2020-03-07
add "add_link()" with bias arg - take 2

### LINKED ISSUE #2227: revisit hb_serialize_context_t::add_link() base arg

blueshade7:
On top of PR#2232 for fixes #2227 , removed “base” arg to `add_link()` when it is actually `current->head` from cmap, gdef, gpos, gsub, cbdt, sbix subsetters.

Two places where non-head base is used, `AnchorMatrix::serialize()` and `MarkArray::serialize()` in hb-ot-layout-gpos-table.hh, convert base pointer to bias by calling `to_bias()`, then calls add_link() which takes bias replacing base. 

add_link() taking base is removed.

### DIFF
--- a/src/hb-open-type.hh
+++ b/src/hb-open-type.hh
@@ -316,7 +316,6 @@ struct OffsetTo : Offset<OffsetType, has_null>
   bool serialize_subset (hb_subset_context_t *c,
 			 const OffsetTo& src,
 			 const void *src_base,
-			 const void *dst_base,
 			 Ts&&... ds)
   {
     *this = 0;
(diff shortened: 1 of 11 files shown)


ebraminio:
Please rebase the change, it looks also nice and understandable as well :)

blueshade7:
branch rebased

ebraminio:
Thank you very much :)

PR #2240, opened 2020-03-11
[meson] Add needed compiler flags

ebraminio:
Are based on https://github.com/harfbuzz/harfbuzz/blob/2.6.4/configure.ac#L86

@tp-m Can you help on the comments on the patch and review please?

@khaledhosny @behdad just have merged meson port https://github.com/harfbuzz/harfbuzz/pull/1437/files and applied 1c3f80ba136bffec00343bae87269bbc9c33ecde c494d7abcd626c274477319859b9bcb873aca388 on top, let's reach to some acceptable level at least to remove CMake port before the release, please have a look at the changes and give me and @tp-m your feedbacks, thanks!

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -40,6 +40,19 @@ if cpp.get_id() == 'msvc'
   # noseh_link_args = ['/SAFESEH:NO']
 endif
 
+add_global_arguments(cpp.get_supported_arguments([
+  '-fno-rtti',
+  '-fno-exceptions',
+  '-fno-threadsafe-statics',
+  '-fvisibility-inlines-hidden', # maybe shouldn't be applied for mingw
+]), language : 'cpp')
+
+if host_machine.cpu_family() == 'arm' and cpp.alignment('struct { char c; }') != 1
+  if cpp.has_argument('-mstructure-size-boundary=8')
+    add_global_arguments('-mstructure-size-boundary=8', language : 'cpp')
+  endif
+endif
+
 python3 = import('python').find_installation('python3')
 
 check_headers = [


meson.build
```diff
@@ -38,6 +38,16 @@ if cpp.get_id() == 'msvc'
   add_project_arguments(msvc_args, language : 'cpp')
   # Disable SAFESEH with MSVC for libs that use external deps that are built with MinGW
   # noseh_link_args = ['/SAFESEH:NO']
+else
+  add_global_arguments('-fno-rtti', language : 'cpp')
+  add_global_arguments('-fno-exceptions', language : 'cpp')
+  add_global_arguments('-fno-threadsafe-statics', language : 'cpp')
```
tp-m: In `configure.ac` this looks conditional based on having gcc as compiler.

In meson you can do that with `if cpp.get_id() == 'gcc'` (table here: https://mesonbuild.com/Reference-tables.html#compiler-ids) or (perhaps) better don't hard-code a specific compiler but check if the arguments are supported, e.g. like this:
```
  add_global_arguments(cpp.get_supported_arguments(['-fno-rtti', '-fno-exceptions', ...]), language: 'cpp')
```

meson.build
```diff
@@ -38,6 +38,16 @@ if cpp.get_id() == 'msvc'
...
+  if (true) # how to check it isn't mingw?
```
tp-m: Why not use this for mingw?

Ideally you'd just check if the compiler supports it as above without special-casing.

If special-casing is needed then perhaps `if cpp.get_id() = 'gcc' and host_machine.system() == 'windows'`

meson.build
```diff
@@ -38,6 +38,16 @@ if cpp.get_id() == 'msvc'
...
+    add_global_arguments('-fvisibility-inlines-hidden', language : 'cpp')
+  endif
+  if (false) # how to check it is ARM?
+    add_global_arguments('-mstructure-size-boundary=8', language : 'cpp')
```
tp-m: `if host_machine.cpu_family() == 'arm'` (for 32-bit arm, 64-bit is `aarch64` as per https://mesonbuild.com/Reference-tables.html#cpu-families

meson.build
ebraminio: oh, ok, as is needed for clang also, clang impersonate gcc in autotools apparently, how should I check that here?

meson.build
ebraminio: > Why not use this for mingw?

https://github.com/harfbuzz/harfbuzz/commit/560d68af8168d1baff607b9616a3590af70fe9ec and https://github.com/harfbuzz/harfbuzz/commit/e3320ecc1b5a7eaccc7c9370b2d1b76850f054be

as Windows are by default are trying to hide the symbols by default

meson.build
ebraminio: Thank you :)

meson.build
tp-m: I think you should just use `cpp.get_supported_arguments()` or individual `if cpp.has_argument(xyz)` if you want to do it one-by-one. Without special-casing any compilers. If that doesn't work, we can fix it once we know the exact failure.

ebraminio: Oh ok.

meson.build
tp-m: IIRC mingw by default exports all symbols (in general, don't know about inline functions, I would expect those not to be exported), msvc by default exports no symbols.

Personally, I would use the meson port as an opportunity to start over and ignore cruft like that, and only fix it if it still causes issues.

I don't fully understand these commits, probably missing context (didn't look what that script does).

ebraminio: > Personally, I would use the meson port as an opportunity to start over and ignore cruft like that, and only fix it if it still causes issues.

Ok, let's comment it, what do you think about those ARM ones? Hide those as well?

tp-m: Also, we need to check for `!= 'windows'` here if we want to do if it *not* mingw. My bad, sorry.

And you don't need the braces in `if (foo and bar)` for what it's worth.

tp-m: or even `if cpp.get_id() != 'gcc' or .... != 'windows'` (sorry for spam)

ebraminio: > And you don't need the braces in if (foo and bar) for what it's worth.

Oh, nice.

> sorry for spam

I'm better at spamming actually :)

So WDYT "Ok, let's comment it, what do you think about those ARM ones? Comment those as well?"?

tp-m: IMHO:

1) The `-fvisibility-inlines-hidden` I would just add unconditionally, and if it causes problems we'll notice it soon enough and can figure out how to best fix it then.

2) the ARM thing I would keep, like this:
```
if host_machine.cpu_family() == 'arm' and cpp.alignment('struct { char c; }') != 1
  add_global_arguments('-mstructure-size-boundary=8', language : 'cpp')
endif
```
This best matches the autotools logic as I see it. I don't fully understand the rationale behind this though based on commit 3f38c113 and commit 824d00dc - the commit messages don't say *why* this is being done.

tp-m: Actually:
```
if host_machine.cpu_family() == 'arm' and cpp.alignment('struct { char c; }') != 1
  add_global_arguments(cpp.get_supported_arguments(['-mstructure-size-boundary=8']), language : 'cpp')
endif
```
or
```
if host_machine.cpu_family() == 'arm' and cpp.alignment('struct { char c; }') != 1
  if cpp.has_argument('-mstructure-size-boundary=8')
    add_global_arguments('-mstructure-size-boundary=8', language : 'cpp')
  endif
endif
```

And one more thing: we shouldn't use `add_global_arguments()` but `add_project_arguments()` - the difference matters if harfbuzz is used as a meson subproject. We only want these things to apply to harfbuzz.

ebraminio: Oh. At this point the patch has come more of your work than mine, can I close this so you upload it again? :)

tp-m: Nah, don't worry about it. I don't need any credit for that, have enough in the history of the meson port. Thanks for handling this instead of just leaving it to me!

ebraminio: Oh, so have a look :)

tp-m: lgtm

PR #2243, opened 2020-03-11
[meson] Add amalgam build option

ebraminio:


khaledhosny:
Meson has `unity` build option which basically does the same thing, might be a good idea to test it as well.

ebraminio:
> Meson has unity build option which basically does the same thing, might be a good idea to test it as well.

Oh, I didn't know. Done.

Stated C++11 need explicitly as https://circleci.com/gh/harfbuzz/harfbuzz/132417 need.

PR #2246, opened 2020-03-12
[subset] Limit number of lookup indices that can be visited during closures.

### LINKED ISSUE #2224: [subset] lookup closure needs limiting
oss-fuzz reports as https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21025 that hb-subset-fuzzer is timing out in `hb_ot_layout_lookups_substitute_closure()` with lookup counts in GSUB table fuzzed to become very large numbers.

[clusterfuzz-testcase-minimized-hb-subset-fuzzer-5662792105590784.zip](https://github.com/harfbuzz/harfbuzz/files/4283032/clusterfuzz-testcase-minimized-hb-subset-fuzzer-5662792105590784.zip)

garretrieger:
Fixes #2224


ebraminio:
Reduces https://crbug.com/oss-fuzz/21025

from:
./a.out   70.41s user 0.00s system 99% cpu 1:10.47 total

to:
./a.out   2.60s user 0.00s system 99% cpu 2.600 total

and https://crbug.com/oss-fuzz/21091
from:
./a.out   22.62s user 0.01s system 99% cpu 22.657 total

to:
./a.out   2.34s user 0.00s system 99% cpu 2.342 total

So you've basically fixed https://crbug.com/oss-fuzz/21091 also which had the same timeout also AFAIK, can recheck and put its testcase on the PR? Thanks!

ebraminio:
Reuploaded your patch on #2247

Issue #2244, opened 2020-03-11
[meson] Enable CI bots on Windows


ebraminio:
cmake bots on Windows https://github.com/harfbuzz/harfbuzz/blob/master/appveyor.yml needs to be replaced with meson

Issue #490, opened 2017-05-23
meson build?


behdad:
That's what the Cool Kids(TM) are using these days... Ie. GNOME is switching to it.  It looks like cmake done right kind of project...  I'd be happy to replace all our current ones with meson if that works for our Windows developers.

http://mesonbuild.com/

@ebraminio?

ebraminio:
Very compelling, specially the autotools removal part if is planned :) My concern was vcpkg compatibility part which it seems that is also not [an issue](https://github.com/Microsoft/vcpkg/blob/master/ports/libepoxy/portfile.cmake#L19) so this works well for Windows developers also I guess. I suggest some branch for that inside the repo.

behdad:
Yeah that's what I thought.  Let's see how pango turns out...

pedronavf:
I'd vote for CMake, its Windows support strikes me as more mature than what meson could have (being a newer project and considering that Microsoft contributes to CMake). Also, having to install Python 3 on Windows just to build seems to me like a very strong requirement.

behdad:
Maybe @ebassi and I can take a look at this at GUADEC.

behdad:
> I'd vote for CMake

We already have a CMake one.  But no, I'm not going to switch to that as the main build system.

nacho:
FWIW in https://github.com/wingtk/gvsbuild we have support to build with meson without any issues

nacho:
did you guys see https://github.com/centricular/harfbuzz ?

tp-m:
I talked about it at GUADEC with behdad and he said we should submit a PR but we haven't gotten around to it yet, sorry!

behdad:
@tp-m Ready to send a PR?

tp-m:
> @tp-m Ready to send a PR?

Let me rebase and see what happens (wip is at https://github.com/tp-m/harfbuzz/tree/meson-rebased).

PR #2252, opened 2020-03-12
Remove cmake build files

ebraminio:
Not that excited about the removal of cmake as I put much time on it also and it is still in use in our Windows bots, #2244, but since we agreed on replacing all our build tools with meson https://github.com/harfbuzz/harfbuzz/issues/490#issue-230590621 not cmake https://github.com/harfbuzz/harfbuzz/issues/490#issuecomment-315604015 have merged meson PR and tried to reach it to cmake level, usable in dist, and now we can remove that I guess.

But who knows, maybe in far future we will need to drop meson also and use cargo :) Nobody likes such instabilities of course.


khaledhosny:
I think this needs to wait to after at least one release and warn that CMake support is being dropped, otherwise people currently building with CMake will be caught off-guard. I’d even wait a few releases until we are confident that meson build is good enough.

ebraminio:
> I think this needs to wait to after at least one release and warn that CMake support is being dropped, otherwise people currently building with CMake will be caught off-guard. I’d even wait a few releases until we are confident that meson build is good enough.

Sure, makes sense, so should bring it back to release tarball and possibly checking its build again on some CI bots.

ebraminio:
Brought back cmake to release tarball, closing this for now

anthrotype:
yes please, thanks
We rely on CMake to build the uharfbuzz python bindings and need to find the time to figure out how to use meson to build Cython extension modules, see https://github.com/harfbuzz/uharfbuzz/issues/29

past-due:
I appreciate the desire to stabilize on a single build system, but I'm disappointed in the decision to remove CMake, given how everything else that's currently used by various projects I'm involved with uses (or at least supports) CMake.

Just my two cents. (I realize decisions on this were made back in 2017.)

As a workaround, I appreciate whatever contributions you can provide to vcpkg to make sure that the meson build of Harfbuzz is a first-class citizen, and fully supported on all vcpkg-supported platforms without compromise.

behdad:
I don't understand the rush.

ebraminio:
> I don't understand the rush.

The fact I have to sync patches being merged https://github.com/harfbuzz/harfbuzz/commit/3c5bdd9bb8eb97b8ff7109884cbf03c645e4eaa7 https://github.com/harfbuzz/harfbuzz/commit/593e58c84287968f5625c8aa0dd7c60f1a03b705#diff-b139705338b6aaf499d3116a5ffda22c https://github.com/harfbuzz/harfbuzz/commit/d7c012a08e347dcfd1ed4229d70c0a4be496c02b#diff-b139705338b6aaf499d3116a5ffda22c for one build system?

I followed what https://github.com/harfbuzz/harfbuzz/pull/2252#issuecomment-598236380 Khaled said, you seem to have a different idea.

behdad:
> The fact I have to sync patches being merged [3c5bdd9#diff-b139705338b6aaf499d3116a5ffda22c](https://github.com/harfbuzz/harfbuzz/commit/3c5bdd9bb8eb97b8ff7109884cbf03c645e4eaa7#diff-b139705338b6aaf499d3116a5ffda22c) [593e58c#diff-b139705338b6aaf499d3116a5ffda22c](https://github.com/harfbuzz/harfbuzz/commit/593e58c84287968f5625c8aa0dd7c60f1a03b705#diff-b139705338b6aaf499d3116a5ffda22c) [d7c012a#diff-b139705338b6aaf499d3116a5ffda22c](https://github.com/harfbuzz/harfbuzz/commit/d7c012a08e347dcfd1ed4229d70c0a4be496c02b#diff-b139705338b6aaf499d3116a5ffda22c) for one build system?

I thought CMake one reads Makefile.sources but I understand I was wrong.



> I followed what [#2252 (comment)](https://github.com/harfbuzz/harfbuzz/pull/2252#issuecomment-598236380) Khaled said, you seem to have a different idea.

No.  I have the same exact opinion.  Let me quote what here:

> I think this needs to wait to after at least one release and warn that CMake support is being dropped, otherwise people currently building with CMake will be caught off-guard. I’d even wait a few releases until we are confident that meson build is good enough.

I fully exactly support that.  Key parts you are missing are "at least one release", "warn", "even wait a few releases until we are confident that meson build is good enough".  You basically just selectively read "one release" instead.

ebraminio:
> I thought CMake one reads Makefile.sources but I understand I was wrong.

It does, it even can coexist with meson actually. Is that what you want?

> Key parts you are missing are "at least one release", "warn", "even wait a few releases until we are confident that meson build is good enough". You basically just selectively read "one release" instead.

We've introduced meson here https://github.com/harfbuzz/harfbuzz/blob/fb46a32ed0322d3f617dbfc7b982ff26f47d789c/NEWS#L23 and https://github.com/harfbuzz/harfbuzz/blob/fb46a32ed0322d3f617dbfc7b982ff26f47d789c/NEWS#L9 and noticed cmake removal on https://github.com/harfbuzz/harfbuzz/blob/fb46a32ed0322d3f617dbfc7b982ff26f47d789c/NEWS#L7 so is "at least one release" and "warn".

>> "even wait a few releases until we are confident that meson build is good enough"
>  You basically just selectively read "one release" instead.

This makes sense for autotools, which, we are doing, not for cmake that meson is on par with it, running more of the test suite than cmake ever had.

ebraminio:
As said [here](https://github.com/harfbuzz/harfbuzz/commit/19ecabed37914bfff22dd79c633021fd5c1abd93#commitcomment-39674869) and given revealed .pc issue, let's slow things down and see what happens. We will reconsider this, as we were [notifying people since 2.6.5](https://github.com/harfbuzz/harfbuzz/commit/e860000db5c393139c546a1273ba37983c729b41#diff-af3b638bc2a3e6c650974192a53c7291), anytime later so no need to rush.

ebraminio:
As https://github.com/harfbuzz/harfbuzz/issues/2586#issuecomment-663877560 I am more confident now that we can merge this now, let me rebase to see what happens.

PR #2255, opened 2020-03-13
Improve Meson build support for Visual Studio

### LINKED ISSUE #2251: [meson] GDI
Our meson port apparently lacks GDI https://github.com/harfbuzz/harfbuzz/commit/eb8bd2f7eccde483d33406f102c69260fde6fe23#diff-67e997bcfdac55191033d57a16d1408a

Maybe we should rename uniscribe option to gdi, merging the two anyway.

fanc999:
Hi,

This attempts to improve build support with Visual Studio for Meson builds, by:

*  Only applying ```-DHB_DLL_EXPORT``` for the library builds, when built as a shared library
*  Skip tests that involve FreeType, if FreeType support is not enabled
*  Enable finding ICU, Cairo and FreeType manually via looking for headers and .lib's,  GObject/GLib and Graphite2 provide pkg-config files with their build systems for Visual Studio builds, so we could leave these alone.
*  Gave DirectWrite and Uniscribe build support their greenlight, with a minor update
*  Cleaned up DLL builds a bit, since HB_DLL_EXPORT would make the vs_module_defs obsolete as it activates macros to do __declspec(dllexport).  The .def files may still be useful in other situations (MinGW builds?), but that would be outside the scope of this PR.

Few catches:
*  All the tests are either passing or skipped, except for the macos test.  I was getting a Python error, as follows:
'''
  File "c:\gnome.build\hb\..\..\Users\fanc999\harfbuzz\test\shaping\run-tests.py", line 53, in <module>
    for line in f:
  File "c:\python37\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 6498: character maps to <undefined>
'''

*  Introspection builds don't work, I was getting: ```error parsing file src/HarfBuzz-0.0.gir: Line 19, character 56: The attribute 'name' on the element 'type' must be specified```.  I didn't touch any of the introspection stuff, so I am out of ideas for this for now.

With blessings, thank you!

Fixes #2251

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -181,7 +181,7 @@ if fontconfig_dep.found()
   deps += [fontconfig_dep]
 endif
 
-# uniscribe (windows) - FIXME: untested
+# uniscribe (windows)
 if host_machine.system() == 'windows' and not get_option('uniscribe').disabled()
   # TODO: make nicer once we have https://github.com/mesonbuild/meson/issues/3940
   if cpp.has_header('usp10.h') and cpp.has_header('windows.h')
@@ -194,9 +194,9 @@ if host_machine.system() == 'windows' and not get_option('uniscribe').disabled()
   endif
 endif
 
-# DirectWrite (windows) - FIXME: untested
+# DirectWrite (windows)
 if host_machine.system() == 'windows' and not get_option('directwrite').disabled()
-  if cpp.has_header('dwrite.h')
+  if cpp.has_header('dwrite_1.h')
     deps += [cpp.find_library('dwrite', required: true)]
     conf.set('HAVE_DIRECTWRITE', 1)
   elif get_option('directwrite').enabled()
--- a/meson.build
+++ b/meson.build
@@ -90,8 +90,7 @@ glib_dep = dependency('glib-2.0', required: get_option('glib'),
                       fallback: ['glib', 'libglib_dep'])
 gobject_dep = dependency('gobject-2.0', required: get_option('gobject'),
                          fallback: ['glib', 'libgobject_dep'])
-cairo_dep = dependency('cairo', required: get_option('cairo'),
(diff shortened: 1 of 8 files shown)


fanc999:
(for records, this PR should now cover for issue #2251)

ebraminio:
Looks great and thank you @fanc999 for your support all these years! :)

xclaesse:
Note that commit makes a subtle difference that I'm not sure was intentional: https://github.com/harfbuzz/harfbuzz/commit/733414b286b544b54c48e72ba8ae78d5584c157d

`if not freetype_dep.found() and get_option('freetype').enabled()` means that if `freetype` option is set to `auto` (the default) it won't use the fallback any more.

Ironically I found this because it fix an issue I had: when both freetype2 and harfbuzz are built as subprojects, Meson abort because there is a cyclic dependency, see https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/37#note_472587. But I'm not sure that's the right way of fixing that cycle, on Ubuntu they do the opposite: they build freetype2 without harfbuzz instead of harfbuff without freetype2. Which way is preferable?

ebraminio:
> if not freetype_dep.found() and get_option('freetype').enabled() means that if freetype option is set to auto (the default) it won't use the fallback any more.

It is changed a little know, can you latest version? Your feedback would be really valuable as we try to stabilize meson port.

> they build freetype2 without harfbuzz instead of harfbuff without freetype2. Which way is preferable?

This one, as far I know and last time I check freetype use, but of course that isn't our official recommendation.

xclaesse:
@ebraminio I opened: https://github.com/harfbuzz/harfbuzz/pull/2498

PR #2258, opened 2020-03-13
[subset] Fix crash in gpos subsetting.

garretrieger:
Happens if the serialization buffer runs out of space.


ebraminio:
Makes sense. Thanks!

PR #2259, opened 2020-03-14
Meson test api c files build fix

tp-m:
commit 1322ebae33618c78f8e5779b05c17cdaca44c092:

```
[meson] use add_project_arguments() instead of add_global_arguments()
.. and simplify, can pass two languages in one go.

add_global_arguments() won't work if harfbuzz is used as a
meson subproject.
```

commit 7c24cc3f4e0c40d149479475c89c445993975629:

```
[meson] fix spurious warning when building test/api C sources
Fixes compiler warning

  test-unicode.c:589:1: warning: ‘test_unicode_properties_lenient’ defined but not used

which didn't happen with autotools.

Reason it does with meson is that the setup for C was slightly wrong.
We would only add -DHAVE_CONFIG_H to cpp_args which is only valid when
compiling C++ code, but not plain C code, and many of these tests were
plain C.

Instead pass -DHAVE_CONFIG_H via add_project_arguments() and make sure
to set both c_args and cpp_args when building test executables.

Fixes https://github.com/harfbuzz/harfbuzz/issues/2257
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -144,13 +144,14 @@ deps = []
 
 conf = configuration_data()
 incconfig = include_directories('.')
-cpp_args = ['-DHAVE_CONFIG_H']
+
+add_project_arguments('-DHAVE_CONFIG_H', language: ['c', 'cpp'])
 
 warn_cflags = [
   '-Wno-non-virtual-dtor',
 ]
 
-cpp_args += cpp.get_supported_arguments(warn_cflags)
+cpp_args = cpp.get_supported_arguments(warn_cflags)
 
 if m_dep.found()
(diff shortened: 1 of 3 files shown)


ebraminio:
Our meson-gcc bot https://github.com/harfbuzz/harfbuzz/blob/master/.circleci/config.yml#L227-L238 fails https://app.circleci.com/pipelines/github/harfbuzz/harfbuzz/jobs/133013 with the changes apparently, can you have a look? Thanks :)

tp-m:
Can reproduce the CI failure locally with `-Damalgam=true`, will take a look.

PR #2261, opened 2020-03-14
Add messages for GPOS/GSUB phases

simoncozens:
I know the messaging API is a bit ad hoc, and I'm sorry for adding to the adhockery without stopping to plan more broadly - but currently the messaging API tells us that a lookup has been called but you don't know what table the lookup is in!

### DIFF
--- a/src/hb-ot-layout.cc
+++ b/src/hb-ot-layout.cc
@@ -1981,12 +1981,14 @@ inline void hb_ot_map_t::apply (const Proxy &proxy,
 void hb_ot_map_t::substitute (const hb_ot_shape_plan_t *plan, hb_font_t *font, hb_buffer_t *buffer) const
 {
   GSUBProxy proxy (font->face);
+  if (!buffer->message (font, "start GSUB stage")) return;
   apply (proxy, plan, font, buffer);
 }
 
 void hb_ot_map_t::position (const hb_ot_shape_plan_t *plan, hb_font_t *font, hb_buffer_t *buffer) const
 {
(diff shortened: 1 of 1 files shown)


justvanrossum:
This will make the message API much more useful. I'd love to see this merged.

src/hb-ot-layout.cc
```diff
@@ -1981,12 +1981,14 @@ inline void hb_ot_map_t::apply (const Proxy &proxy,
 void hb_ot_map_t::substitute (const hb_ot_shape_plan_t *plan, hb_font_t *font, hb_buffer_t *buffer) const
 {
   GSUBProxy proxy (font->face);
+  if (!buffer->message (font, "start GSUB phase")) return;
```
khaledhosny: HarfBuzz calls this stage instead of phase.

behdad:
Can I request a few changes please?

- Stage is actually a different concept *within* GSUB/GPOS application.  We are not exposing stages in this patch.

- The message should simply be "start table GSUB".  We're not fully consistent, but aiming for "action item-type identifier" where action is start here, item-type is table, and identifier is GSUB.

- There should be matching "end ..." messages.  start/end messages should always be in pairs.

Thanks.

behdad:
If you do a `git grep "buffer->message " and you see a few in morx/kerx...  Not all consistent / complete.  But aiming for...

Maybe the "start lookup %d" should also mention type of lookup as in substitution/positioning.

khaledhosny:
I reverted this commit in master for now.

PR #2264, opened 2020-03-15
Update to Unicode 13.0.0

dscorbett:
I used [the current `master` branch of packTab](https://github.com/harfbuzz/packtab/tree/a7897286d78778f8d4aa4ca98d2e9786ea507f01) except that I had to comment out [lines 135 to 137 of \_\_init\_\_.py](https://github.com/harfbuzz/packtab/blob/a7897286d78778f8d4aa4ca98d2e9786ea507f01/Lib/packTab/__init__.py#L135-L137).


src/gen-arabic-table.py
```diff
@@ -5,7 +5,7 @@
 if len (sys.argv) != 4:
 	print ("""usage: ./gen-arabic-table.py ArabicShaping.txt UnicodeData.txt Blocks.txt
 
-Input files, as of Unicode 12:
```
ebraminio: better to use Unicode 13 here as an indication the script has recognized to be compatible with the indicated Unicode version I believe.

src/hb-unicode-emoji-table.hh
```diff
@@ -7,15 +7,15 @@
...
  * # Unicode and the Unicode Logo are registered trademarks of Unicode, Inc. in the U.S. and other countries.
- * # For terms of use, see https://www.unicode.org/terms_of_use.html
+ * # For terms of use, see http://www.unicode.org/terms_of_use.html
  * #
  * # Emoji Data for UTS #51
- * # Version: 12.0
+ * # Version: 13.0
  * #
- * # For documentation and usage, see https://www.unicode.org/reports/tr51
+ * # For documentation and usage, see http://www.unicode.org/reports/tr51
```
ebraminio: Oh, where these http links, instead https, are coming from? Just nice to have https links back :)

src/gen-arabic-table.py
dscorbett: All these URLs include “latest”: they aren’t specific to Unicode 13.

src/hb-unicode-emoji-table.hh
dscorbett: This is copied from emoji-data.txt.

src/OS2UnicodeRanges.txt
```diff
@@ -0,0 +1,169 @@
...
+116	Kayah Li	A900-A92F
+117	Rejang	A930-A95F
+118	Cham	AA00-AA5F
+119	Ancient Symbols	10190-101CF
+120	Phaistos Disc	101D0-101FF
+121	Carian	102A0-102DF
+	Lycian	10280-1029F
+	Lydian	10920-1093F
+122	Domino Tiles	1F030-1F09F
+	Mahjong Tiles	1F000-1F02F
```
khaledhosny: Do we really need to keep this file, I was under the impression that all bit are saturated and no new Unicode ranges are going to be added.

dscorbett: The bits are saturated. Do they ever add ranges to assigned bits for new “Supplement” or “Extended” blocks?

khaledhosny: Apparently not, the list seems to not have been updated since OpenType 1.5 (released in 2008).

dscorbett: Okay, I’ll revert this.

bobh0303:
Is accepting this PR what is needed to get the ball rolling on shaping for new chars in Unicode 13? If so, what needs to happen to move this forward?

ebraminio:
Waiting for Behdad to comment then guessing will be very good to release afterward.

bobh0303:
Assuming you mean @behdad .

behdad:
> I used [the current `master` branch of packTab](https://github.com/harfbuzz/packtab/tree/a7897286d78778f8d4aa4ca98d2e9786ea507f01) except that I had to comment out [lines 135 to 137 of __init__.py](https://github.com/harfbuzz/packtab/blob/a7897286d78778f8d4aa4ca98d2e9786ea507f01/Lib/packTab/__init__.py#L135-L137).

I still hope to review this work soon.  But that condition is warning of something going wrong and should not be blindly ignored.

dscorbett:
The problem is that packTab checks `any(v < 0 for v in values)` but `v` could be a string like `'0x003Bu'`. Changing it to `any(not isinstance(v, str) and v < 0 for v in values)` works around the problem. The generated hb-ucd-table.hh is unchanged.

behdad:
I'll try to get to this soooon.  Thanks.

LornaSIL:
@behdad Is there any chance this will get accepted soon? That would really be helpful.

behdad:
> The problem is that packTab checks `any(v < 0 for v in values)` but `v` could be a string like `'0x003Bu'`. Changing it to `any(not isinstance(v, str) and v < 0 for v in values)` works around the problem. The generated hb-ucd-table.hh is unchanged.

Thanks.  That was unintended debugging leftover!  Removed it.

In the future would be easier to review if you make multiple commits, one per change... eg:

  * [Unicode 13] Add new hb_script_t values
  * [Unicode 13] Update Arabic table
  ...

Thanks again!!

PR #2274, opened 2020-03-24
[avar] Prevent mul overflow

ebraminio:
Fixes https://crbug.com/oss-fuzz/21350

This `__builtin_mul_overflow` is introduced in GCC 5.1 per https://lwn.net/Articles/651678/ available in clang also but no msvc, having an in house __signed_mul_overflow is a bit more work than unsigned one which we already have in harfbuzz, per https://github.com/torvalds/linux/blob/b2ca74d/include/linux/overflow.h#L163 so not sure given the place we need it right now does worth it to go for research and provide a real fallback anyway.

Thanks!


src/hb-ot-var-avar-table.hh
```diff
@@ -89,10 +89,21 @@ struct SegmentMaps : ArrayOf<AxisValueMap>
...
+# define hb_mul_overflow(x, y, result) (result = (x) * (y), false)
+#endif
+    int factor;
+    if (hb_mul_overflow (arrayZ[i].toCoord - arrayZ[i-1].toCoord, value - arrayZ[i-1].fromCoord, factor))
+      return arrayZ[i-1].toCoord;
+#undef hb_mul_overflow
+
     int denom = arrayZ[i].fromCoord - arrayZ[i-1].fromCoord;
-    return arrayZ[i-1].toCoord +
-	   ((arrayZ[i].toCoord - arrayZ[i-1].toCoord) *
```
ebraminio: Alternatively we can go for a line patch by adding a cast to `(long)` here which feels like moving the overflow which guess isn't that appropriate.

behdad:
The machinery definitely shouldn't go in that file.

ebraminio:
> The machinery definitely shouldn't go in that file.

Have a look again

ebraminio:
Merged but am waiting for any other reviews. Thanks!

behdad:
I'm not happy with this.  The fallback is wrong.

behdad:
You just literally knowingly introduced a very hidden bug into the codebase.

ebraminio:
> You just literally knowingly introduced a very hidden bug into the codebase.

Correct. The fix is somehow worse than the issue itself in long run now I see.

src/hb-algs.hh
```diff
@@ -640,6 +634,29 @@ hb_in_ranges (T u, T lo1, T hi1, T lo2, T hi2, T lo3, T hi3)
...
+/* Right now we only have one use for signed overflow and as it
+ * is GCC 5.1 > and clang we don't care about its fallback ATM */
+#ifndef __has_builtin
+# define __has_builtin(x) 0
+#endif
+#if __has_builtin(__builtin_mul_overflow)
+# define hb_signed_mul_overflows(x, y, result) __builtin_mul_overflow(x, y, &result)
+#else
+# define hb_signed_mul_overflows(x, y, result) (result = (x) * (y), false)
+#endif
```
behdad, line 657: Inline functions please so we benefit from the compiler.

In this case the fallback should simply cast to int64 and do the multiplication and check that it equals the signed multiplication is enough.

behdad:
Or change the avar math to use float, since there's still an addition done after the mul, and there's a division as well so float might actually be faster and avoid overflow problem.

ebraminio:
uploaded as #2322

PR #2275, opened 2020-03-24
[meson] Couple of fixes

### LINKED ISSUE #2262: [meson] No 'main' executable
I could not find an executable for `main.cc` and other `src/test-*` sources.

ebraminio:
Fixes #2262


test/subset/meson.build
```diff
@@ -30,9 +30,10 @@ foreach t : tests
       hb_subset,
       join_paths(meson.current_source_dir(), 'data', 'tests', fname),
     ],
-    workdir: join_paths(meson.current_build_dir(), '..', '..'),
-    # as the test are ran concurrently and we have numerous ones here let's raise acceptable time
+    # as the tests are ran concurrently let's raise acceptable time here
+    # ideally better to break and let meson handles them in parallel
```
tp-m: I don't understand this last line ("ideally...")

ebraminio: Oh, these are running lots of tests inside that's why in parallel running it causes timeout, better to after migration to meson drop this timeout tweaking here and add all the tests inside of it to the main runner I believe.

PR #2279, opened 2020-03-25
[subset] fix fuzzer timeout in layout closure.

### LINKED ISSUE #2224: [subset] lookup closure needs limiting

garretrieger:
Fixes #2224.

This timeout is caused by a single chain rule with a huge lookup expansion. Despite the lookup limit being hit, the expansion is still fully processed. So bail out of chain context lookup expansion as soon as the lookup limit is encountered.

Speeds up this test case from 37s to 3s.


ebraminio:
Makes sense. Thanks!

PR #2284, opened 2020-03-29
GDEF variation store subset

qxliu76:
This PR is dependent on #2206 
I noticed that for subsetting HVAR and VVAR tables, VariationStore subset () will require additional input arguments, that will be added later. This PR is for GDEF VarStore subset only.


src/hb-ot-layout-gdef-table.hh
```diff
@@ -544,28 +577,67 @@ struct GDEF
...
+  void closure_variation_indices (hb_closure_variation_indices_context_t *c) const
+  { (this+ligCaretList).closure_variation_indices (c); }
+
+  void remap_layout_variation_indices (const hb_set_t *layout_variation_indices,
+				       hb_map_t *layout_variation_idx_map /* INOUT */) const
+  {
+    if (version.to_int () < 0x00010003u || !varStore) return;
+    if (layout_variation_indices->is_empty ()) return;
+    
+    hb_hashmap_t<uint16_t, hb_set_t *, 0xffff, nullptr> used_varidxes;
```
garretrieger: There's a possibly simpler approach to this:
1. Create an iterator of hb_pair's with (major, minor).
2. Sort the iterator by major, then minor.
3. Filter out all pairs with major > varStore.dataSets.len
4. Then you can iterate through and do the remapping in one pass. Increment new major every time you encounter a new value for major.

src/hb-ot-layout-common.hh
```diff
@@ -2419,6 +2491,36 @@ struct VariationStore
...
+    {
+      uint16_t major = idx >> 16;
+      uint16_t minor = idx & 0xFFFF;
+
+      if (major >= inner_maps.length) return_trace (false);
+      inner_maps[major].add (minor);
+    }
+    varstore_prime->serialize (c->serializer, this, inner_maps.as_array ());
+
+    for (unsigned i = 0; i < inner_maps.length; i++)
```
garretrieger: There's a possibility of an early return above, which would prevent the items in inner_maps from being cleaned up

behdad:
> I noticed that for subsetting HVAR and VVAR tables, VariationStore subset () will require additional input arguments

Like what?  I like to discuss VariationStore subsetting before any implementation.

behdad:
> > I noticed that for subsetting HVAR and VVAR tables, VariationStore subset () will require additional input arguments
> 
> Like what?

Okay I remember that.  The direct-mapping.  Happy to put that aside for now.

So, the variation subsetting looks like this in general:

* collect all used varidxes
* subset VariationStore retaining only used varidxes; this produces a remapping.
* subset other tables while remapping.

However, since we want our `subset()` calls to only follow a plan and not produce things, I suggest we add a subset-planning object to VariationStore that returns a remapping that when fed back to the VariationStore subset itself, it will subset to that same specification.  Keep that thing your subset-plan.

When we also have designspace-subsetting in the future, this changes a bit, in that the VariationStore, for each retained varidx, might tell us that after designspace subset (ie. partial instantiation), that variation might not be needed anymore.  Ie. all variations were removed from the design space for that entry.  So that culls down the retained varidxes, and the subsetquent subset of GPOS etc need to reflect that.

Also some partial-instancing can also generate "bias" values for varidxes.  Ie. a scalar term that needs to be added to the value that was being varied.  And then GPOS subsetting should add that bias to the value at hand.  That means, the remapping will be from old-varidx to pair{bias,new-varidx} where new-varidx might be `HB_OT_LAYOUT_NO_VARIATIONS_INDEX`.

src/hb-ot-layout.cc
```diff
@@ -318,6 +318,37 @@ hb_ot_layout_get_glyphs_in_class (hb_face_t                  *face,
...
+ * retained in the final subset
+ *
+ * Since: REPLACEME
+ **/
+void
+hb_ot_layout_collect_variation_indices (hb_face_t      *face,
+					const hb_set_t *glyphset,
+					const hb_map_t *gpos_lookups,
+					hb_set_t       *layout_variation_indices /* OUT */,
+					hb_map_t       *layout_variation_idx_map /* OUT */)
```
garretrieger, line 339: Let's avoid making this a new public API method since it's only useful within subset plan creation. I threw together a quick commit which shows how this could be moved into subset plan: https://github.com/googlefonts/harfbuzz/pull/new/GDEF-VariationStore-subset_no_public_api

qxliu76: Done. thanks!

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -2936,6 +2948,8 @@ struct ExtensionFormat1
     return_trace (get_subtable<typename T::SubTable> ().dispatch (c, get_type (), hb_forward<Ts> (ds)...));
   }
 
+  void collect_variation_indices (hb_collect_variation_indices_context_t *c) const {}
```
behdad, line 2951: This doesn't look right.

qxliu76: Fixed. Thanks!

PR #2286, opened 2020-03-30
Add support for big objects on win64

chouquette:
In order to fix this kind of issues when building: https://code.videolan.org/videolan/vlc/-/jobs/372040/

The meson change hasn't been tested, I'm hopping the CI will, to be honest.
The cmake change is tested and fixes the build in the case of VLC


ebraminio:
We are using unity build in cmake by default (harfbuzz.cc that includes all the compilation units) that I think ` /tmp/ccuLPmtO.s: Fatal error: can't close CMakeFiles/harfbuzz.dir/src/harfbuzz.cc.obj: File too big` is related to but not in meson so I think the meson part isn't needed. Is it feasible to use meson in VLC? I will also try to remove unity use in cmake port to check whether use of harfbuzz.cc has caused the issue to see if this is the reason VLC build has failed. Thanks! :)

chouquette:
>  Is it feasible to use meson in VLC?

We're slowly moving there yes :) For dependencies it's already doable, I will give it a try for harfbuzz.

I can remove the meson commit for the time being, and push an updated version which checks if the flag is available in the toolchain being used by cmake

ebraminio:
Can you check what happens if you revert https://github.com/harfbuzz/harfbuzz/commit/9fea6b4dd41bfe2b85f788523162658a7ab9bd49#diff-af3b638bc2a3e6c650974192a53c7291 locally? Maybe some of the other variables extraction are also removed on the consequent edits which can be brought back. I will try that also ASAP.

chouquette:
Using meson to build harfbuzz seems to work (I only tested with the current master for win64)

chouquette:
Reverting the commit you mentioned also fixes the build

ebraminio:
That's great! Is it possible to proceed with meson? That would be just fantastic!

chouquette:
Gladly, but not until the meson.build & friends are part of a release, which if I'm not mistaking, they are not yet
For the time being we'll most likely fix our contrib build by enabling big-obj support, but for the next release I'll gladly switch to meson :)

ePirat:
@ebraminio I was wondering why aren't you just using CMakes unity build feature? Forcing a unity build upon users seems like a bad idea in general, even the CMake manual advises against it:

> Projects should not directly set the UNITY_BUILD property or its associated CMAKE_UNITY_BUILD variable to true. Depending on the capabilities of the build machine and compiler used, it might or might not be appropriate to enable unity builds.

ebraminio:
> @ebraminio I was wondering why aren't you just using CMakes unity build feature? Forcing a unity build upon users seems like a bad idea in general, even the CMake manual advises against it:

Sure, the motivation was to test our unity support somewhere, we now now have that into a meson option and are testing it in our CI, thus https://github.com/harfbuzz/harfbuzz/commit/9fea6b4dd41bfe2b85f788523162658a7ab9bd49#diff-af3b638bc2a3e6c650974192a53c7291 can be just reverted but not much of motivation as it was decided that we migrate to meson instead both autotools and meson.

ebraminio:
@chouquette Hi! Now we have a release with meson, feel free to use that instead cmake. If everything went smoothly there please let us know and close this otherwise I will fix this by reverting that amalgam use in our cmake. Thanks :)

chouquette:
Hi! I'm trying that out now :)

Is it expected that the releases are not offered on http://www.freedesktop.org/software/harfbuzz/release/ ? That's where we use to download the tarballs from

ebraminio:
#2333 please use GitHub's, thanks :)

chouquette:
Fine by me :)

chouquette:
It seems to roughly work! But AFAICS some dependencies are checked/built even when disabled, I'll investigate a bit more and open some MRs if I find some actual issues

ebraminio:
Sound great :)

PR #2287, opened 2020-03-30
Add messages for GPOS/GSUB (take 2)

simoncozens:
See feedback in #2261.

### DIFF
--- a/src/hb-aat-layout-kerx-table.hh
+++ b/src/hb-aat-layout-kerx-table.hh
@@ -891,7 +891,7 @@ struct KerxTable
       reverse = bool (st->u.header.coverage & st->u.header.Backwards) !=
 		HB_DIRECTION_IS_BACKWARD (c->buffer->props.direction);
 
-      if (!c->buffer->message (c->font, "start table %c%c%c%c subtable %d", HB_UNTAG (thiz()->tableTag), c->lookup_index))
+      if (!c->buffer->message (c->font, "start subtable %d", c->lookup_index))
 	goto skip;
 
       if (!seenCrossStream &&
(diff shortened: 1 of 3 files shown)


justvanrossum:
Ping @behdad, any chance to get a review on this?

behdad:
Can we take a step back and design this together please?  What I had in mind was that start/end will signify what you'd think of as XML open/close tags.  So to me:
```
start table GSUB
start lookup 10
end lookup 10
end table GSUB
```
will roughly translate to:
```
  <table tag="GSUB">
    <lookup index="10"/>
  </table>
```
The point is, receiving endpoint needs to keep some state.  So repeating the table tag in each lookup line sounds unnecessary to me.  While that info is currently available in `GSUB`/`GPOS` code, note that the `morx` code is shared by `mort` table, so current PR is wrong in that.  `kerx` and `kern` also share code.  I suggest we reduce duplication and rely on client tracking where we are right now than duplicating.

behdad:
Also, eg. any message starting with `start table morx` should only mean one thing: beginning of morx table processing.  Current PR will emit messages like `start table morx chainsubtable %d`.  That's just wrong.

simoncozens:
OK. So how does this sound:

```
start table XXXX
end table XXXX
start lookup NNNN
end lookup NNNN
```

Only question is what to do with morx/kerx. Should we regard their subtables as "lookups"? In OpenType terms that's essentially what they are.

simoncozens:
Linking #2468. With those patches in place, you get a nicely nested "XML tree":

```
trace: start lookup 19	buffer: ...
trace: start lookup 16	buffer: ...
trace: end lookup 16	buffer: ... 
trace: end lookup 19	buffer: ... 
```

behdad:
> Only question is what to do with morx/kerx. Should we regard their subtables as "lookups"? In OpenType terms that's essentially what they are.

I agree in general.  From my debugging experience is obviously revealing what's going on when i trace and see "chain" instead of "lookup".  But your point is valid.  Same as kern subtables.

simoncozens:
I don't mind calling them `chain`.

simoncozens:
OK, with this patch, messages can be:

start/end table  XXXX
start/end lookup NNNN
start/end subtable NNNN (for kerx/kern)
start/end chainsubtable NNNN (for morx)

PR #2289, opened 2020-03-30
Set sequence number when adding feature to aat map builder

### LINKED ISSUE #2288: [AAT] Sequence field in feature_info_t is never set
In `hb_aat_map_builder_t::feature_info_t`, there's a field `seq` with the helpful comment `/* For stable sorting only. */`.

This is clearly meant to be used by the `static int cmp` method to ensure that the ordering of records with identical `type` will be maintained.

Except that the `seq` field is never set anywhere! Checking at runtime confirms that when we're ready to sort the `features` array in `hb_aat_map_builder_t::compile`, its entries all have `seq == 0`.

jfkthame:
To properly support stable sort in compile() method.

Fixes #2288.


ebraminio:
Makes sense. Thanks

PR #2291, opened 2020-03-30
Handle non-exclusive features when mapping OpenType features to AAT

### LINKED ISSUE #2290: hb_aat_map_builder_t mishandles non-exclusive features
In the course of exploring #2285, I've observed what I think is incorrect behavior of `hb_aat_map_builder_t` when it tries to map OpenType-style feature tags to AAT feature type & selector settings.

The problem arises because of the efforts the map-builder makes to eliminate duplicates. In the case of an "exclusive" AAT feature type, only a single selector can be applied, so multiple feature requests that map to the same type must be eliminated; the code (arbitrarily, I guess) aims to respect the first such request and drop any subsequent ones.

However, for non-exclusive features (like, for example, `HB_AAT_LAYOUT_FEATURE_TYPE_LIGATURES`) this is not appropriate. It's perfectly valid to have multiple LIGATURES / <selector> settings in effect. The only ones that would override each other would be multiple features that map to the same on/off pair of selectors. But 'dlig', 'liga', 'hlig', etc each map to *independent* pairs of on/off selectors for the LIGATURES feature type.

There are two issues, AFAICS: first, in `hb_aat_map_builder_t::compile` when it attempts to "Sort features and merge duplicates", the code looks only at the feature type, and so it will discard additional settings that share a common type, even if they have independent selector values.

And second, even if non-exclusive settings aren't dropped at that stage, the code in `Chain::compile_flags` again only uses the feature type when looking up entries, and therefore it will not correctly find the flags to enable/disable if there are multiple feature/selector pairs of the same type.

jfkthame:
AFAICS something like this should work better. (We really should have some tests, too...)

Fixes #2290.

### DIFF
--- a/src/hb-aat-layout-morx-table.hh
+++ b/src/hb-aat-layout-morx-table.hh
@@ -948,8 +948,10 @@ struct Chain
 	hb_aat_layout_feature_type_t type = (hb_aat_layout_feature_type_t) (unsigned int) feature.featureType;
 	hb_aat_layout_feature_selector_t setting = (hb_aat_layout_feature_selector_t) (unsigned int) feature.featureSetting;
       retry:
-	const hb_aat_map_builder_t::feature_info_t *info = map->features.bsearch (type);
-	if (info && info->setting == setting)
+        // Check whether this type/setting pair was requested in the map, and if so, apply its flags.
+        // (The search here only looks at the type and setting fields of feature_info_t.)
+        hb_aat_map_builder_t::feature_info_t info = { type, setting, false, 0 };
+	if (map->features.bsearch (info))
 	{
 	  flags &= feature.disableFlags;
 	  flags |= feature.enableFlags;
--- a/src/hb-aat-map.cc
+++ b/src/hb-aat-map.cc
@@ -48,12 +48,15 @@ void hb_aat_map_builder_t::add_feature (hb_tag_t tag, unsigned value)
     info->type = HB_AAT_LAYOUT_FEATURE_TYPE_CHARACTER_ALTERNATIVES;
     info->setting = (hb_aat_layout_feature_selector_t) value;
     info->seq = features.length;
+    info->is_exclusive = true;
     return;
   }
 
   const hb_aat_feature_mapping_t *mapping = hb_aat_layout_find_feature_mapping (tag);
   if (!mapping) return;
-  if (!face->table.feat->exposes_feature (mapping->aatFeatureType))
+
+  const AAT::FeatureName* feature = &face->table.feat->get_feature (mapping->aatFeatureType);
+  if (!feature->has_data ())
   {
     /* Special case: Chain::compile_flags will fall back to the deprecated version of
      * small-caps if necessary, so we need to check for that possibility.
@@ -61,7 +64,8 @@ void hb_aat_map_builder_t::add_feature (hb_tag_t tag, unsigned value)
     if (mapping->aatFeatureType == HB_AAT_LAYOUT_FEATURE_TYPE_LOWER_CASE &&
         mapping->selectorToEnable == HB_AAT_LAYOUT_FEATURE_SELECTOR_LOWER_CASE_SMALL_CAPS)
     {
-      if (!face->table.feat->exposes_feature (HB_AAT_LAYOUT_FEATURE_TYPE_LETTER_CASE)) return;
+      feature = &face->table.feat->get_feature (HB_AAT_LAYOUT_FEATURE_TYPE_LETTER_CASE);
+      if (!feature->has_data ()) return;
     }
     else return;
   }
@@ -70,6 +74,7 @@ void hb_aat_map_builder_t::add_feature (hb_tag_t tag, unsigned value)
   info->type = mapping->aatFeatureType;
   info->setting = value ? mapping->selectorToEnable : mapping->selectorToDisable;
   info->seq = features.length;
+  info->is_exclusive = feature->is_exclusive ();
 }
 
 void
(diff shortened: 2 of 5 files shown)


ebraminio:
> We really should have some tests, too...

Does macos.tests make you able to test this?

behdad:
Thanks for this and other work. I'm sorry I didn't understand how to set non-exclusive features before.

Looks generally good to me.

src/hb-aat-layout-morx-table.hh
```diff
@@ -948,8 +948,8 @@ struct Chain
 	hb_aat_layout_feature_type_t type = (hb_aat_layout_feature_type_t) (unsigned int) feature.featureType;
 	hb_aat_layout_feature_selector_t setting = (hb_aat_layout_feature_selector_t) (unsigned int) feature.featureSetting;
       retry:
-	const hb_aat_map_builder_t::feature_info_t *info = map->features.bsearch (type);
-	if (info && info->setting == setting)
+	const hb_aat_map_builder_t::feature_info_t *info = map->features.bsearch (hb_aat_map_builder_t::feature_info_t { type, setting, false, 0 });
```
behdad, line 951: This is first time we'd be using universal initialization syntax `type { initializers }`.  Should be fine.  Normally would have put in a new local variable.

src/hb-aat-map.cc
```diff
@@ -66,10 +67,18 @@ void hb_aat_map_builder_t::add_feature (hb_tag_t tag, unsigned value)
     else return;
   }
 
+  unsigned int default_index;
+  hb_aat_layout_feature_type_get_selector_infos (face,
```
behdad, line 71: No need to call public API.  That API simply calls `face->table.feat->get_selector_infos`.  Do instead.  Also, maybe add `is_exclusive` to feat table.

src/hb-aat-map.cc
```diff
@@ -81,7 +90,7 @@ hb_aat_map_builder_t::compile (hb_aat_map_t  &m)
     features.qsort ();
     unsigned int j = 0;
     for (unsigned int i = 1; i < features.length; i++)
-      if (features[i].type != features[j].type)
+      if (features[i].type != features[j].type || (!features[i].is_exclusive && (features[i].setting ^ features[j].setting) > 1))
```
behdad, line 93: I don't understand the check for xor result `> 1`.  Can you clarify please?

src/hb-aat-map.hh
```diff
@@ -64,19 +64,24 @@ struct hb_aat_map_builder_t
...
 
     HB_INTERNAL static int cmp (const void *pa, const void *pb)
     {
       const feature_info_t *a = (const feature_info_t *) pa;
       const feature_info_t *b = (const feature_info_t *) pb;
-      return (a->type != b->type) ? (a->type < b->type ? -1 : 1) :
-	     (a->seq < b->seq ? -1 : a->seq > b->seq ? 1 : 0);
+      if (a->type != b->type) return (a->type < b->type ? -1 : 1);
+      if (a->is_exclusive) return (a->seq < b->seq ? -1 : a->seq > b->seq ? 1 : 0);
+      if ((a->setting & ~1) != (b->setting & ~1)) return (a->setting < b->setting ? -1 : 1);
```
behdad, line 76: Same question here.  Are all non-exclusive settings `> 1`?

src/hb-aat-map.hh
```diff
@@ -64,19 +64,24 @@ struct hb_aat_map_builder_t
...
     unsigned  seq; /* For stable sorting only. */
 
     HB_INTERNAL static int cmp (const void *pa, const void *pb)
     {
       const feature_info_t *a = (const feature_info_t *) pa;
       const feature_info_t *b = (const feature_info_t *) pb;
-      return (a->type != b->type) ? (a->type < b->type ? -1 : 1) :
-	     (a->seq < b->seq ? -1 : a->seq > b->seq ? 1 : 0);
+      if (a->type != b->type) return (a->type < b->type ? -1 : 1);
+      if (a->is_exclusive) return (a->seq < b->seq ? -1 : a->seq > b->seq ? 1 : 0);
```
behdad, line 75: This line tripped me.  I suggest you remove this line and add `!a->is_exclusive` condition to the next line instead.

src/hb-aat-map.cc
jfkthame: Per https://developer.apple.com/fonts/TrueType-Reference-Manual/RM09/AppendixF.html:

> For nonexclusive features, selectors come in pairs consisting of an even number (0, 2, 4,…) and the odd number following it (1, 3, 5,…). The even number indicates that the selector is on, and the odd number that it is explicitly off.

Or to look at it another way, the low bit of the selector indicates whether to turn this setting of the given type on (low bit = 0) or off (low bit = 1).

So if we have multiple feature requests for the same type, and they differ *only* in the low bit, they're referring to the same setting and we only want to keep the first one; but if they differ in any higher bits (i.e. xor > 1) then they're distinct settings (like Common vs Discretionary, both under the Ligatures type) and we want to keep both.

src/hb-aat-map.hh
jfkthame: Same answer -- if the settings differ only in the low bit, they're the on/off selectors for the same feature, so they're "equal" for the purpose of sorting features here.

src/hb-aat-map.cc
jfkthame: FWIW, I rewrote this in the `((a & ~1) != (b & ~1))` form, as that seems a bit less cryptic.

src/hb-aat-map.cc
jfkthame: Done.

jfkthame:
@behdad Thanks for the comments, updated accordingly. Are you happy for this to be merged now?

behdad:
Thanks @jfkthame 
LGTM.

PR #2294, opened 2020-03-31
[subset] Bail out of context lookup expansion once the lookup limit 

garretrieger:
Is exceeded. Fixes hb-shape-fuzzer timeout (#2224).


ebraminio:
Looks good. Can you look beforehand this time on similar places the issue can happen or some way to apply these checks in a higher level ensuring all the uses are guarded, I should go for that myself also. Thanks!

garretrieger:
Any of the lookup types that can recurse to a large number of sub lookups could also be affected by this. I'll go through all of the lookup types and identify any others that have those properties.

garretrieger:
I don't see any other lookup types that should be problematic, so I believe this should cover all cases now.

behdad:
All such checks could use an `unlikely ()` around them.

ebraminio:
oh, done https://github.com/harfbuzz/harfbuzz/commit/0c65a232fa00a1eb09ca0e9d1d8d26747b27416e

PR #2295, opened 2020-04-01
[subset] Limit lookup indices processed subsetting Feature.

garretrieger:
Also, remove two unnessecary full iterations of the lookup index iterator during serialization of the index array. Fixes fuzzer found timeout.

Reduces run time of the fuzzer input from ~30s to ~1.5s.


ebraminio:
Looks good! Thanks!

PR #2297, opened 2020-04-02
[msvc] Disable C4172 warning

### LINKED ISSUE #2293: hb-algs.hh(352) : warning C4172: returning address of local variable or temporary
When compiling the latest HarfBuzz (2.6.4) MS Visual C++ 15 gives the following 3 warnings:

```
hb-algs.hh(352) : warning C4172: returning address of local variable or temporary
hb-algs.hh(359) : warning C4172: returning address of local variable or temporary
hb-algs.hh(366) : warning C4172: returning address of local variable or temporary
```

The relevant code snippet is below. Lines 352, 359 and 366 are in bold:

```C++
/* Note.  In min/max impl, we can use hb_type_identity<T> for second argument.
 * However, that would silently convert between different-signedness integers.
 * Instead we accept two different types, such that compiler can err if
 * comparing integers of different signedness. */
struct
{
  template <typename T, typename T2> constexpr auto
  **operator () (T&& a, T2&& b) const HB_AUTO_RETURN <!-- THIS IS LINE 352**
  (hb_forward<T> (a) <= hb_forward<T2> (b) ? hb_forward<T> (a) : hb_forward<T2> (b))
}
HB_FUNCOBJ (hb_min);
struct
{
  template <typename T, typename T2> constexpr auto
  **operator () (T&& a, T2&& b) const HB_AUTO_RETURN <!-- THIS IS LINE 359**
  (hb_forward<T> (a) >= hb_forward<T2> (b) ? hb_forward<T> (a) : hb_forward<T2> (b))
}
HB_FUNCOBJ (hb_max);
struct
{
  template <typename T, typename T2, typename T3> constexpr auto
  **operator () (T&& x, T2&& min, T3&& max) const HB_AUTO_RETURN <!-- THIS IS LINE 366**
  (hb_min (hb_max (hb_forward<T> (x), hb_forward<T2> (min)), hb_forward<T3> (max)))
}
HB_FUNCOBJ (hb_clamp);
```

However, I do not understand what this code does.

Is this a known warning? It doesn't look like it should be ignored -- returning an address of a local variable or temporary probably results in undefined behaviour. Not sure...

ebraminio:
Closes #2293, just as a possible fix if we like to

### DIFF
--- a/src/hb-algs.hh
+++ b/src/hb-algs.hh
@@ -346,6 +346,11 @@ HB_FUNCOBJ (hb_second);
  * However, that would silently convert between different-signedness integers.
  * Instead we accept two different types, such that compiler can err if
  * comparing integers of different signedness. */
+#ifdef _MSC_VER
+/* Disable MSVC warning C4172: returning address of local variable or temporary
+ * https://github.com/harfbuzz/harfbuzz/issues/2293 */
+#pragma warning(disable:4172)
+#endif
 struct
 {
   template <typename T, typename T2> constexpr auto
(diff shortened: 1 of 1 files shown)


behdad:
No please!

PR #2299, opened 2020-04-02
[AAT] Don't map OT tags to AAT feature types that are not exposed.

### LINKED ISSUE #2285: Disabling `liga` on Kokonoor AAT font breaks Tibetan shaping
`$ util/hb-view  /Library/Fonts/Kokonor.ttf "རྒྱུ" > kokonor.png`

kokonor

`util/hb-view --features=-liga /Library/Fonts/Kokonor.ttf "རྒྱུ" > kokonor_broken.png`

kokonor_broken

Expected: Disabling "common ligatures" / `liga` should not lead to disabling ligatures required to have readable text. Is there a way to identify required ligatures in AAT and stop those from being disable by the OpenType / AAT feature mapping?

jfkthame:
If an AAT feature type is not exposed in the 'feat' table, we assume it is not intended
to be user-controllable and so we should not map any OT feature tag requests to it.

Fixes #2285.


jfkthame:
@drott I believe this should fix the Tibetan issue; if you could double-check, that'd be great - thanks.

drott:
I can confirm this works on a local Mac OS Chromium build with your commit in it:

kokonor_liga

Thanks for the quick patch!

drott:
So, LGTM!

src/hb-aat-layout-feat-table.hh
```diff
@@ -172,6 +172,9 @@ struct feat
     return featureNameCount;
   }
 
+  bool exposes_feature (hb_aat_layout_feature_type_t feature_type) const
+  { return &get_feature (feature_type) != &Null(AAT::FeatureName); }
```
ebraminio, line 176: Can this be turned to `{ return get_feature (feature_type).feature; }` we prefer not to compare with nullpool on possible cases which in here is possible AFAICS.

jfkthame: We could do, but I'm a bit hesitant -- in principle, zero is a valid feature type constant. We don't currently map any tag to it (and it's unlikely that we ever will), but it still feels wrong to me. This method is supposed to be asking "does the 'feat' table have an entry for feature_type", and it should be valid to ask that for feature_type = 0.

ebraminio: Correct. Anything else from the struct or combination of things which shouldn't be zero you believe? A has_data method can be added to FeatureName to write the logic instead hard coding it here.

ebraminio:
Please add a test to https://github.com/harfbuzz/harfbuzz/blob/master/test/shaping/data/in-house/tests/macos.tests for both fixed and intact cases so we ensure this won't break correct uses. Thanks

src/hb-aat-layout-feat-table.hh
jfkthame: I guess we could assume that `nSettings` must be non-zero in any valid `FeatureName` record. (The spec doesn't explicitly say so, afaics, but having zero settings would be kinda meaningless, and would defeat the basic point of the 'feat' table to determine how to expose settings to the user.)

jfkthame:
> Please add a test to https://github.com/harfbuzz/harfbuzz/blob/master/test/shaping/data/in-house/tests/macos.tests (the used hash is sha1) for both fixed and intact cases so we ensure this won't break correct uses. Adding test only for one of the macos version, perhaps you already have is enough, will add the rest myself if needed. Thanks

I've added a couple basic tests, using fonts on my 10.15.4 machine; if you want to adapt them for other versions, that'd be fine. (I see the Times version, at least, matches what you already have listed for 10.15.)

ebraminio:
Looks great. Thanks!

PR #2306, opened 2020-04-03
Fix handling of multiple AAT feature settings

jfkthame:


ebraminio:
Trusting on your analysis, it make sense. Thanks a lot!

PR #2311, opened 2020-04-05
[subset] Fail ClassDefFormat1 serialization if no space available

ebraminio:
Fixes https://crbug.com/oss-fuzz/21580

Guess we should mark `bool serialize (hb_serialize_context_t *c, unsigned int items_len)` as [`[[nodiscard]]`](https://en.cppreference.com/w/cpp/language/attributes/nodiscard), if it was marked so it was failing with
`src/hb-ot-layout-common.hh:1654:5: error: ignoring return value of function declared with 'nodiscard' attribute [-Werror,-Wunused-result]`


behdad:
> Guess we should mark `bool serialize (hb_serialize_context_t *c, unsigned int items_len)` as [`[[nodiscard]]`](https://en.cppreference.com/w/cpp/language/attributes/nodiscard), if it was marked so it was failing with
> `src/hb-ot-layout-common.hh:1654:5: error: ignoring return value of function declared with 'nodiscard' attribute [-Werror,-Wunused-result]`

No.  The serializer is specifically designed  to keep track of error such that return values of all serialize need NOT be propagated.

It's only when one accessing the serializer buffer manually like the code here was doing that is the code's responsibility to first check and bail if serializer is in error.

behdad:
cc @rsheeter @garretrieger 

We need to document these design decisions better and enforce them as much as we can using compiler assistance.

behdad:
The *real* problem is that the system was designed to use iterators and pipelines to sink the output to serializer and ignore error.  Accessing th serialized results or modifying them should be used in exceptional cases, and in those cases error status of the serializer should be checked manually.

If code is not written in that way then yes a lot of bugs will hide in there.  But that's the real problem here.

PR #2313, opened 2020-04-05
[tests] Replace space with colon in batch mode

### LINKED ISSUE #2303: Support space in font name in hb-shape's --batch mode
[This](https://github.com/harfbuzz/harfbuzz/blob/a0b4ac4dce392326284138fc47cf3741e2798e21/util/hb-shape.cc#L173-L179) makes us unable to use space in font name in batch mode so we have to do [this hack](https://github.com/harfbuzz/harfbuzz/blob/03564fd/test/shaping/run-tests.py#L104-L108), figure out how we can avoid copying in --batch maybe by adding quote around font name in batch more or something, or, let's ignore the hack which isn't used much anyway.

ebraminio:
Fixes #2303

### DIFF
--- a/test/shaping/run-tests.py
+++ b/test/shaping/run-tests.py
@@ -1,10 +1,11 @@
 #!/usr/bin/env python3
 
-import sys, os, subprocess, hashlib, tempfile, shutil
+import sys, os, subprocess, hashlib
 
 def cmd(command):
+	print (command)
 	global process
-	process.stdin.write ((' '.join (command) + '\n').encode ("utf-8"))
+	process.stdin.write ((':'.join (command) + '\n').encode ("utf-8"))
 	process.stdin.flush ()
 	return process.stdout.readline().decode ("utf-8").strip ()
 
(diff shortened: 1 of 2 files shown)


behdad:
LGTM.

behdad:
LGTM if all happy.

behdad:
If we go slightly further maybe we can also handle colon in DOS-style paths `C:\...`.

behdad:
Or not.  Ignore me plz!

ebraminio:
> If we go slightly further maybe we can also handle colon in DOS-style paths C:\....

Apparently that is the issue Windows bots are failing with the change, should debug the things in Windows I guess

ebraminio:
As https://github.com/harfbuzz/harfbuzz/pull/2521 I am a bit undecided whether batch is that useful and as colon apparently isn't the appropriate fix way for this, let's close this till we get a better solution. Thanks!

behdad:
I'm really confused. I like this and is still valid.

behdad:
And you deleted the branch so I can't reopen.
Would have been easier to finish and merge thank close a legit PR.

ebraminio:
> I'm really confused. I like this and is still valid.

Windows's use of `:` in path and msvc build… can have another look to see why full path is passed there, let's reopen so

behdad:
Good enough. Let's also make an issue and fix the fixed-size `buf[4096]`. Same about the `args[32]` but that's less critical.

PR #2314, opened 2020-04-05
Add a not discardable bool type, hb_success_t

### LINKED ISSUE #2312: Use C++17 [[nodiscard]]
In order to avoid issues as #2311 in future, it would be nice if we could add [`[[nodiscard]]`](https://en.cppreference.com/w/cpp/language/attributes/nodiscard) somehow to ArrayOf's `bool serialize (hb_serialize_context_t *c, unsigned int items_len)`, of course we can't use C++17 just right now (or we can?) so some macro would needed to automatically disable the attribute in compilers that don't have the feature.

ebraminio:
Fixes #2312

Guess went too far, let's find a balance, I think all `sanitize ()` methods also should be marked as nodiscard as accidentally discarding their output can be catastrophic.

### DIFF
--- a/src/hb-open-type.hh
+++ b/src/hb-open-type.hh
@@ -604,7 +604,7 @@ struct ArrayOf
   hb_array_t<Type> sub_array (unsigned int start_offset, unsigned int *count = nullptr /* IN/OUT */)
   { return as_array ().sub_array (start_offset, count); }
 
-  bool serialize (hb_serialize_context_t *c, unsigned int items_len)
+  hb_success_t serialize (hb_serialize_context_t *c, unsigned items_len)
   {
     TRACE_SERIALIZE (this);
     if (unlikely (!c->extend_min (*this))) return_trace (false);
@@ -614,7 +614,7 @@ struct ArrayOf
   }
   template <typename Iterator,
 	    hb_requires (hb_is_source_of (Iterator, Type))>
-  bool serialize (hb_serialize_context_t *c, Iterator items)
+  hb_success_t serialize (hb_serialize_context_t *c, Iterator items)
   {
     TRACE_SERIALIZE (this);
     unsigned count = items.len ();
--- a/src/hb.hh
+++ b/src/hb.hh
@@ -318,6 +318,18 @@ extern "C" void  hb_free_impl(void *ptr);
 #  define HB_FALLTHROUGH /* FALLTHROUGH */
 #endif
 
+/* A tag to enforce use of return value for a function */
+#if __cplusplus >= 201703L
+#  define HB_NODISCARD [[nodiscard]]
+#elif defined(__GNUC__) || defined(__clang__)
+#  define HB_NODISCARD __attribute__((warn_unused_result))
+#elif defined(_MSC_VER)
+#  define HB_NODISCARD _Check_return_
+#else
+#  define HB_NODISCARD
+#endif
+#define hb_success_t HB_NODISCARD bool
+
 /* https://github.com/harfbuzz/harfbuzz/issues/1852 */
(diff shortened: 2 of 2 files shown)


ebraminio:
I remeber there was a compiler warning that could be turned to warn on all discarded return values instead polluting each method name HB_NODISCARD, maybe we should go with that instead.

behdad:
Or define a nodiscard book type?

sanitize result should never be discarded.  serialize is more complicated since serialize context keeps the error status as well, so this is actually not correct.

Let's first figure out what exactly the semantics are without putting patch together.

cc @garretrieger @rsheeter

behdad:
Can we get a nodiscard impl in the tree first and start experimenting with using it very slowly?

behdad:
and I feel like we should have a bool type that has nodiscard already.  hb_success_t?

src/hb-open-type.hh
```diff
@@ -604,7 +604,7 @@ struct ArrayOf
   hb_array_t<Type> sub_array (unsigned int start_offset, unsigned int *count = nullptr /* IN/OUT */)
   { return as_array ().sub_array (start_offset, count); }
 
-  bool serialize (hb_serialize_context_t *c, unsigned int items_len)
+  hb_success_t serialize (hb_serialize_context_t *c, unsigned items_len)
```
ebraminio, line 607: As this initially was because of https://github.com/harfbuzz/harfbuzz/commit/57b7de032f60d0499ae2debb293d0f8456acfdfb will gradually other methods also.

behdad:
Please.  Typedef, not macro.

Also, I'm very uneasy about changing a couple of methods in this patch.  Ugh.  I feel like we need to clearly discuss and documents return types of any family of functions before changing anything.  I feel rushed.

behdad:
Like, there's a chance to rethink how things are done and document better and change as necessary.  Instead, we're pushing thoughtless change in and hope for the best.

behdad:
I apologize that I let my frustration out quick.

One thing to think, am I'm not very positive about: can we find a 4-letter name so it doesn't destroy formatting?  Maybe not a big deal and we fix all formatting.

sanitize() sounds like correct thing to change first and see if things fall apart.

ebraminio:
> Please. Typedef, not macro.

Will that work? I doubt it.

> we're pushing thoughtless change in and hope for the best

Correct.

>  I feel like we need to clearly discuss and documents return types of any family of functions before changing anything.

Feels a little contradicting with "Can we get a nodiscard impl in the tree first and start experimenting with using it very slowly?".

behdad:
What I meant is: let's get an implementation in, test that it works, figure out how to discard it, only then roll it out widely.

Things to check:
- Does a typedef work?  I think we do that in cairo using attributes.
- How to discard it.  HB_UNUSED?  What syntax are we happy about?  `(hb_discard) ....`?   I like having something that pairs with the new type, separate from what we use to discard other unused values.

All I'm asking is to let's see what infrastructure works and document it before rolling out.

ebraminio:
> What I meant is: let's get an implementation in, test that it works, figure out how to discard it, only then roll it out widely.

Communication fail so I should say, in fact I feel more guilty of wrong things have done before than this 14 line patch :))

> Does a typedef work? I think we do that in cairo using attributes.

Checked that first as said and apparently no

```
src/hb.hh:331:9: error: an attribute list cannot appear here
typedef [[nodiscard]] bool hb_success_t;
        ^~~~~~~~~~~~~
1 error generated.
```

not in cairo also https://github.com/freedesktop/cairo/blob/6112212/src/cairo-compiler-private.h#L132

> How to discard it. HB_UNUSED? What syntax are we happy about? (hb_discard) ....? I like having something that pairs with the new type, separate from what we use to discard other unused values.

Understandably correct point.

PR #2315, opened 2020-04-06
[subset] fixes dangling object_t issue in subsetting FeatureVariationRecord

### LINKED ISSUE #2310: [subset] null substitutions in FeatureVariationRecord trips assert in assign_offset()
Found this issue while investigating https://crbug.com/oss-fuzz/21560

While serializing`FeatureVariationRecord` with non-null`conditions` and null `substitutions`, `subset()` adds links for conditions then fails because of null `substitutions`. As a recovery attempt, the code below calls `hb_serialize_t::revert()` to restore the serialize buffer, as the result the same address in the serialize buffer may be reused for another link to a condition in the next `FeatureVariation`.
https://github.com/harfbuzz/harfbuzz/blob/c8cc1e378df500122d8082a2038d68179eec63d0/src/hb-ot-layout-common.hh#L245-L251
Multiple links at the same address eventually triggers an assert in `hb_serialize_t::assign_offset()`:
https://github.com/harfbuzz/harfbuzz/blob/c8cc1e378df500122d8082a2038d68179eec63d0/src/hb-serialize.hh#L485-L491
I believe `FeatureVariationRecord::sanitize()` should reject null `substitutions`. Currently it calls the generic `OffsetTo::sanitize()` which accepts null.

qxliu76:
Fixes https://crbug.com/oss-fuzz/21560
revert () does not clean up useless object_t. Adjust the order of
subsetting substitutions and conditions to avoid dangling object_t.

Fixes #2310 
The issue here is that with non-null conditions (subset succeeded) and null substitutions (subset failed), the "object_t"s allocated for conditions offset are not cleaned up by revert (), and then this serializer buffer is allocated for next successful featureVariationRecord, whose value may not be null. Finally this leads to crash in resolve_links() for dangling object_t.


ebraminio:
Merged this but we need a systematic analysis also to prevent such issues beforehand.

behdad:
I'm struggling to understand this change.  I think the underlying problem is somewhere else.

behdad:
This change is fine (ie. avoids some work if all substitutions were subsetted out).  But the underlying issue should be fixed and can hit us elsewhere.

behdad:
> Merged this but we need a systematic analysis also and maybe some internal API redesign to prevent such issues beforehand. Thanks

This could should have worked fine if the underlying issue with `revert` is fixed.  The return value of `subset()` *can* be safely ignored.  That's not the issue.

PR #2322, opened 2020-04-16
Add an appropriate fallback to hb_signed_mul_overflows

ebraminio:
Fixes #2320

using float decreases it conversion accuracy as revealed by `test-var-coords.c` test.

PR #2323, opened 2020-04-16
Use float in avar calculation instead ints and checking their overflows

ebraminio:
<s>The problem is in `test-var-coords.c`, the conversion error with it should be turned from 0.5 > to 0.8 > to allow tests be passed.</s> fixed by the comment below

### DIFF
--- a/src/hb-ot-var-avar-table.hh
+++ b/src/hb-ot-var-avar-table.hh
@@ -89,14 +89,9 @@ struct SegmentMaps : ArrayOf<AxisValueMap>
     if (unlikely (arrayZ[i-1].fromCoord == arrayZ[i].fromCoord))
       return arrayZ[i-1].toCoord;
 
-    int factor;
-    if (unlikely (hb_int_mul_overflows (arrayZ[i].toCoord - arrayZ[i-1].toCoord,
-					value - arrayZ[i-1].fromCoord,
-					factor)))
-      return arrayZ[i-1].toCoord;
-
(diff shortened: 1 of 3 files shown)


src/hb-ot-var-avar-table.hh
```diff
@@ -89,14 +89,9 @@ struct SegmentMaps : ArrayOf<AxisValueMap>
...
-    int factor;
-    if (hb_signed_mul_overflows (arrayZ[i].toCoord - arrayZ[i-1].toCoord,
-				 value - arrayZ[i-1].fromCoord,
-				 factor))
-      return arrayZ[i-1].toCoord;
-
     int denom = arrayZ[i].fromCoord - arrayZ[i-1].fromCoord;
-    return arrayZ[i-1].toCoord + (factor + denom/2) / denom;
+    return roundf (arrayZ[i-1].toCoord + ((float) (arrayZ[i].toCoord - arrayZ[i-1].toCoord) *
+					  (value - arrayZ[i-1].fromCoord) + denom/2) / denom);
```
behdad, line 94: You should remove the `+ denom/2` now that you are calling `roundf`.

ebraminio: Thanks! Apparently applying that fixes its accuracy issue.

ebraminio: So now to decide which way to go, this or #2322

behdad:
I'm happy with the float version.  As I said, since there's a division involved, float is probably faster even.

ebraminio:
Thanks!

PR #2329, opened 2020-04-17
Move hb_font_get_var_coords_design behind HB_EXPERIMENTAL_API

ebraminio:
let's decide about this in some other time


behdad:
Thanks.

behdad:
Need change in .h file as well?

ebraminio:
> Need change in .h file as well?

Was running local tests in background… done

PR #2335, opened 2020-04-18
implement SID to glyph ID mapping with predefined Charset

blueshade7:
Added a test case to API test `test-ot-gnames' to exercise `ExpertSubsetCharset`.

This PR also fixes oss-fuzz 21769. Added the fuzzed file to fuzzing/fonts folder.


src/hb-ot-cff1-table.hh
```diff
@@ -1217,6 +1217,31 @@ struct cff1
       }
     }
 
+    hb_codepoint_t sid_to_glyph (hb_codepoint_t sid) const
+    {
+      if (charset != &Null(Charset))
```
ebraminio, line 1222: nullptr feels even better than nullpool for such cases :/

blueshade7: Maybe so. This set up is not new though. There are about two dozens places like this.

ebraminio:
Thanks for the quick response, much appreciated, and it looks good. Even the fact I see why you've put the fuzz fix and the implementaion of custom charsets at the same change, any chance you think we can have two PR for them? Guess lack of charset to nullpool check address has made the issue at the first place? Such address comparison feels like a deeper issue which should be addressed after the change.

blueshade7:
>any chance you think we can have two PR for them?

This PR implements a missing implementation of mapping from glyph name to glyph ID when a CFF font uses a predefined Charset. Without it, even a good font may cause a crash. The fuzzed font happened to discover this oversight. Thus, implementing predefined Charset directly fixes a crash by the fuzzed font. They are inseparable.
`charset` is explicitly set to `&Null` when a font uses a predefined Charset (pls see `init()`). The code you quoted above is not testing if Charset is bad, rather testing if it is predefined, so could be rewritten to call `is_predef_charset()` instead if you prefer, but I don't think this is essential.

src/hb-ot-cff1-table.hh
ebraminio: Understandable, feel free to remove some of them if you got some time to not have such surprises.

ebraminio:
> charset is explicitly set to &Null when a font uses a predefined Charset (pls see init()). The code you quoted above is not testing if Charset is bad, rather testing if it is predefined [..]

I understand, the convention feels weird however, we should put some effort to improve them some time.

> [..] so could be rewritten to call is_predef_charset() instead if you prefer, but I don't think this is essential.

I believe making it explicit will be better than what we have already. Feel free to have a look if you got sometime.

Thanks so much for reaching to it :)

src/hb-ot-cff1-table.cc
```diff
@@ -158,6 +245,18 @@ hb_codepoint_t OT::cff1::lookup_expert_subset_charset_for_sid (hb_codepoint_t gl
     return 0;
 }
 
+hb_codepoint_t OT::cff1::lookup_expert_charset_for_glyph (hb_codepoint_t sid)
+{
+  const auto  *pair = hb_bsearch (sid, expert_charset_sid_to_gid, ARRAY_LENGTH (expert_charset_sid_to_gid));
```
behdad, line 250: You can also try `hb_array (expert_charset_sid_to_gid).bsearch (sid)`.  Same below.

blueshade7: ah good to know

ebraminio: #2349

PR #2340, opened 2020-04-18
autotools: Add fix_get_types.py to EXTRA_DIST

### LINKED ISSUE #2337: meson: fix_get_types.py missing from tarball
```
../harfbuzz-2.6.5/src/meson.build:529:2: ERROR: File fix_get_types.py does not exist.
```

lazka:
So it ends up in the release tarball and can be used in the meson build.

Fixes #2337


ebraminio:
Thank you very much :)

PR #2344, opened 2020-04-18
[meson] Make harfbuzz-icu separate module optout-able

### LINKED ISSUE #2343: Add --with-icu=builtin like option to meson
As #2341, hb-icu is separated in meson also which now matches with defaults of our autotools and cmake porst but autotools had an option to not have hb-icu as a separate module also https://github.com/harfbuzz/harfbuzz/blob/f9bc373/configure.ac#L278-L280 and https://github.com/harfbuzz/harfbuzz/blob/dced2fb/src/Makefile.am#L163-L180 guess would be nice to add such option in meson also.

ebraminio:
Fixes #2343

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -190,6 +190,9 @@ endif
 
 if icu_dep.found()
   conf.set('HAVE_ICU', 1)
+endif
+
+if get_option('icu-builtin')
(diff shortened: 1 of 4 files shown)


lazka:
lgtm

ebraminio:
Thanks :)

PR #2346, opened 2020-04-19
meson: install gobject/icu headers if they are enabled

lazka:
The first commit fixes that hb-icu.h was missing, the second skips the header install in case gobject/icu are disabled.


ebraminio:
Looks great! Thanks :)

PR #2349, opened 2020-04-19
Minor, use hb_sorted_array::bsearch where possible

ebraminio:
Resolves https://github.com/harfbuzz/harfbuzz/pull/2335#discussion_r410767510

### DIFF
--- a/src/hb-ot-os2-unicode-ranges.hh
+++ b/src/hb-ot-os2-unicode-ranges.hh
@@ -222,10 +222,9 @@ static const OS2Range _hb_os2_unicode_ranges[] =
 static unsigned int
 _hb_ot_os2_get_unicode_range_bit (hb_codepoint_t cp)
 {
-  auto* range = hb_bsearch (cp, _hb_os2_unicode_ranges, ARRAY_LENGTH (_hb_os2_unicode_ranges));
-  if (range != nullptr)
-    return range->bit;
-  return -1;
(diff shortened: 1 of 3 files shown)


src/hb-ot-os2-unicode-ranges.hh
```diff
@@ -222,10 +222,9 @@ static const OS2Range _hb_os2_unicode_ranges[] =
 static unsigned int
 _hb_ot_os2_get_unicode_range_bit (hb_codepoint_t cp)
 {
-  auto* range = hb_bsearch (cp, _hb_os2_unicode_ranges, ARRAY_LENGTH (_hb_os2_unicode_ranges));
-  if (range != nullptr)
-    return range->bit;
-  return -1;
+  auto *range = hb_sorted_array (_hb_os2_unicode_ranges).bsearch (cp);
+  if (unlikely (range == nullptr)) return -1;
+  return range->bit;
```
behdad, line 227: FWIW I find explicit comparison to `nullptr` really useless and a distraction.  The object (pointer in this case) readily provides a boolean conversion.  Comparing it to a sentinel value to do same is bizarre.

Also this is perfect situ where `return range ? range->bit : -1` is a lot nicer.

ebraminio: Here you go https://github.com/harfbuzz/harfbuzz/pull/2353

Issue #2334, opened 2020-04-17
[subset] FeatureVariations subsetting is wrong


behdad:
We cannot drop FeatureVariationsRecords that have empty substitutions.  That changes the behavior.  A FeatureVariationsRecords stops the search when the conditions match; so even if there's no susbstitutions we cannot drop the record as that will keep the search going.

The only optimizations we can do are 1. dropping from the end of the array any records with no susbtitutions, and 2. dropping records earlier with no susbstitution only after analysing the conditions and proving that the dropping does not change functionality.

Both of those are unnecessary right now.  We should fix the functionality.

This seems to be wrong in FontTools as well: https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/subset/__init__.py#L1325

cc @anthrotype @qxliu76

behdad:
The spec is here: https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2

qxliu76:
@anthrotype please let me know when this is fixed in fonttools, I'll then fix it in harfbuzz

ebraminio:
@qxliu76 and colleagues, hadn't the chance to state this somewhere else, thanks for all  the tireless efforts you've put on implementing this and other in progress layout subset works, hopefully we will soon fix all the mentioned issues providing layout subset as a quality works to the users. Thanks again :)

behdad:
Here's fonttools fix.  We don't need to implement the pruning here initially. https://github.com/fonttools/fonttools/pull/1882

qxliu76:
Now fonttools is pruning empty record at the end, while harfbuzz is not doing the same thing. Is it an issue in testing?

qxliu76:
@ebraminio, thank you!

behdad:
> Now fonttools is pruning empty record at the end, while harfbuzz is not doing the same thing. Is it an issue in testing?

We'll deal with that when we get to.

PR #2352, opened 2020-04-20
[subset] FeatureVariations subsetting is wrong

behdad:
Never drop FeatureVariationRecord for now.

Fixes https://github.com/harfbuzz/harfbuzz/issues/2334


ebraminio:
Will follow https://github.com/fonttools/fonttools/pull/1882 to add their test also

PR #2353, opened 2020-04-20
minor, replace nullptr checks with implicit ones

ebraminio:
feel free to disagree with ones you don't like ;)


behdad:
LGTM.  Thanks!

PR #2355, opened 2020-04-20
[WIP] [serialize] Deprecated Offset::serialize()

behdad:
That's bad old design.  We should convert all uses to "new" 'tetris'
serializer.  Takes some more design work to make conversion easy.
Putting this in to flag all problem callsites.

Just creating this WIP PR now to experiment with proposed designs.


behdad:
This was mostly fixed by @garretrieger in 0da9158f620077ec46dcdfa4c1556253b23337eb

I'm trying to remove `Offset::serialize()` as well, which has one use currently.

Issue #2242, opened 2020-03-11
[meson] build ragel artifacts


ebraminio:
our meson port doesn't have building HB_BASE_RAGEL_GENERATED_sources from HB_BASE_RAGEL_sources equivalent but it should.

tp-m:
How are these supposed to work?

It looks like the generated sources are actually checked into git, so would it be ok to hide them behind a custom target so they can be updated manually as/when needed? Or we only generate them if `ragel` is found and otherwise use the checked-into-git versions?

If we make `ragel` a hard dependency that would be unfortunate for building on Windows and pretty much all non-Linux OS.

khaledhosny:
Currently `configure.ac` checks for ragel with `AM_MISSING_PROG`, so only when `ragel` executable is needed to regenerate the files a warning will be issued. This makes it harder to miss updating the generated files while still not requiring ragel when building from tarball.

tp-m:
Right, just haven't entirely figured out yet under what circumstances it would actually be considered 'neeed' and via what mechanism in `Makefile.am`. Would it be done as part of disting or when?

khaledhosny:
Currently it would be the regular timestamp checking of make.

behdad:
For a developer, we assume they have ragel, and we want everything to automatically regenerate and work as they edit the ragel sources.

When built from tarball, we don't want to require ragel.

Note that currently we have a bug such that the timestamps from the tarball may confuse `make` into thinking that the ragel sources are newer than the bundled generated file.  We couldn't find any easy solution to that.

tp-m:
Took a stab at this here: https://github.com/tp-m/harfbuzz/commits/meson-ragel-gen

But it breaks `make check` on autotools because `check-includes.sh` doesn't like the fact that I changed the includes for the generated headers to angle brackets (which is required to make them get picked up from the build dir in Meson and probably cmake too, and probably autotools too if one uses builddir != srcdir there, though I guess no one does that for development).

What is the rationale/background for this check?

The installed `hb-*.hh` header files are hidden in a `harfbuzz` subdir so can't really be accidentally picked up from the system include dirs.

khaledhosny:
> and probably autotools too if one uses builddir != srcdir there, though I guess no one does that for development

I always do out-of-tree builds, no issue with includes from the generated files. `make distcheck` does an out-of-tree build as well, so it would have caught such an issue.

tp-m:
[...]

PR #2357, opened 2020-04-21
[meson] Generate harfbuzz.cc, hb-version.h and ragel artifacts

### LINKED ISSUE #2245: [meson] Generate harfbuzz.cc
Meson should be able to generate harfbuzz.cc as autotools [here](https://github.com/harfbuzz/harfbuzz/blob/1dd389668847bcbc5e94573c80bb71bb87ee0d9c/src/Makefile.am#L311-L325)

ebraminio:
Fixes #2245 and #2242 however has one big issue, it generates ragel artifacts each time not only when the dependent file is updated, same as harfbuzz.cc, ideally harfbuzz.cc should be generated whenever `meson.build` itself is changed and ragel artifacts only when `.rl` files are changed which not considering that causes rest of the project in need for rebuild as timestamp updates.

This is considered the official way to define command targets https://mesonbuild.com/Run-targets.html what we need is basically always running them when some of the files are updated, like what make offers.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -98,12 +98,7 @@ BUILT_SOURCES += \
 	hb-version.h
 
 $(srcdir)/hb-version.h: hb-version.h.in $(top_srcdir)/configure.ac
-	$(AM_V_GEN) $(SED) \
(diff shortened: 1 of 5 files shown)


ebraminio:
Have asked that in https://github.com/mesonbuild/meson/issues/7156 also

PR #2359, opened 2020-04-21
[CBLC] Fix assert failure

ckitagawa-work:
If there are multiple IndexSubtableRecord and at least one succeeds in serialization before another fails we currently fail to cleanup extra push() calls before returning. This leads to an assert failure downstream when trying to add a link. The fix is to simply discard these push() calls before returning.


ckitagawa-work:
@garretrieger PTAL Thanks!

behdad:
There might be a better way to do it.  Let me check.  THis is not the kind of cleanup code you should need to write.

garretrieger:
I think you can do this using snapshot()/revert() on the serializer. Start a snapshot before you begin adding the records and then call revert() if you need to bail out.

behdad:
Creating objects for records without popping them is plain wrong.

behdad:
You cannot push each record, then pop each record.  I mean, you can.  But there's no reason you should.

behdad:
Right.  If specific order is need you shouldn't be using the object packer.

You can use Offset::serialize() currently for that.  But I'm going to slightly change that.

garretrieger:
In the meantime should we merge this to fix the fuzzer error?

ebraminio:
I believe we should

behdad:
I'm fine with that.  We just need to clean this as soon as we develop better solution.  THat's is happening at https://github.com/harfbuzz/harfbuzz/issues/2164#issuecomment-617436277

Issue #2152, opened 2020-02-04
Consider providing std::unique_ptr based scope helpers?


drott:
In [this Chromium CL](https://chromium-review.googlesource.com/c/chromium/src/+/2034111/3/third_party/harfbuzz-ng/utils/hb_scoped.h) we're introducing Chromium project wide std::unique_ptr based scope helpers for easier memory management when using HarfBuzz API from C++ in Blink or the fuzzers. The location for this code is not optimal and a wider number of clients would benefit if we could add those to upstream.

WDYT?

Example code below:

```
template <typename T>
struct always_false : std::false_type {};

template <class T>
struct HbSpecializedDeleter {
  inline void operator()(T* obj) {
    static_assert(always_false<T>::value,
                  "HbScoped is only allowed for HarfBuzz types that have a "
                  "deleter specialization.");
  }
};

// Defines a scoped pointer type HbScoped based on std::unique_ptr and using the
// corresponsing HarfBuzz destructor to commonly used public HarfBuzz types. The
// interface of HbScoped is the same as that of std::unique_ptr.
//
//  void MyFunction() {
//    HbScoped<hb_blob_t> scoped_harfbuzz_blob(
//        hb_blob_create(mydata, mylength));
//
//    DoSomethingWithBlob(scoped_harfbuzz_blob.get());
//  }
//
// When |scoped_harfbuzz_buffer| goes out of scope, hb_blob_destroy() is called
// for the hb_blob_t* created from hb_blob_create().
template <class T>
using HbScoped = std::unique_ptr<T, HbSpecializedDeleter<T>>;

#define SPECIALIZED_DELETER_FOR_HARFBUZZ_TYPE(TYPE, DESTRUCTOR) \
  template <>                                                   \
  struct HbSpecializedDeleter<TYPE> {                           \
    inline void operator()(TYPE* obj) { DESTRUCTOR(obj); }      \
  };

#define HB_TYPE_DESTRUCTOR_PAIRS_REPEAT(F) \
  F(hb_blob_t, hb_blob_destroy)            \
  F(hb_buffer_t, hb_buffer_destroy)        \
  F(hb_face_t, hb_face_destroy)            \
  F(hb_font_t, hb_font_destroy)            \
  F(hb_set_t, hb_set_destroy)              \
  F(hb_subset_input_t, hb_subset_input_destroy)

[...]

PR #2362, opened 2020-04-22
Require c++11

behdad:


behdad:
Humm.  Visual Studio doesn't define `__cplusplus`?

behdad:
> Humm. Visual Studio doesn't define `__cplusplus`?

Ugh..
"""
The __cplusplus preprocessor macro is commonly used to report support for a particular version of the C++ standard. Because lots of existing code appears to depend on the value of this macro matching "199711L", the compiler does not change the value of the macro unless you explicitly opt-in by using the /Zc:__cplusplus compiler option. The /Zc:__cplusplus option is available starting in Visual Studio 2017 version 15.7, and is off by default. In earlier versions of Visual Studio, and by default, or if /Zc:__cplusplus- is specified, Visual Studio returns the value "199711L" for the __cplusplus preprocessor macro. The /permissive- option does not enable /Zc:__cplusplus.
"""
https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=vs-2019

Any advice anyone?  I don't want to rewrite it blindly...

behdad:
The reason I ended up touching these bits is that I'm working on adding `hb-cplusplus.h` as C++ integration as discussed in https://github.com/harfbuzz/harfbuzz/issues/2152

For that I wanted to only enable the code if C++ >= 11 to avoid getting bugreports of it not working with older C++.  But looks like one cannot reliably use __cplusplus even with MSVC.  That's soooo typical of....

cc @drott

behdad:
The autotools bit is safe to go in or did we also hit snags there before?  I have a feeling that we had tried it for sure.

behdad:
Am learning from this and doing other.

PR #2364, opened 2020-04-22
[kern] Simplify hb_kern_machine_t

RazrFalcon:
This commit removes some rudimentary code like `skippy_iter.init (&c);` which isn't needed since `set_lookup_mask` and `set_lookup_props` already called it.

Also, it uses an iterator directly, without a temporary reference.

Overall, this is just a simple refactoring.


behdad:
The reason I like keeping the reference is that that's where we choose which iter from `c` to use, and we won't repeat that choice later.  c has `input_iter` as well as another one.

behdad:
I agree the `init` needs to go.

RazrFalcon:
Done.

behdad:
Lovely.  Thank you.

PR #2368, opened 2020-04-23
Dont build tests when disabled

chouquette:
When building for linux with tests disabled, this drops the number of build files from 112 to 62, and the number of targets from 24 to 4.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -396,20 +396,36 @@ libharfbuzz_subset_dep = declare_dependency(
   dependencies: deps)
 
 if get_option('tests').enabled()
-# TODO: MSVC gives the following,
-# error LNK2019: unresolved external symbol "unsigned __int64 const * const _hb_NullPool"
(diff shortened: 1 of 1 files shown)


src/meson.build
```diff
@@ -395,97 +395,99 @@ libharfbuzz_subset_dep = declare_dependency(
...
-
-  foreach name : dist_check_script
-    test(name, find_program(name), env: env)
-  endforeach
-
-  test('check-symbols.sh', find_program('check-symbols.sh'),
-    depends: [harfbuzz_def, harfbuzz_subset_def], # TODO: harfbuzz_icu_def and harfbuzz_gobject_def also
-    env: env)
+if get_option('tests').enabled()
+ # TODO: MSVC gives the following,
```
ebraminio, line 399: two space indention please :)

chouquette: Oops, done!

ebraminio:
Looks good, thanks :)

PR #2374, opened 2020-04-24
[metrics] Normalize rise and run values

### LINKED ISSUE #2366: Should we scale caret related values at all?
Reading https://docs.microsoft.com/en-us/typography/opentype/spec/vhea values, we probably shouldn't have scaled [hv]hea->caretSlopeRise and [hv]hea->caretSlopeRun values at https://github.com/harfbuzz/harfbuzz/blob/6461143b44f81a4190d3f1cb02238750536f08e4/src/hb-ot-metrics.cc#L149-L154 and to be honest don't know the difference between them.

ebraminio:
[metrics] Normalize rise and run values

Many fonts use values like 1 or 0 for rise and run which is allowed
per spec when font is not slanted but they get rounded to zero when
we do font object based scaling, this, while considering variations
scales font values in scales of face's upem.

fixes #2366


behdad:
I actually was thinking that we calculate multiplier from rise/run **without variations**.  That would drastically simplify the code, you won't need a function I think.

ebraminio:
> [..] without variations.

Hmm, so you mean applying variation after calculating multiplier? That makes it to have not aligned rise/run with upem when MVAR tweaks it.

> [,,] That would drastically simplify the code [..]

I can't use the same path AFAICS anyway.

> [..] you won't need a function I think.

Not adding a function reduces noise, done.

ebraminio:
Goes nowhere IIUC, let's close

PR #2375, opened 2020-04-24
meson: Use find_program instead of the python module

chouquette:
Using the python module introduces a dependency on python3-setuptools,
which may be uninstalled on some machines, especially buildbots.
find_programs will run less checks but all since all that's needed is a
python3 installation to run some scripts, that seems good enough

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -54,7 +54,7 @@ if host_machine.cpu_family() == 'arm' and cpp.alignment('struct { char c; }') !=
   endif
 endif
 
-python3 = import('python').find_installation('python3')
+python3 = find_program('python3')
 
 check_headers = [
   ['unistd.h'],


chouquette:
https://code.videolan.org/chouquette/vlc/-/jobs/386119 is an exemple of what this PR tries to fix

ebraminio:
Make sense but hmm, now our Windows CI bot says 'Program python3 found: NO', makes sense as pyhon3 named python in Windows anyway but I should see whether it was really working before and we some find_program('*.py') and not sure how they are working right now :/

chouquette:
I have no idea about how python works on windows, is it guaranteed to be called python3.exe and not python?

That's the kind of extra checks that the python module does, which are quite handy... I'll see if I can find some other way first

ebraminio:
> I have no idea about how python works on windows, is it guaranteed to be called python3.exe and not python?

Me neither :) Apparently it is called python.exe no matter the version so find_program('python') should work except linux machines that doing may call python2. We have shebang /usr/bin/env python3 above in our scripts, hopefully we can have something similar in meson for *nix. Still not sure how our `find_program('*.py')` can work, I mean, should we use the found python3 variable here? https://github.com/harfbuzz/harfbuzz/blob/master/test/fuzzing/meson.build#L24

Also please review https://github.com/harfbuzz/harfbuzz/pull/2357 if you like to :)

chouquette:
It seems find_program('/path/to/script.py') is the recommended way, I'll try to open a PR to do that then :)

ebraminio:
> It seems find_program('/path/to/script.py') is the recommended way, I'll try to open a PR to do that then :)

Sounds great!

PR #2378, opened 2020-04-24
[metrics] Add a private tag for Italic correction, Ital

### LINKED ISSUE #1880: Expose post table’s italicAngle?

ebraminio:
Fixes #1880 (or, it can be considered for)

Compare https://github.com/harfbuzz/harfbuzz/pull/2378/files#diff-3e0886e0824e462dddb236ac6d8b2fbbR78 to https://github.com/harfbuzz/harfbuzz/pull/2374/files#diff-3e0886e0824e462dddb236ac6d8b2fbbR67 (-426 vs 430)


test/api/test-ot-metrics.c
```diff
@@ -68,11 +68,27 @@ test_ot_metrics_get_var (void)
...
+test_ot_metrics_get_ital (void)
+{
+  hb_face_t *face = hb_test_open_font_file ("fonts/DroidSerif-Italic.ttf");
+  hb_font_t *font = hb_font_create (face);
+  hb_position_t value;
+  g_assert (hb_ot_metrics_get_position (font, HB_TAG ('I','t','a','l'), &value));
+  g_assert_cmpint (value, ==, -426);
+  int x_scale;
+  hb_font_get_scale (font, &x_scale, NULL);
+  g_assert_cmpint ((int) roundf (asinf ((float) value / x_scale) * 180.f / (float) M_PI), ==, -12);
```
ebraminio, line 81: How to get degree back

src/hb-ot-metrics.cc
```diff
@@ -107,6 +107,7 @@ _get_gasp (hb_face_t *face, float *result, hb_ot_metrics_tag_t metrics_tag)
 #define _HB_OT_METRICS_TAG_HORIZONTAL_DESCENDER_HHEA HB_TAG ('H','d','s','c')
 #define _HB_OT_METRICS_TAG_HORIZONTAL_LINE_GAP_OS2   HB_TAG ('O','l','g','p')
 #define _HB_OT_METRICS_TAG_HORIZONTAL_LINE_GAP_HHEA  HB_TAG ('H','l','g','p')
+#define _HB_OT_METRICS_TAG_ITALIC_CORRECTION         HB_TAG ('I','t','a','l')
```
ebraminio, line 110: Feel free to suggest something better

test/api/test-ot-metrics.c
```diff
@@ -68,11 +68,27 @@ test_ot_metrics_get_var (void)
...
+test_ot_metrics_get_ital (void)
+{
+  hb_face_t *face = hb_test_open_font_file ("fonts/DroidSerif-Italic.ttf");
+  hb_font_t *font = hb_font_create (face);
+  hb_position_t value;
+  g_assert (hb_ot_metrics_get_position (font, HB_TAG ('I','t','a','l'), &value));
+  g_assert_cmpint (value, ==, 426);
+  int x_scale;
+  hb_font_get_scale (font, &x_scale, NULL);
+  g_assert_cmpint ((int) roundf (-asinf ((float) value / x_scale) * 180.f / (float) M_PI), ==, -12);
```
ebraminio, line 81: How to get post->italicDegree back, can be done by HB_OT_METRICS_TAG_HORIZONTAL_CARET_RUN / HB_OT_METRICS_TAG_HORIZONTAL_CARET_RISE right now also of course

jfkthame:
I don't understand this. "Italic correction" is a specific TeX-related concept that is not the same as "italic angle", and cannot be directly computed from it; it's a per-glyph metric that is up to the font designer to specify.

In XeTeX (for example), we do synthesize a "fake" value for italic correction -- using italic angle as one of the inputs -- but that's just a fallback heuristic to make up for the lack of this glyph metric in OpenType fonts. I don't think harfbuzz should be attempting to do anything like that; it's purely an application-level function.

ebraminio:
> [..] I don't think harfbuzz should be attempting to do anything like that; it's purely an application-level function.

#1890 is what I suggest but nobody shows any interest and Behdad says no client will use that (but I believe a font manager easily can take advantage of it, don't you know any other client willing to use it) so the only option will be providing one `hb_face_get_italic_angle` like API.

ebraminio:
Another doomed attempt :/ let's agree on having hb-style or a `hb_font_get_italic_angle` like API instead I believe.

PR #2388, opened 2020-05-05
[aat] Fix implementation of AAT kerning for Geeza Pro.

### LINKED ISSUE #2358: "لَمَّا" shapes incorrectly with "Geeza Pro.ttc"
$ ~/dev/harfbuzz/util/hb-view /System/Library/Fonts/GeezaPro.ttc  "لَمَّا"

LAM_MEEM_ALEF

jfkthame:
Despite what the comment in the code used to say, it appears that Geeza Pro
does rely on accumulating kerning values from successive subtables. With
this change, the results now match Core Text rendering (and avoid the clear
visual breakage reported in #2358).

Testcase: U+0644,U+064E,U+0645,U+064E,U+0651,U+0627

Fixes #2358.


jfkthame:
@behdad PTAL - note that this contradicts an explicit comment that you had put in the code previously, but AFAICS this gives results that match Core Text.

(There's an existing testcase using Geeza Pro that fails with this patch, but the new result better matches Core Text so I believe it's correct and we just need to fix the expectation.)

behdad:
Oh thanks for this!  Unfortunately I cannot find any evidence of why I thought CoreText does it that way.

Can we test one thing?  Was the problem in this case more that we were resetting x_offset when v = magic-value?  I'll try to find more but looks good to me.

behdad:
I'm really glad you figured this one out.  I was so stuck didn't know where to start.

behdad:
Messaged Ned to get his input.

jfkthame:
> Can we test one thing? Was the problem in this case more that we were resetting x_offset when v = magic-value? I'll try to find more but looks good to me.

I did wonder if that was the cause when I noticed it, and tested just fixing that, but it wasn't the issue. Fixing that alone didn't make any visible difference in this case, IIRC -- certainly didn't make it right.

behdad:
Thank you.  Ned confirmed CoreText only keeps advances (one glyph position to next) so can only accumulate.

I'm happy with the change.  Let's fix test expectation and add a new test and get it in next release.

jfkthame:
Huh, looks like I didn't get that quite right - will check tomorrow.

drott:
Thanks for the fix, @jfkthame - awesome!

PR #2389, opened 2020-05-06
[subset] Check vector resize() call for failure

jfkthame:
Other .resize() calls are checked, presumably this one should be as well.

### DIFF
--- a/src/hb-subset-cff2.cc
+++ b/src/hb-subset-cff2.cc
@@ -343,7 +343,7 @@ static bool _serialize_cff2 (hb_serialize_context_t *c,
 {
   /* private dicts & local subrs */
   hb_vector_t<table_info_t>  private_dict_infos;
-  private_dict_infos.resize (plan.subset_fdcount);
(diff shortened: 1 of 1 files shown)


behdad:
LGTM but I let Ariza chime in.  How did you run into this?

jfkthame:
> LGTM but I let Ariza chime in. How did you run into this?

Coverity noticed it in a patch to update harfbuzz in gecko.

ebraminio:
Why our coverity check didn't mark it :/ Guess resize also should be considered nodiscard, #2314

behdad:
> Why our coverity check didn't mark it :/

It might have.  I handled some but the bulk of CFF stuff were left for Ariza to check.



> Guess resize also should be considered nodiscard

Not necessarily.  Our vector already handles resize failure and keeps error status.  So is perfectly fine dealing with error later.  But yeah, I'm fine if we nodiscard it and ask call-sites to explicitly `(void)` to communicate they know what they doing.

PR #2390, opened 2020-05-07
Mark hb_vector_t::resize as undiscardable, fix all the uses

ebraminio:
Follow up to #2389


behdad:
This is thoughtless.

behdad:
Again.  The library was for the most part designed such that objects keep their own error status.  So this kind of blanket change is wrong.

ebraminio:
> This is thoughtless.

Right, guess we should discuss them per case.

behdad:
I do think vector.resize() should get one.  But each call site should be analyzed and some silenced if the error is handle later.  Or maybe not.  I haven't reviewed enough of the callsites to form a firm opinion.  I apologize that I let my frustration out quick.

ebraminio:
> But each call site should be analyzed and some silenced if the error is handle later.

I did of course checked each case carefully but I don't claim it is error proof, it is just to see whether we are happy with whole thing or not, one thing it revealed quickly that -Wall will still complain about this even with casting to `(void)`.

PR #2391, opened 2020-05-07
[docs] Don’t recommend outdated FDO releases pages

khaledhosny:


ebraminio:
Thanks :)

behdad:
Okay now only if we can edit the fdo page and mark the project moved...  I'll do it eventually.

khaledhosny:
Can we also redirect harfbuzz.org to harfbuzz.github.io?

behdad:
> Can we also redirect harfbuzz.org to harfbuzz.github.io?

@behnam Can we do that?

behnam:
> > Can we also redirect harfbuzz.org to harfbuzz.github.io?
> 
> @behnam Can we do that?

I can do this, if this is still what you want. So, let me know if that's the case.

behdad:
> I can do this, if this is still what you want. So, let me know if that's the case.

I don't see anything needing action right now. Thanks @behnam

PR #2393, opened 2020-05-08
[subset] fix intersects () for Context/ChainContext tables

qxliu76:
return true only when all values in array are intersected with input for ChainRule/Rule.
integration tests will be added later, along with GSUB5 PR.

### DIFF
--- a/src/hb-ot-layout-gsubgpos.hh
+++ b/src/hb-ot-layout-gsubgpos.hh
@@ -791,15 +791,15 @@ static inline bool intersects_coverage (const hb_set_t *glyphs, const HBUINT16 &
   return (data+coverage).intersects (glyphs);
 }
 
-static inline bool intersects_array (const hb_set_t *glyphs,
-				     unsigned int count,
-				     const HBUINT16 values[],
-				     intersects_func_t intersects_func,
-				     const void *intersects_data)
+static inline bool array_is_subset_of (const hb_set_t *glyphs,
+				       unsigned int count,
+				       const HBUINT16 values[],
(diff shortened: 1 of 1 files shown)


src/hb-ot-layout-gsubgpos.hh
```diff
@@ -798,8 +798,8 @@ static inline bool intersects_array (const hb_set_t *glyphs,
 				     const void *intersects_data)
```
garretrieger, line 798: 'intersects' implies that at least one element is shared between the two sets. Since your changing this to require that all elements in values are present in glyphs can you change the name of this function to something like: array_is_subset_of. That will better communicate what is being checked.

qxliu76: renamed. Thanks!

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -791,15 +791,15 @@ static inline bool intersects_coverage (const hb_set_t *glyphs, const HBUINT16 &
...
+				       unsigned int count,
+				       const HBUINT16 values[],
+				       intersects_func_t intersects_func,
+				       const void *intersects_data)
 {
   for (const HBUINT16 &_ : + hb_iter (values, count))
-    if (intersects_func (glyphs, _, intersects_data)) return true;
-  return false;
+    if (!intersects_func (glyphs, _, intersects_data)) return false;
+  return true;
```
behdad, line 802: I have a feeling that this can use `hb_any` or `hb_all` these days.

PR #2395, opened 2020-05-10
[meson] Prefer C linker and improve exported symbols check

### LINKED ISSUE #2276: [meson] Use C linker for output library
Equivalent of [this](https://github.com/harfbuzz/harfbuzz/blob/748e1cf/src/Makefile.am#L120-L129), our meson script needs someway to avoid linking to stdc++ by using a C linker instead C++.

ebraminio:
Fixes #2276

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -334,6 +334,16 @@ if have_icu and have_icu_builtin
   deps += [icu_dep]
 endif
 
+if host_machine.system() == 'windows' or get_option('with-libstdcxx')
+  chosen_linker = 'cpp'
+else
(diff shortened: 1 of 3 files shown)


src/meson.build
```diff
@@ -368,6 +381,8 @@ libharfbuzz = library('harfbuzz', hb_sources,
   soversion: hb_so_version,
   version: version,
   install: true,
+  #vs_module_defs: harfbuzz_def,
```
ebraminio, line 384: If I enable `vs_module_defs` I will get https://ci.appveyor.com/project/harfbuzz/harfbuzz/builds/32769174/job/jrekl36yixwitpgo#L2131 (and lots warnings above)

PR #2396, opened 2020-05-10
[number] fold strtod_rl into hb_parse_double

ebraminio:
making parser simpler and faster, why all these efforts into this however? because it is also used for font parser inside the library thus I believe worth to have a faster and more reliable code in sense of acting consistently across different `libc`s.


ebraminio:
Have merged its less controversial part to master, let's still have strtod_rl for now. Thanks

PR #2397, opened 2020-05-11
[subset] GSUB5/GPOS7 Contextual Subst/Pos Subtbale Subsetting support

qxliu76:


src/hb-ot-layout-gsubgpos.hh
```diff
@@ -1450,6 +1450,58 @@ struct Rule
     return_trace (context_apply_lookup (c, inputCount, inputZ.arrayZ, lookupCount, lookupRecord.arrayZ, lookup_context));
   }
 
+  bool serialize (hb_serialize_context_t *c,
+		  const hb_map_t *mapping,
```
garretrieger, line 1454: Maybe call this input_mapping to make it more clear what it's a map of.

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -1774,8 +1874,46 @@ struct ContextFormat2
...
+      return_trace (false);
+
+    hb_map_t klass_map;
+    out->classDef.serialize_subset (c, classDef, this, &klass_map);
+
+    hb_vector_t<unsigned> rulesets;
+    const hb_map_t *lookup_map = c->table_tag == HB_OT_TAG_GSUB ? c->plan->gsub_lookups : c->plan->gpos_lookups;
+    bool ret = true;
+    for (const hb_pair_t<unsigned, const OffsetTo<RuleSet>&>& _ : + hb_enumerate (ruleSet))
+    {
```
garretrieger, line 1890: Can you add a comment that explains why you need to retain rules that are otherwise subsetted out. From my understanding this is because we need an entry for each class in the subsetted class map.

qxliu76: oh, my bad, we should only keep rulesets for class numbers that survived.

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -1774,8 +1874,46 @@ struct ContextFormat2
...
+    auto *out = c->serializer->start_embed (*this);
+    if (unlikely (!c->serializer->extend_min (out))) return_trace (false);
+    out->format = format;
+    if (unlikely (!out->coverage.serialize_subset (c, coverage, this)))
+      return_trace (false);
+
+    hb_map_t klass_map;
+    out->classDef.serialize_subset (c, classDef, this, &klass_map);
+
+    hb_vector_t<unsigned> rulesets;
```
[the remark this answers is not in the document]
garretrieger: Thanks for changing to use index. You should be able to remove this vector now.

PR #2411, opened 2020-05-16
meson: error out if introspection is enabled but gobject isn't

### LINKED ISSUE #2404: [meson] error out if introspection=enabled, but gobject is disabled
I was shortly puzzled by this two weeks ago, and then I forgot and was just confused again, so for the sake of future me:

`meson -Dintrospection=enabled` doesn't result in introspection being built because you also need `meson -Dgobject=enabled`.

It would be nice if the built would error out in that case.

If that's something that is wanted I can look into providing a PR.

lazka:
In case a user passed -Dintrospection=enabled the build would just ignore
it by default because gobject defaults to disabled and the introspection build
gets skipped.

Instead, if introspection is explicitly enabled but gobject is for some reason
missing error out.

Fixes #2404

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -657,6 +657,10 @@ if have_gobject
 
   install_headers(hb_gobject_headers, subdir: meson.project_name())
 else
+  if get_option('introspection').enabled()
+    error('introspection requires gobject to be enabled')
+  endif
+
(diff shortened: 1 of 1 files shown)


ebraminio:
tyvm! :)

Issue #2409, opened 2020-05-16
2.6.6: meson build suite does not build documentation using gtk-doc


kloczek:
Looks like currently meson does not allow build documentation using gtk-doc.

tp-m:
I'll have a look at this.

PR #2412, opened 2020-05-16
meson: build documentation with gtk-doc

tp-m:
Fixes https://github.com/harfbuzz/harfbuzz/issues/2409
---

Does something, only lightly tested.

Note that gtk-doc docs are not built by default, only on install. Build `harfbuzz-doc` target explicitly to force a gtk-doc docs build.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -25,6 +25,8 @@ option('tests', type : 'feature', value : 'enabled', yield : true,
   description: 'Enable or disable unit tests')
 option('introspection', type : 'feature', value : 'disabled', yield : true,
   description : 'Generate gobject-introspection bindings (.gir/.typelib files)')
+option('gtk_doc',  type : 'feature', value : 'auto', yield : true,
+  description : 'Generate documentation with gtk-doc')
 
 option('icu-builtin', type: 'boolean', value: false,
   description: 'Don\'t separate ICU support as harfbuzz-icu module')
--- a/docs/Makefile.am
+++ b/docs/Makefile.am
@@ -107,7 +107,7 @@ include $(top_srcdir)/gtk-doc.make
 
 # Other files to distribute
 # e.g. EXTRA_DIST += version.xml.in
-EXTRA_DIST += version.xml.in
+EXTRA_DIST += version.xml.in meson.build
 
 # Files not to distribute
 # for --rebuild-types in $(SCAN_OPTIONS), e.g. $(DOC_MODULE).types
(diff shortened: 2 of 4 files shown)


ebraminio:
Please add docs/meson.build to https://github.com/harfbuzz/harfbuzz/blob/master/docs/Makefile.am#L110 Thanks!

tp-m:
> Please add meson.build to https://github.com/harfbuzz/harfbuzz/blob/master/docs/Makefile.am#L110 to resolve `distcheck` bot issue. Thanks! :)

Done.

meson_options.txt
```diff
@@ -25,6 +25,8 @@ option('tests', type : 'feature', value : 'enabled', yield : true,
   description: 'Enable or disable unit tests')
 option('introspection', type : 'feature', value : 'disabled', yield : true,
   description : 'Generate gobject-introspection bindings (.gir/.typelib files)')
+option('gtk_doc',  type : 'feature', value : 'auto', yield : true,
+  description : 'Generate API documentation with gtk-doc')
```
khaledhosny, line 29:
```suggestion
  description : 'Generate documentation with gtk-doc')
```

We use gtk-doc for the user manual as well.

tp-m: Updated, thanks.

meson_options.txt
```diff
@@ -25,6 +25,8 @@ option('tests', type : 'feature', value : 'enabled', yield : true,
   description: 'Enable or disable unit tests')
 option('introspection', type : 'feature', value : 'disabled', yield : true,
   description : 'Generate gobject-introspection bindings (.gir/.typelib files)')
+option('gtk_doc',  type : 'feature', value : 'auto', yield : true,
```
ebraminio, line 28: maybe `gtk-doc` instead underscore is more natural as a meson option, I'm not sure and is just a suggestion

tp-m: I don't mind. Personally I prefer to have options with hyphens, but `gtk_doc` is what is used everywhere else. IMHO meson should canonicalise hyphens/underscores anyway, but we're not there yet.

While we're at it, I should also point out that other modules use a boolean option here instead of a feature option. To me a feature option (with explicit auto detection) makes more sense, but again, your call.

ebraminio: > but gtk_doc is what is used everywhere else.

Oh I see, even meson documentations prefer that apparently.

> To me a feature option (with explicit auto detection) makes more sense

These all have more to do with my unfamiliarity with meson :)

ebraminio:
Lets merge and address those normalization issues also later. Thanks!

meson_options.txt
ebraminio: Hopefully addressed in https://github.com/harfbuzz/harfbuzz/pull/2419

PR #2415, opened 2020-05-20
minor: fix hb_ot_layout_get_ligature_carets

pipcet:
Hi!
I'm finding it necessary to apply this patch in order to get anything back from `hb_ot_layout_get_ligature_carets`. I'm using `Libertinus Serif Display` (`Regular`). If I'm reading the code correctly, we need to pass in the original caret count this way.

Please let me know if there's anything else you need for this PR to be applied, or if I'm mistaken somehow and the original code is correct.

Thanks!

### DIFF
--- a/src/hb-ot-layout.cc
+++ b/src/hb-ot-layout.cc
@@ -369,7 +369,7 @@ hb_ot_layout_get_ligature_carets (hb_font_t      *font,
 				  unsigned int   *caret_count /* IN/OUT */,
 				  hb_position_t  *caret_array /* OUT */)
 {
-  unsigned int result_caret_count = 0;
+  unsigned int result_caret_count = caret_count ? *caret_count : 0;
   unsigned int result = font->face->table.GDEF->table->get_lig_carets (font, direction, glyph, start_offset, &result_caret_count, caret_array);
   if (result)
   {


src/hb-ot-layout.cc
```diff
@@ -369,7 +369,7 @@ hb_ot_layout_get_ligature_carets (hb_font_t      *font,
 				  unsigned int   *caret_count /* IN/OUT */,
 				  hb_position_t  *caret_array /* OUT */)
 {
-  unsigned int result_caret_count = 0;
+  unsigned int result_caret_count = caret_count ? *caret_count : 0;
   unsigned int result = font->face->table.GDEF->table->get_lig_carets (font, direction, glyph, start_offset, &result_caret_count, caret_array);
```
khaledhosny, line 373: `result_caret_count` is now redundant. Drop it and pass `caret_count` to `get_lig_carets()` directly. The `if (result)` check below becomes no-op as well, so the `else` block should then be `if (!result)`.

khaledhosny: Scrap that. It will break AAT font if `result` is zero. LGTM, but I think we need to test this code since it was broken for a bit now (since b986fead0).

pipcet: I'm trying to test this, but running into issues because, apparently, nothing checks the lcar table exists before dereferencing it. It's just dereferencing random memory, as far as I can tell.

Do we need something like https://github.com/pipcet/harfbuzz/commit/4f4fa18964a090ca919dfde479161f7d86b2557e ?

ebraminio: > but running into issues because

Wow, I see that specific issue, totally unacceptable from my side, am trying to make a fix.

pipcet: Thanks for looking into it!

Any hints as to which utilities would work for building a minimum font with a GDEF table and a ligature caret entry in it? I could start with the Libertinus SFD source code, but that would potentially mean licensing trouble.

ebraminio: Oh, I think I've fixed it in #2416, please have a look, thanks!

ebraminio: > Any hints as to which utilities would work for building a minimum font with a GDEF table and a ligature caret entry in it? I could start with the Libertinus SFD source code, but that would potentially mean licensing trouble.

fonttools is best of course,

We had two fonts with having such property already in our source, Amiri and Noto Urdu Nastaliq, I wanted to use Amiri but saw a gid with three in Noto Urdu Nastaliq so went with that.

PR #2416, opened 2020-05-20
Fix _get_ligature_caret's oob read issue

ebraminio:
AAT::Lookup has no other way to detect whether it is returned from
a real and sanitized font data or from a null pool, this checks if
the table has been recognized valid by sanitizer by checking
table's major version which is zero if returned from a null pool and
non-zero if is from a sanitized font data, it is expected the other
calls of the table (unlikely to have more calls however) also do a
similar version check before calling the lookups used on the table.

cc: @khaledhosny @pipcet

### DIFF
--- a/src/hb-aat-layout-lcar-table.hh
+++ b/src/hb-aat-layout-lcar-table.hh
@@ -116,6 +116,8 @@ struct lcar
 {
   static constexpr hb_tag_t tableTag = HB_AAT_TAG_lcar;
 
+  bool has_data () const { return version.major; }
+
   unsigned int get_lig_carets (hb_font_t      *font,
 			       hb_direction_t  direction,
 			       hb_codepoint_t  glyph,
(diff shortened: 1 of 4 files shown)


src/hb-aat-layout-lcar-table.hh
```diff
@@ -123,6 +123,20 @@ struct lcar
...
+       non-zero if is from a sanitized font data, it is expected the other
+       calls of the table (unlikely to have more calls however) also do a
+       similar version check before calling the lookups used on the table */
+    if (!version.major)
+    {
+      if (caret_count)
+        *caret_count = 0;
+      return 0;
+    }
+
```
behdad, line 139: This is wrong.  Why do u need to know that?  Just add a `has_data()` method and call that.  No comment needed.

ebraminio: > Why do u need to know that? 

As explained I see no other way to check whether lookup is returned from a null pool or not IIUC

PR #2417, opened 2020-05-21
[subset] ChainContext subsetting fix: add lookup index remapping

qxliu76:
Add lookup index remapping in ChainContext subset code, which is added before lookup pruning and remapping code.
Also optimizes the code to get rid of using vector when pruning out empty trailing ruleSets.


qxliu76:
> Can you add a test case, which would have failed with the previous code, but now passes?

Tests are added. Thanks!

PR #2419, opened 2020-05-21
[meson] Normalize options names

ebraminio:
As https://github.com/harfbuzz/harfbuzz/pull/2412#discussion_r426275027

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -28,11 +28,11 @@ option('introspection', type : 'feature', value : 'disabled', yield : true,
 option('gtk_doc',  type : 'feature', value : 'auto', yield : true,
   description : 'Generate documentation with gtk-doc')
 
-option('icu-builtin', type: 'boolean', value: false,
+option('icu_builtin', type: 'boolean', value: false,
   description: 'Don\'t separate ICU support as harfbuzz-icu module')
-option('with-libstdcxx', type: 'boolean', value: false,
-   description: 'Allow linking with libstdc++ (ignored and always enabled in Windows)')
-option('experimental-api', type: 'boolean', value: false,
+option('with_libstdcxx', type: 'boolean', value: false,
+   description: 'Allow linking with libstdc++ (always enabled in Windows)')
+option('experimental_api', type: 'boolean', value: false,
   description: 'Enable experimental APIs')
-option('amalgam', type : 'boolean', value : false,
+option('amalgam', type: 'boolean', value: false,
   description : 'Enable amalgam builds')
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ jobs:
       - run: export PKG_CONFIG_PATH="/usr/local/opt/icu4c/lib/pkgconfig:/usr/local/opt/libffi/lib/pkgconfig" && ./autogen.sh --with-freetype --with-glib --with-gobject --with-cairo --with-icu --with-coretext --with-graphite2
       - run: make -j4
       - run: make check || .ci/fail.sh
-      - run: make clean && meson build -Damalgam=true && ninja -Cbuild test
+      - run: make clean && meson build -Dcoretext=enabled -Damalgam=true && ninja -Cbuild test
 
   distcheck:
     docker:
@@ -235,7 +235,7 @@ jobs:
       # test meson's own unity feature also
       - run: meson build --unity on && ninja -Cbuild && rm -rf build
       # test experimental APIs
-      - run: meson build -Dexperimental-api=true && ninja -Cbuild test && rm -rf build
+      - run: meson build -Dexperimental_api=true && ninja -Cbuild test && rm -rf build
 
   crosscompile-notest-djgpp:
     docker:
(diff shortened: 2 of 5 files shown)


meson_options.txt
```diff
@@ -28,11 +28,11 @@ option('introspection', type : 'feature', value : 'disabled', yield : true,
 option('gtk_doc',  type : 'feature', value : 'auto', yield : true,
   description : 'Generate documentation with gtk-doc')
 
-option('icu-builtin', type: 'boolean', value: false,
+option('icu_builtin', type: 'feature', value: 'auto',
```
khaledhosny, line 31: What does `auto` mean in this context?

ebraminio: Fixed

meson_options.txt
```diff
@@ -28,11 +28,11 @@ option('introspection', type : 'feature', value : 'disabled', yield : true,
 option('gtk_doc',  type : 'feature', value : 'auto', yield : true,
   description : 'Generate documentation with gtk-doc')
 
-option('icu-builtin', type: 'boolean', value: false,
+option('icu_builtin', type: 'feature', value: 'disabled',
   description: 'Don\'t separate ICU support as harfbuzz-icu module')
-option('with-libstdcxx', type: 'boolean', value: false,
-   description: 'Allow linking with libstdc++ (ignored and always enabled in Windows)')
-option('experimental-api', type: 'boolean', value: false,
+option('with_libstdcxx', type: 'feature', value: 'disabled',
```
khaledhosny, line 33: Would it be a good idea to default to `auto` here, and honor explicit `disabled` even on Windows?

ebraminio: > Would it be a good idea to default to auto here

As https://mesonbuild.com/Build-options.html

> If the value of a feature option is set to auto, that value is overridden by the global auto_features option (which defaults to auto). This is intended to be used by packagers who want to have full control on which dependencies are required and which are disabled, and not rely on build-deps being installed (at the right version) to get a feature enabled. They could set auto_features=enabled to enable all features and disable explicitly only the few they don't want, if any.

this feels like something that encourage packagers to enable thus makes sense to keep disabled I believe.

> and honor explicit disabled even on Windows?

I don't have clear memory on why I was thinking it isn't possible, it makes sense, let's see what happens.

ebraminio: >  and honor explicit disabled even on Windows?

Oh, this is why https://github.com/harfbuzz/harfbuzz/commit/20a840c7cd08069720050782b4533cde505df09a

ebraminio: Hmm, but I think we can try again after 8 years.

For testing please note that `link_language: chosen_linker,` is a recent and not released yet addition, expected to be added in 0.55.

(just FYI, feel free to ignore) cc: @lazka (as msys2's harfbuzz packager) and @tp-m (as your previous advise on reevaluate hacks we did before)

tp-m: Why switch to feature options at all if none of them default to `auto`? Not arguing one way or another, mind. I personally prefer the feature option ui to the boolean one, but just wondering what your rationale was.

Also, when you use feature options you need to think about what `auto` should mean for each option (even if it's not the default in `meson_options.txt`) - e.g. for each option you could use `if not option.disabled()` or `if option.enabled()`, so first would capture `auto` inside the if and the second wouldn't. Not a big deal in any case, just idly commenting from the sidelines.

The main advantages of feature options is for optional features/dependencies. You can default to `auto` in `meson_options.txt` and skip optional things if they are not available, but packagers can flip a switch and set all `auto` defaults to `enabled`, and then they'll get an error for any missing optional bits and can decide if they want to explicitly disable that option or update their package's build dependencies so that it works. This way they won't accidentally miss things if new dependencies/requirements are added.

ebraminio: > Why switch to feature options at all if none of them default to auto? Not arguing one way or another, mind. I personally prefer the feature option ui to the boolean one, but just wondering what your rationale was.

I misinterpreted what you said :) Keeping option names normalization.

> While we're at it, I should also point out that other modules use a boolean option here instead of a feature option. To me a feature option (with explicit auto detection) makes more sense, but again, your call.

Which modules you were referring to?

Still not sure on whether we should keep https://github.com/harfbuzz/harfbuzz/commit/20a840c7cd08069720050782b4533cde505df09a or not. Some pointer on that also would be nice. Thanks!

tp-m: > Which modules you were referring to?

glib, gtk, pango, for example.

ebraminio: > glib, gtk, pango, for example.

We don't have gtk and pango around, hmm. Our other yes/no/auto matches with `configure.ac` so guess we are fine here, or not? Guess am missing some other thing also.

lazka: > Still not sure on whether we should keep [20a840c](https://github.com/harfbuzz/harfbuzz/commit/20a840c7cd08069720050782b4533cde505df09a) or not. Some pointer on that also would be nice. Thanks!

No idea what this changes, and if this helps, but harfbuzz builds fine without it using autotools+msys2+mingw32

ebraminio: > No idea what this changes, and if this helps, but harfbuzz builds fine without it using autotools+msys2+mingw32

Done https://github.com/harfbuzz/harfbuzz/commit/d9177a594bc2846b03a43b8d00949d23ab0d91c2 Thank you very much :)

PR #2423, opened 2020-05-21
Add shell script and cross-compilation file to compile Harfbuzz with meson

### LINKED ISSUE #2413: meson and mingw
Hello

for compilation of hb with msys2 + mingw-w64, i would use cross-compilation. Because technically  speaking, it is cross-compilation, and it allows also cross-compilation on linux. I use this kind of file with a sed command to replacewhat is between @ : 

https://github.com/vtorri/ewpi/blob/master/packages/glib2/cross_toolchain.txt

vtorri:
This allow also to cross-compile on Linux for Windows

Fixes #2413

### DIFF
--- a/meson-mingw-cross.txt
+++ b/meson-mingw-cross.txt
@@ -1,0 +1,17 @@
+[binaries]
+c = '@host@-gcc'
+cpp = '@host@-g++'
+ar = '@host@-ar'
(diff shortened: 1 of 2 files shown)


ebraminio:
Looks good! Thanks :)

PR #2429, opened 2020-05-29
[tests] Port check scripts to python

ebraminio:


src/check-c-linkage-decls.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python3
+
+import sys, os
+
+os.chdir (os.environ.get ('srcdir', os.path.dirname (__file__)))
+
+HBHEADERS = os.environ.get ('HBHEADERS', '')
+if len (HBHEADERS):
+	HBHEADERS = [x for x in HBHEADERS.split (' ') if x]
```
behdad, line 9: Just use `HBHEADERS.split ()`.  In other places as well.

ebraminio: needed to trim empty entries passed by autotools

behdad: Did you try what I suggested?

behdad: `.split()` by default does exactly what you are doing here.

ebraminio: wow, nice one, done

src/check-c-linkage-decls.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python3
+
+import sys, os
+
+os.chdir (os.environ.get ('srcdir', os.path.dirname (__file__)))
+
+HBHEADERS = os.environ.get ('HBHEADERS', '')
```
khaledhosny, line 7:
```suggestion
HBHEADERS = os.environ.get ('HBHEADERS')
```

src/check-c-linkage-decls.py
```diff
@@ -0,0 +1,37 @@
...
+	HBHEADERS = HBHEADERS.split ()
+else:
+	HBHEADERS = [x for x in os.listdir ('.') if x.startswith ('hb') and x.endswith ('.h')]
+
+HBSOURCES = os.environ.get ('HBSOURCES', '')
+if len (HBSOURCES):
+	HBSOURCES = HBSOURCES.split ()
+else:
+	HBSOURCES = [x for x in os.listdir ('.')
+				 if x.startswith ('hb') and (x.endswith ('.cc') or x.endswith ('.hh'))]
```
khaledhosny, line 18:
```suggestion
				 if x.startswith ('hb') and x.endswith ((.cc', '.hh'))]
```

src/check-c-linkage-decls.py
```diff
@@ -0,0 +1,37 @@
...
+
+fails = 0
+
+for x in HBHEADERS:
+	with open (x, 'r', encoding='utf-8') as f:
+		content = f.read ()
+		if ('HB_BEGIN_DECLS' not in content) or ('HB_END_DECLS' not in content):
```
khaledhosny, line 25: You can de-indent this block to make it a bit more readable.

src/check-c-linkage-decls.py
khaledhosny: Or shorter (also the all cap variables look weird):
```python
headers = os.environ.get ('HBHEADERS', '').split()
if not headers:
    haeders = [x for x in os.listdir ('.') if x.startswith ('hb') and x.endswith ('.h')]
```

src/check-c-linkage-decls.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python3
+
+import sys, os
+
+os.chdir (os.environ.get ('srcdir', os.path.dirname (__file__)))
+
+HBHEADERS = os.environ.get ('HBHEADERS', '')
+if len (HBHEADERS):
```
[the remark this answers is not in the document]
ebraminio: Niiice, done, thanks :)

src/check-header-guards.py
```diff
@@ -0,0 +1,32 @@
...
+HBSOURCES = os.environ.get ('HBSOURCES', '')
+if HBSOURCES:
+	HBSOURCES = HBSOURCES.split ()
+else:
+	HBSOURCES = [x for x in os.listdir ('.')
+				 if x.startswith ('hb') and (x.endswith ('.cc') or x.endswith ('.hh'))]
+
+fails = 0
+
+for x in (HBHEADERS + HBSOURCES):
```
behdad, line 23: The parentheses not needed.

src/check-c-linkage-decls.py
ebraminio: Done, excellent

src/check-c-linkage-decls.py
ebraminio: Niiiiice, thanks!

src/check-c-linkage-decls.py
ebraminio: Done, see if now is better, thanks :)

src/check-header-guards.py
ebraminio: Done, thanks!

src/check-c-linkage-decls.py
```diff
@@ -0,0 +1,34 @@
...
+
+import sys, os
+
+os.chdir (os.environ.get ('srcdir', os.path.dirname (__file__)))
+
+HBHEADERS = os.environ.get ('HBHEADERS')
+if HBHEADERS:
+	HBHEADERS = HBHEADERS.split ()
+else:
+	HBHEADERS = [x for x in os.listdir ('.') if x.startswith ('hb') and x.endswith ('.h')]
```
behdad, line 11: I think I like Khaled's other suggestion to rewrite these as:
```
headers = os.environ.get ('HBHEADERS', '').split ()
if not headers:
        headers = ...
```

These repetitive 5-line blocks are annoying.

ebraminio: Reduced to a one liner using `or` :)

behdad: Hehe.  BTW, you can also omit the `'.'` to `os.listdir`.

Actually the original scripts pass a `maxdepth=1` to `find`.  So you want `glob` here, which will further simplify things eg. `glob.glob ('hb*.h')`... 

The original script was using `find` for some other reasons.  I see you are `chdir`ing into `srcdir`.  All of this had good reasons for how they were done.  Can you please test this with out-of-tree builds with autotools and make sure it works?

behdad:
You don't need to force-push the commit all the time either.  You can keep pushing tiny address-review commits and we can Squash and merge if needed.

ebraminio:
> You don't need to force-push the commit all the time either. You can keep pushing tiny address-review commits and we can Squash and merge if needed.

Correct, that makes review easier.

src/check-static-inits.py
```diff
@@ -0,0 +1,38 @@
...
+	print ("check-static-inits.py: object files not found; skipping test")
+	sys.exit (77)
+
+stat = 0
+
+for obj in OBJS:
+	result = subprocess.check_output ([objdump, '-t', obj]).decode ('utf-8')
+
+	# Checking that no object file has static initializers
+	for l in re.findall (r'^.*\.[cd]tors.*$', result, re.MULTILINE):
```
ebraminio, line 28: This matches https://github.com/harfbuzz/harfbuzz/blob/0d8442a55f6b34ca3b5f444de2a3cdcce27421ff/src/check-static-inits.sh#L27 logic AFAIK but I couldn't reproduce and test the check using `__attribute__((constructor))` or C++ static initialized variables, something tried more or less for other tests.

behdad: Did you try a static initialized variable in a method?

These tests were developed according to what common compilers were producing at the time.  It's possible some of them are not relevant anymore.  But this one should be still relevant...

behdad: > C++ static initialized variables

With a non-trivial constructor?

I made the following trivial patch to test, but am not on Linux and don't have `objdump`, so this test is skipping for me.
```patch
+static hb_set_t oh_oh;
 /**
  * hb_set_previous_range:
  * @set: a set.
@@ -537,5 +538,6 @@ hb_set_previous_range (const hb_set_t *set,
                       hb_codepoint_t *first,
                       hb_codepoint_t *last)
 {
+  if (oh_oh.has (*first)) return true;
   return set->previous_range (first, last);
 }
```

ebraminio: The example triggers the next test and exposes `__cxa_atexit` but not `.ctors` and `.dtors`

> It's possible some of them are not relevant anymore.

Sure! Completely understandable and nice to have tests, thank you very much! :)

behdad: > Sure! Completely understandable and nice to have tests, thank you very much! :)

IIRC Chrome was porting to harfbuzz-ng then and they brought this to my attention.  It's possible that the first test is not what I could reproduce and I added the second one to work on whatever setup I had.

I just tested both gcc and clang on Fedora and neither emit those segments in anything I tried.  I can dig harder to find, but as you said, doesn't matter. :)

PR #2432, opened 2020-05-30
[test/text-rendering-tests] Adopt the update script with meson

ebraminio:
Is in three changes, porting the script to python, updating using the script, adopting the script with meson.


test/shaping/data/text-rendering-tests/DISABLED
```diff
@@ -7,3 +7,15 @@ tests/SHBALI-1.tests
 tests/SHBALI-2.tests
 tests/SHKNDA-2.tests
 tests/SHKNDA-3.tests
+
+# TODO: Needs investigation
```
ebraminio, line 11: This needs an investigation outside I guess of this change scope.

test/shaping/data/text-rendering-tests/tests/SHLANA-1.tests
```diff
@@ -0,0 +1,52 @@
+../fonts/TestShapeLana.ttf:--font-size=1000 --ned --remove-default-ignorables --font-funcs=ft:U+1A20,U+1A6B:[uni1A20|uni1A6B@762,0]
```
khaledhosny, line 1: Why all these tests are using `ft` font functions?

ebraminio: Some where ok with ot callbacks also and some weren't, not sure why, needs more investigation.

khaledhosny: I see, thanks.

behdad: Might be because we didn't have glyph names implemented in ot-font back when I integrated this test suite.  So, please someone investigate and report. :)

PR #2434, opened 2020-05-31
Build win32 artifacts using meson

ebraminio:
As a part of meson migration, this builds `hb-shape` dependencies using meson subprojects and removes different unrelated mingw files we've put on source root about things that aren't really our concern. This also simplifies all the instructions we've put in `README.mingw.md` and makes it easy to build the project on distros that no mingw libraries (such i686-win32-{freetype,harfbuzz,...} are packaged like ArchLinux and guess Debian also. The known catch is however lack of `hb-view` as its need for build of cairo which apparently isn't that straightforward.


README.mingw.md
```diff
@@ -11,21 +11,12 @@ steps are recommended:
    On Fedora, with `dnf install mingw32-gcc-c++`, or `dnf install mingw64-gcc-c++` for the
    64-bit Windows.
 
-3. Install cross-compiled dependency packages.  Alternatively see [^1] below.
-   On Fedora that would be `dnf install mingw32-glib2 mingw32-cairo mingw32-freetype`
-   for 32-bit, or `dnf install mingw64-glib2 mingw64-cairo mingw64-freetype` for 64-bit.
+3. `.ci/build-win32.sh`
```
khaledhosny, line 14: If we are going to recommend this to people, then `.ci` is not the most appropriate place for it.

ebraminio: Maybe renaming `.ci` to something like `.maintaining` or something? I believe we shouldn't offer any of these as a regular part of the library and README.mingw.md is mostly useful for maintainers and not clients anyway, maybe removing README.mingw.md from repo and putting it somewhere less public like GitHub wiki pages.

ebraminio: A correct point given it is used as a part of release process also but like to avoid putting them again somewhere that accessible.

khaledhosny: Up to you then.

Makefile.am
```diff
@@ -10,7 +10,6 @@ EXTRA_DIST = \
 	autogen.sh \
 	harfbuzz.doap \
 	README.md \
-	README.mingw.md \
```
behdad, line 13: Why is this?

ebraminio: It is really not something intended for the release package.

ebraminio:
> What happened to the mingw-ldd? 

Gone, no need to, all the dependencies are compiled right at the place, libc/winpthread is static linked, wished could static link the others also.

> Now the mingw target strings (x86_64-w64-mingw32 etc) are duplicated all over the place. 

image

Only these, how hard is that, the way it works on other places, https://gitlab.gnome.org/GNOME/glib/-/blob/master/.gitlab-ci/cross_file_mingw64.txt our CI is running `.ci/build-win32.sh` and https://circleci.com/gh/harfbuzz/harfbuzz/142176#artifacts/containers/0 is its result that works independently.

> There's a meson cairo dependency fix in there.

Correct. Happened while trying to enable cairo, was making the build to fail but I should've uploaded that one line separately, the line is copied from it three line later however so nothing that new there.

Issue #2436, opened 2020-06-01
Meson migration


ebraminio:
Let's list things our meson port misses so we know when we can do the actual transition, (whom have access, feel free to edit the post)

* https://github.com/harfbuzz/harfbuzz/blob/e248a4e/test/api/Makefile.am#L189-L207 move the logic to some python script and put the CLEANFILES in .gitignore
* Done: https://github.com/harfbuzz/harfbuzz/blob/6058ede/Makefile.am#L47-L62 <s>do we need CHANGELOG? If so, how</s> #2475
* Done: https://github.com/harfbuzz/harfbuzz/blob/7250ade/src/Makefile.am#L271-L296 put them on some bash script I guess. (uploaded #2438)
* Done: Turning `gobject`, `introspection`, <s>`graphite`</s> features https://github.com/harfbuzz/harfbuzz/blob/master/meson_options.txt from 'disabled' to 'auto'? (uploaded #2437)
* Remove https://github.com/harfbuzz/harfbuzz/blob/6058ede/meson.build#L6

Anything else? Of course there are! Let me know!

behdad:
So, where's the `make distcheck` and `make dist` equivalents in your list?

tp-m:
> So, where's the `make distcheck` and `make dist` equivalents in your list?

`ninja dist` or `meson dist`? (Don't think there's a way to skip the checks yet during dist)

ebraminio:
> So, where's the `make distcheck` and `make dist` equivalents in your list?

As indicated by @tp-m, those aren't separated in meson release process. We just need to make sure our regular build in meson is equivalent with distcheck that's why #2437 is needed.

behdad:
> * [ ]  https://github.com/harfbuzz/harfbuzz/blob/e248a4e/test/api/Makefile.am#L189-L207 move the logic to some python script and put the CLEANFILES in .gitignore
> *

There's no CLEANFILES issue since meson builds out-of-tree only.

ebraminio:
> There's no CLEANFILES issue since meson builds out-of-tree only.

Good point, it can do in tree ones which apparently not common anyway, thanks

heftig:
Meson doesn't version the libraries like autotools does (`.so.0.20608.0` but should be `.so.0.20600.8`). If you want to keep the scheme, you need to use the libtool `current`, `age` and `revision` you calculate to determine the `version` and `soversion`.

ebraminio:
[...]

PR #2438, opened 2020-06-02
Move update unicode tables to its own make file

ebraminio:
Also downloads things needed for its job

#2436 related

Needs to be ran in `src/` using `./update-unicode-tables.make`, or in project's root using `make -C src -f update-unicode-tables.make`

### DIFF
--- a/src/update-unicode-tables.make
+++ b/src/update-unicode-tables.make
@@ -2,7 +2,7 @@
 
 all: packtab arabic-table emoji-table indic-table tag-table ucd-table use-table vowel-constraints
 
-.PHONY: all clean packtab arabic-table indic-table tag-table use-table vowel-constraints emoji-table
+.PHONY: all clean packtab arabic-table emoji-table indic-table tag-table ucd-table use-table vowel-constraints
 
 arabic-table: gen-arabic-table.py ArabicShaping.txt UnicodeData.txt Blocks.txt
 	./$^ > hb-ot-shape-complex-arabic-table.hh || (rm hb-ot-shape-complex-arabic-table.hh; false)
--- a/src/update-unicode-tables.make
+++ b/src/update-unicode-tables.make
@@ -1,6 +1,8 @@
 #!/usr/bin/env -S make -f
 
-all: arabic-table emoji-table indic-table tag-table ucd-table use-table vowel-constraints
+all: packtab arabic-table emoji-table indic-table tag-table ucd-table use-table vowel-constraints
+
+.PHONY: all clean packtab arabic-table indic-table tag-table use-table vowel-constraints emoji-table
 
 arabic-table: gen-arabic-table.py ArabicShaping.txt UnicodeData.txt Blocks.txt
 	./$^ > hb-ot-shape-complex-arabic-table.hh || (rm hb-ot-shape-complex-arabic-table.hh; false)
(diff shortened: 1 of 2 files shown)


ebraminio:
@dscorbett see if it looks good, thanks

src/update-unicode-tables.make
```diff
@@ -0,0 +1,59 @@
...
+tag-table: gen-tag-table.py languagetags language-subtag-registry
+	./$^ > hb-ot-tag-table.hh || (rm hb-ot-tag-table.hh; false)
+ucd-table: gen-ucd-table.py ucd.nounihan.grouped.zip hb-common.h
+	./$^ > hb-ucd-table.hh || (rm hb-ucd-table.hh; false)
+use-table: gen-use-table.py IndicSyllabicCategory.txt IndicPositionalCategory.txt UnicodeData.txt Blocks.txt
+	./$^ > hb-ot-shape-complex-use-table.cc || (rm hb-ot-shape-complex-use-table.cc; false)
+vowel-constraints: gen-vowel-constraints.py ms-use/IndicShapingInvalidCluster.txt Scripts.txt
+	./$^ > hb-ot-shape-complex-vowel-constraints.cc || (hb-ot-shape-complex-vowel-constraints.cc; false)
+
+packtab:
```
dscorbett, line 20: Nothing depends on this target. It should be depended on by `emoji-table` and `ucd-table`.

src/update-unicode-tables.make
```diff
@@ -0,0 +1,59 @@
...
+	/usr/bin/env python3 -c "import packTab" 2>/dev/null || /usr/bin/env pip3 install git+https://github.com/harfbuzz/packtab
+
+.PHONY: packtab arabic-table indic-table tag-table use-table vowel-constraints emoji-table
+
+ArabicShaping.txt:
+	wget https://unicode.org/Public/UCD/latest/ucd/ArabicShaping.txt
```
dscorbett, line 26: wget is not always available. This should fall back to curl.

src/update-unicode-tables.make
```diff
@@ -0,0 +1,59 @@
...
+	./$^ > hb-ucd-table.hh || (rm hb-ucd-table.hh; false)
+use-table: gen-use-table.py IndicSyllabicCategory.txt IndicPositionalCategory.txt UnicodeData.txt Blocks.txt
+	./$^ > hb-ot-shape-complex-use-table.cc || (rm hb-ot-shape-complex-use-table.cc; false)
+vowel-constraints: gen-vowel-constraints.py ms-use/IndicShapingInvalidCluster.txt Scripts.txt
+	./$^ > hb-ot-shape-complex-vowel-constraints.cc || (hb-ot-shape-complex-vowel-constraints.cc; false)
+
+packtab:
+	/usr/bin/env python3 -c "import packTab" 2>/dev/null || /usr/bin/env pip3 install git+https://github.com/harfbuzz/packtab
+
+.PHONY: packtab arabic-table indic-table tag-table use-table vowel-constraints emoji-table
```
dscorbett, line 23: These are not all the phony targets in this file.

src/update-unicode-tables.make
```diff
@@ -0,0 +1,59 @@
...
+	wget https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry
+
+ucd.nounihan.grouped.zip:
+	wget https://unicode.org/Public/UCD/latest/ucdxml/ucd.nounihan.grouped.zip
+
+Scripts.txt:
+	wget https://unicode.org/Public/UCD/latest/ucd/Scripts.txt
+
+clean:
+	rm -f \
```
dscorbett, line 56: As a matter of style, I have generally seen makefiles use `$(RM)` instead of `rm -f`.

src/update-unicode-tables.make
ebraminio: a fallback like this https://github.com/talklittle/erlang.mk/commit/9619a3992e15d066f1ae44f9d838756d9fc7f2e6 is a bit overkill (or, how you implement in a simpler way?, as the apparent lack of preinstalled wget in macos let's use `curl` anyway.

src/update-unicode-tables.make
ebraminio: I didn't know to be honest, done :)

src/update-unicode-tables.make
ebraminio: Done, have a look again

src/update-unicode-tables.make
ebraminio: Breaks `./$^` however :/ Added it to `all`.

src/update-unicode-tables.make
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env -S make -f
+
+all: packtab arabic-table emoji-table indic-table tag-table ucd-table use-table vowel-constraints
+
+.PHONY: all clean packtab arabic-table indic-table tag-table use-table vowel-constraints emoji-table
```
dscorbett, line 5: Another phony target is `ucd-table`.

ebraminio: Oh, done, thanks :)

behdad:
Since this is a split Makefile now, we don't need the `gen-*` targets anymore.  Can be just the output file name.  And the packtab dependency is weird.

You are using both `$(RM)` and also plain `rm` in most other places.

The autodownload is not something I'm comfortable with.  Is not like it always makes sure they are latest either.  So one has to "make clean" then run to get latest.  The actual curl invocations for multiple files in the same remote location rule-sharing should be done.

behdad:
As a general rule, can you please split issues such that when you "move X from Y to Z", you don't introduce other things in the change?

ebraminio:
> Since this is a split Makefile now, we don't need the gen-* targets anymore. Can be just the output file name.

Good point, https://github.com/harfbuzz/harfbuzz/pull/2440

> And the packtab dependency is weird.

maybe, but I don't see why not also. I didn't have it and installed a wrong packtab from pip, so.

> You are using both $(RM) and also plain rm in most other places.

I don't remember, feel free to bring those out!

> As a general rule, can you please split issues such that when you "move X from Y to Z", you don't introduce other things in the change?

Better to show the moved still work, that was the idea.

PR #2443, opened 2020-06-03
[meson] Only pass required dependencies to everything

### LINKED ISSUE #2441: meson question: cairo pkg-config dependencies
We currently get

```
prefix=/mingw64
libdir=${prefix}/lib
includedir=${prefix}/include

Name: harfbuzz
Description: HarfBuzz text shaping library
Version: 2.6.6
Requires.private: glib-2.0, gobject-2.0, cairo, cairo-ft, graphite2, freetype2, fontconfig
Libs: -L${libdir} -lharfbuzz
Libs.private: -lm -lusp10 -lgdi32 -lrpcrt4 -ldwrite
Cflags:-I${includedir}/harfbuzz
```

This makes libharfbuzz depend on cairo, gobject, freetype2, fontconfig, I wonder if this is on purpose?

Previously it was just

```
prefix=/mingw64
exec_prefix=/mingw64
libdir=/mingw64/lib
includedir=/mingw64/include

Name: harfbuzz
Description: HarfBuzz text shaping library
Version: 2.6.4

Libs: -L${libdir} -lharfbuzz
Libs.private: -lm  -lusp10 -lgdi32 -lrpcrt4   
Requires.private: glib-2.0 >= 2.19.1 graphite2 >= 1.2.0
Cflags: -I${includedir}/harfbuzz
```

lazka:
Instead of passing dependencies as required we used one giant shared
dependency list containing all dependencies for every library/executable.
While this kinda works, the specified deps are also used for generating
the pkg-config files and this leads to lots of Requires.private and Libs.private
entries which aren't really needed.

This removes the "deps" array and replaces it with a few smaller ones and
makes sure the public libraries only get passed the dependencies actually
needed.

Fixes #2441

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -311,24 +311,30 @@ incsrc = include_directories('.')
 hb_sources = hb_base_sources + hb_base_ragel_generated_sources
 hb_headers = hb_base_headers
 
+harfbuzz_deps = [thread_dep, m_dep] + harfbuzz_extra_deps
+
 if conf.get('HAVE_FREETYPE', 0) == 1
   hb_sources += hb_ft_sources
   hb_headers += hb_ft_headers
+  harfbuzz_deps += [freetype_dep]
 endif
 
 if conf.get('HAVE_GDI', 0) == 1
   hb_sources += hb_gdi_sources
   hb_headers += hb_gdi_headers
+  harfbuzz_deps += gdi_uniscribe_deps
 endif
 
 if conf.get('HAVE_GRAPHITE2', 0) == 1
   hb_sources += hb_graphite2_sources
   hb_headers += hb_graphite2_headers
+  harfbuzz_deps += [graphite2_dep]
 endif
 
 if conf.get('HAVE_GLIB', 0) == 1
   hb_sources += hb_glib_sources
   hb_headers += hb_glib_headers
+  harfbuzz_deps += [glib_dep]
 endif
 
 if conf.get('HAVE_UNISCRIBE', 0) == 1
@@ -339,11 +345,13 @@ endif
 if conf.get('HAVE_DIRECTWRITE', 0) == 1
   hb_sources += hb_directwrite_sources
   hb_headers += hb_directwrite_headers
+  harfbuzz_deps += directwrite_dep
 endif
 
 if conf.get('HAVE_CORETEXT', 0) == 1
   hb_sources += hb_coretext_sources
   hb_headers += hb_coretext_headers
+  harfbuzz_deps += coretext_deps
 endif
 
 have_icu = conf.get('HAVE_ICU', 0) == 1
@@ -352,7 +360,7 @@ have_icu_builtin = conf.get('HAVE_ICU_BUILTIN', 0) == 1
 if have_icu and have_icu_builtin
   hb_sources += hb_icu_sources
   hb_headers += hb_icu_headers
-  deps += [icu_dep]
+  harfbuzz_deps += [icu_dep]
 endif
 
 if get_option('with_libstdcxx')
@@ -390,7 +398,7 @@ endif
 
 libharfbuzz = library('harfbuzz', hb_sources,
   include_directories: incconfig,
-  dependencies: deps,
+  dependencies: harfbuzz_deps,
   cpp_args: cpp_args + extra_hb_cpp_args,
   soversion: hb_so_version,
   version: version,
@@ -401,7 +409,7 @@ libharfbuzz = library('harfbuzz', hb_sources,
 libharfbuzz_dep = declare_dependency(
   link_with: libharfbuzz,
   include_directories: incsrc,
-  dependencies: deps)
+  dependencies: harfbuzz_deps)
 
 # harfbuzz-subset
 harfbuzz_subset_def = custom_target('harfbuzz-subset.def',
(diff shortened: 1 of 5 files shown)


behdad:
This would have been a lot easier to verify if the var naming and configuration of autotools was retained.  I'm new to meson so can you help me understand how the following works:

In Makefile.am, we keep separate list of libs that come from a library with .pc file versus those that don't.  Only libs without a .pc file need to be listed explicitly in our .pc file.  That's why we have, eg.:
```
HBLIBS   += $(GRAPHITE2_LIBS)
HBDEPS   += $(GRAPHITE2_DEPS)
```
but:
```
HBNONPCLIBS += $(DIRECTWRITE_LIBS)
```

How is that differentiation handled in meson build?

lazka:
It's a bit (documented) magic, meson will automatically add various things to the pkgconfig file as it sees fit: https://mesonbuild.com/Pkgconfig-module.html#implicit-dependencies

In the case you mentioned:

Both dependencies get added to `libharfbuzz`, `pkgmod.generate(libharfbuzz, ...)` will then generate the .pc file.

* The directwrite dependency is created with `cpp.find_library()` which will result in it being added to Libs.private
* The graphite2 dependency is generated through `dependency()` which when found internally with pkg-config will add the result in Requires.private

src/meson.build
```diff
@@ -417,7 +425,7 @@ defs_list += [harfbuzz_subset_def]
 
 libharfbuzz_subset = library('harfbuzz-subset', hb_subset_sources,
   include_directories: incconfig,
-  dependencies: deps,
+  dependencies: [],
```
ebraminio, line 428: Looks very good, thank you very much!

I had to put this to get `libm` from deps, https://github.com/harfbuzz/harfbuzz/commit/bb095e1bdd0f7f1de88651d6f7223bc04bf70d70 in C linker which needs [meson master](https://github.com/mesonbuild/meson/commit/1bfeaadd6d1dc87532a1c0ad4be615eb34044495#diff-69c59a112064153b548fa7e9d432af44) (not released yet 0.55) to test, is `dependencies: [],` makes it somehow work? Thanks

lazka: > I had to put this to get `libm` from deps, [bb095e1](https://github.com/harfbuzz/harfbuzz/commit/bb095e1bdd0f7f1de88651d6f7223bc04bf70d70) as it isn't done automatically in C linker which needs [meson master](https://github.com/mesonbuild/meson/commit/1bfeaadd6d1dc87532a1c0ad4be615eb34044495#diff-69c59a112064153b548fa7e9d432af44) (not released yet 0.55) to test, is `dependencies: [],` makes it somehow work?

I didn't find any libm usage there, but I'll have another look.

> And can we drop empty list of `dependencies: [],` in other places?

Sure, I just left it as "documentation" in case it is needed in the future.

ebraminio: Steps to reproduce:

```
# install meson master
pip install git+https://github.com/mesonbuild/meson
# the revision before that change
git checkout dca8ba6b
# build using meson
rm -rf build && meson build && ninja -Cbuild test
```

```
FAILED: src/libharfbuzz-subset.so.0.20605.0 
clang  -o src/libharfbuzz-subset.so.0.20605.0 'src/25a6634@@harfbuzz-subset@sha/hb-number.cc.o' 'src/25a6634@@harfbuzz-subset@sha/hb-ot-cff1-table.cc.o' 'src/25a6634@@harfbuzz-subset@sha/hb-ot-cff2-table.cc.o' 'src/25a6634@@harfbuzz-subset@sha/hb-static.cc.o' 'src/25a6634@@harfbuzz-subset@sha/hb-subset-cff-common.cc.o' 'src/25a6634@@harfbuzz-subset@sha/hb-subset-cff1.cc.o' 'src/25a6634@@harfbuzz-subset@sha/hb-subset-cff2.cc.o' 'src/25a6634@@harfbuzz-subset@sha/hb-subset-input.cc.o' 'src/25a6634@@harfbuzz-subset@sha/hb-subset-plan.cc.o' 'src/25a6634@@harfbuzz-subset@sha/hb-subset.cc.o' -Wl,--as-needed -Wl,--no-undefined -shared -fPIC -Wl,--start-group -Wl,-soname,libharfbuzz-subset.so.0 src/libharfbuzz.so.0.20605.0 -Wl,--end-group '-Wl,-rpath,$ORIGIN/' -Wl,-rpath-link,~/harfbuzz/build/src
/usr/bin/ld: src/25a6634@@harfbuzz-subset@sha/hb-ot-cff1-table.cc.o: in function `hb_font_t::em_scalef(float, int)':
~/harfbuzz/build/../src/hb-font.hh:625: undefined reference to `roundf'
/usr/bin/ld: src/25a6634@@harfbuzz-subset@sha/hb-subset.cc.o: in function `_plan_estimate_subset_table_size(hb_subset_plan_t*, unsigned int)':
~/harfbuzz/build/../src/hb-subset.cc:64: undefined reference to `sqrt'
clang-10: error: linker command failed with exit code 1 (use -v to see invocation)
```

But am not able to reproduce using current state of the build system :)

ebraminio: > But am not able to reproduce using current state of the build system for some reason :)

Okay, now I see, my fault :/ https://github.com/harfbuzz/harfbuzz/commit/d67553d12fc72c4a89fcb75400829b62d53ca064 please rebase to this, thanks

lazka: I've added `m_dep` back now and remove one remaining empty dependency argument.

ebraminio:
@lazka thank you soooo much for your contribution :)

Issue #1389, opened 2018-11-17
CFF glyph extents performance


behdad:
When we merge CFF... maybe add extents cache to CFF?

khaledhosny:
May be cache glyph extents in general? Getting TrueType glyph extents seems to be currently twice as slow as getting glyph advance, or may be that code is not as optimized (not that getting glyph extents is such a common operation, so it might not be worth it for such a small gain).

behdad:
> Getting TrueType glyph extents seems to be currently twice as slow as getting glyph advance, or may be that code is not as optimized (not that getting glyph extents is such a common operation, so it might not be worth it for such a small gain).

If it's only twice slower, it's pretty good deal!  Getting advance width is looking up into an array by glyph id, scaling the value and returning.  For TrueType glyph extents, it's lookup offsets in one array, then look up in second array, fetch four values (compared to one), scale and return them.  There's no way to make this faster.  A hash table / cache can just make the "finding values" part slower.

behdad:
Both calls also have overhead of a atomic ptr get.  For get_h_advance I optimized that by switching to get_h_advances(), which amortizes the atomic ptr get overhead over multiple glyphs.

behdad:
CFF in comparison, we are talking something like 10 to 100 slower, or worse.

behdad:
cc @blueshade7

khaledhosny:
I understand that CFF is much slower since the values are not precalculated as in glyf table, I was just wondering if glyf can be made any faster (context is that LibreOffice caches glyph extents because getting them from GDI or Freetype is pretty slow (not sure about Core Text) and I’m wondering if I switched to HarfBuzz for that would I still need the cache or not and whether caching in HarfBuzz makes sense).

blueshade7:
If performance is a priority, cache may be helpful, but you might as well think about adding a multi glyph API for reducing overhead per glyph. This especially makes sense with VF. Note HVAR is optional with VF TT. When it’s missing, you have to execute glyf for phantom points just to get advance width ala CFF.

behdad:
Right.  Variations causes heavy computations even for TrueType glyph extents (which we don't do right now), and for glyph advances if HVAR is not available (which we also don't do right now).

`hb_font_t` is currently a very lean and fast object.  I like to keep it that way by default.  But it's also the perfect place to optionally add caches.  The user should be able to configure those.  Design questions are:

[...]

PR #2456, opened 2020-06-08
[ot] Implement _get_glyph_contour_point

### LINKED ISSUE #2454: hb_font_funcs_t::glyph_contour_point is not implemented for --font-funcs ot ?
If I've tested it correctly, `hb_font_funcs_t::glyph_contour_point` with `--font-funcs ot` always returns `0 0`. When tested under debugger, it simply jumps to `hb_font_get_glyph_contour_point_nil` and that's it.

It's mainly used by `hb_font_t::get_glyph_contour_point_for_origin`, which is rarely (if never) used, so it may be overlooked. Or is it intentional?

ebraminio:
Fixes #2454


behdad:
I have reservations about this.  I'll write soon.  But you might want to limit how much time you spend on this.  This feature is obsolete and we might not want to support it.

ebraminio:
> I have reservations about this. I'll write soon. But you might want to limit how much time you spend on this. This feature is obsolete and we might not want to support it.

Wished you could bring this up sooner! I'm considering it done at this, at least the patch exists.

test/api/test-ot-face.c
```diff
@@ -150,7 +150,7 @@ test_font (hb_font_t *font, hb_codepoint_t cp)
   hb_draw_funcs_destroy (funcs);
 #endif
 
-  hb_font_get_glyph_contour_point (font, 0, 0, NULL, NULL);
```
ebraminio: Not my assumption but https://github.com/harfbuzz/harfbuzz/blob/2dda6dd744b192616c65ca4aa5fce8e90fd0ff30/src/hb-font.cc#L416

behdad:
> Wished you could bring this up sooner!

Is literally under 24 hours since the issue was opened...   I was sleep!

You could check and see we don't do in `hb-ft` either.  You could discuss.  But that's not your style, very much to my frustration! :)

Here's a couple of things you can work on that *actually* matter:

  * https://github.com/harfbuzz/harfbuzz/issues/566
  * https://github.com/harfbuzz/harfbuzz/issues/1389#issuecomment-624985478

Go make glyph outline loading for CFF and variable fonts 10 times or more faster!

ebraminio:
> Is literally under 24 hours since the issue was opened... I was sleep!

Oh, nvm about it.

> You could check and see we don't do in hb-ft either. 

Huh? I thought this is the equivalent https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ft.cc#L466-L491

ebraminio:
> Go make glyph outline loading for CFF and variable fonts 10 times or more faster!

Yeah, we should start with benchmarking first, I have some plan for this

behdad:
> Yeah, we should start with benchmarking first, I have some plan for this

+1000 for benchmarking!  Thanks.

behdad:
FWIW when we do add glyph-outline API, we probably should wire these up as well. Hope the commits don't disappear.

PR #2458, opened 2020-06-09
meson: Export symbols for a DLL when not building statically

nirbheek:
`default_library` might be 'both', in which case we will fail to link when building with MSVC.


tp-m:
lgtm

ebraminio:
Thank you very much! :)

PR #2471, opened 2020-06-16
[layout] hb_ot_layout_lookup_get_glyph_alternates, a new API

### LINKED ISSUE #673: Add API to get glyph alternates
Such that apps can do fancy GUI.

ebraminio:
Fixes #673

As https://github.com/harfbuzz/harfbuzz/issues/673#issuecomment-644856082 proposal, let's see how it goes.

<details><summary>Used the following to find a font having alternates,</summary>

```C++
#include "src/harfbuzz.cc"

int main(int argc, char **argv) {
  hb_blob_t *blob = hb_blob_create_from_file (argv[1]);
  unsigned count = hb_face_count (blob);
  for (unsigned i = 0; i < count; ++i) {
    hb_face_t *face = hb_face_create (blob, 0);
    unsigned glyph_count = hb_face_get_glyph_count (face);
    unsigned lookup_count = hb_ot_layout_table_get_lookup_count (face, HB_OT_TAG_GSUB);
    for (unsigned lookup = 0; lookup < lookup_count; ++lookup)
      for (unsigned gid = 0; gid < glyph_count; ++gid)
	if (hb_ot_layout_lookup_get_alternates (face, lookup, gid, 0, nullptr, nullptr))
	{
	  printf ("%s\n", argv[1]);
	  goto done;
	}
    done:
    hb_face_destroy (face);
  }
  hb_blob_destroy (blob);
  return 0;
}
```
</details>


ebraminio:
just to mention, it works with Rana Kufi also very well as it was the font I was using for testing initially.

behdad:
Thanks @ebraminio.  This is very nice!

Generally we push most of the parameters from the API down to where they are populated.  Instead you pull out the internal array and copy in the public API.  Any particular reason you did it that way?  If not, I prefer it if you do it the way other methods are done.  Thanks again!

ebraminio:
PTAL

jfkthame:
> Quick show of hands if name should be `get_alternates`, `get_glyph_alternates`, or `get_alternate_glyphs`?
> 
> I feel like `get_alternates` is wrong because it implies it returns alternates to a lookup.
> 
> The `get_glyph_alternates` rightly specifies that this is a lookup method that returns certain data about a glyph.
> 
> The `get_alternate_glyphs` doesn't specify the input, but the output.
> 
> As of now I prefer the second one, `get_glyph_alternates`.

Yeah, `get_glyph_alternates` sounds OK to me.

Does this handle alternates referenced via an Extension subtable?

behdad:
> Does this handle alternates referenced via an Extension subtable?

Doesn't.  I'll address that in my review.

behdad:
Continuing in https://github.com/harfbuzz/harfbuzz/pull/2473

PR #2476, opened 2020-06-19
[layout] Use dagger in hb_ot_layout_feature_get_characters

ebraminio:


ebraminio:
Let me break this into multiple commits

ebraminio:
PTAL

src/hb-ot-layout.cc
```diff
@@ -1710,12 +1710,6 @@ hb_ot_layout_feature_get_name_ids (hb_face_t       *face,
  * Fetches a list of the characters defined as having a variant under the specified
  * "Character Variant" ("cvXX") feature tag.
  *
- * <note>Note: If the char_count output value is equal to its input value, then there
- *       is a chance there were more characters defined under the feature tag than were
- *       returned. This function can be called with incrementally larger start_offset
- *       until the char_count output value is lower than its input value, or the size
- *       of the characters array can be increased.</note>
- *
```
behdad, line 1718: Would be great if we can have a section in our docs that explains this API scheme with examples, and link to it from all such API.  Maybe file an issue..

ebraminio: #2480

PR #2477, opened 2020-06-19
Use dagger in OffsetTable::get_table_tags

ebraminio:


behdad:
Aren't these beautiful?! :)

ebraminio:
Agreed

PR #2478, opened 2020-06-19
[math] Use dagger

ebraminio:


ebraminio:
oops, early upload

src/hb-ot-math-table.hh
```diff
@@ -515,10 +515,10 @@ struct MathGlyphAssembly
     if (parts_count)
     {
       int64_t mult = font->dir_mult (direction);
-      hb_array_t<const MathGlyphPartRecord> arr = partRecords.sub_array (start_offset, parts_count);
-      unsigned int count = arr.length;
-      for (unsigned int i = 0; i < count; i++)
-	arr[i].extract (parts[i], mult, font);
+      + hb_enumerate (partRecords.sub_array (start_offset, parts_count))
```
behdad, line 518: Why enumerate then array-lookup?!  Just zip it with `hb_array (parts, parts_count)`!

src/hb-ot-math-table.hh
```diff
@@ -563,13 +563,11 @@ struct MathGlyphConstruction
...
-      unsigned int count = arr.length;
-      for (unsigned int i = 0; i < count; i++)
-      {
-	variants[i].glyph = arr[i].variantGlyph;
-	variants[i].advance = font->em_mult (arr[i].advanceMeasurement, mult);
-      }
+      + mathGlyphVariantRecord.sub_array (start_offset, variants_count)
+      | hb_map ([&] (const MathGlyphVariantRecord &record) -> hb_ot_math_glyph_variant_t
+		{ return {record.variantGlyph, font->em_mult (record.advanceMeasurement, mult)}; })
+      | hb_sink (hb_array (variants, *variants_count))
```
behdad, line 569: This works, but maybe make it like the other as well.

src/hb-ot-math-table.hh
ebraminio: Hehehe, nice, done

src/hb-ot-math-table.hh
ebraminio: Liked the previous one but also like to make it more consistent also, so, done

PR #2483, opened 2020-06-20
[fvar] minor

ebraminio:
so no repeated calls to get_axes and in multiple levels, which probably was optimized anyway yet feel better personally for clarity.

### DIFF
--- a/src/hb-ot-var-fvar-table.hh
+++ b/src/hb-ot-var-fvar-table.hh
@@ -146,10 +146,9 @@ struct fvar
   {
     if (axes_count)
     {
-      hb_array_t<const AxisRecord> arr = hb_array (&(this+firstAxis), axisCount).sub_array (start_offset, axes_count);
-      const AxisRecord *axes = get_axes ();
+      hb_array_t<const AxisRecord> arr = get_axes ().sub_array (start_offset, axes_count);
       for (unsigned i = 0; i < arr.length; ++i)
-	axes[start_offset + i].get_axis_deprecated (&axes_array[i]);
+	arr[i].get_axis_deprecated (&axes_array[i]);
     }
     return axisCount;
   }
@@ -161,10 +160,9 @@ struct fvar
   {
     if (axes_count)
     {
-      hb_array_t<const AxisRecord> arr = hb_array (&(this+firstAxis), axisCount).sub_array (start_offset, axes_count);
-      const AxisRecord *axes = get_axes ();
+      hb_array_t<const AxisRecord> arr = get_axes ().sub_array (start_offset, axes_count);
       for (unsigned i = 0; i < arr.length; ++i)
-        axes[start_offset + i].get_axis_info (start_offset + i, &axes_array[i]);
+	arr[i].get_axis_info (start_offset + i, &axes_array[i]);
     }
     return axisCount;
   }
--- a/src/hb-ot-var-fvar-table.hh
+++ b/src/hb-ot-var-fvar-table.hh
@@ -140,14 +140,6 @@ struct fvar
   unsigned int get_axis_count () const { return axisCount; }
 
 #ifndef HB_DISABLE_DEPRECATED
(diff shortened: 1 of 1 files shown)


behdad:
Beautiful. Thanks. How about zip & foreach? :D  I hate that we can't use automatic lambda's yet.

ebraminio:
> How about zip & foreach?

That `start_offset + i` ruins everything :/

behdad:
> That `start_offset + i` ruins everything :/

Right. You need to also zip in an `enumerate`. Horrible. :)

ebraminio:
> > That `start_offset + i` ruins everything :/
> 
> Right. You need to also zip in an `enumerate`. Horrible. :)

Maybe something like this ease things https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/for-each-indexed.html

behdad:
> Maybe something like this ease things https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/for-each-indexed.html

We'd have `hb_tuple_t` if it wasn't stupidly hard in current C++.

ebraminio:
> We'd have hb_tuple_t if it wasn't stupidly hard in current C++.

Why it needs that? I mean just passing index in `hb_apply_indexed` like API, it already possible to do with `hb_reduce` also.

behdad:
> Why it needs that?

Because you want to zip two arrays and an enumerator together.

behdad:
> hb_apply_indexed

Ugh no.

behdad:
The thing with this style is that there's only three operations: map, reduce, apply. There's nothing else. :)

ebraminio:
Was thinking something like,

```
+ hb_zip (get_axes ().sub_array (start_offset, axes_count), axes_array)
| hb_apply_indexed ([&] (unsigned i, hb_pair_t<...> &_) { _.first.get_axis_info (start_offset + i, _.second); })
;
```

having indexed suffixed helper will simplified some enumerate uses also.

ebraminio:
Was thinking something like,

```
+ hb_zip (get_axes ().sub_array (start_offset, axes_count), axes_array)
| hb_apply_indexed ([&] (unsigned i, hb_pair_t<...> &_) { _.first.get_axis_info (start_offset + i, _.second); })
;
```

having indexed suffixed helper will simplified some enumerate uses also.

ebraminio:
We have `hb_map_retains_sorting` and `hb_map_sorted` with no documentation,
can you somewhere there were needed?

behdad:
> having indexed suffixed helper will simplified some enumerate uses also.

No no no. :)

Let me fixed that for you since you don't listen. :)

behdad:
> We have `hb_map_retains_sorting` and `hb_map_sorted` with no documentation, can you somewhere there were needed?

It's a workaround for what was a bad / incomplete design. I added the concept of sortedness to the iterators. Then some APIs were marked to require sorted iterator. The intent was to enforce that requirement.

However, then Garret etc found that our zip, map, etc were not propagating that property to their satisfaction. In looking into that it became clear why. So we upgraded sortedness to be more like sorted-monotone or not. That got complicated quick so I didn't finish that if I remember correctly. Instead, added this extra map API for client to tell it if the results will have same sortedness as input, or to declare results sorted. It's a hack and one that I like to clean up.  Let me dig up details...

behdad:
Oh. Actually. The monotone etc stuff was separate issue.

This is needed because we have no way to know if a mapping function retains sortedness. Ie. if you have a sorted iterator and you map it through function `f`, you can't easily know whether output is sorted.

ebraminio:
Feel free to add that somewhere in the code.

> > having indexed suffixed helper will simplified some enumerate uses also.
> 
> No no no. :)
> 
> Let me fixed that for you since you don't listen. :)

The comment is written before seeing yours, as having some bandwidth issue.

behdad:
> Oh. Actually. The monotone etc stuff was separate issue.

This is what I was talking about: https://github.com/harfbuzz/harfbuzz/blob/9510e91/src/hb-iter.hh#L530-L554

PR #2484, opened 2020-06-20
[meson] Mark longer tests with slow

ebraminio:
So one can skip them easily by `meson test -Cbuild --no-suite slow`

This decreases my time to running the test suite to 6s.

```

Ok:                 304 
Expected Fail:      0   
Fail:               0   
Unexpected Pass:    0   
Skipped:            15  
Timeout:            0   

meson test -Cbuild --no-suite slow  26.65s user 12.29s system 586% cpu 6.635 total
```

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -680,7 +680,7 @@ if get_option('tests').enabled()
     test(name, find_program(name + '.py'),
       env: env,
       depends: name == 'check-symbols' ? defs_list : [],
-      suite: ['src'],
+      suite: ['src'] + (name == 'check-static-inits' ? ['slow'] : []),
     )
   endforeach
 endif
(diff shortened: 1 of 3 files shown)


src/meson.build
```diff
@@ -680,7 +680,7 @@ if get_option('tests').enabled()
     test(name, find_program(name + '.py'),
       env: env,
       depends: name == 'check-symbols' ? defs_list : [],
-      suite: ['src'],
+      suite: ['src'] + (name == 'check-static-inits' ? ['slow'] : []),
```
behdad, line 683: Surprised about this one.

ebraminio: Marked ones takes longer than 1s and this ones takes 2s, perhaps can be optimized somehow, yet, marking as slow doesn't mean to skip the test.

ebraminio: Marked ones takes longer than 1s and this ones takes 2s, perhaps can be optimized somehow, yet, marking as slow doesn't mean to skip the test.

behdad: > Marked ones takes longer than 1s and this ones takes 2s, perhaps can be optimized somehow, yet, marking as slow doesn't mean to skip the test.

Okay yeah if 1s is the limit I can see that. Nothing to optimize, it's calling `readelf` or equivalent.

PR #2485, opened 2020-06-21
minor, replace single hb_apply daggers with foreach

ebraminio:


behdad:
Can we discuss when we use foreach vs dagger?

ebraminio:
> Can we discuss when we use foreach vs dagger?

We had some sort of agreement on previous round replacing of dagger to foreach when we have only level of dagger and one apply but sure. I just here tried to apply that consistently.

behdad:
> We had some sort of agreement on previous round replacing of dagger to foreach when we have only level of dagger and one apply but sure. I just here tried to apply that consistently maybe some can be reverted.

I know. And I thank you for that.

Since you have better understanding of the dagger stuff now, I'm suggesting that you and I reconsider what we want to do, and then document it. We should start a documentation page for our iterator framework. If you start I'll try to contribute to. :)

behdad:
Can we have markdown docs in tree that get built and used by our gtk-doc-based docs? That would help me write a lot of documentation. cc @n8willis @simoncozens

ebraminio:
> Can we discuss when we use foreach vs dagger?

I like non purely functional languages providing functional structs so the programmer can choose best way to explain a logic, if something can be explained shorter in dagger, that's better and if dagger won't contribute much (in terms of having smaller and clearer code, the second of course is more subjective so we can focus on first) I like to not have dagger there, I tried dagger for #2483 and really wanted to have dagger there as that unifies our all and saw it can't make the code shorter (let's character count wise) thus clearer (at least at my level of familirarity with dagger) and I didn't turned it to dagger, and in cases here foreach resulted in smaller code so I used it.

behdad:
To me where currently the dagger fails is whenever we need a lambda, since we don't have C++17 auto lambda templates so we have to spell out the parameter types.

behdad:
As for daggers versus foreach, let me say this: if we like to, in the future, we can simply update hb-iter such that many daggers run using multiple threads... Not saying we are going to. But the structure encoded in the dagger allows us full control on such things.  That's what C++ does as well: https://en.cppreference.com/w/cpp/header/execution

ebraminio:
hmmm, makes sense

PR #2488, opened 2020-06-22
Remove concept of stable/unstable from build tools

ebraminio:
We never practiced the concept of stable and unstable releases, let's
remove the code and always use the stable scheme.

With this we will just always use stable scheme, nothing different from "stable" releases but we may need to tag 2.7.0 after this change to not have confused versioning.

About the meson `version =` part, there is no consistent state between the changes those this is in one commit.

using autotools and issuing 2.7.0, before this change,

`/bin/sh ../libtool  --tag=CC   --mode=link gcc  -g -O2  -Bsymbolic-functions -o libharfbuzz.la  -lm -version-info 20600:8:20600 -no-undefined   -rpath /usr/local/lib libharfbuzz_la-hb-aat-layout.lo ... libharfbuzz_la-hb-ft.lo -lglib-2.0  -lfreetype         `

After the change

`/bin/sh ../libtool  --tag=CC   --mode=link gcc  -g -O2  -Bsymbolic-functions -o libharfbuzz.la  -lm -version-info 20608:0:20608 -no-undefined   -rpath /usr/local/lib libharfbuzz_la-hb-aat-layout.lo ... libharfbuzz_la-hb-ft.lo                -lglib-2.0  -lfreetype         `

Of course is different but only for our "unstable" release, but what happens when we want build 2.7.8, before this change but building the version 2.7.8,

`/bin/sh ../libtool  --tag=CC   --mode=link gcc  -g -O2  -Bsymbolic-functions -o libharfbuzz.la  -lm -version-info 20708:0:20708 -no-undefined   -rpath /usr/local/lib libharfbuzz_la-hb-aat-layout.lo ... libharfbuzz_la-hb-ft.lo -lglib-2.0  -lfreetype `

After the change

`/bin/sh ../libtool  --tag=CC   --mode=link gcc  -g -O2  -Bsymbolic-functions -o libharfbuzz.la  -lm -version-info 20708:0:20708 -no-undefined   -rpath /usr/local/lib libharfbuzz_la-hb-aat-layout.lo ... libharfbuzz_la-hb-ft.lo                -lglib-2.0  -lfreetype`

No difference.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -379,7 +379,7 @@ harfbuzz_def = custom_target('harfbuzz.def',
     output: 'harfbuzz.def')
 defs_list = [harfbuzz_def]
 
-version = '0.' + '0'.join(meson.project_version().split('.')) + '.0'
+version = '0.@0@.0'.format(hb_version_int)
 
 extra_hb_cpp_args = []
 if cpp.get_id() == 'msvc'
--- a/configure.ac
+++ b/configure.ac
@@ -46,16 +46,7 @@ AC_SUBST(HB_VERSION)
 # Libtool version
 m4_define([hb_version_int],
 	  m4_eval(hb_version_major*10000 + hb_version_minor*100 + hb_version_micro))
-m4_if(m4_eval(hb_version_minor % 2), [1],
-      dnl for unstable releases
-      [m4_define([hb_libtool_revision], 0)],
-      dnl for stable releases
-      [m4_define([hb_libtool_revision], hb_version_micro)])
-m4_define([hb_libtool_age],
-	  m4_eval(hb_version_int - hb_libtool_revision))
-m4_define([hb_libtool_current],
-	  m4_eval(hb_libtool_age))
-HB_LIBTOOL_VERSION_INFO=hb_libtool_current:hb_libtool_revision:hb_libtool_age
+HB_LIBTOOL_VERSION_INFO=hb_version_int:0:hb_version_int
 AC_SUBST(HB_LIBTOOL_VERSION_INFO)
 
 AC_ARG_WITH([libstdc++],
--- a/meson.build
+++ b/meson.build
@@ -12,14 +12,7 @@ hb_version_micro = hb_version_arr[2].to_int()
 
 # libtool versioning
 hb_version_int = hb_version_major*10000 + hb_version_minor*100 + hb_version_micro
-if hb_version_minor % 2 == 1
-  hb_libtool_revision = 0                # for unstable releases
-else
-  hb_libtool_revision = hb_version_micro # for stable releases
-endif
-hb_libtool_age = hb_version_int - hb_libtool_revision
-hb_libtool_current = hb_libtool_age
-hb_libtool_version_info = '@0@:@1@:@2@'.format(hb_libtool_current, hb_libtool_revision, hb_libtool_age)
+hb_libtool_version_info = '@0@:0:@0@'.format(hb_version_int)
 
 pkgmod = import('pkgconfig')
 cpp = meson.get_compiler('cpp')


src/meson.build
```diff
@@ -379,7 +379,7 @@ harfbuzz_def = custom_target('harfbuzz.def',
     output: 'harfbuzz.def')
 defs_list = [harfbuzz_def]
 
-version = '0.' + '0'.join(meson.project_version().split('.')) + '.0'
```
ebraminio, line 382: Custom and wrong logic to mimic our "stable" versioning.

heftig:
LGTM!

heftig:
I wonder why the code is taking the trouble to set `hb_so_version = ''` on Windows and `'0'` only elsewhere. Doesn't Meson ignore the soversion anyway when the linker is MSVC?

ebraminio:
> I wonder why the code is taking the trouble to set `hb_so_version = ''` on Windows and `'0'` only elsewhere. Doesn't Meson ignore the soversion anyway when the linker is MSVC?

Interesting point, apparently is done on https://github.com/harfbuzz/harfbuzz/commit/4b4d5c295d584809e0281b27687757e895e7edfc by @fanc999 for cmake compatibility.

heftig:
> Interesting point, apparently is done on [4b4d5c2](https://github.com/harfbuzz/harfbuzz/commit/4b4d5c295d584809e0281b27687757e895e7edfc) by @fanc999 for cmake compatibility.

Never mind. [The meson test cases](https://github.com/mesonbuild/meson/tree/master/test%20cases/windows/7%20dll%20versioning) show that `soversion = 0` (or `'0'`) should indeed result in the suffix being `-0.dll`.

nirbheek:
> Never mind. [The meson test cases](https://github.com/mesonbuild/meson/tree/master/test%20cases/windows/7%20dll%20versioning) show that `soversion = 0` (or `'0'`) should indeed result in the suffix being `-0.dll`.

If you folks ever increment the soversion, I recommend having the soversion suffix for all windows builds (not just mingw), so you have versioned DLLs :)

ebraminio:
> > Never mind. [The meson test cases](https://github.com/mesonbuild/meson/tree/master/test%20cases/windows/7%20dll%20versioning) show that `soversion = 0` (or `'0'`) should indeed result in the suffix being `-0.dll`.
> 
> If you folks ever increment the soversion, I recommend having the soversion suffix for all windows builds (not just mingw), so you have versioned DLLs :)

From https://www.freedesktop.org/wiki/Software/HarfBuzz/ "Other, peripheral, headers are more likely to go through minor modifications, but again, we do our best to never change API in an incompatible way. We will never break the ABI." guess that means increasing soversion is very unlikely :) (or maybe not?)

fanc999:
Hi,

The reasoning for my change that was mentioned was that we didn't have versioned DLLs before for Visual Studio builds, but I was not sure about the others, so it was more like a Visual Studio-specific change to not use such named DLL (but keeping everything as-is in the Meson build files, since I could not say the same for the other builds), so that it makes it easier to move to a Meson-based build from a CMake build.

I am actually open to the idea of using the `soversion` for Visual  Studio builds though.

With blessings, and cheers!

ebraminio:
Merged, so @nirbheek and @heftig are somewhat in favor of removing soversion special casing for Visual Studio and @fanc999 isn't opposed also so I will upload soversion normalization patch separately, thank you all and please have a look if is anything is missing or can be improved on meson port :)

nirbheek:
There is benefit to having the same ABI as another build system (whether Autotools or CMake), so I am actually in favour of keeping the current system in place. My comment was just that if you get a chance to break ABI at some point (perhaps that will never happen), versioned DLLs are a honking good idea. :)

ebraminio:
Oh, I see, so let's postpone the change if so. Thank you very much :)

behdad:
Yes we definitely not changing soname. But this change doesn't do that; or at least can be done in a way to not do that.

behdad:
Definitely make sure on Unix systems, as well as Windows, library built after this change will be accepted and loaded by code compiled against HarfBuzz versions before the change. That's the only thing we care about. Can someone confirm that please? If that's not the case, let's discuss why and whether it can be fixed. Thanks.

ebraminio:
> Definitely make sure on Unix systems, as well as Windows, library built after this change will be accepted and loaded by code compiled against HarfBuzz versions before the change. That's the only thing we care about. Can someone confirm that please? If that's not the case, let's discuss why and whether it can be fixed. Thanks.

I really don't like to get caught by this last minute, as the first post shows this really just changes the things to use stable scheme for .so naming but make it always and have shown that using exact commands also, as this of course needs more testing, I went for building the package using ArchLinux aur pkg builder https://aur.archlinux.org/cgit/aur.git/plain/PKGBUILD?h=mingw-w64-harfbuzz (which work upon on top git's master and uses autotools) then installed the package on the system and tested it and things were OK so guess the change is ok but can't do absolute guarantee more this.

While doing so I noticed ArchLinux official repository already uses our meson build https://git.archlinux.org/svntogit/packages.git/commit/trunk/PKGBUILD?h=packages/harfbuzz&id=9b101344f580c72410df9d3b42e7aaa340f2a474 which got very delighted with it.

Then noted Alpine Linux https://git.alpinelinux.org/aports/tree/main/harfbuzz/APKBUILD also has done the switch and this is in addition to msys2 https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-harfbuzz/PKGBUILD which has done the switch from 2.6.6 and discovered and patched its issue.

Thank you all dear packagers, we were provided meson as an experimental build system but you went on testing, discovering its issues and providing fixes for it, and now you have used it already in your distributions, you are just awesome :)

Issue #2489, opened 2020-06-23
Memory leak of blob when allocation of closure fails...


sebras:
I found a memory leak in Harfbuzz 2.6.4, which still appears to be present in git HEAD. This happens when using a custom allocator that can control where it will return `NULL`.

The backtrace relevant for Harfbuzz for the memory leak is here:

```
    #0 0x10dffc6 in hb_blob_t* hb_object_create<hb_blob_t>() harfbuzz/src/hb-object.hh:227:24
    #1 0x10dfcc8 in hb_blob_create harfbuzz/src/hb-blob.cc:94:16
    #2 0xd3055c in hb_ft_face_create harfbuzz/src/hb-ft.cc:617:12
    #3 0xd30c3c in hb_ft_font_create harfbuzz/src/hb-ft.cc:697:10
```

This happens when `calloc()` returns NULL here:

```
#0  0x00000000010e2e0e in _hb_face_for_data_closure_create (blob=0x6060000001a8, index=0) at harfbuzz/src/hb-face.cc:145
#1  0x00000000010e2ca1 in hb_face_create (blob=0x6060000001a8, index=0) at harfbuzz/src/hb-face.cc:203
#2  0x0000000000d305a6 in hb_ft_face_create (ft_face=0x61a000003088, destroy=0x0) at harfbuzz/src/hb-ft.cc:621
#3  0x0000000000d30c3d in hb_ft_font_create (ft_face=0x61a000003088, destroy=0x0) at harfbuzz/src/hb-ft.cc:697
```
Looking at the [source code](https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-face.cc#L203) makes it evident that `hb_blob_reference()` is called on `blob` before being passed to `_hb_face_for_data_closure_create()`. If creating `closure` succeeds `_hb_face_for_data_closure_destroy()` will at a later point call `hb_blob_destroy()` on `closure->blob` to destroy it.

The problem occurs when `calloc()` allocating `closure` fails at the top of `_hb_face_for_data_closure_create()`. Then the reference to `blob` obtained by `hb_face_create()` is never destroyed. The best approach is that `hb_face_create()` calls `hb_blob_destroy()` on `blob` when it fails to obtain the requested `closure`. I will submit a pull request doing this. Hopefully you can include this to resolve this issue. :)

behdad:
Thanks. I'm interested to hear about your use-case as well as experience with using a custom allocator with HarfBuzz if you can share.

[...]

PR #2490, opened 2020-06-23
[face] Destroy blob if face closure can not be created

sebras:
This fixes issue #2489.

### DIFF
--- a/src/hb-face.cc
+++ b/src/hb-face.cc
@@ -200,10 +200,15 @@ hb_face_create (hb_blob_t    *blob,
   if (unlikely (!blob))
     blob = hb_blob_get_empty ();
 
-  hb_face_for_data_closure_t *closure = _hb_face_for_data_closure_create (hb_sanitize_context_t ().sanitize_blob<OT::OpenTypeFontFile> (hb_blob_reference (blob)), index);
+  blob = hb_sanitize_context_t ().sanitize_blob<OT::OpenTypeFontFile> (hb_blob_reference (blob));
+
+  hb_face_for_data_closure_t *closure = _hb_face_for_data_closure_create (blob, index);
 
   if (unlikely (!closure))
+  {
+    hb_blob_destroy (blob);
     return hb_face_get_empty ();
+  }
 
   face = hb_face_create_for_tables (_hb_face_for_data_reference_table,
 				    closure,


sebras:
I see that the coverage test fails to reach 67% coverage for the new allocation error handling code. Looking through `test/api`, `test/fuzzing`, etc. I can't find that there is an existing framework for testing such error handling code. I'll need some advice of how to add that in this case, should that be a requirement for this PR to be merged.

behdad:
We don't currently have a "fork & fail malloc" or similar test setup. Ideas about how to set up are welcome.

behdad:
Oh. This is wrong slightly wrong. The "sanitize" call might return a different blob. Just assign that to blob first before the creation of the closure.

sebras:
> We don't currently have a "fork & fail malloc" or similar 
> test setup. Ideas about how to set up are welcome.

Doing something similar to MuPDF, i.e. `-Dhb_malloc_impl=my_malloc` and have `my_malloc()` have a triggerable allocation limit may be possible. I think it would require HarfBuzz to be linked statically into the test binary, and the execution time might be prohibitive. I can try to play around with something in a separate pull request to illustrate my idea. Doing this will take some time (I'm not well versed in HarfBuzz's test framework).

behdad:
> have a triggerable allocation limit

Yeah but that doesn't do a thorough job...
fork-and-fail shouldn't be that bad. In cairo we used to do these with LD_PRELOAD, so didn't require a differently compiled harfbuzz either.

sebras:
Thanks for guiding me to make a better fix, and merging it. :)

> Yeah but that doesn't do a thorough job...

I'm not sure I understand what it would miss?
If you have an example it would be helpful.

> fork-and-fail shouldn't be that bad.
> In cairo we used to do these with LD_PRELOAD, so 
> didn't require a differently compiled harfbuzz either.

Looking in cairos git I found [sphinx](https://cgit.freedesktop.org/cairo/tree/util/cairo-sphinx/sphinx.c#n2) -- is this what you referred to?

The test parent there transfers information over a socket to each child, for an allocator test that ought to be the desired point at which the child's allocator should return `NULL`. By replacing `malloc()` for via `LD_PRELOAD` it would not just test HarfBuzz, but also all libraries it and HarfBuzz itself depend on. I can't make up my mind whether that's a stability benefit or a problem (not HarfBuzz's code).

behdad:
> I'm not sure I understand what it would miss?

Yeah same as what you said about the runtime.

> The test parent there transfers information over a socket to each child, for an allocator test that ought to be the desired point at which the child's allocator should return `NULL`. By replacing `malloc()` for via `LD_PRELOAD` it would not just test HarfBuzz, but also all libraries it and HarfBuzz itself depend on. I can't make up my mind whether that's a stability benefit or a problem (not HarfBuzz's code).

Right. In cairo we started hitting issues in FreeType. And you don't want to hit issues in FreeType...

> Looking in cairos git I found [sphinx](https://cgit.freedesktop.org/cairo/tree/util/cairo-sphinx/sphinx.c#n2) -- is this what you referred to?

Sorry I don't remember. Has been quite a while. No doesn't look like that. Nevermind. My be my bad memory. I don't want to send you goose-chasing but check DBus as well. I'm fairly sure that one had.

behdad:
at any rate the logic is simple: on every malloc you fork and fail in the child...

PR #2498, opened 2020-06-24
Meson: cleanup how freetype2 dep is looked

xclaesse:


meson.build
```diff
@@ -73,13 +73,13 @@ if not get_option('freetype').disabled()
...
-    endif
+    freetype_dep = cpp.find_library('freetype', required: false,
+                                    has_headers: ['ft2build.h'], )
   endif
 
-  if not freetype_dep.found() and get_option('freetype').enabled()
-    freetype_dep = dependency('freetype2', fallback: ['freetype2', 'freetype_dep'])
+  if not freetype_dep.found()
+    freetype_dep = dependency('freetype2', fallback: ['freetype2', 'freetype_dep'],
+                              default_options: ['harfbuzz=disabled'])
```
ebraminio, line 82: thank you! can we not have this one implicitly?

xclaesse:
Note `default_options: ['harfbuzz=disabled']`, when we build harfbuzz and fallback to build freetype2 we obviously cannot have freetype2 try to find harfbuzz. Similarly freetype2 should also do `default_options: ['freetype2=disabled']` when it fallback to harfbuzz. That way if the main project lookup for freetype2 first it gets freetype2 built with harfbuzz support but harfbuzz without freetype2 support. And if the main project lookup for harfbuzz first it gets harfbuzz with freetype2 support, but freetype2 without harfbuzz support. Does that make sense?

meson.build
xclaesse: what do you mean?

xclaesse:
I did the same for cairo too now.

I also opened a PR on freetype2 meson port: https://github.com/centricular/freetype2/pull/2.

xclaesse:
There is still an issue with cairo dependency check, see https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/37#note_548939

meson.build
ebraminio: do we necessarily need to handle it ourselves?

> Does that make sense?

It is understandable, yes.

ebraminio:
For some reason this has caused our `meson-gcc-mingw` bot to fail now, which can be fixed quickly by disabling cairo explicitly on https://github.com/harfbuzz/harfbuzz/blob/master/.ci/build-win32.sh#L5 (as it was disabled implicitly before) but can you help us on that front also so running `.ci/build-win32.sh` be able to compile cairo also? The reason I ask you is you are providing patches cross the different meson ports so maybe this helps you to make configuration better everywhere now that we have this failure has hit here.

meson.build
```diff
@@ -126,12 +126,15 @@ if not get_option('cairo').disabled()
...
+                                 has_headers: ['cairo.h'])
   endif
 
-  if not cairo_dep.found() and get_option('cairo').enabled()
+  if not cairo_dep.found()
+    # Note that we don't hae harfbuz -> cairo -> freetype2 -> harfbuzz fallback
+    # dependency cycle here because we have configured freetype2 above with
+    # harfbuzz support disabled, so when cairo will lookup freetype2 dependency
+    # it will be forced to use that one.
     cairo_dep = dependency('cairo', fallback: ['cairo', 'libcairo_dep'])
```
ebraminio, line 138: probably this should have `required: get_option('cairo')`

xclaesse: good point, added a commit for that.

meson.build
```diff
@@ -126,13 +127,17 @@ if not get_option('cairo').disabled()
...
-      cairo_dep = cpp.find_library('cairo')
-    endif
+    cairo_dep = cpp.find_library('cairo', required: false,
+                                 has_headers: ['cairo.h'])
   endif
 
-  if not cairo_dep.found() and get_option('cairo').enabled()
-    cairo_dep = dependency('cairo', fallback: ['cairo', 'libcairo_dep'])
+  if not cairo_dep.found()
+    # Note that we don't hae harfbuz -> cairo -> freetype2 -> harfbuzz fallback
```
ebraminio, line 135: hae -> have, harfbuz -> harfbuzz :)

meson.build
xclaesse: Meson currently abort when a cycle is detected but I guess it could instead return not-found, that would make sense. But in the meantime I think it's still better to disable harfbuzz explicitly, I don't think it ever make sense to build freetype with harfbuzz support when we are currently building harfbuzz... does it?

meson.build
xclaesse: fixed.

xclaesse:
> can you help us on that front also so running `.ci/build-win32.sh` be able to compile cairo also?

Cairo meson port is not merged yet, it's in a MR: https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/37. Once that get merged all you need is adding cairo.wrap file in subprojects/

meson.build
```diff
@@ -73,13 +73,14 @@ if not get_option('freetype').disabled()
...
-    endif
+    freetype_dep = cpp.find_library('freetype', required: false,
+                                    has_headers: ['ft2build.h'], )
   endif
 
-  if not freetype_dep.found() and get_option('freetype').enabled()
-    freetype_dep = dependency('freetype2', fallback: ['freetype2', 'freetype_dep'])
+  if not freetype_dep.found()
+    freetype_dep = dependency('freetype2', fallback: ['freetype2', 'freetype_dep'],
+                              default_options: ['harfbuzz=disabled'],
```
ebraminio, line 82: just one thing, can we have yet one other attempt before going for tweaking default option? Just that let's have it last resort, and please rebase the change as I had to fork cairo wrap https://github.com/harfbuzz/harfbuzz/pull/2517/files/d5d15f985e9eb6ef67cb5b29811f1d7d1c50543a#diff-c8106e8bf415bce9acbd51c8f11eb092 to make it work.

ebraminio:
Merged, thank you very much for the contribution, will apply some minor things on top of it :)

meson.build
xclaesse: Commented on https://github.com/harfbuzz/harfbuzz/pull/2519

PR #2504, opened 2020-06-27
Add hb_array_t::lfind

ebraminio:
Based on hb_sorted_array_t::bfind

### DIFF
--- a/src/hb-ot-var-fvar-table.hh
+++ b/src/hb-ot-var-fvar-table.hh
@@ -173,18 +173,11 @@ struct fvar
   bool
   find_axis_deprecated (hb_tag_t tag, unsigned *axis_index, hb_ot_var_axis_t *info) const
   {
-    hb_array_t<const AxisRecord> axes = get_axes ();
     unsigned i;
-    if (axes.lfind (tag, &i))
-    {
-      if (axis_index)
-	*axis_index = i;
-      axes[i].get_axis_deprecated (info);
-      return true;
-    }
-    if (axis_index)
-      *axis_index = HB_OT_VAR_NO_AXIS_INDEX;
-    return false;
+    if (!axis_index) axis_index = &i;
+    *axis_index = HB_OT_VAR_NO_AXIS_INDEX;
+    hb_array_t<const AxisRecord> axes = get_axes ();
+    return axes.lfind (tag, axis_index) && (axes[*axis_index].get_axis_deprecated (info), true);
   }
 #endif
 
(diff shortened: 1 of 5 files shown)


src/hb-ot-var-fvar-table.hh
```diff
@@ -172,14 +174,14 @@ struct fvar
...
-	return true;
-      }
+    unsigned i;
+    if (axes.lfind (tag, &i))
+    {
+      if (axis_index)
+	*axis_index = i;
+      axes[i].get_axis_deprecated (info);
+      return true;
+    }
```
behdad, line 184: Another way to do this, which I typically like, is:
```
unsigned i;
if (!axis_index) axis_index = &i;
return axes.lfind (tag, axis_index) && (axes[*axis_index].get_axis_deprecated (info), true);
```

ebraminio: Done

src/hb-ot-var-fvar-table.hh
```diff
@@ -173,18 +173,11 @@ struct fvar
...
-	*axis_index = i;
-      axes[i].get_axis_deprecated (info);
-      return true;
-    }
-    if (axis_index)
-      *axis_index = HB_OT_VAR_NO_AXIS_INDEX;
-    return false;
+    if (!axis_index) axis_index = &i;
+    *axis_index = HB_OT_VAR_NO_AXIS_INDEX;
+    hb_array_t<const AxisRecord> axes = get_axes ();
```
behdad, line 179: We should more aggressively use `auto` in such cases. What do you think?

ebraminio: was thinking about it but couldn't make myself that enthusiast about it.

ebraminio: There was a time that `var`/`val` proposals to Java were ignored for code clarity reasons and later they later changed their and added the feature, I think we can also keep some balance, use it where really makes things unclean looking, or maybe I'm just incorrect.

behdad: Use it where it helps reading code. In this case it doesn't. `get_axes ()` name suggests it returns an array of some axis record.  The actual type is irrelevant. Since it's an array, we know we can iterate it. And we just case about that.

If it was returning `ArrayOf<>` instead of `hb_array_t`, this code doesn't care. That's why to me it should be `auto` here.

behdad: Ie. we've moved on to writing generic code. Hardcoding types unnecessarily works against that.

PR #2506, opened 2020-06-27
Add dedicated shaping and extents performance measurement tools

ebraminio:
Build the project like,

```
meson build -Dbenchmark=true -Dexperimental_api=true -Doptimization=2
ninja -Cbuild
```

Now you have different options,

* Run all the benchmarks,

```
ninja -Cbuild && build/perf/perf
```

* Run shaping perf tool with a specific font and text,

```
ninja -Cbuild && build/perf/perf-shaping-bench perf/fonts/Amiri-Regular.ttf perf/texts/fa-thelittleprince.txt
```

* Run extents perf tool with a specific font,

```
ninja -Cbuild && build/perf/perf-extents-bench test/api/fonts/Estedad-VF.ttf
```

Other than Google Benchmark based benchmarks there is also `run.sh` which builds `hb-shape` itself
and runs the `perf` tool afterward usually provided by your package manager.

### DIFF
--- a/perf/README.md
+++ b/perf/README.md
@@ -1,0 +1,32 @@
+Benchmarking
+============
+
+Build the project like,
+
+```
+meson build -Dbenchmark=true -Dexperimental_api=true -Doptimization=2
+ninja -Cbuild
+```
+
+Now you have different options,
+
+* Run all the benchmarks,
+
+```
+ninja -Cbuild && build/perf/perf
+```
+
+* Run shaping perf tool with a specific font and text,
+
+```
+ninja -Cbuild && build/perf/perf-shaping-bench perf/fonts/Amiri-Regular.ttf perf/texts/fa-thelittleprince.txt
+```
+
+* Run extents perf tool with a specific font,
+
+```
+ninja -Cbuild && build/perf/perf-extents-bench test/api/fonts/Estedad-VF.ttf
+```
+
+Other than Google Benchmark based benchmarks there is also `run.sh` which builds `hb-shape` itself
+and runs the `perf` tool afterward usually provided by your package manager.
--- a/perf/meson.build
+++ b/perf/meson.build
@@ -7,3 +7,17 @@ benchmark('perf', executable('perf', 'perf.cc',
   link_with: [libharfbuzz],
   install: false,
 ), workdir: meson.current_source_dir() / '..', timeout: 100)
+
+executable('perf-shaping-bench', 'perf-shaping-bench.cc',
+  dependencies : [google_benchmark_dep],
+  include_directories: [incconfig, incsrc],
+  link_with: [libharfbuzz],
+  install: false,
+)
+
+executable('perf-extents-bench', 'perf-extents-bench.cc',
+  dependencies : [google_benchmark_dep, freetype_dep],
+  include_directories: [incconfig, incsrc],
+  link_with: [libharfbuzz],
+  install: false,
+)
--- a/perf/perf-extents-bench.cc
+++ b/perf/perf-extents-bench.cc
@@ -1,0 +1,39 @@
+#define PERF_BENCH
+#include "perf-extents.hh"
+
+#include <stdlib.h>
+
+const char *font_path = "test/api/fonts/Estedad-VF.ttf";
+bool skip_var = false;
+
+static void extents_bench (benchmark::State &state, bool is_var, bool is_ft)
+{
+  if (is_var && skip_var) return;
+  extents (state, font_path, is_var, is_ft);
+}
+BENCHMARK_CAPTURE (extents_bench, ot, false, false);
+BENCHMARK_CAPTURE (extents_bench, ft, false, true);
+BENCHMARK_CAPTURE (extents_bench, ot/var, true, false);
+BENCHMARK_CAPTURE (extents_bench, ft/var, true, true);
+
+int main(int argc, char** argv)
+{
+  if (argc > 1 && argv[1][0] != '-') { font_path = argv[1]; ++argv; --argc; }
+
+  {
+    hb_blob_t *blob = hb_blob_create_from_file (font_path);
+    hb_face_t *face = hb_face_create (blob, 0);
+    hb_blob_destroy (blob);
+    if (!hb_ot_var_get_axis_count (face))
+    {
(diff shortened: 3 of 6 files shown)


ebraminio:
Guess we should revisit the namings, using replacing `perf` in folder and tools names with `bench`, and …?

behdad:
What's the `-Dbenchmark=true` for? Why not just build always like our test suites?

The tool should have one name `hb-perf` for example and take subcommands. Eg.:
```
$ hb-perf --general-perf-args shape font-file text --various-shape-args
```
How it internally works need to be designed, not first implemented!

Also should somehow integrate with Linux `perf` command.

ebraminio:
> What's the -Dbenchmark=true for? Why not just build always like our test suites?

It isn't an installed package thus should be always downloaded and that isn't something packagers like, guess we should have some devmode eventually which isn't that different from that `-Dbenchmark=true` anyway.

> How it internally works need to be designed, not first implemented!
> Also should somehow integrate with Linux perf command.

gbenchmark isn't that flexible so anything can be designed with them, or maybe it is but not using they usual and public APIs, guess your usecase somehow is incompatible with gbenchmark but more with `run.sh` like solution which runs `perf` itself also.

behdad:
Again, you ignored my question about why one needs to enable benchmarking. Just do the right thing. People are busy doing things they care about, not about learning our build systems... Stop adding clutter, stop adding distraction. Make the right decision for our users of it can be made.

behdad:
Okay I'll go see what this gbenchmark is 

I don't like what meson is doing with enabling bad behavior. In an autotools world, we'd just port gbenchmark or gtest in-tree. But with meson everything is depended out.

behdad:
Currently if glib is not available, we can't run most of our tests. Should automatically do same for perf as well.

ebraminio:
> Currently if glib is not available, we can't run most of our tests. Should automatically do same for perf as well.

Yeah. The situation is a bit different as that usually exists as a system package yet this one isn't, at least not in official packages of my distro.

behdad:
> Yeah. The situation is a bit different as that usually exists as a system package yet this one isn't, at least not in official packages of my distro.

In the good old days of autotools there was only two ways:
- You import a package into your tree, or
- User is responsible to install it.

Now with meson that boundary is made very unclear. I understand that much.  So If I understand correctly, meson can look for installed package, and if not found, try git cloning? Does it have a way to tell it to not download anything? If yes, then I still think perf should be enabled by default as "auto" and if dependencies could not be satisfied then it would be off.

And yes in our configure report would be great is we can explain why each feature was turned on or off, like we do in cairo. Eg.:
```
Graphite2: disabled (default; enable with --with-graphite2)
FreeType2: disabled (auto; install freetype2)
```
If the build system can detect the operating system it can even give `dnf`, `apt-get`, `brew`, etc instructions.  *That*'s what I was hoping more sophisticated build systems would do to try to integrate with package managers.

behdad:
@ebraminio can you look into using googletest first/as well? I like to clean up our unit-test suite before we dig into building the perf suite.  Thanks in advance!

ebraminio:
> Does it have a way to tell it to not download anything?

It does, `-Dwrap_mode=nodownload`, packagers should use that

> If yes, then I still think perf should be enabled by default as "auto" and if dependencies could not be satisfied then it would be off.

Correct. Let's fix that separately.

ebraminio:
> Now with meson that boundary is made very unclear. I understand that much.

A bit unrelated but the comment somehow reminds me meson's maintainer take to cargo, https://github.com/mesonbuild/meson/pull/2617#issuecomment-379417837 which maybe is a good read to makes us somehow trust on their decision I guess.

behdad:
> A bit unrelated but the comment somehow reminds me meson's maintainer take to cargo, [mesonbuild/meson#2617 (comment)](https://github.com/mesonbuild/meson/pull/2617#issuecomment-379417837) which maybe is a good read to makes us somehow trust on their decision I guess.

Thanks. Will read.

Many people, like me and I suppose @khaledhosny, @simoncozens, etc work on various projects in the stack at the same time. Ie. you are adding a feature in raqm or crowbar, you end up adding new API in HarfBuzz calling into other new API in FreeType, etc. Not all of them all the time. But still, you want to have one checkout of each in our source directory and install them all in your user directory and they detect each other and work.

This whole "meson will download every dependency by default" is a misguided goal in my opinion. Works for large projects like Chromium, etc, but is not conductive to people who work, across multiple modules equally, not just to use them but to develop multiple modules simultanuously.

I've raised the same issue in our fontmake pipeline as well. @anthrotype's solution has mostly been `pip install -e`, which while good in principle, is a huge hack.

I C libraries there was a similar hack, with `pango-uninstalled.pc` etc to do the same: use the package without `make install`. I'm not asking for that. I'm happy to "make install -j10" each module I touch.

ebraminio:
> across multiple modules equally, not just to use them but to develop multiple modules simultanuously.

maybe one can use soft link for the directory, it isn't that bad as the path it resolves a subproject is deterministic and when sees the directory exists does nothing AFAIK, not much different from git submodules, cmake similar process https://github.com/harfbuzz/uharfbuzz/blob/bcebfee9f659309ef1a1074ad090d41b86a03688/CMakeLists.txt#L10 feels a little more scary as checks out harfbuzz at `_skbuild/linux-x86_64-3.8/cmake-build/hb/src/harfbuzz/`.

khaledhosny:
> This whole "meson will download every dependency by default" is a misguided goal in my opinion. Works for large projects like Chromium, etc, but is not conductive to people who work, across multiple modules equally, not just to use them but to develop multiple modules simultanuously.

The download and compile stuff is not something that meson does by default, it is something our meson build files are actively opting to do. So I think if we don’t want this to be the default behavior, we should change it.

behdad:
> The download and compile stuff is not something that meson does by default, it is something our meson build files are actively doing. So I think if we don’t want this to be the default behavior, we should change it.

That's concerning. Humm. I'll get to learn meson and jump in in a couple weeks I hope.

ebraminio:
> That's concerning. Humm. I'll get to learn meson and jump in in a couple weeks I hope.

Fixed by Khaled's https://github.com/harfbuzz/harfbuzz/pull/2548

Let me close this also as guess that isn't what is wanted. Thanks

Issue #2496, opened 2020-06-24
Variable font: differently rounded x-advance between --font-funcs=ot and --font-funcs=ft


justvanrossum:
With the attached variable font, I get different x-advances, depending on which `--font-funcs` I use:

```
$ hb-shape HBTest-VF.ttf --variations=TEST=491 --font-funcs=ft A
[A=0+496]
$ hb-shape HBTest-VF.ttf --variations=TEST=491 --font-funcs=ot A
[A=0+495]
```
The unrounded x-advance is 495.5.

The [OT spec](https://docs.microsoft.com/en-us/typography/opentype/spec/otvaroverview) contains this bit:

> Multiply the fractional component by 65536, and round the result to the nearest integer (for fractional values of 0.5 and higher, take the next higher integer; for other fractional values, truncate).

However, this applies to "when converting from float or double data types to 16.16", which isn't what's happening here. So strictly speaking, the OT spec is silent on this (I haven't found any other references to rounding).

Still I think there's a point in arguing that the rounded-up .5 value is the correct one: both FreeType and DirectWrite round like this, as do most traditional rounding funcs that do not apply bankers rounding.

Interestingly enough, at another value along the TEST axis, the ot funcs _do_ round up:

```
$ hb-shape HBTest-VF.ttf --variations=TEST=509 --font-funcs=ft A
[A=0+505]
$ hb-shape HBTest-VF.ttf --variations=TEST=509 --font-funcs=ot A
[A=0+505]
```

The unrounded x-advance is 504.5 in this case.

The difference between the two cases is in the *delta* for the x-advance: for `TEST=491` the x-advance delta is negative (the default is 500), and for `TEST=509` it is positive. So that might be a clue as to why the rounding is different at all.

[HBTest-VF.ttf.zip](https://github.com/harfbuzz/harfbuzz/files/4825569/HBTest-VF.ttf.zip)

ebraminio:
just to add uharfbuzz currently uses [this](https://github.com/harfbuzz/harfbuzz/blob/385d64eef1d5dd37187310a08ca6f0408a2e4625/src/hb.hh#L502-L509) implementation which apparently have the result as libc's roundf (as far as I could check by [adding HAVE_ROUNDF](https://github.com/harfbuzz/uharfbuzz/blob/985eda0/setup.py#L16) to defined macros or removing it)

justvanrossum:
[...]

PR #2509, opened 2020-06-28
Always round up

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/2496

### DIFF
--- a/src/hb.hh
+++ b/src/hb.hh
@@ -499,14 +499,9 @@ static_assert ((sizeof (hb_var_int_t) == 4), "");
 #define HB_VAR_ARRAY 1
 #endif
 
-static inline double
-_hb_roundf (float x)
-{
-  return x >= 0 ? floor ((double) x + .5) : ceil ((double) x - .5);
-}
-#ifndef HAVE_ROUNDF
+static inline float
+_hb_roundf (float x) { return floorf (x + .5f); }
 #define roundf(x) _hb_roundf(x)
-#endif
 
 /* Endian swap, used in Windows related backends */
 static inline uint16_t hb_uint16_swap (const uint16_t v)
--- a/configure.ac
+++ b/configure.ac
@@ -68,7 +68,7 @@ GTK_DOC_CHECK([1.15],[--flavour no-tmpl])
 ])
 
 # Functions and headers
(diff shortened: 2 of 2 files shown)


behdad:
meson might need updating to not look for `roundf`.

src/hb.hh
```diff
@@ -499,14 +499,12 @@ static_assert ((sizeof (hb_var_int_t) == 4), "");
 #define HB_VAR_ARRAY 1
 #endif
 
-static inline double
+static inline float
 _hb_roundf (float x)
 {
-  return x >= 0 ? floor ((double) x + .5) : ceil ((double) x - .5);
+  return ceilf (x - .5f);
```
justvanrossum, line 505: This rounds halfway values down, not up. I think it should be `floorf (x + .5f)`.

behdad: Thanks. The tests pass now.

ebraminio:
> meson might need updating to not look for `roundf`.

https://github.com/harfbuzz/harfbuzz/blob/master/meson.build#L67 and https://github.com/harfbuzz/harfbuzz/blob/master/CMakeLists.txt#L91

src/hb.hh
```diff
@@ -499,14 +499,9 @@ static_assert ((sizeof (hb_var_int_t) == 4), "");
...
 
-static inline double
-_hb_roundf (float x)
-{
-  return x >= 0 ? floor ((double) x + .5) : ceil ((double) x - .5);
-}
-#ifndef HAVE_ROUNDF
+static inline float
+_hb_roundf (float x) { return floorf (x + .5f); }
 #define roundf(x) _hb_roundf(x)
```
khaledhosny, line 504: Shouldn’t this be renamed `hb_roundf` (or even `hb_otround` like FontTools, or some other name), and be used directly so that who ever reads the code/uses it in the future knows it is not the libc `roundf`?

behdad: I thought about that. But figured we don't want someone writing code in the future to use `roundf` by mistake. There's no reason I can currently think of that we might want the default `roundf` behavior.

khaledhosny: All good then.

behdad: Thanks. I let Ebrahim check bots and merge.

PR #2511, opened 2020-06-28
[meson] Turn benchmark to a feature

ebraminio:
https://github.com/harfbuzz/harfbuzz/pull/2506#issuecomment-650814046

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -26,6 +26,20 @@ EXTRA_DIST = \
 	subprojects/libffi.wrap \
 	subprojects/proxy-libintl.wrap \
 	subprojects/zlib.wrap \
+	subprojects/google-benchmark.wrap \
+	perf/meson.build \
+	perf/perf-draw.hh \
+	perf/perf-extents.hh \
+	perf/perf-shaping.hh \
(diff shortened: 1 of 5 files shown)


behdad:
Thanks.

Makefile.am
```diff
@@ -26,6 +26,12 @@ EXTRA_DIST = \
 	subprojects/libffi.wrap \
 	subprojects/proxy-libintl.wrap \
 	subprojects/zlib.wrap \
+	subprojects/google-benchmark.wrap \
+	perf/meson.build \
+	perf/perf-draw.hh \
+	perf/perf-extents.hh \
+	perf/perf-shaping.hh \
+	perf/perf.cc \
```
ebraminio, line 34: not happy on putting perf on dist, as there it is hard to check we in autotools dist in meson, yet, isn't that but undecided.

ebraminio: so may is ready to merge if you see the change in the right direction.

behdad:
I'm fine as long as autotools builds fine and ships in tarball what's needed, even if can only be enabled using meson.

PR #2515, opened 2020-06-30
[fvar] minor

ebraminio:


src/hb-ot-var-fvar-table.hh
```diff
@@ -95,20 +92,57 @@ struct AxisRecord
...
+
   bool sanitize (hb_sanitize_context_t *c) const
   {
     TRACE_SANITIZE (this);
     return_trace (c->check_struct (this));
   }
 
-  public:
+  protected:
+  void fill_values (float &default_value, float &min_value, float &max_value) const
```
behdad, line 137: `fill_values` is a very bad and wrong name. And since you are passing references in, reading the call site it's not even clear which direction data flows.

`get_triplet` or whatever.

behdad: `get_shape` maybe even. I also like them ordered as `min, default, max`. Same way that for other things we use `min, peak, max`.

ebraminio: Done

ebraminio: How about `get_coordinates`, that reflect what spec calls them

behdad:
Thanks. Now that you fixed the name, put back the references. Pointers can be null while references can't be.

ebraminio:
> Thanks. Now that you fixed the name, put back the references. Pointers can be null while references can't be.

Done. Thanks!

PR #2517, opened 2020-06-30
[meson] enable cairo as subproject

ebraminio:
Needed to build hb-view in Windows

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -139,12 +139,18 @@ if not get_option('cairo').disabled()
     if cairo_dep.type_name() == 'pkgconfig'
       cairo_ft_dep = dependency('cairo-ft', required: get_option('cairo'))
     else
-      if cpp.has_header('cairo-ft.h') and \
-         cpp.has_function('cairo_ft_font_face_create_for_ft_face', dependencies: cairo_dep)
-        cairo_ft_dep = cairo_dep
-      else
-        cairo_ft_dep = dependency('', required: false)
-      endif
+      cairo_ft_dep = cairo_dep
+      # TODO: Apparently the following isn't reliable to detect whether a dependency provides
+      #       something or not so let's assume it is always enabled as we control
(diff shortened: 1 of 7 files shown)


subprojects/cairo.wrap
```diff
@@ -0,0 +1,5 @@
+[wrap-git]
+directory=cairo
+url=https://github.com/ebraminio/cairo.git
```
ebraminio, line 3: needed to apply https://github.com/ebraminio/cairo/commit/e6f69054671d4c004cb1f0be4a1edbb4c4c1810d to make it work

meson.build
```diff
@@ -139,12 +139,18 @@ if not get_option('cairo').disabled()
...
-        cairo_ft_dep = dependency('', required: false)
-      endif
+      cairo_ft_dep = cairo_dep
+      # TODO: Apparently the following isn't reliable to detect whether a dependency provides
+      #       something or not so let's assume it is always enabled as we control
+      #       cairo subproject anyway for now.  To fix and test run .ci/build-win32.sh and
+      #       make sure hb-view.exe is not missing.
+      #
+      # if cpp.has_header('cairo-ft.h') and \
+      #    cpp.has_function('cairo_ft_font_face_create_for_ft_face', dependencies: cairo_dep)
```
ebraminio, line 149: cairo_ft availability won't be detected correctly as explained above how to reproduce, since pkgconfig case is dealt separately let's consider it is always available.

subprojects/cairo.wrap
ebraminio: some other hard coding done on https://github.com/ebraminio/cairo/commit/c30fa2f0893527af1f7a41bd6f9ac6fc10e99e27 as the options weren't yield: 'true' https://mesonbuild.com/Build-options.html#yielding-to-superproject-option (at least explicitly?)

PR #2519, opened 2020-07-01
Meson freetype

ebraminio:
minor things related to freetype in meson

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -151,18 +151,12 @@ if not get_option('cairo').disabled()
     if cairo_dep.type_name() == 'pkgconfig'
       cairo_ft_dep = dependency('cairo-ft', required: get_option('cairo'))
     else
-      cairo_ft_dep = cairo_dep
-      # TODO: Apparently the following isn't reliable to detect whether a dependency provides
-      #       something or not so let's assume it is always enabled as we control
-      #       cairo subproject anyway for now.  To fix and test run .ci/build-win32.sh and
-      #       make sure hb-view.exe is not missing.
-      #
-      # if cpp.has_header('cairo-ft.h') and \
-      #    cpp.has_function('cairo_ft_font_face_create_for_ft_face', dependencies: cairo_dep)
-      #   cairo_ft_dep = cairo_dep
-      # else
-      #   cairo_ft_dep = dependency('', required: false)
-      # endif
+      cairo_ft_dep = dependency('cairo', fallback: ['cairo', 'libcairo_dep'],
+                                has_headers: ['cairo-ft.h'],
+                                required: get_option('cairo'))
+      if cairo_ft_dep.found()
+        cairo_dep = cairo_ft_dep
+      endif
     endif
   else
     cairo_ft_dep = dependency('', required: false)
--- a/meson.build
+++ b/meson.build
@@ -73,13 +73,20 @@ if not get_option('freetype').disabled()
 
   if not freetype_dep.found() and cpp.get_id() == 'msvc'
     freetype_dep = cpp.find_library('freetype', required: false,
-                                    has_headers: ['ft2build.h'], )
+                                    has_headers: ['ft2build.h'])
   endif
 
   if not freetype_dep.found()
-    freetype_dep = dependency('freetype2', fallback: ['freetype2', 'freetype_dep'],
-                              default_options: ['harfbuzz=disabled'],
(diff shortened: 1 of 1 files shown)


xclaesse:
I don't think this will work. Meson currently treat subproject cycle as fatal error even with `required: false`. That's arguably a bug in Meson, but atm it's like that.

As far as I understand why you're trying to support here is that case:
- The system has harfbuzz installed.
- The system does not have freetype2 installed.
- We build harfbuzz, a newer version than the one from system for example.
- Harfbuzz fallback to freetype2 subproject because it's not found on system.
- freetype2 finds harfbuzz from system and enable harfbuzz support.

Is that even a possible scenario given that, at least on Ubuntu, harfbuzz depends on feetype?

xclaesse:
> use dependency's has_headers to make pulled cairo has ft support 

`dependency()` does not have `has_headers` argument, only find_library does. What you probably need there is:
```
if cairo_dep.type_name() == 'internal'
  # Not sure here, is it the same dependency object?
  cairo_ft_dep = cairo_dep
  # maybe cairo subproject needs to provide a dependency object for it:
  cairo_ft_dep = subproject('cairo').get_variable('libcairo_ft_dep')
elif cairo_dep.type_name() == 'library' and \
      cpp.has_function('cairo_ft_font_face_create_for_ft_face', prefix: '#include <cairo-ft.h>', dependencies: cairo_dep)
  cairo_ft_dep = cairo_dep
else
  cairo_ft_dep = dependency('cairo-ft', required: false)
endif
```

ebraminio:
> Is that even a possible scenario given that, at least on Ubuntu, harfbuzz depends on freetype?

My understanding is if I declare a default_options meson favors subproject mechanism over looking into those fallbacks? I just wanted those fallbacks, global variable in meson configuration hierarchy be tested one last time, 

> dependency() does not have has_headers argument

Oh, you are right, uploaded #2523 based on your code

> # Not sure here, is it the same dependency object?

If I understand correctly from https://github.com/ebraminio/cairo/blob/d48569e/src/meson.build#L298 it is. All we need is to have cairo as a subproject, if you know a better port let us know.

> cairo_ft_dep = subproject('cairo').get_variable('libcairo_ft_dep')

That port doesn't feature that IIUC

xclaesse:
> > Is that even a possible scenario given that, at least on Ubuntu, harfbuzz depends on freetype?
> 
> My understanding is if I declare a default_options meson favors subproject mechanism over looking into those fallbacks? I just wanted those fallbacks (global variable in meson configuration hierarchy I guess?) be tested one last time.

I'm not sure to understand what you mean here. `default_options` is only used by Meson when configuring the fallback when the dependency cannot be found on the system. It tells freetype2 subproject to not lookup for harfbuzz dependency because we are currently building harfbuzz it would create a dependency cycle.

ebraminio:
> I'm not sure to understand what you mean here. default_options is only used by Meson when configuring the fallback when the dependency cannot be found on the system. It tells freetype2 subproject to not lookup for harfbuzz dependency because we are currently building harfbuzz it would create a dependency cycle.

Okay, so :) Just uploaded https://github.com/harfbuzz/harfbuzz/pull/2527

ebraminio:
> I'm not sure to understand what you mean here. default_options is only used by Meson when configuring the fallback when the dependency cannot be found on the system. It tells freetype2 subproject to not lookup for harfbuzz dependency because we are currently building harfbuzz it would create a dependency cycle.

Just asking, can we first go for a `dependency('', fallback: ['freetype2', 'freetype_dep'], required: false)`? Those fallbacks are for things that are already defined in parent project, isn't it?

ebraminio:
> Just asking, can we first go for a dependency('', fallback: ['freetype2', 'freetype_dep'], required: false)? Those fallbacks are for things that are already defined in parent project, isn't it?

Ooh, I see, nvm, my misunderstanding, thank you very much :)

For more info, from https://mesonbuild.com/Dependencies.html

>>     foo_dep = dependency('foo', fallback : ['foo', 'foo_dep'])
> What this declaration means is that first Meson tries to look up the dependency from the system (such as by using pkg-config). If it is not available, then it builds subproject named foo and from that extracts a variable foo_dep. That means that the return value of this function is either an external or an internal dependency object. Since they can be used interchangeably, the rest of the build definitions do not need to care which one it is. Meson will take care of all the work behind the scenes to make this work.

PR #2521, opened 2020-07-02
[tests] Use shared --batch process

ebraminio:
Reverts da95a8c and use another approach for no-ft compiles

Before the change,
`meson test -Cbuild --no-suite slow  31.22s user 15.25s system 569% cpu 8.157 total`

After the change,
`meson test -Cbuild --no-suite slow  25.07s user 7.21s system 503% cpu 6.406 total`

With 1.5s difference however I was thinking also dropping whole --batch support to avoid the need for #2313 or maybe not, we can discuss that also.


behdad:
I don't like that you keep missing things. The `--batch|, just good it even if we don't use (yet).

PR #2522, opened 2020-07-02
meson: Make Requires in .pc files match autotools

heftig:
Libreoffice tries to use only `harfbuzz-icu.pc` and assumes this includes `-lharfbuzz`.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -501,6 +501,7 @@ pkgmod.generate(libharfbuzz,
 
 pkgmod.generate(libharfbuzz_subset,
   description: 'HarfBuzz font subsetter',
+  requires: ['harfbuzz'],
   subdirs: [meson.project_name()],
   version: meson.project_version(),
 )
@@ -531,6 +532,7 @@ if have_icu and not have_icu_builtin
 
   pkgmod.generate(libharfbuzz_icu,
     description: 'HarfBuzz text shaping library ICU integration',
+    requires: ['harfbuzz'],
     subdirs: [meson.project_name()],
     version: meson.project_version(),
   )
@@ -644,6 +646,7 @@ if have_gobject
 
   pkgmod.generate(libharfbuzz_gobject,
     description: 'HarfBuzz text shaping library GObject integration',
+    requires: ['harfbuzz', 'glib-2.0', 'gobject-2.0'],
     subdirs: [meson.project_name()],
     version: meson.project_version(),
   )
--- a/src/meson.build
+++ b/src/meson.build
@@ -501,7 +501,7 @@ pkgmod.generate(libharfbuzz,
 
 pkgmod.generate(libharfbuzz_subset,
   description: 'HarfBuzz font subsetter',
-  requires: ['harfbuzz'],
+  requires: ['harfbuzz = @0@'.format(meson.project_version())],
   subdirs: [meson.project_name()],
   version: meson.project_version(),
 )
@@ -532,7 +532,7 @@ if have_icu and not have_icu_builtin
 
   pkgmod.generate(libharfbuzz_icu,
     description: 'HarfBuzz text shaping library ICU integration',
-    requires: ['harfbuzz'],
+    requires: ['harfbuzz = @0@'.format(meson.project_version())],
     subdirs: [meson.project_name()],
     version: meson.project_version(),
   )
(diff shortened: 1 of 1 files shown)


behdad:
Would be even better if you can fix the exact version of harfbuzz in the dependency. Thanks.

heftig:
Hmm, seems this breaks the MSVC build where the GLib dependencies come from fallback to a subproject. Not sure what to do. @nirbheek?

behdad:
Thank you.

nirbheek:
> Hmm, seems this breaks the MSVC build where the GLib dependencies come from fallback to a subproject. Not sure what to do. @nirbheek?

Instead of using strings, can you try using the library / dependency objects directly? See: https://mesonbuild.com/Pkgconfig-module.html#implicit-dependencies

Also the autotools .pc file does not use `harfbuzz = <version>`, so probably best not to change that.

nirbheek:
Hmm, I get that even with objects.

> src/meson.build:645:9: ERROR: requires argument not a string, library with pkgconfig-generated file or pkgconfig-dependency object, got <InternalDependency null: True>

CC: @xclaesse seems it's not possible to set the `requires:` kwarg in `pkgconfig.generate()` to an internal dependency. Probably because we don't associate internal dependencies with the library's corresponding `pkgconfig.generate()` call.

heftig:
> Instead of using strings, can you try using the library / dependency objects directly?

@nirbheek Actually, that's exactly what I originally used:

```
requires: [libharfbuzz, glib_dep, gobject_dep],
```

Using strings seems to fix the issue I asked about.

---

> Also the autotools .pc file does not use harfbuzz = <version>, so probably best not to change that.

Yes, the version bound was another change I added on Behdad's request.

xclaesse:
> Hmm, I get that even with objects.
> 
> > src/meson.build:645:9: ERROR: requires argument not a string, library with pkgconfig-generated file or pkgconfig-dependency object, got

Are you sure to test with harfbuzz master? That's the error I fixed ;-)

> CC: @xclaesse seems it's not possible to set the `requires:` kwarg in `pkgconfig.generate()` to an internal dependency. Probably because we don't associate internal dependencies with the library's corresponding `pkgconfig.generate()` call.

Meson currently don't accept internal dependencies in requires, I do have vague ideas of merging pkg.generate() and declare_dependency() somehow.

behdad:
> Also the autotools .pc file does not use `harfbuzz = <version>`, so probably best not to change that.

That was an omission on my part in writing the autotools rules. I asked that to be fixed in meson.

behdad:
> That was an omission on my part in writing the autotools rules. I asked that to be fixed in meson.

The reason is that while we try to write the other libraries that HarfBuzz produces to only use HarfBuzz public C API, it's limiting and doesn't work for all. For example the harfbuzz-subset library accesses `hb_face_t` internally so will break if it's linked with a different version of libharfbuzz. harfbuzz-gobject and harfbuzz-icu don't do that, but again, there's no good reason to want to link them to a different version of HarfBuzz. That if ever happens only points to a problem somewhere else. So I like enforcing this at the .pc level.

ebraminio:
> Hmm, seems this breaks the MSVC build where the GLib dependencies come from fallback to a subproject. 

All of our msvc bots are fine with the change https://ci.appveyor.com/project/harfbuzz/harfbuzz/builds/33876219 so guess it is fine to have the change in?

src/meson.build
```diff
@@ -532,7 +532,7 @@ if have_icu and not have_icu_builtin
 
   pkgmod.generate(libharfbuzz_icu,
     description: 'HarfBuzz text shaping library ICU integration',
-    requires: ['harfbuzz'],
+    requires: ['harfbuzz = @0@'.format(meson.project_version())],
```
ebraminio, line 535: Just for educational reasons, is there any reason that `icuuc` isn't here but `glib-2.0` is put on gobject, icu's versioning and so?

heftig:
Yes, because these builds are from *after* I fixed it.
Here's the broken build: https://ci.appveyor.com/project/harfbuzz/harfbuzz/builds/33875456
And the commit it used: https://github.com/heftig/harfbuzz/commit/0b5b0f4bc6f6c5c8d64cfd04d8d55cefa5509ddc

ebraminio:
> Yes, because these builds are from after I fixed it.

Oh, I see, so is it ok to merge now you think?

src/meson.build
heftig: I'm not sure. The template files for autotools have `icuuc` only as a private dependency. The generator already does this so there was no need to list it.

https://github.com/harfbuzz/harfbuzz/blob/22ce0961742df0a213f2cc86ca3d5a72453c00b1/src/harfbuzz-icu.pc.in#L10-L11

heftig:
> Oh, I see, so is it ok to merge now you think?

Yes.

src/meson.build
ebraminio: Ok so, thanks :)

ebraminio:
Thank you very much :)

PR #2532, opened 2020-07-03
[meson] use the recommended way to store generated files

### LINKED ISSUE #2530: [meson] use a more proper way for source alternating tasks
Brought up here https://github.com/harfbuzz/harfbuzz/commit/1853f84abc64ed9dee6ef913760b0295c36a4ef5#commitcomment-40349984

Having source altering tasks is easy in autotools https://github.com/harfbuzz/harfbuzz/blob/76ba619/src/Makefile.am#L274 but isn't the same way in meson for understandable reasons. Currently things work more or less (I use meson for development for a long time) but it isn't ideal and we should fix then remove the hacks mentioned above done done in favor of meson (which is done so touching timestamps not ruin meson's incremental build) 

What is needed here as described in https://github.com/mesonbuild/meson/issues/7156#issuecomment-629622827 is to make `src/gen-hb-version.py`, `src/gen-ragel-artifacts.py` and `src/gen-harfbuzzcc.py` scripts to write their artifacts **both** in build dir and source dir, letting meson to follow the one in build dir (not sure if that can go smoothly), then remove `build_by_default: true,` and hacks above afterward as it won't run the scripts each time.

ebraminio:
As described in https://github.com/mesonbuild/meson/issues/7156#issuecomment-629622827

Fixes #2530

### DIFF
--- a/src/check-c-linkage-decls.py
+++ b/src/check-c-linkage-decls.py
@@ -4,9 +4,9 @@ import sys, os
 
 os.chdir (os.environ.get ('srcdir', os.path.dirname (__file__)))
 
-HBHEADERS = os.environ.get ('HBHEADERS', '').split () or \
+HBHEADERS = [os.path.basename (x) for x in os.environ.get ('HBHEADERS', '').split ()] or \
 	[x for x in os.listdir ('.') if x.startswith ('hb') and x.endswith ('.h')]
-HBSOURCES = os.environ.get ('HBSOURCES', '').split () or \
+HBSOURCES = [os.path.basename (x) for x in os.environ.get ('HBSOURCES', '').split ()] or \
 	[x for x in os.listdir ('.') if x.startswith ('hb') and x.endswith (('.cc', '.hh'))]
 
 stat = 0
--- a/src/meson.build
+++ b/src/meson.build
@@ -301,9 +301,7 @@ else
       input: rl,
       output: hh,
       command: [find_program('gen-ragel-artifacts.py'),
-                '@OUTPUT@',
-                '@CURRENT_SOURCE_DIR@',
-                '@INPUT@'],
+                '@OUTPUT@', '@CURRENT_SOURCE_DIR@', '@INPUT@'],
     )
   endforeach
 endif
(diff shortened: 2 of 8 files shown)


src/meson.build
```diff
@@ -667,8 +674,16 @@ if get_option('tests').enabled()
   env.set('srcdir', meson.current_source_dir())
   env.set('builddir', meson.current_build_dir())
   env.set('libs', meson.current_build_dir()) # TODO: Merge this with builddir after autotools removal
-  env.set('HBSOURCES', ' '.join(hb_sources))
-  env.set('HBHEADERS', ' '.join(hb_headers))
+  HBSOURCES = []
+  foreach f : hb_sources
+    HBSOURCES += '@0@'.format(f).split ('/')[1]
```
ebraminio, line 679: yeah, this won't work.

ebraminio: currently out of idea, guess should `os.path.basename` in check scripts.

src/check-c-linkage-decls.py
```diff
@@ -4,9 +4,9 @@
 
 os.chdir (os.environ.get ('srcdir', os.path.dirname (__file__)))
 
-HBHEADERS = os.environ.get ('HBHEADERS', '').split () or \
+HBHEADERS = [os.path.basename (x) for x in os.environ.get ('HBHEADERS', '').split ()] or \
```
ebraminio, line 7: Why these are needed? File objects in meson (created using by `files()` in this patch) when turned to string are like `src/hb-x.hh` instead just the base name they used to, alternative approach would be top drop the above `chdir` yet I like to keep the script still runnable from command line that's why I am doing this for now.

src/meson.build
```diff
@@ -667,8 +675,16 @@ if get_option('tests').enabled()
   env.set('srcdir', meson.current_source_dir())
   env.set('builddir', meson.current_build_dir())
   env.set('libs', meson.current_build_dir()) # TODO: Merge this with builddir after autotools removal
-  env.set('HBSOURCES', ' '.join(hb_sources))
-  env.set('HBHEADERS', ' '.join(hb_headers))
+  HBSOURCES = []
+  foreach f : hb_sources
+    HBSOURCES += '@0@'.format(f)
+  endforeach
+  env.set('HBSOURCES', ' '.join(HBSOURCES))
```
ebraminio, line 682: The for each above is to generate string out of files objects array

PR #2534, opened 2020-07-04
[py] minor

ebraminio:


src/check-c-linkage-decls.py
```diff
@@ -23,4 +23,4 @@
 		print ('Ouch, file %s has HB_BEGIN_DECLS / HB_END_DECLS, but it shouldn\'t' % x)
 		stat = 1
 
-sys.exit (stat)
+exit (stat)
```
khaledhosny, line 26: https://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python

ebraminio: oh, so let's normalize in the reverse way

ebraminio:
@khaledhosny, is the rest okay? Thank you

ebraminio:
Ok merged as it seems to work, please let me know any other thing with the changes and will try my best to fix them. Thank you so much :-)

PR #2538, opened 2020-07-04
Minor, disable strict-aliasing warning in set fuzzer

ebraminio:
../test/fuzzing/hb-set-fuzzer.cc: In function ‘int LLVMFuzzerTestOneInput(const uint8_t*, size_t)’:
../test/fuzzing/hb-set-fuzzer.cc:38:82: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
   38 |   const instructions_t &instructions = reinterpret_cast<const instructions_t &> (data);
      |

And it is already disabled at project level so let's disable it here also.


behdad:
There's also alignment issue. We assume fuzzer calls us with four-byte aligned pointer. Def true but now enforced in code

Issue #2531, opened 2020-07-03
Broken rendering of AAT font Devanagari Sangam MN


jfkthame:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1650414.
The Devanagari Sangam MN font mis-renders on macOS 10.15 (at least).

Testcase:
```
$ hb-shape --font-file Devanagari\ Sangam\ MN.ttc --unicodes U+092D,U+0942 --shaper coretext
[dn_bha=0+1043|dn_uu_matra.mrk=0@0,11+296]

$ hb-shape --font-file Devanagari\ Sangam\ MN.ttc --unicodes U+092D,U+0942 --shaper ot
[dn_bha=0+1339|dn_uu_matra.mrk=0@-296,1409+0]
```
Note the large y-offset applied to the dn_uu_matra.mrk with shaper=ot.

jfkthame:
I'm beginning to suspect this may be the result of a discrepancy between [the spec for the `kerx` table](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6kerx.html) and what is actually implemented in the Devanagari Sangam MN font and in Core Text.

Specifically, regarding the Format 4 Kerning Subtable, action type 1, the spec says of Anchor Point Actions:

Type | Name | Description
-- | -- | --
uint16 | markAnchorPoint | Anchor point in marked glyph.
uint16 | currAnchorPoint | Anchor point in current glyph.

but this does not appear to match the usage in Devanagari Sangam MN. Tracing execution in harfbuzz for the sequence <U+092D,U+0942>, I'm seeing it retrieving markAnchorPoint=0 for the "marked" glyph `dn_bha`, and currAnchorPoint=1 for the "current" glyph `dn_uu_matra.mrk`; but that makes no sense, as `dn_uu_matra.mrk` only has a single anchor (0) defined in the `ankr` table, while `dn_bha` does have two anchors, and its anchor (1) is the appropriate one for a "below" mark.

So if I reverse the harfbuzz code such that it reads the *first* uint16 here as `currAnchorPoint` and the *second* uint16 as `markAnchorPoint`, that fixes the rendering of this example.

[...]

PR #2539, opened 2020-07-04
[aat] Correct array indexing when looking up actions in KerxSubTableFormat4

jfkthame:
Fixes https://github.com/harfbuzz/harfbuzz/issues/2531.

### DIFF
--- a/src/hb-aat-layout-kerx-table.hh
+++ b/src/hb-aat-layout-kerx-table.hh
@@ -508,8 +508,10 @@ struct KerxSubTableFormat4
 	{
 	  case 0: /* Control Point Actions.*/
 	  {
-	    /* indexed into glyph outline. */
-	    const HBUINT16 *data = &ankrData[entry.data.ankrActionIndex];
+	    /* Indexed into glyph outline. */
+	    /* Each action (record in ankrData) contains two 16-bit fields, so we must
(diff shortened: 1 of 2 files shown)


behdad:
Thanks!

How about changing `ankr_data` to `hb_pair_t<HBUINT16, HBUINT16>? Then instead of `check_array` just do a `check_struct`.

The two `HB_UNUSED` near the first block your are touching can also be removed.

behdad:
LGTM otherwise.

jfkthame:
> Thanks!
> 
> How about changing `ankr_data` to `hb_pair_t<HBUINT16, HBUINT16>? Then instead of `check_array`just do a`check_struct`.

I wondered about changing the type there, but wasn't sure it was worth it for something just used in the one place. Changing the type of the KerxSubTableFormat4::ankrData field isn't entirely satisfactory, because it's not always an array of pairs of uint16 ... in the case where action_type is 3, there are 4 values per action entry.

Which, now that I look at it, means that code fragment is also wrong; it needs a `* 4` inserted. I guess we don't have any test coverage here... :(

behdad:
Thanks. I'll touch up later.

behdad:
I'm on phone. Feel free to merge.

PR #2542, opened 2020-07-05
[aat] Remove lcar support

ebraminio:
Very low use, only two distinct font files, Apple Chancery.ttf and Hoefler Text.ttc
have it so it really doesn't worth the size addition and so, but one may argue that
whole ligature caret is low use but guess we better to encourage GDEF one anyway.

Somehow changed my mind after https://github.com/harfbuzz/harfbuzz/issues/673#issuecomment-646472412 and had some agreement with Behdad before uploading this yet feel free to give your feedback, we can even revert this anytime in future.

https://gist.github.com/ebraminio/d52dd780fec59f8e2dfffe8e1a841823


behdad:
Thanks. LigCaret we want to promote into actually being used.

PR #2545, opened 2020-07-05
Streamline release process

### LINKED ISSUE #2536: Streamline release process
Now that we are migrating to meson and modernizing things, and now that meson's tarball isn't that different from git checkouts (except the recently added .tarball-revision, which we can remove that also), something that I think can increase reliability and trust to releases is to encourage use of GitHub's tarballs instead relying to upload of locally generated tarballs. Currently multiple people have access to release tarball page but GitHub apparently doesn't provide log of who has changed tags' artifacts so packagers need to make a trust at the version they are compiling but with relying to GitHub tarball packagers can be exact sure that which changes has lead to the release and there isn't any suspicious change, and there are other issues that this can help as maybe sometimes tarball upload get interrupted by accident and an incomplete tarball maybe will be put there which this reduces such risks.

Even more reliable link would be https://github.com/harfbuzz/harfbuzz/archive/e4203c14699cc0903a2b4611167a7b99532cfdb0.tar.gz out of e4203c14699cc0903a2b4611167a7b99532cfdb0 that exactly refers to an unchangeable git checkout as what GitHub suggests at below is https://github.com/harfbuzz/harfbuzz/archive/2.6.8.tar.gz 

image

ebraminio:
Fixes #2536

Let's discuss if we like the changes happened to release process on this change.


behdad:
LGTM.

PR #2535, opened 2020-07-04
[tests] rename no_ft_funcs to NO_FT_FUNCS

ebraminio:
As https://github.com/harfbuzz/harfbuzz/commit/43e3ab019c6b6ab8e429b3dd3aa13a7b9e8316e6#r40363339

@khaledhosny would you please have a look? Thanks :)

PR #2546, opened 2020-07-05
[tests] Use HAVE_FREETYPE itself instead NO_FT_FUNCS

ebraminio:
https://github.com/harfbuzz/harfbuzz/pull/2535#issuecomment-653933245

### DIFF
--- a/test/shaping/run-tests.py
+++ b/test/shaping/run-tests.py
@@ -15,7 +15,7 @@ if len (args) and args[0] == "--reference":
 	reference = True
 	args = args[1:]
 
-have_freetype = bool(int(os.getenv ('HAVE_FREETYPE', '0')))
+have_freetype = bool(int(os.getenv ('HAVE_FREETYPE', '1')))
 
 if not args or args[0].find('hb-shape') == -1 or not os.path.exists (args[0]):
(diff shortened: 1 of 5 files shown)


behdad:
Thanks. In the Makefile. Am add:
```
HAVE_FREETYPE = $(HAVE_FREETYPE)
```
Then just add `HAVE_FREETYPE=$(HAVE_FREETYPE)` to the test environment.

Something like that. Or merge as is and I'll fix after.

test/shaping/run-tests.py
```diff
@@ -15,7 +15,7 @@ def cmd(command):
 	reference = True
 	args = args[1:]
 
-no_ft_funcs = bool(int(os.getenv ('NO_FT_FUNCS', '0')))
+have_freetype = bool(int(os.getenv ('HAVE_FREETYPE', '0')))
```
behdad, line 18: Don't we want it to default to 1?

behdad:
Regarding my earlier comment about the autotools part, ignore it. Good enough as is.

test/shaping/run-tests.py
ebraminio: done

PR #2548, opened 2020-07-06
[meson] Set the default wrap mode to nofallback

khaledhosny:
This, I believe, will prevent any automatic use of subprojects be default and will only look for dependencies on the system.
https://mesonbuild.com/Subprojects.html#commandline-options

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -41,7 +41,7 @@ build_script:
   - set "PATH=%PYTHON_ROOT%;%PYTHON_ROOT%\Scripts;%PATH%"
   - pip install --upgrade meson
   - pip install fonttools
-  - 'if "%compiler%"=="msvc" "%vcvarsallpath%" %arch% && meson setup build --buildtype=release -Dglib=enabled -Dfreetype=enabled -Dgdi=enabled -Ddirectwrite=enabled && meson test --print-errorlogs --suite=harfbuzz -Cbuild'
+  - 'if "%compiler%"=="msvc" "%vcvarsallpath%" %arch% && meson setup build --wrap-mode=default --buildtype=release -Dglib=enabled -Dfreetype=enabled -Dgdi=enabled -Ddirectwrite=enabled && meson test --print-errorlogs --suite=harfbuzz -Cbuild'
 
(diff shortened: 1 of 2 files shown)


ebraminio:
wished it had a mode to revert to its default mode, https://mesonbuild.com/Subprojects.html#commandline-options guess we have to add `forcefallback` to msvc bots but that isn't equivalent to its default mode I guess which could use vcpkg for example? And needs gbenchmark be enabled again at https://github.com/harfbuzz/harfbuzz/blob/5f870dfadb6ce3f85e5289131c4291cda31706c6/.circleci/config.yml#L231 also so the removed `-Dbenchmark=true` will be reintroduced somehow as `--force-fallback-for=google-benchmark`

khaledhosny:
> wished it had a mode to revert to its default mode, https://mesonbuild.com/Subprojects.html#commandline-options guess we have to add `forcefallback` to msvc bots but that isn't equivalent to its default mode I guess which could use vcpkg for example?

Looks like `--wrap-mode=default` will use meson’s default so should be enough here.

> And needs gbenchmark be enabled again at [this bot](https://github.com/harfbuzz/harfbuzz/blob/5f870dfadb6ce3f85e5289131c4291cda31706c6/.circleci/config.yml#L231) also so the removed `-Dbenchmark=true` will be reintroduced somehow as `--force-fallback-for=google-benchmark`.

Explicite `subproject()` uses should not not affected by this change.

ebraminio:
Makes sense! Thank you very much :)

xclaesse:
--wrap-mode=nofallback is usually what distro should be doing. The default is easier for devs who don't need to care about dependencies, especially when building on older distro that have too old versions of dependencies.

khaledhosny:
We intentionally want the fallback to be opt-in not opt-out.

behdad:
Yeah let's keep it similar to autotools initially. If any deps were not found we can print how to enable download. In the future we might change the default.

khaledhosny:
Currently meson prints this:
```
Run-time dependency fontconfig found: NO (tried pkgconfig, framework and cmake)
Not looking for a fallback subproject for the dependency fontconfig because:
Use of fallback dependencies is disabled.

meson.build:97:0: ERROR: Dependency 'fontconfig' not found and fallback is disabled
```

It gives some clue, but unless one already knows about `--wrap-mode` option, it will sound cryptic. The message can be improved by mentioning the option.

PR #2566, opened 2020-07-14
[fuzz] Implement failing allocator

ebraminio:
If ever happens… going to be fun!

How to test, 

`CXXFLAGS="-fsanitize=address,fuzzer-no-link" meson build --default-library=static -Dfuzzer_ldflags="-fsanitize=address,fuzzer" && ninja -Cbuild test/fuzzing/hb-{shape,draw,subset,set}-fuzzer` (or `ninja -Cbuild`)

Then run

`build/test/fuzzing/hb-{shape,set,draw,subset}-fuzzer` 

(make sure llvm package thus llvm-symbolizer command is installed)

### DIFF
--- a/src/failing-alloc.c
+++ b/src/failing-alloc.c
@@ -1,0 +1,57 @@
+/*
+ * Copyright © 2020  Ebrahim Byagowi
+ *
+ *  This is part of HarfBuzz, a text shaping library.
+ *
+ * Permission is hereby granted, without written agreement and without
+ * license or royalty fees, to use, copy, modify, and distribute this
+ * software and its documentation for any purpose, provided that the
+ * above copyright notice and the following two paragraphs appear in
+ * all copies of this software.
+ *
+ * IN NO EVENT SHALL THE COPYRIGHT HOLDER BE LIABLE TO ANY PARTY FOR
+ * DIRECT, INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES
+ * ARISING OUT OF THE USE OF THIS SOFTWARE AND ITS DOCUMENTATION, EVEN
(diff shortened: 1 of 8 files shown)


behdad:
Thanks!

I'll play with it sometime. I wanted to implement the fork&fail. But this is a great start!

Let's have one new macro, `HB_CUSTOM_MALLOC` that when defined, we use the `impl` ones, so you don't have to define three macros to expand to their own name.

ebraminio:
> Let's have one new macro, HB_CUSTOM_MALLOC that when defined, we use the impl ones, so you don't have to define three macros to expand to their own name.

Applied on master https://github.com/harfbuzz/harfbuzz/commit/7c433ffde2d566d849cd54a881abddad4d7e2f51 and reduced the noises here.

src/failing-alloc.c
```diff
@@ -0,0 +1,57 @@
...
+
+#include <stdlib.h>
+#include <stdio.h>
+
+int alloc_state = 0;
+
+__attribute__((no_sanitize("integer")))
+static int fastrand ()
+{
+  if (!alloc_state) return 1;
```
ebraminio, line 33: when there is no state, return 1 which means allocs will work correctly.

src/failing-alloc.c
```diff
@@ -0,0 +1,57 @@
...
+  /* Based on https://software.intel.com/content/www/us/en/develop/articles/fast-random-number-generator-on-the-intel-pentiumr-4-processor.html */
+  alloc_state = (214013 * alloc_state + 2531011);
+  return (alloc_state >> 16) & 0x7FFF;
+}
+
+void* hb_malloc_impl (size_t size)
+{
+  return (fastrand () % 16) ? malloc (size) : NULL;
```
ebraminio, line 41: so the failing rate is 1/16

ebraminio:
Ok, have merged this, we can address shape issues #2571 later, the coming days is going to be fun!

PR #2569, opened 2020-07-15
CI: Port MSYS2 jobs from appveyor to github actions

lazka:
This is motivated by our new github action https://github.com/msys2/setup-msys2 which makes spawning a MSYS2 env easy and hides the install/update details..

(and also github actions has 20 parallel jobs.. compared to 1 in appveyor)

### DIFF
--- a/.github/workflows/msys2-ci.yml
+++ b/.github/workflows/msys2-ci.yml
@@ -1,0 +1,66 @@
+name: msys2
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  msys2:
+    runs-on: windows-latest
+
+    strategy:
+      matrix:
+        include:
+          - MSYSTEM: MINGW32
+            MSYS2_ARCH: i686
+          - MSYSTEM: MINGW64
+            MSYS2_ARCH: x86_64
+    name: ${{ matrix.MSYSTEM }}
+
+    defaults:
+      run:
+        shell: msys2 {0}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: msys2/setup-msys2@v1
+        with:
+          msystem: ${{ matrix.MSYSTEM }}
+          update: true
+          install: >-
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-cairo
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-freetype
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-gcc
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-gcc-libs
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-gettext
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-glib2
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-gobject-introspection
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-graphite2
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-icu
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-ninja
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-pkg-config
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-python3
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-python3-pip
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-ragel
+      - name: Install extra dependencies
+        run: |
+          pip install fonttools
+          pip install --upgrade meson==0.53.0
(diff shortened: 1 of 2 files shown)


.github/workflows/msys2-ci.yml
```diff
@@ -0,0 +1,66 @@
...
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-ninja
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-pkg-config
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-python3
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-python3-pip
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-ragel
+      - name: Install extra dependencies
+        run: |
+          pip install fonttools
+          pip install --upgrade meson==0.53.0
+          curl "https://raw.githubusercontent.com/mirror/mingw-w64/023eb04c396d4e8d8fcf604cfababc53dae13398/mingw-w64-headers/include/dwrite_1.h" > "$MINGW_PREFIX/$MINGW_CHOST/include/dwrite_1.h"
```
ebraminio, line 48: This probably can be dropped as your are building hb-directwrite in msys2 successfully now.

ebraminio:
Looks great! Thank you so much!

.github/workflows/msys2-ci.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          meson build \
+            --wrap-mode=nofallback \
+            -Dfreetype=enabled \
+            -Dglib=enabled \
+            -Dcairo=enabled \
+            -Dgobject=enabled \
+            -Dgdi=enabled \
+            -Ddirectwrite=enabled \
+            -Dgraphite=enabled
```
ebraminio, line 59: Guess now can be reduced to `--auto-features=enabled -Dgraphite=enabled`

.github/workflows/msys2-ci.yml
```diff
@@ -0,0 +1,66 @@
...
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-ragel
+      - name: Install extra dependencies
+        run: |
+          pip install fonttools
+          pip install --upgrade meson==0.53.0
+          curl "https://raw.githubusercontent.com/mirror/mingw-w64/023eb04c396d4e8d8fcf604cfababc53dae13398/mingw-w64-headers/include/dwrite_1.h" > "$MINGW_PREFIX/$MINGW_CHOST/include/dwrite_1.h"
+      - name: Build
+        run: |
+          meson build \
+            --wrap-mode=nofallback \
```
ebraminio, line 52: Guess `--wrap-mode=nodownload` will be even better.

.github/workflows/msys2-ci.yml
```diff
@@ -0,0 +1,62 @@
...
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-icu
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-ninja
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-pkg-config
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-python3
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-python3-pip
+            mingw-w64-${{ matrix.MSYS2_ARCH }}-ragel
+      - name: Install extra dependencies
+        run: |
+          pip install fonttools
+          pip install --upgrade meson==0.53.0
```
ebraminio, line 47: We are okay with mingw's meson and even fonttools, I didn't do that as for some reason I weren't getting an updated version with previous, the shiny new setup shouldn't have the issues I'd guess.

ebraminio:
Just to note your CI config is already cleaner than most of our current CI configurations, thank you so much for putting time for this :)

.github/workflows/msys2-ci.yml
lazka: ok, done

.github/workflows/msys2-ci.yml
lazka: ok, switched to the msys2 packages

.github/workflows/msys2-ci.yml
lazka: done, I had to disable benchmark though since that failed.

lazka:
Added the suggestions in a second commit, to separate the porting from the cleanup part.

Should be ready.

ebraminio:
Fantastic. Thank you so much :)

.github/workflows/msys2-ci.yml
ebraminio: have applied some workaround and removed it, just am notifying so you don't need to add it to msys2 pack also :)

PR #2578, opened 2020-07-20
[ci] Update to msys2/setup-msys2@v2

lazka:
This update introduces better error handling and package caching by default.


ebraminio:
Looks great! Thank you very much! :)

PR #2579, opened 2020-07-20
[tests] fix PGO build testing

insilications:
Adds 'mangle_path' to IGNORED_SYMBOLS in order to fix testing under PGO builds


ebraminio:
Thank you very much :)

Issue #2580, opened 2020-07-22
[subset] GPOS lookupType 9 not properly subset.


bungeman:
With HarfBuzz 83c44f69d3724acbb16b25c1c1dee564ca1c3604 with [Lato-Medium.ttf](https://github.com/latofonts/lato-source/blob/353ba57fb6a82049e1dc116d8180ebfbbefc1474/fonts/static/TTF/Lato-Medium.ttf).

```
~/src/harfbuzz$ ./autogen.sh --prefix=$HOME/src/harfbuzz/usr
Build configuration:

Unicode callbacks (you want at least one):
        Builtin                 true
        Glib:                   true
        ICU:                    true

Font callbacks (the more the merrier):
        FreeType:               true

Tools used for command-line utilities:
        Cairo:                  true
        Fontconfig:             true

Additional shapers:
        Graphite2:              false

Platform shapers (not normally needed):
        CoreText:               false
        DirectWrite:            false
        GDI:                    false
        Uniscribe:              false

Other features:
        Documentation:          no
        GObject bindings:       false
        Introspection:          false

~/src/harfbuzz$ make -j32 install
~/src/harfbuzz$ usr/bin/hb-subset --drop-tables-="GSUB GPOS GDEF" -o Lato-Liga.ttf Lato-Medium.ttf "fistﬆﬁ "
```
This runs fine but produces a bad GPOS table. This was initially detected with ttx running like
```
~/src/harfbuzz$ ttx Lato-Liga.ttf
Dumping "src/harfbuzz/Lato-Liga.ttf" to "src/harfbuzz/Lato-Liga.ttx"...
Dumping 'GlyphOrder' table...
Dumping 'head' table...
Dumping 'hhea' table...
Dumping 'maxp' table...
Dumping 'OS/2' table...
Dumping 'hmtx' table...
Dumping 'cmap' table...
Dumping 'fpgm' table...
Dumping 'prep' table...
Dumping 'cvt ' table...
Dumping 'loca' table...
Dumping 'glyf' table...
Dumping 'name' table...
Dumping 'post' table...
Dumping 'gasp' table...
Dumping 'GDEF' table...
ERROR: An exception occurred during the decompilation of the 'GPOS' table
Traceback (most recent call last):
  File "/home/bungeman/src/fonttools/Lib/fontTools/ttLib/ttFont.py", line 372, in __getitem__
    return self.tables[tag]
KeyError: 'GPOS'

[...]

PR #2582, opened 2020-07-22
[subset] Basic support for LookupType 9.

bungeman:
Currently if subseting a LookupType 9 Extension Positioning table the
data written is that of the extension table, not the extension
positioning table which has an offset to the extension table. This
change makes that work in the most basic way possible. It makes no
attempt to place the extesion table later in the output, so may cause
issues as offsets grow large, but no more so than in the existing code.

In a better implementation all LookupType 9 subtables would be placed as
early as possible with the extension subtables referenced placed as
late as possible. An even better implementation would look at the
overall size of the subtables and output LookupType 9 subtables only
when needed.

### DIFF
--- a/src/hb-ot-layout-gsubgpos.hh
+++ b/src/hb-ot-layout-gsubgpos.hh
@@ -3096,6 +3096,24 @@ struct ExtensionFormat1
   void collect_variation_indices (hb_collect_variation_indices_context_t *c) const
   { dispatch (c); }
 
+  /* This is called from may_dispatch() above with hb_subset_context_t. */
+  bool subset (hb_subset_context_t *c) const
+  {
+    TRACE_SUBSET (this);
+
+    auto *out = c->serializer->start_embed (this);
+    if (unlikely (!out || !c->serializer->extend_min (out))) return_trace (false);
+    out->format = format;
+    out->extensionLookupType = extensionLookupType;
+
+    const auto& inOffset = reinterpret_cast<const OffsetTo<typename T::SubTable, HBUINT32>&> (this->extensionOffset);
+    auto& outOffset = reinterpret_cast<OffsetTo<typename T::SubTable, HBUINT32>&> (out->extensionOffset);
+    /*bool ret =*/ outOffset.serialize_subset (c, inOffset, this, get_type ());
+
+    // Cancel dispatch to next level.
+    return_trace (false);
+  }
+
   /* This is called from may_dispatch() above with hb_sanitize_context_t. */
   bool sanitize (hb_sanitize_context_t *c) const
   {
--- a/src/hb-subset.hh
+++ b/src/hb-subset.hh
@@ -54,6 +54,18 @@ struct hb_subset_context_t :
   dispatch (const T &obj, Ts&&... ds) HB_AUTO_RETURN
   ( _dispatch (obj, hb_prioritize, hb_forward<Ts> (ds)...) )
 
+  private:
+  template <typename T, typename F> auto
+  _may_dispatch (const T *obj, const F *format, hb_priority<1>) HB_AUTO_RETURN
+  ( format->subset (this) )
+  template <typename T, typename F> auto
+  _may_dispatch (const T *obj, const F *format, hb_priority<0>) HB_AUTO_RETURN
+  ( true )
+  public:
+  template <typename T, typename F> auto
+  may_dispatch (const T *obj, const F *format) HB_AUTO_RETURN
+  ( _may_dispatch (obj, format, hb_prioritize) )
+
   hb_blob_t *source_blob;
   hb_subset_plan_t *plan;
   hb_serialize_context_t *serializer;


bungeman:
An attempt at an interim fix for https://github.com/harfbuzz/harfbuzz/issues/2580. Provides basic support and shouldn't break any fonts which used to work.

bungeman:
> I feel we could still keep the extension table instead of dropping it.
> Also, could you help to add some tests?

I'm confused, this change outputs the extension positioning table and also the extension table (the one it points at). However, it does more or less defeat the purpose of the extension positioning table in that the extension table is always placed right after it in the stream instead of placing the extension tables toward the end of the stream.

I might be able to figure out how to add a test for this, though a real implementation would probably produce quite different output, so the test inputs would probably be interesting but the outputs wouldn't be interesting to a 'real' implementation.

qxliu76:
> > I feel we could still keep the extension table instead of dropping it.
> > Also, could you help to add some tests?
> 
> I'm confused, this change outputs the extension positioning table and also the extension table (the one it points at). However, it does more or less defeat the purpose of the extension positioning table in that the extension table is always placed right after it in the stream instead of placing the extension tables toward the end of the stream.
> 
> I might be able to figure out how to add a test for this, though a real implementation would probably produce quite different output, so the test inputs would probably be interesting but the outputs wouldn't be interesting to a 'real' implementation.

Oh I did not notice you overloaded may_dispatch(). I wonder is there a specific reason to overload it?  From my point of view, there's no need to do that.

bungeman:
> > > I feel we could still keep the extension table instead of dropping it.
> > > Also, could you help to add some tests?
> > 
> > 
> > I'm confused, this change outputs the extension positioning table and also the extension table (the one it points at). However, it does more or less defeat the purpose of the extension positioning table in that the extension table is always placed right after it in the stream instead of placing the extension tables toward the end of the stream.
> > I might be able to figure out how to add a test for this, though a real implementation would probably produce quite different output, so the test inputs would probably be interesting but the outputs wouldn't be interesting to a 'real' implementation.
> 
> Oh I did not notice you overloaded may_dispatch(). I wonder is there a specific reason to overload it? From my point of view, there's no need to do that.

For the same reason dispatch() is overloaded, not all types F have a 'subset' method. The alternative is to need to implement all other context calls on ExtensionFormat1 (which are fine to allow the dispatch mechanism to allow through, all the implementations would just be pass-through anyway) or re-invent may_dispatch in an ugly way inside ExtensionFormat1::dipatch itself.

bungeman:
> > > > I feel we could still keep the extension table instead of dropping it.
> > > > Also, could you help to add some tests?
> > > 
> > > 
> > > I'm confused, this change outputs the extension positioning table and also the extension table (the one it points at). However, it does more or less defeat the purpose of the extension positioning table in that the extension table is always placed right after it in the stream instead of placing the extension tables toward the end of the stream.
> > > I might be able to figure out how to add a test for this, though a real implementation would probably produce quite different output, so the test inputs would probably be interesting but the outputs wouldn't be interesting to a 'real' implementation.
> > 
> > 
> > Oh I did not notice you overloaded may_dispatch(). I wonder is there a specific reason to overload it? From my point of view, there's no need to do that.
> 
> For the same reason dispatch() is overloaded, not all types F have a 'subset' method. The alternative is to need to implement all other context calls on ExtensionFormat1 (which are fine to allow the dispatch mechanism to allow through, all the implementations would just be pass-through anyway) or re-invent may_dispatch in an ugly way inside ExtensionFormat1::dipatch itself.

To clarify by comparing with 'sanitize' which seems to be (before this change) the only user of the may_dispatch mechanism, it seems every type may_dispatch is used on has a 'sanitize' method (in fact, because 'sanitize was the only user, may_dispatch was only used on such types). The lack of overloading of hb_sanitize_context_t::may_dispatch actually enforces this. If 'subset' were done the same way hb_subset_context_t::may_dispatch could be non-overloaded but then all types would need a 'subset' method (everywhere there is a 'sanitize' there would also need to be a 'subset'). This makes a bit of sense in a strange way, OT::IntType is 'subset' by just writing itself out. However, here I'm just piggy-backing on the existing may_dispatch to self in ExtensionFormat1::dispatch, not adding any, so here there is no need to overload hb_sanitize_context_t::may_dispatch quite yet, but for the sake of being more generic it probably should be.

qxliu76:
thanks for your explanation. 
Traditionally subset() returning false means this table will be dropped. You forced it to return false, which makes may_dispatch() returns false as well, which is confusing. With may_dispatch()/subset() returning false, actually the table is still subsetted successfully. The logic is weird to me.

Also if my understanding is correct, for other tables, because of your overloaded may_dispatch(), subset () is called twice: first time in overloaded may_dispatch(), and they return true, so subset () is called a second time in dispatch().. 

We're currently working on supporting all tables,  so I'd prefer implementing subset() for all other tables before we start implementing subset for extension table. This way there's no need to overload may_dispatch().

bungeman:
> thanks for your explanation.
> Traditionally subset() returning false means this table will be dropped. You forced it to return false, which makes may_dispatch() returns false as well, which is confusing. With may_dispatch()/subset() returning false, actually the table is still subsetted successfully. The logic is weird to me.

Hmmm... yes, this is an awkward bit with how this works. Since I implemented `hb_sanitize_context_t::may_dispatch` here I should have also defined `hb_sanitize_context_t::no_dispatch_return_value` to return `false` (instead of the current default, which is `true`). Doing so would have prevented this from working. The only real way around this seems to be to make `ExtensionFormat1` more fleshed out (implement the required methods for `dispatch` to work on it) so that it can be more normally dispatched to, so that it can intercept the `subset` dispatch and then do whatever it needs to do.

> Also if my understanding is correct, for other tables, because of your overloaded may_dispatch(), subset () is called twice: first time in overloaded may_dispatch(), and they return true, so subset () is called a second time in dispatch()..

This would only be true if any other existing code already called `may_dispatch()` directly on a type with `subset`, which isn't the case because if they did they would have run into the same issues adding this `subset` did.

> We're currently working on supporting all tables, so I'd prefer implementing subset() for all other tables before we start implementing subset for extension table. This way there's no need to overload may_dispatch().

Overloading may_dispatch is more about `ExtensionFormat1` not being an entirely like the other tables, since it doesn't implement all the things it would need to participate in all the dispatches. However `ExtensionFormat1` is strange this way since it's more of an implementation detail. When things are actually written out it may or may not be wanted depending on the size of other things. It may be desirable to write one out when one didn't previously exist and it may be desirable not to write one out when did exist initially. It's more of an extra optional header on top of the existing tables (which also allows for / requires their data to be deferred to after all the lookup headers have been written out).

behdad:
Closing in favor of https://github.com/harfbuzz/harfbuzz/pull/2705

PR #2583, opened 2020-07-23
Remove unused code from the Khmer shaper.

RazrFalcon:
Looks like a dead code to me.


ebraminio:
Thank you very much :)

ebraminio:
Of course it may had some reason which now that removed who has added can have a look why was the propose.

PR #2584, opened 2020-07-23
[meson] favor builtin options over manual compiler flags

ebraminio:
So meson takes care of MSVC specific things also and no warnings when
passed to non C++ targets.

As https://github.com/mesonbuild/meson/issues/7437#issuecomment-662950521


ebraminio:
I just have one concern on this that may projects that like to have harfbuzz as a subproject fail to build if they don't have these default options as we also explicitly indicate the use of C linker, better to postpone this after the release I think.

PR #2588, opened 2020-07-26
Use fonttools's python package instead of cli

HrBDev:
This reduces the time needed for tests depending on fonttool's ttx(e.g 20/42 took ~34 seconds before this now it takes ~14 seconds)
Before:
image
After:
image

### DIFF
--- a/test/subset/run-tests.py
+++ b/test/subset/run-tests.py
@@ -12,13 +12,13 @@ import tempfile
 import shutil
 
 from subset_test_suite import SubsetTestSuite
-
-fonttools = shutil.which ("fonttools")
-ots_sanitize = shutil.which ("ots-sanitize")
-
-if not fonttools:
+try:
+	from fontTools import ttx
+except ImportError:
 	print ("fonttools is not present, skipping test.")
 	sys.exit (77)
+
+ots_sanitize = shutil.which ("ots-sanitize")
 
 def cmd (command):
 	p = subprocess.Popen (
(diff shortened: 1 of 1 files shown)


test/subset/run-tests.py
```diff
@@ -104,7 +104,7 @@ def run_test (test, should_check_ots):
 
 def run_ttx (font_path, ttx_output_path):
 	print ("fonttools ttx %s" % font_path)
-	return cmd ([fonttools, "ttx", "-q", "-o", ttx_output_path, font_path])
+	ttx.main (args=['-q', '-o', ttx_output_path, font_path])
```
khaledhosny, line 107: Probably not much faster, but you can skip ttx altogeyher, e.g.:
```python
from fontTools.ttLib import TTFont
with TTFont(font_path) as font:
	font.saveXML(ttx_output_path)
```

khaledhosny: You need to also return something from this function in case of failure.

khaledhosny: Since the code is saving to a file, just to read it back, you can also skip this and pass a `file`-like object (e.g. `io.StringIO` ) and skip writing to disk.

ebraminio: Thank you so much Khaled for the suggestions, have applied it on https://github.com/harfbuzz/harfbuzz/commit/322426f8c8ef818a60ec9090670ff077647fdfa2 please let me know anything else can be improved. Thanks! :)

PR #2602, opened 2020-07-29
[ENOMEM] make serializer modification operations no-ops if it's in an error state.

garretrieger:
If allocations fail during an operation such as push() then the serializer is left in a bad state. Further modifications can result in assertions being tripped. So disable all modifications once the serializer has entered an error state.


ebraminio:
Makes sense! Thank you!

PR #2607, opened 2020-07-30
Turn README to a soft redirect

### LINKED ISSUE #2591: windows: can't decompress windows archive because of symbolic link
can you remove symbolic link on README (or remove README) please ? tar fails with MSYS2. thank you

ebraminio:
Fixes #2591 guess we can have this one till we go for autotools removal

### DIFF
--- a/README
+++ b/README
@@ -1,1 +1,15 @@
-README.md
+This is HarfBuzz, a text shaping library.
+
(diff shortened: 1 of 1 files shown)


khaledhosny:
I’d make it a copy without the badges, it shouldn’t be to long before we completely remove it.

PR #2609, opened 2020-07-30
[ENOMEM] popragate packed/packed_map errors to the serializer.

garretrieger:
Will disable further modifications based on a bad state.


ebraminio:
tyvm! :)

Issue #2516, opened 2020-06-30
OT/CoreText shaping difference in Noto Nastaliq Urdu


simoncozens:
As mentioned in https://github.com/googlefonts/noto-fonts/issues/1783:

```
simon@upo ~/Downloads/NotoNastaliqUrdu-unhinted $ hb-shape --shapers=ot NotoNastaliqUrdu-Regular.ttf 'بلے'
[YBc1=2@1566,-576+0|YehBarreeFin.2_3=2+133|LamMed.outD2HYB=1@0,400+1037|OneDotBelowYB=0@151,174+0|sp9=0+0|BehxIni.A=0@0,512+1609]
simon@upo ~/Downloads/NotoNastaliqUrdu-unhinted $ hb-shape --shapers=coretext NotoNastaliqUrdu-Regular.ttf 'بلے'
[OneDotBelowYB=2@1566,-374+0|YBc1=2@1566,-576+-1566|YehBarreeFin.2_3=2@1566,0+133|LamMed.outD2HYB=1@1566,400+1037|NullMk=0@1566,400+0|sp9=0@1566,512+0|BehxIni.A=0@1566,512+3175]
```

Dropping the issue here to track it for now. I plan to run the lookups by hand and see where the problem is.

simoncozens:
The problem is in [lookup 42](https://github.com/googlefonts/noto-source/blob/master/src/NotoNastaliqUrdu/Noto%20Nastaliq%20Urdu%20Regular%20GSUB.txt#L7981), which is chained from [lookup 38](https://github.com/googlefonts/noto-source/blob/master/src/NotoNastaliqUrdu/Noto%20Nastaliq%20Urdu%20Regular%20GSUB.txt#L2365). When I apply it by hand, I get the following "trace":

```
38
  Class 5                  Class 29       Class 2
  This fires off two chain rules: lookup 42 pos 1, lookup 143 pos 1

  Chain 1,42
    BehxIni.A OneDotBelowYB LamMed.outD2HYB YehBarreeFin.2_3 YBc1
    Class 1   Class 10      Class 2         Class 4          Class 6
              Another two chain rules: 45 pos 2, 46 pos 2
              Chain 2, 45
              OneDotBelowYB LamMed.outD2HYB YehBarreeFin.2_3 YBc1
              Class 14      Class 2         Class 4          Class 6
                                                             Chain 4, 52
                                                             YBc1 -> YBc1 OneDotBelowYB
              Chain 2, 46
[...]

PR #2556, opened 2020-07-09
2.7.0

ebraminio:
Expected to be published in some days (as I expect some new fuzzing issues to pop out and am softly waiting for #2553 also somehow), uploaded however to get feedback on the release note, is the release note too verbose or has grammar/spell mistakes, or not happy to with things it suggests? Feel free to give me your feedback :)

PR #2616, opened 2020-08-02
Remove autotools build support

ebraminio:
No rush but now can happen just anytime, considering already adoptions listed on https://github.com/harfbuzz/harfbuzz/wiki/Notes-on-migration-to-meson


ebraminio:
@khaledhosny what timeline do you imagine for this? Maybe we can first just remove the tests install and dist and keep the rest for sometime, for regression testing or so, or keeping it for some months or remove it immediately?

khaledhosny:
I’d say keep it as long as feasible (minimal maintenance, no new features) and drop when it starts to be a maintenance hurdle or you are confident that
 no one is still using it.

ebraminio:
> it starts to be a maintenance hurdle

For rewriting tests with the tests, or using Google Tests, as was suggested by Behdad, it is making things tedious. That's why even tests support removal will be good I think. WDYT?

> you are confident that no one is still using it.

Correct. This will be a long long time :/

Thank you so much!

ebraminio:
Remove as it won't happen till some releases as you suggested, I will  justcontinue the notifying the packagers for meson use in the meanwhile to speed up the transition. Thanks for your help Khaled :)

ebraminio:
Had a discussion with Behdad and he was actually in favor of autotools removal, personally I liked to be conservative here but he said our current dists are against autotools philosophy anyway and better to not keep the project in this inconsistent state.

khaledhosny:
Having autotools around was still good for bisecting from before meson support, it allowed me to build quickly each commit without having to check whether meson was present at that point or not.

khaledhosny:
Now I have an idea, create a repo with hb-view/hb-shape built from each HarfBuzz commit (since, say 1.0.0), so bisecting can be quick (LibreOffice has something like this, but building LibreOffice takes hours so makes sense there).

ebraminio:
> Having autotools around was still good for bisecting from before meson support, it allowed me to build quickly each commit without having to check whether meson was present at that point or not.

Agreed, as have mentioned it in the first comment also.

> Now I have an idea, create a repo with hb-view/hb-shape built from each HarfBuzz commit (since, say 1.0.0), so bisecting can be quick (LibreOffice has something like this, but building LibreOffice takes hours so makes sense there).

Wow, that's just fantastic.

behdad:
> No rush

And you already merged! Can you please, for whatever's sake, get feedback before merging things :(. Yes, I told you to go ahead and do please. But properly...

The `INSTALL` file is outdated.

I just sat down to debug and fix #2516 (which is something much more important, you should have reverted the regressing commit and made release with), but can't even build.  On Fedora 31 I'm getting:

```
$ meson build
The Meson build system
Version: 0.52.0
Source dir: /home/behdad/harfbuzz
Build dir: /home/behdad/harfbuzz/build
Build type: native build

meson.build:1:0: ERROR: Meson version is 0.52.0 but project requires >= 0.53.0
```

ebraminio:
> you should have reverted the regressing commit and made release with

I have asked Khaled that https://github.com/harfbuzz/harfbuzz/pull/2556#issuecomment-657112152 he is active on reviewing and is definitely more knowledgeable that me. As can be seen on the issue even delayed 2.7.0 just for that as I knew its importance.

> The INSTALL file is outdated.

Which file is it? I've updated most of docs before the change, like https://github.com/harfbuzz/harfbuzz/blob/master/BUILD.md and `TESTING.md`, `BUILD.md`, `CONFIG.md` and `usermanual-install-harfbuzz.xml`, perhaps something overlooked.

> Meson version is 0.52.0 

We can support support 0.52.0 but we are using [`summary`](https://github.com/harfbuzz/harfbuzz/blob/master/meson.build#L359-L383) feature of meson that we like packagers to have so they now they are building on par features with autotools. An updated meson just easily can be installed in your system, just using pip, `pip3 install meson`

However we will support 0.52.0 if some packagers ask, just that your Fedora version, 31, has 2.6.1 per https://repology.org/badge/vertical-allrepos/harfbuzz.svg

ebraminio:
> The `INSTALL` file is outdated.

Apparently `INSTALL` is autotools generated file and isn't in the repo at all?

PR #2622, opened 2020-08-07
Use printf instead of echo -n in autogen.sh

ryandesign:
"echo -n" is not portable and on some POSIX systems it literally prints "-n".

When I run autogen.sh without this patch on macOS 10.13.6, it prints:

```
-n checking for pkg-config...
/opt/local/bin/pkg-config
```

With this patch, it correctly prints:

```
checking for pkg-config... /opt/local/bin/pkg-config
```

etc.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -7,24 +7,24 @@ test -n "$srcdir" || srcdir=.
 olddir=`pwd`
 cd $srcdir
 
-#echo -n "checking for ragel... "
+#printf "checking for ragel... "
 #which ragel || {
 #	echo "You need to install ragel... See http://www.complang.org/ragel/"
 #	exit 1
 #}
 
-echo -n "checking for pkg-config... "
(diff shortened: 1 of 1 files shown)


ebraminio:
Can macports instead consider for meson migration https://github.com/harfbuzz/harfbuzz/wiki/Notes-on-migration-to-meson just like https://github.com/Homebrew/homebrew-core/blob/master/Formula/harfbuzz.rb homebrew?

ryandesign:
I have had nothing but terrible experiences with meson and don't want to invest time in figuring out how to make it work here right now.

ebraminio:
> I have had nothing but terrible experiences with meson and don't want to invest time in figuring out how to make it work here right now.

We can't keep multiple build systems on par for a long time of course but sure, but sure, can have this innocent looking change in in the meanwhile.

Issue #2597, opened 2020-07-29
CMake Windows & static libs support


lisyarus:
Are there any plans on supporting Windows and/or static version of harfbuzz in CMake? Currently, `harfbuzz-config.cmake.in` just declares Windows as an unsupported platform, and unconditionally declares `SHARED` libraries.

ebraminio:
harfbuzz config is at least somehow considered a legacy thing. Even if not removed at #2601, we may accept fixes for it but not actively maintaining it. So. Thanks :)

Issue #2316, opened 2020-04-06
Incorrect prefix computation in harfbuzz-config.cmake


lisyarus:
On my Gentoo system with harfbuzz-2.6.4, in appears that `harfbuzz-config.cmake` (located in `/usr/lib/cmake/harfbuzz/`) is incorrect. Specifically, it contains

```
get_filename_component(_harfbuzz_prefix "${CMAKE_CURRENT_LIST_DIR}" DIRECTORY)
get_filename_component(_harfbuzz_prefix "${_harfbuzz_prefix}" DIRECTORY)
set(_harfbuzz_libdir "/usr/lib64")
string(REPLACE "${_harfbuzz_remove_string}/" "" _harfbuzz_libdir "${_harfbuzz_libdir}")
set(_harfbuzz_libdir_iter "${_harfbuzz_libdir}")
while (_harfbuzz_libdir_iter)
  set(_harfbuzz_libdir_prev_iter "${_harfbuzz_libdir_iter}")
  get_filename_component(_harfbuzz_libdir_iter "${_harfbuzz_libdir_iter}" DIRECTORY)
  if (_harfbuzz_libdir_prev_iter STREQUAL _harfbuzz_libdir_iter)
    break()
  endif ()
  get_filename_component(_harfbuzz_prefix "${_harfbuzz_prefix}" DIRECTORY)
endwhile ()
unset(_harfbuzz_libdir_iter)
```

which, as I can see, leads to `_harfbuzz_prefix` being just `/` and later the include path to be `//include/harfbuzz`, which is clearly wrong and basically makes the whole config file useless.

I'm attaching the full `harfbuzz-config.cmake`.

[harfbuzz-config.cmake.txt](https://github.com/harfbuzz/harfbuzz/files/4440598/harfbuzz-config.cmake.txt)

ebraminio:
This was provided by @mathstuf in d38f37b2e3794 and to be honest I don't have any expertise on it, would you please see also how it can be fixed? Otherwise its fix may take more time. Thanks!

ebraminio:
Oh @vkrause has apparently something related also in 9e9a36ee65 (very sorry for ping noise btw)

mathstuf:
Hmm. Do you have symlinks in your install path or in the path you're using to refer to the directory? Do the symlinks point to the install prefix itself or directly into it?

lisyarus:
Thanks you a lot for quick response.
[...]

PR #2623, opened 2020-08-07
Bring back cmake config support but declare it as deprecated

ebraminio:
Reverts 75efa89

Just days ago I wanted to use harfbuzz in some cmake based project and I think when I used find_package(harfbuzz) I actually used this bit. Equally I fear many people are relying to this bit that's why am proposing this change.

This change brings back cmake config support only for meson build installs as I wanted to apply some optimizations also. Personally as I wanted to propose cmake as the main build system for HarfBuzz sometime, cmake isn't illegible to me and actually I used to like it so as the so called maintainer of the project am able to maintain this thing also but I just don't like issues like https://github.com/harfbuzz/harfbuzz/issues/2316 and https://github.com/harfbuzz/harfbuzz/issues/2597 popping out that's why I am declaring it as deprecated also (but that can be discussed also)

Let's discuss this first however, @mathstuf @lisyarus and @khaledhosny

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -16,6 +16,7 @@ hb_version_micro = hb_version_arr[2].to_int()
 
 # libtool versioning
 hb_version_int = hb_version_major*10000 + hb_version_minor*100 + hb_version_micro
+hb_libtool_version_info = '@0@:0:@0@'.format(hb_version_int)
 
 pkgmod = import('pkgconfig')
 cpp = meson.get_compiler('cpp')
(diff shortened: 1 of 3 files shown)


khaledhosny:
Unless you are building on a system without pkg-config, you should be able to use it with CMake https://cmake.org/cmake/help/latest/module/FindPkgConfig.html

ebraminio:
So let's close as nobody seems interested. Thanks

PR #2606, opened 2020-07-30
Skip hb_shape if buffer object is immutable

ebraminio:
@khaledhosny does it make sense? Maybe doesn't worth it as it is highly unlikely or maybe not? Please have a look. Thanks!

PR #2625, opened 2020-08-08
[ENOMEM] Return true from _plan_execute only if the buffer is valid

ebraminio:
An alternative fix to https://github.com/harfbuzz/harfbuzz/pull/2606

More details on https://github.com/harfbuzz/harfbuzz/pull/2606#issuecomment-670325875

We are modifying the null pool hosted `hb_buffer_t` (which happens when `hb_buffer_create` gets ENOMEM) by returning true from `hb_shape_plan_execute` which makes `if (res) buffer->content_type = HB_BUFFER_CONTENT_TYPE_GLYPHS;` of `hb_shape_full` be triggered. This changes  returns false on INVALID buffers (which includes null pool hosted `hb_buffer_t`).

### DIFF
--- a/src/hb-shape-plan.cc
+++ b/src/hb-shape-plan.cc
@@ -383,7 +383,7 @@ hb_shape_plan_execute (hb_shape_plan_t    *shape_plan,
 		  shape_plan->key.shaper_name);
 
   if (unlikely (!buffer->len))
-    return true;
+    return buffer->content_type == HB_BUFFER_CONTENT_TYPE_UNICODE;
 
   assert (!hb_object_is_immutable (buffer));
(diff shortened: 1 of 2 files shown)


ebraminio:
I think this makes more sense however doing `if (unlikely (hb_object_is_immutable (buffer))) return false;` was idiomatic between similar calls, so not sure about it.

ebraminio:
tl;dr, we are modifying the null pool hosted `hb_buffer_t` by returning true from `hb_shape_plan_execute` which makes `if (res) buffer->content_type = HB_BUFFER_CONTENT_TYPE_GLYPHS;` of `hb_shape_full` be triggered. This returns false on INVALID buffers (which includes null pool hosted `hb_buffer_t`).

behdad:
Pushed alternate fix.

PR #2627, opened 2020-08-08
[ENOMEM] Don't use raw pointer of GSUBGPOS::accelerator_t::accels

ebraminio:
It can be invalid if [this calloc](https://github.com/harfbuzz/harfbuzz/blob/06dbb6acbbd322b3123820c73ed4371560efc558/src/hb-ot-layout-gsubgpos.hh#L3376) of GSUBGPOS::accelerator_t has failed to allocate memory, better to not use raw pointers.

[This](https://github.com/harfbuzz/harfbuzz/blob/122a2897e4782c5e0a39ec7bb0eedb2e7d568913/src/hb-ot-layout.cc#L1867) is where the raw pointer can be dereferenced incorrectly which is discovered by failing alloc fuzzers.

Fixes https://crbug.com/oss-fuzz/24473 and https://crbug.com/oss-fuzz/24490 and https://crbug.com/oss-fuzz/24508 and https://crbug.com/oss-fuzz/24842

The second change, CI one, is also here as this change is needed to make some of the previously added tests to work also hitting the same issue.

How to reproduce?

`CXXFLAGS="-fsanitize=address,fuzzer-no-link" meson fuzzbuild --default-library=static -Dfuzzer_ldflags="-fsanitize=address,fuzzer" -Dexperimental_api=true && ninja -C fuzzbuild test/fuzzing/hb-{shape,draw,subset,set}-fuzzer && fuzzbuild/test/fuzzing/hb-shape-fuzze test/fuzzing/fonts/clusterfuzz-testcase-hb-shape-fuzzer-5721479954825216`

Of course generally we should also think about what changes to shaping is acceptable when we are in OOM (which I know as OS's overcommit having the chance to handle those isn't that common, still), obviously we shouldn't alter semantics of text on OOM and maybe we should communicate with clients about OOM in some better way. This change however doesn't change the intended behavior as we had that `calloc` result and length check anyway just that we should be careful about them generally.

Or maybe we should think about making those accels store limited and stack allocated?


ebraminio:
Not a good solution I think, somehow relies on calloc return not null on zero sized allocation which isn't always try, a better fix will be to limit [lookup_index](https://github.com/harfbuzz/harfbuzz/blob/122a2897e4782c5e0a39ec7bb0eedb2e7d568913/src/hb-ot-layout.cc#L1867) with accelerator's [lookup_count](https://github.com/harfbuzz/harfbuzz/blob/06dbb6acbbd322b3123820c73ed4371560efc558/src/hb-ot-layout-gsubgpos.hh#L3378)

ebraminio:
@jfkthame @khaledhosny @behdad can one of you review this? Sorry for random pings these day, will be reduce in the coming days :) Thank you very much!

ebraminio:
Closing in favor of #2728, easier yet less costly way to me, I measured shaping performance before and after this and saw this had some effect there which won't be the case that patch AFAICT

PR #2628, opened 2020-08-09
Turn -Wunused-but-set-variable to warning

### LINKED ISSUE #2555: error: variable ‘gapg_temp_atomic’ set but not used [-Werror=unused-but-set-variable]
I'm trying to comile harfbuzz but get 17 errors similar to the following:

```
  CXX      libharfbuzz_gobject_la-hb-gobject-structs.lo
In file included from /srv/devel/waf-develop-8.3/tps/binaries/centos-7/x86_64/include/glib-2.0/glib/gthread.h:32:0,
                 from /srv/devel/waf-develop-8.3/tps/binaries/centos-7/x86_64/include/glib-2.0/glib/gasyncqueue.h:32,
                 from /srv/devel/waf-develop-8.3/tps/binaries/centos-7/x86_64/include/glib-2.0/glib.h:32,
                 from hb-gobject-structs.cc:53:
hb-gobject-structs.cc: In function ‘GType hb_gobject_buffer_get_type()’:
/srv/devel/waf-develop-8.3/tps/binaries/centos-7/x86_64/include/glib-2.0/glib/gatomic.h:128:15: error: variable ‘gapg_temp_atomic’ set but not used [-Werror=unused-but-set-variable]
     gpointer *gapg_temp_atomic = (gpointer *)(atomic);                       \
               ^
/srv/devel/waf-develop-8.3/tps/binaries/centos-7/x86_64/include/glib-2.0/glib/gthread.h:251:7: note: in expansion of macro ‘g_atomic_pointer_get’
     (!g_atomic_pointer_get (location) &&                             \
       ^
hb-gobject-structs.cc:68:8: note: in expansion of macro ‘g_once_init_enter’
    if (g_once_init_enter (&type_id)) { \
        ^
hb-gobject-structs.cc:78:2: note: in expansion of macro ‘HB_DEFINE_BOXED_TYPE’
  HB_DEFINE_BOXED_TYPE (name, hb_##name##_reference, hb_##name##_destroy)
  ^
hb-gobject-structs.cc:91:1: note: in expansion of macro ‘HB_DEFINE_OBJECT_TYPE’
 HB_DEFINE_OBJECT_TYPE (buffer)
 ^
```

**Environment:**
   CentOS 7
   gcc --version = gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-39)
   harfbuzz versions tried: 2.6.8, 2.6.7, 2.6.4, 2.5.3, 2.3.1

**Relevant part of larger Makefile:**
```
LD_LIBRARY_PATH="$(OUTPUT_DIR)/lib:$(OUTPUT_DIR)/lib64" \
PKG_CONFIG_PATH="$(OUTPUT_DIR)/lib/pkgconfig:$(OUTPUT_DIR)/lib64/pkgconfig" \
./configure --prefix=$(OUTPUT_DIR) --enable-static --disable-shared \
--disable-gtk-doc-html --with-gobject --enable-introspection

PATH="$(OUTPUT_DIR)/bin:$(PATH)" \
LD_LIBRARY_PATH="$(OUTPUT_DIR)/lib:$(OUTPUT_DIR)/lib64" \
PKG_CONFIG_PATH="$(OUTPUT_DIR)/lib/pkgconfig:$(OUTPUT_DIR)/lib64/pkgconfig" \
$(MAKE) -C $(dir $@)
```

I don't see any place where any `-Werror*` is set.

ebraminio:
As #2555 turned out some glib headers are imposing that so let's turn it to warning

Fixes #2555

### DIFF
--- a/src/hb.hh
+++ b/src/hb.hh
@@ -107,6 +107,7 @@
 #pragma GCC diagnostic warning "-Wmaybe-uninitialized"
 #pragma GCC diagnostic warning "-Wmissing-format-attribute"
 #pragma GCC diagnostic warning "-Wundef"
+#pragma GCC diagnostic warning "-Wunused-but-set-variable"
 #endif
 
 /* Ignored currently, but should be fixed at some point. */


ebraminio:
@clpo13 @jfrickson can you check whether this fixes your issue? Thanks!

khaledhosny:
Can you disable it only around `glib.h` includes? Looks like its included only 4 times in the repository.

ebraminio:
That will sound much better so let's close this and trying to upload that separately. Thanks

ebraminio:
> Can you disable it only around glib.h includes? Looks like its included only 4 times in the repository.

Done. Feeling a bit better about the previous as this has made a bit more noise, still. As couldn't add it to public APIs have added it where hb-glib.h was actually included inside the project. PTAL

khaledhosny:
My point was it would allow keeping this as error for the rest of the code (since warnings are often missed, and our build is already noisy because of gtk-doc warning, so it might make it even more easy to miss). Up to you to decide which is better.

ebraminio:
> and our build is already noisy because of gtk-doc warning

Correct, those warnings are coming from introspection, in autotools they were silenced, one can pass -Dinstrospection=false to config but that isn't ideal of course :/

ebraminio:
> Up to you to decide which is better.

Okay, am going with the less noisy approach than https://github.com/harfbuzz/harfbuzz/commit/8af9ecb075be60244372a29b855b9b87772b7ed1 and surely we can improve it later. Thanks

behdad:
I hate these "can improve it later" merges I keep discovering... Specially since Khaled clearly raised the same concerns that I have with the code that is merged...

PR #2619, opened 2020-08-06
[var] Apply variations on empty glyphs

ebraminio:
Fixes #2618

PR #2630, opened 2020-08-11
[glyf] Remove an unnecessary check for empty glyphs 

### LINKED ISSUE #2618: hb-ot font functions fail to scale space glyph by phantom points & GVAR method (without HVAR)
Expected, using FreeType based font-funcs, space scales from 180 to 300 font units.
```
$ hb-shape --variations="wdth=1" --font-funcs=ft ~/dev/GT-Alpina-GX.ttf " "
[space=0+180]
$ hb-shape --variations="wdth=1000" --font-funcs=ft ~/dev/GT-Alpina-GX.ttf " "
[space=0+300]
```

Incorrect, using OT font-funcs, space does not scale.
```
$ hb-shape --variations="wdth=1" --font-funcs=ot ~/dev/GT-Alpina-GX.ttf " "
[space=0+180]
$ hb-shape --variations="wdth=1000" --font-funcs=ot ~/dev/GT-Alpina-GX.ttf " "
[space=0+180]
```

See [Chromium bug 1102739](https://bugs.chromium.org/p/chromium/issues/detail?id=1102739) and download the font [from comment 17](https://bugs.chromium.org/p/chromium/issues/detail?id=1102739#c17) on that issue.

Related to https://github.com/harfbuzz/harfbuzz/issues/2365, but that issue was fixed in Chromium by applying variation parameters to the parent hb_font, see [relevant CL](https://chromium-review.googlesource.com/c/chromium/src/+/2161227/2/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_face.cc).

For reference, deltas for space glyph of GT-Alpina GVAR:
```
    <glyphVariations glyph="space">
      <tuple>
        <coord axis="wght" min="0.0" value="0.38776" max="1.0"/>
        <delta pt="0" x="0" y="0"/>
        <delta pt="1" x="-10" y="0"/>
        <delta pt="2" x="0" y="0"/>
        <delta pt="3" x="0" y="0"/>
      </tuple>
      <tuple>
        <coord axis="wght" min="0.38776" value="1.0" max="1.0"/>
        <delta pt="0" x="0" y="0"/>
        <delta pt="1" x="-20" y="0"/>
        <delta pt="2" x="0" y="0"/>
        <delta pt="3" x="0" y="0"/>
      </tuple>
      <tuple>
        <coord axis="wdth" min="0.0" value="0.4995" max="1.0"/>
        <delta pt="0" x="0" y="0"/>
        <delta pt="1" x="45" y="0"/>
        <delta pt="2" x="0" y="0"/>
        <delta pt="3" x="0" y="0"/>
      </tuple>
      <tuple>
        <coord axis="wght" min="0.0" value="0.38776" max="1.0"/>
        <coord axis="wdth" min="0.0" value="0.4995" max="1.0"/>
        <delta pt="0" x="0" y="0"/>
        <delta pt="1" x="-10" y="0"/>
        <delta pt="2" x="0" y="0"/>
        <delta pt="3" x="0" y="0"/>
      </tuple>
      <tuple>
        <coord axis="wght" min="0.38776" value="1.0" max="1.0"/>
        <coord axis="wdth" min="0.0" value="0.4995" max="1.0"/>
        <delta pt="0" x="0" y="0"/>
        <delta pt="1" x="-10" y="0"/>
        <delta pt="2" x="0" y="0"/>
        <delta pt="3" x="0" y="0"/>
      </tuple>
      <tuple>
        <coord axis="wdth" min="0.4995" value="1.0" max="1.0"/>
        <delta pt="0" x="0" y="0"/>
        <delta pt="1" x="120" y="0"/>
        <delta pt="2" x="0" y="0"/>
        <delta pt="3" x="0" y="0"/>
      </tuple>
      <tuple>
        <coord axis="wght" min="0.0" value="0.38776" max="1.0"/>
        <coord axis="wdth" min="0.4995" value="1.0" max="1.0"/>
        <delta pt="0" x="0" y="0"/>
        <delta pt="1" x="-20" y="0"/>
        <delta pt="2" x="0" y="0"/>
        <delta pt="3" x="0" y="0"/>
      </tuple>
      <tuple>
        <coord axis="wght" min="0.38776" value="1.0" max="1.0"/>
        <coord axis="wdth" min="0.4995" value="1.0" max="1.0"/>
        <delta pt="0" x="0" y="0"/>
        <delta pt="1" x="-40" y="0"/>
        <delta pt="2" x="0" y="0"/>
        <delta pt="3" x="0" y="0"/>
      </tuple>
    </glyphVariations>
```

ebraminio:
@khaledhosny can you have a look please? Thanks

### DIFF
--- a/src/hb-ot-glyf-table.hh
+++ b/src/hb-ot-glyf-table.hh
@@ -818,10 +818,7 @@ struct glyf
 
     bool get_extents (hb_font_t *font, const accelerator_t &glyf_accelerator,
 		      hb_glyph_extents_t *extents) const
-    {
-      if (type == EMPTY) return true; /* Empty glyph; zero extents. */
-      return header->get_extents (font, glyf_accelerator, gid, extents);
-    }
+    { return header->get_extents (font, glyf_accelerator, gid, extents); }
 
     hb_bytes_t get_bytes () const { return bytes; }
 


test/api/test-font.c
```diff
@@ -119,6 +119,21 @@ test_face_createfortables (void)
...
+{
+  hb_face_t *face = hb_test_open_font_file ("fonts/empty-glyph.ttf");
+  hb_font_t *font = hb_font_create (face);
+  hb_face_destroy (face);
+  hb_glyph_extents_t extents;
+  g_assert (hb_font_get_glyph_extents (font, 1, &extents));
+  g_assert_cmpint (extents.x_bearing, ==, 0);
+  g_assert_cmpint (extents.y_bearing, ==, 0);
+  g_assert_cmpint (extents.width, ==, 0);
+  g_assert_cmpint (extents.height, ==, 0);
```
ebraminio, line 133: Manually checked that hits the touched part however was more interesting if was non zero one, hard to reach part of the code anyway.

khaledhosny: What was the result before this change?

ebraminio: Was also zero that's why the test isn't that good, apparently fonttools itself sets xMin xMax yMin yMax so not sure how to set those properly in a test font.

khaledhosny: Sorry, I don’t understand what this change is doing then to make any informed review.

ebraminio: You are absolutely right, do you know some I can tweak xMin yMin xMax yMax without having to deal with

```
<!-- The xMin, yMin, xMax and yMax values
         will be recalculated by the compiler. -->
```

of fonttools. Some binary tweak maybe? Thanks

ebraminio: Okay, apparently I don't understand this fully also, let me close so I can came with a better understanding and real test. Thanks

ebraminio:
Ok, in fact https://github.com/harfbuzz/harfbuzz/pull/2619 isn't filed with xMin/xMax/yMin/yMax of the certain empty glyphs, they all are zero and gvar tweaks are turning those zeros to usable values.

So this check really doesn't matter other than our general preference is to not reduce checks by using null pool, at the time [this comment was put](https://github.com/harfbuzz/harfbuzz/blob/5c4fead734b082e0168e6811bec4bcaa19acc36a/src/hb-ot-glyf-table.hh#L448) there were no null pool there AFAICS but now that we are putting [null pool for header](https://github.com/harfbuzz/harfbuzz/blob/d03eecb/src/hb-ot-glyf-table.hh#L828-L836) removing the check doesn't make the difference.

This also doesn't need a new test also as it doesn't change the behavior, it is by hit ot-color's sbix font extents check and is hit by fuzzers also.

So given the explanation guess is worth take another look if you like to @khaledhosny

khaledhosny:
Sorry, but I still don’t understand what this issue is fixing. Empty glyph has no extents, you can’t vary something that does not exist.

ebraminio:
> Sorry, but I still don’t understand what this issue is fixing. Empty glyph has no extents, you can’t vary something that does not exist.

I think this isn't changing that behavior but intended more to remove special casing. This is what is done [since 2018](https://github.com/harfbuzz/harfbuzz/blob/ffcb5cbf71b312cdd4606a3285e10e201ff5bccd/src/hb-ot-glyf-table.hh#L442) somehow, reporting success for empty glyph get_extents.

ebraminio:
Okay doesn't worth the risk to change, let's revisit sometime later maybe. Thanks

PR #2632, opened 2020-08-11
2.7.1

ebraminio:
Let's see if everyone is happy about this. Guess some factual/grammar/wording improvements will be needed also on NEWS.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,11 @@
+Overview of changes leading to 2.7.1
+Thursday, Aug 13, 2020
+====================================
+- ot-funcs now handles variable empty glyphs better when hvar/vvar isn't present.
+- Reverted a GDEF processing regression.
+- A couple of fixes to handle OOM better.
+
+
 Overview of changes leading to 2.7.0
 Saturday, July 25, 2020
 ====================================
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,6 @@
 AC_PREREQ([2.64])
 AC_INIT([HarfBuzz],
-        [2.7.0],
+        [2.7.1],
         [https://github.com/harfbuzz/harfbuzz/issues/new],
         [harfbuzz],
         [http://harfbuzz.org/])
--- a/meson.build
+++ b/meson.build
@@ -1,6 +1,6 @@
 project('harfbuzz', 'c', 'cpp',
   meson_version: '>= 0.47.0',
-  version: '2.7.0',
+  version: '2.7.1',
   default_options: [
     'cpp_eh=none',          # Just to support msvc, we are passing -fno-rtti also anyway
     'cpp_rtti=false',       # Just to support msvc, we are passing -fno-exceptions also anyway
--- a/src/hb-version.h
+++ b/src/hb-version.h
@@ -38,9 +38,9 @@ HB_BEGIN_DECLS
 
 #define HB_VERSION_MAJOR 2
 #define HB_VERSION_MINOR 7
-#define HB_VERSION_MICRO 0
+#define HB_VERSION_MICRO 1
 
-#define HB_VERSION_STRING "2.7.0"
+#define HB_VERSION_STRING "2.7.1"
 
 #define HB_VERSION_ATLEAST(major,minor,micro) \
 	((major)*10000+(minor)*100+(micro) <= \


NEWS
```diff
@@ -1,3 +1,15 @@
+Overview of changes leading to 2.7.0
+Wednesday, Aug 12, 2020
+====================================
+- Autotools and cmake build supports are removed in favor of meson.
```
alerque, line 4:
```suggestion
- Autotools and cmake build supports are removed in favor of meson.
- Removed autotools and cmake build support in favor of meson.
```

NEWS
```diff
@@ -1,3 +1,15 @@
...
+  More on https://github.com/harfbuzz/harfbuzz/wiki/Notes-on-migration-to-meson
+- cmake-config support is removed, we suggest use of pkg-config in cmake projects.
```
alerque, line 6:
```suggestion
- cmake-config support is removed, we suggest use of pkg-config in cmake projects.
- Removed cmake-config support, we suggest using pkg-config in cmake projects.
```

NEWS
```diff
@@ -1,3 +1,15 @@
...
+  See https://github.com/harfbuzz/harfbuzz/issues/2316 for more context.
+- ot-funcs now handles variable empty glyphs better when hvar/vvar isn't present.
+- Reverts a shaping regression on fonts lack GDEF.
```
alerque, line 9:
```suggestion
- Reverted a shaping regression for fonts which lack GDEF.
```

NEWS
```diff
@@ -1,3 +1,15 @@
+Overview of changes leading to 2.7.0
```
alerque, line 1:
```suggestion
Overview of changes leading to 2.7.1
```

ebraminio:
@alerque fantastic, thank you very much! :)

khaledhosny:
I think removing autotools was a bit rushed, and if HarfBuzz [does not build out of box on a last year’s release of a major Linux distribution](https://github.com/harfbuzz/harfbuzz/pull/2616#issuecomment-670969052), that is not good either. I suggest we restore autotools build before release, cut this release and then discuss the remaining issue (e.g. if we can lower the requires meson version a bit).

alerque:
I'll chime in to say I think changing the build system has been a little rushed. Specifically it's kind of weird to do it in a patch level release. I think I'd like to see that actually trotted out for 2.8.0 where at least the minor version is a tip off more package maintainer involvement might be necessary. Having the new system available for early adopters has gone well, but removing the old system just doesn't feel right for a patch release number.

ebraminio:
Brought back autotools support, just to note https://github.com/harfbuzz/harfbuzz/pull/2616#issuecomment-671040278 I didn't want it also, I even have close that PR after Khaled comment, maybe let's call it happened in a communication error. This is my part of the story, in https://github.com/harfbuzz/harfbuzz/issues/2536#issue-650935767 I asked whether we can drop dist tarball upload, I have one clear reason for this, there is not easy way for packager to check whether this is aligning really with the repository content and multiple people have access to change the tarball, and I think you can see that Behdad agreed with it as you see there. Recently he said why there is no autotools dist and I said this is what we agreed for, isn't it, and he said we should either check in all the autotools generated files or remove autotools support in favor of meson, the first option looked horrible to me as need to tweak those .gitignore also so I said explicitly to him I am going with the second option if that is what he wants, I uploaded a PR for this before the discussion but he apparently and undestandibly haven't seen and then he asked why it is merged (the thing I didn't want also) and it was surprising for me, I understand him as he is busy with other things but at least like  to inform others what has happened.

So in conclusion let's upload a autotools dist tarball just like old good days, is that what everyone agrees with?

ebraminio:
> if we can lower the requires meson version a bit

We can but gtk also requires 0.53.0 https://gitlab.gnome.org/GNOME/gtk/blob/master/meson.build#L10 for the same reason as us, `summary`, so we now packagers are OK with it, is there something that different about `gtk` not applies here? I wanted to hear such request from some packagers first.

behdad:
> We can but gtk also requires 0.53.0 https://gitlab.gnome.org/GNOME/gtk/blob/master/meson.build#L10 for the same reason as us, `summary`, so packagers know they are building something intended,

Since it's inception, HarfBuzz could be built using "./configure && make" on any POSIX system with shell, make, and a C++ compiler. It was intentional to not require glib, FreeType, gtk-doc, etc.  So you are telling me that showing a nice `summary` is worth requiring a user of a current Linux distro to install a package before then can build HarfBuzz?

> is there something that different about `gtk` not applies here?

Yes. GTK+ is a framework that has at least a dozen required packages. It also never gained the billion+-user adoption that HarfBuzz has gained.

My philosophy building HarfBuzz was to look at and learn from how other libraries are built, and IMPROVE ON THAT. If I aimed HarfBuzz to be just like any other library, well, I would have at best got it to succeed as much as other libraries.

*You* know better than anyone the troubles I went through to revive g++-4.9 support such that we can support RHEL6. You wanted an amalgam to simplify building. You wanted cmake to simplify building. Yet you tell me is okay to do what you did with 2.7.0. Shipping without autotools build files, your explanation was that "distros seem to have figured out without it". That's a **very** low bar. I want random user who wants to build HarfBuzz on their 6-year-old Linux server to be able to build it ***easily***.

> I wanted to hear such request from some packagers first to see what I can do about.

I have no idea what you are talking about. Just don't break thing without discussing it first. You submit a PR then merge immediately. What's the point then? Just push direct to master if that's what you are going to do.

behdad:
> This is my take of the story, in [#2536 (comment)](https://github.com/harfbuzz/harfbuzz/issues/2536#issue-650935767) I asked whether we can drop dist tarball upload, I have clear reasons for this, there is no easy ways for packager to check whether the dist tarball is really aligning with the repository's content and multiple people have access to upload the tarball afterward so if anything happens it will be blamed on me as there is no log on release page AFAICS,

This is a goal I applaud and agree with.

> and I think you can see that Behdad agreed with it. But then recently he said why there is no autotools dist and I said this is what we agreed for, isn't it,

What I agreed with was that eventually when we switch to meson as our main build system, that's what happens.

> and he said we should either check in all the autotools generated files or remove autotools support in favor of meson, the first option looked horrible to me as need to tweak those automatic generated .gitignore also so I said explicitly to him I am going with the second option if that is what he wants,

Yes. I told you to remove autotools if it's being left dysfunctional.  I assumed the meson system is **usable**.

> I uploaded a PR for this before the discussion

And merged it before anyone had a chance to give any feedback.

> but he apparently and now understandibly haven't seen and then he asked why it is merged (the thing I didn't want also)

I want it merged, but I found that, unlike what I've been made to believe, the meson system was NOT functional on a very recent Linux distro.

> and it was surprising for me as I thought this is what he wants, I understand him as he is busy with other things but at least like to inform others what has happened.

In summary: we all agree on the end goal. You seem to deem it okay to break things on the road getting there. I want the project to maintain it's extremely high standards of quality and usability while getting there.

ebraminio:
> I want random user who wants to build HarfBuzz on their 6-year-old Linux server to be able to build it easily.

This is understandable, something I didn't know of course.

ebraminio:
> You wanted an amalgam to simplify building. You wanted cmake to simplify building. Yet you tell me is okay to do what you did with 2.7.0. Shipping without autotools build files, your explanation was that "distros seem to have figured out without it". That's a very low bar.

This is what I though your agreement here https://github.com/harfbuzz/harfbuzz/issues/2536#issuecomment-653913805 and your LGTM on change release process https://github.com/harfbuzz/harfbuzz/pull/2545/files meant. I don't like link hounding or anything, I see you points and can agree with them and like you to also see my reasoning also.

NEWS
```diff
@@ -1,3 +1,11 @@
+Overview of changes leading to 2.7.1
+Wednesday, Aug 12, 2020
+====================================
+- ot-funcs now handles variable empty glyphs better when hvar/vvar isn't present.
+- Reverted a shaping regression for fonts which lack GDEF.
```
khaledhosny, line 5: The regression happens to fonts with GDEF, the issue that caused the regression was a fix for fonts lacking GDEF table.

ebraminio: Oh correct, so does `Reverted a GDEF processing regression.` sound better?

khaledhosny:
> > if we can lower the requires meson version a bit
> 
> We can but gtk also requires 0.53.0 https://gitlab.gnome.org/GNOME/gtk/blob/master/meson.build#L10 for the same reason as us, `summary`, so packagers know they are building something intended, is there something that different about `gtk` not applies here? I wanted to hear such request from some packagers first to see what I can do about.

That is GTK 4, a still in beta major new version of the library. GTK 3 still has autotools and its meson build requires 0.48.0.

ebraminio:
> That is GTK 4, a still in beta major new version of the library. GTK 3 still has autotools and its meson build requires 0.48.0.

Correct. Can you give me which version should be adopted? I can make it work I think as what we 0.48.0 initially.

ebraminio:
https://github.com/harfbuzz/harfbuzz/pull/2641 0.49.0 compatibility, I can try more so have a look at https://repology.org/badge/vertical-allrepos/meson.svg and let me know which version you prefer. Thanks!

ebraminio:
2.7.1 is released with meson supporting from 0.47.0 onward (including `summary` polyfill there), each minor version of meson is tested manually but 0.47.0 itself is tested on our CI so we won't lose its support in future, the autotools release is still there and its dist is also uploaded, one can use meson on top of that dist or better to use GitHub generated tarballs as suggested in 2.7.0.

Thanks everyone for their collaboration and feedback :)

PR #2633, opened 2020-08-11
Revive cmake

ebraminio:
Quoting Behdad,

> Can we get 2.7.1 out WITHOUT removing cmake?

However I am not sure which one is referred to, cmake-config or cmake build itself. I see cmake-config more risky but cmake removal itself less problematic.


behdad:
What I want is this: Just revert any build system changes. Restore sanity, get release out, let the dust settle. Then we discuss how to run the project moving forward. Thank you.

ebraminio:
Done

PR #2635, opened 2020-08-11
[ENOMEM] skip asserts in to_bias if serializer is in an error state.

garretrieger:


ebraminio:
Ok, traced this a bit to see what happen if bias is zero and reached to `hb_serialize_context_t::add_link` and saw that simply returns if is `in_error`, have a feeling that we should look back a minute to see what we should be done to make whole thing subset more idiomatic to rest of the project as spreading `in_error` check doesn't give the best filling still let's have this in. Thanks!

garretrieger:
> Ok, traced this a bit to see what happen if bias is zero and reached to `hb_serialize_context_t::add_link` and saw that simply returns if is `in_error`, have a feeling that we should look back a minute to see what we should be done to make whole thing subset more idiomatic to rest of the project as spreading `in_error` check doesn't give the best filling still let's have this in. Thanks!

For subset the approach we take with the serializer is:
- Once the serializer is in an error state the result will be tossed at the end of the subset operation.
- If the serializer is in an error state then the subsetting code doesn't have to do anything reasonable or correct as long as it doesn't segfault or trip assertions.
- Therefore calling any methods on serializer once it's in error should have no effect, since the output's going to be tossed.

This scheme allows us to avoid having to constantly check if the serializer is in error throughout the subsetting code.

ebraminio:
Sounds good. Thanks for the explanation

PR #2639, opened 2020-08-12
[ENOMEM] Return gracefully if stages isn't initialized correctly

ebraminio:
This happens if calls to ['m.lookups[table_index].push ()'](https://github.com/harfbuzz/harfbuzz/blob/b2628f1a3e3a6875dcdf7cb89f5b10a4ae76e6f9/src/hb-ot-map.cc#L127) has been
silently failed due to lack of memory.

This change just returns gracefully instead issuing the assert.

Fixes https://crbug.com/oss-fuzz/24494

Not claiming is the best solution just that I haven't found other chance to check the failure around.

How to reproduce?

`CXXFLAGS="-fsanitize=address,fuzzer-no-link" meson fuzzbuild --default-library=static -Dfuzzer_ldflags="-fsanitize=address,fuzzer" -Dexperimental_api=true && ninja -C fuzzbuild test/fuzzing/hb-{shape,draw,subset,set}-fuzzer && fuzzbuild/test/fuzzing/hb-shape-fuzze test/api/fonts/clusterfuzz-testcase-minimized-hb-shape-fuzzer-5753845452636160`


src/hb-ot-map.hh
```diff
@@ -140,12 +140,12 @@ struct hb_ot_map_t
   void get_stage_lookups (unsigned int table_index, unsigned int stage,
 			  const struct lookup_map_t **plookups, unsigned int *lookup_count) const
   {
-    if (unlikely (stage == UINT_MAX)) {
+    if (unlikely ((stage == UINT_MAX) || (stage > stages[table_index].length)))
```
behdad, line 143: Just remove the first check, as is covered by the second. Thanks. LGTM otherwise. Please merge.

behdad:
Thank you Ebrahim. Good to see you around again.

PR #2640, opened 2020-08-12
[ENOMEM] Check whether serialize context isn't in error

ebraminio:
Happens if serialize is in_error due to OOM and we are executing copy.

Fixes https://crbug.com/oss-fuzz/24780

@garretrieger I didn't know parts of subset is used in shaper also… shouldn't this happen a bit earlier thus affecting `ret` anyway? Please have a look. Thanks!

How to reproduce?

`CXXFLAGS="-fsanitize=address,fuzzer-no-link" meson fuzzbuild --default-library=static -Dfuzzer_ldflags="-fsanitize=address,fuzzer" -Dexperimental_api=true && ninja -C fuzzbuild test/fuzzing/hb-{shape,draw,subset,set}-fuzzer && fuzzbuild/test/fuzzing/hb-shape-fuzze test/fuzzing/fonts/clusterfuzz-testcase-minimized-hb-shape-fuzzer-5138182937772032`


src/hb-ot-shape-complex-arabic-fallback.hh
```diff
@@ -170,7 +170,7 @@ arabic_fallback_synthesize_lookup_ligature (const hb_ot_shape_plan_t *plan HB_UN
   c.end_serialize ();
   /* TODO sanitize the results? */
 
-  return ret ? c.copy<OT::SubstLookup> () : nullptr;
```
ebraminio, line 173: The one that fuzzer has discovered is this one but the other has the same code so anything happens here should happen this also most likely.

garretrieger:
So in general with the serializer, whenever you attempt to pull data out of it you need to check in_error(). I talked about this in the other PR, the way the serializer is intended to be used:

- start serialize
- Do serialization operations. Don't need to check for in_error() here. Operations are no-ops if the serializer is in_error.
- end serialize
- Check if in_error() is set, if so toss the result.
- otherwise, pull data out of the serializer.

ebraminio:
Oh, sounds nice, thank you, does it look good now?

garretrieger:
> Oh, sounds nice, thank you, does it look good now or should I keep checking ret?

I think the safer option is to continue checking return value in addition to in_error(). We have a goal to remove the return values from all of the serialize calls and rely only on context.in_error(), but until that's done we should probably continue to check the return value as well.

ebraminio:
> I think the safer option is to continue checking return value in addition to in_error(). We have a goal to remove the return values from all of the serialize calls and rely only on context.in_error(), but until that's done we should probably continue to check the return value as well.

Done. Do you think it looks now? Thanks

PR #2642, opened 2020-08-12
[ENOMEM] Fix subsetting timeout

garretrieger:
Timeout caused by visited lookups not working during closure after a failed allocation. https://oss-fuzz.com/testcase-detail/5741735372914688


ebraminio:
Nice to put oss-fuzz link also, does the second one has file associated or just spotted? Thanks!

garretrieger:
Added link to the test case to the comment above. The second one has no associated fuzz file I just spotted that while fixing the original issue.

PR #2648, opened 2020-08-19
Fix get_baseline when `baseline_tag` does not exist

kojiishi:
This patch fixes `Axis::get_baseline` to return `false` if the specified `baseline_tag` does not exist.

### DIFF
--- a/src/hb-ot-layout-base-table.hh
+++ b/src/hb-ot-layout-base-table.hh
@@ -379,12 +379,20 @@ struct Axis
 		     const BaseCoord **coord) const
   {
     const BaseScript &base_script = (this+baseScriptList).get_base_script (script_tag);
-    if (!base_script.has_data ()) return false;
+    if (!base_script.has_data ())
+    {
+      *coord = nullptr;
+      return false;
+    }
 
     if (likely (coord))
     {
       unsigned int tag_index = 0;
-      if (!(this+baseTagList).bfind (baseline_tag, &tag_index)) return false;
+      if (!(this+baseTagList).bfind (baseline_tag, &tag_index))
+      {
+        *coord = nullptr;
+        return false;
+      }
       *coord = &base_script.get_base_coord (tag_index);
     }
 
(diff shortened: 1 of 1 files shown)


kojiishi:
@ebraminio Can you take a look? It looks like you know this function.

ebraminio:
lgtm and happy to know this is going to be used somewhere.

kojiishi:
Thank you. I'm still experimenting, and hit this issue.

Sorry I have zero knowledge on hb test infra. I can add to Blink when this becomes real, but it'll take a bit more.

Also can you advice what msys2/MINGW32 and codecov failures are? The former does not seem to related with the patch, and no idea what the latter is.

drott:
Koji tells me this happens with the `BASE` table of YuGothic, see: https://gist.github.com/drott/fe857fc0d91833de03e3ab67f16c5c15

kojiishi:
Thank you @drott for the help to explain the case. In his example, YuGothic has:
```
<BaseTagList>
  <!-- BaseTagCount=4 -->
  <BaselineTag index="0" value="icfb"/>
  <BaselineTag index="1" value="icft"/>
  <BaselineTag index="2" value="ideo"/>
  <BaselineTag index="3" value="romn"/>
</BaseTagList>
```
This patch fixes when there's a table for the specified script, but the `BaselineTag` is missing; e.g.:
```
hb_ot_layout_get_baseline(font, HB_TAG('i','d','t','p'), HB_DIRECTION_LTR, HB_OT_TAG_DEFAULT_SCRIPT, HB_OT_TAG_DEFAULT_SCRIPT, &coord);
```
returns `true`, and the `coord` value is taken from the first entry (`icfb` in this case.)

kojiishi:
@ebraminio gentle ping, any chance this gets fixed?

ebraminio:
lgtm

kojiishi:
@ebraminio:
> lgtm

Thank you, but I don't have the right to merge this, nor I'm not familiar with the process in this repository. Can you advise what to do next?

behdad:
Thanks. I think we want to set `*coord = 0` in that case.

kojiishi:
> Thanks. I think we want to set `*coord = 0` in that case.

Thank you, done.

behdad:
Thanks. LGTM.

PR #2652, opened 2020-08-21
[USE] Update to the 2020-08-13 USE specification

dscorbett:
This updates the USE shaper to match the USE spec, last updated August 13, and the USE supplementary data [as of August 18](https://github.com/microsoft/font-tools/tree/78b2134fdca84df41a931d30d5c48fec115eca41/USE).

The changes for Egyptian hieroglyphs are in a separate commit in case you don’t want to merge it. It is ambiguous whether USE supports them yet. The spec implies they are supported but the relevant overrides are commented out. I figure that they were probably meant to be supported.

I did not switch joining scripts from the Arabic shaper to USE. USE’s treatment of topographical features is underspecified and its InSC overrides for joining scripts are incomplete.


dscorbett:
The MinGW builds are failing.
>```console
>D:/a/_temp/msys/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/10.2.0/../../../../i686-w64-mingw32/bin/ld.exe: src/libharfbuzz-0.dll.p/hb-directwrite.cc.obj:hb-directwrite.cc:(.rdata$_ZTV25IDWriteTextAnalysisSource[__ZTV25IDWriteTextAnalysisSource]+0x1c): more undefined references to `__cxa_pure_virtual' follow
>collect2.exe: error: ld returned 1 exit status
>```
Is that related to my changes?

khaledhosny:
> The MinGW builds are failing.

It should be fixed if you rebase on master.

behdad:
Thanks. I didn't do a full review of why you added Arabic joining in but I trust you to know best. Please merge when you feel it's ready.

dscorbett:
Thank you for trusting me; I’ll explain it anyway. gen-use-table.py now consumes ArabicJoining.txt because the new USE spec uses Joining_Type in the definition of the class BASE. It’s not useful yet, but it will be once joining scripts are routed to USE, which I hope I’ll get to soon.

behdad:
Got it. Thanks.

PR #2655, opened 2020-08-25
Glyph class tests

simoncozens:
Tests for two glyph class bugs. These tests may not succeed, but that's OK! Failing tests tell us where there are problems in the codebase...


behdad:
Thanks. Maybe push the one that passes to master?

We need to add a way to mark tests as expected-fail in our test-suite.

ebraminio:
Guess we can add the tests but commented.

simoncozens:
Actually, I wanted to deliberately put a failing test in the CI because there is a bug that we should get around to fixing...

ebraminio:
Understandable, just that it makes it not merge-able as it breaks things more that needed. If all the tests of one file are skipped the whole test will be marked as skipped at least which is similar to what requested "expected fail", not that different from what Chromium project has for example.

khaledhosny:
Closing for #3365

PR #2657, opened 2020-08-27
[ENOMEM] fix access to unitialized memory.

garretrieger:
If the serialize() call fails to write the object then we can't safely read varstore_prime fields. Fixes https://oss-fuzz.com/testcase-detail/5137462782066688.


behdad:
The explicit `bool` cast is not necessary. Lgtm otherwise.

PR #2668, opened 2020-09-03
meson: add hb-version.h to dependencies correctly

### LINKED ISSUE #2667: Parallel build race
A build failed with mysterious errors which suggest that there is a build race:

```
../harfbuzz-2.7.1/src/hb.h:44:10: warning: ../harfbuzz-2.7.1/src/hb-version.h is shorter than expected
   44 | #include "hb-version.h"
      |          ^~~~~~~~~~~~~~
../harfbuzz-2.7.1/src/hb-common.cc:625:1: error: no previous declaration for 'void hb_version(unsigned int*, unsigned int*, unsigned int*)' [-Werror=missing-declarations]
  625 | hb_version (unsigned int *major,
      | ^~~~~~~~~~
../harfbuzz-2.7.1/src/hb-common.cc: In function 'void hb_version(unsigned int*, unsigned int*, unsigned int*)':
../harfbuzz-2.7.1/src/hb-common.cc:629:12: error: 'HB_VERSION_MAJOR' was not declared in this scope
  629 |   *major = HB_VERSION_MAJOR;
      |            ^~~~~~~~~~~~~~~~
../harfbuzz-2.7.1/src/hb-common.cc:630:12: error: 'HB_VERSION_MINOR' was not declared in this scope
  630 |   *minor = HB_VERSION_MINOR;
      |            ^~~~~~~~~~~~~~~~
../harfbuzz-2.7.1/src/hb-common.cc:631:12: error: 'HB_VERSION_MICRO' was not declared in this scope
  631 |   *micro = HB_VERSION_MICRO;
      |            ^~~~~~~~~~~~~~~~
```

My guess is that `hb-version.h` is being repeatedly written.

rossburton:
We need to make sure that the generator for hb-version.h is tracked as a
dependency, otherwise it can be re-generated too late causing compile
errors.

Closes #2667

### DIFF
--- a/src/gen-hb-version.py
+++ b/src/gen-hb-version.py
@@ -4,15 +4,15 @@
 
 import os, sys, shutil
 
-if len (sys.argv) < 5:
+if len (sys.argv) < 4:
 	sys.exit(__doc__)
 
 version = sys.argv[1]
 major, minor, micro = version.split (".")
 
 OUTPUT = sys.argv[2]
-CURRENT_SOURCE_DIR = sys.argv[3]
-INPUT = sys.argv[4]
+INPUT = sys.argv[3]
+CURRENT_SOURCE_DIR = os.path.dirname(INPUT)
 
 with open (INPUT, "r", encoding='utf-8') as template:
 	with open (OUTPUT, "wb") as output:
--- a/src/meson.build
+++ b/src/meson.build
@@ -1,3 +1,10 @@
+hb_version_h = configure_file(
+               command: [find_program('gen-hb-version.py'), meson.project_version(), '@OUTPUT@', '@INPUT@'],
+               input: 'hb-version.h.in',
+               output: 'hb-version.h',
+               install: true,
+               install_dir: join_paths(get_option('includedir'), meson.project_name()))
+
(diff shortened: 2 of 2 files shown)


khaledhosny:
This is breaking almost all CI builds.

rossburton:
Yes, I'm well aware it is breaking.  Working on it, but I can't fire CI without sending a PR.

khaledhosny:
`src/hb-version.h` is kept in the source tree intentionally even though the file is generated. Some major HarfBuzz consumers don’t use our build system and want to be able to build directly from git tree at the same time, so all generated source files are kept under version control.

rossburton:
That MSVC failure happens in master so I believe this actually passes CI now.

khaledhosny:
I’m seeing this warning:
```
 * 0.50.0: {'install arg in configure_file'}
```

We would really like to keep support for older meson versions as much as possible.

dcbaker:
That warning is a bit misleading (I added it). The `install` argument is *honored* in 0.50, before that it was ignored and the file would always be installed if `install_dir` is set.

khaledhosny:
I see, thanks.

PR #2669, opened 2020-09-04
Disambiguate OT language system tags when the primary tag is not first

dscorbett:
'TWI ' has a single primary BCP 47 candidate ("tw"), so gen-tag-table.py treated it as unambiguous. However, 'TWI ' also has the non-primary candidate "ak", which alphabetically precedes it, so 'TWI ' was being converted to "ak". That is wrong.

'TWI ' is now converted to "tw". Similarly, 'HYE ' and 'TOD ' are now converted to "hyw" and "xwo". It is a bit odd to convert 'HYE ' (Armenian) to "hyw" (Western Armenian) instead of the more likely "hy" (Armenian, implicitly Eastern Armenian), but this is necessary to make the mapping bijective.


behdad:
Lgtm. Thanks

PR #2676, opened 2020-09-09
[subset] Fix integer underflow in ContextFormat2.

garretrieger:


garretrieger:
Fixes: https://oss-fuzz.com/testcase-detail/5689082504806400

behdad:
lgtm. This should all be done without manual index tracking. But lgtm for the code as is.

PR #2677, opened 2020-09-09
Demote unregistered vendor-specific language tags

dscorbett:
Abyssinica SIL 2.000 does not use the unregistered language system tags 'GUK ' and 'SGW '. I think it is safe to remove support for them. I’d like to remove support for 'KHN ' too, but I don’t know what Microsoft fonts are supposed to have used it, so I can’t verify that it would be safe. Instead, I demoted it below the correct tag, 'KHT '. (This will only matter for a font supporting both tags but enabling different lookups for each, which is unlikely.)


behdad:
I trust you on these. Thanks.

PR #2678, opened 2020-09-11
Post build artifacts from CI jobs

### LINKED ISSUE #2663: [ci] Upload win32 build artifacts
For tagged builds, we want to upload the win32 build artifacts to the releases page. Latest build lacks them because I don’t have a working MingW setup, but CI is building them already so we should make use of that.

alerque:
Closes #2663.

...and more besides — in addition to posting the prebuilt Windows artifacts (`harfbuzz-win32-<version>.zip`) this also posts the autotools generated dist file (`harfbuz-<version>.tar.xz`).

See [this caveat](https://github.com/harfbuzz/harfbuzz/issues/2663#issuecomment-691089335) about the API token I currently configured in Circle CI.

### DIFF
--- a/.ci/publish_release_artifact.sh
+++ b/.ci/publish_release_artifact.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+set -e
+set -o pipefail
+
+if [[ -z $GITHUB_TOKEN ]]; then
+	echo "No GITHUB_TOKEN secret found, artifact publishing skipped"
+	exit
+fi
+
+mkdir -p $HOME/.local/bin
+export _GHR_VER=v0.13.0
+export _GHR=ghr_${_GHR_VER}_linux_amd64
+curl -sfL https://github.com/tcnksm/ghr/releases/download/$_GHR_VER/$_GHR.tar.gz |
+	tar xz -C $HOME/.local/bin --strip-components=1 $_GHR/ghr
+
+ghr -replace \
+	-u $CIRCLE_PROJECT_USERNAME \
+	-r $CIRCLE_PROJECT_REPONAME \
+	$CIRCLE_TAG \
+	$1
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,9 @@ executors:
   win32-executor:
     docker:
       - image: cimg/base:edge-20.04
+  autotools-executor:
+    docker:
+      - image: cimg/base:edge-20.04
 
 jobs:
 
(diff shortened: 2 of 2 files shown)


alerque:
If anybody really wants to dig into whether this works, the live Harfbuzz repo is not a great place to try it on for size (since it only runs on valid semver tags!). You can however look at [the branch on my fork I was testing with](https://github.com/alerque/harfbuzz/commits/win32-artifacts) (not the same as this PR branch), the [corresponding ci job history](https://app.circleci.com/pipelines/github/alerque/harfbuzz), and [releases](https://github.com/alerque/harfbuzz/releases). The simulated 2.7.3 tag there for example was has automatic artifacts, and you can [see the build outline](https://app.circleci.com/pipelines/github/alerque/harfbuzz/52/workflows/1f8cd938-9aea-4dc8-aea7-060c6054e733) that did it.

behdad:
Happy to try. Don't really know what to review.

ebraminio:
Feel free to consider the fact that our CircleCI setup is a paid service so in long run guess we like to rely free services, GitHub Actions in here, instead that, so what I suggest is to migrate the bot to GitHub Actions while doing this cool thing. Thanks!

ebraminio:
That's just a suggestion based on the fact GitHub Actions based solution will need different than what you've proposed for Circle CI yet simpler solution, see https://github.com/actions/upload-release-asset#example-workflow---upload-a-release-asset for example, however as you wish.

alerque:
Yes the equivalent artifact posting in GitHub Actions would be a bit simpler –and something I've setup many times– but porting the entire build workflow to Actions is not a simple change and will require quite a bit of fiddling and comparing outputs to confirm that it's testing the same things across all platforms. And this work is done. Again I can help with the port later, but porting the build to a different CI platform is out of scope for this PR or the issue it was made to solve.

ebraminio:
> [..] but porting the entire build workflow to Actions is not a simple change and will require quite a bit of fiddling and comparing outputs to confirm that it's testing the same things across all platforms [..]

Oooh, not sure why you are on that feeling that as we don't have complex multilevel workflow as some other project as migrating the two simple bots to GA first in a separate PR add less complexity as for example no need to add GITHUB_TOKEN to Circle CI bot then removing it while the migration as GA already has one.

alerque:
> no need to add GITHUB_TOKEN to Circle CI bot

This is already done. There is nothing left for anybody to do, everything is here and in place. Only somebody pushing merge is left. The result will be a better system that it was before this PR.

Future improvements can of course be made. Migrating to a different CI has a very real potential to introduce unexpected issues (as the build system migration to meson has) that take time to work out the wrinkles and be confident everything is covered. I don't see any reason to hold up this PR that makes concrete improvements to the existing CI just to discard it because a possible later migration to a different CI platform is in the cards. Even if we do migrate, this should be the benchmark by which to compare equivalence when migrating.

ebraminio:
> Migrating to a different CI has a very real potential to introduce unexpected issues

Correct, experienced it while migrating appveyor msvc to GA, didn't expect this can be that complicated but of course can't make guarantee.

> I don't see any reason to hold up this

Just the big picture that while that we will need another round of testings and so, but seems this is what you insist and didn't oppose from the beginning.

khaledhosny (approved):

PR #2682, opened 2020-09-15
[ENOMEM] don't perform set process operations if the other set is in …

garretrieger:
…an error state.

Running a process while the other set is in an error state can potentially corrupt this sets map map (for example by overwritting all of the major values with 0).

Fixes https://oss-fuzz.com/testcase-detail/5345734743031808


alerque (approved):

behdad:
I'm actually not sure if this is a good idea.

rsheeter:
Do you suggest rollback?

behdad:
The way I envisioned objects in HarfBuzz is that the ones that track their own success, will *always* be in a valid state. So whatever it is that you are seeing should be fixed by figuring out what's being violated in that invariant.

behdad:
> Do you suggest rollback?

Yes.

garretrieger:
Alright I can take a look at the implementation of process() and see if there's another way to safe handled error'd out sets. The problem is that process() uses the internal state of the other set (page_map and pages) rather than the public interface. Currently an in error set is safe to be accessed via the public interface, but accessing the internal structures is potentially going to cause things to go wrong.

behdad:
> ...but accessing the internal structures is potentially going to cause things to go wrong.

It shouldn't. Instead of looking into process(), please look into why the internal structures are inconsistent. They shouldn't be. Thanks.

PR #2685, opened 2020-09-16
[ENOMEM] fix cases where set internals can get corrupted from alloc failures.

garretrieger:
Fixes: https://oss-fuzz.com/testcase-detail/5345734743031808


garretrieger:
So there's two issues:
- first with clear(): If there was previously an alloc failure during a resize() call you end up in a state where 'page_map' is not in error, but 'pages' is in_error. When clear() is called it resizes page_map to 0, but since pages is in_error it does not get resized. Then if good_set.process(failed_set) is called since it uses failed_set.pages.length, process ends up trying to access page_map entries at indices that no longer exist which results in in process() breaking the page map on good_set. 
- Second, with compaction we first rewrite all of the page map entries into the front of the page map array and then compact() is used to move all retains pages to the front of pages array and fix the page map pointers. compact() internally uses an allocated vector to store a mapping, if that fails to allocate then compact() call will break the page map. So I changed compact() to bail out and put the set into error on alloc failure. I also moved the allocation of the vector up to before we rewrite the page map entries into the front of the page map array so that when we bail out after the failure page_map isn't left in an invalid state.

src/hb-set.hh
```diff
@@ -266,9 +266,9 @@ struct hb_set_t
   {
     if (unlikely (hb_object_is_immutable (this)))
       return;
-    population = 0;
-    page_map.resize (0);
-    pages.resize (0);
+
+    if (resize (0))
+      population = 0;
```
behdad, line 271: Please make clear() always succeed (even if set is in error state), like it used to.

behdad:
Thanks. I finally understand this. :) LGTM w minor comments.

behdad:
A couple of comments on the `compact()` since you are touching it already:
```
 void compact (unsigned int length) 
  { 
    hb_vector_t<uint32_t> old_index_to_page_map_index; 
    old_index_to_page_map_index.resize(pages.length); 
    for (uint32_t i = 0; i < old_index_to_page_map_index.length; i++) 
      old_index_to_page_map_index[i] = 0xFFFFFFFF; 
 
    for (uint32_t i = 0; i < length; i++) 
      old_index_to_page_map_index[page_map[i].index] =  i; 
 
    compact_pages (old_index_to_page_map_index); 
  } 
```

The `for` loop index should simply use `unsigned`, not `uint32_t`.

Also, both those loops we can use the iterator machinary:
1. First loop is probably: `hb_fill (old_index_to_page_map_index, 0xFFFFFFFF)`,
2. Second loop is probably: `hb_iota () | hb_sink (old_index_to_page_map_index)`.

PR #2686, opened 2020-09-17
More tracing

simoncozens:
Fixes #2683

```
$ ./util/hb-shape -V dalekh_si.ttf '<1a2f><1a6e><1a6c><1a65><1a41>'
trace: begin preprocess        buffer: [U+1A2F|U+1A6E|U+1A6C|U+1A65|U+1A41]
trace: end preprocess   buffer: [U+1A2F|U+1A6E|U+1A6C|U+1A65|U+1A41]
trace: begin normalize  buffer: [U+1A2F|U+1A6E|U+1A6C|U+1A65|U+1A41]
trace: end normalize    buffer: [U+1A2F|U+1A6E|U+1A6C|U+1A65|U+1A41]
trace: start table GSUB buffer: [uni1A2F=0|uni1A6E=0|uni1A6C=0|uni1A65=0|uni1A41=4]
...
trace: start reordering USE     buffer: [uni1A2F=0|uni1A6E=0|.notdef=0|uni1A6C=0|uni1A65=0|uni1A41=4]
trace: end reordering USE       buffer: [uni1A6E=0|uni1A2F=0|.notdef=0|uni1A6C=0|uni25CC=0|uni1A65=0|uni1A41=4]
```

Issue #2468, opened 2020-06-12
When tracing, sanitize buffer between lookup applications


simoncozens:
Given a font with the following feature code:

```
lookup lookup1 { sub uni0061 by uni0062; } lookup1;
lookup lookup2 {
    sub uni0061' lookup lookup1;
    sub uni0062' by uni0061;
} lookup2;

feature rlig { sub uni0061' lookup lookup2 uni0062' lookup lookup2; } rlig;
```

running `hb-shape -V test.otf abc` returns:

```
trace: start GSUB stage	buffer: [uni0061=0|uni0062=1]
trace: start lookup 3	buffer: [uni0061=0|uni0062=1]
trace: end lookup 3	buffer: [uni0062=0|uni0061=1]
[uni0062=0+600|uni0061=1+600]
```

Reporting on the application of lookups 1 and 2 would also be useful for debugging.

simoncozens:
I don't know if it's the *cleanest* approach, but this patch does what I want:

```patch
diff --git a/src/hb-ot-layout-gpos-table.hh b/src/hb-ot-layout-gpos-table.hh
index 5fd848fe..b23b9177 100644
--- a/src/hb-ot-layout-gpos-table.hh
+++ b/src/hb-ot-layout-gpos-table.hh
@@ -2709,9 +2709,11 @@ template <typename context_t>
   const PosLookup &l = c->face->table.GPOS.get_relaxed ()->table->get_lookup (lookup_index);
   unsigned int saved_lookup_props = c->lookup_props;
   unsigned int saved_lookup_index = c->lookup_index;
+  if (!c->buffer->message (c->font, "start lookup %d", lookup_index)) return false;
   c->set_lookup_index (lookup_index);
   c->set_lookup_props (l.get_props ());
   bool ret = l.dispatch (c);
+  (void) c->buffer->message (c->font, "end lookup %d", lookup_index);
   c->set_lookup_index (saved_lookup_index);
   c->set_lookup_props (saved_lookup_props);
   return ret;
diff --git a/src/hb-ot-layout-gsub-table.hh b/src/hb-ot-layout-gsub-table.hh
index 612720e6..ff64c5d6 100644
--- a/src/hb-ot-layout-gsub-table.hh
+++ b/src/hb-ot-layout-gsub-table.hh
@@ -1587,9 +1587,11 @@ template <typename context_t>
[...]

PR #2687, opened 2020-09-17
Serialize Unicode buffers

simoncozens:
This adds support in hb-buffer-serialize for serializing Unicode buffers as well as glyph buffers, including a new function `hb_buffer_serialize` which chooses between `hb_buffer_serialize_glyphs` and `hb_buffer_serialize_unicode` depending on the content of the buffer.

We also use this in the utils serialization, which both simplifies that code and allows for tracing of Unicode buffers from inside Harfbuzz.

Please check my implementation of `_hb_buffer_serialize_unicode_text` carefully for potential bad string access bugs.


behdad:
Thanks!

src/hb-buffer-serialize.cc
```diff
@@ -379,6 +449,117 @@ hb_buffer_serialize_glyphs (hb_buffer_t *buffer,
   }
 }
 
+unsigned int
```
khaledhosny, line 452: This is public API and needs documentation.

src/hb-buffer-serialize.cc
```diff
@@ -379,6 +449,117 @@ hb_buffer_serialize_glyphs (hb_buffer_t *buffer,
...
+ * @format: the #hb_buffer_serialize_format_t to use for formatting the output.
+ * @flags: the #hb_buffer_serialize_flags_t that control what glyph properties
+ *         to serialize.
+ *
+ * Serializes @buffer into a textual representation of its content, whether
+ * Unicode codepoints or glyph identifiers and positioning information. This is
+ * useful for showing the contents of the buffer, for example during debugging.
+ * There are currently two supported serialization formats:
+ *
+ * ## text
```
khaledhosny, line 512: The description if the format should not be repeated here, it should instead refer to the documentation of the other two functions.

util/options.cc
```diff
@@ -910,47 +910,36 @@ format_options_t::add_options (option_parser_t *parser)
...
+format_options_t::serialize (hb_buffer_t *buffer,
 				    hb_font_t   *font,
 				    hb_buffer_serialize_format_t output_format,
 				    hb_buffer_serialize_flags_t flags,
 				    GString     *gs)
 {
-  g_string_append_c (gs, '[');
+  if (hb_buffer_get_content_type(buffer) == HB_BUFFER_CONTENT_TYPE_UNICODE)
+    g_string_append_c (gs, '<');
+  else
```
khaledhosny, line 921: We need tests for the new code, should be somewhere under `tests/api`.

util/options.hh
```diff
@@ -635,9 +635,7 @@ struct format_options_t : option_group_t
 
   void add_options (option_parser_t *parser) override;
 
-  void serialize_unicode (hb_buffer_t  *buffer,
-			  GString      *gs);
-  void serialize_glyphs (hb_buffer_t  *buffer,
+  void serialize (hb_buffer_t  *buffer,
 			 hb_font_t    *font,
```
khaledhosny, line 639: Please fix the indentation after the name change.

khaledhosny (changes requested):
Looks good, thanks!

Do we also need the equivalent of `hb_buffer_deserialize_glyphs ()`? Might not be that useful public buffer API can be used to create unicode buffers, unlike glyph buffers.

behdad:
So about the API:

- In theory, we can just add `hb_buffer_serialize()` and deprecated `hb_buffer_serialize_glyphs()`. Or keep `serialize_glyphs()` and add `serialize_unicode()` as you have and have them just assert content type and call `serialize()`.

- The existing documentation and your new one both mention that serialized output is delimited in `[/]`. That doesn't seem to be the case and is only added in util. So, docu needs update. Which is good, because for Unicode I like the `<U+1234,U+5678>` syntax more, and if buffer is empty and of content-type UNKNOWN, there's no ambiguity without the delimiters.

- For deserialize API, I think it would be nice to have a `hb_buffer_deserialize_unicode()` as well.

src/hb-buffer-serialize.cc
```diff
@@ -196,6 +196,47 @@ _hb_buffer_serialize_glyphs_json (hb_buffer_t *buffer,
...
+	{
+		char b[1024];
+		char *p = b;
+		if (i)
+			*p++ = ',';
+
+		*p++ = '{';
+
+		APPEND ("\"u\":");
+		p += hb_max (0, snprintf (p, ARRAY_LENGTH (b) - (p - b), "%u", info[i].codepoint));
```
behdad, line 220: Does JSON accept hex integers? I think we should use those if doable.

src/hb-buffer-serialize.cc
```diff
@@ -282,6 +323,35 @@ _hb_buffer_serialize_glyphs_text (hb_buffer_t *buffer,
...
+	char *p = buf;
+	/* We can use a much simpler implementation for this and copy text
+	 * directly into the output buffer. */
+	for (unsigned int i = start; i < end; i++)
+	{
+		if (i) {
+			*p++ = '|';
+		}
+
+		p += hb_max (0, snprintf (p, buf_size - *buf_consumed, "U+%X", info[i].codepoint));
```
behdad, line 346: Correct way to print Unicode codepoints is `U+%04X`.

src/hb-buffer-serialize.cc
```diff
@@ -282,6 +323,35 @@ _hb_buffer_serialize_glyphs_text (hb_buffer_t *buffer,
...
+{
+	hb_glyph_info_t *info = hb_buffer_get_glyph_infos (buffer, nullptr);
+	*buf_consumed = 0;
+	char *p = buf;
+	/* We can use a much simpler implementation for this and copy text
+	 * directly into the output buffer. */
+	for (unsigned int i = start; i < end; i++)
+	{
+		if (i) {
+			*p++ = '|';
```
behdad, line 343: Use `,` please, not `|`. Unless you have a strong preference?

behdad: I'm actually not fully sold on that. Can be talked into liking `|`.

behdad: Yeah I think I like `|`. Keep it. Thanks. :)

src/hb-buffer-serialize.cc
```diff
@@ -282,6 +323,35 @@ _hb_buffer_serialize_glyphs_text (hb_buffer_t *buffer,
...
+					unsigned int start,
+					unsigned int end,
+					char *buf,
+					unsigned int buf_size,
+					unsigned int *buf_consumed)
+{
+	hb_glyph_info_t *info = hb_buffer_get_glyph_infos (buffer, nullptr);
+	*buf_consumed = 0;
+	char *p = buf;
+	/* We can use a much simpler implementation for this and copy text
```
behdad, line 338: Not sure what this comment means or why it's retained. :)

src/hb-buffer-serialize.cc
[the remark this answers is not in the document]
behdad: Yeah seems like the code was written with the aux-buffer style of other functions, but the aux buffer removed to write directly to output buffer without adding proper checks.

src/hb-buffer-serialize.cc
simoncozens: No, sadly not.

src/hb-buffer-serialize.cc
```diff
@@ -196,6 +196,47 @@ _hb_buffer_serialize_glyphs_json (hb_buffer_t *buffer,
...
 
+static unsigned int
+_hb_buffer_serialize_unicode_json (hb_buffer_t *buffer,
+					unsigned int start,
+					unsigned int end,
+					char *buf,
+					unsigned int buf_size,
+					unsigned int *buf_consumed)
+{
+	hb_glyph_info_t *info = hb_buffer_get_glyph_infos (buffer, nullptr);
```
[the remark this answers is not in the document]
simoncozens: This is because in the .editorconfig, there is the following line:

```
indent_style = tab # should be space
```

Maybe that should be "space"? :-)

simoncozens:
> The existing documentation and your new one both mention that serialized output is delimited in [/]. That doesn't seem to be the case and is only added in util. So, docu needs update.

Actually I don't like this. Currently `hb_serialize_glyphs( ... HB_BUFFER_SERIALIZE_FORMAT_JSON)` doesn't add the `[]` delimiters, so produces invalid JSON; that seems wrong for a function which promises to return JSON!

So I'm going to suggest that I move the delimiters from utils to `hb-buffer-serialize`. This is obviously a breaking change, but the previous behaviour was arguably a bug, and I think we probably control everywhere that uses `HB_BUFFER_SERIALIZE_FORMAT_JSON` (harfbuzz and harfbuzzjs).

src/hb-buffer-serialize.cc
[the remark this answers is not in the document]
khaledhosny: I have no idea, I don’t use an editor that understands this config file.

src/hb-buffer-serialize.cc
```diff
@@ -91,26 +91,26 @@ hb_buffer_serialize_format_to_string (hb_buffer_serialize_format_t format)
...
-				  unsigned int *buf_consumed,
-				  hb_font_t *font,
-				  hb_buffer_serialize_flags_t flags)
+                                  unsigned int start,
+                                  unsigned int end,
+                                  char *buf,
+                                  unsigned int buf_size,
+                                  unsigned int *buf_consumed,
+                                  hb_font_t *font,
+                                  hb_buffer_serialize_flags_t flags)
```
khaledhosny, line 109: There is now a bunch of unrelated spacing changes. I have to admit, I never fully understood HarfBuzz indentation style and I probably mislead you.

src/hb-buffer-serialize.cc
```diff
@@ -379,6 +487,131 @@ hb_buffer_serialize_glyphs (hb_buffer_t *buffer,
...
+ * Since: 2.7.3
+ **/
+unsigned int
+hb_buffer_serialize_unicode (hb_buffer_t *buffer,
+                             unsigned int start,
+                             unsigned int end,
+                             char *buf,
+                             unsigned int buf_size,
+                             unsigned int *buf_consumed,
+                             hb_buffer_serialize_format_t format)
```
khaledhosny, line 537: This function does not currently use any flags, but I’m wondering if it should take flags anyway for parity with the glyphs function and in case flags were needed in the future? (it should be documented as unused in this case).

src/hb-buffer-serialize.cc
simoncozens: I'm not convinced that there is a consistent indentation style; that's why I converted everything to spaces (which is used *nearly* everywhere) in that file so there would be one.

Screenshot 2020-09-18 at 14 28 45

src/hb-buffer-serialize.cc
simoncozens: We're already different from the glyphs function in that we don't take a font parameter either. I can't imagine what flags you would want. But now that I think about it, I wonder if when serializing Unicode buffers we also want to output cluster IDs as well as codepoints...

src/hb-buffer-serialize.cc
jfkthame: I believe the usual HB style (which TBH I've never liked, but that's beside the point) is a two-space unit of indentation, with real tabs set to a tab width of 8 and used where there are line-initial runs of 8+ spaces.

src/hb-buffer-serialize.cc
khaledhosny: I’d make it take a font as well, but that is just me.

simoncozens: Yeah, I'm going to rework this, and think about how to show clusters.

src/hb-buffer-serialize.cc
ebraminio: Don't know whether editorconfig has a plugin in your editor or not but I use that to apply tab configuration on my editor, there is .editorconfig file in root of the project to guide the extension, not a fan of the style either.

src/hb-buffer-serialize.cc
behdad: > Yeah, I'm going to rework this, and think about how to show clusters.

Showing clusters is great idea. Same flags can be used. But taking font to serialize Unicodes buffer I can't see how helps.

src/hb-buffer-serialize.cc
behdad: > I believe the usual HB style (which TBH I've never liked, but that's beside the point) is a two-space unit of indentation, with real tabs set to a tab width of 8 and used where there are line-initial runs of 8+ spaces.

Right. And I inherited this from GNOME / FreeDesktop projects I was acquainted with. Open to alternative proposals.

simoncozens: I use the `.editorconfig`, but it indents using tabs:

```
indent_style = tab # should be space
```

src/hb-buffer-serialize.cc
```diff
@@ -508,18 +525,25 @@ hb_buffer_serialize_glyphs (hb_buffer_t *buffer,
...
+ * - If #HB_BUFFER_SERIALIZE_FLAG_NO_CLUSTERS is not set, the cluster
+ *   will be indicated with a `=` then #hb_glyph_info_t.cluster.
  *
  * ## json
  * A machine-readable, structured format.
- * The serialized codepoints will be a list of Unicode codepoints as
- * decimal integers. For example:
+ * The serialized codepoints will be a list of objects with the following
+ * properties:
+ * - `g`: the Unicode codepoint as a decimal integer
```
khaledhosny, line 540: Use of `g` here is confusing, shouldn’t be `u` or `c`?

simoncozens: Sorry, yes, typo. (u)

simoncozens:
@khaledhosny are you otherwise happy with this now?

behdad:
I'm typing my review. A few things still need work.

src/hb-buffer-serialize.cc
behdad: See: https://github.com/editorconfig/editorconfig-vscode/issues/190

behdad:
@simoncozens

I had to think this through again and remembered: There's something *really* tricky that you missed, and that's why we cannot emit the surrounding `[]`/`<>` delimiters; same reason has implementations on some other parts of your work, that I enumerate below: an empty buffer with `HB_BUFFER_TYPE_UNKNOWN` is accepted in all places; that is, in both API that expects `TYPE_GLYPHS` *and* API that expects `TYPE_UNICODE`. This is done such that after you clear buffer, you can use it with any API without having to set the content type explicitly. This is very important and must be maintained. Buffer content-type was added years after the base API. So it was a MUST to keep code like the following working:

```
  while (read input)
  {
    hb_buffer_clear_contents (buffer);
    hb_buffer_add_utf8 (buffer, input);
    hb_shape(..., buffer, ...);
    // Use buffer glyphs
  }
```
Since `hb_shape` expects buffer with Unicodes and modifies the buffer content type to glyphs, this code would have broken if the buffer type was not manually reset back to unicodes. Instead, I made `clear_contents` set content type to UNKNOWN, which is correct, and `add_utf8` et al check that buffer is Unicodes like this:
```
  assert ((buffer->content_type == HB_BUFFER_CONTENT_TYPE_UNICODE) ||
          (!buffer->len && (buffer->content_type == HB_BUFFER_CONTENT_TYPE_INVALID)));
```
and set content type to UNICODE upon success:
```
  buffer->content_type = HB_BUFFER_CONTENT_TYPE_UNICODE;
```

and `hb_shape()` et al check buffer type like this:
```
  if (unlikely (!buffer->len))
    return true;
  assert (buffer->content_type == HB_BUFFER_CONTENT_TYPE_UNICODE);
```
[I now see that if buffer is empty we don't check that it's NOT `GLYPHS`. That can be improved.]
and upon successful shaping, set it to UNICODE.

Also note that the buffer returned by `hb_buffer_get_empty()` has length 0 and `CONTENT_TYPE_UNKNOWN` and it should be accepted everywhere a buffer is accepted.

So!!! Now what should `hb_buffer_serialize (hb_buffer_get_empty ())` produce: `[]` or `<>`?!?!?!?!?!

*That* is the reason and the question.

behdad:
With that out of the way, here's things I like to see fixed:

- Remove the appending of the surrounding `[]`/`<>` as reasoned. Update docs to mention the outer thing is not printed. Humm. Now I realize you can keep this for JSON! We just can't output it for TEXT output. I'm happy to accept that!

- Your `hb_buffer_serialize()` does not handle `CONTENT_TYPE_UNKNOWN`. It should assert type is not UNKNOWN if length > 0.

- Your deserialize-json code accepts mix & match of glyph items and unicode items! That's wrong. It should only parse unicodes or only parse glyphs, depending on what function is called, OR if the type is not clear from the called function, ie. if the combined `hb_buffer_deserialize()` is called, then buffer type must be inferred from the JSON data.

- Similarly but differently, your deserialize-text.rl accepts a run of all glyphs, or a run of all unicodes, BUT! you accept both no matter what API was called or what buffer content type was!

To fix the above we might need a bit more ragel foo, or multiple machines. I'm happy to work with you on doing that. I might actually do those parts myself if time later tonight.

- Moreover, when parsing `unicode`, please accept all delimiters that `util/options.cc` uses:
```
#define DELIMITERS "<+>{},;&#\\xXuUnNiI\n\t\v\f\r "
```
except for `<>` since those two we handle separately. This list was carefully created to accept XML-style hex entities (`&#x1234`), programming-style numbers (`0x1234`), simple `u1234`, etc. I like to keep it and use it in `util/options.cc:parse_unicodes()` as well IMO. WDYT?

behdad:
Oh, while reviewing this I noticed and fixed a bug. Please update your patch to address same issue in your `unicodes` deserialize:
https://github.com/harfbuzz/harfbuzz/commit/a1f4d285e1305fda5da771c3c5555c912b8e2d65

simoncozens:
It’s really annoying that you can’t have a ragel machine with two heads and shared rules between them. Perhaps we could switch the main rule with an IFDEF an generate two outputs?

behdad:
> It’s really annoying that you can’t have a ragel machine with two heads and shared rules between them.

What about allowing all paths from the main rule and use actions to reject ones that are not applicable in a context? Works great in this case because if buffer is empty and content type UNKNOWN, then both paths are acceptable. Basically just add a rule at `item` level to check content type and reject if mismatch...

> Perhaps we could switch the main rule with an IFDEF an generate two outputs?

Nah. I think sharing is possible.

simoncozens:
That’s genius. I’ll work on it tomorrow.

behdad:
> That’s genius. I’ll work on it tomorrow.

Thanks.
Eg:
```
  if (unlikely (buffer->content_type != HB_BUFFER_CONTENT_TYPE_GLYPHS))
  {
    if (buffer->content_type != HB_BUFFER_CONTENT_TYPE_UNKNOWN)
      goto err;
    assert (buffer->length == 0);
    buffer->content_type = HB_BUFFER_CONTENT_TYPE_GLYPHS;
  }
```

simoncozens:
> Your hb_buffer_serialize() does not handle CONTENT_TYPE_UNKNOWN. It should assert type is not UNKNOWN if length > 0.

It does, but indirectly - each of the `hb_buffer_serialize_...` functions called by it checks for INVALID.

simoncozens:
> We just can't output it for TEXT output. I'm happy to accept that!

Urgh, this is tricky. It means that the input to `deserialize` will have delimiters, but the output of `serialize` will have no delimiters for text but will have them for JSON. That's ugly. :-)

I'm not sure this is as big a deal as you think. The issue is about how to serialize and deserialize empty buffers with type==INVALID. But deserializing them is a no-op anyway. How about if we make the serialization format distinct ("!!" or something)?

behdad:
> "!!" or something

Yes, let's do that!

Thanks. We'll just add enough `set_content_type()` to `hb-shape` et al to never see that string in our output.

Thank you Simon!

simoncozens:
There we go. Only thing I haven't dealt with was your comment about parsing all delimiters when deserializing. I don't quite know what you were referring to. Delimiters at start and end, or between items? Anyway, I think it's best to keep to the `U+xxxx` format for deserializing Unicode buffers, because we have to keep Unicode buffers in text format strictly disambiguated from potential glyph names. If we allow `[u0640=0]`, it would be parsable either as a Unicode buffer or a glyphs buffer.

behdad:
Thank you. Looks good to me in principle. Will try to do a full reivew.

behdad (commented):
> Anyway, I think it's best to keep to the `U+xxxx` format for deserializing Unicode buffers, because we have to keep Unicode buffers in text format strictly disambiguated from potential glyph names. If we allow `[u0640=0]`, it would be parsable either as a Unicode buffer or a glyphs buffer.

Humm. I wasn't going to rely on being able to distinguish those.  Okay. Let's go with this for now.

I'm looking into this now and if needed changing / finishing then continuing on integrating your https://github.com/harfbuzz/harfbuzz/pull/2686 and then some buffer work for https://github.com/harfbuzz/harfbuzz/issues/2468

Wish me attention!

behdad:
Finished and merged manually. Thanks Simon! More coming soon.

PR #2692, opened 2020-09-24
src/meson.build: allow introspection when cross-compiling

ffontaine:
introspection can be enabled when cross-compiling on certains conditions
(for example it is supported by buildroot)

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


khaledhosny:
This goes back to the original commit 6abe7983edcf820d7b8b35e1a7dddaaf3d0f1346 by @tp-m but no explanation was given.

tp-m:
> This goes back to the original commit [6abe798](https://github.com/harfbuzz/harfbuzz/commit/6abe7983edcf820d7b8b35e1a7dddaaf3d0f1346) by @tp-m but no explanation was given.

gobject-introspection usually doesn't just work out of the box when cross-compiling, but it probably also shouldn't be disabled without providing a way to override it for those who want to.

What we do in GStreamer is this:
```
build_gir = gir.found() and (not meson.is_cross_build() or get_option('introspection').enabled())
```

That way it's disabled by default for cross builds unless the option was enabled explicitly (vs. the default which is `auto`), in which case we assume the person building knows what they're doing and what they want.

ffontaine:
Thanks for your feedback, I can update the PR with your proposal or wait that you commit the fix to retrieve it.

tp-m:
Please feel free to just update your PR if you like this approach and it works for you.

ffontaine:
I updated the PR as requested, do you have more comments or do you plan on merging it soon?

khaledhosny:
Please rebase on master to see if it fixes the CI failure.

ffontaine:
Done, all tests passed

khaledhosny:
Thanks!

PR #2694, opened 2020-09-25
Expand windows build documentation

kg:


khaledhosny:
We don’t document specific package managers, there are 10s of them. And the options you are adding to the meson invocation are specific to your setup, and not something that is generally applicable. What options the build system takes are already documented.

alerque:
Note upcoming PR #2695 includes a bit more clarity in the Windows instructions section without getting into system or use-case specific options.

PR #2696, opened 2020-09-29
[use] Fix tests with MSVC

### LINKED ISSUE #2689: CI job that tests build using MSVC compiler broken
See discussion in #2688.

dscorbett:
I don’t understand why, but this seems to fix #2689. The MSYS2 timeouts are unrelated: they first appeared in #2648, before the MSVC tests started failing.

### DIFF
--- a/src/hb-ot-shape-complex-use-machine.hh
+++ b/src/hb-ot-shape-complex-use-machine.hh
@@ -359,7 +359,8 @@ find_syllables_use (hb_buffer_t *buffer)
   auto p =
     + hb_iter (info, buffer->len)
     | hb_enumerate
-    | hb_filter (not_standard_default_ignorable, hb_second)
+    | hb_filter ([] (const hb_glyph_info_t &i) { return not_standard_default_ignorable (i); },
+		 hb_second)
     | hb_filter ([&] (const hb_pair_t<unsigned, const hb_glyph_info_t &> p)
 		 {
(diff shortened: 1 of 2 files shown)


behdad:
Thanks. I haven't read the comments offered in the other issue yet either. Will try to get to the bottom of this tomorrow. In the mean time, feel free to merge.

alerque:
> I don’t understand why, but this seems to fix #2689. The MSYS2 timeouts are unrelated: they first appeared in #2648, before the MSVC tests started failing.

Did you mean to say this "seems to fix #2293"? This does not fix #2689 as you wrote, whether that is a timeout issue or not I'm not sure. I just skimmed the failure log and didn't see anything to indicate that it is timeout related. Can you point me to where you saw that?

dscorbett:
I meant it as I wrote it: this PR fixes the bug where the CI job that tests the build using the MSVC compiler is broken. There are green check marks next to all four MSVC runs. Why do you say that this doesn’t fix #2689?

There are still warnings about hb-algs.hh, so #2293 is not fixed.

There are no timeouts in the MSVC logs. They are in the MSYS2 logs. They are irrelevant to my change. I only mentioned the timeouts to explain why this PR still has Windows-related build failures, in case a reviewer might think this fix was incomplete or that I broke something new.

alerque:
You're right, I had my tabs, retinas, or some visual cortex cross-wired. Carry on.

alerque (approved):

PR #2698, opened 2020-09-29
Revert "Fix get_baseline when `baseline_tag` does not exist"

### LINKED ISSUE #2697: Ci build tests on MSYS2 encountering timeouts
As a follow up to #2689, the CI job for MSYS2 is encountering timeout errors and hence not very useful. These should be tracked down and resolved somehow

khaledhosny:
Reverts harfbuzz/harfbuzz#2648

Lets see if this fixes #2697.


alerque:
No love.

khaledhosny:
Yes, might be coincidental.

khaledhosny:
I don’t know what is going on, but these timeouts make no sense since the failing tests are not that slow. Something else is broken.

PR #2699, opened 2020-09-29
[subset] Add a more complex layout subsetting test case and fix the issues it exposed.

garretrieger:
This adds a new layout subsetting test case using the Amiri font. This exposes several bugs/differences in harfbuzz subsetting versus fontTools. This PR includes fixes for all found issues:

- Adds GDEF subsetting to the layout tests.
- ChainContextFormat3 currently subsets the backtrack, input, and lookahead sequences. These should not be subset, but instead the subtable should only be retained if all three sequences are a subset of the glyph set.
- In ChainContextFormat2 empty offsets are pruned from the end, but the serialization allocation wasn't being shrunk to match.
- Lookup closure problems:
  - [Chain]ContextFormat3: don't recurse into the referenced lookups if the current subtable can't be activated by the retained glyph set.
 - [Chain]ContextFormat(1|2): don't recurse into the referenced lookups from a [Chain]Rule if the rule can't be activated by the retained glyph set.
- Use plan->glyphset_sub instead of plan->glyphset for all lookup subsetting in GPOS/GSUB. glyphset can contain extra glyphs which don't participate in layout, so shouldn't be considered for layout subsetting.
- Prune features from the final subset that aren't reachable via the script table. Current behaviour was to retain any features which intersected the set of retained lookups.


src/hb-ot-layout-gsubgpos.hh
```diff
@@ -2701,9 +2734,28 @@ struct ChainContextFormat2
...
+      {&backtrack_class_def,
+       &input_class_def,
+       &lookahead_class_def}
+    };
+
     + hb_iter (ruleSet)
     | hb_map (hb_add (this))
-    | hb_apply ([&] (const ChainRuleSet &_) { _.closure_lookups (c); })
+    | hb_enumerate
+    | hb_filter([&] (unsigned gid)
```
behdad, line 2754: This is not gid. It's class index.

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -2986,13 +3043,16 @@ struct ChainContextFormat3
...
+      return_trace (false);
 
-    + it
-    | hb_apply (subset_offset_array (c, *out, base))
-    ;
+    for (auto& offset : it) {
+      auto *o = out->serialize_append (c->serializer);
+      if (unlikely (!o) || !o->serialize_subset (c, offset, base))
+        return_trace (false);
+    }
```
behdad, line 3053: I don't understand this chunk.

behdad (approved):
Thanks. This looks great! I added a few minor comments. Please merge after addressing.

src/hb-ot-layout-gsubgpos.hh
behdad: Ah okay. I understand now.

PR #2700, opened 2020-10-04
meson: Fix cairo_ft_dep when fontconfig is a subproject

xclaesse:
When cairo comes from a subproject fallback it should still take
    cairo-ft dependency, otherwise it miss freetype2 cflags.
    
    Also update wrap file to use upstream repository now that meson support
    has been merged.
    
    This requires Meson >= 0.54.0 because cairo itself requires that version
    and we use the new `fallback: 'cairo'` syntax without specifying a
    variable name because cairo does not provide a variable name for
    cairo-ft.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,7 +142,8 @@ jobs:
     executor: win32-executor
     steps:
       - checkout
-      - run: sudo apt update && DEBIAN_FRONTEND=noninteractive sudo apt install -y ninja-build binutils meson gcc g++ pkg-config gtk-doc-tools libfreetype6-dev libglib2.0-dev libcairo2-dev libicu-dev libgraphite2-dev python3 python3-pip git g++-mingw-w64-i686 zip
+      - run: sudo apt update && DEBIAN_FRONTEND=noninteractive sudo apt install -y ninja-build binutils gcc g++ pkg-config gtk-doc-tools libfreetype6-dev libglib2.0-dev libcairo2-dev libicu-dev libgraphite2-dev python3 python3-pip git g++-mingw-w64-i686 zip
+      - run: pip3 install meson==0.56.0 --upgrade
       - run: .ci/build-win32.sh
       - store_artifacts:
           path: harfbuzz-win32.zip
--- a/meson.build
+++ b/meson.build
@@ -127,34 +127,31 @@ cairo_dep = null_dep
 cairo_ft_dep = null_dep
 if not get_option('cairo').disabled()
   cairo_dep = dependency('cairo', required: false)
+  cairo_ft_dep = dependency('cairo-ft', required: false)
 
   if (not cairo_dep.found() and
       cpp.get_id() == 'msvc' and
       cpp.has_header('cairo.h'))
     cairo_dep = cpp.find_library('cairo', required: false)
+    if cairo_dep.found() and cpp.has_function('cairo_ft_font_face_create_for_ft_face',
+                                              prefix: '#include <cairo-ft.h>',
+                                              dependencies: cairo_dep)
+      cairo_ft_dep = cairo_dep
+    endif
   endif
 
   if not cairo_dep.found()
-    # Note that we don't have harfbuzz -> cairo -> freetype2 -> harfbuzz fallback
-    # dependency cycle here because we have configured freetype2 above with
-    # harfbuzz support disabled, so when cairo will lookup freetype2 dependency
-    # it will be forced to use that one.
-    cairo_dep = dependency('cairo', fallback: ['cairo', 'libcairo_dep'],
-                           required: get_option('cairo'))
-  endif
-
-  # Ensure that cairo-ft is fetched from the same library as cairo itself
-  if cairo_dep.found()
-    if cairo_dep.type_name() == 'internal'
-      # It is true at least for the port we have
-      cairo_ft_dep = cairo_dep
-    elif (cairo_dep.type_name() == 'library' and
-          cpp.has_function('cairo_ft_font_face_create_for_ft_face',
-                           prefix: '#include <cairo-ft.h>',
-                           dependencies: cairo_dep))
-      cairo_ft_dep = cairo_dep
-    else # including the most important type for us, 'pkgconfig'
-      cairo_ft_dep = dependency('cairo-ft', required: get_option('cairo'))
+    # Requires Meson 0.54.0 to use cairo subproject
+    if meson.version().version_compare('>=0.54.0')
+      # Note that we don't have harfbuzz -> cairo -> freetype2 -> harfbuzz fallback
+      # dependency cycle here because we have configured freetype2 above with
+      # harfbuzz support disabled, so when cairo will lookup freetype2 dependency
+      # it will be forced to use that one.
+      cairo_dep = dependency('cairo', fallback: 'cairo', required: get_option('cairo'))
+      cairo_ft_dep = dependency('cairo-ft', fallback: 'cairo', required: get_option('cairo'))
+    elif get_option('cairo').enabled()
+      error('cairo feature is enabled but it cannot be found on the system and ' +
+            'meson>=0.54.0 is required to build it as subproject')
     endif
   endif
 endif
--- a/subprojects/cairo.wrap
+++ b/subprojects/cairo.wrap
@@ -1,5 +1,5 @@
 [wrap-git]
 directory=cairo
-url=https://github.com/ebraminio/cairo.git
+url=https://gitlab.freedesktop.org/cairo/cairo.git
 depth=1
-revision=meson
+revision=master
--- a/subprojects/expat.wrap
+++ b/subprojects/expat.wrap
@@ -1,10 +1,9 @@
 [wrap-file]
-directory = expat-2.2.5
+directory = expat-2.2.9
+source_url = https://github.com/libexpat/libexpat/releases/download/R_2_2_9/expat-2.2.9.tar.xz
+source_filename = expat-2.2.9.tar.bz2
+source_hash = 1ea6965b15c2106b6bbe883397271c80dfa0331cdf821b2c319591b55eadc0a4
+patch_url = https://wrapdb.mesonbuild.com/v1/projects/expat/2.2.9/3/get_zip
+patch_filename = expat-2.2.9-3-wrap.zip
+patch_hash = e9aaace62e9a158b5e96f5c38c9f81f369179206acd87697653d777c0d3975d3
 
-source_url = https://github.com/libexpat/libexpat/releases/download/R_2_2_5/expat-2.2.5.tar.bz2
-source_filename = expat-2.2.5.tar.bz2
-source_hash = d9dc32efba7e74f788fcc4f212a43216fc37cf5f23f4c2339664d473353aedf6
-
-patch_url = https://wrapdb.mesonbuild.com/v1/projects/expat/2.2.5/4/get_zip
-patch_filename = expat-2.2.5-4-wrap.zip
-patch_hash = 25748839be2bbdd2ff586d1a05aa6fc37aeada75c78416df6e8347a6321abaac
--- a/subprojects/fontconfig.wrap
+++ b/subprojects/fontconfig.wrap
@@ -1,6 +1,6 @@
 [wrap-git]
 directory=fontconfig
-url=https://github.com/centricular/fontconfig.git
+url=https://gitlab.freedesktop.org/fontconfig/fontconfig.git
 depth=1
-push-url=git@github.com:centricular/fontconfig.git
-revision=meson
+push-url=git@gitlab.freedesktop.org/fontconfig/fontconfig.git
+revision=master
--- a/subprojects/google-benchmark.wrap
+++ b/subprojects/google-benchmark.wrap
@@ -1,8 +1,9 @@
 [wrap-file]
-directory = benchmark-1.4.1
-source_url = https://github.com/google/benchmark/archive/v1.4.1.zip
-source_filename = benchmark-1.4.1.zip
-source_hash = 61ae07eb5d4a0b02753419eb17a82b7d322786bb36ab62bd3df331a4d47c00a7
-patch_url = https://wrapdb.mesonbuild.com/v1/projects/google-benchmark/1.4.1/1/get_zip
-patch_filename = google-benchmark-1.4.1-1-wrap.zip
-patch_hash = 4cc5fe02ebd4fc82e110919b7977d7463eb2a99e4ecb9feca920eab6fd911d67
+directory = benchmark-1.5.2
+source_url = https://github.com/google/benchmark/archive/v1.5.2.zip
+source_filename = benchmark-1.5.2.zip
+source_hash = 21e6e096c9a9a88076b46bd38c33660f565fa050ca427125f64c4a8bf60f336b
+patch_url = https://wrapdb.mesonbuild.com/v1/projects/google-benchmark/1.5.2/1/get_zip
+patch_filename = google-benchmark-1.5.2-1-wrap.zip
+patch_hash = 49f41e4a7e68ac258b6509b9de9857441903be4fb473454c4cba8be885f0c6c3
+
--- a/subprojects/libpng.wrap
+++ b/subprojects/libpng.wrap
@@ -1,10 +1,12 @@
 [wrap-file]
-directory = libpng-1.6.35
+directory = libpng-1.6.37
+source_url = https://github.com/glennrp/libpng/archive/v1.6.37.tar.gz
+source_filename = libpng-1.6.37.tar.gz
+source_hash = ca74a0dace179a8422187671aee97dd3892b53e168627145271cad5b5ac81307
+patch_url = https://wrapdb.mesonbuild.com/v1/projects/libpng/1.6.37/3/get_zip
+patch_filename = libpng-1.6.37-3-wrap.zip
+patch_hash = 6c9f32fd9150b3a96ab89be52af664e32207e10aa9f5fb9aa015989ee2dd7100
 
-source_url = https://github.com/glennrp/libpng/archive/v1.6.35.tar.gz
-source_filename = libpng-1.6.35.tar.gz
-source_hash = 6d59d6a154ccbb772ec11772cb8f8beb0d382b61e7ccc62435bf7311c9f4b210
+[provide]
+libpng = libpng_dep
 
-patch_url = https://wrapdb.mesonbuild.com/v1/projects/libpng/1.6.35/5/get_zip
-patch_filename = libpng-1.6.35-5-wrap.zip
-patch_hash = da42b18e8d75a88615bdbc1c7bbf1f739ae19f63a8e70d96c90bc448326ae6b7
--- a/subprojects/zlib.wrap
+++ b/subprojects/zlib.wrap
@@ -1,5 +1,12 @@
-[wrap-git]
-directory=zlib
-url=https://github.com/centricular/zlib.git
-depth=1
-revision=meson
+[wrap-file]
+directory = zlib-1.2.11
+source_url = http://zlib.net/fossils/zlib-1.2.11.tar.gz
+source_filename = zlib-1.2.11.tar.gz
+source_hash = c3e5e9fdd5004dcb542feda5ee4f0ff0744628baf8ed2dd5d66f8ca1197cb1a1
+patch_url = https://wrapdb.mesonbuild.com/v1/projects/zlib/1.2.11/5/get_zip
+patch_filename = zlib-1.2.11-5-wrap.zip
+patch_hash = 728c8e24acbc2e6682fbd950fec39e2fc77528af361adb87259f8a8511434004
+
+[provide]
+zlib = zlib_dep
+


khaledhosny:
I’m feeling really uneasy about the growing complexity of the fallback dependencies. HarfBuzz build system should not be a mini package manager. I’m inclined to completely drop it and let the users handle the dependencies as they see fit.

alerque:
> I’m feeling really uneasy about the growing complexity of the fallback dependencies. HarfBuzz build system should not be a mini package manager. I’m inclined to completely drop it and let the users handle the dependencies as they see fit.

Honestly I don't really make the connection between declaring dependencies (however complex they are) using a build system and becoming a package manager.

...that being said I do agree that the series of fallbacks is adding a lot of complexity we might not need. I think it also adds quite a bit of unpredictability. I would tend to favor explicit build flags, and the user setting those flags would either enable a feature and *enforce* it's dependencies or disable it and not look for them at all. With so many automatic detection and fallbacks, two people can build the same thing on relatively similar platforms and not come away with the same library at all. If there was a default baseline config and builders had to explicitly set anything they wanted differently for their platform and what dependencies they wanted to pull in or not things would be a bit more predictable.

But this PR is fine ... this is discussion for future direction. I wouldn't suggest redoing anything in the build stack to hastily, and certainly fixing issues like this is fine for now.

alerque (approved):

khaledhosny:
> Honestly I don't really make the connection between declaring dependencies (however complex they are) using a build system and becoming a package manager.

This is done for ”fallback” dependency, which in meson speak a subproject that we build when the dependency is not found. Now we depend on `cairo-ft`, it is not our job to also mange `cairo-ft` own dependencies, which is what is being done here.

khaledhosny:
There is also no explanation what this change is fixing, only what is doing.

alerque:
Okay you might be right.

xclaesse:
The problem here is `cairo_ft_dep = cairo_dep` is wrong because it miss cflags for fontconfig. It is true that all harfbuzz should only care about is getting `cairo-ft` dependency, but unfortunately cairo's meson port is missing that, it's not even generating cairo-ft.pc. That's a bug in cairo.

The proper fix should be done in both cairo in harfbuzz:
- Cairo meson should provide cairo-ft dependency
- Harfbuzz should always do `cairo_ft_dep = dependency('cairo-ft', required: get_option('cairo'))` regardless if cairo comes from system or subproject.

Side question: Does harfbuzz requires that cairo has been built with fontconfig? Or can it work either way?

khaledhosny:
> The proper fix should be done in both cairo in harfbuzz:

Agree. We will also need to switch to Cairo upstream not the fork we are using now.

>  Does harfbuzz requires that cairo has been built with fontconfig

We don’t seem to use any of Cairo’s FontConfig function, and we check only for cairo-ft.

xclaesse:
> We don’t seem to use any of Cairo’s FontConfig function, and we check only for cairo-ft.

Seems I got this wrong, cairo-ft is for FreeType and not FontConfig. Sorry for the confusion.

I created that cairo PR: https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/68

xclaesse:
Sorry for the confusing. I updated this PR to use the fix I did cairo and updated description and commit message. Hopfully it makes things a bit more clear.

xclaesse:
CC @tp-m

xclaesse:
Hmm this requires Meson 0.54 in the CI.

behdad:
I share Khaled's concerns.

tp-m:
As far as I can tell this MR is a clean-up of some meson code plus a correctness fix so that harfbuzz can be consumed correctly as subproject by other projects such as cairo, Gtk, GStreamer in certain situations.

This is independent of the .wrap files in `harfbuzz/subprojects`. Those are for convenience, and they make it very easy to build and hack on harfbuzz and its stack in some situations, e.g. on windows or older distros or when doing fuzzing/CI. You can remove these wrap files of course, but I think they're quite cool to have, and shouldn't be too much hassle once things are settled. Removing them will only make it more difficult to build harfbuzz.

The fix would be needed even if you decided not to ship wrap files in harfbuzz any more. The fix is for other projects that depend on you.

The cairo meson port landed upstream only recently, and we're still shaking out some bugs, sorry about that!

xclaesse:
Right, sorry for all the confusion here, my initial patch was wrong and the right fix for the problem I had is actually in cairo: https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/68/diffs?commit_id=4395f7a5b60004bfe1ac62c2bae993385cb8b504.

Since then I updated this PR (probably should have closed and opened a new one) with a cleanup how harfbuzz lookup the cairo dependency. It's actually simplifying the code (+12/-23 lines).

alerque:
No worries on the PR update, GitHub's tooling actually makes it easier to review this way than closing/re-opening PRs. There is no easy way to compare what changed between PRs, but there is a way to see what changed since your last visit to a PR and also see where the force-pushed commits fit into the discussion timeline.

These changes are definitely looking better! My only question is will this be a regression for building against systems that don't have the latest Cairo with the upstream build fixes?

xclaesse:
> These changes are definitely looking better! My only question is will this be a regression for building against systems that don't have the latest Cairo with the upstream build fixes?

This PR should make no difference when getting cairo from system. 

> Right, sorry for all the confusion here, my initial patch was wrong and the right fix for the problem I had is actually in cairo: https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/68/diffs?commit_id=4395f7a5b60004bfe1ac62c2bae993385cb8b504.

And now I realize even that patch is wrong, cairo.pc is supposed to require fontconfig already, so it's not needed in cairo-ft.pc. But it's not harfbuzz issue anyway...

xclaesse:
@behdad @tp-m @alerque The cairo side of this has been merged, any objection to merge this in harfbuzz? I updated the PR to not require a newer version of Meson.

xclaesse:
Argh, cairo requires meson >= 0.54.0, so it needs to be updated in the CI too since the CI uses cairo as subproject.

alerque:
> Argh, cairo requires meson >= 0.54.0

Meson distribution is still very uneven. Releases with different bugs have meant some platforms have gotten stuck upgrading on various different releases. Given earlier troubles and the effort that went into getting the current build system working across older meson releases (and the unresolved tension with deprecating other build systems) I don't think this is a good juncture to narrow the range of supported meson versions.

xclaesse:
I'm not aware of bugs in meson affecting harfbuzz, can you please be more specific and report issues of any? It is true that. 0 release have their regressions but there are point releases that fix most reported issues.

xclaesse:
Who knows how to update meson version in CI runners?

khaledhosny:
We are intentionally requiring meson >= 0.47.0, we do not wish to require a newer version of meson for the time being. Some of the CI jobs intentionally install meson==0.47.0 to enforce this.

xclaesse:
@khaledhosny why requiring a very old version? Meson support just got merged in harfbuzz very recently, users on very old system won't use it any way.  In the early days it's better to require latest version to not accumulate legacy.

khaledhosny:
https://github.com/harfbuzz/harfbuzz/pull/2632#issuecomment-672265606

xclaesse:
I strongly disagree, but if you prefer broken Cairo detection that's your call. I guess you'll be able to merge this in about 2 or 3 years.

alerque:
> Meson support just got merged in harfbuzz very recently, users on very old system won't use it any way.

The hope with Meson support is that we would be able to *drop* the other build systems. Currently there are 3: automake, cmake, and meson. You are correct that Meson was just recently added, but the others were (a bit hastily) removed entirely. Then they were fetched back because it became clear that Meson didn't work for everybody and we needed to support a longer transition period. At that point a huge effort was made to bring the Meson support from only working on very new builds to working on much older Meson. If and when we do keep moving in the direction of dropping cmake and/or automake, having support for a broad range of Meson versions will be crucial. Harfbuzz deploys in a *lot* of places with a wide variety of system tooling. Requiring very new build tooling would be a huge regression and work against keeping the Harfbuzz ecosystem as up to date as possible. As long as people can keep dropping new Harfbuzz versions into old systems without breakage those old systems stand a chance of getting good HB support. As soon as you require fresh dependencies or tooling you start limiting what platforms can keep up.

> I strongly disagree, but if you prefer broken Cairo detection that's your call. I guess you'll be able to merge this in about 2 or 3 years.

I think we'd prefer broken Cairo detection to *requiring* the freshest Meson. But is that really a compromise we have to make? Is there really no other way? I believe there must be work-arounds so we can continue to support building under older Meson so that we can keep moving forward with deprecating other build systems.

xclaesse:
Hmmm ok, sorry for having been harsh, I miss understood the linked comment, I thought you wanted to keep autotools indefinitely. If the plan is to be meson only then it makes much more sense.

What I could do is do the cairo fallback only if meson version is new enough. With older version of meson it will only look on system.

xclaesse:
@alerque problem remains that the Windows CI requires cairo as subproject. Maybe we can update meson version only for that CI job?

QuLogic:
FYI, this fixes building the full GStreamer set, with their overarching `gst-build` project, on Windows. On current `master`, it fails to build due to missing includes (presumably from missing search paths.)

khaledhosny:
> problem remains that the Windows CI requires cairo as subproject. Maybe we can update meson version only for that CI job?

This should be fine since this CI job is also testing the use of subprojects.

xclaesse:
Hmm build of cairo fails:
```
FAILED: subprojects/cairo/src/libcairo-2.dll 
i686-w64-mingw32-gcc  -o subprojects/cairo/src/libcairo-2.dll 'subprojects/cairo/src/da09890@@cairo@sha/cairo-analysis-surface.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-arc.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-array.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-atomic.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-base64-stream.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-base85-stream.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-bentley-ottmann-rectangular.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-bentley-ottmann-rectilinear.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-bentley-ottmann.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-botor-scan-converter.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-boxes-intersect.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-boxes.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-cache.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-clip-boxes.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-clip-polygon.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-clip-region.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-clip-surface.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-clip-tor-scan-converter.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-clip.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-color.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-composite-rectangles.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-compositor.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-contour.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-damage.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-debug.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-default-context.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-device.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-error.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-fallback-compositor.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-fixed.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-font-face-twin-data.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-font-face-twin.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-font-face.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-font-options.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-freed-pool.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-freelist.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-gstate.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-hash.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-hull.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-image-compositor.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-image-info.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-image-source.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-image-surface.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-line.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-lzw.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-mask-compositor.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-matrix.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-mempool.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-mesh-pattern-rasterizer.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-misc.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-mono-scan-converter.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-mutex.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-no-compositor.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-observer.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-output-stream.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-paginated-surface.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-path-bounds.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-path-fill.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-path-fixed.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-path-in-fill.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-path-stroke-boxes.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-path-stroke-polygon.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-path-stroke-traps.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-path-stroke-tristrip.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-path-stroke.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-path.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-pattern.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-pen.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-polygon-intersect.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-polygon-reduce.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-polygon.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-raster-source-pattern.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-recording-surface.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-rectangle.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-rectangular-scan-converter.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-region.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-rtree.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-scaled-font.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-shape-mask-compositor.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-slope.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-spans-compositor.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-spans.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-spline.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-stroke-dash.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-stroke-style.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-surface-clipper.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-surface-fallback.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-surface-observer.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-surface-offset.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-surface-snapshot.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-surface-subsurface.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-surface-wrapper.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-surface.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-time.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-tor-scan-converter.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-tor22-scan-converter.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-toy-font-face.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-traps-compositor.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-traps.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-tristrip.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-unicode.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-user-font.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-version.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-wideint.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-cff-subset.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-scaled-font-subsets.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-truetype-subset.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-type1-fallback.c.obj' 'subprojects/cairo/src/da09890@
@cairo@sha/cairo-type1-glyph-names.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-type1-subset.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-type3-glyph-surface.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-pdf-operators.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-pdf-shading.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-tag-attributes.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-deflate-stream.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-png.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-svg-surface.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-ft-font.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/win32_cairo-win32-debug.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/win32_cairo-win32-device.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/win32_cairo-win32-gdi-compositor.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/win32_cairo-win32-system.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/win32_cairo-win32-surface.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/win32_cairo-win32-display-surface.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/win32_cairo-win32-printing-surface.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/win32_cairo-win32-font.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-composite.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-device.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-dispatch.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-glyphs.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-gradient.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-info.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-msaa-compositor.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-operand.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-shaders.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-source.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-spans-compositor.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-surface.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-traps-compositor.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-wgl-context.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-script-surface.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-ps-surface.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-pdf-surface.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-pdf-interchange.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-tag-stack.c.obj' 'subprojects/cairo/src/da09890@@cairo@sha/cairo-xml-surface.c.obj' -Wl,--allow-shlib-undefined -Wl,-O1 -shared -Wl,--start-group -Wl,--out-implib=subprojects/cairo/src/libcairo.dll.a -static-libgcc -Wl,-Bstatic -lpthread subprojects/zlib/libz.dll.a subprojects/libpng-1.6.35/libpng16.dll.a subprojects/freetype2/libfreetype.dll.a subprojects/pixman/pixman/libpixman-1.dll.a -lm -lgdi32 -lmsimg32 -lopengl32 -lkernel32 -luser32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 -Wl,--end-group
/usr/bin/i686-w64-mingw32-ld: subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-composite.c.obj:cairo-gl-composite.c:(.text+0x156): undefined reference to `glActiveTexture@4'
/usr/bin/i686-w64-mingw32-ld: subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-composite.c.obj:cairo-gl-composite.c:(.text+0x252): undefined reference to `glActiveTexture@4'
/usr/bin/i686-w64-mingw32-ld: subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-composite.c.obj:cairo-gl-composite.c:(.text+0xb0a): undefined reference to `glBlendFuncSeparate@16'
/usr/bin/i686-w64-mingw32-ld: subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-composite.c.obj:cairo-gl-composite.c:(.text+0xbd0): undefined reference to `glBlendFuncSeparate@16'
/usr/bin/i686-w64-mingw32-ld: subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-device.c.obj:cairo-gl-device.c:(.text+0xaaf): undefined reference to `glActiveTexture@4'
/usr/bin/i686-w64-mingw32-ld: subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-device.c.obj:cairo-gl-device.c:(.text+0xaca): undefined reference to `glActiveTexture@4'
/usr/bin/i686-w64-mingw32-ld: subprojects/cairo/src/da09890@@cairo@sha/cairo-gl-glyphs.c.obj:cairo-gl-glyphs.c:(.text+0xc1): undefined reference to `glActiveTexture@4'
collect2: error: ld returned 1 exit status
```

CC @tp-m @nirbheek

xclaesse:
@alatiera Added a commit here to update all wraps while at it. This should replace https://github.com/harfbuzz/harfbuzz/pull/2746.

Remains that cairo cannot be cross built to windows currently, I reported the bug there if someone wants to investigate: https://gitlab.freedesktop.org/cairo/cairo/-/issues/435.

xclaesse:
Ok, it seems the cairo issue has been already fixed, or at least worked around enough that the CI now pass here. So I think this is ready to be merged now.

khaledhosny:
Thanks! I think freetype needs to be updated as well since upstream has meson support now.

xclaesse:
> Thanks! I think freetype needs to be updated as well since upstream has meson support now.

Indeed, but the meson port that got merged upstream got developed independently to the port we are using here, and it has a few issues. They are in the process of migrating to gitlab which will make it much easier to contribute, I made a MR already: https://gitlab.freedesktop.org/freetype/freetype/-/merge_requests/1

PR #2162, opened 2020-02-06
[subset] GPOS 5 MarkToLigature subsetting support

qxliu76:

PR #2701, opened 2020-10-05
[subset] GPOS 5 MarkToLigature subsetting support

garretrieger:
Replaces PR #2162 

- Fixes merge conflicts and updates the PR with respect to variation indexes changes.
- Removes unnecessary vector allocation during subsetting.
- Fixes incorrect offset base in LigatureArray.


test/subset/data/Makefile.sources
```diff
@@ -16,6 +16,7 @@ TESTS = \
 	tests/layout.gpos2.tests \
 	tests/layout.gpos3.tests \
 	tests/layout.gpos4.tests \
+	tests/layout.gpos5.tests \
```
ebraminio, line 19: Feel free to update respective meson.build list also, thanks :)

garretrieger: Good catch, fixed.

src/hb-ot-layout-gpos-table.hh
```diff
@@ -566,6 +566,26 @@ struct AnchorMatrix
...
+  bool subset (hb_subset_context_t *c,
+	       unsigned cols,
+	       const hb_map_t *klass_mapping) const
+  {
+    TRACE_SUBSET (this);
+    auto *out = c->serializer->start_embed (*this);
+
+    auto indexes =
+    + hb_range (rows * cols)
+    | hb_filter ([=] (unsigned index) { return klass_mapping->has (index % cols); })
```
behdad, line 578: I wonder if we can later speed this up to avoid so many remainder operations.

behdad, line 578: There's an assumption in the output of `Markclass_closure_and_remap_indexes` that is not reflected on the type of the `hb_map_t *klass_map` but relied on in the code here and in other places, that basically the mapping is monotone and continuous. Let's commit this and work on improving internals later.

behdad (approved):
This looks good!

src/hb-ot-layout-gpos-table.hh
garretrieger: Yeah I believe we frequently make that assumption for the glyph map as well. It might be nice to have a special version of map which enforces that.

behdad: > Yeah I believe we frequently make that assumption for the glyph map as well. It might be nice to have a special version of map which enforces that.

Yep. We'll switch to a different data-structure. I think the CFF subsetter added something along those lines.

behdad: > Yep. We'll switch to a different data-structure. I think the CFF subsetter added something along those lines.

hb_inc_bimap_t I think.

Issue #2703, opened 2020-10-08
PairPosFormat2::intersects() possibly wrong?


garretrieger:
https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ot-layout-gpos-table.hh#L1331

Currently returns true if the glyph set intersects both coverage and classDef2. However, PairPosFormat2 implicitly includes class 0 (which is matches anything not assigned a class). This means that any glyph will match classDef2. I believe then that we should change intersects to just test for intersection against coverage.

The current implementation is causing issues in subsetting fonts with PairPosFormat2 lookups. For example consider a lookup where the retained glyph set intersects coverage and classDef1, but does not intersect classDef2. Since PairPosFormat2 spec says class 0 should be included then you end up with classCount1 = 2 and classCount2 = 1 (for class 0). FontTools will retain the lookup in the final subset in this case, but harfbuzz subset will drop it during lookup collection since intersects() currently reports false.

For reference here's the fontTools subsetting output:
```
          <PairPos Format="2">
            <Coverage Format="1">
              <Glyph value="at"/>
            </Coverage>
            <ValueFormat1 value="4"/>
            <ValueFormat2 value="0"/>
            <ClassDef1 Format="1">
              <ClassDef glyph="at" class="1"/>
            </ClassDef1>
            <ClassDef2 Format="2">
            </ClassDef2>
            <!-- Class1Count=2 -->
            <!-- Class2Count=1 -->
            <Class1Record index="0">
              <Class2Record index="0">
                <Value1 XAdvance="0"/>
              </Class2Record>
            </Class1Record>
            <Class1Record index="1">
              <Class2Record index="0">
                <Value1 XAdvance="0"/>
              </Class2Record>
            </Class1Record>
          </PairPos>
```

garretrieger:
I tested changing intersects() to only check coverage for intersection and all tests are still passing, plus the subsetter starts correctly retaining the lookups it was previously dropping.

garretrieger:
My proposed fix: https://github.com/harfbuzz/harfbuzz/pull/2704

The other option would be to change fontTool's behaviour.

behdad:
Proposed fix is wrong. I'll comment in detail soon.

behdad:
This is what I suggest instead, fixing fonttools first, then making harfbuzz match it:

https://github.com/fonttools/fonttools/pull/2221

PR #2704, opened 2020-10-08
[subset] Make PairPosFormat2 subsetting consistent with fontTools.

garretrieger:
- Don't drop pair pos format 2 lookups if the classDef2 subtable does not intersect the glyph set. classDef2 implicitly contains class 0 which matches everything.
- Make classDef2 a non-null offset, this matches how fontTools subsetter treats it.

Add an additional test case which demonstrates these issues.

Fix for #2703


behdad:
Closing. See https://github.com/harfbuzz/harfbuzz/issues/2703#issuecomment-796026680 for proposed solution.

PR #2705, opened 2020-10-08
[subset] Add subset support for Extension lookups (GPOS 9, GSUB 7).

garretrieger:


src/hb-ot-layout-gsubgpos.hh
```diff
@@ -3143,6 +3161,18 @@ struct Extension
     }
   }
 
+  // Specialization of dispatch for subset. dispatch() normally just
+  // dispatches to the sub table this points too, but for subset
+  // we need to run subset on this subtable too.
```
behdad, line 3166: I don't understand why you need this. The `hb_subset_context_t::dispatch()` does the same already:
```
struct hb_subset_context_t : 
       hb_dispatch_context_t<hb_subset_context_t, bool, HB_DEBUG_SUBSET> 
{ 
  const char *get_name () { return "SUBSET"; } 
  static return_t default_return_value () { return true; } 
 
  private: 
  template <typename T, typename ...Ts> auto 
  _dispatch (const T &obj, hb_priority<1>, Ts&&... ds) HB_AUTO_RETURN 
  ( obj.subset (this, hb_forward<Ts> (ds)...) ) 
  template <typename T, typename ...Ts> auto 
  _dispatch (const T &obj, hb_priority<0>, Ts&&... ds) HB_AUTO_RETURN 
  ( obj.dispatch (this, hb_forward<Ts> (ds)...) ) 
  public: 
  template <typename T, typename ...Ts> auto 
  dispatch (const T &obj, Ts&&... ds) HB_AUTO_RETURN 
  ( _dispatch (obj, hb_prioritize, hb_forward<Ts> (ds)...) ) 
```

and the `Extension::dispatch()`:
```
  template <typename context_t, typename ...Ts> 
  typename context_t::return_t dispatch (context_t *c, Ts&&... ds) const 
  { 
    TRACE_DISPATCH (this, u.format); 
    if (unlikely (!c->may_dispatch (this, &u.format))) return_trace (c->no_dispatch_return_value ()); 
    switch (u.format) { 
    case 1: return_trace (u.format1.dispatch (c, hb_forward<Ts> (ds)...)); 
    default:return_trace (c->default_return_value ()); 
    } 
  } 
```

so the specialization looks unnecessary to me.

test/subset/data/Makefile.sources
```diff
@@ -18,6 +18,7 @@ TESTS = \
 	tests/layout.gpos4.tests \
 	tests/layout.gpos6.tests \
 	tests/layout.gpos8.tests \
+	tests/layout.gpos9.tests \
```
ebraminio, line 21: Something should've be added to `test/subset/data/Makefile.am` list also.

behdad: Ha. That's weird. Thanks for pointing out. The parallel list should be moved to `Makefile.sources` as well...

src/hb-ot-layout-gsubgpos.hh
garretrieger: Lookup::dispatch calls sub_table->dispatch (Extension::dispatch in this case) directly (https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ot-layout-common.hh#L1096) and likewise
Extension::dispatch calls dispatch() on the referenced lookup directly. So subset_context->dispatch(...) doesn't get called and the subset() call on ExtensionFormat1 ends up getting skipped.

 I tried changing Extension::dispatch to use context->dispatch(...), but that ran into issues because the other context's don't implement automatically picking between the dispatch or other method if present like subset does.

So just specializing the dispatch call here was the simplest way to get this to work without needing to make changes to all of the other contexts.

garretrieger: The other attempt at implementing this (https://github.com/harfbuzz/harfbuzz/pull/2582) ran into the same issue and added a may_subset() method like sanitize context uses as an alternate solution, but I'm not a fan of that approach.

test/subset/data/Makefile.sources
garretrieger: Fixed

src/hb-ot-layout-gsubgpos.hh
behdad: Okay let me think.

behdad: Thanks. Let's merge this, and I'll add a note and keep in mind for future `dispatch` mechanism improvements.

behdad: I see an alternative implementation that would remove all Extension tables during subsetting. WHich is what we want long term. Should I aim for that? Will do after you merge I guess.

PR #2706, opened 2020-10-10
Fix and tweak gen-tag-table.py

dscorbett:
These are a few unrelated changes to gen-tag-table.py.

The first is purely cosmetic for more consistency of language names in comments.

The second adds a mapping from "hy-arevmda" (Western Armenian) to 'HYE ' (Armenian). Previously, the variant subtag was unrecognized, so it mapped to HYE0 (Eastern Armenian) as plain "hy" does. "arevmda" is deprecated, but it should be recognized to support documents created in the 11 years it was recommended.

The third fixes the mapping from 'MNK ' to "man". Previously it was being mapped to "emk". It was a general bug but it happened to only affect this single mapping.

The fourth maps every individual language to its macrolanguage’s language system tag (if any). For example, "prs" maps to 'DRI ' as before, but now falls back to 'FAR '. Previously it would not do that if the individual language had its own language system tag. That was a bad idea because, given a document tagged as "prs" and a font that supports 'FAR ' but not 'DRI ', it would be better to apply the 'FAR ' lookups than not.

### DIFF
--- a/src/gen-tag-table.py
+++ b/src/gen-tag-table.py
@@ -591,7 +591,9 @@ class BCP47Parser (object):
 					elif not has_preferred_value and line.startswith ('Macrolanguage: '):
 						self._add_macrolanguage (line.split (' ')[1], subtag)
 				elif subtag_type == 'variant':
-					if line.startswith ('Prefix: '):
+					if line.startswith ('Deprecated: '):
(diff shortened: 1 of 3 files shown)


behdad:
Oooh. Thank you for the last piece (macrolanguages)!.

Issue #2711, opened 2020-10-11
hb-subset Unable to open output file


typoman:
I've installed harfbuzz using homebrew on OSX Catalina and it seems that hb-subset cannot write the font into the disk. I even used sudo but got the same error: 
```
> hb-subset /path/test.otf text
Unable to open output file
```
Does anyone know why this is happening?
Thanks

khaledhosny:
You need to specify output file path using `-o` option.

typoman:
I see! `-o` option solved it.  I got the command from the documentation. I guess that was old. Thank you very much Khaled!

PR #2713, opened 2020-10-12
[hb-subset] Improve error handling a bit

khaledhosny:
* Check that output-file option is actually set before trying to open it.
* Print file name and `errno` when opening the output file fails.
* Be more resilient when writing output file and use `ferror()` to check for errors.

Fixes https://github.com/harfbuzz/harfbuzz/issues/2711

### DIFF
--- a/util/hb-subset.cc
+++ b/util/hb-subset.cc
@@ -70,27 +70,27 @@ struct subset_consumer_t
 
   hb_bool_t
   write_file (const char *output_file, hb_blob_t *blob) {
-    unsigned int data_length;
-    const char* data = hb_blob_get_data (blob, &data_length);
+    unsigned int size;
+    const char* data = hb_blob_get_data (blob, &size);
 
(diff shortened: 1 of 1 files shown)


alerque (approved):

behdad:
If output-file is required, the option-parser should take care of that. No?

khaledhosny:
It should, but this part seems to be shared with other utils so I don’t know how to do this without affecting the others. May be it shouldn’t be shared:
```
$ hb-subset --help-output
Usage:
  hb-subset [OPTION…] [FONT-FILE] [TEXT]

Output destination & format options:
  -o, --output-file=filename                                          Set output file-name (default: stdout)
  -O, --output-format=format                                          Set output serialization format
```

Wording for `--output-file` is misleading, and `--output-format` is probably just ignored.

behdad:
Okay let me merge this and see what I can do to clean up.

Issue #2707, opened 2020-10-11
Questions about HB_BUFFER_MAX_OPS_FACTOR


dscorbett:
I am developing a Duployan font that uses many complex lookups to work around the lack of [spacing marks in OpenType](https://github.com/OpenType/opentype-layout/blob/63d169c92160832cf5b2333f28b559c80c9a4389/proposals/20151104-spacemark.md). Compiling HarfBuzz with `CPPFLAGS=-DHB_BUFFER_MAX_OPS_FACTOR=256` is sufficient for my own testing, but the font won’t work for anyone else, so I have some questions.

`max_ops` was added for #429 to “curb DoS fonts”. How I can tell if I accidentally made a DoS font? My font doesn’t have any recursive lookup loops and produces only linear growth (about 192 glyphs per character as of the latest commit). Is there a way for HarfBuzz to distinguish a legitimate complex font from a DoS font other than by counting operations?

If not, would it be acceptable to increase the default `HB_BUFFER_MAX_OPS_FACTOR` from 64? I don’t know if 256 will ultimately suffice for the final font; what is the highest value to which you would agree to increase it? I’m asking now, well before the font is complete, because I will have wasted a lot of time and effort if I can’t count on HarfBuzz supporting it. I’d like to know now so I can plan accordingly.

behdad:
I just increased it to 1024 and `HB_BUFFER_MAX_OPS_MIN` while we discuss hopefully finding a better mechanism.

dscorbett:
Wow! That is quite a high limit. Thanks!

behdad:
A legit font shouldn't have to fight our limits...

khaledhosny:
This is now breaking the `shape_fuzzer` test.

behdad:
> This is now breaking the `shape_fuzzer` test.

Ugh..

So, we shouldn't be running such tests under valgrind or other mechanisms that slow tests down significantly. We can also rely on oss-fuzz to discover most new issues for us.

I'm too lost in our current testing setup to know where to start, but can look into it later.

khaledhosny:
The test runner seems to default to a 2 second timeout:
https://github.com/harfbuzz/harfbuzz/blob/fa771a7f85d6d07ed64db438c8f07cd6b3f65e3e/test/fuzzing/run-shape-fuzzer-tests.py#L12

[...]

PR #2715, opened 2020-10-13
[test] Remove timeout from test runners

khaledhosny:
See https://github.com/harfbuzz/harfbuzz/issues/2707#issuecomment-707744079

This wasn’t inconsistent as well, `HB_TEST_SUBSET_FUZZER_TIMEOUT` defaulted to 12 in the test runner, but it was overridden to 50 in meson.build, and then meson has its own test timeout.


behdad:
Okay let's see what that does to the bots.

khaledhosny:
All bots are happy, except msys2 bots that has been mysteriously failing.

behdad:
Merging this since master is failing now.

PR #2722, opened 2020-10-17
[use] Skip WJ and ZWJ when clustering

dscorbett:
This is a follow-up to #1695 for U+2060 WORD JOINER and U+200D ZERO WIDTH JOINER. These default ignorable code points do not need any special handling. This is intentionally different from what the USE spec specifies: the spec requires many more dotted circles to be inserted, but the spec is wrong. The point of a character being default ignorable is so it gets ignored for rendering if the renderer doesn’t know what to do with it. A user should always be allowed to explicitly forbid a line break or request a ligature, even in contexts Microsoft didn’t think of.

The USE spec says that WJ is always in a cluster by itself. This is wrong because WJ is default ignorable. Because WJ has Joining_Type=Transparent, we can infer that it is not meant to have any visual effect when there is no line break. WJ is generally not useful within a cluster since a cluster does not normally have a line break opportunity anyway, but it is none of the shaper’s business if someone wants to use it.

The USE spec says that if there are two consecutive ZWJs, there is a cluster break between them. Why should there be? It would be acceptable, though unnecessary, to break the cluster in contexts like ⟨C, H, ZWJ, ZWJ, C⟩ but it is not acceptable to break the cluster when that would require a dotted circle. Again, the shaper should not be second-guessing the user’s questionable choices.

The USE spec says that ZWJ merges two adjacent clusters into a single cluster. However, according to MicrosoftDocs/typography-issues#232, DirectWrite does not implement it. I don’t want to go out on a limb supporting an otherwise unimplemented feature. Also, I think it is a bad feature: it would make character sequences like ⟨C, ZWJ, C, VPre⟩ be rendered as ⟨VPre, C, C⟩ or ⟨VPre, C_C⟩, which a reader would probably pronounce wrong. A ZWJ is not a virama.


behdad:
Thanks David!

PR #2723, opened 2020-10-17
[use] Minor clean-up of unused categories

behdad:
Unfortunately Fedora ships ragel 7.0.0, which, ehem, is "developmental"
version and breaks things badly. So I cannot even compile with this
change, let alone the lots of whitespace and other unnecessary changes.

The compile error is a legit unused-variable issue in the
ragel-generated code:

  CXX      libharfbuzz_la-hb-ot-shape-complex-use.lo
In file included from hb-ot-shape-complex-use.cc:202:
hb-ot-shape-complex-use-machine.rl: In function ‘void
find_syllables_use(hb_buffer_t*)’:
hb-ot-shape-complex-use-machine.rl:195:15: error: unused variable ‘act’
[-Werror=unused-variable]
  195 |   unsigned int act;
      |               ^~~

Someone without older ragel please amend this commit if you can.


rofl0r:
how about turning off -Werror ?

dscorbett:
I have amended the commit. I don’t have Fedora, but it’s my understanding that you can install old versions of packages if you can find them archived somewhere, like [this package of Ragel 6.8](https://koji.fedoraproject.org/koji/buildinfo?buildID=660479). For reference, I have Ragel 6.10. You could probably also build your own: see https://github.com/harfbuzz/harfbuzz/pull/1695#issuecomment-490614199 for how I built Ragel 7; building Ragel 6 is probably similar.

dscorbett (approved):

PR #2724, opened 2020-10-17
[CI] Don’t install ragel on circleci jobs

khaledhosny:
Latest ragel version is broken (https://github.com/adrian-thurston/ragel/issues/56), but we also don’t need to regenerate state machine on CI jobs.


behdad:
Oof. All greens!

PR #2725, opened 2020-10-17
[use] Merge IND and Rsv classes into O

dscorbett:
IND, Rsv, and O are effectively identical, so they might as well all be O.


behdad:
Lgtm. Thanks.

PR #2726, opened 2020-10-17
Avoid duplication of categories between C++ and Ragel

dscorbett:
A bunch of Ragel machines for complex shapers had comments like “Same order as enum indic_category_t.  Not sure how to avoid duplication.” I put the enums into their own headers and included them in both C++ and Ragel. This relies on the difference between the two languages’ include statements so that the same enum value is called e.g. `OT_ZWJ` in C++ and `ZWJ` in Ragel.

I merged `khmer_category_t` and `myanmar_category_t` into `indic_category_t`. I guess this wasn’t strictly necessary. It required renaming Myanmar’s `OT_V` (visarga or Shan tone) because it clashed with Indic’s `OT_V` (independent vowel).


khaledhosny:
Unrelated, but I don’t know what is wrong with this codecov thing.

behdad:
Thank you!

I wasn't aware of Ragel's `import` mechanism. It's pretty sloppy but works. However, I'm concerned that removing the list of categories used by each machine from the machine's definition will become a headache.

I like to explore the alternative: use Ragel's `export` mechanism. Let me give it a try.

behdad:
Err. Merged by mistake :(.

behdad:
I made https://github.com/harfbuzz/harfbuzz/pull/2836 as proposed alternate solution.

PR #2728, opened 2020-10-23
[ENOMEM] Drop layout table if layout lookup accelerator allocation failed

drott:
If the table is kept, when the table is present but only the accelerator
allocation failed, then we run into nullptr dereferences downstream in
hb_ot_map_t::apply when trying to access proxy.accels[i].

To fix this, instead of only setting lookup_count of accelerator_t
itself, drop the whole table to avoid hb-ot-map construction assuming
that the lookups will work correctly despite accelerator allocation
having failed.

Fixes:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24490
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24473

### DIFF
--- a/src/hb-ot-layout-gsubgpos.hh
+++ b/src/hb-ot-layout-gsubgpos.hh
@@ -3378,7 +3378,11 @@ struct GSUBGPOS
 
       this->accels = (hb_ot_layout_lookup_accelerator_t *) calloc (this->lookup_count, sizeof (hb_ot_layout_lookup_accelerator_t));
       if (unlikely (!this->accels))
+      {
 	this->lookup_count = 0;
+	hb_blob_destroy (this->table.get_blob ());
+	this->table = hb_blob_get_empty ();
+      }
 
       for (unsigned int i = 0; i < this->lookup_count; i++)
 	this->accels[i].init (table->get_lookup (i));


drott:
Not 100% sure this is the right fix, but to me it seems the alternative fix would be to have hb-ot-map construction or shape plan construction double-check the existence of `accels` in the accelerator class, which seems it's not intended.

drott:
CC @garretrieger

behdad:
Thanks. This definitely works in principle.

The previous design was that any part that needs access to the accelerator will use lookup_count and as such not try to access further. But that does indeed seem error-prone.

What's the code path that was faulty?

Maybe we don't need to keep separate lookup_count now.

drott:
> What's the code path that was faulty?

I'm not so familiar with where it goes wrong in the first place, and I only have a rough theory. (You have access to the stack trace in OSS-fuzz, do you?) but IIUC my observation is that hb_ot_map_t construction and shape plan determination use only values from GSUBProxy.table, not from the accelerators, and thus come to look at different number for lookup_count than what's available in the accelerator (=reset to 0) when the allocation fails. In other words, the lookup_count of the accelerators goes out of sync with what's read directly from the the .table structs.

drott:
> The previous design was that any part that needs access to the accelerator will use lookup_count and as such not try to access further. But that does indeed seem error-prone.

PS: Yes, in that sense https://github.com/harfbuzz/harfbuzz/blob/5091ea7e247b118009c719076650a93b3e2dde6d/src/hb-ot-layout.cc#L1800 apply(), and apply_forward() and friends are not checking the lookup_count before accessing the `accels`. Perhaps checks for that could be added there and that might be an alternative fix? The code here looked like it was on the hot path and we would want to do checks earlier, that's why I was looking for a sport where to do drop the table / mark it as invalid up front.

drott:
I didn't know #2627 existed as well, which guards the access to the accelerators. @behdad please advice, which approach would be preferred.

behdad:
Thanks @drott. Can you add the tests from #2627? No idea about the bot change there...

PR #2729, opened 2020-10-23
[autotools] Require pkg-config 0.28

### LINKED ISSUE #1565: pc file misses dependency
If harfbuzz creates a .pc file and uses/depends on freetype, it should call this out in its .pc file, see #720

khaledhosny:
Follow up 0615c7b64b63400b2a70a1a6dd8a57d0b0fff10c, which fixes #1565.

This seems to be the version that fixed the circular dependency, based on the ChangeLog:
https://cgit.freedesktop.org/pkg-config/commit/?h=pkg-config-0.28&id=74ceac54ef6f9247c00f08eecd8cca811a3c5934


khaledhosny:
The last to commits are not strictly related to this PR, but they are here because CI was otherwise failing.

alerque (approved):
This LGTM. Even Ubuntu Xenial has 0.29, so requiring 0.28 is not unreasonable.

Not to block this because I'll take your word for something I don't understand, but I'm slightly confused the whole XCode 9 test removal and your comment doesn't clear it up for me. Isn't the point to keep test coverage as far as possible in the *older* platform direction as well well as recent ones?

khaledhosny:
> Not to block this because I'll take your word for something I don't understand, but I'm slightly confused the whole XCode 9 test removal and your comment doesn't clear it up for me. Isn't the point to keep test coverage as far as possible in the _older_ platform direction as well well as recent ones?

The was supposed to be running macOS 10.12.6, but according to https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions XCode 9.3.1 images is actually running macOS 10.13.3 and we don’t have 10.13.3 specific tests on test/shaping/data/in-house/tests/macos.tests, so this job was effectively not testing any macOS fonts. The replacement XCode 9.4.1 images run the same macOS version.

alerque:
> > Not to block this because I'll take your word for something I don't understand, but I'm slightly confused the whole XCode 9 test removal and your comment doesn't clear it up for me. Isn't the point to keep test coverage as far as possible in the _older_ platform direction as well well as recent ones?
> 
> The was supposed to be running macOS 10.12.6, but according to [circleci.com/docs/2.0/testing-ios/#supported-xcode-versions](https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions) XCode 9.3.1 images is actually running macOS 10.13.3 and we don’t have 10.13.3 specific tests on test/shaping/data/in-house/tests/macos.tests, so this job was effectively not testing any macOS fonts. The replacement XCode 9.4.1 images run the same macOS version.

Oh, got it! Then yes by all means ditch it ;-)

Issue #2730, opened 2020-10-24
Clarify whether buffer context must have same script/language


astiob:
https://harfbuzz.github.io/adding-text-to-the-buffer.html says:

> For longer text runs, such as full paragraphs, it might be tempting to only add smaller sub-segments to a buffer and shape them in piecemeal fashion. […] The safest approach is to add all of the text available, then use `item_offset` and `item_length` to indicate which characters you want shaped, so that HarfBuzz has access to any context.

How much does this apply when the text contains multiple scripts and languages? Is it right to split on a script boundary and pass each part as context to the other, or must each script part be placed in the shaping buffer in full isolation?

(This is a question, but it would be great to have an answer in the manual, rather than only on the issue tracker.)

khaledhosny:
Ideally, the full paragraph text is passed to HarfBuzz each time with the specific run boubdry being shaped, regardless of the script/language/direction/font.

khaledhosny:
I thonk the manual text is clear enough, but feel free to suggest improvements (pregerably as a PR).

Eli-Zaretskii:
If passing the full paragraph is preferred, does that mean the direction, the script, and the language should *not* be set, but instead left to HarfBuzz's discretion?  What if HarfBuzz decides on one of these incorrectly?

khaledhosny:
The [`hb_buffer_add_*()`](https://harfbuzz.github.io/harfbuzz-hb-buffer.html#hb-buffer-add-codepoints) functions take a text string and a length, this should ideally be a full paragraph of text, then item offset and length, this is the single script/language/direction/font run that is being shaped.

So instead of itemizing then splitting paragraph into items and passing only the text of that item to HarfBuzz, you itemize then store item offsets and lengths and pass to HarfBuzz the full text, then item offset and length one by one.

The first way still works, but it is a sub-optimal use of HarfBuzz API and you loss some functionality, like HarfBuzz detecting the proper Arabic joining behavior of the item boundary based on the context, so when say “ال**ع**ربية” is shaped (with the ayn set in bold or a fallback font), you don’t get a unjoined word (Chrome and Firefox will show the word above differently, possibly illustrating this very issue of HarfBuzz API use).

Eli-Zaretskii:
[...]

PR #2733, opened 2020-10-24
[docs] Describe buffer context applicability more explicitly

astiob:
Make it abundantly clear that the context's script etc. don't matter.

Fixes https://github.com/harfbuzz/harfbuzz/issues/2730


khaledhosny:
Thanks!

PR #2735, opened 2020-10-27
Don’t map BCP 47 tags to coincidentally similar language system tags

dscorbett:
If nothing else works, `hb_ot_tags_from_language` falls back to uppercasing a three-letter BCP 47 language subtag in the hope that it might match a language system tag. For example, "ach" falls back to matching 'ACH '. That is a fine optimization, but there are some false positives. For example, "far" falls back to matching 'FAR '; however, the former is Fataleka and the latter is Persian. Such a mapping does not make sense. I have updated gen-tag-table.py to detect these tricky language system tags and preemptively map them in `hb_ot_tags_from_complex_language` to nothing.

### DIFF
--- a/src/gen-tag-table.py
+++ b/src/gen-tag-table.py
@@ -47,6 +47,8 @@ def expect (condition, message=None):
 			raise AssertionError
 		raise AssertionError (message)
 
+DEFAULT_LANGUAGE_SYSTEM = ''
+
 # from https://www-01.sil.org/iso639-3/iso-639-3.tab
 ISO_639_3_TO_1 = {
 	'aar': 'aa',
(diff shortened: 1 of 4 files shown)


behdad:
Why does your diff includes two-letter tags?

Also that "if" hierarchy is becoming too long for my performance taste...

dscorbett:
> Why does your diff includes two-letter tags?

Do you mean as in `if (lang_matches (&lang_str[1], "ba"))`? That’s not a two-letter tag: it’s the last two letters of a three-letter tag. That is because of an optimization you suggested, when I first added this function, to switch on the first character.

> Also that "if" hierarchy is becoming too long for my performance taste...

I could do another layer of switching, on the second character; or I could add these false positives to `ot_languages`, mapping them to `HB_TAG('d','f','l','t')` or `HB_TAG_NONE` or something like that, and rely on the binary search to be fast.

eternal-winter:
> That is because of an optimization you suggested, when I first added this function, to switch on the first character.

Ah. Got it.

> or I could add these false positives to `ot_languages`, mapping them to `HB_TAG('d','f','l','t')` or `HB_TAG_NONE` or something like that, and rely on the binary search to be fast.

Yes, that's what I originally expected to see in the patch based on your description.

Thanks.

behdad:
> That is because of an optimization you suggested, when I first added this function, to switch on the first character.

Ah. Got it.

> or I could add these false positives to `ot_languages`, mapping them to `HB_TAG('d','f','l','t')` or `HB_TAG_NONE` or something like that, and rely on the binary search to be fast.

Yes, that's what I originally expected to see in the patch based on your description.

Thanks.

PR #2744, opened 2020-11-15
Bump minimum meson version to 0.53.0

CoffeeFlux:
From a local build:
```
|WARNING: Project specifies a minimum meson_version '>= 0.47.0' but uses features which were added in newer versions:
| * 0.50.0: {'install arg in configure_file'}
| * 0.53.0: {'Fallback without variable name'}
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,5 +1,5 @@
 project('harfbuzz', 'c', 'cpp',
-  meson_version: '>= 0.47.0',
+  meson_version: '>= 0.53.0',
   version: '2.7.2',
   default_options: [
     'cpp_eh=none',          # Just to support msvc, we are passing -fno-rtti also anyway


CoffeeFlux:
Looking elsewhere, seems this lower version is intentional. Sorry for the noise.

alerque:
Yes, the lower version is intentional. We need to support as old a version as possible so that distro adoption can proceed without holdups due to older meson availability.

That being said there shouldn't be messages about using features from newer versions because we shouldn't be using anything newer than what we claim as the minimum supported version!

khaledhosny:
The first warning was introduced in https://github.com/harfbuzz/harfbuzz/pull/2668, but according to https://github.com/harfbuzz/harfbuzz/pull/2668#issuecomment-689296225 it can be ignored.

I don’t get the other warning here and I have no idea what is triggering it.

PR #2746, opened 2020-11-16
libpng.wrap: update to 1.6.37

alatiera:


khaledhosny:
This breaks building Cairo subproject.

khaledhosny (changes requested):
This breaks building Cairo subproject.

alatiera:
ah indeed, didn't notice harfbuzz was still pointing at a fork.

The meson port got merged upstream and we also updated libpng in cairo yday. 

I added a commit to switch cairo and fontconfig wraps to the upstream repos.

alatiera:
blah, cairo from git looks like it requires meson 0.54+ which breaks the win32 build.

```
subprojects/cairo/meson.build:1:0: ERROR: Meson version is 0.53.2 but project requires >= 0.54.0
```

khaledhosny:
For Cairo see #2700.

alatiera:
indeed, maybe @xclaesse  could bump libpng and included in #2700

alatiera:
Now part of #2700

PR #2747, opened 2020-11-17
Map more ISO 15924 variants to `hb_script_t`s

dscorbett:


behdad:
Also, differentiating between Traditional vs Simplified Chinese via a script tag is definitely common practice. Making it impossible is not good :).

PR #2748, opened 2020-11-17
meson.build: check for pthread.h

ffontaine:
Check for pthread.h otherwise the build will fail with some toolchains that have libphtread.so but not pthread.h:

```
Run-time dependency threads found: YES

../src/hb-mutex.hh:53:10: fatal error: pthread.h: No such file or directory
 #include <pthread.h>
          ^~~~~~~~~~~
```

Moreover, fix detection of pthread fallback

Fixes:
 - http://autobuild.buildroot.org/results/70c98e89b1d5e5b651d1f6928dc53f465103f57a

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -284,13 +284,15 @@ endif
 # threads
 thread_dep = null_dep
 if host_machine.system() != 'windows'
-  thread_dep = dependency('threads', required: false)
+  if cpp.has_header('pthread.h')
+    thread_dep = dependency('threads', required: false)
+  endif
 
   if thread_dep.found()
     conf.set('HAVE_PTHREAD', 1)
   else
     check_headers += ['sched.h']
-    check_funcs += ['sched_yield', {'link_with': 'rt'}]
+    check_funcs += [['sched_yield', {'link_with': 'rt'}]]
   endif
 endif
 


khaledhosny:
Shouldn’t `dependency('threads')` be doing this already? That sounds like a bug/limitation of it or may be we shouldn’t be using it if it does not do what we need despite being the [recommended way](https://mesonbuild.com/howtox.html#enable-threads).

behdad:
Anyone want to respond to Khaled's point?

Patch also need updating since the `sched_yield()` use has since been removed.

ffontaine:
I don't know if this is a meson or a toolchain bug as having `libpthread.so` but not `pthread.h` seems a bit strange. However, this is still failing with latest meson (0.57.1)

xclaesse:
That's a meson bug indeed, I opened a quick PR but it could be more complicated than it sounds: https://github.com/mesonbuild/meson/pull/8610.

OOC, harfbuzz always requires pthread.h, it can't use winthreads?

behdad:
> OOC, harfbuzz always requires pthread.h, it can't use winthreads?

We use pthread.h for mutex only. And we have a native win32 impl using `InitializeCriticalSection` et al if `defined(_WIN32)`.

behdad:
Should we close this? We're moving towards using C++11 std::mutex if nothing else is found:
https://github.com/harfbuzz/harfbuzz/pull/3100

behdad:
Or at least someone please update it and recommend merging it if it's still relevant. Thanks.

khaledhosny:
AFAIC, this is a meson issue. `dependency('threads', required: false)` should automatically detect pthreads, we shouldn’t have to second guess it.

behdad:
> AFAIC, this is a meson issue. `dependency('threads')` should automatically detect pthreads, we shouldn’t have to second guess it.

Works for me.

PR #2755, opened 2020-11-24
meson: use meson.current_source_dir() instead of @CURRENT_SOURCE_DIR@

Myaamori:
@CURRENT_SOURCE_DIR@ is not listed as a valid string substitution for custom targets in the Meson reference, and in practice it does not get substituted when using the vs2019 backend.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -299,7 +299,7 @@ else
       build_by_default: true,
       input: rl,
       output: hh,
-      command: [ragel_helper, '@OUTPUT@', '@CURRENT_SOURCE_DIR@', '@INPUT@'],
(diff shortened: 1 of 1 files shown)


alerque:
This sounds right, but can we confirm from release docs when `meson.current_source_dir()` was added? Remember we're trying to keep supporting pretty old Meson versions here.

Myaamori:
It seems to have been present since version 0.14.0, hopefully that's old enough: https://github.com/mesonbuild/meson/commit/9318203257c85209b5c60a9dc2cb59c5feaae23d

alerque (approved):

CoffeeFlux:
Anything blocking a merge here? Would be great to get this in so we no longer have to work around it on our end.

PR #2762, opened 2020-12-09
CMake: Define HAVE_GOBJECT when harfbuzz-gobject is built

fanc999:
Hi,

From the commit message (I am aware the CMake build system is going out, but, anyways :)):

<i>Without this, the built library for harfbuzz-gobject is unuseful.</i>

The reason is because the code in there are only built if `HAVE_GOBJECT` is defined during compilation, so without that, we are more or less building an empty harfbuzz-gobject dll|lib|so.

With blessings, thank you!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -302,6 +302,7 @@ if (WIN32 AND HB_HAVE_DIRECTWRITE)
 endif ()
 
 if (HB_HAVE_GOBJECT)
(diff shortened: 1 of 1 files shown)


behdad:
Lgtm but I let someone knowing cmake to merge (@khaledhosny / @ebraminio?)

ebraminio:
lgtm, merged, thanks

PR #2764, opened 2020-12-11
Make failing allocator an opt-in for fuzzer runs

ebraminio:
Now that basic issues regarding malloc limits are fixed let's
turn failing allocator to an opt-in for fuzzer runs to separate
their runs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,6 +134,12 @@ jobs:
       # msan, needs --force-fallback-for=glib,freetype2 also which doesn't work yet but runs fuzzer cases at least
       - run: rm -rf build && meson build --default-library=static -Db_sanitize=memory --buildtype=debugoptimized --wrap-mode=nodownload -Dauto_features=disabled -Dtests=enabled -Dexperimental_api=true
       - run: ninja -Cbuild -j8 && meson test -Cbuild --print-errorlogs | asan_symbolize | c++filt
+      # asan fuzzer
+      - run: rm -rf build && CXXFLAGS="-fsanitize=address,fuzzer-no-link" meson build --default-library=static -Dfuzzer_ldflags="-fsanitize=address,fuzzer" --buildtype=debugoptimized --wrap-mode=nodownload
+      - run: ninja -Cbuild -j8 && meson test -Cbuild --suite fuzzing --print-errorlogs | asan_symbolize | c++filt
+      # asan fuzzer+failing-alloc
+      - run: rm -rf build && CXXFLAGS="-fsanitize=address,fuzzer-no-link" meson build --default-library=static -Dfuzzer_ldflags="-fsanitize=address,fuzzer" -Dfailing_alloc=true --buildtype=debugoptimized --wrap-mode=nodownload
+      - run: ninja -Cbuild -j8 && meson test -Cbuild --suite fuzzing --print-errorlogs | asan_symbolize | c++filt
       # test -std=c++2a and -Weverything of nightly clang builds
       - run: clang -c src/harfbuzz.cc src/hb-subset*.cc -DHB_NO_MT -Werror -std=c++2a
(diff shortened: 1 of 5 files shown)


alerque (approved):

behdad:
I don't understand what this is doing. But with @garretrieger we agreed that we want to make subset library to a scheme that doesn't support failing malloc. Possibly using exceptions to just jump to the top and fail.

Also wondering what our major clients do with memory allocation failures; are those handled, or crashed? @jfkthame @drott @nona-google

ebraminio:
This disables the feature while keeping the found cases tested on CI and makes it possible to enable it again on oss-fuzz if someone is interested. I have a bit concern on possibility of some code paths aren't tested enough in this mode, fortunately there is an independent fuzzer on Chrome project which doesn't do the allocation failure testing yet better to have this disabled now that basics issues regarding to such issues are resolved.

drott:
> Also wondering what our major clients do with memory allocation failures; are those handled, or crashed?

In what way do you mean handled? As in overriding HB_CUSTOM_ALLOC and then checking in those if allocations succeeded? Or in general, in Chrome? I think in Chrome V8 has some OOM handling, layout code usually crashes for OOM. Is that what you meant?

behdad:
> In what way do you mean handled?

My question is: does the malloc() called by HB in Chrome ever return NULL? And if yes, is it expected that HB not crash?

We currently don't crash, and recover. But for the subset that's a pain. What we want to do is either be allowed to just crash in subsetting code if malloc failed, or possibly use exceptions and catch at top API boundary.

cc @bungeman

bungeman:
I'm familiar with the mess that is [Skia's allocator in Chromium](https://source.chromium.org/chromium/chromium/src/+/master:skia/ext/SkMemory_new_handler.cpp;drc=c86618b300799fd70be83a72afbe7e15f124493d), but I'm less sure about the general one. The [documentation for Chromium's allocator](https://chromium.googlesource.com/chromium/src/+/c1a1483fd0d3f047e811a54832f44b4fe2b607c5/base/allocator/) implies that by default it will try to crash if malloc returns nullptr. I'd have to look a bit more to see under which circumstances the Chromium allocator itself will force a crash if a malloc cannot be satisfied.

It generally seems to be mostly fine for libraries to crash on small allocations if malloc returns nullptr (really out of memory, so no real progress is going to happen anyway), but it is somewhat expected that large allocations will be checked (couldn't reserve space for that 1GB image, try to carry on). Note that Skia does something like this, naming them `sk_malloc_throw` and `sk_malloc_canfail`, as well as using 'new' when allocation failure should abort. Of course, it's nice to always fail gracefully.

I would be fine with HarfBuzz using exceptions for this internally if it works well for HarfBuzz. We'd need to update the build flags on the Chromium side, but it shouldn't be difficult. If the trade off is between HarfBuzz using exceptions and returning a nice "the subset didn't work" result and the alternative being a crash, I'd prefer the exceptions. Even if Chromium choses to crash, other more normal applications would probably prefer a graceful failure.

behdad:
> I would be fine with HarfBuzz using exceptions for this internally if it works well for HarfBuzz. We'd need to update the build flags on the Chromium side, but it shouldn't be difficult. If the trade off is between HarfBuzz using exceptions and returning a nice "the subset didn't work" result and the alternative being a crash, I'd prefer the exceptions. Even if Chromium choses to crash, other more normal applications would probably prefer a graceful failure.

Right. I feel the same about it.

The current "any allocation can fail" has proved to be too hard to handle gracefully in the subsetter code. In the shaper we can get away, because we're mostly just consuming readonly data. But in the subsetter we're constantly creating objects...

khaledhosny:
What is the status of this, merge or close?

behdad:
I'm hesitant to change our current setup which other clients (clusterfuzz, oss-fuzz) depend on.

What I suggest we do is that within the fuzzer harness, disable failing-alloc for 50% of the runs.

behdad:
I created https://github.com/harfbuzz/harfbuzz/pull/3701 instead.

Issue #2501, opened 2020-06-25
Allocation failure causes a dereference of NULL


sebras:
More testing of memory allocation failing using clang+ASAN revealed a `NULL` dereference in HarfBuzz 2.6.4, which still seems to be present in git HEAD. This is the fourth and final issue I've discovered this way.

The crash occurs due to dereferencing of `NULL`  here:

```
    #0 0xd312fc in hb_ft_font_changed harfbuzz/src/hb-ft.cc:712:30
    #1 0xd310ee in hb_ft_font_create harfbuzz/src/hb-ft.cc:701:3
```

To get into this situation `calloc()` must first return `NULL` here:

```
#0  0x0000000000d22887 in hb_object_create<hb_font_funcs_t> () at harfbuzz/src/hb-object.hh:227
#1  0x0000000000d227fd in hb_font_funcs_create () at harfbuzz/src/hb-font.cc:532
#2  0x0000000000d3aead in hb_ot_font_funcs_lazy_loader_t::create () at harfbuzz/src/hb-ot-font.cc:257
#3  0x0000000000d3ae21 in hb_data_wrapper_t<void, 0u>::call_create<hb_font_funcs_t, hb_ot_font_funcs_lazy_loader_t> (this=0x4256220 <static_ot_funcs>) at harfbuzz/src/hb-machinery.hh:171
#4  0x0000000000d3ad86 in hb_lazy_loader_t<hb_font_funcs_t, hb_ot_font_funcs_lazy_loader_t, void, 0u, hb_font_funcs_t>::get_stored (this=0x4256220 <static_ot_funcs>) at harfbuzz/src/hb-machinery.hh:222
#5 0x0000000000d3acf5 in hb_lazy_loader_t<hb_font_funcs_t, hb_ot_font_funcs_lazy_loader_t, void, 0u, hb_font_funcs_t>::get_unconst (this=0x4256220 <static_ot_funcs>) at harfbuzz/src/hb-machinery.hh:247
#6 0x0000000000d3a163 in _hb_ot_get_font_funcs () at harfbuzz/src/hb-ot-font.cc:296
#7 0x0000000000d3a0e9 in hb_ot_font_set_funcs (font=0x60d0000006c8) at harfbuzz/src/hb-ot-font.cc:309
#8 0x0000000000d2b2c2 in hb_font_create (face=0x6140000c8648) at harfbuzz/src/hb-font.cc:1380
[...]

Issue #2493, opened 2020-06-24
NULL pointer access due to memory allocation failing...


sebras:
While testing what happens when memory allocations fail using clang+ASAN I stumbled upon a crash in HarfBuzz 2.6.4, which still seems to be present in git HEAD.

The backtrace relevant for HarfBuzz for the crash is here, where it dereferences the NULL pointer argument passed to it:

```
    #0 0xd3019f in _hb_ft_font_destroy(void*) harfbuzz/src/hb-ft.cc:120:7
    #1 0xd2c525 in hb_font_set_funcs harfbuzz/src/hb-font.cc:1673:7
    #2 0xd30d8e in _hb_ft_font_set_funcs(hb_font_t*, FT_FaceRec_*, bool) harfbuzz/src/hb-ft.cc:559:3
    #3 0xd30c45 in hb_ft_font_create harfbuzz/src/hb-ft.cc:698:3
```

To get into this state is a bit complicated as `calloc()` needs to return `NULL` in multiple locations, the first being here:

```
#0  0x00000000010dffa7 in hb_object_create<hb_blob_t> () at harfbuzz/src/hb-object.hh:227
#1  0x00000000010dfca9 in hb_blob_create (data=0x384c475 "", length=86828, mode=HB_MEMORY_MODE_READONLY, user_data=0x61a000008488, destroy=0x0) at harfbuzz/src/hb-blob.cc:94
#2  0x0000000000d3053d in hb_ft_face_create (ft_face=0x61a000008488, destroy=0x0) at harfbuzz/src/hb-ft.cc:615
#3  0x0000000000d30c1d in hb_ft_font_create (ft_face=0x61a000008488, destroy=0x0) at harfbuzz/src/hb-ft.cc:695
```

Next `calloc()` returns `NULL` again here:

```
#0  0x00000000010e2e0e in _hb_face_for_data_closure_create (blob=0x17e46c0 <_hb_NullPool>, index=0) at harfbuzz/src/hb-face.cc:145
#1  0x00000000010e2ca1 in hb_face_create (blob=0x17e46c0 <_hb_NullPool>, index=0) at harfbuzz/src/hb-face.cc:205
#2  0x0000000000d30586 in hb_ft_face_create (ft_face=0x61a000008488, destroy=0x0) at harfbuzz/src/hb-ft.cc:619
[...]

PR #2770, opened 2020-12-21
Check input info length when assigning next glyph

drott:
Fixes oss-fuzz:28234 but needs further review and approval from someone
who knows this code better.

### DIFF
--- a/src/hb-buffer.hh
+++ b/src/hb-buffer.hh
@@ -262,7 +262,7 @@ struct hb_buffer_t
   void
   next_glyph ()
   {
-    if (have_output)
+    if (have_output && idx < len)
     {
       if (out_info != info || out_len != idx)
       {


behdad:
Please put the full issue URL (even for our own issues) in the commit log.  Also, please include your assessment of what the problem is. For example, here's two excellent reports by someone new to the project:

https://github.com/harfbuzz/harfbuzz/issues/2501
https://github.com/harfbuzz/harfbuzz/issues/2493

The change doesn't look right to me. `idx < len` shall hold always when `next_glyph()` is called. So we need to figure out why we get into a position that this is violated.

drott:
Yes, sorry, this was a bit brief and incomplete. I don't have the full analysis yet - my intention was to nudge and start a discussion on https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28234 - as this issue has been reported early in December and has not been picked up so far. 

Unfortunately, I am not sure which of the allocations fails yet, but it looks like at the point in question, `len` may not be in sync with the length of `info`. I tried setting a breakpoint in `rr` on the `mov 0x0, %eax` instruction of all the allocation `hb_*alloc` functions (i.e. the case when they return `nullptr`) after building `hb-shape-fuzzer` with `-DHB_CUSTOM_MALLOC` but those did not get hit and the build was not build with `llvm` and not with `-fsanitize`. Is there an easy way to locally build with `libFuzzer` and the MSAN or other sanitisation options?

behdad:
I haven't built any of that myself. Let's see if @ebraminio can help in.

ebraminio:
`CXXFLAGS="-fsanitize=address,fuzzer-no-link" meson fuzzbuild --default-library=static -Dfuzzer_ldflags="-fsanitize=address,fuzzer" -Dexperimental_api=true && ninja -C fuzzbuild test/fuzzing/hb-{shape,draw,subset,set}-fuzzer && fuzzbuild/test/fuzzing/hb-shape-fuzze YOURTESTCASE` this builds with failing alloc

behdad:
This *should* be fixed by the audit I did a couple weeks ago.

behdad:
Yep. Status of referenced oss-fuzz issue is Verified.

drott:
Thanks for the follow-up.

PR #2775, opened 2020-12-23
meson: avoid regenerating hb-version.h unnecessarily

CoffeeFlux:
The file is a dependency for the library as a whole, so previously editing the Meson build files at all would force a full rebuild. This will only rebuild it if the version has changed.


CoffeeFlux:
msys2 build failure looks unrelated.

PR #2776, opened 2020-12-23
2.7.3

khaledhosny:
Not sure if NEWS entries are good, I didn’t really understand most of the changes so the entries are terse without much specifics, hoping people would check the commit log for further details.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,26 @@
+Overview of changes leading to 2.7.3
+Wednesday, December 23, 2020
+====================================
+- Update USE shaper to 2020-08-13 specification, and other improvements.
+- Don’t disable liga feature in myanmar shaper, to match Uniscribe.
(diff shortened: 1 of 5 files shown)


alerque:
I'll try to give it a gander today. I'll also be checking in when the actually release tag is placed to make sure the CI based release process goes smoothly. This will be the first release since I setup automatic source release tarballs/zips being posted to releases automatically from CI…

behdad:
LGTM. THanks.

alerque (approved):

khaledhosny:
> This will be the first release since I setup automatic source release tarballs/zips being posted to releases automatically from CI…

Seems to work fine. If possible, though, I’d prefer if the ci made a draft release so one can edit it and add the text from the MEWS file before publishing the release. Also RELEASING.md needs to be updated.

PR #2778, opened 2020-12-24
meson: update minimum meson_version to 0.50.0

heitbaum:
As per the log error during meson.build, update the minimum to 0.50.0

```
 * 0.50.0: {'install arg in configure_file'}
```


khaledhosny:
See https://github.com/harfbuzz/harfbuzz/pull/2744

Issue #2779, opened 2020-12-25
2.7.3 configure doesn't support --enable-introspection


rworkman:
and this breaks quite a few things depending on HarfBuzz-0.0.typelib remaining in place from 2.7.2 to 2.7.3

rworkman:
Solution appears to be "use meson to build" and that's fine *if* there's some notification that it's necessary...

alerque:
We started recommending distros packagers switch to Meson in the 2.7.0 announcement. We don't have an end date for completely deprecating autotools support yet, but it's been mentioned many times.

I don't think the `--enable-introspection` breakage was intentional though. I agree such a removal would have deserved an announcement. However I don't think there was a removal. Possibly a regression, but one can't exactly expect announcement of unintentional regressions.

Also I'm not convinced the only solution is to use Meson (although that is recommended). Discussion on #1202 suggests to me that you *can* still do this with autotools, but that you need to explicitly enable GObject support too. Given how many other issues mention this with various other iterations of confusion it does sound like at the very least documentation about this feature might be lacking.

rworkman:
Seems like a good time to officially kill the autotools support :-)

khaledhosny:
This is fixed in 2.7.4 release.

PR #2781, opened 2020-12-26
[ci] Install gobject-introspection on distcheck job

khaledhosny:
To fix missing --enable-introspection from release tarballs.

Fixes https://github.com/harfbuzz/harfbuzz/issues/2779

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,7 +54,7 @@ jobs:
     executor: autotools-executor
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


khaledhosny:
@alerque any idea where to download the dist archive (short of tagging a new release) to verify this actually works?

alerque:
@khaledhosny Hmmm, yes ... somewhere. Stand by.

I can't find it. I thought when I was setting this up that there were artifacts being attached to the *distcheck* jobs [on all Circle CI runs](https://app.circleci.com/pipelines/github/harfbuzz/harfbuzz) but that does not appear to be the case. I'll look into getting them added because that would be a useful thing to have on hand for debugging.

PR #2788, opened 2020-12-27
[docs] Fix generated docs for deprecated symbols

khaledhosny:
Without ignoring `HB_DEPRECATED`, most deprecated symbols get missing from documentation. Ignoring `HB_DEPRECATED_FOR` does not work, but this seems to be a GTK-Doc bug.


khaledhosny:
https://gitlab.gnome.org/GNOME/gtk-doc/-/issues/134

Issue #1810, opened 2019-07-02
Merge various `insert_dotted_circles` copies


behdad:
The four Indic-like shapers each have a copy.  They're mostly the same.  We should merge them to save binary size.

behdad:
Closing in preference to https://github.com/harfbuzz/harfbuzz/pull/2789

PR #2789, opened 2020-12-28
Refactor shared code in syllabic shapers

simoncozens:
This is an idea, and I'm submitting it here as a draft, to get a read on whether or not we want to go in this direction.

While implementing fontFeatures.shaperLib, I realised how much repeated code there is in Harfbuzz to handle the different syllable-based complex shapers (Indic, Myanmar, Khmer, USE). All of these things are actually structurally (and in terms of actual lines of code) very similar. In Python we can handle that with inheritance from a common base class. In C it's a bit harder because all the syllable type enums are different, but...

So far, this PR:

* Proposes a common `enum syllable_type_t` shared by all syllabic shapers.
* Refactors all the `insert_dotted_circles_...` into a common routine.
* Provides a "base class" into which we can also put other common routines (`setup_syllables_...`, `reorder_...`, `reorder_syllable...`)

We could go a lot further done this road of refactoring if it's desirable. I think we could have much of the syllabic shaper behaviour driven by a data structure similar to the `hb_ot_complex_shaper_t`: e.g. a function pointer to reorder each syllable type, the reph class to insert dotted circles, the kinds of position/category data structure to allocate, and so on.

I haven't moved any other common code over yet, because I want to see if we feel this is a good thing to be doing before potentially wasting time on implementation. I can see benefits in keeping each shaper completely independent of the others, but it does mean a lot of repetition with all of the complexity and maintenance nightmare that implies.

### DIFF
--- a/src/hb-ot-shape-complex-use.cc
+++ b/src/hb-ot-shape-complex-use.cc
@@ -31,6 +31,7 @@
 #ifndef HB_NO_OT_SHAPE
 
 #include "hb-ot-shape-complex-use.hh"
+#include "hb-ot-shape-complex-syllabic.hh"
 #include "hb-ot-shape-complex-arabic.hh"
 #include "hb-ot-shape-complex-arabic-joining-list.hh"
 #include "hb-ot-shape-complex-vowel-constraints.hh"
@@ -186,19 +187,6 @@ data_destroy_use (void *data)
   free (data);
 }
 
-enum use_syllable_type_t {
-  use_independent_cluster,
-  use_virama_terminated_cluster,
-  use_sakot_terminated_cluster,
-  use_standard_cluster,
-  use_number_joiner_terminated_cluster,
-  use_numeral_cluster,
-  use_symbol_cluster,
-  use_hieroglyph_cluster,
-  use_broken_cluster,
-  use_non_cluster,
-};
-
 #include "hb-ot-shape-complex-use-machine.hh"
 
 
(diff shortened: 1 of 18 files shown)


behdad:
Thanks Simon. I'm deeply into code-sharing. So indeed I had recognized that the insert-dotted-circle can and should be merged:

  https://github.com/harfbuzz/harfbuzz/issues/1810

I don't think there's much else that can be shared amongst those shapers though. They are similar but not in exact ways worth sharing.

Let me take a look and see how I like your work.

I'm also looking into https://github.com/harfbuzz/harfbuzz/pull/2726 now.

behdad:
I like to commit a less drastic change that doesn't merge syllable types. Also, patch as is has a bug because it defines:
```
#define syllabic_category() complex_var_u8_0()
```
whereas the USE shaper uses a different definition. Let me see if I get this in tonight.

src/hb-ot-shape-complex-use.cc
```diff
@@ -454,71 +442,13 @@ reorder_syllable_use (hb_buffer_t *buffer, unsigned int start, unsigned int end)
...
-      has_broken_syllables = true;
-      break;
-    }
-  if (likely (!has_broken_syllables))
-    return;
-
-  hb_glyph_info_t dottedcircle = {0};
-  if (!font->get_nominal_glyph (0x25CCu, &dottedcircle.codepoint))
-    return;
-  dottedcircle.use_category() = hb_use_get_category (0x25CC);
```
behdad, line 482: You also seem to drop this...

behdad:
Closing in favor of https://github.com/harfbuzz/harfbuzz/pull/2817

PR #2792, opened 2020-12-30
[docs] Hide private enum values

khaledhosny:
`/*< skip >*/` does not seem to hide them.

### DIFF
--- a/src/hb-ot-layout.h
+++ b/src/hb-ot-layout.h
@@ -446,6 +446,7 @@ typedef enum {
   HB_OT_LAYOUT_BASELINE_TAG_IDEO_EMBOX_TOP_OR_RIGHT	= HB_TAG ('i','d','t','p'),
   HB_OT_LAYOUT_BASELINE_TAG_MATH			= HB_TAG ('m','a','t','h'),
 
+  /*< private >*/
   _HB_OT_LAYOUT_BASELINE_TAG_MAX_VALUE = HB_TAG_MAX_SIGNED /*< skip >*/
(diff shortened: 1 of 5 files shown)


behdad:
What about gobject-introspection? Some might have been because of that.

khaledhosny:
> What about gobject-introspection? Some might have been because of that.

Do you mean `/*< skip >*/` was there for gobject-introspection, or `/*< private >*/` would have an effect on it? If the former, I can put them back, as I assumed it was a typo since GTK-Doc does not mention it, but it didn’t occur to me to check gobject-introspection.

khaledhosny:
(I don’t know what is wrong with msys2, but it is terribly unreliable for CI!).

behdad:
> (I don’t know what is wrong with msys2, but it is terribly unreliable for CI!).

Yes....

behdad:
> Do you mean `/*< skip >*/` was there for gobject-introspection, or `/*< private >*/` would have an effect on it?

I meant the former, but am not sure. LGTM if doesn't change g-i output. Thanks.

khaledhosny:
> I meant the former, but am not sure. LGTM if doesn't change g-i output. Thanks.

It does change the `.gir` file, so I put it back.

khaledhosny:
> > (I don’t know what is wrong with msys2, but it is terribly unreliable for CI!).
> 
> Yes....

Anyone remembers why do we have it? Is it for MINGW? We already cross-build with it on Linux and seems to be much more reliable.

PR #2794, opened 2020-12-30
Various doc additions

khaledhosny:
This brings doc coverage (as calculated in `docs/harfbuzz-undocumented.txt`) to 87%, with most of the missing docs being for gobject (which I’m not sure we need to document) and subset (which is probably deliberately undocumented).

We might want to find a solution for this and make missing docs an error going forward, or otherwise issues with new documentation goes unnoticed and things deteriorates over time.


src/hb-blob.h
```diff
@@ -38,10 +38,10 @@ HB_BEGIN_DECLS
 
 /**
  * hb_memory_mode_t:
- * @HB_MEMORY_MODE_DUPLICATE
- * @HB_MEMORY_MODE_READONLY
- * @HB_MEMORY_MODE_WRITABLE
- * @HB_MEMORY_MODE_READONLY_MAY_MAKE_WRITABLE
+ * @HB_MEMORY_MODE_DUPLICATE: HarfBuzz immediately makes a copy of the data.
+ * @HB_MEMORY_MODE_READONLY: HarfBuzz client will never modify the data
```
behdad, line 42: add "and HarfBuzz will never modify the data."

src/hb-blob.h
```diff
@@ -38,10 +38,10 @@ HB_BEGIN_DECLS
...
+ * @HB_MEMORY_MODE_WRITABLE: HarfBuzz client made a copy of the data solely for HarfBuzz.
```
behdad, line 43: add ", so HarfBuzz may modify the data."

PR #2801, opened 2021-01-01
[use] Move most joining scripts to USE

dscorbett:


behdad:
Thanks. Would be nice to do some performance measurement on this as well. Also, maybe group them separately from the Brahmi-derived scripts in the big switch just for clarity? I'm merging and hoping that it doesn't break things in the wild.

PR #2811, opened 2021-01-12
Fix incorrect cairo color content detection

AaronWebster:
hb-view triggers a use-of-uninitialized-value MSAN (https://github.com/google/sanitizers/wiki/MemorySanitizer) error when --background and --foreground are solid colors with no alpha.  The current logic (which I don't fully understand) will set CAIRO_CONTENT_COLOR when the RGB components of the foreground/background are non-equal.  This doesn't take in to account solid colors (e.g. #000000).  This change makes it such that if the foreground/background have no opacity, CAIRO_CONTENT_COLOR is set rather than CAIRO_CONTENT_ALPHA.

### DIFF
--- a/util/helper-cairo.cc
+++ b/util/helper-cairo.cc
@@ -406,9 +406,7 @@ helper_cairo_create_context (double w, double h,
 
   if (content == CAIRO_CONTENT_ALPHA)
   {
-    if (view_opts->annotate ||
-	br != bg || bg != bb ||
-	fr != fg || fg != fb)
+    if (view_opts->annotate || (ba == 255 && fa == 255))
       content = CAIRO_CONTENT_COLOR;
   }
   if (ba != 255)


AaronWebster:
Friendly ping!

behdad:
I don't understand this change.

Show us the MSan error instead. Thanks.

AaronWebster:
(build with MSan)

`./hb-view --text=ABC --font-size=72 --margin=36 --background='#000000' --foreground='#ffffff' --output-file /tmp/ABC.png --font-file /usr/share/fonts/truetype/roboto/unhinted/RobotoTTF/Roboto-Regular.ttf`

Relevant error:

```
==178312==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x55f392739ac8 in png_write_bKGD png/src/pngwutil.c:1460:11
    #1 0x55f39272c7e4 in png_write_info png/src/pngwrite.c:237:7
    #2 0x55f39201d1da in write_png cairo/v1_12_16/src/cairo-png.c:400:5
    #3 0x55f39201db23 in cairo_surface_write_to_png_stream_param cairo/v1_12_16/src/cairo-png.c:611:12
    #4 0x55f39201da38 in cairo_surface_write_to_png_stream cairo/v1_12_16/src/cairo-png.c:575:10
    #5 0x55f391f2faa0 in finalize_png(finalize_closure_t*) harfbuzz/util/helper-cairo.cc:249:12
    #6 0x55f391f2ea96 in helper_cairo_destroy_context(_cairo*) harfbuzz/util/helper-cairo.cc:463:5
    #7 0x55f391f3847a in view_cairo_t::render(font_options_t const*) harfbuzz/util/view-cairo.cc:138:3
    #8 0x55f391f2ccad in view_cairo_t::finish(hb_buffer_t*, font_options_t const*) harfbuzz/util/view-cairo.hh:70:5
    #9 0x55f391f2b0ef in shape_consumer_t<view_cairo_t>::finish(font_options_t const*) harfbuzz/util/shape-consumer.hh:82:12
    #10 0x55f391f299d9 in main_font_text_t<shape_consumer_t<view_cairo_t>, 256, 6>::main(int, char**) harfbuzz/util/main-font-text.hh:83:14
    #11 0x55f391f2930e in main harfbuzz/util/hb-view.cc:39:17

SUMMARY: MemorySanitizer: use-of-uninitialized-value png/src/pngwutil.c:1460:11 in png_write_bKGD
Exiting
```
Note that this error does not occur if you set the background to a value where all the channels are not equal (e.g. --background-'#00000001').

behdad:
As you can see, this is happening deep in the bowels of libpng, which we don't even directly use. The fact that if you change the image from grayscale to colorful makes the error go away doesn't mean that it's the right fix.

I suggest you report to libpng, or at least check the code involved. I tried doing so but without knowing which version of libpng, it is hard to map that line number back to the right place in the code.

PR #2812, opened 2021-01-13
[subset] GSUB Glyph closure

qxliu76:
previous GSUB glyph closure includes more glyphs than needed compared to fontTools.
This change makes glyph closure the same as in fontTools by considering the current glyph set passed by parent lookup. 

Also added a commit to remove redundant cmap12 table, this is to match fonttools behavior: https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/subset/__init__.py#L2345


src/hb-ot-layout-gsubgpos.hh
```diff
@@ -83,7 +92,9 @@ struct hb_closure_context_t :
     if (is_lookup_done (lookup_index))
       return false;
 
-    done_lookups->set (lookup_index, glyphs->get_population ());
+    /* a lookup might be visited twice from different rules/starting glyphs, while c->glyphs is
+     * only updated with output glyphs at the end of recursion*/
+    done_lookups->set (lookup_index, glyphs->get_population () + output->get_population ());
```
garretrieger, line 97: I think there are some cases where this doesn't work. Consider a case where you recurse into a look with a specific set of parent glyphs. No additional glyphs are added to output. Then in the same iteration with a different set of parent glyphs you go back to the same lookup. It would be treated as already being done since output hasn't changed.

Unfortunately I think we're going to have to switch to a more complicated done_lookups value, that actually records the specific set of glyphs  that have been used so far. Similar to what fontTools does: https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/subset/__init__.py#L1176

garretrieger (approved):

khaledhosny:
Please rebase to check the CI failures

src/hb-ot-cmap-table.hh
```diff
@@ -1468,7 +1503,155 @@ struct cmap
...
+    for (auto _ : format12_tables.iter ())
+    {
+      unsigned language =_.second;
+      const EncodingRecord *format4_tbl = nullptr;
+      const EncodingRecord *format12_tbl = hb_addressof (encodingRecord[_.first]);
+      if (format12_tbl->platformID == 0 && format12_tbl->encodingID == 4)
+      {
+        if (!unicode_bmp_tables.has (language))
+        {
+          format12_tables.del (_.first);
```
garretrieger, line 1516: I don't think the table should be removed in this case. In the fontTools PR the table is only removed if there is a matching language cmap4 table.

src/hb-ot-cmap-table.hh
```diff
@@ -1468,7 +1503,155 @@ struct cmap
...
+          continue;
+        }
+        format4_tbl = unicode_bmp_tables.get (language);
+      }
+
+      if (format12_tbl->platformID == 3 && format12_tbl->encodingID == 10)
+      {
+        if (!ms_bmp_tables.has (language))
+        {
+          format12_tables.del (_.first);
```
garretrieger, line 1526: Same here.

src/hb-ot-cmap-table.hh
```diff
@@ -1422,15 +1426,16 @@ struct cmap
     cmap *cmap_prime = c->serializer->start_embed<cmap> ();
     if (unlikely (!c->serializer->check_success (cmap_prime))) return_trace (false);
 
+    unsigned count = encodingRecord.len;
```
garretrieger, line 1429: With these changes this method has become very big and complex. It needs to be split up into helper methods to make it easier to follow what's happening. I'll recommend some places where I think things can be split off in following comments.

src/hb-ot-cmap-table.hh
```diff
@@ -1441,17 +1446,47 @@ struct cmap
 
     const EncodingRecord *unicode_bmp= nullptr, *unicode_ucs4 = nullptr, *ms_bmp = nullptr, *ms_ucs4 = nullptr;
     bool has_format12 = false;
+    hb_hashmap_t<unsigned, const EncodingRecord *, (unsigned)-1, nullptr> unicode_bmp_tables;
+    hb_hashmap_t<unsigned, const EncodingRecord *, (unsigned)-1, nullptr> ms_bmp_tables;
+    hb_map_t format12_tables;
 
-    for (const EncodingRecord& _ : encodingrec_iter)
+    for (auto _ : encodingrec_iter)
```
garretrieger, line 1453: This for loop, and down to line 1493 could become a separate method.

It would take encoding iter and the three maps as output parameters and return a boolean, which returns the result of the checks on lines 1492 and 1493.

Name could be something like, "categorize_sub_tables"

src/hb-ot-cmap-table.hh
```diff
@@ -1468,7 +1503,155 @@ struct cmap
     | hb_filter ([&] (const hb_pair_t<hb_codepoint_t, hb_codepoint_t> _)
 		 { return (_.second != HB_MAP_VALUE_INVALID); })
     ;
-    cmap_prime->serialize (c->serializer, it, encodingrec_iter, this, c->plan);
+
+    for (auto _ : format12_tables.iter ())
```
garretrieger, line 1507: This loop could become a helper method, "prune_redundant_format12_tables"

src/hb-ot-cmap-table.hh
```diff
@@ -1468,7 +1503,155 @@ struct cmap
...
+
+        auto format12_iter =
+        + hb_zip (format12_unicodes_set.iter(), format12_unicodes_set.iter() | hb_map(format12_cp_glyphid_map))
+        | hb_filter (c->plan->_glyphset, hb_second)
+        | hb_filter ([c] (const hb_pair_t<hb_codepoint_t, hb_codepoint_t>& p)
+                     {
+                       return c->plan->unicodes->has (p.first) ||
+                              c->plan->glyphs_requested->has (p.second);
+                     })
+        | hb_map (hb_first)
```
garretrieger, line 1548: Add a helper method which creates this iterator. It could then be reused for the format4 table and again in the serialize method which I believe has the exact same iterator

src/hb-ot-cmap-table.hh
garretrieger, line 1507: Is it safe to iterate a map that's being modified?

src/hb-ot-cmap-table.hh
```diff
@@ -1468,7 +1503,155 @@ struct cmap
...
+        ;
+
+        if (format12_iter.len () != format4_iter.len ())
+        {
+          format12_tables.del (_.first);
+          continue;
+        }
+
+        bool is_equal = true;
```
garretrieger, line 1578: The equality check between the iterators could be a helper method.

src/hb-ot-cmap-table.hh
```diff
@@ -1468,7 +1503,155 @@ struct cmap
...
+        for (auto pair : + hb_zip (format12_iter, format4_iter))
```
garretrieger, line 1626: Should be able to re-use the equality helper here.

src/hb-ot-cmap-table.hh
```diff
@@ -1468,7 +1503,155 @@ struct cmap
...
+    }
+
+    auto encoding_record_iter =
+    + encodingrec_iter
+    | hb_filter ([&] (hb_pair_t<const EncodingRecord&, unsigned> _)
+                 {
+                   unsigned format = (this+_.first.subtable).u.format;
+                   if (format != 12) return true;
+                   if (format12_tables.has (_.second)) return false;
+                   return true;
```
garretrieger, line 1650: slightly simpler to write:

return format != 12 || !format12_tables.has (_.second);

src/hb-ot-cmap-table.hh
```diff
@@ -1468,7 +1503,155 @@ struct cmap
...
+        | hb_filter (c->plan->_glyphset, hb_second)
+        | hb_filter ([c] (const hb_pair_t<hb_codepoint_t, hb_codepoint_t>& p)
+                     {
+                       return c->plan->unicodes->has (p.first) ||
+                              c->plan->glyphs_requested->has (p.second);
+                     })
+        | hb_map (hb_first)
+        ;
+
+        if (format12_iter.len () != format4_iter.len ())
```
garretrieger, line 1572: len()'s are probably reasonably expensive to compute, so I'd recommend caching the value from the .len() call above and re-using here.

src/hb-ot-cmap-table.hh
garretrieger: Looking at the map impl, I don't think it is. It's possible for the map's resize() to be triggered by a del() call since that internally calls set_with_hash(). Specifically if the last addition to the map pushes it past the occupancy limit for a resize, then the next del() call will trigger a resize() and the array which the iterator is iterating through will be deleted.

garretrieger: So you'll instead need to create keep a set/list of the format12_tables to be deleted and add to it in this loop.

src/hb-ot-cmap-table.hh
qxliu76: Actually format12_tables keeps format12 tables that should be removed. I'll rename it so it's not misleading.

src/hb-ot-cmap-table.hh
qxliu76: In set_with_hash (), resize () won't be triggered if the value is kINVALID. 
{
if (unlikely (key == kINVALID)) return;
if ((occupancy + occupancy / 2) >= mask && !resize ()) return;
}
It seems safe to call del () while iterating a map.

qxliu76: oh, sorry, the value is invalid not the key. I will change it.

src/hb-ot-cmap-table.hh
```diff
@@ -1348,6 +1352,181 @@ struct cmap
...
 
+  template<typename Iterable1, typename Iterable2,
+           hb_requires (hb_is_iterator (Iterable1)),
+           hb_requires (hb_is_iterator (Iterable2))>
+  bool compare_iterators (Iterable1 Iter_1, Iterable2 Iter_2) const
+  {
+    if (Iter_1.len () != Iter_2.len ())
+      return false;
+
+    bool is_equal = true;
```
garretrieger, line 1363: this variable can be dropped, just return false immediately when you find a not equal pair.

src/hb-ot-cmap-table.hh
```diff
@@ -1348,6 +1352,181 @@ struct cmap
...
+    if (unlikely (has_format12 && (!unicode_ucs4 && !ms_ucs4))) return false;
+
+    return true;
+  }
+
+  template<typename Iterator,
+           hb_requires (hb_is_iterator (Iterator))>
+  void prune_redundant_format12_tables (hb_subset_plan_t *plan,
+                                        hb_map_t& format12_table_languages,
+                                        Iterator it,
```
garretrieger, line 1433: Suggest renaming this parameter to unicode_gid _pairs (or something that gives an indication what it iterates over)

garretrieger (approved):

behdad:
LGTM. Thanks.

PR #2821, opened 2021-01-18
Meson: Look for FreeType2 via CMake too

fanc999:
Hi,

From the commit message:

<i>The package naming (and versioning) scheme for FreeType2 are different with pkg-config and CMake, so try to look for FreeType2 via CMake if we can't find it via pkg-config, as Visual Studio builds of FreeType2 is likely done with CMake as well as FreeType2 nowadays have more configurable options for its builds.  This means, we try to find 'freetype2' with pkg-config and then 'freetype' with CMake if necessary.</i>

<i>We continue to fallback to manually look for the FreeType2 libraries and headers for Visual Studio builds if we can't find it via pkg-config nor CMake</i>

With blessings, thank you!

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -75,7 +75,11 @@ m_dep = cpp.find_library('m', required: false)
 
 freetype_dep = null_dep
 if not get_option('freetype').disabled()
-  freetype_dep = dependency('freetype2', required: false)
+  freetype_dep = dependency('freetype2', method: 'pkg-config', required: false)
+
+  if not (freetype_dep.found())
+    freetype_dep = dependency('freetype', method: 'cmake', required: false)
+  endif
 
   if (not freetype_dep.found() and
       cpp.get_id() == 'msvc' and


khaledhosny:
I’m not sure why this needs to be done manually, isn’t this what meson does by default (according to https://mesonbuild.com/Dependencies.html#dependency-method)?

khaledhosny:
Sorry, didn’t read the description closely enough.

meson.build
```diff
@@ -75,7 +75,11 @@ m_dep = cpp.find_library('m', required: false)
 
 freetype_dep = null_dep
 if not get_option('freetype').disabled()
-  freetype_dep = dependency('freetype2', required: false)
+  freetype_dep = dependency('freetype2', method: 'pkg-config', required: false)
```
khaledhosny, line 78: Can we keep the first check without specifying `method`, does not seem necessary to override the default here.

khaledhosny: But this whole finding FreeType business is really messed up, too much work for something that should’ve been one line. Since meson seem to have builtin support for finding specific packages with special requirements, and FreeType is such a popular package, it might be good idea to try to teach meson how to handle all of this instead of hacking every package that has a FreeType dependency.

fanc999: Hi @khaledhosny,

Yes, it would make sense, except this becomes even more complicated when we want to look for FreeType that is newer than a specific version (which we don't at this moment), meaning that Meson needs to be updated with a mapping of versions when there is a Freetype release, for instance:

From https://fossies.org/diffs/freetype/2.9.1_vs_2.10.0/docs/VERSIONS.TXT-diff.html, one will see the following:

libtool version (the one that is used in the .pc files): 22.1.16
Actual version (the one that is used in the CMake files): 2.9.1

Let me know what is the best to do here.

khaledhosny: Why would CMake report a different version that pkg-config, that looks like an odd choice from FreeType developers!

behdad:
What's the status of this?

fanc999:
Hi,

Sorry, this one somehow slipped out of my radar.  I think it would be the best to close this one and just use the existing manual library discovery in Meson for Visual Studio builds, to make things simpler.

With blessings, thank you!

PR #2825, opened 2021-01-21
Disable benchmark feature by default

khaledhosny:
It downloads a dependency from the internet, and in the same time not used anywhere (one needs to explicitly call ninja benchmark), not even on the CI (not sure if it that would make much sense since it does not seem to ever fail).


behdad:
LGTM.

PR #2831, opened 2021-01-25
[subet/COLR] Default initialize firstLayerIdx

### LINKED ISSUE #2829: [subset/COLR] Uninitialized `firstLayerIdx`
From a Coverity report:
```
src/hb-ot-color-colr-table.hh: 220 in OT::COLR::subset(hb_subset_context_t *) const::[lambda(unsigned int) (instance 1)]::operator ()(unsigned int) const()
214                                     if (unlikely (!old_record))
215                                       return hb_pair_t<bool, BaseGlyphRecord> (false, Null (BaseGlyphRecord));
216     
217                                     BaseGlyphRecord new_record;
218                                     new_record.glyphId = new_gid;
219                                     new_record.numLayers = old_record->numLayers;
>>>     CID 365706:  Uninitialized variables  (UNINIT)
>>>     Using uninitialized value "new_record". Field "new_record.firstLayerIdx" is uninitialized when calling "hb_pair_t".
220                                     return hb_pair_t<bool, BaseGlyphRecord> (true, new_record);
221                                   })
222         | hb_filter (hb_first)
223         | hb_map_retains_sorting (hb_second)
224         ;
225     
```

Is right. Was this code tested at all?

ckitagawa-work:
Coverity reported that this value was uninitialized. It is set later during serialization (line 181) as the number of layers is needed to compute this value. However, the value should be default initialized (to 0).

This should fix #2829

### DIFF
--- a/src/hb-ot-color-colr-table.hh
+++ b/src/hb-ot-color-colr-table.hh
@@ -214,10 +214,9 @@ struct COLR
 				if (unlikely (!old_record))
 				  return hb_pair_t<bool, BaseGlyphRecord> (false, Null (BaseGlyphRecord));
 
-				BaseGlyphRecord new_record;
+				BaseGlyphRecord new_record = {0};
 				new_record.glyphId = new_gid;
 				new_record.numLayers = old_record->numLayers;
-				new_record.firstLayerIdx = 0; // Updated during serialization.
 				return hb_pair_t<bool, BaseGlyphRecord> (true, new_record);
 			      })
     | hb_filter (hb_first)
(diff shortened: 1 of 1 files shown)


src/hb-ot-color-colr-table.hh
```diff
@@ -217,6 +217,7 @@ struct COLR
 				BaseGlyphRecord new_record;
```
behdad, line 217: Alternative fix is to add `= {0}` initialization here.

ckitagawa-work: Done. Thanks for the suggestion :)

behdad: Yep I suggested `{0}` instead of `{}` because I had a vague memory of the latter being a GNU extension. But you are correct that that's only in C. We're good. Thanks!

PR #2835, opened 2021-01-28
[subset] Remove redundant LangSys

qxliu76:
This change is dependent on  #2699 
fonttools does remove_redundant_langsys() which is missing in harfbuzz. After this removal, fonttools collects active features again in prune_features() and some features are not active because of redundant langsys removal.
This change is trying to get the same result between harfbuzz and fonttools, so collect_features is done again, which might look ugly. Any comment/suggestions would be welcome!


src/hb-ot-layout-gsubgpos.hh
```diff
@@ -3405,6 +3418,42 @@ struct GSUBGPOS
           && !alternate_feature_indices.has (i))
         feature_indices->del (i);
     }
+
+    //find out duplicate features after subset
```
garretrieger, line 3422: Finding duplicate features is a pretty distinct operation from the pruning. So I would make this it's own method, like:
```
 find_duplicate_features ( const hb_map_t *lookup_indices,
                                            const hb_set_t *feature_indices,
                                            hb_map_t *duplicate_index_map /* OUT */)
```

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -3405,6 +3418,42 @@ struct GSUBGPOS
...
+
+      if (f_iter.len () != prev_iter.len ())
+      { prev = i; continue; }
+
+      bool is_equal = true;
+      for (auto _ : + hb_zip (f_iter, prev_iter))
+        if (_.first != _.second) { is_equal = false; break; }
+
+      if (is_equal == true) duplicate_index_map->set (i, prev);
+      else prev = i;
```
garretrieger, line 3455: This is only check for adjacent features that are duplicates. Is that done to match fontTools behaviour? Otherwise I would expect this to check each feature against all other features for duplicates.

src/hb-subset-plan.cc
```diff
@@ -96,7 +125,13 @@ _gsub_closure_glyphs_lookups_features (hb_face_t *face,
                                  nullptr,
                                  nullptr,
                                  &feature_indices);
-  gsub->prune_features (gsub_lookups, &feature_indices);
+
+  hb_map_t duplicate_feature_map;
+  gsub->prune_features (gsub_lookups, &feature_indices, &duplicate_feature_map);
+  _remap_feature_indexes (&feature_indices, &duplicate_feature_map, gsub_features_org);
```
garretrieger, line 131: I think this step (remap_feature_indexes) can be eliminated. If you change the duplicate feature finder method to put all retained feature indices into the duplicate feature map (mapped to themselves if they are not duplicates) then duplicate_feature_map can be used in place of gsub_features_org. ie. the subset plan would retain a reference to duplicate_feature_map instead of gsub_features_org. and collect_features would use that instead.

I found the name 'gsub_features_org' to not be intuitive as to what it's purpose was, using duplicate_feature_map instead throughout makes it more clear.

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -3340,6 +3340,18 @@ struct GSUBGPOS
     hb_set_subtract (lookup_indexes, &inactive_lookups);
   }
 
+  void collect_features (const hb_map_t *org_feature_index_map,
+                         hb_set_t       *new_feature_indexes /* OUT */) const
```
garretrieger, line 3344: I would update this method to also output the set of script and langsys indices which are retained (ie. a map from script index to a set of langsys indices). If that is placed into the subset plan then later one when you're subsetting Script and LanSys you don't need to redo the langsys comparison again to locate which ones to drop. This would also mean it's not necessary to ship around the org_feature_index_map with the subset plan.

If this method is changed to do that, then it would also make sense to rename this and the context, which will help distinguish it from the other collect_features. Suggested name:

prune_langsys

src/hb-ot-layout-gsubgpos.hh
qxliu76: According to the spec, FeatureRecords are listed alphabetically by feature tag. See also: https://github.com/fonttools/fonttools/blob/9c0bc9e554766fe9755f070d1c048c8866922f0e/Lib/fontTools/feaLib/builder.py#L699

So checking for adjacent featureRecords should be enough.

garretrieger: Ok cool, that makes sense then.

garretrieger (approved):
Looks good, I'll get this merged once the PR it depends on is merged.

khaledhosny:
Please rebase on master to double check the CI failures.

qxliu76:
> Please rebase on master to double check the CI failures.

done.

behdad (approved):
I hate how ugly the code has become. But as discussed with Garret, let's merge this, then start on a major refactor / redesign / cleanup. Thank you!

PR #2836, opened 2021-01-29
Use Ragel export expressions to reduce duplication

behdad:
Alternative approach to https://github.com/harfbuzz/harfbuzz/pull/2726


behdad:
The USE is simplified and converted to a point that I'm happy with.

I'm going to do the same for Indic, Khmer, and Myanmar. But first am going to (finally) separate the Khmer/Myanmar data tables from the Indic one.

@dscorbett @jfkthame @mhosken can you please advise if just providing category data for the Unicode Khmer and Myanmar blocks is enough for those two shapers? Or are there other marks (vedic, etc) that can be used with those two scripts that need category data?

behdad:
Also. Maybe best to merge this and continue on the others separately. I don't see any reason to hold this back.

behdad:
I made https://github.com/harfbuzz/harfbuzz/pull/2836 as my proposed alternate approach.

src/hb-ot-shape-complex-use-machine.rl
```diff
@@ -31,58 +31,81 @@
...
-VMAbv	= 37; # VOWEL_MOD_ABOVE
-VMBlw	= 38; # VOWEL_MOD_BELOW
-VMPst	= 39; # VOWEL_MOD_POST
-VMPre	= 23; # VOWEL_MOD_PRE
-SMAbv	= 41; # SYM_MOD_ABOVE
-SMBlw	= 42; # SYM_MOD_BELOW
-FMAbv	= 45; # CONS_FINAL_MOD	UIPC = Top
-FMBlw	= 46; # CONS_FINAL_MOD	UIPC = Bottom
-FMPst	= 47; # CONS_FINAL_MOD	UIPC = Not_Applicable
+# Cateories used in the Universal Shaping Engine spec:
```
dscorbett, line 63: Typo: “Cateories”.

dscorbett (approved):
I approve these USE changes. I agree that this could be merged without changing the other shapers.

dscorbett:
U+A92E KAYAH LI SIGN CWI has Script_Extensions=Mymr. I can’t think of any other characters relevant to Myanmar or Khmer outside the various Myanmar and Khmer blocks, except of course for generic ones like U+00D7 and U+25CC. Neither of the OpenType script development specs mentions any Vedic etc. characters.

PR #2845, opened 2021-02-06
Fix to export target include directories

tksuoran:
- Without `target_include_directories()` harfbazz library will not export include directories. This is the main fix.
- Requiring too old cmake version causes warning from cmake that support for old versions can be removed from cmake, so I upped the version requirement. I don't mind if this gets excluded, but I think this should be fixed as well.
- I also silenced the warning about cmake support being maybe removed. I don't see that happening, at least for a while. Yes, I still think this even though I would happy with meson as well. I don't mind if this gets excluded, but I don't believe cmake support can be removed for a while.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
-cmake_minimum_required(VERSION 2.8.0)
+cmake_minimum_required(VERSION 3.0.0)
 project(harfbuzz)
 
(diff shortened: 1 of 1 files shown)


behdad:
LGTM  as well as much as I understand CMake (none).

khaledhosny:
I’d keep the warning. We really want people to migrate to meson even if we don’t currently have a time line for dropping CMake support. Maintaining support for three build systems is not practical.

PR #2846, opened 2021-02-07
Pointer usage fixes.

ihsinme:
given the use of a smart pointer, the link will become invalid when destroyed.
I find it better to use a full-fledged object rather than a link.

### DIFF
--- a/src/hb-ot-stat-table.hh
+++ b/src/hb-ot-stat-table.hh
@@ -316,7 +316,7 @@ struct STAT
   hb_ot_name_id_t get_axis_record_name_id (unsigned axis_record_index) const
   {
     if (unlikely (axis_record_index >= designAxisCount)) return HB_OT_NAME_ID_INVALID;
-    const StatAxisRecord &axis_record = get_design_axes ()[axis_record_index];
+    const StatAxisRecord axis_record (get_design_axes ()[axis_record_index]);
     return axis_record.get_name_id ();
   }
 


jfkthame:
This looks unnecessary; the lifetime of the const reference is fine. What exactly do you think is destroyed, and when?

ihsinme:
I'm sorry to be late with the reply.
looking at https://godbolt.org/z/d6jjhs it is clear that the release will occur immediately.
Ideally, use the move constructor or option `Func3`.
but I believe that using simple changes makes it possible to draw attention to the problem.

behdad:
Please tell us why you think the code as is has a problem.

The object pointed to is owned by the face and won't disappear even if the immediate rvalue `hb_array_t` disappears. There's no smart-pointer involved.

ihsinme:
sorry i looked and don't see a smart pointer.
this is my mistake in the description.

most likely my analyzer found similar release processes.
this is possible for vector objects.
But I need to check it out.

give those to me a little time, and I will describe the problem again.

again sorry for the confusion with smart pointers.

ihsinme:
good day.

I confirm that the vector behaves the same as a smart pointer in this situation.
in fact, as far as I can tell `axis_record` refers to an object that will be returned from the function by value. unfortunately this object has an automatic release property, and will be deallocated immediately after initialization `axis_record `.
this is undefined behavior.

I agree that your code is quite complex and perhaps my suggestion is not the most beautiful one.
but for me the main thing is to draw your attention to a possible security issue.
therefore, if something is not clear to you or you are in doubt about something, write, I will try to give examples in assembler of my words.

behdad:
We are sure our code is fine.

These are not C++-managed objects we are passing around. We are passing around pointer/references to POD objects that are guaranteed to remain there as long as the relevant code is running. The memory is owned by the face.

PR #2852, opened 2021-02-10
[build] Fix make distcheck

khaledhosny:
A regression from 7b77ce0507e18fb981a9b865f3eaac0c2ae06044.
It was caught by the CI build in the PR, but apparently we are now used to CI failures that they are effectively ignored.

Also, yay for multiple build systems.


behdad:
Sighz.  Thanks.

Issue #1017, opened 2018-05-01
GSUB/GPOS subsetting


behdad:
- [x] Clean up `UnsizedArray` type and use it everywhere.
- [x] Implement population caching in `hb_set_t`.
- [x] Implement `hb_auto_t<>` that calls init/fini on the object inside, for auto lifecycle management.
- [ ] Write down Supplier protocol. A Supplier will have `iter_forward()` and `iter_backward()`. Check for C++ idioms.
- [x] Write `sized_array_t<>` that takes unsized array-like object and a len and enforces bound checking.
- [ ] Write `SubsetView<>` that takes a set and an array-like object (ensure it has .len) and provides iterators for a subsetted view of the array.
- [x] Change/implement `serialize()` methods that take a `typename Supplier` and work with it.
- [ ] Implement `subset()` methods that build `SubsetView<>`s and calls `serialize()`.

behdad:
This is done now. We have a lot of refactoring / cleanup to do. But that's separate.

behdad:
Waiting for @garretrieger to enable layout subsetting by default.

yisibl:
Apart from COLRv1, which is still working, are there any other tables that cannot be subsetted?

behdad:
> Apart from COLRv1, which is still working, are there any other tables that cannot be subsetted?

All the relevant tables in OpenType are there:
https://github.com/harfbuzz/harfbuzz/blob/4cc4cd3b24ded40cfff651ebf63d20d19253c505/src/hb-subset.cc#L243

yisibl:
> Subsetter now fully supports GSUB/GPOS/GDEF tables (including variations); as such, layout tables are retained by subsetter by default.

https://github.com/harfbuzz/harfbuzz/pull/2958#issuecomment-831396202

PR #2857, opened 2021-02-12
[subset] add a GSUB/GPOS table repacker to resolve offset overflows.

garretrieger:
Adds hb-repacker.hh which can be used to repack GPOS and GSUB tables to avoid offset overflows. Updates hb-subset to call into the repacker if overflows are encountered.

The repacker works by extracting a graph of the subtables in GSUB/GPOS from the serialization context's internal object stack. This allows the repacker to operate on a generic graph of where the sub table binary blobs are the vertices and offsets form edges.

The repacker first attempts to generate a better topological sorting by using Kahn's and/or Djikstra's algorithm to sort by distance to a vertex. Then if overflows still exist it attempts different resolutions iteratively until a non-overflowing topological ordering is found. Currently supported resolution strategies:

- Duplicating shared sub-tables which allows the duplicate to be placed closer to the parent.
- Moving individual sub-tables and their children backwards in the ordering.

Currently this is able to resolve most of the overflow cases I've encountered, but more work is still needed.


src/hb-serialize.hh
```diff
@@ -119,24 +119,41 @@ struct hb_serialize_context_t
...
   bool check_equal (T1 &&v1, T2 &&v2)
-  { return check_success ((long long) v1 == (long long) v2); }
+  {
+    if ((long long) v1 != (long long) v2)
+    {
+      err_offset_overflow ();
+      return false;
+    }
+    return true;
+  }
```
behdad, line 156: So this changes `check_equal()` to always result in an offset-overflow error. While we currently don't call this in any other instance, the assumption is not good.

Let's add a flag-enum for serialization error, which would be "ran-out-of-room" error, "other" error, and this new "offset overflow" error. And then all of `check_success()`, `check_equal()`, `check_assign()`, etc. take an optional error type.

src/Makefile.sources
```diff
@@ -268,6 +269,7 @@ HB_SUBSET_sources = \
 	hb-subset-plan.hh \
 	hb-subset.cc \
 	hb-subset.hh \
+        hb-repacker.hh \
```
behdad, line 272: Keep sorted.

src/hb-serialize.hh
```diff
@@ -119,24 +119,41 @@ struct hb_serialize_context_t
...
     this->successful = true;
     this->ran_out_of_room = false;
+    this->offset_overflow = false;
     this->head = this->start;
     this->tail = this->end;
     this->debug_depth = 0;
 
     fini ();
     this->packed.push (nullptr);
+    this->packed_map.init ();
```
behdad, line 141: Okay I see this is a bugfix.

src/hb-priority-queue.hh
```diff
@@ -0,0 +1,149 @@
...
+#include "hb.hh"
+#include "hb-vector.hh"
+
+/*
+ * hb_priority_queue_t
+ *
+ * Priority queue implemented as a binary heap. Supports extract minimum
+ * and insert operations.
+ */
+struct hb_priority_queue_t
```
behdad, line 39: Ideally I'd make this a template like `hb_hashmap_t`, and order the parameters as priority,value; you currently have value,priority.  I'm fine if you don't templatize right now. Just swap the order and I'd be happy to maintain it from there.

src/hb-priority-queue.hh
```diff
@@ -0,0 +1,149 @@
...
+  bool in_error () const { return heap.in_error (); }
+
+  void insert (unsigned value, int64_t priority)
+  {
+    heap.push (item_t (value, priority));
+    bubble_up (heap.length - 1);
+  }
+
+  item_t extract_minimum ()
+  {
```
behdad, line 65: Rename to `pop_minimum()`?

src/hb-priority-queue.hh
```diff
@@ -0,0 +1,149 @@
...
+
+    return result;
+  }
+
+  const item_t& minimum ()
+  {
+    return heap[0];
+  }
+
+  bool is_empty () const { return heap.length == 0; }
```
behdad, line 80: Add after it:
```
  explicit operator bool () const { return !is_empty (); }
```
Also:
```
  unsigned int get_population () const { return ...; }
```
ala `hb_map_t` / `hb_set_t`.

src/hb-priority-queue.hh
```diff
@@ -0,0 +1,149 @@
...
+
+  unsigned left_child (unsigned index)
+  {
+    return 2 * index + 1;
+  }
+
+  unsigned right_child (unsigned index)
+  {
+    return 2 * index + 2;
+  }
```
behdad, line 101: The three above please write as single-line. Also, they are static methods.

behdad (approved):
Thanks. This looks amazing. Small comments left; good to go otherwise.

src/hb-priority-queue.hh
behdad: `static constexpr`. I'm trying to push as much `constexpr` into the code as possible.

src/hb-cff-interp-common.hh
```diff
@@ -263,7 +263,7 @@ struct UnsizedByteStr : UnsizedArrayOf <HBUINT8>
 
     T *ip = c->allocate_size<T> (T::static_size);
     if (unlikely (!ip)) return_trace (false);
-    return_trace (c->check_assign (*ip, value));
+    return_trace (c->check_assign (*ip, value, HB_SERIALIZE_ERR_OTHER));
```
behdad, line 266: Maybe since we are touching everything, we can add a `ERR_INT_OVERFLOW` as well for when numerical fields overflow.

src/hb-open-type.hh
```diff
@@ -209,7 +209,9 @@ struct Offset : Type
   void *serialize (hb_serialize_context_t *c, const void *base)
   {
     void *t = c->start_embed<void> ();
-    c->check_assign (*this, (unsigned) ((char *) t - (char *) base));
+    c->check_assign (*this,
+                     (unsigned) ((char *) t - (char *) base),
+                     HB_SERIALIZE_ERR_OTHER);
```
behdad, line 214: This should be `ERR_OFFSET_OVERFLOW`, right? This is the one case we talked about.

src/hb-open-type.hh
```diff
@@ -621,7 +623,7 @@ struct ArrayOf
   {
     TRACE_SERIALIZE (this);
     if (unlikely (!c->extend_min (*this))) return_trace (false);
-    c->check_assign (len, items_len);
+    c->check_assign (len, items_len, HB_SERIALIZE_ERR_OTHER);
```
behdad, line 626: For this and the next two maybe a `ARRAY_OVERFLOW`.

src/hb-ot-cmap-table.hh
```diff
@@ -276,7 +276,9 @@ struct CmapSubtableFormat4
     HBUINT16 *idRangeOffset = serialize_rangeoffset_glyid (c, format4_iter, endCode, startCode, idDelta, segcount);
     if (unlikely (!c->check_success (idRangeOffset))) return;
 
-    if (unlikely (!c->check_assign(this->length, c->length () - table_initpos))) return;
+    if (unlikely (!c->check_assign(this->length,
+                                   c->length () - table_initpos,
+                                   HB_SERIALIZE_ERR_OTHER))) return;
```
behdad, line 281: One of the overflows. I'm trying to separate systemic field overflows, from "something went wrong" while we are at it. Not worth extra effort, but since we're doing this already, maybe okay.

src/hb-ot-hmtx-table.hh
```diff
@@ -146,7 +146,7 @@ struct hmtxvmtx
 
     _mtx.fini ();
 
-    if (unlikely (c->serializer->ran_out_of_room || c->serializer->in_error ()))
+    if (unlikely (c->serializer->ran_out_of_room () || c->serializer->in_error ()))
```
behdad, line 149: Just `in_error()` should be enough.

src/hb-ot-name-table.hh
```diff
@@ -230,7 +230,8 @@ struct name
     c->copy_all (records, src_string_pool);
     free (records.arrayZ);
 
-    if (unlikely (c->ran_out_of_room)) return_trace (false);
+
+    if (unlikely (c->ran_out_of_room ())) return_trace (false);
```
behdad, line 234: Replace with `in_error ()`.

src/hb-serialize.hh
```diff
@@ -41,6 +41,14 @@
  * Serialize
  */
 
+enum hb_serialize_error_type_t {
+  HB_SERIALIZE_ERR_NONE =            0x00000000u,
+  HB_SERIALIZE_ERR_OTHER =           0x00000001u,
+  HB_SERIALIZE_ERR_OFFSET_OVERFLOW = 0x00000002u,
+  HB_SERIALIZE_ERR_OUT_OF_ROOM =     0x00000004u,
+};
+HB_MARK_AS_FLAG_T (hb_serialize_error_type_t);
```
behdad, line 50: The enum name and the value names should match. Rename enum to `hb_serialize_error_t` and values to `HB_SERIALIZE_ERROR_...`.

Now, since this is internal and won't be exposed to C API, you can use enum classes with shorter value names, if that feels better. Guess not here, since the enum name needs to be repeated all the time anyway.

src/hb-serialize.hh
```diff
@@ -117,7 +127,12 @@ struct hb_serialize_context_t
     object_pool.fini ();
   }
 
-  bool in_error () const { return !this->successful; }
+  bool in_error () const { return errors & HB_SERIALIZE_ERR_OTHER; }
```
behdad, line 130: If the check the old code, the only place where `ran_out_of_room` was being set to true, `successful` was also being set to false. So, `in_error` should simply return `bool (errors)`. *Any* error will make `in_error()` fire. The `ERROR_OTHER` simply means "something else".

src/hb-serialize.hh
```diff
@@ -415,20 +432,20 @@ struct hb_serialize_context_t
...
   { return start_embed (hb_addressof (obj)); }
 
-  /* Following two functions exist to allow setting breakpoint on. */
-  void err_ran_out_of_room () { this->ran_out_of_room = true; }
-  void err_offset_overflow () { this->offset_overflow = true; }
-  void err_other_error () { this->successful = false; }
+  void set_error (hb_serialize_error_type_t err_type)
+  {
+    errors = errors | err_type;
+  }
```
behdad, line 438: Maybe call this `err()` and make it `return bool (errors = errors | err_type)`.

src/hb-serialize.hh
```diff
@@ -415,20 +432,20 @@ struct hb_serialize_context_t
...
   {
-    if (unlikely (!this->successful)) return nullptr;
+    if (unlikely (in_error ())) return nullptr;
 
     if (this->tail - this->head < ptrdiff_t (size))
     {
-      err_ran_out_of_room ();
-      this->successful = false;
+      set_error (HB_SERIALIZE_ERR_OUT_OF_ROOM);
+      set_error (HB_SERIALIZE_ERR_OTHER);
```
behdad, line 448: This code just needs to set out-of-room. No need to set OTHER based on suggestions above.

src/hb-subset.cc
```diff
@@ -86,7 +86,7 @@ _repack (hb_tag_t tag, const hb_serialize_context_t& c)
   hb_serialize_context_t repacked ((void *) buf, buf_size);
   hb_resolve_overflows (c.object_graph (), &repacked);
 
-  if (unlikely (repacked.ran_out_of_room || repacked.in_error () || repacked.offset_overflow))
+  if (unlikely (repacked.ran_out_of_room () || repacked.in_error () || repacked.offset_overflow ()))
```
behdad, line 89: This just becomes `repacked.in_error ()` as well.

src/hb-open-type.hh
garretrieger: Yes this is the case, so offset overflow is not treated as a fatal error and since this overflow happens outside of the add_link/resolve_links mechanism the repacker can't deal with it so we need to treat it as a fatal error by instead using ERR_OTHER. I could maybe instead add a ERROR_OFFSET_OVERFLOW_FATAL?

src/hb-ot-hmtx-table.hh
garretrieger: atm ERR_RAN_OUT_OF_ROOM is not treated as fatal, so in_error() won't return true if ran out of room is set.

src/hb-serialize.hh
garretrieger: in_error() is currently treated as a check for fatal errors. In my first attempt at this change I did indeed just use bool (errors) but that caused subsetting to fail when offset overflows happened.

Instead, I think I'll make this more clear and rename to fatal_error () and then check if any error flag other than offset overflow is set.

src/hb-open-type.hh
behdad: I suggest considering all errors the same. Just that **inside resolve_links** if you start with no error and end up with offset-overflow then you know you have repacking to do.

src/hb-serialize.hh
```diff
@@ -127,12 +129,13 @@ struct hb_serialize_context_t
...
+  bool in_error () const { return bool (errors); }
 
-  bool successful () const { return !in_error (); }
+  bool successful () const { return !bool (errors); }
 
-  bool ran_out_of_room () const { return errors & HB_SERIALIZE_ERR_OUT_OF_ROOM; }
-  bool offset_overflow () const { return errors & HB_SERIALIZE_ERR_OFFSET_OVERFLOW; }
+  bool ran_out_of_room () const { return errors & HB_SERIALIZE_ERROR_OUT_OF_ROOM; }
+  bool offset_overflow () const { return errors & HB_SERIALIZE_ERROR_OFFSET_OVERFLOW; }
+  bool only_offset_overflow () const { return errors == HB_SERIALIZE_ERROR_OFFSET_OVERFLOW; }
```
behdad, line 138: Please add `HB_NODISCARD` so we'll notice if someone calls them thinking that they mark error. Or maybe we should remove / rename completely and deal with errors directly. WDUT?

behdad:
LGTM. Thanks!

src/hb-serialize.hh
garretrieger: I think just the NO_DISCARD should be fine. Will do that for now.

behdad:
THANKS!

yisibl:
Great work, did this PR solve this issue https://github.com/harfbuzz/harfbuzz/issues/1017?

garretrieger:
Yes, we're currently working on changing GSUB/GPOS subsetting to happen by default.

PR #2858, opened 2021-02-12
[subset] fix heap buffer overflow found by fuzzer.

garretrieger:
Access to arrayZ should be checked against it's length.


behdad:
Thanks. The idiomatic way to fix this is to zip the coverage and ligature array and let the zip take care of different sizes. Not sure if there's good reason to not do that here.

behdad:
> Not sure if there's good reason to not do that here.

I guess the need to call `serialize_append` is one here. I'll implement good idiom for that later.

behdad:
> I guess the need to call `serialize_append` is one here

I'll take that back. This can already be written with a dagger and iterated through with the range for loop.

PR #2868, opened 2021-02-16
Do not round added deltas; fixes #2866

### LINKED ISSUE #2866: Odd rounding artifacts with variable font
The glyphs in the attached test font draw a rectangle, the top edge of which is a long string of aligned points. There are two variable axes; each one pulls up the top edge from one side into a trapezoid. If the axes have the same value, the top edge should once again be a straight line. There are several characters with different dimensions for the shape, but I ended up using "C".

With FreeType, this works reasonable well, although it is visible in the trapezoids that the final points are rounded to the unit grid.

However, using HB for outline extraction (via `hb_font_draw_glyph`), this results in weird rounding artifacts:
test
Red: FreeType/Skia, Blue: HarfBuzz

Detail:
image

It's as if the scaled deltas are rounded individually and then added, instead of having the end result rounded after adding:

`result = round(scalar1 * delta1) + round(scalar2 * delta2)`

vs.

`result = round(scalar1 * delta1 + scalar2 * delta2)`

I haven't started searching for the relevant source code yet, I first wanted to post my observation and (possibly wrong) theory of what goes wrong.

The attached archive contains the test font and a drawbot(-skia) test script that produces the above image.

(drawbot-skia is a skia-python-based cross platform 2D drawing lib using uharfbuzz. Install with `$ pip install drawbot-skia`, then run the script as `$ drawbot test_drawing.py test_image.pdf`)

[test_font+script.zip](https://github.com/harfbuzz/harfbuzz/files/5988611/test_font%2Bscript.zip)

(Using the DrawBot macOS app gives the same result, although the red outline then comes from the OS font machinery instead of FreeType.)

justvanrossum:


justvanrossum:
I adjusted one expected test result: the extents of a glyph from a font without HVAR. But it turns out there are more differences.

The lack of rounding in gvar could mean that this value is now truncated instead of rounded, esp. given the value changes from 551 to 550.

Please advice: is it safe to continue adjusting expected test results, or should I look into `hb_font_get_glyph_extents` and see whether rounding should be applied there or elsewhere?

khaledhosny:
I think you now need to round at API boundry, and this seems to be in https://github.com/harfbuzz/harfbuzz/blob/9c8032cb4b436e093c16dd8476284d8fdcc1c93e/src/hb-ot-glyf-table.hh#L1092-L1176

khaledhosny:
`em_scalef*()` call `roundf()` actually, so may be truncation is happening somewhere before this is called (if truncation is happening at all)?

justvanrossum:
Yeah indeed, upon further studying, I don't see any truncation going on, so I'm tempted to think that perhaps my changes to those expected extent values are correct.

justvanrossum:
Although _perhaps_ these min/max values need to be rounded _before_ being used to calculate width and height:

https://github.com/harfbuzz/harfbuzz/blob/9c8032cb4b436e093c16dd8476284d8fdcc1c93e/src/hb-ot-glyf-table.hh#L946-L949

justvanrossum:
If I do that, I only need to adjust a single expected extent value.

justvanrossum:
Or rather do this:
`extents->width = font->em_scalef_x (max_x) - font->em_scalef_x (min_x);`
And the same for height. Let me check.

justvanrossum:
Same result: if I do that (which is more correct) I only need to adjust a single expected extent value (the one on line 162 of test-ot-metrics-tt-var.c)

behdad:
> Or rather do this:
> `extents->width = font->em_scalef_x (max_x) - font->em_scalef_x (min_x);`
> And the same for height. Let me check.

Yes. This is the correct fix. Thanks.

There might be a few places to do this. One in `glyf` and one or two in the CFF tables perhaps?

justvanrossum:
Are those two failing checks something to worry about?

Other than that, this PR is good to go.

behdad:
> Are those two failing checks something to worry about?

One is a build fail with fontconfig (ignore). The other one looks possibly relevant:
```


--- command ---
09:24:19 MALLOC_CHECK_='2' G_DEBUG='gc-friendly' G_TEST_BUILDDIR='/root/project/build/test/api' G_TEST_SRCDIR='/root/project/test/api' G_SLICE='always-malloc' /root/project/build/test/api/test-draw
--- stdout ---
# random seed: R02S65bf4ef5ac52d25baebed2020f856494
1..12
ok 1 /itoa
# Start of hb tests
# Start of draw tests
ok 2 /hb/draw/empty
Bail out! ERROR:../test/api/test-draw.c:228:test_hb_draw_glyf: assertion failed (str == expected3)
--- stderr ---
**
ERROR:../test/api/test-draw.c:228:test_hb_draw_glyf: assertion failed (str == expected3)
-------
```

behdad:
> Other than that, this PR is good to go.

Other than the one CI fail LGTM as well.

khaledhosny:
The FontConfig build failure should go away if you rebase on master.

justvanrossum:
Noob question: how do I enable `HB_EXPERIMENTAL_API` when running `meson build && meson test -Cbuild test-draw`?

When I enable `HB_EXPERIMENTAL_API` (in the wrong way, my Q above stands), I do not see the values of the strings that should compare equal but don't. I'm not seeing more than this:
```
1/1 harfbuzz:api / test-draw FAIL            0.14s   killed by signal 6 SIGABRT
18:58:32 G_DEBUG=gc-friendly G_TEST_BUILDDIR=/Users/just/code/git/harfbuzz/build/test/api G_SLICE=always-malloc G_TEST_SRCDIR=/Users/just/code/git/harfbuzz/test/api MALLOC_CHECK_=2 MALLOC_PERTURB_=90 /Users/just/code/git/harfbuzz/build/test/api/test-draw
----------------------------------- output -----------------------------------
stdout:
# random seed: R02Sbf99f12c8662e8520ef49caf0cd7fd2d
1..12
ok 1 /itoa
# Start of hb tests
# Start of draw tests
ok 2 /hb/draw/empty
Bail out! ERROR:../test/api/test-draw.c:228:test_hb_draw_glyf: assertion failed (str == expected3)
stderr:
**
ERROR:../test/api/test-draw.c:228:test_hb_draw_glyf: assertion failed (str == expected3)
------------------------------------------------------------------------------
```

khaledhosny:
`meson -Dexperimental_api=true build`

justvanrossum:
> meson -Dexperimental_api=true build

Unfortunately that doesn't work for me: it still doesn't run the test-draw tests.

khaledhosny:
It does for me. Are you re-configuring an existing `build` directory, it might be simpler to delete it or use `meson configure`.

justvanrossum:
Ah thanks, deleting the build dir worked!

My next problem is that I don't see the values of the unequal strings of the failing test. All I see in testlog.txt the snippet I posted above. How do I even debug this?

justvanrossum:
> How do I even debug this?

Duhhh printf to the rescue!

But now: what does that codecov/path failure mean and how do I fix it?

justvanrossum:
Could it be this: The test-draw tests are not executed by most checks, so my changes there are now counted as "not covered by tests"?

khaledhosny:
I can’t make much sense of codecov report, but it runs on the linux-ci job and it does not enable experimental API.

justvanrossum:
Thanks @behdad and @khaledhosny!

Issue #2860, opened 2021-02-15
Too frequent unsafe-to-break flags output for Lucida?


drott:
```
$ hb-shape --show-flags /System/Library/Fonts/LucidaGrande.ttc "test multiple words"
[t=0+766|e=1+1141#1|s=2+1044#1|t=3+766#1|space=4+648#1|m=5+1912#1|u=6+1271#1|l=7+592#1|t=8+766#1|i=9+592#1|p=10+1289#1|l=11+592#1|e=12+1141#1|space=13+648#1|w=14+1578#1|o=15+1258#1|r=16+838#1|d=17+1289#1|s=18+1044#1]
```

shows that for the Lucida system font on Mac there's surprisingly few safe-to-break positions, in this case only the first glyph, if I interpret the output correctly? That seems wrong - although I haven't looked at Lucida's AAT tables more closely.

Lucida does have a morx table, [table directory here](https://gist.github.com/drott/4caca4bf4a02823a86a63de3cd3ee63f).

@jfkthame @javifernandez @kojiishi

behdad:
So.. the "unsafe-to-break" handling for AAT `morx` table is implemented this way: as we running a state-machine, every time we are not in state=0 (start-of-text), we deem as unsafe-to-break.

As it happens, the chain 8 in this face, which seems to be implementing Thai Sara-Am handling, spends all its time in state=2 for non-Thai text.

I checked the transitions from state=0 vs state=2 to see if they are deemed identical (and as such maybe we can detect), but they are not. I'm not sure if we can do anything in this case short of a full state-machine analysis implementation.

Ideas? @jfkthame 

@drott is this just affecting performance, or have you seen other adverse effects? How critical is "fixing" this to Chrome?

drott:
I first thought this was essential for a fix for https://bugs.chromium.org/p/chromium/issues/detail?id=1173219 - but it looks like @javifernandez found a different solution. So the priority is now lower.

drott:
Thanks for looking into the state machine behaviour of Lucida, that's what I was hoping to get further insight in. So do I understand correctly, that the font immediately tries to go to state 2 to attempt handling Thai Sara-Am, then stays in state=2 for the Latin sequence all the time? 
Indeed, at the moment, with @javifernandez fix, it looks like this only affects performance when it comes to determining the size of to-be-reshaped portion of text during line layout.

behdad:
[...]

PR #2871, opened 2021-02-18
[aat] Improve unsafe-to-break logic

behdad:
Reduces false-positives.  See comments for new logic.

Fixes https://github.com/harfbuzz/harfbuzz/issues/2860

Need to add test.


jfkthame:
The logic here looks sensible to me, though I'm a little concerned that Dominik indicates in #2860 that he's still not getting a good result.

behdad:
> I'm a little concerned that Dominik indicates in #2860 that he's still not getting a good result.

Looks like Dominik assumed I merged the fix so he was testing master.

PR #2874, opened 2021-02-20
Constexpr2

behdad:
Following up on yesterday's work, this is reviewing and rebasing the rest of my commits from June in the https://github.com/harfbuzz/harfbuzz/commits/constexpr-wip-behdad branch.


behdad:
Can someone who knows cmake fix the failure?  Where the heck does cmake detect pthread?

cc @khaledhosny @ebraminio

khaledhosny:
I’ve no idea and wouldn’t know where to start looking. If it were up to me, I’ll disable CMake testing on CI jobs and let CMake users deal with the breakages, they seem to be so smart unlike me.

ebraminio:
> Where the heck does cmake detect pthread?

Probably never implemented, guess something like `if (NOT WIN32) add_definitions("-DHAVE_PTHREAD") endif ()` should be enough for the most clients and agreeing with Khaled, even vcpkg, more or less like a cross platform cmake package manager, has moved to hb's `meson` port https://github.com/microsoft/vcpkg/pull/12860 recently.

behdad:
I revived all commits from that branch except for the following, which causes test failures and I need to debug and fix separately:
https://github.com/harfbuzz/harfbuzz/commit/fc83ea6279653a500a4ca36fab3956c297eb1132

I intend to merge this when bots pass.

behdad:
I suspended our LGTM analysis setup since they never delivered anything useful to us and it was using cmake, which fails now.

PR #2875, opened 2021-02-23
[constexpr] IntType

behdad:
This breaks tests on my Linux box, but passes on Mac. Sending to
bots to see if helps me figure out what's going on.

The main change is in cast operators of BEInt used to return Type,
now return Wide. I don't understand why that can fail tests (and
with no compiler diagnostic message...).

Nor do I remember why I made this change back in June...  I *think*
I made it to make the types be closer to regular small-int types by
involving just one conversion to signed/unsigned instead of two,
which becomes important in template matching. However, I don't
remember the exact details. Sigh...

### DIFF
--- a/src/hb-aat-layout-common.hh
+++ b/src/hb-aat-layout-common.hh
@@ -576,7 +576,7 @@ struct StateTable
 	  if (unlikely (stop > states))
 	    return_trace (false);
 	  for (const HBUSHORT *p = states; stop < p; p--)
-	    num_entries = hb_max (num_entries, *(p - 1) + 1);
+	    num_entries = hb_max (num_entries, *(p - 1) + 1u);
 	  state_neg = min_state;
 	}
       }
@@ -597,7 +597,7 @@ struct StateTable
 	  if (unlikely (stop < states))
 	    return_trace (false);
 	  for (const HBUSHORT *p = &states[state_pos * num_classes]; p < stop; p++)
-	    num_entries = hb_max (num_entries, *p + 1);
+	    num_entries = hb_max (num_entries, *p + 1u);
 	  state_pos = max_state + 1;
 	}
       }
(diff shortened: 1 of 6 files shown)


behdad:
Continuing from https://github.com/harfbuzz/harfbuzz/pull/2874

behdad:
Okay this fails on my Mac as well if fonttools is available. Surprising that only fails subsetting tests it seems. Anyway. I think I know where to look next.

behdad:
I believe this is how it can fail: imagine a `USHORT` (16bit unsigned). If you add that to an `hb_position_t` (a `int32_t`), previously the `USHORT` was lifted to `uint16_t`, and a `uint16_t` when combined with a `int32_t` would result in `int32_t`.

But if I change the cast operator to `uint32_t`, then it becomes `uint32_t` against `int32_t`, and the `uint32_t` wins in such situation.

So I know what's wrong with this patch. Going to revert that part. But still am curious to know why it fails tests without any compiler warning.

behdad:
Ah, I see what I'm doing wrong. The BEInt cast operators for signed types were relying on the size of the return value to be the same as the size of the read value for the sign to appear. That is:
```
operator uint16_t () const { return (v[0] << 8) + v[1]; }
```
fails miserably if `uint16_t` is replaced by `uint32_t`.

behdad:
Not sure if this narrowing fight is worth fighting...
```
../src/hb-ot-layout-gsubgpos.hh: In instantiation of ‘void OT::ChainRule::serialize_array(hb_serialize_context_t*, OT::HBUINT16, Iterator) const [with Iterator = hb_map_iter_t<hb_array_t<const OT::IntType<short unsigned int> >, const hb_map_t*&, (hb_function_sortedness_t)0, 0>; typename hb_enable_if<hb_is_iterator_of<Lhs, typename Lhs::item_t>::value>::type* <anonymous> = 0; OT::HBUINT16 = OT::IntType<short unsigned int>]’:
../src/hb-ot-layout-gsubgpos.hh:2341:30:   required from here
../src/hb-ot-layout-gsubgpos.hh:2326:15: error: narrowing conversion of ‘(unsigned int)g’ from ‘unsigned int’ to ‘short unsigned int’ inside { } [-Werror=narrowing]
       c->copy (HBUINT16 {g});
       ~~~~~~~~^~~~~~~~~~~~~~
```

Issue #2426, opened 2020-05-27
Add get_serial()/changed() API


behdad:
To allow synching internal and external caches, we like to provide serial/changed scheme.  I did that in Pango years ago.  I think cairo does as well but don't remember.

It wouldn't solve all our problems, but is a good start.  Might be enough to fix:
https://github.com/harfbuzz/harfbuzz/issues/1620
https://github.com/harfbuzz/harfbuzz/issues/2270

We should fully consider if this can be of any use, or we should find a better solution, to fix:
https://github.com/harfbuzz/harfbuzz/issues/2365

In general change-detection is against the initial design philosophy of our face and font objects, which was you configure the object once and never change again.  There's two reasons that people have deviated from that:

  - hb_font_t doesn't hold much.  A few numbers.  So systems like Chrome, Android, etc, makes sense to keep one around and configure it and call hb_shape(), instead of keeping multiple (one per size) around or recreating every time.

  - hb-ft is it's own mess...  People want to change FT_Face and hb_font_t to follow.  That's much harder since we don't want to check things (even just serial, even though FreeType doesn't provide serials) in every single font-funcs callback.  To that end I'm thinking that when we have a single-glyph and multiple-glyphs version of a callback, is easy to make the multiple-glyphs version fast and that's enough for a fast shaping operation.  The single one can either forgo caches or have checks and be slow...

Anyway.  The main idea is that `hb_face_t` and `hb_font_t` will have a `get_serial()` that returns a number.  Whenever a `set_...` thing is called on them, they increase serial if the actual internal value of something changed.

There will also be `data_changed()` call to inform them that some of the set callbacks' underlying data changed so caches if any should be dropped.  This operation increases serial as well.

behdad:
I don't think any other object need this.  For Unicode data we decided early on to not associate private data with the unicode-funcs.  So if someone wants to change underlying data there they need to recreate the funcs.  But there's also no caching there so that's not an issue.

Issue #2270, opened 2020-03-19
[hb-ft] FT_Face should get updated every time hb_font_t coords changes


ebraminio:
Apparently `FT_Set_Var_Blend_Coordinates` is [only called](https://github.com/harfbuzz/harfbuzz/blob/d5338ba/src/hb-ft.cc#L920) when `hb_ft_font_set_funcs` is called on a hb_font_t but possibly it should be called everytime `hb_font_t` coords changes via `hb_font_set_variations` or `hb_font_set_var_*` calls, shouldn't it? If is true, how it should be implemented?

behdad:
This is the kind of property-change notification we do not currently support.  I'm not sure how to do, if at all.  Maybe adding a `changed` callback to font-funcs.

Issue #1620, opened 2019-03-16
hb_font caching sizes after 2.3.1


math4origami:
I recently updated to harfbuzz 2.3.1 and am now experiencing a change in behavior that used to work for us.  

I took this example and modified it to demonstrate my issue.
https://github.com/harfbuzz/harfbuzz-tutorial/blob/master/hello-harfbuzz-freetype.c

```
#include <stdlib.h>
#include <stdio.h>
#include <math.h>
#include "third_party/harfbuzz/src/hb.h"
#include "third_party/harfbuzz/src/hb-ft.h"
#include <string>
#include <vector>

int
main(int argc, char **argv)
{
  const char *fontfile;

  if (argc < 2)
  {
    fprintf (stderr, "usage: hello-harfbuzz font-file.ttf\n");
    exit (1);
  }

  fontfile = argv[1];
  std::vector<std::string> texts = {"i", "aia"};
  std::vector<int> font_sizes = {32, 64};

  /* Initialize FreeType and create FreeType font face. */
  FT_Library ft_library;
  FT_Face ft_face;
  FT_Error ft_error;
  if ((ft_error = FT_Init_FreeType (&ft_library)))
    abort();
  if ((ft_error = FT_New_Face (ft_library, fontfile, 0, &ft_face)))
    abort();

  /* Create hb-ft font. */
  hb_font_t *hb_font;
  hb_font = hb_ft_font_create(ft_face, NULL);

  for (int i = 0; i < texts.size(); ++i) {
    /* Create hb-buffer and populate. */
    hb_buffer_t *hb_buffer;
    hb_buffer = hb_buffer_create ();

    int font_size = font_sizes[i];
    if ((ft_error = FT_Set_Char_Size (ft_face, font_size*64, font_size*64, 0, 0)))
      abort();

    hb_buffer_add_utf8 (hb_buffer, texts[i].c_str(), -1, 0, -1);
    hb_buffer_guess_segment_properties (hb_buffer);

    /* Shape it! */
    hb_shape (hb_font, hb_buffer, NULL, 0);

    /* Get glyph information and positions out of the buffer. */
    unsigned int len = hb_buffer_get_length (hb_buffer);
    hb_glyph_info_t *info = hb_buffer_get_glyph_infos (hb_buffer, NULL);
    hb_glyph_position_t *pos = hb_buffer_get_glyph_positions (hb_buffer, NULL);

    /* Print them out as is. */
    printf ("Raw buffer contents:\n");
[...]

PR #2878, opened 2021-02-24
Check start/peak/end coords in VarRegionAxis::sanitize.

jfkthame:
This saves us checking these conditions at runtime, as noted in the TODO.

### DIFF
--- a/src/hb-ot-layout-common.hh
+++ b/src/hb-ot-layout-common.hh
@@ -2165,12 +2165,6 @@ struct VarRegionAxis
   {
     int start = startCoord, peak = peakCoord, end = endCoord;
 
-    /* TODO Move these to sanitize(). */
-    if (unlikely (start > peak || peak > end))
-      return 1.;
-    if (unlikely (start < 0 && end > 0 && peak != 0))
-      return 1.;
-
     if (peak == 0 || coord == peak)
       return 1.;
 
@@ -2187,9 +2181,9 @@ struct VarRegionAxis
   bool sanitize (hb_sanitize_context_t *c) const
   {
     TRACE_SANITIZE (this);
-    return_trace (c->check_struct (this));
-    /* TODO Handle invalid start/peak/end configs, so we don't
-     * have to do that at runtime. */
+    return_trace (c->check_struct (this) &&
+		  startCoord <= peakCoord && peakCoord <= endCoord &&
+		  (startCoord >= 0 || endCoord <= 0 || peakCoord == 0));
   }
 
   public:


behdad:
Actually, does that even work? I think not.

Currently sanitize of arrays of type `T` do not call into individual `T` members' `sanitize()` if such sanitize doesn't take extra arguments and is trivially-copyable:

https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-open-type.hh#L495
```
  template <typename ...Ts>
  bool sanitize (hb_sanitize_context_t *c, unsigned int count, Ts&&... ds) const
  {
    TRACE_SANITIZE (this);
    if (unlikely (!sanitize_shallow (c, count))) return_trace (false);
    if (!sizeof... (Ts) && hb_is_trivially_copyable (Type)) return_trace (true);
      if (unlikely (!c->dispatch (arrayZ[i], hb_forward<Ts> (ds)...)))
	return_trace (false);
    return_trace (true);
  }
```
The intent was to optimize away if all such sanitize() calls would do is check boundaries.

You've found a good reason to *not* do thing this way anymore.

When I wrote `sanitize()` initially, we weren't into SFINAE, etc. Now we are.

I suggest we remove the trivial `sanitize()` implementations, ie. the ones that just check bounds of a static-size struct, eg.:
```
  {
    TRACE_SANITIZE (this);
  }
```
and never skip calling `sanitize()` if it's actually implemented.

Maybe also making Arrray `sanitize()` also replace an insane member by copying its Null object over it under certain conditions. I'm not sure about that.

jfkthame:
> Thanks. LGTM in general. Two questions:
> 
>     1. What drove you to this? Was it showing up on any profiles?

I'm seeing VarRegionList::evaluate() showing up in the profile when we need to measure/render a very long run of text in the macOS system font, so I was looking around this area to see if there were any easy optimizations to be had, and noticed the TODO here. (I don't suppose this by itself would make much difference, but on principle it seemed like something we should fix.)

(See https://bugzilla.mozilla.org/show_bug.cgi?id=1694420 for the example & profile where this showed up.)

At a higher level, if we could for example cache the result of VariationDevice::get_delta() for a given store and font, that might make a real difference, as we end up calling it numerous times with the same parameters.

>     2. The way you have it in the patch, it will essentially disable an entire `VariationStore` I think. That's because this object (`VarRegionAxis`) is _not_ referenced to by an offset. The closest offset is to `VarRegionList`, which encodes all regions in a `VariationStore`.
> 
> 
> I think the latter is _not_ desired, so I suggest we fix-up the values in sanitize instead of returning false. WDYT? If you want to give that a try, grep for `try_set`. Is trivial.

If this fails, the font is broken per spec anyway, so doesn't really have any right to work. But I'd be fine with fixing-up rather than just disabling if you'd prefer. (Does OTS check/fix this, I wonder? Haven't looked.)

I don't really have cycles to dig deeper here at the moment but if you're up for fixing it properly, that'd be great - thanks!

behdad:
> > Thanks. LGTM in general. Two questions:
> > ```
> > 1. What drove you to this? Was it showing up on any profiles?
> > ```
> 
> I'm seeing VarRegionList::evaluate() showing up in the profile when we need to measure/render a very long run of text in the macOS system font, so I was looking around this area to see if there were any easy optimizations to be had, and noticed the TODO here. (I don't suppose this by itself would make much difference, but on principle it seemed like something we should fix.)
> 
> (See https://bugzilla.mozilla.org/show_bug.cgi?id=1694420 for the example & profile where this showed up.)
> 
> At a higher level, if we could for example cache the result of VariationDevice::get_delta() for a given store and font, that might make a real difference, as we end up calling it numerous times with the same parameters.

Honestly I'm surprised it took so long for anyone to ask for caching VariationStore results! Indeed, I designed that structure in a way that allows various caching mechanisms.

Full caching of interpolation results per entry is one way. But a much more attractive tradeoff is to just cache the scalar value of regions.

Adding a cache in front of the `get_h_advances` in hb-ot-font the way hb-ft does is another, easier way.

The main problem with adding caches is that we currently don't have change-tracking / cache-invalidation logic built in. Here's a few related issues:

https://github.com/harfbuzz/harfbuzz/issues/2270
https://github.com/harfbuzz/harfbuzz/issues/1389
https://github.com/harfbuzz/harfbuzz/issues/2426
https://github.com/harfbuzz/harfbuzz/issues/1620

We currently don't have a single place to invalidate `hb_font_t` caches. Though, maybe that's fine. We can add shape-specific caches onto `hb_font_t` and let the `ensure` mechanism clear them. We already have to do that for hb-coretext, but that has an inherent race:
https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-coretext.cc#L346-L374

So I need to find a better way. Suggestions welcome.
 
> > ```
> > 2. The way you have it in the patch, it will essentially disable an entire `VariationStore` I think. That's because this object (`VarRegionAxis`) is _not_ referenced to by an offset. The closest offset is to `VarRegionList`, which encodes all regions in a `VariationStore`.
> > ```
> > 
> > I think the latter is _not_ desired, so I suggest we fix-up the values in sanitize instead of returning false. WDYT? If you want to give that a try, grep for `try_set`. Is trivial.
> 
> If this fails, the font is broken per spec anyway, so doesn't really have any right to work.

Yeah but I *think* we encoded exactly how those should be interpretted into the spec. I would quote exact part of the spec, except that the spec parts written by a specific someone have become tomes of "Intro to How Things Work 101" prose instead of succinct specifications.... Somewhere in this long page towards the end before the example:

https://docs.microsoft.com/en-us/typography/opentype/spec/otvaroverview

So, anyway. I think caching of scalars is better route forward and wouldn't require touching this current mess.

Cleaning sanitize issues I highlighted is something worth doing though. I'll open another issue for that.


> I don't really have cycles to dig deeper here at the moment but if you're up for fixing it properly, that'd be great - thanks!

Sure. Is the kind of thing I enjoy doing.

Wish we had proper microbenchmarking infrastructure though...

cc @drott

jfkthame:
> Adding a cache in front of the `get_h_advances` in hb-ot-font the way hb-ft does is another, easier way.

Caching advances is all very well, but wouldn't help in this case; it's the font's kerning that is the issue, not the advances. The major part of the time is under PairPosFormat1::apply and PairPosFormat2::apply.

behdad:
> Caching advances is all very well, but wouldn't help in this case; it's the font's kerning that is the issue, not the advances. The major part of the time is under PairPosFormat1::apply and PairPosFormat2::apply.

Great. That makes it easier to decide. :)

That's a great profile viewer firefox has! Thanks.

behdad:
Closing in favor of https://github.com/harfbuzz/harfbuzz/issues/2924

PR #2881, opened 2021-02-27
Fix compilation on Nintendo Switch

Mstrodl:
Switch needs climits included for `LLONG_MIN`, `LLONG_MAX`, and `ULLONG_MAX` to be defined


behdad:
We had an issue for this. Is a bad `limits.h` in the SDK. Anyway. I'm pushing a fix try.

Issue #2883, opened 2021-03-02
Substituting ZWJ


khaledhosny:
I have a font that contextually substitutes ZWJ by a visible glyph, something like (in AFDKO feature syntax):
```fea
sub heh-ar.init zerowidthjoiner' by joinerTerminalar.fina;
```

(the idea is to give a tapering end to these glyphs)

This does not work in HarfBuzz, and it seems like the `joinerTerminal-ar.fina` glyph gets replaced by `space` glyph anyway:
```
$ hb-shape font.ttf -u "0647,200D" -V
trace: start table GSUB	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: start lookup 2	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: end lookup 2	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: start lookup 3	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: end lookup 3	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: start lookup 5	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: end lookup 5	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: start lookup 10	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: end lookup 10	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: start lookup 11	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: end lookup 11	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: start lookup 8	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: end lookup 8	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: start lookup 9	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: end lookup 9	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: start lookup 6	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: end lookup 6	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: start lookup 7	buffer: [heh-ar=0|zerowidthjoiner=0]
trace: end lookup 7	buffer: [heh-ar.init=0|zerowidthjoiner=0]
trace: start lookup 12	buffer: [heh-ar.init=0|zerowidthjoiner=0]
[...]

PR #2886, opened 2021-03-03
[ot] Keep substituted Default_Ignorables

khaledhosny:
Don’t replace Default_Ignorables with zero-width space if they are substituted or multiplied, not just when ligated.

After this change, HarfBuzz output matches that of Uniscribe and CoreText for the new tests.

Fixes https://github.com/harfbuzz/harfbuzz/issues/2883


src/hb-ot-layout.hh
```diff
@@ -314,13 +314,17 @@ _hb_glyph_info_get_unicode_space_fallback_type (const hb_glyph_info_t *info)
...
+static inline bool _hb_glyph_info_multiplied (const hb_glyph_info_t *info);
 
 static inline bool
 _hb_glyph_info_is_default_ignorable (const hb_glyph_info_t *info)
 {
   return (info->unicode_props() & UPROPS_MASK_IGNORABLE) &&
-	 !_hb_glyph_info_ligated (info);
+	 !_hb_glyph_info_substituted (info) &&
+	 !_hb_glyph_info_ligated (info) &&
+	 !_hb_glyph_info_multiplied (info);
```
khaledhosny, line 327: I’m not entirely sure the last two checks are actually needed, in my limited tests whenever `_hb_glyph_info_ligated()` or `_hb_glyph_info_multiplied()` return `true`, `_hb_glyph_info_substituted()` return `true` as well (but the reverse is not true).

behdad: Correct. Let's do that.

behdad: LGTM otherwise.

PR #2893, opened 2021-03-10
[subset] prune redundant format12 cmap table

garretrieger:
This is to match latest behavior in fonttools (https://github.com/fonttools/fonttools/pull/2146)
Testcases are updated with newer fonttools version


behdad:
Thanks.

Most of the complexity of cmap table is unused and just a historical artifact. For example, language must always be zero for 0/3 and 3/1 encodings (as touched in this PR):

  https://docs.microsoft.com/en-us/typography/opentype/spec/cmap#use-of-the-language-field-in-cmap-subtables

I'll open an issue at fonttools to simplify that, and ideally add a (ideally default) mode to fonttools subsetter to synthesize a cmap based on requested characters and ignoring the source's cmap repertoire.

In the modern world, a cmap should be an optional Format 4, optional Format 12, and optional Format 14. That's all. I suggest we switch HB subsetter to always synthesizing cmap from scratch.

Common wisdom is that if Format 12 exist, a Format 4 should exist as well. That's in the Recommendations part of the spec:

  https://docs.microsoft.com/en-us/typography/opentype/spec/recom#cmap-table

but same page makes it clear that a Format 4 is NOT required. (This used to require it; was changed in recent years)

I suggest it's time to throw that away as well. So, if we add a Format 12, then no need for a Format 14.

behdad:
As for which plat/enc to include, the Recommendation says:
```
'cmap' Table
New fonts should use Unicode encoding:

If the font supports only characters in the Unicode Basic Multilingual Plane (U+0000 to U+FFFF): either platform 3, encoding 1; or platform 0, encoding 3. With either encoding, use a format 4 subtable.

If the font supports any characters in a Unicode supplementary plane (U+10000 to U+10FFFF): either platform 3, encoding 10; or platform 0, encoding 4. With either encoding, use a format 12 subtable.

If the font supports Unicode variation sequences: platform 0, encoding 5, with a format 14 subtable. This is in addition to a format 4 or format 12 subtable with encodings as listed above.

If the font is designed as a font of last resort, using single glyphs for entire Unicode character ranges: platform 0, encoding 6, with a format 13 subtable. A last-resort font should not use any other encodings or subtable formats.
```

I suggest going for plat0 (Unicode) instead of plat3 (Windows). But doesn't matter, either works.

behdad:
@garretrieger Please advise. Should we abandon this and just always cook cmap from scratch? Or you want to rebase and land this?

garretrieger:
Yeah, I'd prefer to rewrite cmap handling to generate the tables from scratch. This change isn't critical to land in the mean time as it's only a small size saving.

behdad:
Cool. The cmap subsetting code is one of the messier parts. Great to have it redone. Thanks.

PR #2894, opened 2021-03-11
[subset] Update PairPosFormat2 subsetting to match fontTools

garretrieger:
Matches changes from https://github.com/fonttools/fonttools/pull/2221:
- subset class def 1 against the coverage table.
- Don't re-use class 0 in class def 2.
- Ignore class 0 glyphs for the purposes of determing format 1 vs format 2 encoding for ClassDef.

Add an additional test case which demonstrates these issues.


behdad:
My first pass review:

- Ideally I think ClassDef.subset() should always take an explicit glyphset to susbset to. Same about Coverage. That's how we do it in fonttools. Though, am happy to postpone that to a later time.

- You can still keep using a reference instead of pointer in `const hb_set_t &glyphset = *c->plan->glyphset_gsub ();` When a pointer cannot be null, we prefer using a reference.

- We should add a constructor to `hb_set_t` that takes an iterator and populates the set. Then such a set can be created and intersected easily. The dagger we currently have to do instead looks jarring. Though, again, happy to wait for this to be done laters.

garretrieger:
> * Ideally I think ClassDef.subset() should always take an explicit glyphset to susbset to. Same about Coverage. That's how we do it in fonttools. Though, am happy to postpone that to a later time.

Makes sense, I'll take a look and see if this can be easily done without changing too much.

> * We should add a constructor to `hb_set_t` that takes an iterator and populates the set. Then such a set can be created and intersected easily. The dagger we currently have to do instead looks jarring. Though, again, happy to wait for this to be done laters.

Sounds good as well, should be easy to add.

garretrieger:
Fixed the remaining failures for var store, removed the cmap12 changes from this PR, and did a small refactor to eliminate the need to allocate a new set to use as the additional filter.

PR #2899, opened 2021-03-15
[ci] Don’t install meson from its master branch

khaledhosny:
We are testing Harfbuzz not meson!

### DIFF
--- a/.github/workflows/linux-ci.yml
+++ b/.github/workflows/linux-ci.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


behdad:
Thank you Khaled.

khaledhosny:
It does not error now, but the generated coverage file that gets uploaded to codecov.io is empty, seems like something is broken on the CI machines.

behdad:
Thanks. Just having one coverage would be more than enough. Or if it's too much trouble, none would be fine as well.. I don't think any of us use it much.

khaledhosny:
I _think_ it is because GitHub Actions are migrating to Ubuntu 20.04 (from 18.04) and gcovr is broken on 20.04 for some reason. We request the latest version whatever it is, but I think for CI builds it is better to not have moving targets and to migrate to new version deliberately, so I switched to explicitly requesting 18.04. Lets see if this fixes it.

khaledhosny:
> I _think_ it is because GitHub Actions are migrating to Ubuntu 20.04 (from 18.04) and gcovr is broken on 20.04 for some reason.

Yup, that was it.

khaledhosny:
This should be good to go now (codecov ”failure” is because of the code that changed while generating coverage was broken).

PR #2900, opened 2021-03-15
Mark buffer modification methods that can fail with nodiscard

behdad:
Related to https://github.com/harfbuzz/harfbuzz/issues/2312

Motivated by fuzzer-found issues with failing-malloc, I added nodiscard to buffer manipulation methods that can fail, and audited all callsites.


behdad:
cc @drott @jfkthame

PR #2905, opened 2021-03-17
meson.build: Improve FreeType2 dependency search

fanc999:
Hi,

From the commit message:

<i>The package names for the pkg-config files and CMake config files for FreeType2 are different, so match the package names to the dependency search type, and look for FreeType2 first with pkg-config, then with CMake, and then on Visual Studio-style compilers, the actual .lib file.</i>

This can become useful as CMake has built-in support for looking for FreeType2.

With blessings, thank you!

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -75,7 +75,17 @@ m_dep = cpp.find_library('m', required: false)
 
 freetype_dep = null_dep
 if not get_option('freetype').disabled()
-  freetype_dep = dependency('freetype2', required: false)
+  # The package names for FreeType2 are different for pkg-config and CMake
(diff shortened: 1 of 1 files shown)


khaledhosny:
Isn’t this basically the same as https://github.com/harfbuzz/harfbuzz/pull/2821?

fanc999:
Oh right, I must have forgotten... Sorry!

alerque:
No worries. Might I suggest you look over the two implementations, pick the best bits, then update the old PR. I see there are some outstanding questions there too, so maybe after pushing whatever implementation you think is best to that PR make sure the questions are and we'll give it a review.

PR #2906, opened 2021-03-19
hb-config: Include config-override earlier

xNombre:
Currently config-override.h is included at the end of this file. This caused a problem for me while undefing HB_DISABLE_DEPRECATED, namely HB_IF_NOT_DEPRECATED was defined before actual undef took place and broke the whole build. I believe it would break builds for some other defines, too. Moving config-override.h include right after predefined configs is more sane and fixes all potential problems with includes.

### DIFF
--- a/src/hb-config.hh
+++ b/src/hb-config.hh
@@ -86,6 +86,9 @@
 #define HB_NO_LEGACY
 #endif
 
+#ifdef HAVE_CONFIG_OVERRIDE_H
+#include "config-override.h"
+#endif
 
 /* Closure of options. */
 
@@ -153,10 +156,6 @@
 #ifndef HB_OPTIMIZE_SIZE
 #define HB_OPTIMIZE_SIZE
 #endif
-#endif
-
-#ifdef HAVE_CONFIG_OVERRIDE_H
-#include "config-override.h"
 #endif
 
 


alerque:
I can see how this would fix your particular case (and probably a lot of others) but doesn't it also cut off a bunch of potential configurations? I'm not saying any of them are actually useful, but having the override before many of the flags are set does seem to limit the potential scope. For example what if you wanted to define these two things differently? Perhaps enable one and not the other:

```
#define HB_NO_OT_FONT_CFF
#define HB_NO_SUBSET_CFF
```

Given your new arrangement that would become impossible. Again I'm not sure if it makes any sense to ever split up any of those combinations, but I'm not sure why you would want to take a way that possibility.

At least at cursory glance it looks like `HB_IF_NOT_DEPRECATED` might be not like the other things in that set and maybe *it* should move to the end after the override include leaving all the other things in the order they are now.

xNombre:
Hello,
You're wrong with your example actually. Disabling one of the mentioned flags will break the build as, from what I understand, these depend on each other. That's is a case for other ifdefs, too. With my change all of the additional ifdefs are treated as internal configuration which end-user shouln't worry about, at least in most of the cases.

alerque:
> With my change all of the additional ifdefs are treated as internal configuration which end-user shouln't worry about, at least in most of the cases.

I realize my example was contrived, but this is the question. You see "at least in most of the cases", but this change makes it *impossible* to worry about in *any* case. Perhaps most users shouldn't worry about most of these in most cases, but if there are *any cases at all* which would be restricted by this change then I think the order is wrong. Overrides are typically not for most users' benefit anyway, they are there to support the edge cases. If these are truly all internal and should never be messed with that's fine, but if there are any edge cases that might require this then we should revert the order and only fix the one case that's actually known to be broken.

behdad:
I can see both of your sides. Reading the original report cursorily and checking hb-config.hh my first thought was that indeed the override should come before the "Closure of options" section, and I checked and indeed that's what the change offers.

The current structure of hb-config.hh is:
1. From intentional categories to functional categories (eg. `HB_LEAN` enables `HB_NO_CFF`)
2. From functional categories to implementation categories (eg. `HB_NO_CFF` enables `HB_NO_SUBSET_CFF`)

Now, the original reporter is correct that with the override file at the end, one has to duplicate the logic from second step to undo a functional category. Eg. to re-enable CFF, they have to copy the facts of this block:
```
#ifdef HB_NO_CFF
#define HB_NO_OT_FONT_CFF
#define HB_NO_SUBSET_CFF
#endif
```
That duplication (which also needs constant updating) is undesirable. So in that sense the proposed change is good, because it makes the override file encode functional overrides, not implementation overrides.

The other side of the argument, that Caleb brought up, is that the override is there as a last resort, to give full control to the client. And it would completely fail at that if it comes any earlier, because then the implementation categories cannot be overridden at all.

So my first reaction to Caleb's point was that he is correct: that 1. the override is a tool of last resort, and is okay if some duplication has to happen, and 2. not being able to override certain features is an unacceptable situation for an override mechanism.

However, I then convinced myself that the proposed patch is probably an improvement, because 1. it provides a functional, instead of implementational, API to overriding. And 2. it does NOT limit what can be overridden. The second part of the file only has rules of same "if feature XYZ enabled, then enable features ABC, DEF, ...". So if an earlier override wants to override that rule, they can just undefine XYZ and define whichever of ABC, DEF, ... they want. Now, *that* does involve some duplication indeed. So it becomes like the current situation. But if we agree that this extra control over implementation features is less common as wanting to override the functional features, then the proposed change is an improvement.

Given that I mostly added this mechanism such that users of `HB_TINY` can undo some of its functional decisions, I think I like the proposed change as it makes it easier to achieve that, without duplicating what's our implementation details.

It is a slight break; but then again, I don't know of anyone so far who uses the override mechanism other than maybe harfbuzzjs, which we can fix.

alerque:
Fair enough! Thanks for the contribution @xNombre (and explanation Behdad!).

PR #2908, opened 2021-03-22
[build] Don’t look for FontConfig

khaledhosny:
We don’t currently use it anywhere. It was used briefly for the `hb-fc-list` tool, but this tool have not been built since 2015:

```
commit cd042fc8c4a3984c3647cd22a27c34f00636f6e8
Author: Behdad Esfahbod <behdad@behdad.org>
Date:   Thu Jun 18 10:55:13 2015 -0700

    [util] Disable hb-fc-list for now

    Until I figure out what to do about the API, and finalize
    the tool.
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,7 +54,7 @@ jobs:
     executor: autotools-executor
     steps:
       - checkout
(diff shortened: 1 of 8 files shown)


alerque (approved):
Fair enough. This is easy enough to revert if/when needed and there isn't much sense in dragging it around unused.

PR #2912, opened 2021-03-25
[subset] Add more variable font subsetting tests

garretrieger:
Fix a couple issues that popped up:
- Don't keep FeatureVariationRecord's which have no substitutions.
- Don't drop Features if they are referenced in a retained FeatureVariation substitution even if they are otherwise empty.


behdad:
> * Don't keep FeatureVariationRecord's which have no substitutions.

That's definitely not what your code does. "conditions" vs "substitutions".

Which can be dropped is non-trivial. See: https://github.com/harfbuzz/harfbuzz/issues/2334#issue-602270174

Since we are not instancing, the Conditions don't really change...

garretrieger:
Good catch, I'll rework this.

src/hb-ot-layout-common.hh
```diff
@@ -3188,9 +3211,19 @@ struct FeatureVariations
...
+        keep_up_to = i;
+        break;
+      }
+    }
+
+    for (int i = 0; (unsigned) i < varRecords.len; i++) {
+      if (i <= keep_up_to) {
+        subset_record_array (l, &(out->varRecords), this) (varRecords[i]);
+      }
+    }
```
behdad, line 3226: Why not:
```
  unsigned count = (unsigned) (keep_up_to + 1);
  for (unsigned i = 0; i < count; i++)
    ...
```

src/hb-ot-layout-common.hh
```diff
@@ -1117,8 +1117,9 @@ struct Feature
     ;
 
     out->lookupIndex.serialize (c->serializer, l, it);
-    return_trace (bool (it) || subset_featureParams
-		  || (tag && *tag == HB_TAG ('p', 'r', 'e', 'f')));
```
behdad, line 1121: What happens to this `pref` case? There was a reason we added that. Breaks shaping with certain scripts (Khmer?) I think. Probably fonttools doesn't have it?

garretrieger: I'll look into it. If we want this then we'll need to check for it during feature collection since if we drop a feature here that the feature collection was planning to keep the remapped feature indices get out of sync with the actual list of subsetted features.

behdad: > If we want this then we'll need to check for it during feature collection

Right.

garretrieger: Oh misread that, this always keeps "pref" which fontTools does too: https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/subset/__init__.py#L1280. I'll check the feature collection code to make sure it's doing the same.

src/hb-ot-layout-common.hh
garretrieger: Done.

src/hb-ot-layout-common.hh
garretrieger: I'll fix this in a separate PR, this change won't make the current behaviour any worse since this method now always returns true.

behdad:
Sure. I trust that you would. :) LGTM.

PR #2913, opened 2021-03-26
[subset] add Noto Nastaliq subsetting test.

garretrieger:
Fix exposed issues with glyph closure: when computing closure for context lookups if a sequence index has already had a lookup recursed then don't use a current glyph set (matching the chaos behaviour from fontTools).


behdad:
So, this just matches fonttools chaos behavior? LGTM in that case.

@garretrieger Do we know why bots are unhappy?

garretrieger:
Looks to be transient failures, I'll try re-running them.

khaledhosny:
One of the subset tests is timing out, unless there is a way to speed it up, you might want to increase the meson timeout for the subset tests.

garretrieger:
> One of the subset tests is timing out, unless there is a way to speed it up, you might want to increase the meson timeout for the subset tests.

I'll do some profiling and see if there's any low hanging fruit for the slow case.

behdad:
So, a legit subset operation (not a fuzzed one) is timing out? I guess under sanitizers? How long does it take? I thought we're in the fast-enough-that-can't-be-measured ballpark still.

garretrieger:
> So, a legit subset operation (not a fuzzed one) is timing out? I guess under sanitizers? How long does it take? I thought we're in the fast-enough-that-can't-be-measured ballpark still.

It's the fuzzer test that's timing out. With this PR it's taking almost twice as long to complete. I assume that's because the closure can end up doing more work with this change. I'll do some profiling to see if there's anything unreasonable happening.

PR #2916, opened 2021-03-29
[subset] Fix fuzzer found null dereference case.

garretrieger:


behdad:
Would `hb_set_get_empty()` do?

garretrieger:
> Would `hb_set_get_empty()` do?

That looks like it returns the null pool set. In this case the set will be added to later so can't use that.

behdad:
Got it. Thanks. LGTM.

PR #2917, opened 2021-03-29
[subset] fix memory leak when map insert fails.

garretrieger:


behdad:
Wish we could mark this and previous similar change to be reverted when we add proper lifetime-tracking to our hashmap.

LGTM. Thanks.

behdad:
Then we can / should insert a reference-wrapper into the hashmap instead of raw pointers.

behdad:
Also, we can make `set()` return bool. Sounds nicer.

garretrieger:
sounds good.

PR #2918, opened 2021-03-30
[subset] avoid calling clear on null pool set.

garretrieger:
Fixes: https://oss-fuzz.com/testcase-detail/5617065093365760


behdad:
This is becoming magic :(.

Doesn't it return nullptr on failed insert? Where does the null set come in?

behdad:
Also, please include the oss-fuzz URL in the commit logs. Thanks. Bonus for adding the fuzzed tests.

behdad:
> > This is becoming magic :(.
> > Doesn't it return nullptr on failed insert? Where does the null set come in?
> 
> Agreed, this is a case where a unique_ptr like construct would make this much simpler. The null set can be created as a result of memory allocation failing for the hb_set_create () call.

Right. As discussed recently, `hb_set_create()` and the null stuff is considered C API. Ideally you would just insert `set_t` directly in there and let move semantics work for you. We're not that far from that. I expect a few more weeks and we'll be there.

garretrieger:
Ah yes, using move() would be even better. Shall I merge this for now and we'll come back an clean up once we have that in place?

behdad:
Actually I think all you need to do is to use the C API `hb_set_clear()`, which does the null-object check for you.

garretrieger:
> Actually I think all you need to do is to use the C API `hb_set_clear()`, which does the null-object check for you.

Done.

PR #2921, opened 2021-03-30
COLRv1 subset step 1: struct definitions and closure

qxliu76:
hb_ot_colr_closure_forV1() is added just for testing purpose, can be removed later.

### DIFF
--- a/src/hb-ot-color-colr-table.hh
+++ b/src/hb-ot-color-colr-table.hh
@@ -29,6 +29,7 @@
 #define HB_OT_COLOR_COLR_TABLE_HH
 
 #include "hb-open-type.hh"
+#include "hb-ot-layout-common.hh"
 
 /*
  * COLR -- Color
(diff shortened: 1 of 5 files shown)


behdad:
Let's get the data types in before you jump to closure / subsetting.

I just added `Offset24` and am adding `Offset16To` / `Offset24To` / `Offset32To`. Just use those. You are using `NNOffset`, which is wrong.

I also added `VarIdx`.

Last but not least, for the Paint types that have a Var and no-Var version, we want to templatize those. Make them take a `template Variation` argument. Then add a `template NoVariation` ala `template Variation`.

Just do these, with basic `sanitize` and send back to me. Let's get those in, then we'll move to closure/subsetting. I'll read your design doc, but my first thought is that since the offsets are positive, there can't be paint loops. So not sure why you are trying to detect those.

qxliu76:
> Let's get the data types in before you jump to closure / subsetting.
> 
> I just added `Offset24` and am adding `Offset16To` / `Offset24To` / `Offset32To`. Just use those. You are using `NNOffset`, which is wrong.
> 
> I also added `VarIdx`.
> 
> Last but not least, for the Paint types that have a Var and no-Var version, we want to templatize those. Make them take a `template Variation` argument. Then add a `template NoVariation` ala `template Variation`.
> 
> Just do these, with basic `sanitize` and send back to me. Let's get those in, then we'll move to closure/subsetting. I'll read your design doc, but my first thought is that since the offsets are positive, there can't be paint loops. So not sure why you are trying to detect those.

Thank you! I'll address your comments and create a new PR for review. 
The reason I wanted to check for loops is 2 subtable formats below:
PaintColrLayers table can reference any contiguous slice of layers in the LayerV1List table.
PaintColrGlyph references a base glyph ID, which is used to access a corresponding BaseGlyphV1Record.

PR #2923, opened 2021-03-31
[sanitize] change max ops to track number of bytes processed.

garretrieger:
Counting bytes as the operations is likely to be a better proxy for how
much work processing the table will cost vs. the current approach of
counting the number of sub-objects.

This should allow checks for max features, max scripts, etc. to be removed.

I tested this change against the full collection of fonts at https://github.com/google/fonts
and a max ops factor of 3 was sufficient to successfully sanitize all of them.


behdad:
Thanks!

We should consider reducing timeouts now that we don't expect ops to take long.

PR #2926, opened 2021-04-01
[subset] never drop the 'pref' feature.

garretrieger:
Never ever drop feature 'pref', even if it's empty. Harfbuzz uses it to choose the shaper for Khmer.


behdad:
Lgtm!

PR #2929, opened 2021-04-02
[subset] COLRv1 struct definitions

qxliu76:
updated struct definitions

### DIFF
--- a/src/hb-ot-color-colr-table.hh
+++ b/src/hb-ot-color-colr-table.hh
@@ -29,6 +29,7 @@
 #define HB_OT_COLOR_COLR_TABLE_HH
 
 #include "hb-open-type.hh"
+#include "hb-ot-layout-common.hh"
 
 /*
  * COLR -- Color
@@ -38,7 +39,6 @@
 
 
 namespace OT {
-
 
 struct LayerRecord
 {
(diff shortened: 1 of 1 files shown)


src/hb-ot-color-colr-table.hh
```diff
@@ -90,6 +90,474 @@ struct BaseGlyphRecord
...
+template <template<typename> typename V>
+struct ColorStop
+{
+  bool sanitize (hb_sanitize_context_t *c) const
+  {
+    TRACE_SANITIZE (this);
+    return_trace (c->check_struct (this));
+  }
+
+  F2DOT14		stopOffset;
```
garretrieger, line 152: Should be `V<F2DOT14>` and update the static size to match.

src/hb-ot-color-colr-table.hh
```diff
@@ -90,6 +90,474 @@ struct BaseGlyphRecord
...
+  T      value;
+  public:
+  DEFINE_SIZE_STATIC (T::static_size);
+};
+
+// Color structures
+
+// The ColorIndex alpha is multiplied into the alpha of the CPAL entry
+// (converted to float -- divide by 255) looked up using paletteIndex to
+// produce a final alpha.
```
behdad, line 127: Remove this comment. Doesn't belong here.

src/hb-ot-color-colr-table.hh
```diff
@@ -90,6 +90,474 @@ struct BaseGlyphRecord
...
+struct ColorIndex
+{
+  bool sanitize (hb_sanitize_context_t *c) const
+  {
+    TRACE_SANITIZE (this);
+    return_trace (c->check_struct (this));
+  }
+
+  HBUINT16		paletteIndex;
+  V<F2DOT14>	alpha; /* Default 1.0. Values outside [0.,1.] reserved. */
```
behdad, line 138: Not sure what "Default" means in this context, since it cannot be omitted. Remove maybe.

src/hb-ot-color-colr-table.hh
```diff
@@ -90,6 +90,474 @@ struct BaseGlyphRecord
...
+{
+  bool sanitize (hb_sanitize_context_t *c) const
+  {
+    TRACE_SANITIZE (this);
+    return_trace (c->check_struct (this));
+  }
+
+  protected:
+  T      value;
+  VarIdx varIdx; /* Use 0xFFFFFFFF to indicate no variation. */
```
behdad, line 104: The comment is unnecessary. Belongs to where VarIdx is declared.  But then again, it is there already:
```
struct VarIdx : HBUINT32 {
  static constexpr unsigned NO_VARIATION = 0xFFFFFFFFu;
  VarIdx& operator = (uint32_t i) { HBUINT32::operator= (i); return *this; }
};
```

src/hb-ot-color-colr-table.hh
```diff
@@ -90,6 +90,474 @@ struct BaseGlyphRecord
...
+  public:
+  DEFINE_SIZE_STATIC (T::static_size);
+};
+
+// Color structures
+
+// The ColorIndex alpha is multiplied into the alpha of the CPAL entry
+// (converted to float -- divide by 255) looked up using paletteIndex to
+// produce a final alpha.
+template <template<typename> typename V>
```
behdad, line 128: Any reason why you have the template argument called `V` in some and `Var` in other places? I like the `Var` more. Please unify.

src/hb-ot-color-colr-table.hh
```diff
@@ -90,6 +90,474 @@ struct BaseGlyphRecord
...
+    COMPOSITE_HSL_COLOR      = 26,  // https://www.w3.org/TR/compositing-1/#blendingcolor
+    COMPOSITE_HSL_LUMINOSITY = 27,  // https://www.w3.org/TR/compositing-1/#blendingluminosity
+  };
+  public:
+  DEFINE_SIZE_STATIC (1);
+};
+
+// Affine 2D transformations
+
+// This is a standard 2x3 matrix for 2D affine transformation.
```
behdad, line 236: These two comments don't seem to add any information.

src/hb-ot-color-colr-table.hh
```diff
@@ -90,6 +90,474 @@ struct BaseGlyphRecord
...
+  }
+
+  HBUINT8			format; /* format = 4(noVar) or 5 (Var) */
+  Offset24To<ColorLine<Var>>	colorLine; /* Offset (from beginning of PaintLinearGradient
+                                              table) to ColorLine subtable.*/
+  Var<FWORD>			x0;
+  Var<FWORD>			y0;
+  Var<FWORD>			x1;
+  Var<FWORD>			y1;
+  Var<FWORD>			x2; /* Normal; Equal to (x1,y1) in simple cases. */
```
behdad, line 302: This comment is also confusing; remove please.

src/hb-ot-color-colr-table.hh
```diff
@@ -90,6 +90,474 @@ struct BaseGlyphRecord
...
+{
+  bool sanitize (hb_sanitize_context_t *c) const
+  {
+    TRACE_SANITIZE (this);
+    return_trace (c->check_struct (this) && colorLine.sanitize (c, this));
+  }
+
+  HBUINT8			format; /* format = 4(noVar) or 5 (Var) */
+  Offset24To<ColorLine<Var>>	colorLine; /* Offset (from beginning of PaintLinearGradient
+                                              table) to ColorLine subtable.*/
```
behdad, line 297: For multiline comments, please add a `*` to the beginning of the subsequent lines, like is done in existing code.

behdad: And a space before `*/`.

src/hb-ot-color-colr-table.hh
```diff
@@ -90,6 +90,474 @@ struct BaseGlyphRecord
...
+{
+  bool sanitize (hb_sanitize_context_t *c) const
+  {
+    TRACE_SANITIZE (this);
+    return_trace (c->check_struct (this) && paint.sanitize (c, this));
+  }
+
+  HBUINT8                        format; /* format = 10 */
+  Offset24To<Paint>    paint;  /* Offset (from beginning of PaintGlyph table) to Paint subtable. */
+  HBUINT16                       gid;    /* not a COLR-only gid */
```
behdad, line 362: Remove the comments for `gid`. They are irrelevant.

src/hb-ot-color-colr-table.hh
```diff
@@ -90,6 +90,474 @@ struct BaseGlyphRecord
...
+struct PaintColrGlyph
+{
+  bool sanitize (hb_sanitize_context_t *c) const
+  {
+    TRACE_SANITIZE (this);
+    return_trace (c->check_struct (this));
+  }
+
+  HBUINT8                  format; /* format = 11 */
+  HBUINT16                 gid;    /* shall be a COLR gid */
```
behdad, line 377: Remove the comment for `gid`.

src/hb-ot-color-colr-table.hh
```diff
@@ -150,9 +618,18 @@ struct COLR
   bool sanitize (hb_sanitize_context_t *c) const
   {
     TRACE_SANITIZE (this);
-    return_trace (likely (c->check_struct (this) &&
-			  (this+baseGlyphsZ).sanitize (c, numBaseGlyphs) &&
-			  (this+layersZ).sanitize (c, numLayers)));
+    if (unlikely (!(c->check_struct (this) &&
```
behdad, line 621: The `check_struct` here is not very nice. We should also check for version 0.
Copy the sanitize style from `GDEF` table.

src/hb-ot-color-colr-table.hh
```diff
@@ -198,6 +675,17 @@ struct COLR
     return record;
   }
 
+  const BaseGlyphV1Record* get_base_glyphV1_record (hb_codepoint_t gid) const
+  {
+    if ((unsigned int) gid == 0) // Ignore notdef.
+      return nullptr;
+
+    const BaseGlyphV1Record* record = &(this+baseGlyphsV1List).bsearch ((unsigned) gid);
+    if ((record && (hb_codepoint_t) record->glyphId != gid))
```
behdad, line 684: Many things wrong about this function. The gid==0 check, the extra check to return nullptr. We don't do those things. Remove this function and add back when we need to use it.

src/hb-ot-color-colr-table.hh
```diff
@@ -260,16 +748,20 @@ struct COLR
     return_trace (colr_prime->serialize (c->serializer, version, base_it, layer_it));
   }
 
-  protected:
+  public:
```
behdad, line 751: Why?

behdad (changes requested):

src/hb-ot-color-colr-table.hh
```diff
@@ -150,9 +611,11 @@ struct COLR
   bool sanitize (hb_sanitize_context_t *c) const
   {
     TRACE_SANITIZE (this);
```
garretrieger, line 613: I believe this still needs a check_struct() to make sure it's safe to access the offset fields. Will need to check beyond the min size if it's a v1 table.

behdad: Either the check_struct, or individually checking each item. The latter is what we do in GDEF / GSUB / GPOS. Should also check version == 0 for the first few fields, then version == 1 for the newer ones.

garretrieger (approved):

src/hb-ot-color-colr-table.hh
```diff
@@ -150,9 +611,12 @@ struct COLR
...
-    return_trace (likely (c->check_struct (this) &&
-			  (this+baseGlyphsZ).sanitize (c, numBaseGlyphs) &&
-			  (this+layersZ).sanitize (c, numLayers)));
+    return_trace (c->check_struct (this) &&
```
behdad, line 614: This still doesn't check version for 0. I don't know what I'm not successfully communicating.

qxliu76: I added explicit check for version 0. I think I misunderstood you. I thought "check for version 0" is to check the first few fields. :)

behdad: Thanks.

behdad (approved):

PR #2931, opened 2021-04-02
[subset] copy Lookup::markFilteringSet when subseting.

garretrieger:
Also enable layout subsetting for the basic tests.


behdad:
Thanks.

PR #2934, opened 2021-04-05
[subset] don't visit lookup if covered glyph set has failed.

garretrieger:
If covered glyph set is in error then the same lookup can be recursed into repeatedly potentially causing a fuzzer timeout. Fixes: https://oss-fuzz.com/testcase-detail/5416421032067072.


src/hb-ot-layout-gsubgpos.hh
```diff
@@ -120,6 +120,8 @@ struct hb_closure_context_t :
     }
 
     hb_set_t *covered_glyph_set = done_lookups_glyph_set->get (lookup_index);
+    if (covered_glyph_set->in_error ())
```
behdad, line 123: Add `unlikely`.

behdad (approved):
LGTM. Please adjust and merge.

PR #2945, opened 2021-04-10
Fix typo in docs.

AmbientLighter:


alerque (approved):

alerque (approved):
Thanks for taking the time to contribute!

perf/texts/en-words.txt
```diff
@@ -2841,7 +2841,6 @@ Deseret
 DESERET
 deserialize
 DESERIALIZE
-desgined
```
khaledhosny, line 2844: FWIW, this file is used for testing performance and it does not have to be correctly spelled English.

alerque: Yes, but it does seem to at include words from the source and it didn't seem to make sense to keep a matching miss-spelling in it ;-)

PR #2951, opened 2021-04-15
[ci] Try to stop distcheck job from crashing gcc

khaledhosny:


khaledhosny:
I reran this build 3 times and it passed in the three, so I’m hopeful that is might actually stop the random CI failures.

behdad:
Thanks. Maybe -j2 at least? Even on one core that's a speedup.

khaledhosny:
It does not seem like the overall run is much slower than before this change. It seems to be usually in the range of 8 to 10 minutes before this change, and the three builds of this change were around 10 minutes. It is our most slower CI job, but the majority of time is spent on `make distcheck` which was not parallelized before either. May be I should try `-j2` for both and see if they do not crash?

behdad:
Why do we even to two builds in one run, which makes it so slow? Let's split them.

khaledhosny:
> Why do we even to two builds in one run, which makes it so slow? Let's split them.

I was wondering about this as well, but meson build is done in the autotools generated dist, but this does not require running make distcheck first, right?

behdad:
> but meson build is done in the autotools generated dist

Is it? Doesn't look that way to me.

khaledhosny:
With -j2 for make and ninja it now takes 6 minutes, still slow, but the 3 aat jobs that run sequentially are still slower, so I’m tempted to leave it at this for now.

behdad:
> With -j2 for make and ninja it now takes 6 minutes, still slow, but the 3 aat jobs that run sequentially are still slower, so I’m tempted to leave it at this for now.

Sure. Thanks.

behdad:
It also does another `make dist` at the end for good measure.

behdad:
We can also remove one of the mac jobs AFAIC, the middle one.

khaledhosny:
> Is it? Doesn't look that way to me.

I’m seing:

```yaml
      - run: make -j2 distcheck
      - run: rm harfbuzz-* && make distdir
      - run: cd harfbuzz-* && meson build && ninja -j2 -Cbuild test
```

so I’m guessing `harfbuzz-*` is what `make distdir` generates (otherwise why run `make distdir` right after `make distcheck`).

khaledhosny:
OK, I’ll try to see if we can clean this up, and also look at using ccache which might also speed the build a bit.

behdad:
> so I’m guessing `harfbuzz-*` is what `make distdir` generates (otherwise why run `make distdir` right after `make distcheck`).

Ah you are right.  Sorry.

Thanks for taking a look. Just not crashing is plenty for now. :)

PR #2955, opened 2021-04-17
add subset() method for COLRv1 paint tables, , BaseGlyphV1List and LayerV1List

qxliu76:
also support Offset24 in serializer and repacker. 
This depends on PR #2937

### DIFF
--- a/src/hb-repacker.hh
+++ b/src/hb-repacker.hh
@@ -531,7 +531,7 @@ struct graph_t
 
         const auto& child = vertices_[link.objidx].obj;
         int64_t child_weight = child.tail - child.head +
-                               (!link.is_wide ? (1 << 16) : ((int64_t) 1 << 32));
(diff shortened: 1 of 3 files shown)


src/hb-repacker.hh
```diff
@@ -627,22 +629,26 @@ struct graph_t
...
+    {
       if (link.is_signed)
       {
         serialize_link_of_type<OT::HBINT16> (link, head, c);
       } else {
         serialize_link_of_type<OT::HBUINT16> (link, head, c);
       }
     }
+    else
+      serialize_link_of_type<OT::HBUINT24> (link, head, c);
```
behdad, line 651: Can we rewrite this as a switch or otherwise not assume we're left with 24bit?

behdad:
LGTM. Thanks.

PR #2957, opened 2021-04-21
enable layout subsetting by default and several fixes for discrepancies

qxliu76:


garretrieger (approved):

behdad:
This is wrong. We cannot blindly drop all Device tables, because that's where variations are stored as well. So this change will remove variations if hinting is dropped!

Also, passing `drop_hinting` to `serialize` API is definitely a design mismatch.

The serialize API for involved tables already take `ValueFormat`. So you should pass the desired output `ValueFormat` there. Determining that is less than trivial unfortunately, since it would be the union of the value format needed to express the surviving values.

Hinting-related Device tables in GSUB/GPOS are extremely rare. I wonder where you even encountered them.

behdad:
FontTools fix: https://github.com/fonttools/fonttools/pull/2275

PR #2958, opened 2021-04-24
[subset] Enable layout subsetting by default.

garretrieger:
Fix all differences in the subset tests exposed by this change:
- Correct variation index collection.
- Match fontTool's ValueRecord optimization behaviour (dropping Value's that are zero where possible).


behdad:
Thanks Garret.

I believe it's time to get a release out. :) @khaledhosny can you do that?

khaledhosny:
If someone can prepare the NEWS entries I can do it (or anyone, really, it is just a git tag and the CI will take care of the rest).

khaledhosny:
(git tag, and updated version in meson.build and Makefile.am)

behdad:
> If someone can prepare the NEWS entries I can do it

Subsetter now fully supports GSUB/GPOS/GDEF tables (including variations); as such, layout tables are retained by subsetter by default.

Thanks.

khaledhosny:
Done.

PR #2959, opened 2021-04-24
[hb-view] Use Chafa for terminal graphics if available

### LINKED ISSUE #2430: [hb-view] Use libchafa if available...
https://hpjansson.org/chafa/

We can also make it work without cairo with a bit of bitmap code.

hpjansson:
Link with and use Chafa if available. This produces high-quality terminal graphics using symbols or sixels according to the detected terminal capabilities.

Fixes #2430.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -154,6 +154,8 @@ if not get_option('cairo').disabled()
   endif
 endif
 
+chafa_dep = dependency('chafa', version: '>= 1.6.0', required: get_option('chafa'))
+
 conf = configuration_data()
 incconfig = include_directories('.')
 
@@ -179,6 +181,10 @@ endif
 
 if cairo_ft_dep.found()
   conf.set('HAVE_CAIRO_FT', 1)
+endif
+
+if chafa_dep.found()
+  conf.set('HAVE_CHAFA', 1)
 endif
 
 if graphite2_dep.found()
@@ -363,6 +369,7 @@ build_summary = {
     },
   'Dependencies used for command-line utilities':
     {'Cairo': conf.get('HAVE_CAIRO', 0) == 1,
+     'Chafa': conf.get('HAVE_CHAFA', 0) == 1,
     },
   'Additional shapers':
     {'Graphite2': conf.get('HAVE_GRAPHITE2', 0) == 1,
--- a/util/helper-cairo-ansi.cc
+++ b/util/helper-cairo-ansi.cc
@@ -26,9 +26,102 @@
 
 #include "helper-cairo-ansi.hh"
 #include "options.hh"
-
 #include "ansi-print.hh"
 
+#ifdef HAVE_CHAFA
+# include <chafa.h>
+
+/* Similar to ansi-print.cc */
+# define CELL_W 8
+# define CELL_H (2 * CELL_W)
+
+static void
+chafa_print_image_rgb24 (const void *data, int width, int height, int stride)
+{
+  ChafaTermInfo *term_info;
+  ChafaSymbolMap *symbol_map;
+  ChafaCanvasConfig *config;
+  ChafaCanvas *canvas;
+  GString *gs;
+  unsigned int cols = (width +  CELL_W - 1) / CELL_W;
+  unsigned int rows = (height + CELL_H - 1) / CELL_H;
+  gchar **environ;
+  ChafaCanvasMode mode;
+  ChafaPixelMode pixel_mode;
+
+  /* Adapt to terminal; use sixels if available, and fall back to symbols
+   * with as many colors as are supported */
+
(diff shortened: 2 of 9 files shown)


behdad:
Lovely!

One minor suggestion. Can we keep including and call the `ansi-print` code even when Chafa is available, such that we compile-check them, and let the compiler not use it?

hpjansson:
> One minor suggestion. Can we keep including and call the `ansi-print` code even when Chafa is available, such that we compile-check them, and let the compiler not use it?

Sure thing, it's done.

meson.build
```diff
@@ -154,6 +154,17 @@ if not get_option('cairo').disabled()
...
 
+chafa_dep = null_dep
+if not get_option('chafa').disabled()
+  chafa_dep = dependency('chafa', version: '>= 1.6.0', required: false)
+
+  if (not chafa_dep.found() and
+      cpp.get_id() == 'msvc' and
+      cpp.has_header('chafa.h'))
+    chafa_dep = cpp.find_library('chafa', required: false)
+  endif
```
khaledhosny, line 165: I’d not bother with MSVC here, and simplify this into:
```
chafa_dep = dependency('chafa', version: '>= 1.6.0', required: get_option('chafa'))
```

hpjansson: Done!

alerque (approved):

util/helper-cairo-ansi.cc
```diff
@@ -26,9 +26,102 @@
...
+
+  symbol_map = chafa_symbol_map_new ();
+  chafa_symbol_map_add_by_tags (symbol_map,
+                                (ChafaSymbolTags) (CHAFA_SYMBOL_TAG_BLOCK
+                                                   | CHAFA_SYMBOL_TAG_SPACE));
+
+  config = chafa_canvas_config_new ();
+  chafa_canvas_config_set_canvas_mode (config, mode);
+  chafa_canvas_config_set_pixel_mode (config, pixel_mode);
+  chafa_canvas_config_set_cell_geometry (config, 10, 20);
```
behdad, line 87: Curious. How are these numbers used?

behdad:
Humm. On my Mac terminal I'm getting bad rendering. This seems to be same as the chafa cmdline's default 24bit color which renders wrong as well:

Screen Shot 2021-04-26 at 12 30 49 PM

@hpjansson any ideas?

behdad:
I disabled the truecolor mode for now.

behdad:
> I disabled the truecolor mode for now.

After:

Screen Shot 2021-04-26 at 12 35 09 PM

behdad:
Oh. It might be my bad $TERM:
```
$ echo $TERM
xterm-256color
```

behdad:
Humm. I'm not setting it.

util/helper-cairo-ansi.cc
hpjansson: The cell geometry is the pixel size of a single character cell, and the canvas geometry is the size of the canvas in character cells. The cell size is only used in sixel mode: Let's say you have a canvas that's 10 cells wide, and we supply that character cells are 10 pixels wide. Chafa will then produce a sixel image that's 100 pixels wide, scaling the input image if necessary.

I decided to use 10x20 because that's the fixed cell size of VT240 and VT340 terminals (and possibly Windows Terminal too when/if it gets sixel support).

This will lead to sixel output being a bit bigger than expected on modern Linux terminals (character cells tend to be bigger than 10x20, but the aspect is usually about the same). It would be possible to get the cell size with the  `TIOCGWINSZ` ioctl (.ws_xpixel and .ws_ypixel fields divided by .ws_col and .ws_row). The `chafa` command-line tool does this, but I avoided it in this patch because it seemed minor and might make Harfbuzz less portable.

See https://github.com/hpjansson/chafa/blob/2af7edec1adfdf1825cfb0dba46e8aaed7504918/tools/chafa/chafa.c#L726-L753 for how I did it in `chafa`. I can add something similar to `hb-view` if you think it's worth it.

Also see [here](https://gitlab.gnome.org/GNOME/vte/-/issues/312) for a long and somewhat opinionated discussion of sixel/cell dimensions arcana :)

hpjansson:
Disabling the truecolor mode seems fine as a stopgap. Does your terminal set anything else in the environment that would make it possible for me to identify it? E.g. VTE sets COLORTERM and VTE_VERSION. The issue is that there's a whole bunch of terminals out there that support truecolor/directcolor but identify themselves as xterm-256color.

Worst case, I'll have to bite the bullet and make Chafa more conservative when identifying the terminal.

behdad:
Here:
```
TERM_PROGRAM=Apple_Terminal
TERM=xterm-256color
```

PR #2961, opened 2021-04-30
[subset] correct ClassDef format determination.

garretrieger:
It was possible for glyph_min to be a glyph mapped to class zero which threw off the number of glyph calculation.


behdad:
Thanks. Bots not very happy.

PR #2964, opened 2021-05-01
[ci] Drop macos-10_13_6-aat-fonts job

khaledhosny:
This version of macOS is no longer supported by Homebrew, it takes 26 minutes to brew the dependencies before it fails.


behdad:
Lgtm. Thanks.

PR #2968, opened 2021-05-04
Update glyph properties during AAT substitutions so that mark zeroing works as expected

### LINKED ISSUE #2967: OT shaping fails to form mixed skin tone emoji sequence, CoreText succeeds
For the WOMAN MAN HOLDING HANDS sequence,

With ToT build:

`build/util/hb-shape /System/Library/Fonts/Apple\ Color\ Emoji.ttc -u 1F469,1F3FD,200D,1F91D,200D,1F468,1F3FE`

I get

`[u1F469.3.L=0+800|space=0+0|space=0+0|u1F468.4.RA=0+800]`

Expected, with CoreText:

`build/util/hb-shape --shaper=coretext /System/Library/Fonts/Apple\ Color\ Emoji.ttc -u 1F469,1F3FD,200D,1F91D,200D,1F468,1F3FE`

`[u1F469.3.L=0+0|u1F468.4.RA=5+1066]`

CC @jfkthame

jfkthame:
When zeroing by GDEF late, we need to re-check the GDEF table
(if it is the source of glyph classes) because substitutions may
mean that we now have mark glyphs in slots that were not flagged
as such initially.

Fixes #2967.

### DIFF
--- a/src/hb-ot-shape.cc
+++ b/src/hb-ot-shape.cc
@@ -193,6 +193,12 @@ hb_ot_shape_planner_t::compile (hb_ot_shape_plan_t           &plan,
 				   script_fallback_mark_positioning;
 
 #ifndef HB_NO_AAT_SHAPE
+  /* If we're using morx shaping, we cancel mark position adjustment because
+     Apple Color Emoji assumes this will NOT be done when forming emoji sequences;
+     https://github.com/harfbuzz/harfbuzz/issues/2967. */
+  if (plan.apply_morx)
+    plan.adjust_mark_positioning_when_zeroing = false;
+
   /* Currently we always apply trak. */
   plan.apply_trak = plan.requested_tracking && hb_aat_layout_has_tracking (face);
 #endif
@@ -1164,7 +1170,7 @@ _hb_ot_shape (hb_shape_plan_t    *shape_plan,
  * @lookup_indexes: (out): The #hb_set_t set of lookups returned
  *
  * Computes the complete set of GSUB or GPOS lookups that are applicable
- * under a given @shape_plan. 
+ * under a given @shape_plan.
  *
  * Since: 0.9.7
  **/
--- a/src/hb-aat-layout-common.hh
+++ b/src/hb-aat-layout-common.hh
@@ -30,6 +30,9 @@
 #include "hb-aat-layout.hh"
 #include "hb-open-type.hh"
 
+namespace OT {
+struct GDEF;
+};
 
 namespace AAT {
 
@@ -863,6 +866,7 @@ struct hb_aat_apply_context_t :
   hb_buffer_t *buffer;
   hb_sanitize_context_t sanitizer;
   const ankr *ankr_table;
+  const OT::GDEF *gdef_table;
 
   /* Unused. For debug tracing only. */
   unsigned int lookup_index;
--- a/src/hb-aat-layout-morx-table.hh
+++ b/src/hb-aat-layout-morx-table.hh
@@ -30,6 +30,7 @@
 #include "hb-open-type.hh"
 #include "hb-aat-layout-common.hh"
 #include "hb-ot-layout-common.hh"
+#include "hb-ot-layout-gdef-table.hh"
 #include "hb-aat-map.hh"
 
 /*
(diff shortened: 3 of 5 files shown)


behdad:
Based on my comment on the issue, I suggest this is not the approach we want to take.

https://github.com/harfbuzz/harfbuzz/issues/2967#issuecomment-832246409

What do you think?

jfkthame:
> Based on my comment on the issue, I suggest this is not the approach we want to take.
> 
> [#2967 (comment)](https://github.com/harfbuzz/harfbuzz/issues/2967#issuecomment-832246409)
> 
> What do you think?

SGTM; I've pushed a new version that appears to work as expected in local testing.

src/hb-ot-shape.cc
```diff
@@ -923,15 +923,19 @@ zero_mark_width (hb_glyph_position_t *pos)
...
 {
+  unsigned int prev_cluster = (unsigned int) -1;
   unsigned int count = buffer->len;
   hb_glyph_info_t *info = buffer->info;
   for (unsigned int i = 0; i < count; i++)
+  {
     if (_hb_glyph_info_is_mark (&info[i]))
     {
-      if (adjust_offsets)
+      if (adjust_offsets && likely (info[i].cluster == prev_cluster))
```
behdad, line 933: This can't be good. Fails with other cluster levels. I need to study why you are doing this. Typically we just insert a dotted-circle if there's no base before a mark. But that can only happen at start of text.

jfkthame: The motivation for this is that the Apple Color Emoji case here (the couple-with-skintones-holding-hands sequence) relies on substituting the first "person" glyph (MAN or WOMAN) with a version that has GDEF class = 3, and expects that to result in zeroing its advance. But it must not be repositioned leftwards when its advance gets zeroed, because it isn't supposed to overprint the preceding character; it's supposed to stay in place so that the second "person" glyph overprints it.

But you're right, this would be a problem with other cluster levels...

(And there's no dotted circle because although this is now at the start of text in the testcase, it wasn't a mark glyph at the time when dotted-circle insertion would have happened. It only becomes a mark during morx substitution.)

It would've been easier, perhaps, if they'd simply given these holding-hands people glyphs half the advance width each, instead of creating glyphs with the full width but the person image only in half of the space, and then overprinting the two glyphs at the same position -- using the GDEF class to trigger the zeroing of the first glyph's advance.

behdad: Right...

I feel like we should disable that adjustment for morx/kerx somehow. That is, let's adjust the logic here:
```
  plan.zero_marks = script_zero_marks &&
                    !plan.apply_kerx &&
                    (!plan.apply_kern
#ifndef HB_NO_OT_KERN
                     || !hb_ot_layout_has_machine_kerning (face)
#endif
                    );
  plan.has_gpos_mark = !!plan.map.get_1_mask (HB_TAG ('m','a','r','k'));

  plan.adjust_mark_positioning_when_zeroing = !plan.apply_gpos &&
                                              !plan.apply_kerx &&
                                              (!plan.apply_kern
#ifndef HB_NO_OT_KERN
                                               || !hb_ot_layout_has_cross_kerning (face)
#endif
                                              );

  plan.fallback_mark_positioning = plan.adjust_mark_positioning_when_zeroing &&
                                   script_fallback_mark_positioning;
```

jfkthame: OK, I think we can just disable adjust_mark_positioning_when_zeroing (but not fallback_mark_positioning) if we're using apply_morx. This resolves the problem here, and so far doesn't seem to break expectations anywhere else. I've updated the PR to use this approach.

Also added a testcase for the sequence involved here. Maybe @ebraminio can check if older versions of the font supported it (might be quite a recent addition?), and if so add the appropriate entries for wider coverage.

ebraminio: I don't have access to older version of the macOS fonts, I used to access those in the CI bots but Circle CI won't let me log into it anymore as they have added some country based blocks.

drott: Thanks very much for the fix!

> Also added a testcase for the sequence involved here. Maybe @ebraminio can check if older versions of the font supported it (might be quite a recent addition?), and if so add the appropriate entries for wider coverage.

According to https://blog.emojipedia.org/ios-13-2-emoji-changelog/ the first release that had the mixed holding hands emoji was iOS 13 (September 2019). Mac OS 10.15 came out after (October 2019) that so maybe 10.15 Apple Color Emoji may have it. I only have archived copies of 10.10 to 10.13 here and in 10.13 definitely the sequence does not exist as a combination.

Issue #2972, opened 2021-05-06
Warnings with gcc 8 and earlier


tstuefe:
I'm from the OpenJDK project. We recently upgraded to HB 2.8.0. Since then, the build is broken for older gcc releases (tried 8 and 7):

```
In file included from /shared/projects/openjdk/jdk-jdk/source/src/java.desktop/share/native/libharfbuzz/hb-ot-shape-complex-use.cc:33:0:
hb-ot-shape-complex-use-machine.rl: In instantiation of 'machine_index_t<Iter>::machine_index_t(const machine_index_t<Iter>&) [with Iter = hb_zip_iter_t<hb_iota_iter_t<unsigned int, unsigned int>, hb_filter_iter_t<hb_filter_iter_t<hb_zip_iter_t<hb_iota_iter_t<unsigned int, un
signed int>, hb_array_t<hb_glyph_info_t> >, find_syllables_use(hb_buffer_t*)::<lambda(const hb_glyph_info_t&)>, const<unnamed struct>&, 0>, find_syllables_use(hb_buffer_t*)::<lambda(hb_pair_t<unsigned int, const hb_glyph_info_t&>)>, const<unnamed struct>&, 0> >]':
hb-ot-shape-complex-use-machine.rl:249:11: required from here
hb-ot-shape-complex-use-machine.rl:194:9: error: base class 'struct hb_iter_with_fallback_t<machine_index_t<hb_zip_iter_t<hb_iota_iter_t<unsigned int, unsigned int>, hb_filter_iter_t<hb_filter_iter_t<hb_zip_iter_t<hb_iota_iter_t<unsigned int, unsigned int>, hb_array_t<hb_glyp
h_info_t> >, find_syllables_use(hb_buffer_t*)::<lambda(const hb_glyph_info_t&)>, const<unnamed struct>&, 0>, find_syllables_use(hb_buffer_t*)::<lambda(hb_pair_t<unsigned int, const hb_glyph_info_t&>)>, const<unnamed struct>&, 0> > >, hb_pair_t<unsigned int, hb_pair_t<unsigned
 int, hb_glyph_info_t&> > >' should be explicitly initialized in the copy constructor [-Werror=extra]
cc1plus: error: unrecognized command line option '-Wno-class-memaccess' [-Werror]
cc1plus: all warnings being treated as errors
[...]

PR #2973, opened 2021-05-06
Fix for: Warnings with gcc 8 and earlier #2972

tstuefe:
Hi,

Please consider this fix for a build warning with gcc 8 and earlier. This causes problems in OpenJDK, see issue for details.

Thank you!

..Thomas


behdad:
Thanks Thomas. I'll review tomorrow.

tstuefe:
Thanks @behdad !

PR #2977, opened 2021-05-13
[subset] subset both CPAL and COLRv1

qxliu76:
Final PR for subsetting COLRv1.
Some tests are not enabled because fonttools crashes while generating subset files.


drott:
Cool stuff, great to see COLRv1 subsetting already!

garretrieger (approved):

behdad:
LGTM when @garretrieger's happy.

PR #2980, opened 2021-05-18
Chromium build fixes for C++ 17 warning and missing _remap_indexes

### LINKED ISSUE #2979: Chromium build failures with 89df3539966d7d544b49495a667ccebf94bb045c
Chromium build complains about

```
template <template<typename> template Var>
HB_INTERNAL void PaintTransform<Var>::closurev1 (hb_colrv1_closure_context_t* c) const
{
```

being a C++ 17 extension and can be fixed by changing `template Var` to `class Var`.

in `hb-subset-plan.cc` `_remap_indexes` is now used outside of `#ifndef HB_NO_SUBSET_LAYOUT`. 

Will submit a PR.

drott:
Use class instead of typename, move _remap_indexes out of #ifndef.

Fixes #2979


drott:
Thanks for the quick merge.

PR #2981, opened 2021-05-19
[subset] Add ability for subsetter to set the overlaps flag on all glyphs.

garretrieger:
Equivalent to SetMacOverlapFlags in fontTools: https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/varLib/instancer/__init__.py#L1069


behdad:
Is ugly to add it in the subsetter but LGTM.

PR #2982, opened 2021-05-19
[set] Compute is_subset by comparing pages.

kkartaltepe:
Test subsets one page at a time instead of by codepoints. On my machine
this is about 250x (not enough samples in the profiler after so this could be
off by a lot) faster than the previous implementation.

I have no clue what the preferred style is for this project so please let me know if if something is off.

### DIFF
--- a/src/hb-set.hh
+++ b/src/hb-set.hh
@@ -506,14 +506,44 @@ struct hb_set_t
 
   bool is_subset (const hb_set_t *larger_set) const
   {
-    if (get_population () > larger_set->get_population ())
-      return false;
+    if (unlikely(larger_set->is_empty ()))
+    {
+      return is_empty ();
+    }
 
-    /* TODO Optimize to use pages. */
-    hb_codepoint_t c = INVALID;
-    while (next (&c))
-      if (!larger_set->has (c))
-	return false;
+    uint32_t spi = 0;
+    for (uint32_t lpi = 0; spi < page_map.length && lpi < larger_set->page_map.length; lpi++)
+    {
+      uint32_t spm = page_map[spi].major;
+      uint32_t lpm = larger_set->page_map[lpi].major;
+      auto sp = page_at (spi);
+      auto lp = larger_set->page_at (lpi);
+
+      if (spm < lpm && !sp.is_empty ())
+      {
+        return false;
+      }
+
+      if (lpm < spm)
+      {
+        continue;
+      }
+
+      for (int j = 0; j < ARRAY_LENGTH_CONST (sp.v); j++)
+      {
+        if ((~lp.v[j] & sp.v[j]) != 0) { return false; }
+      }
+
+      spi++;
+    }
+
+    while (spi < page_map.length)
+    {
+      if (!page_at (spi++).is_empty ())
+      {
+        return false;
+      }
+    }
 
     return true;
   }
--- a/test/api/test-set.c
+++ b/test/api/test-set.c
@@ -216,6 +216,58 @@ static void test_set_union (void)
 }
 
 static void
+test_set_subsets (void)
+{
+  hb_set_t *s = hb_set_create ();
+  hb_set_t *l = hb_set_create ();
+
+  hb_set_add (l, 0x0FFFF);
+  hb_set_add (s, 0x1FFFF);
+  g_assert (!hb_set_is_subset (s, l));
+  hb_set_clear (s);
+
+  hb_set_add (s, 0x0FFF0);
+  g_assert (!hb_set_is_subset (s, l));
+  hb_set_clear (s);
+
+  hb_set_add (s, 0x0AFFF);
+  g_assert (!hb_set_is_subset (s, l));
+
+  hb_set_clear (s);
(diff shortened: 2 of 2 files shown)


behdad:
Thanks! Curious, how did you end up noticing this?

kkartaltepe:
I was sorting fonts using the unicode coverage from hb_face_t but it was painfully slow compared to fontconfig, when I profiled all the time was in this function.

kkartaltepe:
Small update, the case where the smaller set had elements in a higher page was not checked correctly and getting the next page's major may have read past the page_map boundary. Should all be correct now.

src/hb-set.hh
```diff
@@ -506,16 +506,34 @@ struct hb_set_t
 
   bool is_subset (const hb_set_t *larger_set) const
   {
-    if (get_population () > larger_set->get_population ())
-      return false;
+	if(unlikely(larger_set->page_map.length == 0)) {
+		return false;
```
garretrieger, line 510: Should be:
`return this->is_empty ();`

Since if this set has no elements then it is a subset of larger_set, which has no elements.

src/hb-set.hh
```diff
@@ -506,16 +506,34 @@ struct hb_set_t
...
+	}
+	if(page_map[page_map.length].major > larger_set->page_map[larger_set->page_map.length].major) {
```
garretrieger, line 512: Pages may exist that are all zeros, so it isn't valid to return early here.

src/hb-set.hh
```diff
@@ -506,16 +506,34 @@ struct hb_set_t
...
-    /* TODO Optimize to use pages. */
-    hb_codepoint_t c = INVALID;
-    while (next (&c))
-      if (!larger_set->has (c))
-	return false;
+	uint32_t spi = 0;
+	for(uint32_t lpi = 0; spi < page_map.length && lpi < larger_set->page_map.length; lpi++) {
+		uint32_t spm = page_map[spi].major;
+		uint32_t lpm = larger_set->page_map[lpi].major;
+		if(spm < lpm) {
```
garretrieger, line 520: similar thing here, the page at spi may be all zeros. So you'll only want to return false if `!sp.is_empty()`

src/hb-set.hh
```diff
@@ -506,16 +506,34 @@ struct hb_set_t
 
   bool is_subset (const hb_set_t *larger_set) const
```
garretrieger, line 507: Can you update the formatting of the new code to match the style in this file. Notably:
- 2 space indent
- if( -> if (
- function_call( -> function_call (
- { should be on their own line

kkartaltepe: Done

src/hb-set.hh
kkartaltepe: Done

src/hb-set.hh
kkartaltepe: Fixed up the check to ensure small set pages beyond the large set are all empty.

src/hb-set.hh
kkartaltepe: Ah nice I didnt see those earlier, updated to use some of the convenience functions.

kkartaltepe:
Updated for comments, extended the subset tests to catch some common mistakes.

garretrieger (approved):

behdad:
Thanks again @kkartaltepe. I've been also meaning to look into replacing fontconfig with something based on harfbuzz. Would love to hear about your findings.

kkartaltepe:
At least for my simple case most things were pretty smooth.

I only deal with `names` derived properties, coverage, and spacing. The first two are all available directly and with this change coverage is fast to compare. For spacing I iterate over `hb_set_next`  to check spacings via `hb_font_get_nominal_glyph` and `hb_font_get_glyph_h_advance` similar to what FC does. I think this is probably slower than FC, but I didnt check as it was fast enough for me on startup with the full noto font set.

Since I didn't want to implement any table parsing myself I was quite pleased with the existing harfbuzz interfaces. If I wanted to do anything tricky referencing tables directly I assume referencing the face/table blobs would be sufficient. Maybe the only gap was that old style bitmap tables are not exposed just emoji tables (hb-ot-color).

behdad:
Cool.

I think FC bails as soon as it determines the spacing is proportional. That's the only trick it has in its sleeve. Other than that, loading this using HB should be faster than FT already.

If you want to make it even faster, you can use the plural versions, ie. `hb_font_get_nominal_glyph` and `hb_font_get_glyph_h_advances`, to batch them, to reduce overhead inside HB.

behdad:
> Since I didn't want to implement any table parsing myself I was quite pleased with the existing harfbuzz interfaces. If I wanted to do anything tricky referencing tables directly I assume referencing the face/table blobs would be sufficient.

Right. We have a lot of facilities to make table-parsing easier; so that part is not exposed. I like to make that available as a C++ (header-only?) library eventually.

> Maybe the only gap was that old style bitmap tables are not exposed just emoji tables (hb-ot-color). But I can at least check for the tables.

Right. Isn't impossible to add..

PR #2983, opened 2021-05-20
[subset] support option: --layout-features

qxliu76:


src/hb-ot-layout.h
```diff
@@ -312,15 +312,17 @@ hb_ot_layout_collect_features (hb_face_t      *face,
 			       hb_tag_t        table_tag,
 			       const hb_tag_t *scripts,
 			       const hb_tag_t *languages,
-			       const hb_tag_t *features,
+			       const hb_set_t *features,
```
garretrieger, line 315: We can't change the signature of an existing API method. Instead I would add a second version method that takes features as a set. Then that could internally convert features to an array and call this method, or you could update this method to convert the array to a set and call the new one.

Additionally I'd drop the retain_all_features bool and instead stick with the current behaviour that if the features set is a nullptr that means keep all.

khaledhosny: > if the features set is a nullptr that means keep all.

What if one wants to drop all features?

garretrieger: Passing an empty set would drop all features.

khaledhosny: Right.

src/hb-subset-plan.cc
```diff
@@ -68,20 +68,95 @@ _remap_indexes (const hb_set_t *indexes,
...
+				lookup_indices);
+
+  free (features);
+}
+
+static void _collect_subset_features (hb_face_t      *face,
+				      hb_tag_t        table_tag,
+				      const hb_set_t *layout_features_to_retain,
+				      bool            retain_all_features,
+				      hb_set_t       *lookup_indices /* OUT */)
```
garretrieger, line 113: This function and the above are pretty much identical except for the function they call. You should be able to have just one function that takes a function point to the collect function to call.

src/hb-subset-plan.cc
```diff
@@ -68,20 +68,95 @@ _remap_indexes (const hb_set_t *indexes,
...
+				 table_tag,
+				 nullptr,
+				 nullptr,
+				 features,
+				 lookup_indices);
+
+  free (features);
+}
 static inline void
 _gsub_closure_glyphs_lookups_features (hb_face_t *face,
```
garretrieger, line 146: This function and _gpos_closure_glyphs_lookups_features are basically identical (except gsub has an additional call to   hb_ot_layout_lookups_substitute_closure). They could be merged into one function by templating the function with the table class (ie. OT:GSUB/OT:GPOS). The tag can be grabbed from the table class via 'tableTag' and then you can put an if (tableTag == HB_OT_TAG_GSUB) hb_ot_layout_lookups_substitute_closure(...

test/subset/data/tests/basics.tests
```diff
@@ -12,6 +12,7 @@ name-ids.txt
 name-languages.txt
 name-legacy.txt
 gids.txt
+layout-features.txt
```
garretrieger, line 16: Can you also add a test that uses --layout-features=*

behdad:
LGTM w Garret's comments addressed. Thanks.

PR #2985, opened 2021-05-22
[meson] fix warning: use feature which were added in newer versions

inkydragon:
```
WARNING: Project targeting '>= 0.47.0' but tried to use feature introduced in '0.50.0': install arg in configure_file.

harfbuzz| * 0.50.0: {'install arg in configure_file'}
```


khaledhosny:
The warning is erroneous, the use of newer features is guarded by version check.

khaledhosny:
See https://github.com/harfbuzz/harfbuzz/pull/2744

PR #2988, opened 2021-05-26
[meson] Remove wraps we don’t use directly

khaledhosny:


khaledhosny:
We should also switch freetype2.wrap from https://github.com/centricular/freetype2 fork to upstream, but that is for another day.

behdad:
Thanks.

PR #2993, opened 2021-05-27
fuzzer fix: null pointer dereference

qxliu76:


behdad:
We don't expect `current` to be `nullptr`. Do you have more info on why / when this happens? I feel like the best fix belongs somewhere else.

qxliu76:
> We don't expect `current` to be `nullptr`. Do you have more info on why / when this happens? I feel like the best fix belongs somewhere else.

it happens in COLR subset(), after V0 part, I'll try to get more details

qxliu76:
> We don't expect `current` to be `nullptr`. Do you have more info on why / when this happens? I feel like the best fix belongs somewhere else.

The issue is that calloc () failed in hb_pool_t::alloc (), which is called by serializer push () in start_serialize (). This happened before subset method.

garretrieger (approved):

behdad:
Much better. Thank you.

PR #2995, opened 2021-06-02
Removed unused variable `supp_size` from plan_subset_encoding(...).

abpostelnicu:
Variable `supp_size` is set but unused and on newer clang versions, see [this](https://reviews.llvm.org/rGcf49cae278b4e972cd2547d72f9ee7d9d69a3af4), this represents an issue.


abpostelnicu:
@behdad any idea who can review this? Thanks.

abpostelnicu:
@jfkthame do you think we can have this in the next version, for the time being we will have to have a patch locally to `mozilla-central` in order to be able to build with `clang-trunk`.

Issue #1829, opened 2019-07-11
[graphite2] Threadsafety?


behdad:
The code in hb-graphite2.cc, in particular `hb_graphite2_get_table` and `hb_graphite2_release_table`  looks suspicious.  One *cannot* safely remove a linked-list element and destroy it as another thread might be traversing the list at the same location...  We should just keep all tables and destroy them at the end.

This might be the cause for https://bugzilla.redhat.com/show_bug.cgi?id=1689037

cc @mhosken

behdad:
Going to revert e4e74c2751ac24178086cce2811d34d8019b6f85 since seems to be causing crashes:
https://bugzilla.redhat.com/show_bug.cgi?id=1689037#c32

behdad:
I'm not quite sure *why* we hit this in anaconda.  But it's clear that this is causing the issue.

mhosken:
The way harfbuzz loads a graphite face, (using preloadAll), all font tables are released by the time the gr_make_face_with_ops call returns. This means that the engine is single threaded for the purposes of loading a face. But after that access may be multithreaded since none of its face data is written to during segment creation and also no tables are requested or released during segment creation. I'm not sure if that helps.

I'm not sure why two different threads would be accessing the same data blob of tables given all read and write to that linked list is inside the one function gr_make_face_with_ops, which is a single function thread.

behdad:
I'm not familiar with graphite2 code to comment.  But clearly old code used to work and new work was buggy.  Can we just close this?  Any reason why the new code was preferable?

mhosken:
No. Just trying to be tidy. Feel free to revert.

behdad:
I reverted already.  While I like to get to the bottom of what's happening, I won't have time for it.  Closing, unless someone wants to investigate.

PR #2999, opened 2021-06-03
Avoid a deprecation warning in graphite2

pkgw:
As of graphite2 1.3.7, `gr_make_face` is deprecated in favor of `gr_make_face_with_ops`. It's a one-liner to port over to using it.

This is potentially a compatibility break since I'm not sure when the `with_ops` API was added, but the minimum version of graphite2 that's supported by Harfbuzz doesn't seem to be documented anywhere anyway.

### DIFF
--- a/src/hb-graphite2.cc
+++ b/src/hb-graphite2.cc
@@ -128,7 +128,8 @@ _hb_graphite2_shaper_face_data_create (hb_face_t *face)
     return nullptr;
 
   data->face = face;
-  data->grface = gr_make_face (data, &hb_graphite2_get_table, gr_face_preloadAll);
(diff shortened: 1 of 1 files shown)


khaledhosny:
See https://github.com/harfbuzz/harfbuzz/issues/1829

pkgw:
In the version of graphite2 that I'm using, at least, the current function is literally a two-liner equivalent to the code in my PR, so I don't see how using the new API could cause any new thread-safety issues. Maybe that wasn't always the case?

behdad:
The issue with #1829 was that it also introduced a `release_table` callback. Without it, as is in this PR, I don't think there will be any threading issue. Thanks.

PR #3005, opened 2021-06-04
Dw cleanup

arzanilorestan:
A very, very important article by me (Satoshi Nakamoto), the true and true creator of digital currencies for the whole world .____ Important ___ Important ___ Important ___ Important ___ Important ___ Important ___
http://sendanywhe.re/ZHTRREP9

### DIFF
--- a/src/hb-directwrite.cc
+++ b/src/hb-directwrite.cc
@@ -49,16 +49,6 @@ typedef HRESULT (* WINAPI t_DWriteCreateFactory)(
   IUnknown            **factory
 );
 
-/*
- * hb-directwrite uses new/delete syntatically but as we let users
- * to override malloc/free, we will redefine new/delete so users
- * won't need to do that by their own.
- */
-void* operator new (size_t size)        { return malloc (size); }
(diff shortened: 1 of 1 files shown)


behdad:
Can we keep the experimental width version but mark it static? Just want to have it for future reference. Though, we can always look back in history.

The dynamic-loading, I agree shouldn't be necessary since we don't expect to support systems old-enough to not have dwrite. Though if we ever want to use newer features (variations, etc), then dynamic loading might become necessary again. We do that in uniscribe as well. So maybe leaving that one as is is fine.

behdad:
Yeah I think I'm fine with these all. We can always revive any bits.

khaledhosny:
I didn't open a PR because this is still work in progress, furthermore this looks like a spam account. I don't know if we can block it, at least it should be reported.

behdad:
> I didn't open a PR because this is still work in progress, furthermore this looks like a spam account. I don't know if we can block it, at least it should be reported.

I reported and blocked it already. Ah, I did wonder why it's the first comment. Haha.

Issue #3008, opened 2021-06-02
GPOS does not apply only in some fonts


kojiishi:
I'm not entirely sure whether this is an issue of uharfbuzz or not, but I can't figure it out after some days of investigations. Appreciate any feedback in advance.

I'm porting my Python program from `hb-shape` to `uharfbuzz`. This is so much faster and looks great, thank you for the work. But I've found one different behavior I could not figure out why. The difference occurs:
* Only in two fonts on Mac; 'ヒラギノ丸ゴ ProN W4.ttc' and 'ヒラギノ明朝 ProN.ttc'. I tested other Japanese fonts on Mac, Noto CJK, and some CJK fonts on Windows. So far only these two fonts hit this issue.
* `hb-shape` works fine for these two fonts.

The [test code at gist](https://gist.github.com/kojiishi/1c6b2f616a88d251b2f749102a51ae1b). Here's an sample output of the failure:
```
start table GSUB
start lookup 2
end lookup 2
start lookup 4
end lookup 4
start lookup 7
end lookup 7
start lookup 8
end lookup 8
start lookup 16
end lookup 16
start lookup 19
end lookup 19
end table GSUB
start subtable 0
end subtable 0
gid781=0@1000,0+0
hb-shape '--font-file=fonts/g/mac/10.15.7/Hiragino Mincho ProN.ttc' --trace '--text=１' --features=palt
trace: start lookup 2   buffer: [gid781=0]
trace: end lookup 2     buffer: [gid781=0]
trace: start lookup 4   buffer: [gid781=0]
trace: end lookup 4     buffer: [gid781=0]
trace: start lookup 7   buffer: [gid781=0]
trace: end lookup 7     buffer: [gid781=0]
trace: start lookup 8   buffer: [gid781=0]
trace: end lookup 8     buffer: [gid781=0]
trace: start lookup 16  buffer: [gid781=0]
trace: end lookup 16    buffer: [gid781=0]
trace: start lookup 19  buffer: [gid781=0]
trace: end lookup 19    buffer: [gid781=0]
trace: start lookup 1   buffer: [gid781=0+1000]
trace: end lookup 1     buffer: [gid781=0+1000]
trace: start lookup 2   buffer: [gid781=0+1000]
trace: end lookup 2     buffer: [gid781=0@-106,0+783]
[gid781=0@-106,0+783]
```
The advance is 783 when using `hb-shape`, but 1000 when using `uharfbuzz`. The output of other fonts is:
```
[...]

PR #3009, opened 2021-06-06
[aat] Add start table/end table to buffer messages

khaledhosny:
Related to https://github.com/harfbuzz/harfbuzz/issues/3008


behdad:
Weird. I thought @simoncozens added those already.

PR #3015, opened 2021-06-08
Config.h filename configuration

1bsyl:
Hi,

We want to be able to have several config.h configuration, for various platform.
Maybe we can add HB_CONFIG_H_FILENAME and HB_CONFIG_OVERRIDE_H_FILENAME to allow config filenames customization.

### DIFF
--- a/src/hb-config.hh
+++ b/src/hb-config.hh
@@ -32,7 +32,10 @@
 #endif
 
 #ifdef HAVE_CONFIG_H
-#include "config.h"
+#ifndef HB_CONFIG_H_FILENAME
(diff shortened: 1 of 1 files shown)


behdad:
You can create your own `config.h` and `config-override.h` that internally `#include HB_CONFIG_H_FILENAME` etc. Multiarch distros have been doing that for years.

behdad:
That said, I'm open to this.

behdad:
Sure. Send a PR.

1bsyl:
Goal is to keep several different config.h in our repository
(one for android, linux, mingw probably .. )
to still be able to compile for each platform

PR #3020, opened 2021-06-11
[subset] Always serialize objects point to by OffsetTo to a new serializer object.

garretrieger:
Fixes GSUB/GPOS table size inflation noticed in #2986

There's an often used pattern in the current serialization code to serialize an object point to an offset by calling serialize() on the OffsetTo which embeds the pointed to structure at the current serializer head. This bypasses the push()/pop_pack() mechanism which can prevent the point to structure from being deduplicated. Ultimately this can result in inflated table sizes.

This PR introduces a new serialize_serialize() method on OffsetTo which wraps push()/pop_pack() around the serialization of the pointed to structure to enable correct deduplication.

### DIFF
--- a/src/hb-open-type.hh
+++ b/src/hb-open-type.hh
@@ -355,6 +355,23 @@ struct OffsetTo : Offset<OffsetType, has_null>
     return ret;
   }
 
+
+  template <typename ...Ts>
+  bool serialize_serialize (hb_serialize_context_t *c, Ts&&... ds)
+  {
+    *this = 0;
+
+    Type* obj = c->push<Type> ();
+    bool ret = obj->serialize (c, hb_forward<Ts> (ds)...);
+
+    if (ret)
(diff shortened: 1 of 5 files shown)


behdad:
Ah, thanks! I'm fairly sure I had an issue open to remove that old pattern that predates the push/pop mechanism.

behdad:
LGTM!

Any idea why tests are failing?

behdad:
Humm. Seems to break GSUB.

garretrieger:
> Humm. Seems to break GSUB.

Strange, I'll track it down.

garretrieger:
> > Humm. Seems to break GSUB.
> 
> Strange, I'll track it down.

Alright, found it. The arabic shaping code can serialize lookups which it does from methods like "serialize_single" on SubstLookup. These are not using the push()/pop_pack() mechanism either so the base for the offset ends up wrong.

Issue #3012, opened 2021-06-08
[subset] API (and ABI) stability of subset


mapreri:
Hi,

scribus has started (with version 1.5.7) to optionally link to hb-subset if available.

While packaging that in Debian, I've been told that hb-subset is not publicly available because back in 2018 it was marked as "API still subject to changes" (and I suppose ABI too, which from my POV is more important than the API in this case).

See https://bugs.debian.org/988781

Could I ask what's the status of hb-subset and if it can be considered API/ABI stable like the rest of harfbuzz?

behdad:
Thanks. @garretrieger let's sit down and do a review of the `hb-subset` API, create roadmap to stablization. I can do that tomorrow if we start early.

@mapreri thanks for the report. I'm interested to hear more about the scribus experience.

mapreri:
> @mapreri thanks for the report. I'm interested to hear more about the scribus experience.

I asked and these is what I got:

> for now we had no issue with harfbuzz-subset API stability
> other distros such as OpenSuse are already packaging it
> fox now we are using harfbuzz-subset only for CID keyed fonts
> so mostly asian fonts

garretrieger:
Alright me and Behdad worked out a plan of what we want to change before making the API stable:
- hb_subset(...) the input should be a const.
- Add set inversion to allow for efficiently specifying all things in a set. Drop the retain all features method.
- Expose set/get user data methods.
- name lang id's should work with hb_lang_t's instead of a set.
- For boolean properties use an enum instead of a method for each property.
- Add a pass thru tables set.

garretrieger:
@qxliu76 FYI

garretrieger:
For the boolean properties, here's a proposed signature:

enum hb_subset_input_property_t {
  ...
};

void hb_subset_input_set_property (hb_subset_input_property_t property, bool value);
bool hb_subset_input_get_property (hb_subset_input_property_t property);

behdad:
> For the boolean properties, here's a proposed signature:

alternatively suggesting: `set_bool`/`get_bool`.

behdad:
> hb_subset_input_property_t

Would this naming be confusing as well? Since we only want to do it for boolean properties, not everything.

khaledhosny:
Any thing else to be tracked here?

behdad:
> Maybe just a final check the .so to make sure exposed symbols are all good and intended, and proper release notes.

And maybe autoconf/meson options to disable building it?

[...]

PR #3021, opened 2021-06-11
[subset] update hb_subset api with final changes before going stable.

garretrieger:
Makes some changes to the hb-subset.h API before we stabilize it:

- Remove individual get/set methods for bool properties and instead use an enum of property names.
- hb_subset_input is now const on the hb_subset() call.
- Add hb_subset_or_fail ()
- Adds set/get user data methods that are standard on hb_objects.
- Adds passthrough tables and no subset tables to match fontTools table dropping and retention logic.

I've still got a couple things left to do before this is ready:
- Documentation on the properties enum.
- Documentation for all methods.
- Remove hb_subset() in favor of hb_subset_or_fail()
- Re-order get/set user data methods.


behdad:
Thanks Garret!

CC @jfkthame @drott @bungeman @mapreri #3012

drott:
Thanks, @garretrieger! API change generally LGTM but I am not too familiar with the API and its usage. `third_party/skia/src/pdf/SkPDFSubsetFont.cpp` owners are better suited to answer.

What's `hb_subset_input_set_user_data` used for? 

IIUC this will need a migration off of `hb_subset_input_set_retain_gids` to `hb_subset_set_flags` in Skia & Chromium. Usually the way that's done is to define a compile flag in Skia, set this compile flag in Chromium's top level `skia/BUILD.gn` `defines` list when rolling HarfBuzz in Chromium. Are you up for handling that migration?

behdad:
> What's `hb_subset_input_set_user_data` used for?

That's generic part of HB object model; so you can attach arbitrary (say, `SkFace`) data to a HB object.

behdad:
> That's generic part of HB object model; so you can attach arbitrary (say, `SkFace`) data to a HB object.

Sometimes used by language bindings.

garretrieger:
> IIUC this will need a migration off of `hb_subset_input_set_retain_gids` to `hb_subset_set_flags` in Skia & Chromium. Usually the way that's done is to define a compile flag in Skia, set this compile flag in Chromium's top level `skia/BUILD.gn` `defines` list when rolling HarfBuzz in Chromium. Are you up for handling that migration?

Yes I can try and do that migration.

src/hb-subset-input.cc
```diff
@@ -237,18 +256,43 @@ hb_subset_input_glyph_set (hb_subset_input_t *subset_input)
   return subset_input->glyphs;
 }
 
+/**
+ * hb_subset_input_nameid_set:
+ * @subset_input: a subset_input.
```
khaledhosny, line 261: That should be `an #hb_subset_input_t`.

src/hb-subset-input.cc
```diff
@@ -269,98 +313,153 @@ hb_subset_input_get_retain_all_features (hb_subset_input_t *subset_input)
...
+      break;
+  }
 }
 
 /**
- * hb_subset_input_get_retain_gids:
- * Returns: value of retain_gids.
- * Since: 2.4.0
+ * hb_subset_input_set_user_data: (skip)
+ * @face: A subset input object
```
khaledhosny, line 426: This should be `@input`. There probably should be some consistency in naming this variable, either input or subset_input.

src/hb-subset-input.cc
```diff
@@ -269,98 +313,153 @@ hb_subset_input_get_retain_all_features (hb_subset_input_t *subset_input)
...
 {
-  subset_input->notdef_outline = notdef_outline;
+  return hb_object_set_user_data (input, key, data, destroy, replace);
 }
 
-HB_EXTERN hb_bool_t
-hb_subset_input_get_notdef_outline (hb_subset_input_t *subset_input)
+/**
+ * hb_subset_input_get_user_data: (skip)
+ * @face: A subset input object
```
khaledhosny, line 450: This should be `@input`.

src/hb-subset-plan.cc
```diff
@@ -389,8 +389,8 @@ _nameid_closure (hb_face_t *face,
  * Since: 1.7.5
  **/
 hb_subset_plan_t *
-hb_subset_plan_create (hb_face_t         *face,
-		       hb_subset_input_t *input)
+hb_subset_plan_create (hb_face_t	 *face,
```
khaledhosny, line 392: The function arguments need to be documented.

khaledhosny (changes requested):
I don't think subset documentation is currently generated, appropriate section need to be added to the files unders /docs.

src/hb-subset-plan.cc
garretrieger: Done.

src/hb-subset-input.cc
garretrieger: Alright, I standardized everything to use "input".

behdad:
Thanks. Garret, can you also add a `SECTION:hb-subset` comment to `hb-subset.cc` to provide some description for this section of the docs? See top of `hb-shape.cc` for example.

garretrieger:
> Thanks. Garret, can you also add a `SECTION:hb-subset` comment to `hb-subset.cc` to provide some description for this section of the docs? See top of `hb-shape.cc` for example.

Done.

garretrieger:
> I don't think subset documentation is currently generated, appropriate section need to be added to the files unders /docs.

Done.

khaledhosny (approved):

Issue #3019, opened 2021-06-11
Build with LLVM 12 and Clang LTO Fails with check-static-inits. Meson replace linker.


h0tc0d3:
Build with LLVM 12 and Clang LTO  Fails: harfbuzz:src / check-static-inits FAIL. Meson replace linker.

```bash
export CC=clang
export CXX=clang++
export LD=ld.lld
export CC_LD=lld
export CXX_LD=lld
export AR=llvm-ar
export NM=llvm-nm
export STRIP=llvm-strip
export OBJCOPY=llvm-objcopy
export OBJDUMP=llvm-objdump
export READELF=llvm-readelf
export HOSTCC=clang
export HOSTCXX=clang++
export HOSTAR=llvm-ar
export HOSTLD=ld.lld

export CPPFLAGS="-D_FORTIFY_SOURCE=2"
export CFLAGS="-fdiagnostics-color=always -pipe -O2 -march=native -fstack-protector-strong"
export CXXFLAGS="-fdiagnostics-color=always -pipe -O2 -march=native -fstack-protector-strong -fuse-ld=lld"
export LDFLAGS="-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now"

meson harfbuzz build \
  --prefix /usr \
  --libexecdir lib \
  --sbindir bin \
  --auto-features enabled \
  --wrap-mode=nodownload \
  -D debug=false \
  -D optimization=2 \
  -D b_ndebug=true \
  -D b_lto=true \
  -D b_lto_mode=thin \
  -D glib=enabled \
  -D gobject=enabled \
  -D cairo=enabled \
  -D icu=enabled \
  -D graphite=enabled \
  -D freetype=enabled \
  -D chafa=enabled \
  -D introspection=enabled

```

`file hb-icu.cc.o` return `hb-icu.cc.o: LLVM IR bitcode`

<details>
  <summary>Error log</summary>

```text

22/361 harfbuzz:src / check-static-inits                                 FAIL             0.08s   exit status 1
[...]

PR #3024, opened 2021-06-12
[src/check-*] Pickup $(NM), $(OBJDUMP), $(LDD), $(OTOOL)

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3019


h0tc0d3:
Thank you. Not helped for me. The same error.

Issue #2567, opened 2020-07-14
hb_blob_create_from_file return the same thing when a file is not found or empty


ebraminio:
> Can you open an issue please? I have ideas.

https://github.com/harfbuzz/harfbuzz/commit/1339351105ce798ec5b701dfecc6d34b84273b4e#r40599376

Guess you are referring to this issue, honestly I like this uniform property of harfbuzz APIs as we even see a file system with the idea but I like to hear about what you are proposing.

behdad:
I suggest we add a `hb_blob_create_from_file_or_fail()` variant that returns `nullptr` if file can't be opened.

PR #3026, opened 2021-06-14
[blob] Add failing versions of create API

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/2567

New API:
+hb_blob_create_or_fail()
+hb_blob_create_from_file_or_fail()

Use these in util/ to distinguish empty file from not-found file.
Only err on the latter.

### DIFF
--- a/src/hb-blob.cc
+++ b/src/hb-blob.cc
@@ -72,14 +72,48 @@ hb_blob_create (const char        *data,
 		void              *user_data,
 		hb_destroy_func_t  destroy)
 {
+  if (!length)
+    return hb_blob_get_empty ();
+
+  hb_blob_t *blob = hb_blob_create_or_fail (data, length, mode,
+					    user_data, destroy);
+  return likely (blob) ? blob : hb_blob_get_empty ();
+}
+
+/**
+ * hb_blob_create_or_fail: (skip)
+ * @data: Pointer to blob data.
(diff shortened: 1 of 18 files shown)


behdad:
This, specifically, doesn't change any semantics of the existing API. In particular, `hb_blob_create()` with length=0 still just returns the empty blob; we depend on this for non-existing font tables to not allocate a blob wastefully. The new `hb_blob_create_or_fail()` however doesn't do that and gives flexibility back to the caller to decide.

src/hb-blob.cc
```diff
@@ -71,15 +71,49 @@ hb_blob_create (const char        *data,
 		hb_memory_mode_t   mode,
 		void              *user_data,
 		hb_destroy_func_t  destroy)
+{
+  if (!length)
+    return hb_blob_get_empty ();
```
khaledhosny, line 76: Do we need to call `destroy()` here like in the (old) code below?

khaledhosny (approved):
This seems to have been merged into master (by accident?).

src/hb-blob.cc
behdad: Riiiiight.

behdad:
> This seems to have been merged into master (by accident?).

Ouch.

khaledhosny:
Since this is merged already, should we close this PR?

PR #3028, opened 2021-06-16
[subset] don't alloc zero bytes.

garretrieger:
It will be leaked later since hb_blob_create() won't set up the blob to cleanup since it has length zero.


behdad:
LGTM. Maybe add a comment.

PR #3030, opened 2021-06-17
[subset] support option "--notdef-outline"

qxliu76:


src/hb-subset-input.hh
```diff
@@ -45,12 +45,13 @@ struct hb_subset_input_t
   hb_set_t *drop_tables;
   hb_set_t *layout_features;
 
-  bool drop_hints;
-  bool desubroutinize;
-  bool retain_gids;
-  bool name_legacy;
-  bool overlaps_flag;
-  bool retain_all_layout_features;
+  int32_t drop_hints;
```
garretrieger, line 48: Can you add a comment explaining why these are int32's intsead of booleans

garretrieger (approved):

src/hb-ot-glyf-table.hh
```diff
@@ -209,7 +209,10 @@ struct glyf
 		if (!plan->old_gid_for_new_gid (new_gid, &subset_glyph.old_gid))
 		  return subset_glyph;
 
-		subset_glyph.source_glyph = glyf.glyph_for_gid (subset_glyph.old_gid, true);
+                if (new_gid == 0 && !plan->notdef_outline)
+                  subset_glyph.source_glyph = Glyph ();
+                else
```
behdad, line 214: Tabs instead of spaces please.

src/hb-ot-var-gvar-table.hh
```diff
@@ -461,6 +462,7 @@ struct gvar
       else
 	((HBUINT16 *) subset_offsets)[gid] = glyph_offset / 2;
 
+      if (gid == 0 && !c->plan->notdef_outline) continue;
```
behdad, line 465: Can't you adjust the start of the loop instead?

behdad:
LGTM w comments. Thanks.

behdad:
Both loops in gvar can just adjust theirs start, instead of `gid = 0` can become `gid = 1` if notdef outline isn't desired. That's what I meant.

qxliu76:
> Both loops in gvar can just adjust theirs start, instead of `gid = 0` can become `gid = 1` if notdef outline isn't desired. That's what I meant.

Done. :)

behdad:
Lovely. Thanks.

PR #3034, opened 2021-06-20
[directwrite] Support feature ranges

khaledhosny:
Code copied from hb-uniscribe.cc and adapted to use DirectWrite types.
May be there is a better way to de-duplicate the code, though I'm not exactly sure how or if it is worth it.

### DIFF
--- a/src/hb-directwrite.cc
+++ b/src/hb-directwrite.cc
@@ -510,6 +510,44 @@ protected:
  * shaper
  */
 
+struct active_feature_t {
+  DWRITE_FONT_FEATURE fea;
+  unsigned int order;
+
+  HB_INTERNAL static int cmp (const void *pa, const void *pb) {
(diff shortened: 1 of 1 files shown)


behdad:
I'm okay merging this first, then moving common code to a shared template.

We might do something similar for `hb-ot-shape` as well, to avoid doing a buffer loop per user feature.

khaledhosny:
(I don’t know much about C++ templates, except that some people hated them decades ago).

behdad:
> (I don’t know much about C++ templates, except that some people hated them decades ago).

It's simple and nice. Give it a try if you feel like.

C++ has a 1000 problems. Simple use of templates is fine.

behdad:
> (I don’t know much about C++ templates, except that some people hated them decades ago).

It's simple and nice. Give it a try if you feel like.

C++ has a 1000 problems. Simple use of templates is fine.

Issue #501, opened 2017-06-23
Verse numbers of "KFGQPC Uthmanic Script HAFS" are not shaped right


ebraminio:
Filed from [this mailing list topic](https://lists.freedesktop.org/archives/harfbuzz/2017-June/005953.html) as a possible bug, steps to repro:

1. Download the font from http://fonts.qurancomplex.gov.sa/?page_id=42
2. ./hb-view UthmanicHafs1\ Ver09.otf --direction=ltr ١٢

Actual:
21

Expected:
12

(this works if we issue `./hb-view UthmanicHafs1\ Ver09.otf --shapers=coretext --direction=ltr ١٢` on macOS and perhaps uniscribe, and with other fonts)

khaledhosny:
I think this is because HarfBuzz processes the characters in the “native” direction of the script, so Arabic-Indic digits are processed from right-to-left so gives the wrong ligature in the font. I think that is one of the major remaining differences between HarfBuzz and Uniscribe. I stumbled upon this issue years ago and after Behdad’s explanation I managed to come up with a workaround in my font, but I wasn’t using ligatures so it was doable unlike here.

ebraminio:
Thanks for you insightful comment @khaledhosny,

@behdad, I remember you wanted to put your comment here also, perhaps it would be useful for the future developments of the project and other collaborators.

behdad:
Here's full details of this issue:

In Unicode, one can force-change direction of text using RLO and LRO characters.  HarfBuzz wants to support this correctly.  Ie. if you have <RLO,e,f,f,i> you should get "i,f,f,e" rendered.  If you try with Uniscribe, you get "ffi,e", ie. it wrongly forms the 'ffi' ligature. Ie. it processes lookups in whatever input text order was there.  At least it used to work this way two years ago.

Now, Uniscribe does handle this correctly in their Arabic and I suppose Hebrew shapers.  In HarfBuzz I want to support it correctly all the time.  The way I do this is to have a per-script "native" horizontal direction, which reflects the natural order of the script.  When `hb_shape()` and requested shaping direction is reverse of the natural horizontal direction of the script, we flip the buffer first such that text becomes in natural order before we apply the lookups, because the font designer designed lookups for the natural order.

[...]

PR #3035, opened 2021-06-23
 [ot-shape] Numeric runs native direction is LTR

khaledhosny:
See inline comments. Slightly modified version of the code from Jonathan Kew on the linked issue.

Fixes https://github.com/harfbuzz/harfbuzz/issues/501.


behdad:
In the Arabic shaper there's this comment:
```
static void
apply_stch (const hb_ot_shape_plan_t *plan HB_UNUSED,
            hb_buffer_t              *buffer,
            hb_font_t                *font)
{
  if (likely (!(buffer->scratch_flags & HB_BUFFER_SCRATCH_FLAG_ARABIC_HAS_STCH)))
    return;

  /* The Arabic shaper currently always processes in RTL mode, so we should
   * stretch / position the stretched pieces to the left / preceding glyphs. */
```

Does this change break that assumption? If yes, we should either review & fix, or at least check and bail if not RTL.

behdad:
> Does this change break that assumption?

At the least the good news is that there are no Syriac numberals.

jfkthame:
No, but the ASCII digits have Script=Common, so presumably they would be merged with a surrounding Syriac context for the purposes of script-run analysis. (Not sure if this is a potential problem or not...?)

khaledhosny:
It is potentially an issue if Syriac abbreviation mark gets applied to the digits, but this does not happens since HarfBuzz does not allow `HB_UNICODE_GENERAL_CATEGORY_DECIMAL_NUMBER` to be covered by the abbreviation mark.

jfkthame:
Just FTR, I confirmed that this does break an existing testcase we have in Gecko (once we revert the hack from https://bugzilla.mozilla.org/show_bug.cgi?id=1716029, which otherwise "hides" this), where we use bidi-override to force directionality on an example with one of the format controls (see layout/reftests//bugs/1599841-1.html and the corresponding -ref file in the Mozilla tree). However, I'm OK with that; with this change, more things will work in the "natural" way authors would expect, so we'll adapt our (rather artificial/arbitrary) testcase as needed. In particular, it'll be possible to use the footnote marker, year sign, etc in Noto Nastaliq Urdu *without* needing to apply any direction overrides.

behdad:
> In particular, it'll be possible to use the footnote marker, year sign, etc in Noto Nastaliq Urdu _without_ needing to apply any direction overrides.

Right. I think that's a worthwhile change. Thanks for the comment.

PR #3039, opened 2021-06-29
[subset] bail if region_map is in error.

garretrieger:
Fixes fuzzer case https://oss-fuzz.com/testcase-detail/5499162606043136.


behdad:
LGTM. Why are the sanitizers failing?

garretrieger:
> LGTM. Why are the sanitizers failing?

Still trying to figure that out. Haven't been able to replicate locally.

behdad:
Please merge if you are satisfied with this.

PR #3042, opened 2021-06-30
Don’t build Cairo subproject with FontConfig on win32 cross-build

khaledhosny:
We don’t need it.


khaledhosny:
This should also fix recent CI breakage because Cairo wants to build FontConfig from master branch which is apparently no more (renamed to main). Building from random git commits is a totally fine idea 🤦🏽.

PR #3045, opened 2021-07-07
[subset] fuzzer fix

qxliu76:
Don't serialize UnsizedArray when the offset to the array is 0.
In this fuzzer testcase, offset to UnsizedArray is 0 but num of elements in the array is not 0


behdad:
Can't we use one of the templated `serialize_copy` methods here?

qxliu76:
> Can't we use one of the templated `serialize_copy` methods here?

Thanks! Changed to use this method in 2 places.

Issue #3048, opened 2021-07-08
Time for a release


behdad:
There are a couple of important shaping fixes in (emoji; Arabic digits), and misc improvements, specially to the subsetter. So, let's get a release out.

muratmat:
It works, thanks!

PR #3049, opened 2021-07-08
Update release notes to reflect actual current procedure

alerque:
In anticipation of #3048 I checked the release notes file in the repo and realized it is a bit out of date. At least two points are completely obsoleted by CI changes. Possibly more but I did the ones I know about.

While I was at it I made the list a Markdown checklist suitable for copy/pasting into an issue or PR to track release progress if that is ever desired. I find it handy but that's really up to the fellow doing the work more than anything else.


behdad:
Thank you!

PR #3057, opened 2021-07-12
[ci] Add macOS GitHub workflow

khaledhosny:
To collect codecov coverage on macOS as well (hopefully it will run the macOS-specific tests).

### DIFF
--- a/.github/workflows/linux-ci.yml
+++ b/.github/workflows/linux-ci.yml
@@ -47,7 +47,7 @@ jobs:
     #     fail-on-alert: true
 
     - name: cov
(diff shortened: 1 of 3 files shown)


behdad:
Thanks.

khaledhosny:
`gcovr` is failing to find some ragel files, and I have no idea why and why only on macOS but not on Linux.

khaledhosny:
> `gcovr` is failing to find some ragel files, and I have no idea why and why only on macOS but not on Linux.

Generating XML coverage data seems to work, since that is what we are uploading anyway, we now generate only it and build seems to pass.

Issue #3058, opened 2021-07-13
[subset] Missing TrueType table after subsetting NotoSerifMyanmar-Regular.otf


joycode:
[NotoSerifMyanmar-Regular.otf.zip](https://github.com/harfbuzz/harfbuzz/files/6805845/NotoSerifMyanmar-Regular.otf.zip)

OS: Debian Buster
Compiler: g++ (Debian 8.3.0-6) 8.3.0
harfbuzz version: 63e15eac4f443fa53565d1e4fb9611cdd7814f28 (HEAD -> main, tag: 2.8.2, origin/main, origin/HEAD)
test command: ./hb-subset NotoSerifMyanmar-Regular.otf "႒" -o subset-myanmar.ttf

"႒" is a glyph from NotoSerifMyanmar-Regular.otf, it's Glyph Name is "two_shn", Unicode Value is U+1092

Output file "subset-myanmar.ttf" seems a invalid TrueType file, when open with FontForge, it warns "subset-myanmar.ttf is not in a known format", and fails to open it. I also tried to open the file with FreeType library, it fails in tt_face_lookup_table(), which returns NULL.

behdad:
Works fine for me.

behdad:
Ah, you are right. FreeType doesn't like to open it.

behdad:
@garretrieger any idea why `CFF ` table is dropped?

behdad:
@joycode where did you get that font file? We've tracked down the issue and have a fix, but in my understanding the font is wrong for having a `FontName` op in `TopDict`. So I'm wondering what compiler built that font. Thanks.

cc @blueshade7 @anthrotype

joycode:
@behdad Thanks for your fixing, It's a font from AOSP: https://android.googlesource.com/platform/external/noto-fonts/+/refs/heads/master/other/NotoSerifMyanmar-Regular.otf

anthrotype:
I'm also curious to know how those .otf files were built.
ufo2ft (used by fontmake) never adds a `FontName` op to a CFF TopDict (see it's `setupTable_CFF` method [here](https://github.com/googlefonts/ufo2ft/blob/86c3357aa535d3eec73f8440435838c112eae639/Lib/ufo2ft/outlineCompiler.py#L1213-L1372)).
If I am not mistaken, the `FontName` op is specific to CID-keyed fonts and we always output name-keyed CFF fonts.
[...]

PR #3059, opened 2021-07-14
[subset] Add missing FontName op code in top dict parsing.

garretrieger:
Fix for #3058


behdad:
Thanks. cc @blueshade7

PR #3060, opened 2021-07-15
[subset] de-duplicate the logic that finds unicodes corresponding to …

garretrieger:
…requested glyphs.

Move the logic into subset planning and then re-use the results in cmap and OS2 subsetting. Removes depedency on cmap from os2. Which is needed for #681


behdad:
Thank you Garret. Looks great.

PR #3063, opened 2021-07-18
Arabic PUA shaping

khaledhosny:
TODO:
- [x] Support 3-letter ligatures
- [x] Add tests for the “traditional” fontpage
- [ ] Final ligatures in the “simplified” fontpage
- [x] mark ligatures
- [x] Hook in `hb-ft` as well.


src/hb-ot-cmap-table.hh
```diff
@@ -1497,7 +1499,19 @@ struct cmap
...
+	  break;
+	case OS2::font_page_t::FONT_PAGE_SIMP_ARABIC:
+	  this->get_glyph_funcZ = get_glyph_from_symbol_arabic_pua1<CmapSubtable>;
+	  break;
+	case OS2::font_page_t::FONT_PAGE_TRAD_ARABIC:
+	  this->get_glyph_funcZ = get_glyph_from_symbol_arabic_pua2<CmapSubtable>;
+	  break;
+	}
+      }
       else
```
khaledhosny, line 1515: So this currently works only for `ot` font functions (I was surprised again to discover that `ft `are still the default for the command line tools, it confused me more than once). I don’t fell easy about this since it is unlikely that clients implementing their own font functions will try to support this, we also need to support it for the `ft` functions, not sure what is the best way, though.

src/hb-ot-os2-table.hh
```diff
@@ -223,9 +223,11 @@ struct OS2
     *max_cp = hb_min (0xFFFFu, codepoints->get_max ());
   }
 
-  /* https://github.com/Microsoft/Font-Validator/blob/520aaae/OTFontFileVal/val_OS2.cs#L644-L681 */
+  /* https://github.com/Microsoft/Font-Validator/blob/520aaae/OTFontFileVal/val_OS2.cs#L644-L681
+   * https://docs.microsoft.com/en-us/typography/legacy/legacy_arabic_fonts */
   enum font_page_t
   {
+    FONT_PAGE_NONE		= 0,
```
khaledhosny, line 230: I haven’t used this, but if felt odd not to define 0 when it is the fallback value returned by `get_font_page()` below.

src/hb-ot-shape-complex-arabic-fallback.hh
```diff
@@ -254,6 +256,137 @@ struct arabic_fallback_plan_t
...
+				     first_glyphs_indirection,
+				     ligature_per_first_glyph_count_list,
+				     ligature_list,
+				     component_count_list,
+				     component_list,
+				     ligature_table_pua1,
+				     ARRAY_LENGTH_CONST (ligature_table_pua1),
+				     buf,
+				     sizeof (buf),
+				     0);
```
khaledhosny, line 331: This 0 is for lookup flags, there are a few mark ligatures so can’t use `IgnoreMarks`. I should have probably split the marks into there own looks, but by testing shows that Uniscribe breaks the ligatures when there are marks, so the bad behaviour is actually closer to what it does.

src/hb-ot-shape-complex-arabic-pua.hh
```diff
@@ -0,0 +1,385 @@
...
+#define SHAPING_TABLE_PUA2_FIRST	0x0622u
+#define SHAPING_TABLE_PUA2_LAST		0x064Au
+#define SHAPING_TABLE_PUA2_LEN		(SHAPING_TABLE_PUA2_LAST - SHAPING_TABLE_PUA2_FIRST + 1)
+
+static const struct ligature_set_t ligature_table_pua2[] =
+{
+  { 0xF2BBu, {
+//  { 0xF2BCu, 0xF2C9u, 0xF201u }, /* LAM WITH LAM WITH HEH ISOLATED FORM */
+//  { 0xF2C0u, 0xF25Au, 0xF210u }, /* LAM WITH MEEM WITH HAH INITIAL FORM */
+//  { 0xF2C0u, 0xF256u, 0xF211u }, /* LAM WITH MEEM WITH JEEM INITIAL FORM */
```
khaledhosny, line 290: I still need to figure out if the current code can easily support 3-letter ligatures.

src/hb-ot-shape-complex-arabic-pua.hh
```diff
@@ -0,0 +1,385 @@
...
+#define SHAPING_TABLE_PUA1_LEN		(SHAPING_TABLE_PUA1_LAST - SHAPING_TABLE_PUA1_FIRST + 1)
+
+static const struct ligature_set_t ligature_table_pua1[] =
+{
+  { 0xF17Bu, {
+    { 0xF144u, 0xF1C0u }, /* LAM WITH ALEF WITH MADDA ABOVE ISOLATED FORM */
+    { 0xF144u, 0xF1BEu }, /* LAM WITH ALEF WITH HAMZA ABOVE ISOLATED FORM */
+    { 0xF148u, 0xF1C2u }, /* LAM WITH ALEF WITH HAMZA BELOW ISOLATED FORM */
+    { 0xF142u, 0xF1BCu }, /* LAM WITH ALEF ISOLATED FORM */
+  }},
```
khaledhosny, line 148: There should be final version of these lam-alef ligatures here (the fonts has the glyphs), but the fonts use the same glyph for initial and medial lam, so I don’t think this can work in a ligature lookup. I don’t know if I can form the ligature first then add them to the fina feature (would fina be applied to a ligature made of medial + final letters)?

behdad:
One thing I don't understand is why do you need separate shaping tables for these? If you correctly decode cmap, then our existing Arabic fallback will synthesize tables for them. We just need to add more ligatures if you wish.

khaledhosny:
Two reasons, I didn't want to change the behavior of the existing fallback by enabling more ligatures, and there are at least two ligatures without corresponding Unicode presentation forms code points.

behdad:
> Two reasons, I didn't want to change the behavior of the existing fallback by enabling more ligatures, and there are at least two ligatures without corresponding Unicode presentation forms code points.

Given how legacy these fonts are, I'm happy ignoring both of those issues.

khaledhosny:
This is basically working, PUA is remapped to Unicode Arabic code points in cmap table and the existing Unicode fallback shaping is used. The extra ligatures have been enabled in `gen-arabic-table.py` so they are now will used in any font that goes through the Arabic fallback shaping and has these code points.

The three remaining issues that I’m not sure how to fix:
* Some of the new ligatures have three components, but the ligature building code and data structure allows only for two components.
* Some ligatures are mark glyphs, these currently do nothing because the ligature lookup sets the `IgnoreMarks` lookup flag, so they need a different lookup.
* 3 ligatures have no equivalent Unicode Presentation Forms code points (one is لله ligature and I’d love to have it), I made `gen-arabic-table.py` generate entries for them using the PUA code points but this does not feel right (they are effectively unused though, because one is a mark ligature and the other two have three components).

khaledhosny:
And making this work with other font functions.

behdad:
> This is basically working, PUA is remapped to Unicode Arabic code points in cmap table and the existing Unicode fallback shaping is used. The extra ligatures have been enabled in `gen-arabic-table.py` so they are now will used in any font that goes through the Arabic fallback shaping and has these code points.

Sounds good.

> The three remaining issues that I’m not sure how to fix:
> 
> * Some of the new ligatures have three components, but the ligature building code and data structure allows only for two components.

Easy enough to add code for that. Can you collect those in a different array? I can help with code to generate lookup (which isn't that different really).

> * Some ligatures are mark glyphs, these currently do nothing because the ligature lookup sets the `IgnoreMarks` lookup flag, so they need a different lookup.

Same; can you modify the generator to collect those in different array?

> * 3 ligatures have no equivalent Unicode Presentation Forms code points (one is لله ligature and I’d love to have it), I made `gen-arabic-table.py` generate entries for them using the PUA code points but this does not feel right (they are effectively unused though, because one is a mark ligature and the other two have three components).

This I suggest you let these go. Remember, how legacy and uncommon these fonts are...

src/hb-ot-cmap-table.hh
```diff
@@ -1383,6 +1385,21 @@ struct EncodingRecord
...
+_hb_remap_symbol_pua (hb_codepoint_t codepoint)
+{
+  if (codepoint <= 0x00FFu)
+  {
+    /* For symbol-encoded OpenType fonts, we duplicate the
+     * U+F000..F0FF range at U+0000..U+00FF.  That's what
+     * Windows seems to do, and that's hinted about at:
+     * https://docs.microsoft.com/en-us/typography/opentype/spec/recom
+     * under "Non-Standard (Symbol) Fonts". */
+    return 0xF000u + codepoint;
```
behdad, line 1398: While at it, can you change this (and comments etc) to use `os2.usFirstCharIndex`?

behdad:
> And making this work with other font functions.

Is not too complicated to make it work with `hb-ft`. Custom funcs I don't think we need to deal with.

src/hb-ot-shape-complex-arabic-pua.hh
```diff
@@ -0,0 +1,289 @@
+
```
behdad, line 1: So, you hand-generated this file? Let's reuse packtab to make it much more compact. There's so many zeros in there.

I don't understand why it maps from more than just the first page though.

behdad:
I rebased this onto main.

behdad:
This mostly looks good to me. I think hooking it up to hb-ft, and adding option to disable it and hooking that up to HB_NO_LEGACY is all needed. The other todo items in your list are rather improvements than blockers IMO.

The tables are a bit large, but given the option above I don't think it's a major issue.

behdad:
> The tables are a bit large, but given the option above I don't think it's a major issue.

Or maybe we should only enable this on Windows.

khaledhosny:
> > The tables are a bit large, but given the option above I don't think it's a major issue.
> 
> Or maybe we should only enable this on Windows.

These fonts are rather popular, and people often complain they don’t work outside Windows or in apps with their own layout engine, so I’d rather we enable this by default.

behdad:
> These fonts are rather popular, and people often complain they don’t work outside Windows or in apps with their own layout engine, so I’d rather we enable this by default.

Sounds good.

src/hb-ft.cc
```diff
@@ -310,14 +312,25 @@ hb_ft_get_nominal_glyph (hb_font_t *font HB_UNUSED,
...
-    if (unlikely (ft_font->symbol) && unicode <= 0x00FFu)
+    if (unlikely (ft_font->symbol))
     {
-      /* For symbol-encoded OpenType fonts, we duplicate the
-       * U+F000..F0FF range at U+0000..U+00FF.  That's what
-       * Windows seems to do, and that's hinted about at:
-       * https://docs.microsoft.com/en-us/typography/opentype/spec/recom
-       * under "Non-Standard (Symbol) Fonts". */
-      g = FT_Get_Char_Index (ft_font->ft_face, 0xF000u + unicode);
+      switch ((unsigned) font->face->table.OS2->get_font_page ()) {
```
khaledhosny, line 317: Should this instead use FreeType API to access OS/2 table?

src/hb-ot-shape-complex-arabic-pua.hh
khaledhosny: I can generate a text file in a format similar to Unicode encoding mapping files (e.g. https://unicode.org/Public/MAPPINGS/VENDORS/MICSFT/WINDOWS/CP1256.TXT), would that be OK?

khaledhosny: In f61add0ec8ca114377d042a5d37258d76f4ac64b I committed two files using the Unicode mapping format (which mean they map from PUA to Unicode, the reverse of the C arrays).

src/hb-ft.cc
behdad: Umm. I don't see any advantage of one over the other.

behdad:
Can you also remind me again, why this can't be done in the `preprocess_text` shaper hook?

behdad:
> Can you also remind me again, why this can't be done in the `preprocess_text` shaper hook?

Maybe even the symbol stuff we can preprocess it, such that works with custom font funcs as well.

src/hb-ot-shape-complex-arabic-pua.hh
behdad: I did a packTab impl of it to generate table from those two files. Takes only 2kb instead of 3.5 current implementation takes. Is slower but we don't care.

behdad:
> Can you also remind me again, why this can't be done in the `preprocess_text` shaper hook?

I think we should do this. See the Thai shaper for an example of that.

behdad:
And here's a proof-of-concept of using packTab for these tables:
https://gist.github.com/behdad/6cdb9a9a2f029b91ed8a1a6a253a5723

khaledhosny:
> > The three remaining issues that I’m not sure how to fix:
> > 
> > * Some of the new ligatures have three components, but the ligature building code and data structure allows only for two components.
> 
> Easy enough to add code for that. Can you collect those in a different array? I can help with code to generate lookup (which isn't that different really).
> 
> > * Some ligatures are mark glyphs, these currently do nothing because the ligature lookup sets the `IgnoreMarks` lookup flag, so they need a different lookup.
> 
> Same; can you modify the generator to collect those in different array?

Done in 1c156b67b. The struct definitions are no ugly, but I tried to make the change as minimal as possible.

khaledhosny:
> Done in [1c156b6](https://github.com/harfbuzz/harfbuzz/commit/1c156b67b9781290bfaa9f4c60acfc08747c896d). The struct definitions are no ugly, but I tried to make the change as minimal as possible.

This breaks the tests since some of the now split ligatures are used in the tests.

behdad:
> > > The three remaining issues that I’m not sure how to fix:
> > > 
> > > * Some of the new ligatures have three components, but the ligature building code and data structure allows only for two components.
> > 
> > 
> > Easy enough to add code for that. Can you collect those in a different array? I can help with code to generate lookup (which isn't that different really).
> > > * Some ligatures are mark glyphs, these currently do nothing because the ligature lookup sets the `IgnoreMarks` lookup flag, so they need a different lookup.
> > 
> > 
> > Same; can you modify the generator to collect those in different array?
> 
> Done in [1c156b6](https://github.com/harfbuzz/harfbuzz/commit/1c156b67b9781290bfaa9f4c60acfc08747c896d). The struct definitions are no ugly, but I tried to make the change as minimal as possible.

This doesn't yet generate those ligatures, right?

khaledhosny:
> Can you also remind me again, why this can't be done in the `preprocess_text` shaper hook?

I don’t remember the details, but I can try it again and see what happens.

behdad:
> > Can you also remind me again, why this can't be done in the `preprocess_text` shaper hook?
> 
> I don’t remember the details, but I can try it again and see what happens.

Thanks. I think it will be much nicer.

I think you need to make the cmap/ft symbol handling to only do the previous mapping if OS/2 thing is 0.

khaledhosny:
> This doesn't yet generate those ligatures, right?

Yes. The code that generates the ligatures is too clever for me to modify.

behdad:
Can you email me a test font please?

behdad:
ligature_mark_set_t and ligature_set_t can be merged...

behdad:
> ligature_mark_set_t and ligature_set_t can be merged...

Nvmind.

khaledhosny:
> Can you email me a test font please?

There is two test fonts in this PR,  test/shape/data/in-house/fonts/SimpArabicTest.ttf and test/shape/data/in-house/fonts/TradArabicTest.ttf

khaledhosny:
> This breaks the tests since some of the now split ligatures are used in the tests.

Actually it doesn’t, I should expand the tests to cover these later.

khaledhosny:
> > Can you also remind me again, why this can't be done in the `preprocess_text` shaper hook?
> 
> I don’t remember the details, but I can try it again and see what happens.

We need to check if the selected cmap subtable is a symbol one, that is why symbol mapping is implemented in the cmap table/get_nominal_glyph callback.

behdad:
> We need to check if the selected cmap subtable is a symbol one, that is why symbol mapping is implemented in the cmap table/get_nominal_glyph callback.

So, the OS/2 check is not enough?

behdad:
> We need to check if the selected cmap subtable is a symbol one, that is why symbol mapping is implemented in the cmap table/get_nominal_glyph callback

> So, the OS/2 check is not enough?

It might be, right? And we can also sniff the font for a few of the PUA code-points to be sure.

khaledhosny:
> > We need to check if the selected cmap subtable is a symbol one, that is why symbol mapping is implemented in the cmap table/get_nominal_glyph callback.
> 
> So, the OS/2 check is not enough?

The [MS documentation](https://docs.microsoft.com/en-us/typography/legacy/legacy_arabic_fonts) says these fonts use symbol encoding, so we check for this and the font page stuff, but may be skipping the symbol check is good enough for Arabic?

behdad:
I implemented mark ligatures. Will work on 3-letter ligatures now, though that one is trickier.

behdad:
> I implemented mark ligatures. Will work on 3-letter ligatures now, though that one is trickier.

I tested and works. Can you add a test to the test suite please?

behdad:
> ligature_mark_set_t and ligature_set_t can be merged...

Actually yes they can.

behdad:
> > I implemented mark ligatures. Will work on 3-letter ligatures now, though that one is trickier.
> 
> I tested and works. Can you add a test to the test suite please?

I'm fixing the existing test.

behdad:
3-letter ligatures work now. I test manually. Please add a test if you can. Thanks!

behdad:
> 3-letter ligatures work now. I test manually. Please add a test if you can. Thanks!

Again, just the existing tests need updating. I'm doing that.

khaledhosny:
> 3-letter ligatures work now. I test manually. Please add a test if you can. Thanks!

Something is not quite right, though:
```
$ ./util/hb-shape ../test/shape/data/in-house/fonts/TradArabicTest.ttf "لمحة لمجة الله"
```
resut

The first ligature is fine, the second ligature has its last two components repeated unligated, the last ligature has its last component repeated.

behdad:
Yeah I noticed. Debugging.

behdad:
> > 3-letter ligatures work now. I test manually. Please add a test if you can. Thanks!
> 
> Something is not quite right, though:
> 
> ```
> $ ./util/hb-shape ../test/shape/data/in-house/fonts/TradArabicTest.ttf "لمحة لمجة الله"
> ```
> 
> resut
> 
> The first ligature is fine, the second ligature has its last two components repeated unligated, the last ligature has its last component repeated.

Which is really weird given that we use the same GSUB lookup facilities to apply...

behdad:
Ah, found an error at least.

behdad:
Okay I think I got it right this time.

khaledhosny:
There remains the issue with the ”simplified” font page sharing initial/medial and final/isolated glyphs, but having different ligatures. These fonts are rarer, so I think we can just ignore this. This PR is functionally complete AFAIC, yay!

khaledhosny:
> There remains the issue with the ”simplified” font page sharing initial/medial and final/isolated glyphs, but having different ligatures. These fonts are rarer, so I think we can just ignore this. This PR is functionally complete AFAIC, yay!

```
$ ./util/hb-view ../test/shape/data/in-house/fonts/SimpArabicTest.ttf "لا بلا"
```
result

(the lam-alef ligature is always talking the final form)

behdad:
> > There remains the issue with the ”simplified” font page sharing initial/medial and final/isolated glyphs, but having different ligatures. These fonts are rarer, so I think we can just ignore this. This PR is functionally complete AFAIC, yay!
> 
> ```
> $ ./util/hb-view ../test/shape/data/in-house/fonts/SimpArabicTest.ttf "لا بلا"
> ```
> 
> result
> 
> (the lam-alef ligature is always talking the final form)

I see.

khaledhosny:
> 

I _think_ this can be fixed by creating the isolated ligatures in, say `ccmp` feature, that in `fina` map the isolated ligature to the final one. Not sure if it is worth the trouble, though.

behdad:
> I _think_ this can be fixed by creating the isolated ligatures in, say `ccmp` feature, that in `fina` map the isolated ligature to the final one. Not sure if it is worth the trouble, though.

Agree.

behdad:
I tried doing it in preprocess_text. It's not possible with our current architecture. So that's down.

Only remains using packTab. But that can be done later.

src/hb-ot-shape-complex-arabic-pua.hh
behdad: Here's the packtab gist:
https://gist.github.com/behdad/6cdb9a9a2f029b91ed8a1a6a253a5723

behdad:
Just hooking it up to NO_LEGACY, which I'm doing now.

khaledhosny:
> Only remains using packTab. But that can be done later.

Working on a PR.

src/hb-ot-shape-complex-arabic-pua.hh
khaledhosny: Does not look like we can use `hb-ot-shaper-arabic-table.hh` here as it expects `JOINING_TYPE_*` to be defined before including it, which is why I created a new file.

behdad: Yeah a new file is fine. Then make a new tiny generator. Which makes sense since these files are static and need no Unicode data file.

khaledhosny: https://github.com/harfbuzz/harfbuzz/pull/3668

PR #3067, opened 2021-07-22
[subset] support option --glyph-names

qxliu76:


src/hb-ot-post-table.hh
```diff
@@ -55,6 +55,91 @@ struct postV2Tail
...
+
+  bool subset (hb_subset_context_t *c) const
+  {
+    TRACE_SUBSET (this);
+
+    const hb_map_t &reverse_glyph_map = *c->plan->reverse_glyph_map;
+    hb_map_t old_new_index_map;
+    unsigned i = 0;
+
+    auto index_iter =
```
garretrieger, line 114: Re-using index_iter will be problematic since it modifies the local variable 'i'. Each time the iterator is run i will be incremented further.

I suggest:
- Moving this iteration code into a for loop that builds the map and computes the max old index.
- Use old_new_index_map.iter() to provide the iterator for serialize.

src/hb-ot-post-table.hh
```diff
@@ -55,6 +55,91 @@ struct postV2Tail
...
+                  const hb_map_t& index_map) const
+  {
+    TRACE_SERIALIZE (this);
+    auto *out = c->start_embed (this);
+    if (unlikely (!c->check_success (out))) return_trace (false);
+    if (!out->glyphNameIndex.serialize (c, + it
+                                           | hb_map (hb_second)))
+        return_trace (false);
+
+    const UnsizedArrayOf<const HBUINT8> &names = StructAfter<const UnsizedArrayOf<const HBUINT8>> (glyphNameIndex);
```
garretrieger, line 86: No sanitization is done on the strings array currently. So it's not safe to access the strings memory directly. You could instead use the accelerator_t::find_glyph_name() which does bounds checking for you.

garretrieger (approved):

garretrieger (approved):

Issue #2743, opened 2020-11-13
[hb-view] Doesn't understand linked axes


ctrlcctrlv:
It is a recognized font creation trick that if you make multiple axes with the same tag in the fvar and STAT tables, and mark all but one hidden, that changing the value of one changes the value of all. _Cf._ https://twitter.com/jmsoleb/status/1325735951730155521

`hb-view`, however, acts as if only one axis was set to the chosen value.

```bash
for i in `seq -f "%04g" 0 10 1000`; do hb-view --variations="0001=$i" NLI_unmangled.ttf -o $i.png --text-file=../gif-text.txt; echo $i; done
```

Therefore I am for now producing "mangled" and "unmangled" versions of my FontForge NLI/HOI example. See https://github.com/ctrlcctrlv/FontForge-Higher-Order-Interpolation

behdad:
Problem code is in `hb_font_set_variations()`. Isn't too hard to fix.

behdad:
Duplicate of https://github.com/harfbuzz/harfbuzz/issues/1673

Issue #1673, opened 2019-04-18
[variations] Support multiple axes with same tag


behdad:
Currently doesn't work.

behdad:
Needs no API change though.

jfkthame:
Do renderers actually support fonts having multiple axes with the same tag? I guess I've always assumed axis tags needed to be distinct, though now that you mention it, I don't know that the spec says anything about this.

behdad:
> Do renderers actually support fonts having multiple axes with the same tag? I guess I've always assumed axis tags needed to be distinct, though now that you mention it, I don't know that the spec says anything about this.

The geniuses that discovered this trick claim that it works in Firefox and Chrome.  Which might mean just CoreText, or CoreText and FreeType.  cc @drott @lemzwerg FYI.  We should check FreeType.

But yeah, this would allow higher-order polynomial interpolations.  I'm going to recommend that spec allow this as long as at most one axis per tag is non-HIDDEN.

lemzwerg:
Is there a sample font for this?

behdad:
> Is there a sample font for this?

I'll send you privately.

justvanrossum:
Fwiw, this is a nice write-up, and links to a test font: https://bungeman.github.io/hoi.html

lemzwerg:
Indeed, thanks!  And the demo font seems to exhibit a problem with FreeType: If you test with `ftmulti` (from current git), there is a jump of the point glyph at 'Weight' value 450...

justvanrossum:
I also see the jump with FreeType 2.10.2.

lemzwerg:
It' an artifact of the font, I guess – if you manually enter value 450 into Samsa, you see exactly the same jump.  @bungeman, can you have a look, please?

bungeman:
Yes, the example font purposely has the 'bad' way to encode multiple curves into a single contour on the 'negative' side of the axes. In fact that's why it has that strange range on the negative side, so that the most viewers would actually hit the discontinuity (it's really easy to make a font which doesn't show the discontinuity in most viewers since the viewer may not be able to exactly hit the point).

That discontinuity is the motivation for the 'second attempt' which is demonstrated on the 'positive' side of the axes which doesn't have any discontinuity between curves on the contour. So yes, this jump is expected. Maybe should be more explicit about that in the 'example' text.

lemzwerg:
Thanks!

PR #3068, opened 2021-07-22
[variations] Support multiple axes with same tag, aka HOI

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/1673
Fixes https://github.com/harfbuzz/harfbuzz/issues/2743

Test from https://github.com/ctrlcctrlv/FontForge-Higher-Order-Interpolation


behdad:
cc @bungeman @lemzwerg @justvanrossum @ctrlcctrlv

lemzwerg:
I don't know enough about HarfBuzz to make any qualified review, but I'm all for it, so thanks for working on this!

Is there anything necessary on the FreeType side?

src/hb-font.cc
```diff
@@ -2067,13 +2067,16 @@ hb_font_set_variations (hb_font_t            *font,
   const OT::fvar &fvar = *font->face->table.fvar;
   for (unsigned int i = 0; i < variations_length; i++)
   {
-    unsigned axis_index;
-    if (fvar.find_axis_index (variations[i].tag, &axis_index) &&
-	axis_index < coords_length)
+    unsigned axis_index_start, axis_index_end;
+    if (fvar.find_axis_index (variations[i].tag, &axis_index_start, &axis_index_end))
```
drott, line 2071: If there are multiple axes with the same tag, is it ensured that they are ordered and identical tags are lined up consecutively? In the font, or does HB sort them somewhere?

drott: PS: If yes, then LGTM.

justvanrossum: I don't think the OT spec demands the axes to be ordered by tag, and `find_axis_index ` seems to iterate using the order in the font: I suspect that `find_axis_index()` is not correct.

behdad: Ah, I assumed they are sorted by tag. Let me check.

behdad: Okay yes, not sorted. I'll rework the code.

behdad: I pushed a reworked version that doesn't make sorted assumption. PTAL.

behdad:
> Is there anything necessary on the FreeType side?

Not that I know of. From what I've heard, FreeType works fine. Thanks.

drott (approved):
LGTM with this change - although I have limited knowledge of this code.

src/hb-font.cc
justvanrossum: Looks good to me.

PR #3071, opened 2021-07-26
[sebset] Use [s|g]et_flag for retain_all_layout_features

khaledhosny:
Looks like this was missed when adding `hb_subset_input_[s|g]et_flag()`.


src/hb-subset.h
```diff
@@ -61,6 +61,8 @@ typedef struct hb_subset_input_t hb_subset_input_t;
  * retained in the final subset. Defaults to false.
  * @HB_SUBSET_FLAG_NO_PRUNE_UNICODE_RANGES: If set then the unicode ranges in
  * OS/2 will not be recalculated.
+ * @HB_SUBSET_FLAG_RETAIN_ALL_FEATURES: If set all layout features will be
+ * retained.
```
garretrieger, line 65: Can you also add:

If unset then the layout_features_set will be used to determine the set of features to be retained.

garretrieger (approved):
Thanks!

src/hb-subset.h
khaledhosny: Done.

PR #3073, opened 2021-07-26
Do not synthesize the Ideographic space

### LINKED ISSUE #2889: Ideographic spaces are synthesized using non-CJK fonts in Blink
When author specifies non-CJK fonts before CJK fonts in the `font-family` property in Blink, following 3 symptoms are observed. 

Originally found the symptom 1, then found this is probably related with #2885, and found other symptoms. These symptoms are not seen in Gecko. I'm not sure whether this is a Blink issue or not. Any advice is very much appreciated in advance. /cc @drott

**1. GPOS that involves U+3000 are not applied**

The font in [this test page](https://jsbin.com/mosomej/edit?html,output) has a GPOS feature for U+3000 using format 2 and format 8.

**2. Width of U+3000 in the fonts are ignored**

Ideographic Spaces are meant to match Kana and Ideographic characters in the fonts. It is usually 1em, but not always so. An example when Kana and Ideographic do not have the same advance can be observed in [a test page](https://jsbin.com/quqibi/edit?html,output), or by using `hb-shape`:
```sh
% hb-shape msgothic.ttc --face-index 0 -u 3000
[gid2057=0+256]
% hb-shape msgothic.ttc --face-index 1 -u 3000
[gid20621=0+170]
% hb-shape msgothic.ttc --face-index 2 -u 3000
[gid18103=0+170]
```
It is also seen in some news paper fonts, such as [Motoya Newspaper Mincho/Gothic](http://www.motoyafont.jp/font-list/news.html) or [Font Garage AFS series](https://font.designers-garage.jp/products/detail/7792).

**3. Pure CJK runs are split at Ideographic spaces in Blink**

In the following example, Blink decides to split the run and use Arial to render U+3000. No visual differences in this case, but slightly less efficient as it is rendered as two separate runs.
```html
<p style="font-family: Arial, 'MS Gothic'">&#x3031;&#x3000;</p>
```

KrasnayaPloshchad:
This should fix #2889


khaledhosny:
Closing pet the discussion in the linked issue.

PR #3075, opened 2021-07-26
[ci] Bin FontTools version due to recent COLRv1 changes

khaledhosny:
Should be reverted once HarfBuzz updated to match FontTools.


khaledhosny:
This is to unbreak the CI builds.

behdad:
Thanks. Though the root cause really is https://github.com/harfbuzz/harfbuzz/issues/3077

PR #3076, opened 2021-07-26
[subset] add a note to hb-subset on what is and isn't supported

garretrieger:


behdad:
LGTM; suggest removing mention of COLRv0/COLRV1 and just say COLR.

behdad:
Also add that notably EBLC/EBDT are not supported. sbix, I don't know.

garretrieger:
> Also add that notably EBLC/EBDT are not supported. sbix, I don't know.

Ah right forgot we just did CBLC/CBDT, I also note we don't support SVG. For sbix we do have support: https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-ot-color-sbix-table.hh#L387

src/hb-subset.cc
```diff
@@ -64,6 +64,15 @@
  * that is no longer needed. A subset input describes the desired subset. The input is
  * provided along with a font to the subsetting operation. Output is a new font file
  * containing only the data specified in the input.
+ *
+ * Currently most outline and bitmap tables are supported: glyf, CFF, CFF2, sbix,
+ * COLR, and CBDT/CBLC. This also includes fonts with variable outlines via open
+ * type variations. Notably EBDT/EBLC and SVG are not supported. Layout subsetting is supported
```
behdad, line 70: => `OpenType Variations`.

behdad:
LGTM. Thanks.

Issue #3070, opened 2021-07-24
Amalgam builds are broken when both DirectWrite and Uniscribe are enabled


khaledhosny:
Both files define structs with similar names since #3034. Need to move that code to a shared place and templatize it after all.

behdad:
You can rename them to be unique for now.

I would do it but I can't build the directwrite code.

[15mins later]

I checked the code. I believe the DWrite structs were designed to be byte-compatible with the uniscribe ones. So I believe we can use non-templatized shared code. Just need to define two structs and have static asserts that their sizes match the uniscribe / dwrite one, and use those.

- Unify `OPENTYPE_FEATURE_RECORD` and `DWRITE_FONT_FEATURE` into `hb_ms_feature_t`.
- Unify `TEXTRANGE_PROPERTIES` and `DWRITE_TYPOGRAPHIC_FEATURES` into `hb_ms_features_t`.

PR #3080, opened 2021-07-27
[uniscribe/directwrite] Move feature setup to shared file

khaledhosny:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3070

### DIFF
--- a/src/hb-ms-feature-ranges.hh
+++ b/src/hb-ms-feature-ranges.hh
@@ -1,0 +1,229 @@
+/*
+ * Copyright © 2011,2012,2013  Google, Inc.
+ * Copyright © 2021  Khaled Hosny
+ *
+ *  This is part of HarfBuzz, a text shaping library.
+ *
+ * Permission is hereby granted, without written agreement and without
+ * license or royalty fees, to use, copy, modify, and distribute this
+ * software and its documentation for any purpose, provided that the
+ * above copyright notice and the following two paragraphs appear in
+ * all copies of this software.
+ *
+ * IN NO EVENT SHALL THE COPYRIGHT HOLDER BE LIABLE TO ANY PARTY FOR
+ * DIRECT, INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES
+ * ARISING OUT OF THE USE OF THIS SOFTWARE AND ITS DOCUMENTATION, EVEN
+ * IF THE COPYRIGHT HOLDER HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ *
+ * THE COPYRIGHT HOLDER SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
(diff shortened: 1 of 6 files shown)


src/hb-ms-feature-ranges.hh
```diff
@@ -0,0 +1,229 @@
...
+  }
+};
+
+struct range_record_t {
+  hb_ms_features_t features;
+  unsigned int index_first; /* == start */
+  unsigned int index_last;  /* == end - 1 */
+};
+
+HB_INTERNAL static bool
```
behdad, line 80: Since this is `static`, you probably don't need `HB_INTERNAL`. Maybe prefix with `_` just to be sure.

src/hb-ms-feature-ranges.hh
```diff
@@ -0,0 +1,229 @@
...
+  hb_tag_t tag;
+  uint32_t value;
+} hb_ms_feature_t;
+
+typedef struct hb_ms_features_t {
+  hb_ms_feature_t *features;
+  uint32_t         num_features;
+} hb_ms_features_t;
+
+struct active_feature_t {
```
behdad, line 43: Maybe prefix these `ms` or `hb_ms` as well. `active_feature_t` etc are overly general.

src/hb-ms-feature-ranges.hh
```diff
@@ -0,0 +1,229 @@
...
+ * Google Author(s): Behdad Esfahbod
+ */
+
+#ifndef HB_MS_FEATURE_RANGES_HH
+#define HB_MS_FEATURE_RANGES_HH
+
+#include "hb.hh"
+
+typedef struct hb_ms_feature_t {
+  hb_tag_t tag;
```
behdad, line 34: This tag is in different byteorder than ours. So I suggest using `uint32_t` or a separate typedefed one, and maybe rename `tag` to `tag_le`.

src/hb-ms-feature-ranges.hh
```diff
@@ -0,0 +1,229 @@
...
+  {
+    range_record_t *range = &range_records[i];
+    range->features.features = (hb_ms_feature_t *) feature_records + reinterpret_cast<uintptr_t> (range->features.features);
+  }
+
+  return !!num_features;
+
+}
+
+HB_INTERNAL static void
```
behdad, line 186: Same here.

Alternatively, we can move the code into a separate `hb-ms.cc` and mark these only `HB_INTERNAL`, such that they are not duplicated in the binary if both backends are enabled.

behdad (commented):
Thanks for doing this!

src/hb-ms-feature-ranges.hh
khaledhosny: Moved to .cc file.

src/hb-ms-feature-ranges.hh
khaledhosny: Done.

src/hb-ms-feature-ranges.hh
khaledhosny: Done.

src/hb-ms-feature-ranges.hh
khaledhosny: Moved to `hb-ms-feature-ranges.cc` file.

behdad (approved):
Great. Thanks!

PR #3083, opened 2021-07-28
Update AOTS tests and fix some failures

khaledhosny:


util/main-font-text.hh
```diff
@@ -47,7 +47,7 @@ locale_to_utf8 (char *s)
   return t;
 }
 
-template <typename consumer_t, int default_font_size, int subpixel_bits>
+template <typename consumer_t, int default_font_size, int subpixel_bits, bool oneline = false>
```
behdad, line 50: How about adding a `char eol = '\n'` instead?

util/options.cc
```diff
@@ -258,12 +258,21 @@ parse_features (const char *name G_GNUC_UNUSED,
...
+  size_t l = strlen (s);
   char *p;
 
   shape_opts->num_features = 0;
   g_free (shape_opts->features);
   shape_opts->features = nullptr;
 
+  /* if the string is quoted strip the quotes */
+  if ((s[0] == '\"' && s[l - 1] == '\"') ||
+      (s[0] == '\'' && s[l - 1] == '\''))
```
behdad, line 270: Slightly nicer would be to check that `s[0] == s[l - 1]` first.

behdad:
Thanks Khaled. LGTM.

util/main-font-text.hh
khaledhosny: Done.

util/options.cc
khaledhosny: Done.

util/options.hh
```diff
@@ -548,7 +548,7 @@ struct text_options_t : option_group_t
 		   "Only one of text and text-file can be set");
   }
 
-  const char *get_line (unsigned int *len);
+  const char *get_line (unsigned int *len, char eol);
```
behdad, line 551: Maybe add the default `\n` here?

behdad: Did it myself. Thanks agian.

khaledhosny: I considered that, but found it is called only in one place. But thinking again, I agree, it is more elegant/self documenting.

Issue #3017, opened 2021-06-09
Not all emoji sequences recommended for general interchange (RGI) cluster


rsheeter:
In testing emoji rgi clustering it seems polar bears and regional flags don't cluster. https://github.com/rsheeter/hb-emoji-clusters/blob/main/try_shape-stdout.txt has a list. For context, I tested all the sequences in https://unicode.org/Public/emoji/14.0/emoji-test.txt, which enumerates emoji sequences.

IIUC per Mark Davis all emoji should be grapheme clusters. I thought that would mean HB would cluster them but seemingly not. I see discussion on #2265 where a fix to make it so was discussed.

In the event emoji were to span multiple files it would help Chrome itemization if emoji rgi consistently formed clusters. If it's desirable that they don't I'd appreciate if someone could ELI5.

behdad:
Regional-indicators are Complicated(TM) as seen in #2265.

Polar bear is the weirdest thing I've seen in Unicode:  🐻 U+1F43B, ZWJ U+200D,❄ U+2744.
️
This is our code implementing the grapheme logic: https://github.com/harfbuzz/harfbuzz/blob/bd5502f09f09b3f464754395880f9db30a5fcf60/src/hb-ot-shape.cc#L466-L517

For emoji, we append any ZWJ,Extended_Pictograph sequence to the previous cluster. U+2744 SNOWFLAKE *is* in that list. So I expect that we handle this sequence correctly. Let me check.

behdad:
Maybe we should add an equivalent of your code to the test suite.

behdad:
Oops... Bad bug in emoji table generator...

behdad:
@rsheeter Can you please rerun your script against master and attach conclusion?

rsheeter:
It would appear you have rescued the polar bear! https://github.com/rsheeter/hb-emoji-clusters/commit/85bfd1fe44434b1d69305082e335a48401a63506

behdad:
Re the regional-indicator pairs: https://github.com/harfbuzz/harfbuzz/issues/2265#issuecomment-858191957

rsheeter:
#3018 probably fixes this;.

behdad:
> #3018 probably fixes this;.

Interesting to see if it actually does.

rsheeter:
`0 / 4702 failures`, looks like it does :)

drott:
Big thanks for doing this list, @rsheeter! And glad we found an actual issue here.

[...]

PR #3087, opened 2021-07-28
[test] Add generated tests for emoji clusters

khaledhosny:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3017

Uses AdobeBlank2.ttf from:
  https://github.com/adobe-fonts/adobe-blank-2

instead of a dummy empty font so that everything maps to GID 1 and control code points are kept instead of being dropped because there is
not space glyph (otherwise we’d need to identify control code points somehow when generating the expectations).


behdad:
Thank you!!!

behdad:
How much slower will it be if we don't chunk?

khaledhosny:
> How much slower will it be if we don't chunk?

About 4 times slower on my machine, but still below 1 second over all.

khaledhosny:
(apparently batch mode makes things pretty fast, the standalone test that from @rsheeter repo took about 30 seconds.

behdad:
> (apparently batch mode makes things pretty fast, the standalone test that from @rsheeter repo took about 30 seconds.)

Yes, I was wondering since we are using batch mode.

If it's still under 1s, I"m inclined to leave it unchunked.

And I'd be happy to speed batch mode up instead.

khaledhosny:
To give exact numbers, it goes from 0.15s to 1.15s (it was 0.8s when I checked earlier), but yeah not worth it and easier to read failure messages should be worth it.

behdad:
> To give exact numbers, it goes from 0.15s to 1.15s (it was 0.8s when I checked earlier), but yeah not worth it and easier to read failure messages should be worth it.

I see. But yeah, do it the proper way and I'll take a stab at optimizing batch mode. I think currently it still reloads the face every time. Just not doing that would help.

PR #3088, opened 2021-07-29
Region indices

behdad:
Trying to fix failing sanitizer test. But this makes bunch of our tests fail. Not sure if tests are wrong or the code. Pushing here to get help from @garretrieger


behdad:
If I comment out the `del_range` line, then tests pass. Which makes me really suspicious of the test.

behdad:
Nah; something wrong with my patch.

behdad:
I'm positive this one should do it..

PR #3094, opened 2021-07-29
Construct user_data in hb_object

bungeman:
hb_object's user_data is created lazily. The previous implementation of hb_object_set_user_data created space for the user_data but did not actually construct it. This means that hb_user_data_array_t's lock was not constructed. If hb_mutex_t is backed by an implementation which requires that it be constructed (not just zero initialized) then errors will occur when taking the lock when setting the user data.

Change hb_object_set_user_data to construct the user_data in the created space and hb_object_fini to call the destructor.

### DIFF
--- a/src/hb-object.hh
+++ b/src/hb-object.hh
@@ -37,6 +37,7 @@
 #include "hb-mutex.hh"
 #include "hb-vector.hh"
 
+#include <new>
 
 /*
  * Lockable set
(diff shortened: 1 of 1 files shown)


bungeman:
This was discovered while attempting to use std::mutex to back hb_mutex_t on macOS is debug configuration. See [proposed Skia roll](https://skia-review.googlesource.com/c/skia/+/433737/3). Reproduced this issue locally and this change alleviates the issue.

behdad:
Thanks.

Garret is also working on contributing `std::mutex` to HB.

I'm also working on moving away from our custom init/fini and more into default C++ idioms. Let's see how this `<new>` works out.

Issue #3089, opened 2021-07-29
Speed-up subset test suite


behdad:
I suggest adding `--batch` to `hb-subset`, like `hb-shape` has. That should immensely help.

behdad:
Currently `make check -j10` in `test/shaping` takes 14s for me, whereas in `test/subset` takes 100s.

behdad:
Tentatively assigning to @garretrieger; but @khaledhosny feel free to try since you're on a roll! Thank you.

khaledhosny:
The subset tests also use FontTools to compare the fonts (saving to ttx XML first then diffing it), we can probably speed it by saving the expected output in XML format instead of generating it each time. This has the downside of having to update the expected results each time FontTools changes its output.

behdad:
If anything, making the test-runner a bit more verbose would be great. Right now each test takes multiple seconds with no further details.

khaledhosny:
> It has expected binary from fonttools saved? Then converts that and hb-subset output to ttx to compare?

Yes.

behdad:
Thanks Khaled. That definitely helped.

Next: the subset tests are a matrix of cartesian product of fonts x profiles x text. For example:

  https://github.com/harfbuzz/harfbuzz/blob/5d283aa0bd65739f78b384590e070e8cfb531d74/test/subset/data/tests/cff-japanese.tests#L1-L19

That's one font, 8 profiles, 5 texts, for a total of 40 subset tests.  I think we should sparse this up. We don't need to test every combination.

khaledhosny:
> Next: the subset tests are a matrix of cartesian product of fonts x profiles x text. For example:

I’ll leave this for someone who is actually familiar with how subset and its test suite works.

khaledhosny:
These two tests are currently the biggest offenders, all other tests are under 2 seconds and most under 1 second:
```
54/55 harfbuzz:subset+slow / cff-japanese                    OK               9.01s
55/55 harfbuzz:subset+slow / basics                          OK              20.35s
```

Breaking them up so they can be parallelized probably helps, even if we keep testing all the profiles.

khaledhosny:
Splitting basics test into 3 tests one for each font does not seem to help either, most of the time is spent working on the 3rd font (NanumMyeongjo-Regular) anyway.

khaledhosny:
Correction, most of the time is spent on Comfortaa-Regular-new.ttf.

khaledhosny:
[...]

PR #3095, opened 2021-07-29
[test] Speed-up subset tests by saving TTX dump

khaledhosny:
Speed-up subset tests by saving TTX dump of expected output instead of generating it each time the tests are run.

Cuts down meson `test --suite=subset` on my system from:
```
real	0m38.977s
user	1m12.024s
sys	0m10.547s
```

to:
```
real	0m22.291s
user	0m44.548s
sys	0m9.221s
```

Part of https://github.com/harfbuzz/harfbuzz/issues/3089


behdad:
Thanks. What's the effect on the release tarball?

behdad:
I just did a tar.xz of all the ttx and that's under 1MB. I'm happy with this. Thanks again!

Issue #3086, opened 2021-07-28
[subset] Consider adding a bulk flag setting method to the subset api


garretrieger:
Now that we're using enum's to for flag setting we should add a variant of the flag setting method that allows multiple flags to be set in a single call. I'm thinking something like this:

hb_subset_input_set_flags(HB_SUBSET_FLAG_HINTING, HB_SUBSET_FLAG_RETAIN_GIDS)

or to clear flags:

hb_subset_input_clear_flags(HB_SUBSET_FLAG_HINTING, HB_SUBSET_FLAG_RETAIN_GIDS)

using variadic arguments to accept multiple flags.

behdad:
I don't like variadic API..

I'm leaning towards:
```
hb_subset_input_set_flags (*input, hb_subset_flag_t mask, hb_subset_flag_t value);
``

but that makes me realize that these flags are not bit-flags. Maybe we *should* make them bitflags, rename `flag_t` to `flags_t` to match other similar API.

PR #3096, opened 2021-07-29
[subset] change input and plan flags to be bit sets.

garretrieger:
By switching both to bit sets we can easily support a batch set/clear method and as an added bonus makes adding new flags trivial.

One other notable change is this updates the subset fuzzer to take the entire flags bit set from the fuzzer input, which will now allow the fuzzer to toggle any combination of flags.

For #3086

### DIFF
--- a/src/hb-subset-input.cc
+++ b/src/hb-subset-input.cc
@@ -55,7 +55,7 @@ hb_subset_input_create_or_fail (void)
   input->drop_tables = hb_set_create ();
   input->no_subset_tables = hb_set_create ();
 
-  input->flags = HB_SUBSET_FLAGS_NONE;
+  input->flags = HB_SUBSET_FLAGS_DEFAULT;
 
   hb_tag_t default_drop_tables[] = {
     // Layout disabled by default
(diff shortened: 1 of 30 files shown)


src/hb-subset-input.cc
```diff
@@ -362,98 +354,39 @@ hb_subset_input_no_subset_tables_set (hb_subset_input_t *input)
...
 /**
- * hb_subset_input_get_flag:
+ * hb_subset_input_get_flags:
  * @input: a #hb_subset_input_t object.
- * @flag: which flag to check.
- *
- * Get the value of the specified flag.
  *
- * Return value: value of the specified flag.
+ * Return value: the subsetting flags bit array.
```
behdad, line 360: Maybe "bit-field"? Array can be confusing.

src/hb-subset-input.cc
```diff
@@ -362,98 +354,39 @@ hb_subset_input_no_subset_tables_set (hb_subset_input_t *input)
...
-      input->retain_all_layout_features = value;
-      break;
-    default:
-      // Do nothing.
-      break;
-  }
+  // Set desired flags.
+  input->flags = (hb_subset_flags_t) (input->flags | (mask & value));
+  // Clear desired flags.
+  input->flags = (hb_subset_flags_t) (input->flags & ~(mask & ~value));
```
behdad, line 389: Just:
```
  intput->flags = (input->flags & ~mask) | (value & mask)
```

src/hb-subset.h
```diff
@@ -40,51 +40,56 @@ HB_BEGIN_DECLS
...
+  HB_SUBSET_FLAGS_NO_HINTING = 1,
+  HB_SUBSET_FLAGS_RETAIN_GIDS = 1 << 1,
+  HB_SUBSET_FLAGS_DESUBROUTINIZE = 1 << 2,
+  HB_SUBSET_FLAGS_NAME_LEGACY = 1 << 3,
+  HB_SUBSET_FLAGS_SET_OVERLAPS_FLAG = 1 << 4,
+  HB_SUBSET_FLAGS_PASSTHROUGH_UNRECOGNIZED = 1 << 5,
+  HB_SUBSET_FLAGS_NOTDEF_OUTLINE = 1 << 6,
+  HB_SUBSET_FLAGS_GLYPH_NAMES = 1 << 7,
+  HB_SUBSET_FLAGS_NO_PRUNE_UNICODE_RANGES = 1 << 8,
+  HB_SUBSET_FLAGS_RETAIN_ALL_FEATURES = 1 << 9,
```
behdad: Please tabulate the `=` signs using enough tabs.

In other places we use hex eg `0x00000080u`. Maybe match that?

src/hb-subset.h
```diff
@@ -40,51 +40,56 @@ HB_BEGIN_DECLS
...
+  HB_SUBSET_FLAGS_ALL = -1,
```
behdad: For `ALL` I had in mind the `|` of all the currently defined ones only.

util/options.hh
```diff
@@ -690,7 +690,31 @@ struct subset_options_t : option_group_t
...
+    {
+      hb_subset_input_set_flags (input,
+                                 (hb_subset_flags_t) (1 << i),
+                                 flags[i] ? HB_SUBSET_FLAGS_ALL : HB_SUBSET_FLAGS_NONE);
+    }
+    return input;
+  }
+
+  hb_bool_t flags[sizeof(int) * 8] = {0};
+
```
behdad, line 717: So, instead of this `bool_for` and `get_input`, I suggest you use `G_OPTION_ARG_CALLBACK`.

src/hb-subset.h
garretrieger: ALL has been removed.

PR #3098, opened 2021-07-29
[WIP] Buffer shift forward

behdad:
From https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=36236
https://oss-fuzz.com/testcase-detail/5061207689134080


behdad:
Humm. Failure only happens on certain memory alloc fail; our sanitizers can't replicate, nor can I :(.

PR #3099, opened 2021-07-30
[mutex] for the other use of hb_mutex_t ensure the constructor/destructor is called.

garretrieger:


behdad:
I feel like this (and previous change) are not the best way to approach this. We should contain the change in `hb-mutex.hh`. So far there was an assumption that mutexes can be initialized by value. We just need to change to call init on them, instead of initializing random objects using `new`.

behdad:
See https://github.com/harfbuzz/harfbuzz/pull/3100. No need.

bungeman:
The issue is that hb_object types have a hb_mutex_t in them, and if hb_mutex_t isn't POD (which it isn't if hb_mutex_impl_t is std:mutex) then just casting the hb_calloc'ed memory to a pointer and then calling through it is already undefined behavior, so calling `new` in the `init` doesn't help.

behdad:
> The issue is that hb_object types have a hb_mutex_t in them, and if hb_mutex_t isn't POD (which it isn't if hb_mutex_impl_t is std:mutex) then just casting the hb_calloc'ed memory to a pointer and then calling through it is already undefined behavior, so calling `new` in the `init` doesn't help.

I see how legally that analysis is correct. We're transitioning into using new instead of our custom allocations. Would using a reinterpret_cast instead of C-style cast make it any different? Guess not.

It will take a while to clean this all up to be legally safe. I'm fine landing init/fini change with this caveat and fix in time. Do sanitizers call that out?

PR #3100, opened 2021-07-30
Add optional hb_mutex_t std::mutex implementation

bungeman:


src/hb-mutex.hh
```diff
@@ -48,6 +48,17 @@
 /* Defined externally, i.e. in config.h; must have typedef'ed hb_mutex_impl_t as well. */
 
 
+#elif !defined(HB_NO_MT) && defined(HB_MUTEX_IMPL_STD_MUTEX)
```
behdad, line 51: No need for extra symbol. We should just use this as backup.  When we added `std::atomic` to `hb-atomic.hh`, I removed all other implementations except for the compiler builtins.

The only reason to keep pthreads and win32 impls above this is that `std::mutex` probably requires linking to libstdc++; right?  We're moving towards that anyway; and trying to use this instead of erring sounds better to me.

behdad:
Please see https://github.com/harfbuzz/harfbuzz/pull/3099#issuecomment-890082224

I feel like we should just make sure we call `init/fini` on the mutex always, and move the `new` construction to this new implementation.

bungeman:
The way the user opts into this should probably be reviewed (I'm just guessing at a name for build time flag here).

Also `HB_MUTEX_IMPL_INIT` is only used by `HB_MUTEX_INIT`, but `HB_MUTEX_INIT` is not currently used. Based on how everything is defined I'm making an educated guess at the correct `HB_MUTEX_IMPL_INIT` for std::mutex. However, since it isn't used it isn't tested.

behdad:
Oh. We already call init/fini as needed. YOu just need to call placement new here. And revert the previous change you sent.

behdad:
> Also `HB_MUTEX_IMPL_INIT` is only used by `HB_MUTEX_INIT`, but `HB_MUTEX_INIT` is not currently used. Based on how everything is defined I'm making an educated guess at the correct `HB_MUTEX_IMPL_INIT` for std::mutex. However, since it isn't used it isn't tested.

Lemme remove that completely.

behdad:
Okay. Just call placement new/delete in your init/fini and no extra check for enabling it, and this should be good to go. Thank you!

bungeman:
As I was commenting on the other change, calling placement new in `init` is too late. It may work in a lot of current compilers, but making the call though the cast pointer from hb_calloc does not put an object at that location (and won't ever even with proposed changes) when the mutex isn't POD (which std::mutex isn't).

behdad:
How about this instead: Instead of embedding `std::mutex` in `hb_mutex_t`, do a union that just allocates the space for it. Then you construct/destruct in init/fini; that's legal, no?

behdad:
> How about this instead: Instead of embedding `std::mutex` in `hb_mutex_t`, do a union that just allocates the space for it. Then you construct/destruct in init/fini; that's legal, no?

Even if union doesn't work, we can use `sizeof/alignas` etc to allocate bytes for it and construct into it.

bungeman:
I suppose a union could make things work out, but I still feel uncomfortable with all of the shenanigans of calling object methods when an object has not yet been constructed, though I suppose HarfBuzz currently heavily relies on that more or less working with at least POD types.

And also, yes, in general it looks like every time a mutex is created memory is being allocated anyway, so might make sense to defer and encapsulate that way as well.

behdad:
> And also, yes, in general it looks like every time a mutex is created memory is being allocated anyway, so might make sense to defer and encapsulate that way as well.

I know in glib we ended up always allocating mutexes; that is, use a pointer. I don't think we need to do that here, since the user-data array is already allocated on-demand only.

The weird/dangerous C-style casting and init/fini was necessary before C++11. With C++11 we can have `constexpr` consturctors, which still allow simple objects to act like POD. So I'm moving in that direction.

src/hb-mutex.hh
```diff
@@ -81,20 +81,26 @@ typedef int hb_mutex_impl_t;
 
 #else
 
-#error "Could not find any system to define mutex macros."
-#error "Check hb-mutex.hh for possible resolutions."
+#include <mutex>
+#include <new>
+typedef std::mutex              hb_mutex_impl_t;
+#define hb_mutex_impl_init(M)   HB_STMT_START { new (M) hb_mutex_impl_t; } HB_STMT_END
```
bungeman, line 87: This is also a bit squirrel-y, technically one needs to access the object from the pointer returned from `new`, the original pointer passed in to `new` is not blessed with "objectness". Technically, a compiler is allowed to optimize this `new` away, since the resulting object can never be accessed. I can't store it in `m`  since that's actually an array, not a pointer.

src/hb-mutex.hh
```diff
@@ -81,20 +81,26 @@ typedef int hb_mutex_impl_t;
 
 #else
 
-#error "Could not find any system to define mutex macros."
-#error "Check hb-mutex.hh for possible resolutions."
+#include <mutex>
+#include <new>
+typedef std::mutex              hb_mutex_impl_t;
```
bungeman, line 86: After ensuring that this is actually being used everywhere and building, it appears that most std::mutex implementations are going to want to link against the c++ standard library (mostly for exception types apparently), so may need to have a feature test for that?

bungeman:
So a few issues.

* Ignoring the return value of the placement new can either be fixed in various ways or just lived with for the time being until such time as HarfBuzz can new/delete things instead of init/fini them.
* The use of std::mutex is going to require linking against a c++ standard library. Will need a feature macro of some sort to determine if linking against a c++ standard library so we know if std::mutex can be used.
* This will need to be tested somehow. In the current proposed change it's a fallback and a build flag can be used to select it, but it is not currently selected by any test bot.

Not sure if I can work on all of that (I had started this hoping it would be a ~10 line change), so feel free to run with this if you have the time.

behdad:
> Not sure if I can work on all of that (I had started this hoping it would be a ~10 line change), so feel free to run with this if you have the time.

Yeah don't worry about it. I'll take it from here. I probably will merge as is first, then figure out what to  do.

src/hb-mutex.hh
```diff
@@ -81,20 +81,26 @@ typedef int hb_mutex_impl_t;
...
 {
-  hb_mutex_impl_t m;
+  // Create space for, but do not initialize m
+  alignas(hb_mutex_impl_t) char m[sizeof(hb_mutex_impl_t)];
 
-  void init   () { hb_mutex_impl_init   (&m); }
-  void lock   () { hb_mutex_impl_lock   (&m); }
-  void unlock () { hb_mutex_impl_unlock (&m); }
-  void fini ()   { hb_mutex_impl_finish (&m); }
+  void init   () { hb_mutex_impl_init   ((hb_mutex_impl_t*)&m); }
```
behdad, line 100: Aren't the `&` redundant here?

src/hb-mutex.hh
```diff
@@ -81,20 +81,26 @@ typedef int hb_mutex_impl_t;
...
 #else
 
-#error "Could not find any system to define mutex macros."
-#error "Check hb-mutex.hh for possible resolutions."
+#include <mutex>
+#include <new>
+typedef std::mutex              hb_mutex_impl_t;
+#define hb_mutex_impl_init(M)   HB_STMT_START { new (M) hb_mutex_impl_t; } HB_STMT_END
+#define hb_mutex_impl_lock(M)   (*(M)).lock ()
+#define hb_mutex_impl_unlock(M) (*(M)).unlock ()
```
behdad, line 89: Any reason you didn't use `->` here?

src/hb-mutex.hh
behdad: I'm fairly sure that's not the case for placement new. Trying to find some reference to that.

behdad: Here, for example: https://en.cppreference.com/w/cpp/memory/allocator_traits/construct
We can call that if you prefer.

bungeman: It seems even stranger to use `std::allocator_traits<Alloc>::construct` since the underlying memory was not allocated with any allocator, and `construct` is spec'ed to call placement new anyway. The reality is that I don't think any compiler can really take advantage of not using the return value of placement new since it would break so much existing code (even if it is technically undefined behavior).

So mostly, I think this is probably ok, it just makes me uncomfortable. Essentially every vector like container has this issue.

src/hb-mutex.hh
bungeman: I suppose arrays do like to decay to pointers. Note that this code is casting `char[N]*`. The proposal to remove the `&` (not taking the the address) would cast `char*` instead. It makes no difference here since we're just casing the pointer either way, but it would make a difference if doing pointer arithmetic.

behdad:
Superseded by https://github.com/harfbuzz/harfbuzz/pull/3225.

PR #3102, opened 2021-08-01
[test] Add batch mode to hb-subset and use it

khaledhosny:
time meson test --suite=subset down from:
```
real	0m22.822s
user	0m44.561s
sys	0m9.255s
```
to:
```
real	0m19.418s
user	0m38.171s
sys	0m3.587s
```

Does not seem to help much, but it is something.

Part of https://github.com/harfbuzz/harfbuzz/issues/3089

### DIFF
--- a/test/subset/run-tests.py
+++ b/test/subset/run-tests.py
@@ -22,6 +22,12 @@ except ImportError:
 
 ots_sanitize = shutil.which ("ots-sanitize")
 
+def subset_cmd (command):
+	global process
+	process.stdin.write ((';'.join (command) + '\n').encode ("utf-8"))
+	process.stdin.flush ()
+	return process.stdout.readline().decode ("utf-8").strip ()
+
 def cmd (command):
 	p = subprocess.Popen (
 		command, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
@@ -51,10 +57,10 @@ def run_test (test, should_check_ots):
 		    "--drop-tables-=sbix"]
 	cli_args.extend (test.get_profile_flags ())
 	print (' '.join (cli_args))
-	_, return_code = cmd (cli_args)
+	ret = subset_cmd (cli_args)
 
-	if return_code:
-		return fail_test (test, cli_args, "%s returned %d" % (' '.join (cli_args), return_code))
+	if ret != "success":
+		return fail_test (test, cli_args, "%s failed" % ' '.join (cli_args))
 
 	expected_file = os.path.join (test_suite.get_output_directory (), test.get_font_ttx_name ())
 	with open(expected_file, encoding="utf-8") as expected_ttx:
@@ -114,6 +120,11 @@ if not len (args):
 	sys.exit ("No tests supplied.")
 
 has_ots = has_ots()
+
+process = subprocess.Popen ([hb_subset, '--batch'],
+                            stdin=subprocess.PIPE,
+                            stdout=subprocess.PIPE,
+                            stderr=sys.stdout)
 
 fails = 0
 for path in args:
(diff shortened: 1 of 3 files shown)


util/hb-subset.cc
```diff
@@ -130,6 +130,36 @@ struct subset_consumer_t
...
+      unsigned start_offset = 0;
+      while ((e = strchr (p + start_offset, ';')) && argc < (int) ARRAY_LENGTH (args))
+      {
+	*e++ = '\0';
+	while (*e == ':')
+	  e++;
+	args[argc++] = p = e;
+      }
+      ret |= driver.main (argc, args);
+      fprintf (stdout, "Done!\n");
```
behdad, line 155: Can we make it write out the return status? Either 0/1, or success/fail...

test/subset/run-tests.py
```diff
@@ -51,10 +57,7 @@ def run_test (test, should_check_ots):
 		    "--drop-tables-=sbix"]
 	cli_args.extend (test.get_profile_flags ())
 	print (' '.join (cli_args))
-	_, return_code = cmd (cli_args)
-
-	if return_code:
-		return fail_test (test, cli_args, "%s returned %d" % (' '.join (cli_args), return_code))
```
behdad, line 57: You are not tracking individual failures anymore. That's not very friendly. Let's read back the status.

util/hb-subset.cc
```diff
@@ -130,6 +130,36 @@ struct subset_consumer_t
...
 {
+  if (argc == 2 && !strcmp (argv[1], "--batch"))
+  {
+    unsigned int ret = 0;
+    char buf[4092];
+    while (fgets (buf, sizeof (buf), stdin))
+    {
+      size_t l = strlen (buf);
+      if (l && buf[l - 1] == '\n') buf[l - 1] = '\0';
+      main_font_text_t<subset_consumer_t, 10, 0> driver;
```
behdad, line 141: Pass `eol` here as well?

test/subset/run-tests.py
khaledhosny: If hb-subset failed, it will be caught when comparing the output with the expected on. The shaping tests don't check the return value either.

behdad: Well. Shape results are communicated via stdout. Subset results are communicated via a file. What if the file exists from previous runs, but subset fails now without removing the old file?

khaledhosny: The test runner is using temporary folders, so this shouldn't happen. But lets be more through and communicate the failure right away.

behdad: Thanks.

util/hb-subset.cc
khaledhosny: Done.

test/subset/run-tests.py
khaledhosny: Done.

util/hb-subset.cc
khaledhosny: Done.

behdad:
Any reason you used `;` instead of `:` to separate args in `hb-subset` batch mode? `hb-shape` uses ':'...

Now I wish we had just used space. Or.. I suppose there was a problem with that but I don't remember.

behdad:
> Now I wish we had just used space. Or.. I suppose there was a problem with that but I don't remember.

ba810ce83d392daeaa0c69af5549d020436a5ac7

behdad:
Now I'm more inclined on `;` instead of `:` as well, if it means we can remove that `C:...` hack as well. cc @ebraminio

behdad:
I tried to changing `;`. Makes one test fail in `aat-trak.tests` because that one uses `;` as a feature range separator, since `:` cannot be used in test files...

I'm leaning towards using `\0` as separator in batch mode...

behdad:
Maybe I switch it back to space, and add a `--batch0` that uses `\0`, like `-print0` does in `find` and have our test runners use that.

PR #3105, opened 2021-08-02
[test] Add API test for lookup_collect_glyphs()

khaledhosny:
We don’t seem to have any tests exercising this API at all, and it is a good check of GSUB/GPOS code.


behdad:
Thanks Khaled!

khaledhosny:
I believe all lookup types and subtable formats are now covered.

behdad:
> I believe all lookup types and subtable formats are now covered.

Wow. Thanks!

PR #3110, opened 2021-08-04
[test] Make subset tests fast

khaledhosny:


test/subset/run-tests.py
```diff
@@ -72,6 +71,10 @@ def run_test (test, should_check_ots):
 	if expected_hash == actual_hash:
 		return 0
 
+	if TTFont is None:
+		print ("fonttools is not present, skipping test.")
+		sys.exit (77)
```
khaledhosny, line 76: The still skips the tests, but at this point we know there is a chance that the test is actually failing, so should we fail the test if FontTools is missing?

behdad: Yeah I think the test should be considered failed if hashes not equal. FontTools really is then just used to report the failure in TTX diff.

Issue #3066, opened 2021-07-22
--variations and --shaper=directwrite


simoncozens:
CoreText *does* support FeatureVariations (I checked with a HTML demo in Safari). But it seems like `hb-shape --shaper=coretext` does not:

```
$ hb-shape bahnschrift.ttf '$' --variations=wght=100 --shaper=ot
[gid696=0+1181]
 hb-shape bahnschrift.ttf '$' --variations=wght=1000 --shaper=ot
[gid730=0+1181]

$ hb-shape bahnschrift.ttf '$' --variations=wght=100 --shaper=coretext
[gid730=0+1181]
$ hb-shape bahnschrift.ttf '$' --variations=wght=1000 --shaper=coretext
[gid730=0+1181]
```

behdad:
We don't call CoreText variations APIs. Any idea what the API is called? cc @drott

khaledhosny:
This is probably true for `directwrite` shaper (and `uninscribe`, but I don’t think GDI has a variations API).

simoncozens:
Yes, directwrite doesn't work either (but then DW doesn't [really support variations](https://twitter.com/simoncozens/status/1418162669069246465) too well anyway...). An example of the DW API can be found [here](https://github.com/microsoft/DWriteShapePy/blob/16b7ea0990e94af342598ab28601cfc1b9dc3f3a/src/cpp/DWriteShapeLib.cpp#L300-L313).

drott:
> We don't call CoreText variations APIs. Any idea what the API is called? cc @drott @jfkthame

You mean what API is needed for applying variations to a `CTFontRef`? I think that would be [CTFontCreateCopyWithAttributes](https://developer.apple.com/documentation/coretext/1511225-ctfontcreatecopywithattributes?language=objc) - and specifying the variations for the `kCTFontVariationAttribute` key of the attributes dictionary, example [here in Skia](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/src/ports/SkTypeface_mac_ct.cpp;l=1215).

> Yes, directwrite doesn't work either (but then DW doesn't [really support variations](https://twitter.com/simoncozens/status/1418162669069246465) too well anyway...). An example of the DW API can be found [here](https://github.com/microsoft/DWriteShapePy/blob/16b7ea0990e94af342598ab28601cfc1b9dc3f3a/src/cpp/DWriteShapeLib.cpp#L300-L313).

[...]

PR #3112, opened 2021-08-04
Support passing variations to coretext shaper

simoncozens:
Deals with the coretext half of #3066.


behdad:
Thanks Simon!

Would be nice to add tests as well.

PR #3116, opened 2021-08-04
[subset] Prune cmap12

garretrieger:
Matches changes in fonttools to prune cmap12's that are equivalent to other cmap subtables: https://github.com/fonttools/fonttools/pull/2146


khaledhosny:
I personally prefer to update the tests in the same commit that changed the code, otherwise when bisecting the first commit will have failing tests.

garretrieger:
> I personally prefer to update the tests in the same commit that changed the code, otherwise when bisecting the first commit will have failing tests.

Squashed the two commits into one.

khaledhosny:
`distcheck` is failing because you removed the colrv1 expected output directories, but they are still referenced from `test/subset/data/Makefile.am`.

garretrieger:
> `distcheck` is failing because you removed the colrv1 expected output directories, but they are still referenced from `test/subset/data/Makefile.am`.

Fixed.

behdad:
> I personally prefer to update the tests in the same commit that changed the code, otherwise when bisecting the first commit will have failing tests.

Right. But github UI makes it impossible to review individual commits touching thousands of files. We can "squash and merge" the PR.

PR #3119, opened 2021-08-04
[subset] discard extra copies of a table in face builder.

### LINKED ISSUE #2361: [subset] Might end up with multiple table entries for same tag
Will fill in.  Submitting before I forget.

garretrieger:
Fixes #2361. Stores tables in the builder in a hashmap so you end up with at most one copy of each table. Table serialization order is now based on tag sort order instead of order of insertion into the builder.


behdad:
Thanks. I was thinking about what order we should use. Mind implementing a better order at the same time?  I suggest ignoring the OT "recommended order" and just order by `(blob_size, table_tag)`. WDYT?

src/hb-face.cc
```diff
@@ -623,20 +624,7 @@ hb_face_collect_variation_unicodes (hb_face_t *face,
...
-      if (t > tag) return -1;
-      return 0;
-    }
-
-    hb_tag_t   tag;
-    hb_blob_t *blob;
-  };
-
-  hb_vector_t<table_entry_t> tables;
+  hb_hashmap_t<hb_tag_t, hb_blob_t*, (unsigned)-1, nullptr> tables;
```
behdad, line 627: The template defaults for the last two args should be enough.

src/hb-face.cc
```diff
@@ -706,9 +707,8 @@ _hb_face_builder_reference_table (hb_face_t *face HB_UNUSED, hb_tag_t tag, void
   if (!tag)
     return _hb_face_builder_data_reference_blob (data);
 
-  hb_face_builder_data_t::table_entry_t *entry = data->tables.lsearch (tag);
-  if (entry)
-    return hb_blob_reference (entry->blob);
+  if (data->tables.has (tag))
+    return hb_blob_reference (data->tables[tag]);
```
behdad, line 711: No need to check. Always return. Hashmap will return `nullptr` if not found, and `blob_reference` is safe on that.

src/hb-open-file.hh
```diff
@@ -119,10 +119,10 @@ typedef struct OpenTypeOffsetTable
 
   public:
 
-  template <typename item_t>
+  template <typename Iterator>
```
behdad, line 122: Can you maybe add a `hb_is_iterator_of` qualifier here?

src/hb-face.cc
```diff
@@ -682,10 +670,23 @@ _hb_face_builder_data_reference_blob (hb_face_builder_data_t *data)
   c.propagate_error (data->tables);
   OT::OpenTypeFontFile *f = c.start_serialize<OT::OpenTypeFontFile> ();
 
-  bool is_cff = data->tables.lsearch (HB_TAG ('C','F','F',' ')) || data->tables.lsearch (HB_TAG ('C','F','F','2'));
+  bool is_cff = (data->tables.has (HB_TAG ('C','F','F',' '))
+                 || data->tables.has (HB_TAG ('C','F','F','2')));
   hb_tag_t sfnt_tag = is_cff ? OT::OpenTypeFontFile::CFFTag : OT::OpenTypeFontFile::TrueTypeTag;
 
-  bool ret = f->serialize_single (&c, sfnt_tag, data->tables.as_array ());
+  // Sort the tags so that produced face is deterministic.
```
behdad, line 677: Maybe add `tag,blob` into a vector and sort it with a custom cmp.

src/hb-face.cc
garretrieger: I tried this but unfortunately it seems to choke finding the default for the hb_blob_t* pointer.

src/hb-open-file.hh
garretrieger: tried iterator_of but that failed, I suspect similar issue to the map default value parameter. Using hb_is_iterator for now.

src/hb-face.cc
garretrieger: Done.

behdad:
Thank you Garret.

src/hb-face.cc
behdad: Okay fixed it.

src/hb-open-file.hh
behdad: Right, that had some problems. I'll see if I can fix.

Issue #3109, opened 2021-08-03
Support ranges in hb-subset --unicodes


behdad:
Currently `--unicodes` docs on `hb-subset` say:
```
  -u, --unicodes=list of hex numbers                                  Set input Unicode codepoints
```
ie. it doesn't support ranges. Which is not surprising since it's sharing implementation with `hb-shape`.

We shouldd implement ranges for `hb-subset --unicodes`; and also add `--unicodes-file` (issue https://github.com/harfbuzz/harfbuzz/issues/3002).

_Originally posted by @behdad in https://github.com/harfbuzz/harfbuzz/issues/3038#issuecomment-871735705_

Issue #3002, opened 2021-06-04
[util] Add --unicodes-file


behdad:
Make it also ignore comments starting with '#'.

arzanilorestan:
A very, very important article by me (Satoshi Nakamoto), the true and true creator of digital currencies for the whole world .____ Important ___ Important ___ Important ___ Important ___ Important ___ Important ___
http://sendanywhe.re/ZHTRREP9

behdad:
Make sure it can accept fontconfig `.orth` files natively. I believe the fonttools subsetter does that, but haven't checked.

Issue #2455, opened 2020-06-08
Add hb-subset --glyphs / --glyphs-file


madig:
IIUC, hb-subset will currently only take a text to choose what to subset. pyftsubset will also take a list of glyph names or a file with a list of glyph names (`--glyphs` and `--glyphs-file`). This allows to explicitly select glyphs without a Unicode code point attached. Would be nice to have in hb-subset :)

garretrieger:
In addition to --text, hb-subset also supports the "--gids" option which can be used to specify the glyph ids to keep. So if you're able to map glyph names to ids for your use case you could use that.

madig:
In version 2.6.7, there is no `--gids` option?

behdad:
We should add the `--glyphs-file` and `--gids-file` etc that match pyftsubset IMO.

PR #3121, opened 2021-08-05
Fix up `hb-subset --help-all`

behdad:
Currently it uses `font_opts_t` and `text_opts_t` from `hb-shape/hb-view`, which is not what we need, so there's lots of extra unused args. Trying to fix that as well as:

  - [ ] https://github.com/harfbuzz/harfbuzz/issues/3002
  - [ ] https://github.com/harfbuzz/harfbuzz/issues/3109
  - [ ] https://github.com/harfbuzz/harfbuzz/issues/2455


behdad:
Okay. The refactoring is done. Now I can actually implement those missing features :)).

PR #3122, opened 2021-08-05
[subset] don't leak memory for duplicate tables.

garretrieger:


behdad:
This would keep the old table. We should remove it and add the new one.

garretrieger:
> This would keep the old table. We should remove it and add the new one.

Fixed.

behdad:
Lovely. Thanks.

PR #3124, opened 2021-08-06
[subset] fail in face builder if allocation for table sorting fails.

garretrieger:
Fixes https://oss-fuzz.com/testcase-detail/5041767803125760


src/hb-face.cc
```diff
@@ -690,6 +690,12 @@ _hb_face_builder_data_reference_blob (hb_face_builder_data_t *data)
   // Sort the tags so that produced face is deterministic.
   hb_vector_t<hb_pair_t <hb_tag_t, hb_blob_t*>> sorted_entries;
   data->tables.iter () | hb_sink (sorted_entries);
+  if (sorted_entries.in_error ())
```
behdad, line 693: Add `unlikely` please.

behdad (approved):

PR #3126, opened 2021-08-07
Fix documentation of hb_subset_flags_t defaults

bungeman:
HB_SUBSET_FLAGS_DEFAULT is 0, and hb_subset_input_create_or_fail
initializes its flags to HB_SUBSET_FLAGS_DEFAULT, so currently all flags
are set to false by default. The documentation for
HB_SUBSET_FLAGS_NO_HINTING is fixed to state that it is false by
default, and the default state is documented for
HB_SUBSET_FLAGS_NO_PRUNE_UNICODE_RANGES and
HB_SUBSET_FLAGS_RETAIN_ALL_FEATURES.

### DIFF
--- a/src/hb-subset.h
+++ b/src/hb-subset.h
@@ -41,27 +41,24 @@ typedef struct hb_subset_input_t hb_subset_input_t;
 
 /**
  * hb_subset_flags_t:
- * @HB_SUBSET_FLAGS_DEFAULT: all flags at their default value.
+ * @HB_SUBSET_FLAGS_DEFAULT: all flags at their default value of false.
  * @HB_SUBSET_FLAGS_NO_HINTING: If set hinting instructions will be dropped in
  * the produced subset. Otherwise hinting instructions will be retained.
- * Defaults to true.
  * @HB_SUBSET_FLAGS_RETAIN_GIDS: If set glyph indices will not be modified in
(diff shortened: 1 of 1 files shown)


behdad:
Thanks.

Do we even need to document each flag defaulting to false? @garretrieger

bungeman:
I noticed that too. At the moment `hb_subset_input_create_or_fail` could set some flags to `true` by default if it wanted to (by not setting its flags to `HB_SUBSET_FLAGS_DEFAULT`). It seems that either

- `HB_SUBSET_FLAGS_DEFAULT` should not be public (so no one assumes that all flags will be default false in the future) and `hb_subset_input_create_or_fail` set the flags however they are documented on `hb_subset_flags_t` or
- `hb_subset_flags_t` document that the default for all flags is `false`, both now and in the future.

behdad:
Garrett and I decided to keep the default zero. That's why done flags are negated (`NO_`), which left the docs out of sync as you found.

So we can document it that way now.

bungeman:
Changed proposed documentation so that `HB_SUBSET_FLAGS_DEFAULT` states that the defaults are `false` and remove the `Defaults to false.` (and the one errant `Defaults to true.`) on the other flags' documentation.

behdad:
Thank you Ben.

PR #3132, opened 2021-08-10
Use semicolon in shaping test suite

behdad:


khaledhosny:
Looks good, thanks @behdad!

PR #3133, opened 2021-08-10
[subset] fix bug in parsing glyf flags.

### LINKED ISSUE #3131: Subset glyph lacks outline.
Originally reported as [Chromium issue 1231818](https://crbug.com/1231818).

Building current tip of tree HarfBuzz ( 8940409e3c108e2c26c0f22679eb3c6cea4fee9d ) with the default meson build and running
```
    util/hb-subset --retain-gids testing-font.ttf "cdeopqr" -o test-font.subset.ttf
```
produces a subset font where 'd', 'p', and 'q' are present but have no outline, even though they have an outline in the original font. Even after laundering this font through `ttx` (decompile and recompile to ensure sane data) the same happens. The missing glyph outlines are strange in that they have a large number of points near each other (not sure if this is intentional or not).

Providing the font in [testing-font.zip](https://github.com/harfbuzz/harfbuzz/files/6963252/testing-font.zip), since `.ttf` files are blocked.

garretrieger:
the bytes of the flag stream were being treated as signed integers instead of unsigned as specified in the spec. Fixes #3131


behdad:
LGTM

PR #3134, opened 2021-08-10
[map] fix incorrect population count in hash map.

garretrieger:
If the same key was set twice the population was being incorrectly incremented. Fixes https://oss-fuzz.com/testcase-detail/5522792714993664


behdad:
Lgtm.

PR #3142, opened 2021-08-12
[util/hb-subset] preload face

behdad:
This loads the font face before parsing other args, so we can parse glyph names / etc as we go, which allows for correctly removing glyphs by name as well, eventually, after we also implement `hb_set_invert` (to be used for handling `--unicodes='*'` and `--gids='*'` etc).


behdad:
We should do something similar in fonttools subsetter as well. cc @anthrotype

Issue #3130, opened 2021-08-09
[subset] Use a central API for various sets


behdad:
Reviewing the subset library API (as part of https://github.com/harfbuzz/harfbuzz/issues/3078) I see:
```
0000000000046ac0 T _hb_subset_input_create_or_fail
0000000000046fe0 T _hb_subset_input_destroy
00000000000471d0 T _hb_subset_input_drop_tables_set
0000000000047210 T _hb_subset_input_get_flags
00000000000473e0 T _hb_subset_input_get_user_data
0000000000047150 T _hb_subset_input_glyph_set
00000000000471b0 T _hb_subset_input_layout_features_set
0000000000047170 T _hb_subset_input_nameid_set
0000000000047190 T _hb_subset_input_namelangid_set
00000000000471f0 T _hb_subset_input_no_subset_tables_set
0000000000046f10 T _hb_subset_input_reference
0000000000047230 T _hb_subset_input_set_flags
0000000000047250 T _hb_subset_input_set_user_data
0000000000047130 T _hb_subset_input_unicode_set
00000000000c2060 T _hb_subset_or_fail
```

we merged all boolean flags into one flags API. I suggest we also merge all the "set" APIs into one `hb_subset_input_???_set` with an enum. We will keep adding such sets for sure (for variation axes, other aspects).

behdad:
WDYT? @garretrieger @bungeman

behdad:
So, of these sets:

> 0000000000047130 T _hb_subset_input_unicode_set
> 00000000000471d0 T _hb_subset_input_drop_tables_set
> 0000000000047150 T _hb_subset_input_glyph_set
> 00000000000471b0 T _hb_subset_input_layout_features_set
> 0000000000047170 T _hb_subset_input_nameid_set
> 0000000000047190 T _hb_subset_input_namelangid_set
> 00000000000471f0 T _hb_subset_input_no_subset_tables_set

Garret expressed that my proposed unifying them doesn't make the API easier to use. I still like to do it because it makes the various sets enumerable.

A couple naming issues: "layout_features" and "no_subset_tables" and "drop_tables" are plural but shouldn't be.

I'm fine having direct API for glyph set and unicode set as well, treating them above others, but the rest to me should go in an enum. We can make the names also identify whether it's a tag set or index set. Like: `layout_feature_tag` `name_lang_index` etc.

garretrieger:
[...]

PR #3146, opened 2021-08-13
[subset] proposed update to subset set apis.

garretrieger:
For #3130


src/hb-subset.h
```diff
@@ -84,6 +84,29 @@ typedef enum { /*< flags >*/
...
+ * List of sets that can be configured on the subset input.
+ *
+ * Since: REPLACE
+ **/
+typedef enum { /*< flags >*/
+  HB_SUBSET_SETS_LAYOUT_FEATURE_TAG = 0,
+  HB_SUBSET_SETS_NAME_INDEX,
+  HB_SUBSET_SETS_NAME_LANG_INDEX,
+  HB_SUBSET_SETS_NO_SUBSET_TABLE_TAG,
+  HB_SUBSET_SETS_DROP_TABLE_TAG,
```
behdad, line 107: Let's add a `HB_SUBSET_SETS_COUNT` at the end? Similarly probably add `HB_SUBSET_FLAG_DEFINED` as or of all.

src/hb-subset.h
```diff
@@ -84,6 +84,29 @@ typedef enum { /*< flags >*/
...
+ * subsetted.
+ * @HB_SUBSET_SETS_DROP_TABLE_TAG: the set of table tags which specifies tables which will be dropped
+ * in the subset.
+ *
+ * List of sets that can be configured on the subset input.
+ *
+ * Since: REPLACE
+ **/
+typedef enum { /*< flags >*/
+  HB_SUBSET_SETS_LAYOUT_FEATURE_TAG = 0,
```
behdad, line 103: Make 0 be "UNICODE" and 1 be "GLYPH_INDEX".

bungeman:
Already commented over on the API proposal itself, but the implementation of this is going to look like a switch, or worse these sets will get placed into an array and be accessed by enum. But to what end is this extra level of indirection? What is the improvement? Is it fewer lines in the documentation? That's just a documentation formatting issue. Fewer overall API functions? But there is no actual reduction in complexity here. In fact, it's really confusing that a bunch of unrelated data happens to be going through one API function simply because they accidentally have the same type. Also, I can detect the existence of a function, but not an enum value, so this is actually a step backwards in API usefulness for users trying to stay near tip of tree.

src/hb-subset-input.hh
```diff
@@ -40,23 +42,110 @@ struct hb_subset_input_t
...
+  {
+    return sizeof (set_ptrs) / sizeof (hb_set_t*);
+  }
+
+  inline hb_array_t<hb_set_t*> sets_iter ()
+  {
+    return hb_array_t<hb_set_t*> (set_ptrs, num_sets ());
+  }
+
+  inline hb_set_t* unicodes()
```
behdad, line 70: Are these functions really necessary? Instead of just typing `input->sets.unicodes`?

garretrieger: Removed.

src/hb-subset.h
```diff
@@ -79,6 +79,33 @@ typedef enum { /*< flags >*/
...
+ * @HB_SUBSET_SETS_DROP_TABLE_TAG: the set of table tags which specifies tables which will be dropped
+ * in the subset.
+ * @HB_SUBSET_SETS_NAME_ID: the set of name ids that will be retained.
+ * @HB_SUBSET_SETS_NAME_LANG_ID: the set of name lang ids that will be retained.
+ * @HB_SUBSET_SETS_LAYOUT_FEATURE_TAG: the set of layout feature tags that will be retained
+ * in the subset.
+ *
+ * List of sets that can be configured on the subset input.
+ *
+ * Since: REPLACE
```
behdad, line 97: "REPLACEME" please. We grep for that.

Issue #3078, opened 2021-07-27
HarfBuzz 3.0.0 tracker issue


behdad:
With the subsetter library approaching API/ABI stability I think we should transition to HB 3.x. Here's a tracker for what needs fixing before that release. Mainly it is just:

  - [x] https://github.com/harfbuzz/harfbuzz/issues/3012
    - [x] Final check of the .so to make sure exposed symbols are all good and intended. https://github.com/harfbuzz/harfbuzz/issues/3130
    - [ ] Ideally, a "getting started" section to the subset library; maybe exists already.
    - [ ] Proper release notes.
    - [ ] autoconf/meson options to disable building it? Perhaps not.

 

But a couple small issues as well:

  - [x] https://github.com/harfbuzz/harfbuzz/pull/3076

  - [x] https://github.com/harfbuzz/harfbuzz/issues/3077

  - [ ] This oss-fuzz issue:
     https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=36236

     I think I introduced it in https://github.com/harfbuzz/harfbuzz/pull/3027 so I want to fix before a release.

Would be nice to fix, but definitely no release-blocker:

  - [x] https://github.com/harfbuzz/harfbuzz/issues/3069
  - [ ] https://github.com/harfbuzz/harfbuzz/issues/3074

behdad:
Ideally also:
  - [ ] https://github.com/harfbuzz/harfbuzz/issues/3002
  - [ ] https://github.com/harfbuzz/harfbuzz/issues/3109
  - [ ] https://github.com/harfbuzz/harfbuzz/issues/3038
  - [ ] https://github.com/harfbuzz/harfbuzz/issues/2455

stha09:
Is it possible to get a release soon? Latest Chromium Dev update already depends on unreleased harfbuzz features.

behdad:
We can make a release today, call it 2.9.0. There's still API changes I like to make so we're not at 3.0.0 mark yet.

@khaledhosny can you roll a release please?

khaledhosny:
Should I list the changes to subset API in the NEWS file, so far we never listed any subset API there.

khaledhosny:
Also, I don’t like the fact we are going to break the subset API twice in a short period of time (even with being unstable and all), we should change it once for good. Chromioum depending on an unreleased HarfBuzz looks like something that should be fixed there.

behdad:
[...]

PR #3147, opened 2021-08-13
[subset] restore old subsetting api methods.

garretrieger:
These will be left in place until the 3.0.0 release. As discussed in #3078

### DIFF
--- a/src/hb-subset-input.cc
+++ b/src/hb-subset-input.cc
@@ -448,7 +448,7 @@ hb_subset_input_set_retain_all_features (hb_subset_input_t *subset_input,
 hb_bool_t
 hb_subset_input_get_retain_all_features (hb_subset_input_t *subset_input)
 {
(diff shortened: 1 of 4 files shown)


src/hb-subset-input.cc
```diff
@@ -426,3 +426,133 @@ hb_subset_input_get_user_data (const hb_subset_input_t *input,
...
+{
+  return set_flag_value (subset_input,
+                         HB_SUBSET_FLAGS_RETAIN_ALL_FEATURES,
+                         value);
+}
+
+hb_bool_t
+hb_subset_input_get_retain_all_features (hb_subset_input_t *subset_input)
+{
+  return hb_subset_input_get_flags (subset_input) & HB_SUBSET_FLAGS_RETAIN_ALL_FEATURES;
```
behdad, line 451: Cast to (bool) before returning.

behdad:
Thanks. One issue. Lgtm otherwise.

PR #3148, opened 2021-08-15
Build win64 binaries

khaledhosny:


behdad:
Thanks! I never had luck with this.

Issue #1843, opened 2019-07-19
[buffer] hb_buffer_append doesn't deal with pre/post-context


behdad:
It should.

behdad:
I want to fix this.

behdad:
Or, any volunteers? @khaledhosny ?

behdad:
To do that we should change `hb_utf32_xe_t` to also take an optimize `stride` parameter. Then we can cast such a template to read the `hb_glyph_info_t::codepoint`. Then move the two blocks for adding pre/post-context from `hb_buffer_add_utf` into their own template and use those in `hb_buffer_append` as well.

PR #3150, opened 2021-08-16
[test] Add tests for item context

khaledhosny:
Test pre-context and post-context in Arabic shaper, as well as interaction of pre-context with BOT flag


src/hb-buffer.cc
```diff
@@ -1755,6 +1755,22 @@ hb_buffer_append (hb_buffer_t *buffer,
...
+      for (unsigned i = 0; i < source->context_len[0]; i++)
+	buffer->context[0][i] = source->context[0][i];
+    }
+    if (end == source->len && source->context_len[1] && !buffer->context_len[1])
+    {
+      buffer->context_len[1] = source->context_len[1];
+      for (unsigned i = 0; i < source->context_len[1]; i++)
+	buffer->context[1][i] = source->context[1][i];
+    }
+  }
```
khaledhosny, line 1773: This looked like the easiest way to make `verify_buffer_safe_to_break ()` copy the context when appending various buffer, but I’m not comfortable with the fact that it changes the behavior of public API.

behdad: I'm fine with behavior change, as I filed https://github.com/harfbuzz/harfbuzz/issues/1843

Let me review correctness though.

behdad: The logic seems like can use some fine-tuning to me.  See similar logic in add_utf. Ideally we should reuse that as I described in #1843. I'll see if I can help with that.

I'll comment above about what doesn't look right to me.

src/hb-buffer.cc
```diff
@@ -1755,6 +1755,22 @@ hb_buffer_append (hb_buffer_t *buffer,
   memcpy (buffer->info + orig_len, source->info + start, (end - start) * sizeof (buffer->info[0]));
   if (buffer->have_positions)
     memcpy (buffer->pos + orig_len, source->pos + start, (end - start) * sizeof (buffer->pos[0]));
+
+  if (source->content_type == HB_BUFFER_CONTENT_TYPE_UNICODE)
+  {
+    if (start == 0 && orig_len == 0 && source->context_len[0] && !buffer->context_len[0])
```
behdad, line 1761: If `start > 0 && orig_len == 0` then part of `source` before `start` should also be added as `pre_context`, followed by `source`'s `pre_context`.

src/hb-buffer.cc
```diff
@@ -1755,6 +1755,22 @@ hb_buffer_append (hb_buffer_t *buffer,
...
+    {
+      buffer->context_len[0] = source->context_len[0];
+      for (unsigned i = 0; i < source->context_len[0]; i++)
+	buffer->context[0][i] = source->context[0][i];
+    }
+    if (end == source->len && source->context_len[1] && !buffer->context_len[1])
```
behdad, line 1767: If `end < source->len` then `source`'s after `len` should also be added (unconditionally) as post-context, followed by `source`'s post-context.

src/hb-buffer.cc
behdad: That seems too much work. Let's duplicate logic.

src/hb-buffer.cc
khaledhosny: Shouldn’t it be `source`’s pre-context followed by the part of `source` before `start`?

src/hb-buffer.cc
khaledhosny: Should this replace the existing condition or be in addition to it?

src/hb-buffer.cc
behdad: Depends on which direction you look at. Pre-context moves backward, so I said it in the order it would be coded.

src/hb-buffer.cc
behdad: Right. Replace it. The `end == source->len` check is arbitrary. In reality, the two pieces should be walked.

khaledhosny: I did that, but if t`end == source->len`, no context will be added and my tests will fail, so i changed it to `end <= source->len`.

behdad: What I mean is that you'd have two loops, one walking from `end` to `source->len` and adding it, followed by another walking `source->context[1]` and adding it.

khaledhosny: This is what I have now:
```C++
    if (end <= source->len)
    {
      while (end < source->len && buffer->context_len[1] < buffer->CONTEXT_LENGTH)
        buffer->context[1][buffer->context_len[1]++] = (source->info + end++)->codepoint;
      for (unsigned i = 0; i < source->context_len[1] && buffer->context_len[1] < buffer->CONTEXT_LENGTH; i++)
        buffer->context[1][buffer->context_len[1]++] = source->context[1][i];
    }
```

behdad: And it doesn't work?

The `end <= source->len` is not needed as we have ensured that already.

`(source->info + end++)->codepoint` is same as `source->info[end++].codepoint`.

behdad: Here, why don't you dip your toes into our iterator design? Makes it much nicer.

```
  auto writer = hb_array (context[1]);
  hb_copy (hb_array (&source->info[end].codepoint, source->len - end, sizeof (source->info[0])), writer);
  hb_copy (hb_array (source->context[1], source->context_len[1]), writer);
```

however, currently doesn't work since `hb-array` doesn't accept stride. I like to fix that.

behdad: Even if that worked, we'd need to write to `context_len[1]` after.

khaledhosny: I _think_ I finally got what you are saying, and I updated the code according to that understanding.

PR #3154, opened 2021-08-17
Resurrect hb_set_invert()

behdad:


behdad:
Okay, this is ready for review.

behdad:
Umm. Needs tests to be written. Anyone can help with that?

src/hb-bit-set-invertible.hh
```diff
@@ -0,0 +1,307 @@
...
+    {
+      if (unlikely (inverted))
+	process (hb_bitwise_or, other);
+      else
+	process (hb_bitwise_and, other); /* Main branch. */
+    }
+    else
+    {
+      if (unlikely (inverted))
+	process (hb_bitwise_ge, other);
```
garretrieger: I believe this should be lt

src/hb-bit-set-invertible.hh
```diff
@@ -0,0 +1,307 @@
...
+      else
+	process (hb_bitwise_le, other);
```
garretrieger: gt

src/hb-bit-set-invertible.hh
```diff
@@ -0,0 +1,307 @@
...
+    {
+      if (unlikely (inverted))
+	process (hb_bitwise_lt, other);
+      else
+	process (hb_bitwise_gt, other); /* Main branch. */
+    }
+    else
+    {
+      if (unlikely (inverted))
+	process (hb_bitwise_non, other);
```
garretrieger: or

src/hb-bit-set-invertible.hh
```diff
@@ -0,0 +1,307 @@
...
+    auto old = *codepoint;
+    auto v = old;
+    s.next (&v);
+    if (old + 1 < v)
+    {
+      *codepoint = old + 1;
+      return true;
+    }
+
+    if (unlikely (old + 1 == INVALID))
```
garretrieger, line 192: I think you also want to check if old == INVALID

behdad: No, `old == INVALID` is how we start iteration. `old+1` becomes 0 and it all works.

PR #3156, opened 2021-08-18
Fix CI workflow for publishing releases

alerque:
This should fix the issue [noted here](https://github.com/harfbuzz/harfbuzz/issues/3078#issuecomment-900896994) in the 2.9.0 release cycle where the automatic release artifact publishing did not work.

The culprit here was d203267e5 which on first glance is quite logical. The trouble is Circle CI is not logical here and these filters were not "effectively ignored", they were effectively helping the jobs to not be ignored. In order for a job to run as a dependency of another job that has filters it must itself have a filter.

Since clearly the intent of d203267e5 was to also run `distcheck` on PRs and branch pushes, I went ahead and setup filters that will work for that as well as for use as a dependency for the ‘publish-*’ series of jobs.

The expected workflow output can be [previewed on CI for my fork](https://app.circleci.com/pipelines/github/alerque/harfbuzz) including a phony release cycle.

### DIFF
--- a/.ci/publish_release_artifact.sh
+++ b/.ci/publish_release_artifact.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+
 set -e
 set -o pipefail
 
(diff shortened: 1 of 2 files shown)


alerque:
Marking this as draft for a moment because while the Circle CI workflows ran as expected on this the GitHub UI isn't showing the expected release in my fork. It is probably unrelated, but let me check into that before this gets merged...

alerque:
Good to go. The release and some of the artifacts **did** show up, the one artifact that didn't failed because of the way I was doing phony releases (tagging without editing the source to tell autotools what version was being built).

khaledhosny (approved):

khaledhosny:
Thanks!

Issue #1300, opened 2018-10-23
Non-nullable offsets in null objects...


behdad:
And other objects that have variable length external to the struct itself, cannot be safely placed on the null pool.

behdad:
Humm.  While I've now fixed disallowing taking Null() of objects with unbounded size, the original report in the subject still is valid.  Ie. if a struct has a non-nullable offsets to an unbounded struct (currently allowed), the outer struct itself essentially becomes unbounded and should become non-nullable.  Unless it guarantees to access inner struct (referred to by offset) properly.   I don't think there's a way to enforce that...

behdad:
> I don't think there's a way to enforce that...

Now that we use C++11 enforcing these kinds of things has become easier, I mean, maybe possible.

behdad:
> Humm. While I've now fixed disallowing taking Null() of objects with unbounded size, the original report in the subject still is valid. Ie. if a struct has a non-nullable offsets to an unbounded struct (currently allowed), the outer struct itself essentially becomes unbounded and should become non-nullable. Unless it guarantees to access inner struct (referred to by offset) properly. I don't think there's a way to enforce that...

So a non-nullable offset to an unbounded size should make the offset refuse to be embedded in Null objects by way of metaprogramming magic.  Let me try.

PR #3157, opened 2021-08-18
Don't serialize null offsets in CPAL.

garretrieger:
Fixes https://oss-fuzz.com/testcase-detail/5443213648330752


behdad:
I don't understand. `serialize_copy` starts like this:

```
  template <typename ...Ts>
  bool serialize_copy (hb_serialize_context_t *c, const OffsetTo& src,
                       const void *src_base, unsigned dst_bias,
                       hb_serialize_context_t::whence_t whence,
                       Ts&&... ds)
  {
    *this = 0;
    if (src.is_null ())
      return false;
    ...
```

which should do the same.

Ugh. I see... Those fields are defined this way:
```
  protected:
  NNOffset32To<UnsizedArrayOf<HBUINT32>>
                paletteFlagsZ;          /* Offset from the beginning of CPAL table to
                                         * the Palette Type Array. Set to 0 if no array
                                         * is provided. */
  NNOffset32To<UnsizedArrayOf<NameID>>
                paletteLabelsZ;         /* Offset from the beginning of CPAL table to
                                         * the palette labels array. Set to 0 if no
                                         * array is provided. */
  NNOffset32To<UnsizedArrayOf<NameID>>
                colorLabelsZ;           /* Offset from the beginning of CPAL table to
                                         * the color labels array. Set to 0
                                         * if no array is provided. */
```

See those `NNOffset...`? The `NN` stands for non-null. Clearly these offsets do allow null. So just remove the `NN`. I never understood why the person writing this code added those to begin with. I've seen other instances....

behdad:
As per offline discussion, trying to remove the `NN` fails because `UnsizedArrayOf` doesn't declare null_size. That's intentional, and indeed you cannot safely have a nullable offset to an unbounded type.  That is caught by the compiler.

What is NOT caught by the compiler, and indeed that's the subject of https://github.com/harfbuzz/harfbuzz/issues/1300, is that non-nullable offset to unbounded-size object are even more problematic. But those are currently not caught by the compiler, as is shown in this issue.

PR #3158, opened 2021-08-19
[subset] make glyph-names option match fonttools behavior

qxliu76:
Don't encode psNames that are contained in standard glyph names set


src/hb-ot-standard-glyph-order-static.hh
```diff
@@ -0,0 +1,330 @@
...
+    const char *g = glyph;
+    for (unsigned i = 0; i < key.length; i++)
+    {
+      char k = key.arrayZ[i];
+      if (*g == '\0' || k > *g) return 1;
+      if (k < *g) return -1;
+      g++;
+    }
+    if (*g == '\0') return 0;
+    return -1;
```
behdad, line 45: Why are you not using `strcmp`?!

qxliu76: oh, I found arrayZ in hb_bytes_t doesn't include a trailing '\0'. I will change to strcmp

qxliu76: done

behdad:
Thanks. See if you can use `hb-string-array` instead. That should save a couple kb.

qxliu76:
> Thanks. See if you can use `hb-string-array` instead. That should save a couple kb.

Done, thank you! I should use already existing utilities. :)

behdad:
Cool. I thought we might have that name list already. :)

test/subset/meson.build
```diff
@@ -36,6 +36,7 @@ tests = [
   'colr_with_components',
   'cbdt',
   'variable',
+  'glyph_names',
```
khaledhosny, line 39: You need to add this to the relevant `Makefile.sources` to fix the distcheck failure.

qxliu76: Done, thank you!

PR #3162, opened 2021-08-23
[set] During subset input creation check for memory alloc failures.

garretrieger:


behdad:
I see what's going on. I had thought about this in length. For these operators, I prefer to NOT let the null set get here at all.

The null set is only returned when `hb_set_create()` is called.

IMO the problem is that `hb_subset_input_create()` should check and fail if any of the set/map creations failed.

garretrieger:
> null set is only returned when `hb_set_create()` is called.

Looks like the null set is also returned from hb_set_get_empty() (https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-set.cc#L76)

garretrieger:
> > null set is only returned when `hb_set_create()` is called.
> 
> Looks like the null set is also returned from hb_set_get_empty() (https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-set.cc#L76)

Which means that without the guards added here it's possible to segfault if you attempt to mutate the empty set. Prior to the invertible set change those operations would have been safe. We very likely want to retain the existing public behaviour of the api instead of introducing possible segfault.

behdad:
> We very likely want to retain the existing public behaviour of the api instead of introducing possible segfault.

I've taken care of that in the C-API layer: 1d832693e1

garretrieger:
> > We very likely want to retain the existing public behaviour of the api instead of introducing possible segfault.
> 
> I've taken care of that in the C-API layer: [1d83269](https://github.com/harfbuzz/harfbuzz/commit/1d832693e17935e025201905236b9fa34e1f310d)

Sounds good, I'll update subset plan to check for failed allocations.

src/hb-subset-input.cc
```diff
@@ -45,15 +45,22 @@ hb_subset_input_create_or_fail (void)
...
   hb_set_add (input->name_languages, 0x0409);
-  input->layout_features = hb_set_create ();
-  input->drop_tables = hb_set_create ();
-  input->no_subset_tables = hb_set_create ();
+  if (unlikely ((input->layout_features = hb_set_create ())->in_error ()))
+    return nullptr;
+  if (unlikely ((input->drop_tables = hb_set_create ())->in_error ()))
+    return nullptr;
+  if (unlikely ((input->no_subset_tables = hb_set_create ())->in_error ()))
+    return nullptr;
```
behdad, line 63: Doesn't this block leak earlier objects?

Much easier when we switch to a common API for all sets.

garretrieger: Yes, good catch. I'll fix.

garretrieger:
Updated to only add the hb_subset_input allocation failure checking.

PR #3166, opened 2021-08-24
[subset] Remove retain all layout features flag.

garretrieger:
Instead use inverted sets to handle requesting all features. Modifies feature collection in subset plan to intersect the set of requested features against the features in the font. This prevents iterating a fully filled feature tag set.

### DIFF
--- a/src/hb-subset-plan.cc
+++ b/src/hb-subset-plan.cc
@@ -87,12 +87,13 @@ _remap_indexes (const hb_set_t *indexes,
 #ifndef HB_NO_SUBSET_LAYOUT
 typedef void (*layout_collect_func_t) (hb_face_t *face, hb_tag_t table_tag, const hb_tag_t *scripts, const hb_tag_t *languages, const hb_tag_t *features, hb_set_t *lookup_indexes /* OUT */);
 
+
 template <typename T>
-static void _collect_subset_layout (hb_face_t		 *face,
-                                    const T&              table,
-				    const hb_set_t	 *layout_features_to_retain,
-				    layout_collect_func_t layout_collect_func,
-				    hb_set_t		 *lookup_indices /* OUT */)
+static void _collect_layout_indices (hb_face_t		  *face,
+                                     const T&              table,
+                                     const hb_set_t	  *layout_features_to_retain,
+                                     layout_collect_func_t layout_collect_func,
+                                     hb_set_t		  *lookup_indices /* OUT */)
 {
   hb_vector_t<hb_tag_t> features;
   if (!features.alloc (table.get_feature_count () + 1))
@@ -143,11 +144,11 @@ _closure_glyphs_lookups_features (hb_face_t	     *face,
   hb_blob_ptr_t<T> table = hb_sanitize_context_t ().reference_table<T> (face);
   hb_tag_t table_tag = table->tableTag;
   hb_set_t lookup_indices;
-  _collect_subset_layout<T> (face,
-                             *table,
-                             layout_features_to_retain,
-                             hb_ot_layout_collect_lookups,
-                             &lookup_indices);
+  _collect_layout_indices<T> (face,
+                              *table,
+                              layout_features_to_retain,
+                              hb_ot_layout_collect_lookups,
+                              &lookup_indices);
 
   if (table_tag == HB_OT_TAG_GSUB)
     hb_ot_layout_lookups_substitute_closure (face,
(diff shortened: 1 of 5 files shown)


src/hb-subset.h
```diff
@@ -61,8 +61,6 @@ typedef struct hb_subset_input_t hb_subset_input_t;
  * in the final subset.
  * @HB_SUBSET_FLAGS_NO_PRUNE_UNICODE_RANGES: If set then the unicode ranges in
  * OS/2 will not be recalculated.
- * @HB_SUBSET_FLAGS_RETAIN_ALL_FEATURES: If set all layout features will be
- * retained. If unset then the set accessed by
  * hb_subset_input_layout_features_set() will be used to determine the features
  * to be retained.
```
khaledhosny, line 65: Shouldn’t these two lines go as well?

garretrieger: Removed.

util/hb-subset.cc
```diff
@@ -495,11 +495,9 @@ parse_layout_features (const char *name,
     if (last_name_char == '-')
     {
       hb_set_clear (layout_features);
-      hb_subset_input_set_flags (subset_main->input,
-				 hb_subset_input_get_flags (subset_main->input) & ~HB_SUBSET_FLAGS_RETAIN_ALL_FEATURES);
     } else {
-      hb_subset_input_set_flags (subset_main->input,
-				 hb_subset_input_get_flags (subset_main->input) | HB_SUBSET_FLAGS_RETAIN_ALL_FEATURES);
+      hb_set_clear (layout_features);
```
behdad, line 499: The `hb_set_clear` is common two both branches. Move outside.

src/hb-subset-plan.cc
```diff
@@ -157,12 +164,11 @@ _closure_glyphs_lookups_features (hb_face_t	     *face,
 
   // Collect and prune features
   hb_set_t feature_indices;
-  _collect_subset_layout (face,
-			  table_tag,
-			  layout_features_to_retain,
-			  retain_all_features,
-			  hb_ot_layout_collect_features,
-			  &feature_indices);
+  _collect_subset_layout<T> (face,
```
behdad, line 167: This name, `_collect_subset_layout` means nothing. What is this doing?

src/hb-subset-plan.cc
```diff
@@ -87,42 +87,51 @@ _remap_indexes (const hb_set_t *indexes,
...
+template <typename T>
 static void _collect_subset_layout (hb_face_t		 *face,
 				    hb_tag_t		  table_tag,
 				    const hb_set_t	 *layout_features_to_retain,
-				    bool		  retain_all_features,
 				    layout_collect_func_t layout_collect_func,
 				    hb_set_t		 *lookup_indices /* OUT */)
 {
-  if (retain_all_features)
+  hb_blob_ptr_t<T> table = hb_sanitize_context_t ().reference_table<T> (face);
```
behdad, line 97: You shouldn't be sanitizing yourself.  Use existing facilities to avoid duplicate work.

src/hb-subset-plan.cc
```diff
@@ -87,42 +87,51 @@ _remap_indexes (const hb_set_t *indexes,
...
+  hb_tag_t *features = (hb_tag_t *) hb_malloc ((table->get_feature_count () + 1) * sizeof (hb_tag_t));
```
behdad, line 98: Use vector.

behdad:
My suggestion: use existing `hb-ot-layout` API to get num features in the font, presize a vector, fetch all such features, then filter them.

I was going to do it myself, and was going to experiment a dagger filtering and writing in-place.

behdad:
Sanitizer unhappy about memory leak.

src/hb-subset-plan.cc
garretrieger: Renamed to _collect_layout_indices which better indicates what it does. It runs the provided function to collect a type of indices from a layout table.

garretrieger:
> My suggestion: use existing `hb-ot-layout` API to get num features in the font, presize a vector, fetch all such features, then filter them.
> 
> I was going to do it myself, and was going to experiment a dagger filtering and writing in-place.

We happen to already have a table reference around,  so I've changed to re-use that. It's a bit cleaner to use the table directly than the hb_ot_layout_... methods.

khaledhosny:
> It's a bit cleaner to use the table directly than the hb_ot_layout_... methods.

I prefer using public API in the tools, in one hand it exercises the API, and on the other hand can serve as an example of how to use the API.

PR #3177, opened 2021-08-26
[subset] handle cmap4 overflows.

### LINKED ISSUE #3173: [hb-subset] Failure to subset NotoSansCJK-Regular.ttc
HarfBuzz 2.9.0 from the provided Windows binaries.

1. Get https://github.com/googlefonts/noto-cjk/blob/165c01b46ea533872e002e0785ff17e44f6d97d8/Sans/OTC/NotoSansCJK-Regular.ttc
2. `hb-subset.exe --unicodes-file subset.txt --layout-features="*" NotoSansCJK-Regular.ttc --face-index 0 -o out.ttf`
3. Observe hb-subset exits with error code 1 and leaves a 0 bytes out.ttf.
4. Remove last Unicode value from subset.txt and try again
5. See it working this time?!

Some times it works with the last value, too, for some reason?!?! If you put the value as the only one into a new txt file and use it, hb-subset also finishes properly...

[subset.txt](https://github.com/harfbuzz/harfbuzz/files/7060110/subset.txt)

garretrieger:
If a cmap4 subtable overflows during serialization drop it and the corresponding EncodingRecord. Don't drop the corresponding cmap12 table if it would have otherwise been removed. Fixes #3173.

### DIFF
--- a/src/hb-ot-cmap-table.hh
+++ b/src/hb-ot-cmap-table.hh
@@ -282,9 +282,17 @@ struct CmapSubtableFormat4
     HBUINT16 *idRangeOffset = serialize_rangeoffset_glyid (c, format4_iter, endCode, startCode, idDelta, segcount);
     if (unlikely (!c->check_success (idRangeOffset))) return;
 
-    if (unlikely (!c->check_assign(this->length,
-                                   c->length () - table_initpos,
-                                   HB_SERIALIZE_ERROR_INT_OVERFLOW))) return;
+    this->length = c->length () - table_initpos;
+    if ((long long) this->length != (long long) c->length () - table_initpos)
+    {
+      // Length overflowed. Discard the current object before setting the error condition, otherwise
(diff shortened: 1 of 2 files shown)


behdad:
Thanks Garret.

I wonder if we should simply go with only cmap 12 or only cmap 4...

Basically, if there's any non-BMP, just do cmap 12. Otherwise, do both cmap4 and cmap 12, and if both succeed, store the shorter.

behdad:
I'm fine landing this though, if you think we can remove the extra complexity later.

garretrieger:
> Thanks Garret.
> 
> I wonder if we should simply go with only cmap 12 or only cmap 4...
> 
> Basically, if there's any non-BMP, just do cmap 12. Otherwise, do both cmap4 and cmap 12, and if both succeed, store the shorter.

Yes that sounds good. We'll need to update fonttools to match of course. I'll merge this as is for now and we can do that separately.

PR #3180, opened 2021-08-27
[subset] support notdef-outline option for CFF

qxliu76:


behdad:
I don't understand this commit. What does `closure_subrs` has to do with `notdef_outline`?

qxliu76:
> I don't understand this commit. What does `closure_subrs` has to do with `notdef_outline`?

This is just to match fonttools' result. I observed a diff that with notdef_outline option, numbers before callgsubr are shifted by 1. I just guess that if the CharString for .notdef glyph is emptied,  then no subr is called by this charstring, maybe it should not be included in closure_subrs. I'm not sure whether this is correct since I don't know much about CFF. :)

PR #3182, opened 2021-09-01
Small docs improvement

mildsunrise:
Mention that the returned strings are always nul terminated for convenience.


mildsunrise:
closing, I thought this was specific to these functions but it seems to be general convention across the API. sorry for the fuss

behdad:
It's okay to add them to documentation still. :)

We don't have many places in the API where we return strings. These are some of the few places.

PR #3185, opened 2021-09-02
[subset] fix for (Chain)ContextFormat3: subset lookupRecord

qxliu76:


behdad:
Nowhere I can find what this fixes.

qxliu76:
> Nowhere I can find what this fixes.

oh this is to match fonttools behavior: previously we just copy all lookupRecord.  In the testcase some lookup referenced in the lookupRecord is not retained after subset, and  lookupRecord is still copied since hb_map_get() will return 65535 if this lookup index does not exist in the map. This fix will skip copying this LookupRecord.

khaledhosny:
This should probably be mentioned in the commit message as well.

behdad:
> mentioned in the commit message as well

Correct. And is not merely "to match fonttools". Is a bug fix.

qxliu76:
> This should probably be mentioned in the commit message as well.

Done.

behdad:
I remember these very same code blocks looked smelly to me. I should have raised flags back then.

For starters, a `copy()` method should take no extra parameter like a mapping. It should copy verbatim. If it takes a mapping it's called `subset()` then.

Next, those for loops should pretty much all be written as daggers, and those checks be done as filters. Or better, the `copy`-renamed-`subset` returns `false` if the map doesn't contain the lookup index.

At any rate, something more elegant and less adhoc.

@garretrieger can you please work with @qxliu76 on a cleaner approach? I'm fine merging this for now since it's a bug fix, but a cleanup is called for.

Thanks.

Issue #3179, opened 2021-08-27
not decomposing arabic per UTR#53 


bobh0303:
Observed:

```
$ hb-shape -V /dev/null -u 0622,0670
trace: start reorder    buffer: <U+0622=0|U+0670=0>
trace: end reorder      buffer: <U+0622=0|U+0670=0>
```

Expected:

```
$ hb-shape -V /dev/null -u 0622,0670
trace: start reorder    buffer: <U+0622=0|U+0670=0>
trace: end reorder      buffer: <U+0627=0|U+0670=0|U+0653=0>
```

Description:

The first step in UTR#53 Unicode Arabic Mark Reordering algorithm [AMTRA](https://www.unicode.org/reports/tr53/#AMTRA_Specification) is to decompose to NFD, which would cause:
```
U+0622 ARABIC LETTER ALEF WITH MADDA ABOVE
U+0670 ARABIC SUBSCRIPT ALEF
```
to decompose to 
```
U+0627 ARABIC LETTER ALEF 
U+0653 ARABIC MADDAH ABOVE
U+0670 ARABIC SUBSCRIPT ALEF
```
which would then be correctly reordered (by existing harfbuzz logic) to:
```
U+0627 ARABIC LETTER ALEF 
U+0670 ARABIC SUBSCRIPT ALEF
U+0653 ARABIC MADDAH ABOVE
```
However the initial decomposition to NFD is not happening.

Tested with `hb-shape` 2.9.0

behdad:
> $ hb-shape -V /dev/null -u 0622,0670

With `/dev/null` you won't get the decomposition, since the normalizer would only decompose if the font supports the decomposed sequences.  But if you try with a real font, you'll see that it's actually already working.

bobh0303:
Thanks for your reply, Behdad, and that makes sense. However, something is still not right (or I'm not understanding something):
```
$ hb-shape -V Amiri-Regular.ttf -u 0622,0670
trace: start reorder    buffer: <U+0627=0|U+0653=0|U+0670=0>
trace: end reorder      buffer: <U+0627=0|U+0670=0|U+0653=0>
trace: start table GSUB buffer: [uni0622=0|uni0670=0]
...
```
As you predicted, with a real font the decomposition of U+0622 occurs and marks are reordered.

But then: Why is alefMadda apparently re-composed for the start of GSUB? Is there an NFC normalization pass between `end reorder` and `start table GSUB` ?

behdad:
[...]

PR #3187, opened 2021-09-03
[test] Add tests for Arabic decomposition

khaledhosny:
Test that both NFC and NFD input produces identical results for fonts that used composed fonts internally (Amiri here) and fonts that decompose internally (Noto Nastaliq Urdu here) and that for the former composed forms are used.

See https://github.com/harfbuzz/harfbuzz/issues/3179


behdad:
Thanks. Am fine merging it. What would be better is:

1. Call it `arabic-normalization`;
2. Add smaller tests one per line. They can still share the same font file.

khaledhosny:
OK, I’ll do both.

PR #3188, opened 2021-09-03
[test] Make shape test names match file names

khaledhosny:
No idea why test names are underscorified but it it just makes calling `meson test testname` harder than it should being not able to copy file name directly.


behdad:
I hate such inconsistencies as well. Thanks.

PR #3198, opened 2021-09-09
[repacker] add extension lookup subgraph isolation to the repacker.

garretrieger:
For GSUB/GPOS tables the subgraphs pointed to by an extension lookup subtable via a 32 bit offset can be isolated to significantly improve the effectiveness of the overflow resolver. This change adds that as an explicit step before manual resolutions are attempted.

The subgraphs are isolated by duplicating nodes in the subgraph as necessary to eliminate any incoming edges that originate outside of the subgraph. Additionally the modified distance calculation has been updated to place each extension lookup subgraph into its own space. This prevents interleaving of objects from multiple subgraphs.


behdad:
I'm an opponent of subgraph isolation. That's one of the big problems in the fonttools algorithm and I don't think we should introduce it here. Let's find other ways to fix our algo without adding this. The recursive cloning is part of the fix.

Can you describe what kind of situation you are fixing?

garretrieger:
Roughly the way the sort works is that it packs things into layers based on the number of offsets from root to that particular thing. 32 bit links go into their own layer (by virtue of using 2^32 on the distance modifier). The problem though is that all extension sub table get packed into the same layer (since they all have a single 32 bit offset between them and root). These subtables are typically quite large (which is why they would have been placed into an extension to begin with). So all the objects in the subgraphs end up competing for the same 65kb of space which leads to hundreds of overflows with no easy way to disentangle them.

Since the 32 bit offset allows us to point essentially arbitrarily long distances the most important part of this change is that each 32 bit subgraph is moved into it's own layer by the space multiplier so these subgraphs are no longer competing for the same space.

In the examples I was looking at there was quite a bit a sub table sharing between these subgraphs (hundreds to thousands of links) I assume likely for things like coverage tables that are identical. For the layering above to work effectively these need to be severed. Given the high number of sharing I saw, its more efficient to pre-emptively sever the links between subgraphs than relying on the resolution loop to find and sever the links.

I'm pretty confident we'll need the layering to separate spaces to arrive at resolutions for these cases, but the pre-emptive severing of all links into a subgraph could potentially be dropped if we added recursive duplication to the resolution processing loop. Though I'm not convinced it would produce a significantly better packed result and would probably be slightly slower since multiple resolution loops would probably be needed.

garretrieger:
I thought about this some more and I think there's two main ways to improve this, while making it more targeted:

1. Improve the existing topological sort: the current topological sort is good, but has one major flaw. It doesn't make efficient use of all layers. Particularly the lower layers are likely underutilized and things from higher layers could be moved down freeing up space in the higher more densely packed layers. I have some ideas on how to implement this I'll do a more detailed write up. This would allow the repacker to succeed in cases where it's failing today, but does not fully address the extension lookup issue and will likely still be insufficient for cases with lots of extension lookups.

2. Make extension lookup isolation more intelligent. Roughly:
- Scan the graph to produce a clustering of the extension lookups. Where extension lookups that are connected by shared nodes are in the same cluster.
- Assign a unique space to each cluster of extension lookups.
- If there are overflows within a cluster iteratively isolate lookups from the cluster until overflows are eliminated.

I'll write up more detailed design in a doc for these. For now though I believe we should still try and land this change as it solves the immediate problem and it's done in a way that is very easy to remove it if we implement the above ideas in the future.

src/test-repacker.cc
```diff
@@ -433,7 +529,7 @@ static void test_resolve_overflows_via_sort ()
   void* out_buffer = malloc (buffer_size);
   hb_serialize_context_t out (out_buffer, buffer_size);
 
-  hb_resolve_overflows (c.object_graph (), &out);
+  hb_resolve_overflows (c.object_graph (), 0, &out);
```
behdad, line 532: `HB_TAG_NONE` please.

behdad (commented):
LGTM.

PR #3201, opened 2021-09-10
Release hb-style.h API

### LINKED ISSUE #3199: Release hb-style.h API
Is currently experimental. I like the way it is and seems to work fine. Need to move the enum from `hb-style.cc` to `hb-style.h` and call it stable. cc @ebraminio @khaledhosny @matthiasclasen @jfkthame

khaledhosny:
Fixes #3199.

### DIFF
--- a/src/hb-style.cc
+++ b/src/hb-style.cc
@@ -25,7 +25,6 @@
 #include "hb.hh"
 
 #ifndef HB_NO_STYLE
-#ifdef HB_EXPERIMENTAL_API
 
 #include "hb-ot-var-avar-table.hh"
 #include "hb-ot-var-fvar-table.hh"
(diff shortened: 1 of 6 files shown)


src/hb-style.cc
```diff
@@ -89,13 +61,13 @@ _hb_ratio_to_angle (float r)
...
  * if not set, then tries to get default style values from different
  * tables of the font.
  *
  * Returns: Corresponding axis or default value to a style tag.
  *
- * Since: EXPERIMENTAL
+ * Since: REPLACEME
  **/
 float
 hb_style_get_value (hb_font_t *font, hb_tag_t tag)
```
khaledhosny, line 73: Shouldn’t this be `hb_style_tag_t` instead of plain `hb_tag_t`?

ebraminio: Thanks for this! It supposed to be `hb_style_tag_t` for sure, I couldn't use it as the enum wasn't in public header to not confuse the doc gen.

behdad: Done. Thanks all.

PR #3204, opened 2021-09-13
[subset] fix find_duplicate_features

qxliu76:
We should check each feature against all other features with the same
tag for duplicates.


behdad:
Any idea why the bots are failing?

behdad:
Multiple features can have same tag legitimately. I don't think this change is right. Also, again, this change does not explain what problem it's trying to fix.

khaledhosny:
> Any idea why the bots are failing?

Looks like gitlab.gnome.org was down breaking building glib subproject, I restarted the affected jobs.

qxliu76:
> > Any idea why the bots are failing?
> 
> Looks like gitlab.gnome.org was down breaking building glib subproject, I restarted the affected jobs.

thank you! I did a push to force restart those checks

qxliu76:
> Multiple features can have same tag legitimately. I don't think this change is right. Also, again, this change does not explain what problem it's trying to fix.

Thanks for your comments, I will remember to provide more details next time. 
This change is to fix a problem in pruning redundant LangSys:
When subsetting a Script, we will remove redundant Langsys that are duplicate with the default Langsys in this Script. This will need to check if every feature referenced in the Langsys is duplicate with that in the default LangSys. Find_duplicate_deatures() is doing the check for duplicate features after subset(under same tag, same lookup indices list). Previously we only compare each feature against previous one but actually we should check each feature against all other features with the same tag. This PR is trying to fix this issue.

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -3678,57 +3678,60 @@ struct GSUBGPOS
                                 const hb_set_t *feature_indices,
                                 hb_map_t *duplicate_feature_map /* OUT */) const
   {
+    hb_hashmap_t<hb_tag_t, hb_set_t *, (unsigned)-1, nullptr> tag_features;
```
garretrieger, line 3681: Since feature indices are sorted by tag's alphabetically that means all feature with the same tag are consecutive. So you don't need to store the set for each tag. Rather you could have one set, which you clear every time you encounter a feature with a different tag than the previous.

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -3678,57 +3678,60 @@ struct GSUBGPOS
...
-
-      bool is_equal = true;
-      for (auto _ : + hb_zip (f_iter, prev_iter))
-        if (_.first != _.second) { is_equal = false; break; }
-
-      if (is_equal == true) duplicate_feature_map->set (i, prev);
-      else
-      {
+      
+      same_tag_features->add (i);
```
garretrieger, line 3728: Recommend that you don't add feature indices to this set, if they are duplicate features. Since you only need to test against each unique feature for a given tag.

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -3678,57 +3678,60 @@ struct GSUBGPOS
...
+        + hb_iter (other_f.lookupIndex)
+        | hb_filter (lookup_indices)
+        ;
+  
+        if (f_iter.len () != other_f_iter.len ())
+          continue;
+
+        bool is_equal = true;
+        for (auto _ : + hb_zip (f_iter, other_f_iter))
+          if (_.first != _.second) { is_equal = false; break; }
```
garretrieger, line 3719: For a slightly more compact way of checking iterator equality see: https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-ot-cmap-table.hh#L1488 it skips the len() comparison (which ends up being more efficient since calling .len() triggers a full iteration)

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -3678,55 +3678,60 @@ struct GSUBGPOS
...
+        + hb_iter (f.lookupIndex)
+        | hb_filter (lookup_indices)
+        ;
+  
+        auto other_f_iter =
+        + hb_iter (other_f.lookupIndex)
+        | hb_filter (lookup_indices)
+        ;
+  
+        if (f_iter.len () != other_f_iter.len ())
```
garretrieger, line 3713: This can be removed as

if (is_equal == false || f_iter || other_f_iter) continue; 

below is checking for equal lengths

qxliu76: Good catch! I forgot to remove this line. :)

PR #3205, opened 2021-09-14
Pin ragel version

khaledhosny:


behdad:
Thanks.

khaledhosny:
I’m done with this, but it would be great if someone more knowledgeable about meson can review the change to how exceptions are disabled/enabled for MSVC.

astiob:
Two comments about the last change (ragel subproject):

* It uses `meson.can_run_host_binaries()`, which was introduced in meson 0.55.0, so the minimum meson version should be bumped again.

* It no longer lets me build with no Ragel installed and with `--wrap-mode=nodownload`:

      Program ragel found: NO
      WARNING: Project targeting '>= 0.52.0' but tried to use feature introduced in '0.55.0': meson.can_run_host_binaries.
      
      src/meson.build:295:2: ERROR: Automatic wrap-based subproject downloading is disabled

  Is this intended?

khaledhosny:
Both issues should be addressed in #3209

astiob:
Ah, indeed. My bad! I neglected to check open issues. Thanks!

PR #3206, opened 2021-09-14
Update to Unicode 14.0.0

dscorbett:


behdad:
Thanks!

behdad:
Should we merge, or wait till after 3.0.0?

behdad:
I actually don't see any reason to wait.

Issue #3208, opened 2021-09-15
Build failure on macos


matthiasclasen:
Since the ragel subproject appeared in harfbuzz, the GTK ci build has been failing on macos:

https://gitlab.gnome.org/GNOME/gtk/-/jobs/1517300

khaledhosny:
I might have accidentally depended on too new meson version, as the meson version you are using does not seem to like the second `ragel.found()` call. Unfortunately I’m not able to investigate further at the moment. We probably should also add an option not to build ragel at all and may be make it the default since most harfbuzz builders don't need it.

A quick fix for you is probably to install ragel on the CI job.

behdad:
Let's get 3.0.0 out after this is settled.

PR #3209, opened 2021-09-15
[meson] Add ragel_subproject option

khaledhosny:
Add an option to build fallback ragel subproject when no suitable ragel version is found, and make it off by default since most builder don’t need ragel at all.

Fixes https://github.com/harfbuzz/harfbuzz/issues/3208 (hopefully)

### DIFF
--- a/.github/workflows/linux-ci.yml
+++ b/.github/workflows/linux-ci.yml
@@ -13,10 +13,10 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: install dependencies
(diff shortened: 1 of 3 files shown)


matthiasclasen:
This would be nice to have, thanks

astiob:
Um, gentle prod. (This is blocking CI for downstream projects that try to build without Ragel.)

behdad:
Yay. Finally I can build using meson without it rewriting my sources. Thanks Khaled!

PR #3211, opened 2021-09-16
[subset] In preparation for 3.0.0 release remove legacy subsetting api methods.

garretrieger:
…


src/hb-subset.h
```diff
@@ -126,6 +126,12 @@ HB_EXTERN void *
 hb_subset_input_get_user_data (const hb_subset_input_t *input,
 			       hb_user_data_key_t	   *key);
 
+HB_EXTERN hb_set_t *
+hb_subset_input_unicode_set (hb_subset_input_t *input);
+
+HB_EXTERN hb_set_t *
+hb_subset_input_glyph_set (hb_subset_input_t *input);
+
```
khaledhosny, line 134: I think these two need to be added back to the documentation as I might have mistakingly removed them.

garretrieger: Re-added them.

PR #3212, opened 2021-09-17
[style] Try removing _USE_MATH_DEFINES

behdad:
Internets says I need it on MS compilers. I don't believe the internets.
GCC gives me unused-macros warning. Pushing to CI to check.

### DIFF
--- a/src/hb-style.cc
+++ b/src/hb-style.cc
@@ -26,9 +26,6 @@
 
 #ifndef HB_NO_STYLE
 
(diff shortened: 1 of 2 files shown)


khaledhosny:
It is needed for MSVC, I wanted to not believe too and tried without it first but the build failed. I guess it should be guarded in MSVC ifdef.

behdad:
Ah, you added this, that's why it just appeared.  Okay, I pushed another try.

PR #3215, opened 2021-09-18
Make the code able to compile with MSVC++ from Visual Studio 2015 (Version 14.0.25431.01, Update 3) sucessfully again.

ActuallyaDeviloper:
This is just a tin adjustment to get our build to work again.

### DIFF
--- a/src/hb-subset-input.hh
+++ b/src/hb-subset-input.hh
@@ -1,5 +1,5 @@
 /*
- * Copyright Â© 2018  Google, Inc.
+ * Copyright © 2018  Google, Inc.
  *
  *  This is part of HarfBuzz, a text shaping library.
  *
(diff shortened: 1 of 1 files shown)


alerque:
I take it this was working before and stopped? Is there something that should do in CI so we can confirm this or similar things don't keep getting broken?

behdad:
This is new code. Wet don't test VS 2015 I suppose.

ActuallyaDeviloper:
@alerque Yes, it used to be working with Harfbuzz 2.7.2. Since Unicode 14 is now pusblished we updated first our tables and then Harfbuzz (to version 3.0.0). At that point, we ran into this build problem.

I think only VS 2017 and VS 2019 are currently being checked for builds automatically for Harfbuzz.

PR #3217, opened 2021-09-20
Add support for Pwo Karen / Ason Chin medial la

simoncozens:
See w3c/font-text-cg#43 for context. The Microsoft spec for Burmese clusters is incorrect, in that a medial la can appear after a medial ha; MS has these two characters as mutually exclusive.


src/hb-ot-shape-complex-myanmar-machine.rl
```diff
@@ -76,9 +77,9 @@ k = (Ra As H);			# Kinzi
 
 c = C|Ra;			# is_consonant
 
-medial_group = MY? As? MR? ((MW MH? | MH) As?)?;
+medial_group = MY? As? MR? ((MW MH? ML? | MH ML?) As?)?;
```
behdad, line 80: This doesn't allow a bare `ML As`. Shouldn't it?

behdad:
Closing in favor of https://github.com/harfbuzz/harfbuzz/pull/3220.

PR #3220, opened 2021-09-20
Add support for Pwo Karen / Ason Chin medial la

behdad:
Supersedes @simoncozens's https://github.com/harfbuzz/harfbuzz/pull/3217


dscorbett:
I built this and tested <U+1003, U+1060, U+103A>. It is now getting a dotted circle before the U+103A but I don’t see why.

behdad:
Humm. I don't understand either.

dscorbett (approved):
False alarm: I’m still getting used to this new Meson-based system with a built-in Ragel subproject. This change looks good.

behdad:
> False alarm: I’m still getting used to this new Meson-based system with a built-in Ragel subproject. This change looks good.

Interesting.

Should this be reproducible with NotoSansMyanmar? With that I get same rendering and no dotted-circle with / without this change.

dscorbett:
I was using an hb-view from an earlier make-based build instead of the Ragel-based one I’d just built. I can’t reproduce the error anymore but I’m sure it was all on my end.

PR #3223, opened 2021-09-22
[subset] subset MATH table

qxliu76:
add subset() method for MATH table


src/hb-ot-math-table.hh
```diff
@@ -295,6 +388,18 @@ struct MathKern
 
 struct MathKernInfoRecord
 {
+  MathKernInfoRecord* copy (hb_serialize_context_t *c, const void *base) const
+  {
+    TRACE_SERIALIZE (this);
+    auto *out = c->embed (this);
+    if (unlikely (!out)) return_trace (nullptr);
+
+    for (unsigned i = 0; i < 4; i++)
```
behdad, line 397: Use `for` loop iteration / `zip`, or at least `ARRAY_LENGTH` instead of hard-coding 4.

PR #3224, opened 2021-09-22
[subset] Fix subset_offset_array adding unused space to serializer.

garretrieger:
ArrayOf.serialize_append allocates space for the new item, but ArrayOf.pop() does not recover the allocated space. So in the case where the revert path was entered the extra space added by serialize_append gets left in the serialization buffer. This moves the snapshot to before ArrayOf.serialize_append is called so that revert cleans up the buffer extend.


behdad:
LGTM. Good catch.

PR #3225, opened 2021-09-22
[mutex] Add hb_mutex_t std::mutex implementation

behdad:
Supersedes https://github.com/harfbuzz/harfbuzz/pull/3100.

This implementation requires libc++. But I'm fine with that because it's plugged in a place that we were err'ing and failing before.


behdad:
cc @bungeman @garretrieger

PR #3228, opened 2021-09-23
[repacker] add 'virtual links' to the serializer.

garretrieger:
These aren't associated with an offset field, but instead exist solely to add an ordering constraint to the object graph.

### DIFF
--- a/src/hb-repacker.hh
+++ b/src/hb-repacker.hh
@@ -671,8 +671,9 @@ struct graph_t
         if (visited.has (link.objidx)) continue;
 
         const auto& child = vertices_[link.objidx].obj;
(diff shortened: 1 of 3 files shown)


rsheeter:
Suggest adding a bit more context around when one would use this capability, either to PR or comments or both.

garretrieger:
> Suggest adding a bit more context around when one would use this capability, either to PR or comments or both.

Expanded the comment on add_virtual_link to give examples of when this is useful.

behdad:
Is this obsolete in favor of #3230?

garretrieger:
> Is this obsolete in favor of #3230?

Yes, I'll close this one.

garretrieger:
Superseded by #3230

PR #3233, opened 2021-09-24
[repacker] fix bug in subgraph isolation.

garretrieger:
Prior to this fix id remapping at the end of the isolation operation was fed the old subgraph instead of the new one. Which results in object indices being remapped for the nodes outside of the new subgraph. Adds a test which detects this problem.


behdad:
LGTM when sanitizers are happy.

PR #3235, opened 2021-09-25
[PairPos] Split GPOS kerning to both sides

behdad:
I'm happy with how this is done, just not sure about the performance implication, since the determination can be done at a subtable level but currently is done again and again at every application. Posting for review.

Before:
```
$ hb-shape SourceSerifVariable-Roman.otf VA
[gid23=0+555|gid2=1+663]
```

After:
```
$ hb-shape SourceSerifVariable-Roman.otf VA
[gid23=0+615|gid2=1@-60,0+603]
```


behdad:
This is what it tries to fix:
https://gitlab.gnome.org/GNOME/pango/-/issues/139

behdad:
cc @matthiasclasen

behdad:
I'm trying to reorganize `apply_string` / `apply_forward` to see if I can move the choosing logic there...

behdad:
No idea about the mac fail.

jfkthame:
While I can see the attraction of doing this, I suspect it is responsible for some [test failures](https://treeherder.mozilla.org/jobs?repo=try&revision=085fc6cbe64a702fe8937dd58aca5d09d3d900d4&group_state=expanded) seen when we try to pull the latest HB release into Firefox.

In principle this is a Gecko bug (it doesn't re-shape the text each side of a line-break in cases where it should), and we need to address it there, but as things stand this change aggravates the issue for us. As such, it'd be helpful if it could be made optional, at least for the time being (a build-time flag would be fine, it doesn't need to be a runtime option).

@behdad would you be willing to put this behind `#ifdef HB_BALANCED_KERN` or something like that, so that we can disable it until we're better at handling it?

(I also suspect this is most likely at the root of https://github.com/harfbuzz/harfbuzz/issues/3408 though I haven't investigated that any further.)

behdad:
> @behdad would you be willing to put this behind `#ifdef HB_BALANCED_KERN` or something like that, so that we can disable it until we're better at handling it?

Yes. Though if the default is to be still enabled, it would be `#ifndef HB_NO_BALANCED_KERN`. Happy to put it in now.

> (I also suspect this is most likely at the root of #3408 though I haven't investigated that any further.)

Right... Don't know what's up with that one. I might disable this by default after all as you suggest.

jfkthame:
I don't have a strong opinion on whether it's on or off by default, we can deal with it either way. Though maybe if the old behavior is closer to how other engines (like Uniscribe) handle it, we should have that as the (conservative, more-compatible) default, and tell people to opt in to the new improved behavior if they're confident of compatibility with it.

behdad:
I don't like having random disabled code. If this proves to be problematic, I'd rather just remove it. Let's see if we can make any progress on #3408.

behdad:
> I don't like having random disabled code. If this proves to be problematic, I'd rather just remove it. Let's see if we can make any progress on #3408.

I ended up disabling it.

Issue #3241, opened 2021-09-30
Meson graphite2 dependency is called graphite


behdad:
I don't like that it deviated from autotools.

behdad:
Especially since the report line still says Graphite2:
```
    Graphite2        : YES
```

khaledhosny:
Changing the name would break automated build scripts, but may be we can duplicate the option with the new name and keep the existing name for backward compatibility.

khaledhosny:
Or just rename and let people deal with the breakage.

behdad:
The biggest problem I see right now is that when one configures the project and gets the report, it's very hard to know what to do to change the configuration:
```
harfbuzz 3.0.0

  Additional shapers
    Graphite2        : YES

  Dependencies used for command-line utilities
    Cairo            : YES
    Chafa            : NO

  Directories
    prefix           : /usr/local
    bindir           : bin
    libdir           : lib64
    includedir       : include
    datadir          : share

  Font callbacks (the more the merrier)
    FreeType         : YES

  Other features
    Documentation    : NO
    GObject bindings : YES
    Introspection    : NO
    Experimental APIs: NO

  Platform shapers (not normally needed)
    CoreText         : NO
    DirectWrite      : NO
    GDI/Uniscribe    : NO

  Testing
    Tests            : YES
    Benchmark        : NO

  Unicode callbacks (you want at least one)
    Builtin          : YES
    Glib             : YES
    ICU              : YES
```

If at least the name of the options matches the word we report, it would be obvious.

khaledhosny:
I have no string opinion here. If you want the option renamed we can rename it.

behdad:
If is possible to rename and keep the old name also working, please do. Thanks.

khaledhosny:
I tried to hack something, then thought may be meson have way to deprecate options. It currently doesn’t, but there is a recent PR that does just this: https://github.com/mesonbuild/meson/pull/9184, it will be a while before we can use this deprecation mechanism even if it got merged to day, so I guess I should continue hacking something.

PR #3247, opened 2021-10-05
[meson] Add graphite2 option and deprecate graphite

khaledhosny:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3241

Should be better handled when we can use https://github.com/mesonbuild/meson/pull/9184


meson_options.txt
```diff
@@ -10,6 +10,8 @@ option('chafa', type: 'feature', value: 'auto',
 option('icu', type: 'feature', value: 'auto',
   description: 'Enable ICU library unicode functions')
 option('graphite', type: 'feature', value: 'disabled',
+  description: 'Deprecated use graphite2 option insatead')
```
behdad, line 13: Typo `instead`.

Can we hide this one?

khaledhosny: > Can we hide this one?

I don’t think so, that probably what the meson PR would do.

PR #3256, opened 2021-10-12
Fix Wbitwise-instead-of-logical warnings

nico:
`a || b` only evaluates b if a is false. `a | b` always evaluates
both a and b. If a and b are of type bool, || is usually what you
want, so clang now warns on `|` where both arguments are of type bool.

This warning fires twice in harfbuzz. In both cases, `|` is used
intentionally, with a comment explaining this.

Slightly reorder the code a bit to make the compiler happy, and to
make it obvious even without a comment that both calls should be evaluated.

No intended behavior change.

### DIFF
--- a/src/hb-ot-layout-gpos-table.hh
+++ b/src/hb-ot-layout-gpos-table.hh
@@ -1220,9 +1220,9 @@ struct PairSet
 						record_size);
     if (record)
     {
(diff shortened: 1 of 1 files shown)


nico:
Thanks, Behdad! :)

PR #3259, opened 2021-10-16
API for buffer target length

myoresh:


khaledhosny:
How do you intend to use them s API? Justification happens after lone breaking and by extension shaping. Also a line can involve multiple buffers (different fonts, different scripts etc), how the client will decide what percentage of the line length a buffer should use?

myoresh:
This API is naturally is not intended to be anything more than a draft. Maybe putting it in a PR was not the correct way to ask for a feedback. Of course, the implementation is the hard part, but doing it before agreeing on an API wouldn't be quite an effective approach.

The definition of target length as in https://github.com/harfbuzz/harfbuzz/pull/3259/commits/147e7d15d383fbedb2f51bd37fdc039c69ffdca5 can indeed fit only the simple case of uniform text (one font, one size, one script).

For this case the shaper would do something as follows:
1. Shape the entire buffer as usual
2. Determine the part of buffer which doesn't exceed the target length. For latin scripts the API user might need to do that, because HarfBuzz can't do hyphenation.
3. Shape the relevant part again, fitting it to the target length exactly.

In the more complex case I can think about the same two-pass approach, with a bit more complex API:
1. Call auxiliary function like **hb_compute_adjusted_target_length(hb_font_t *fonts, hb_buffer_t *buffers, unsigned int num_buffers, hb_position_t target_length)**. This function will shape everything and determine the desired expansion coefficient E, e.g. E=1.05, meaning that every buffer fitting into the current line shall be expanded by 5%.
2. Take the first buffer, determine its actual (non-justified) length L and return the adjusted target length E*L to the user.
3. Call hb_justify(..., adjusted_target_length) for the first buffer. This call would expand the buffer exactly as necessary.

I'm not sure whether these steps should be exposed to the user or performed internally. Regarding the hyphenation, the API can accept some sort of callback for that. I will add these changes to the branch, but wanted to ask about the preferred way to put them for review.

khaledhosny:
FWIW, @DigitalKhatt is doing a wonderful job justifying Arabic text using variable fonts, but I think they are using a modified version of HarfBuzz, it would be nice if they can describe what the changes they did and if any of it can be of general use in HarfBuzz proper.

https://twitter.com/DigitalKhatt/status/1434961280062926849

iorsh:
Hello everyone, thank you for your comments! I'll start the initial implementation, and I will share the code once I get some working stuff.

iorsh:
I checked the line-breaking, and it looks like the implementation of line breaking based on dynamic programming (like Knuth-Plass) is quite challenging. I think it would be more feasible to stick to the greedy algorithm for the time being.

@behdad , what do you think? Would it be practical to implement justification without support of advanced line-breaking algorithms?

behdad:
> @behdad , what do you think?

What you are facing is exactly why we have not addressed justification before :).

I'll add my thoughts in a few days.

iorsh:
Hi guys, I pushed the code which performs greedy splitting of buffer into lines - please see https://github.com/harfbuzz/harfbuzz/compare/main...iorsh:justification
Naturally, the interesting stuff is not there yet, but I would be grateful if you could take a look and comment about coding style / conventions or about the overall approach.

behdad:
We do not want justification in HarfBuzz itself.

PR #3282, opened 2021-11-02
[subset] Fix fuzzer found memory leak.

garretrieger:
Happens because an insert into a map with an invalid key reports successful, but this causes the set being inserted to be lost.


src/hb-ot-layout-gsubgpos.hh
```diff
@@ -3664,6 +3664,7 @@ struct GSUBGPOS
     for (unsigned i : feature_indices->iter ())
     {
       hb_tag_t t = get_feature_tag (i);
+      if (t == (unsigned) -1) continue;
```
behdad, line 3667: Please use `HB_MAP_INVALID` or whatever we have. LGTM otherwise. Thanks!

behdad: Or `unique_features::INVALID`?

PR #3284, opened 2021-11-04
[sanitize] add a maximum recursion depth for sanitize.

garretrieger:
Fuzzer has found cases in COLRv1 where an arbitrary depth can be reached.


behdad:
I feel like this is overkill overhead. Can we add recursion counter only in COLRv1 when PaintColrGlyf is invoked? That's how we do it in eg GSUB/GPOS when Lookups are recursed to.

garretrieger:
> I feel like this is overkill overhead. Can we add recursion counter only in COLRv1 when PaintColrGlyf is invoked? That's how we do it in eg GSUB/GPOS when Lookups are recursed to.

The problem isn't in PaintColorGlyf since the sanitizer doesn't recurse via gid's. This is a case where the offset graph is deep enough to cause a stack overflow. So the problem could happen with any of the paint types that have an offset to another Paint, which I believe is quite a few.



> I feel like this is overkill overhead. Can we add recursion counter only in COLRv1 when PaintColrGlyf is invoked? That's how we do it in eg GSUB/GPOS when Lookups are recursed to.

Ok, switched to only do the recursion sanitization inside of COLR

src/hb-sanitize.hh
```diff
@@ -116,13 +116,17 @@
 #ifndef HB_SANITIZE_MAX_SUBTABLES
 #define HB_SANITIZE_MAX_SUBTABLES 0x4000
 #endif
+#ifndef HB_SANITIZE_MAX_DEPTH
+#define HB_SANITIZE_MAX_DEPTH 2500
```
behdad, line 120: Move this into COLR then. And I feel like 128 should be more than enough. That's a lot of stack!

There's already `COLRV1_MAX_NESTING_LEVEL`. Reuse it. Rename to `HB_COLR_MAX_NESTING_LEVEL` please.

garretrieger: Ah hadn't noticed that constant. Updated to reuse it.

PR #3285, opened 2021-11-06
Minor fixes

matthiasclasen:
A documentation clarification and a case of confusing formatting.

### DIFF
--- a/src/hb-ot-layout.cc
+++ b/src/hb-ot-layout.cc
@@ -717,10 +717,14 @@ hb_ot_layout_script_find_language (hb_face_t    *face,
  * @language_tags: The array of language tags
  * @language_index: (out): The index of the requested language
  *
- * Fetches the index of a given language tag in the specified face's GSUB table
- * or GPOS table, underneath the specified script index.
+ * Fetches the index of the first language tag fom @language_tags that is present
+ * in the specified face's GSUB or GPOS table, underneath the specified script
+ * index.
  *
- * Return value: %true if the language tag is found, %false otherwise
+ * If none of the given language tags is found, %false is returned and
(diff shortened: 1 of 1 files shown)


src/hb-ot-layout.cc
```diff
@@ -717,10 +717,15 @@ hb_ot_layout_script_find_language (hb_face_t    *face,
  * @language_tags: The array of language tags
  * @language_index: (out): The index of the requested language
  *
- * Fetches the index of a given language tag in the specified face's GSUB table
- * or GPOS table, underneath the specified script index.
+ * Fetches the index of a the first language tag that is present in the
```
khaledhosny, line 720:
```suggestion
 * Fetches the index of a the first language tag from @language_tags that is present in the
```

src/hb-ot-layout.cc
```diff
@@ -717,10 +717,15 @@ hb_ot_layout_script_find_language (hb_face_t    *face,
...
+ * specified face's GSUB or GPOS table, underneath the specified script
+ * index.
  *
- * Return value: %true if the language tag is found, %false otherwise
+ * If none of the given language tags is found, %false is returned and
+ * @language_index is set to either the index of the default language tag
+ * (`dflt`) or to `HB_OT_LAYOUT_DEFAULT_LANGUAGE_INDEX`.
```
khaledhosny, line 726: Not sure if we want to document the `dflt` tag part. Strictly speaking, there is no `dflt` language in OpenType and this code, IIUC, is working around broken fonts.

matthiasclasen:
Made some revisions. How about this?

khaledhosny (approved):
LGTM, thanks.

PR #3288, opened 2021-11-08
[subset] avoid writing out duplicate extra glyph names in post table

qxliu76:
Add check for possible duplicate with other name index


behdad:
This looks bad because `O(N^2)`. Does the input font have duplicate glyph names? Can't we sort and remove dups? Why not insert in a hash?

qxliu76:
> This looks bad because `O(N^2)`. Does the input font have duplicate glyph names? Can't we sort and remove dups? Why not insert in a hash?

GlyphName strings are stored in Pascal string format and the input font has duplicate glyph names with different indices. I'll try to insert a hash when checking for duplicates.

src/hb-ot-post-table-v2subset.hh
```diff
@@ -79,6 +112,7 @@ HB_INTERNAL bool postV2Tail::subset (hb_subset_context_t *c) const
   post::accelerator_t _post;
   _post.init (c->plan->source);
 
+  hb_hashmap_t<unsigned, hb_set_t *, (unsigned)-1, nullptr> len_to_gids;
```
garretrieger, line 115: Make this a map from hb_bytes_t* to the index of the string. hb_bytes_t has hash() defined so that should work without any additional changes. Then for each string you can check if it already exists, and grab it's index for re-use easily. Should simplify this change.

behdad:
This can be done properly after https://github.com/harfbuzz/harfbuzz/pull/3301 is merged.

src/hb-ot-post-table-v2subset.hh
```diff
@@ -90,22 +91,29 @@ HB_INTERNAL bool postV2Tail::subset (hb_subset_context_t *c) const
     else
     {
       hb_bytes_t s = _post.find_glyph_name (old_gid);
-      int standard_glyph_index = -1;
-      for (unsigned i = 0; i < format1_names_length; i++)
+      if (glyph_name_to_new_index.has (s))
+      { new_index = glyph_name_to_new_index.get (s); }
```
behdad, line 95: Instead of checking first, just try `get` and check for `-1`.

Issue #1463, opened 2018-12-07
UNSAFE_TO_CONCAT [Was Clarify UNSAFE_TO_BREAK and expose OS2V2Tail::usMaxContext as context.]


bungeman:
The current documentation clearly states what UNSAFE_TO_BREAK (or it's absence) can be interpreted to mean. However, it should also be documented as to what it is not, since it's very easy to be lured into a false sense of security with it.

The current behavior (as I understand it) is that one can shape a sequence of code points into glyphs, pick any two safe to break locations in the resulting buffer, reshape the associated sub-sequence of code points in isolation and get the same buffer output as is seen between the original two safe to break locations. Essentially, any pair of safe to break locations is stating the the larger context of the original sequence of code points did not affect the shaping of that run. (This is actually a bit stronger of a statement than the current documentation implies, please correct if this is not true.)

However, it is important to note that the inverse is not true, it is not the case that such a run of code points will always shape this way; any modification to the original sequence of code points can arbitrarily modify the overall shaping. For instance I can make a font which creates ligatures of all the sentences in the works of Shakespeare such that "There is a written scroll" will come back with many safe to break locations, but "There is a written scroll!" will shape into a single glyph which visually represents the icon for the Prince of Morocco. In other words, adding to or modifying any of the original sequence of code points invalidates everything about the shaping before doing so (including any safe to breaks).

In order to avoid requiring complete re-shaping when editing (especially simply adding to the end of a paragraph), it is necessary to let the user know the maximum affected area. For OpenType this is the purpose of OS2V2Tail::usMaxContext, in theory even TrueType knows the answer to this question for at least appending (keep track of the last ground state). With this sort of information it should be possible for a sophisticated user to only re-shape a portion of the overall sequence of code points by dirtying the maximum context area and finding the safe to break locations surrounding that area. The pathological Shakespeare font I propose above would obviously have quite a large value for the maximum context, but for most fonts this value is three or less.

In between these two cases (breaking at safe to break locations and mutating the sequence of code points) there is the case of determining the extent of re-shaping if truncating the original sequence of code points. Is it enough to find some safe to break location before the truncation and shape only the remaining portion? I cannot construct an example to answer in the negative, but it is not obvious that this is always allowable either. I think this is equivalent to asking if any shaping operation can operate by negative look ahead in a way which doesn't affect safe to break. With the maximum context information it seems this could be at least bounded, but it would be nice to know if this common case can be handled even more efficiently (may be common when line breaking).

[...]

PR #3297, opened 2021-11-16
Implement UNSAFE_TO_CONCAT flag

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/1463


behdad:
cc @drott @bungeman @raphlinus @matthiasclasen @khaledhosny

I gave this a go. Wasn't bad at all. Of course, I haven't gone through all shapers and reviewed them specifically. But I don't think this flag needs as much work as the `UNSAFE_TO_BREAK` flag does.

Documentation still needs more work re what guarantee this exactly provides and how to use it.

behdad:
I'll add details about how to test, etc. soon.

src/hb-buffer.h
```diff
@@ -112,15 +111,26 @@ typedef struct hb_glyph_info_t {
  * 				   reshaping to a small piece around the
  * 				   breaking point only, even if the breaking
  * 				   point carries the
- * 				   #HB_GLYPH_FLAG_UNSAFE_TOBREAK, by reshaping
- * 				   a the shortest segment from the closest
- * 				   safe-to-concat point.
+ * 				   #HB_GLYPH_FLAG_UNSAFE_TOBREAK or when
```
matthiasclasen, line 114: Is it really called UNSAFE_TOBREAK, or should that be UNSAFE_TO_BREAK ?

src/hb-buffer.h
```diff
@@ -112,15 +111,26 @@ typedef struct hb_glyph_info_t {
...
+ * 				   hyphenation or other text transformation
+ * 				   happens at line-break point, by reshaping
+ * 				   the segment from the closest
+ * 				   safe-to-concat point, checking whether the
+ * 				   resulting glyph-run also is safe-to-concat
+ * 				   at the point of interest, and if it is, just
+ * 				   append splict it into place. If not, back up
```
matthiasclasen, line 121: splict ?

src/hb-buffer.h
behdad: Fixed. Thanks.

src/hb-buffer.h
behdad: Fixing. Also, more doc updates coming.

behdad:
Okay, another docu rewrite. I *think* I got the guarantee and algorithm right this time. Please review.

matthiasclasen:
I can't really review the algorithm since I'm not familiar enough with hb internals and C++ is disorienting to me.

The docs make sense to me now.

It would be nice to have a testcase that checks some of the cases where UNSAFE_TO_BREAK or UNSAFE_TO_CONCAT
would be set. I guess for UNSAFE_TO_BREAK, that is just s|t (for a font with an st ligature). What is an example for UNSAFE_TO_CONCAT?

behdad:
Let's take a look at how this works.

With Material Icons font from:
https://fonts.google.com/icons?selected=Material+Icons+Outlined

which can be downloaded from:
https://github.com/google/material-design-icons/blob/master/font/MaterialIcons-Regular.ttf

This font has lots of long ligatures, for words like "anchor", or "hourglass_bottom", etc. So it's a great test for this feature.

With `hb-shape`, the `--show-flags` argument is of interest. flags are shown after `#`. And flag `1` is `UNSAFE_TO_BREAK`, flag `2` is `UNSAFE_TO_CONCAT`. A flag of `3` would imply both.

Let's look at shaping of the text "xxhourglass_botto". What I expect is that it tell me that it's unsafe-to-concat all the from from *after* `h` to the end of the string, because appending an `m` will form the `hourglass_bottom` ligature:
```
$ util/hb-shape --no-positions --show-flags MaterialIcons-Regular.ttf  hourglass_botto 
[gid21=0|gid28=1#2|gid34=2#2|gid31=3#2|gid20=4#2|gid25=5#2|gid14=6#2|gid32=7#2|gid32=8#2|gid13=9#2|gid15=10#2|gid28=11#2|gid33=12#2|gid33=13#2|gid28=14#2]
```
and indeed:
```
$ ninja && util/hb-shape --no-positions --show-flags MaterialIcons-Regular.ttf  hourglass_bottom 
[gid1369=0]
```

Note that the unsafe-to-break is missing here, because indeed there was no ligatures in the long sequence, so breaking was safe at any point, and in the short sequence there's only one glyph, so breaking either before or after it is still safe.

That's it for now.

behdad:
What is left is to add `unsafe_to_concat` marker around the rest of the codebase. It's hard to analyze though, where else it is needed. For example, in Arabic shaper, after any letter that is open to further joining, it's unsafe to concat. In the grammar-based shapers, anytime the cluster is open to further items it's technically unsafe to concat, but not if the shaping result isn't different. So...

Easiest way is to add a fuzzer test and fix up the failures... I'll see if I can put that together.

khaledhosny:
cc @zauguin

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -1175,8 +1192,14 @@ static inline bool match_backtrack (hb_ot_apply_context_t *c,
   skippy_iter.set_match_func (match_func, match_data, backtrack);
 
   for (unsigned int i = 0; i < count; i++)
-    if (!skippy_iter.prev ())
+  {
+    unsigned unsafe_from;
+    if (!skippy_iter.prev (&unsafe_from))
```
drott, line 1197: IIUC here, and below in line 1226 is the logic for setting this flag, but I don't have enough knowledge of the codebase to tell what this does. What's the underlying logic of when this is set? What existing internal state of this `skippy_iter` is used to determine when to set the flag? Is there some sort of closure/pre-analysis done of potential ligature/other-shaping-sequence completions?

src/hb-buffer.h
```diff
@@ -76,18 +76,63 @@ typedef struct hb_glyph_info_t {
...
+ * 				   way:
+ *
+ * 				   1. Iterate back from the line-break point till
+ * 				   the the first cluster start point that is
+ * 				   NOT unsafe-to-concat, 2. shape the segment from
+ * 				   there till the end of line, 3. check whether the
+ * 				   resulting glyph-run also is clear of the
+ * 				   unsafe-to-concat at its start-of-text position;
+ * 				   if it is, just splice it into place and the line
+ * 				   is shaped; If not, back up to a point further back
```
drott, line 121:
```suggestion
 * 				   is shaped; If not, move the line-break position candidate further back
```

Is that what's intended?

src/hb-buffer.h
```diff
@@ -76,18 +76,63 @@ typedef struct hb_glyph_info_t {
...
+ * 				   after line-breaking, by limiting the
+ * 				   reshaping to a small piece around the
+ * 				   breaking point only, even if the breaking
+ * 				   point carries the
+ * 				   #HB_GLYPH_FLAG_UNSAFE_TO_BREAK or when
+ * 				   hyphenation or other text transformation
+ * 				   happens at line-break point, in the following
+ * 				   way:
+ *
+ * 				   1. Iterate back from the line-break point till
```
drott, line 114:
```suggestion
 * 				   1. Iterate back from the line-break position candidate till
```

To make it consistent with the suggestion below.

src/hb-buffer.h
```diff
@@ -76,18 +76,63 @@ typedef struct hb_glyph_info_t {
...
+ * 				   the the first cluster start point that is
+ * 				   NOT unsafe-to-concat, 2. shape the segment from
+ * 				   there till the end of line, 3. check whether the
+ * 				   resulting glyph-run also is clear of the
+ * 				   unsafe-to-concat at its start-of-text position;
+ * 				   if it is, just splice it into place and the line
+ * 				   is shaped; If not, back up to a point further back
+ * 				   that is clear of unsafe-to-concat and retry.
+ *
+ * 				   At the start of next line a similar algorithm can
```
drott, line 124: I am trying to grasp this paragraph. Does this mean for computing a good line break that keeps line-breaking free of any shaping changes, one would have to find a candidate both by checking unsafe-to-concat "join compatibility" with the current line as described above, then check the segment after the candidate position for compatibility with the next line - and that's where the problem arises of being unable to check unsafe-to-concat state on the end of the run?

drott (commented):
This is great! Thank you for adding the logic for unsafe-to-concat. Similarly to @matthiasclasen I don't feel I know the code well enough to review the logic thorughly, but I tried to take a closer look. Some questions on the logic, and some suggestions regarding the documentation below. Thank you!

src/hb-buffer.h
behdad: No, not the line-break position candidate. Just the reshape-from "safe-to-concat" candidate. I'll clarify.

src/hb-buffer.h
behdad: We first shape the full paragraph. Then our aim is to reuse the shaping result of the paragraph to produce shaping result for each line. But the line is being broken and we need to patch the line start / end for proper shaping results, right? The algorithm I documented explains what to do for the end of line. But when you start the next line, you need to reshape the beginning of the line. That's where you need to find a safe-to-concat point and splice in a piece. Does that make sense?

behdad: > That's where you need to find a safe-to-concat point and splice in a piece. Does that make sense?

Only if the line-break position wasn't safe-to-break.

behdad: I clarified the documentation a bit. Please check.

src/hb-ot-layout-gsubgpos.hh
behdad: No pre-analysis. Just every time it tries to match a ligature and fails, it marks up to the failure point as unsafe-to-concat.

Eg. if text is "hourglaxxyzdkjtlak" and the ligature we're trying to match is "hourglass_bottom", the matcher matches up the sequence "hourgla" and gives up upon seeing "x". It then marks unsafe-to-concat up to there, because up to the x, if something changes, there's a change of a ligature forming.

That's basically what it's doing.

behdad:
Okay, added buffer verification of the flag and I see 23 tests, which point to places that need more adjustment to the flag. Good start:
```
Summary of Failures:

 74/392 harfbuzz:shaping+in-house / arabic-fallback-shaping      FAIL             0.06s   exit status 1
 79/392 harfbuzz:shaping+in-house / arabic-stch                  FAIL             0.05s   exit status 1
 80/392 harfbuzz:shaping+in-house / cluster                      FAIL             0.06s   exit status 1
 87/392 harfbuzz:shaping+in-house / default-ignorables           FAIL             0.07s   exit status 1
103/392 harfbuzz:shaping+in-house / kern-format2                 FAIL             0.06s   exit status 1
104/392 harfbuzz:shaping+in-house / item-context                 FAIL             0.06s   exit status 1
105/392 harfbuzz:shaping+in-house / indic-vowel-letter-spoofing  FAIL             0.07s   exit status 1
117/392 harfbuzz:shaping+in-house / none-directional             FAIL             0.06s   exit status 1
123/392 harfbuzz:shaping+in-house / simple                       FAIL             0.07s   exit status 1
158/392 harfbuzz:shaping+aots / gpos_chaining1_simple            FAIL             0.06s   exit status 1
164/392 harfbuzz:shaping+aots / gpos_chaining2_simple            FAIL             0.06s   exit status 1
168/392 harfbuzz:shaping+aots / gpos_chaining3_simple            FAIL             0.06s   exit status 1
201/392 harfbuzz:shaping+aots / gpos3                            FAIL             0.07s   exit status 1
202/392 harfbuzz:shaping+aots / gpos4_lookupflag                 FAIL             0.07s   exit status 1
213/392 harfbuzz:shaping+aots / gsub_chaining1_simple            FAIL             0.06s   exit status 1
219/392 harfbuzz:shaping+aots / gsub_chaining2_simple            FAIL             0.06s   exit status 1
224/392 harfbuzz:shaping+aots / gsub_chaining3_simple            FAIL             0.06s   exit status 1
293/392 harfbuzz:shaping+text-rendering-tests / HVAR-1           FAIL             0.07s   exit status 1
324/392 harfbuzz:shaping+text-rendering-tests / MORX-37          FAIL             0.07s   exit status 1
325/392 harfbuzz:shaping+text-rendering-tests / MORX-38          FAIL             0.06s   exit status 1
326/392 harfbuzz:shaping+text-rendering-tests / MORX-39          FAIL             0.07s   exit status 1
328/392 harfbuzz:shaping+text-rendering-tests / MORX-40          FAIL             0.06s   exit status 1
331/392 harfbuzz:shaping+text-rendering-tests / MORX-5           FAIL             0.07s   exit status 1
```

behdad:
Humm. Kerning is going to be a problem. As in, any glyph that is beginning of a kern pair will cause an unsafe-to-concat flag.

behdad:
Failures down to:
```
103/392 harfbuzz:shaping+in-house / kern-format2                 FAIL             0.06s   exit status 1
112/392 harfbuzz:shaping+in-house / mongolian-variation-selector FAIL             0.07s   exit status 1
158/392 harfbuzz:shaping+aots / gpos_chaining1_simple            FAIL             0.06s   exit status 1
164/392 harfbuzz:shaping+aots / gpos_chaining2_simple            FAIL             0.07s   exit status 1
170/392 harfbuzz:shaping+aots / gpos_chaining3_simple            FAIL             0.07s   exit status 1
201/392 harfbuzz:shaping+aots / gpos3                            FAIL             0.07s   exit status 1
202/392 harfbuzz:shaping+aots / gpos4_lookupflag                 FAIL             0.07s   exit status 1
258/392 harfbuzz:shaping+aots / gsub4_1_multiple_ligatures       FAIL             0.06s   exit status 1
262/392 harfbuzz:shaping+aots / lookupflag_ignore_attach         FAIL             0.07s   exit status 1
264/392 harfbuzz:shaping+aots / lookupflag_ignore_combination    FAIL             0.07s   exit status 1
```

src/hb-buffer.h
```diff
@@ -76,18 +76,67 @@ typedef struct hb_glyph_info_t {
...
+ * 				   after line-breaking, by limiting the
+ * 				   reshaping to a small piece around the
+ * 				   breaking positin only, even if the breaking
+ * 				   position carries the
+ * 				   #HB_GLYPH_FLAG_UNSAFE_TO_BREAK or when
+ * 				   hyphenation or other text transformation
+ * 				   happens at line-break position, in the following
+ * 				   way:
+ *
+ * 				   1. Iterate back from the line-break position till
```
drott, line 114: Also, I am coming back to trying to understand this paragraph, and I don't entirely reach clarity.

a) In this algorithm, do we aim to adjust the line break position forwards or backwards? Or do we only aim to reshape so that isolated runs are correctly shaped?

b) By, "splice it into place" do you mean "replace previous shape results with the new result that was shaped in isolation"?

So with this algorithm, we ensure that we grab enough precontext to have long-context rules applied correctly despite line break truncation or end of line hyphenation character additions, right? Perhaps it'd make sense to state that before "1."?

src/hb-buffer.h
drott: > At the start of next line a similar algorithm can be implemented.

I still don't get this line. What does "similar" mean here if we spelled it out. What do we want to achieve for the beginning of the line in terms of safe-to-concat "clean"-ness.

src/hb-buffer.h
behdad: a) The line-break position does not change; We just aim to produce shaping result for the line, ideally without reshaping the entire line: by reusing most of the line shape results from the paragraph shape results and patching up the line start/end shaping only, using information from this new bit.

b) correct.

behdad:
> I still don't get this line. What does "similar" mean here if we spelled it out. What do we want to achieve for the beginning of the line in terms of safe-to-concat "clean"-ness.

Yeah, I'll try to spell out the algorithm.

If at the position of line-break, we have unsafe-to-break flag on, or if transformation happened, then we cannot just reuse the shaping result from the broken paragraph results. So we need to either reshape the entire line, or patch up the results. That's what we need to achieve.

src/hb-buffer.h
```diff
@@ -76,18 +76,67 @@ typedef struct hb_glyph_info_t {
...
+ * 				   the the first cluster start position that is
+ * 				   NOT unsafe-to-concat, 2. shape the segment from
```
kojiishi, line 116: Can you elaborate a bit more how UNSAFE_TO_BREAK can't do what's explained here, but UNSAFE_TO_CONCAT can? I was under an impression that UNSAFE_TO_BREAK is for line breaking, and UNSAFE_TO_CONCAT is for editing/appending.

Currently, we use UNSAFE_TO_BREAK for the line breaking purpose, and use heuristic rules for when editing/appending. I can understand UNSAFE_TO_CONCAT is useful when editing/appending, including when appending hyphenation while line braking, but in this step 1, I thought UNSAFE_TO_BREAK is the flag to use.

It'd be really helpful if the comment can cover in what cases using UNSAFE_TO_BREAK in step 1 can create a wrong result.

behdad: > Can you elaborate a bit more how UNSAFE_TO_BREAK can't do what's explained here, but UNSAFE_TO_CONCAT can? I was under an impression that UNSAFE_TO_BREAK is for line breaking, and UNSAFE_TO_CONCAT is for editing/appending.

Hi @kojiishi. Please see https://github.com/harfbuzz/harfbuzz/issues/1463#issuecomment-448081494

Basically, if the point of line-break is UNSAFE_TO_BREAK, then you need this UNSAFE_TO_CONCAT for optimized line-breaking. You cannot use the UNSAFE_TO_BREAK for optimized line-breaking at that point.

behdad:
I'm getting very close with this. The chaining logic is solid I believe. I have 9 fringe tests failing:
```
104/400 harfbuzz:shaping+in-house / indic-vowel-letter-spoofing  FAIL             0.07s   exit status 1
106/400 harfbuzz:shaping+in-house / kern-format2                 FAIL             0.07s   exit status 1
114/400 harfbuzz:shaping+in-house / mongolian-variation-selector FAIL             0.07s   exit status 1
203/400 harfbuzz:shaping+aots / gpos3                            FAIL             0.06s   exit status 1
204/400 harfbuzz:shaping+aots / gpos4_lookupflag                 FAIL             0.06s   exit status 1
260/400 harfbuzz:shaping+aots / gsub4_1_multiple_ligatures       FAIL             0.06s   exit status 1
264/400 harfbuzz:shaping+aots / lookupflag_ignore_attach         FAIL             0.06s   exit status 1
266/400 harfbuzz:shaping+aots / lookupflag_ignore_combination    FAIL             0.06s   exit status 1
295/400 harfbuzz:shaping+text-rendering-tests / HVAR-1           FAIL             0.06s   exit status 1
```

behdad:
Tests pass! @jfkthame would you be able to review this please?

What's left is annotations in Hangul and Thai shapers.

src/hb-buffer.hh
```diff
@@ -385,15 +385,80 @@ struct hb_buffer_t
...
+  void unsafe_to_break_from_outbuffer (unsigned int start = 0, unsigned int end = -1)
+  {
+    set_glyph_flags (HB_GLYPH_FLAG_UNSAFE_TO_BREAK | HB_GLYPH_FLAG_UNSAFE_TO_CONCAT,
+		     start = start, end = end,
+		     true, true);
+  }
+  void unsafe_to_concat_from_outbuffer (unsigned int start = 0, unsigned int end = -1)
+  {
+    set_glyph_flags (HB_GLYPH_FLAG_UNSAFE_TO_CONCAT,
+		     start = start, end = end,
```
dscorbett, line 459: This isn’t named parameter syntax: it’s just reassigning `start` and `end` to themselves.

behdad: Oops. I thought I fixed. Thanks.

behdad:
> What's left is annotations in Hangul and Thai shapers.

I reviewed those shapers and I don't think there's any work that needs to be done.

behdad:
@SergeyMalkin @raphlinus @mikeday This is ready. Any review is appreciated. Thanks.

src/hb-buffer.h
behdad: Just added this:
https://github.com/harfbuzz/harfbuzz/pull/3297/commits/540af548dc3926c4b9db059c9b20297df0802671

behdad:
I'm merging this just so the code wouldn't rot and dependencies can move forward. I still would appreciate review of the API semantics, documentation, as well as the implementation. Thanks.

mikeday (commented):
Looks sensible!

src/hb-buffer.h
drott: Thanks, yes, I find that more clear.

src/hb-buffer.hh
```diff
@@ -67,8 +67,8 @@ enum hb_buffer_scratch_flags_t {
   HB_BUFFER_SCRATCH_FLAG_HAS_DEFAULT_IGNORABLES		= 0x00000002u,
   HB_BUFFER_SCRATCH_FLAG_HAS_SPACE_FALLBACK		= 0x00000004u,
   HB_BUFFER_SCRATCH_FLAG_HAS_GPOS_ATTACHMENT		= 0x00000008u,
-  HB_BUFFER_SCRATCH_FLAG_HAS_UNSAFE_TO_BREAK		= 0x00000010u,
-  HB_BUFFER_SCRATCH_FLAG_HAS_CGJ			= 0x00000020u,
+  HB_BUFFER_SCRATCH_FLAG_HAS_CGJ			= 0x00000010u,
```
drott, line 70: What's CGJ? Could that be spelled out or would that be too long?

src/hb-buffer.hh
```diff
@@ -385,15 +385,80 @@ struct hb_buffer_t
...
+	unsigned cluster = _infos_find_min_cluster (info, idx, end);
+	cluster = _infos_find_min_cluster (out_info, start, out_len, cluster);
+
+	_infos_set_glyph_flags (out_info, start, out_len, cluster, mask);
+	_infos_set_glyph_flags (info, idx, end, cluster, mask);
+      }
+    }
+  }
+
+  void unsafe_to_break (unsigned int start = 0, unsigned int end = -1)
```
drott, line 438: Maybe `mark_unsafe_to_break` or `set_unsafe_to_break` for this and the following three? When reading this next to `clear_flags()` the latter describes what the function does with a verb, this one here could be a getter or a setter from its name.

src/hb-buffer.hh
```diff
@@ -385,15 +385,80 @@ struct hb_buffer_t
...
+		     start, end,
+		     true);
+  }
+  void unsafe_to_concat (unsigned int start = 0, unsigned int end = -1)
+  {
+    set_glyph_flags (HB_GLYPH_FLAG_UNSAFE_TO_CONCAT,
+		     start, end,
+		     true);
+  }
+  void unsafe_to_break_from_outbuffer (unsigned int start = 0, unsigned int end = -1)
```
drott, line 450: I don't understand what `from_outbuffer` means here. Could that be explained in a comment? I only see the `set_glyph_flags` (true, true) flags difference, but that's not quite clear in this context.

util/shape-options.hh
```diff
@@ -269,6 +269,186 @@ struct shape_options_t
...
+     *    shaping the two buffers, we still get cluster boundaries at the
+     *    segment boundaries, and that those all are safe-to-concat points.
+     *    Moreover, that there are NOT any safe-to-concat points within the
+     *    segments.
+     *
+     * 4. Finally, we reconstruct the shaping results of the original text by
+     *    simply interleaving the shaping results of the segments from the two
+     *    buffers, and assert that the total shaping results is the same as
+     *    the one from original buffer in step 1.
+     */
```
drott, line 307: Nice explanations, that's very helpful.

util/shape-options.hh
```diff
@@ -269,6 +269,186 @@ struct shape_options_t
...
+	  printf("start %d end %d text start %d end %d\n", start, end, text_start, text_end);
+
+#if 0
+	hb_buffer_flags_t flags = hb_buffer_get_flags (fragment);
+	if (0 < text_start)
+	  flags = (hb_buffer_flags_t) (flags & ~HB_BUFFER_FLAG_BOT);
+	if (text_end < num_chars)
+	  flags = (hb_buffer_flags_t) (flags & ~HB_BUFFER_FLAG_EOT);
+	hb_buffer_set_flags (fragment, flags);
+#endif
```
drott, line 365: Not sure what HarfBuzz' practice is for deactivated code. In Blink I would suggest to either move to some deactivatable logging macros, or Skia uses some DEBUGCODE macros IIRC. `if 0` or `#if 0` looks a bit brittle.

util/shape-options.hh
```diff
@@ -269,6 +269,186 @@ struct shape_options_t
...
+     *    shaping results of each segment.  As such, we expect that after
+     *    shaping the two buffers, we still get cluster boundaries at the
+     *    segment boundaries, and that those all are safe-to-concat points.
+     *    Moreover, that there are NOT any safe-to-concat points within the
+     *    segments.
+     *
+     * 4. Finally, we reconstruct the shaping results of the original text by
+     *    simply interleaving the shaping results of the segments from the two
+     *    buffers, and assert that the total shaping results is the same as
+     *    the one from original buffer in step 1.
```
drott, line 306: Would it be a useful addition to interleave them shifted as well (if the original is `1odd-2even-3odd-4even`, then interleave them as `2even-1odd-4even-3odd`, and compare original shaping and reconstructed shaping results for that set as well? This creates some additional new "boundary situations" similar to the folding/pulling together in 2.

drott (commented):
Really not well versed in this code, but a few high level comments. I like the verification approach, is that already run on all existing tests?

src/hb-buffer.hh
khaledhosny: https://en.wikipedia.org/wiki/Combining_grapheme_joiner, but it is a misnomer so spelling it out would be equally confusing, IMO.

src/hb-buffer.hh
behdad: Added comment in `set_glyph_flags`.

PR #3299, opened 2021-11-18
[subset] keep features that have FeatureParams and the tag is "size"

qxliu76:
This is to match fonttools:
https://github.com/fonttools/fonttools/commit/1df4f1eac1fa185e1b838e1203c92d67fcc41067


garretrieger (approved):

behdad:
The change looks good but your summary is misleading.

Better summary would be: "keep empty features with FeatureParams only if feature tag is 'size'".

PR #3300, opened 2021-11-19
[subset] inputSequence could be empty, remove the sanity check

qxliu76:


garretrieger (approved):

behdad:
Why are you doing this? An input length of 0 is invalid.

behdad:
In other words, input sequence cannot be empty.

qxliu76:
> In other words, input sequence cannot be empty.

In the spec:

inputSequence is starting from the second glyph, defined as below in the spec:
uint16  inputSequence[glyphCount - 1]: Sequence of classes to be matched to the input glyph sequence, beginning with the second glyph position

In this testcase, glyphCount=1, so the inputSequence is empty. If we don't allow empty inputSequence, this Rule is dropped in the subset, but fonttools keeps this rule

behdad:
Oh, then the fix is to change the check (and move it one line up), to be:
```
    if (unlikely (!inputCount)) return_trace (false);
```

behdad:
lgtm. thanks.

PR #3302, opened 2021-11-20
[meson] Require 0.55.0

khaledhosny:
We implicitly require it for building ragel subproject. This new version requirement should satisfied in both Fedora 33 and Debian bullseye, and not be too cutting edge for us.


khaledhosny:
Or may be it is too cutting edge for our CI. The oss-fuzz build should be fixed with https://github.com/google/oss-fuzz/pull/6876, but I don’t know why the compiler is crashing in the other failing build.

behdad:
Ooh. Thanks for splitting the sanitizer job!

PR #3311, opened 2021-11-30
Improve GSUB/GPOS packing by sharing nodes despite differing virtual links

garretrieger:
Ligature subtables use virtual links to enforce an ordering constraint between the subtables and the coverage table. Unfortunately this has the side effect of preventing the subtables from being shared by another Ligature with a different coverage table since object equality compares all links real and virtual. This change stores virtual links separately from real links and updates the equality check to only check real links. If an object is de-duped any virtual links it has are merged into the object that replaces it to maintain the desired ordering constraints.

Also introduces hb_concat (a, b) which produces an iterator that iterates over all of a and then all of b.


src/hb-iter.hh
```diff
@@ -581,6 +581,79 @@ struct
...
+    else
+      ++b;
+  }
+
+  void __forward__ (unsigned n)
+  {
+    if (n > a.len ()) {
+      a.__forward__ (a.len ());
+      n -= a.len ();
+    }
```
behdad, line 632: This block doesn't look right to me.

behdad (commented):
Thanks. Left one comment.

src/hb-iter.hh
garretrieger: oh good catch, a.len() will be incorrect since we jumped forward. I'll fix and add a test for forward.

src/hb-iter.hh
```diff
@@ -627,10 +627,12 @@ struct hb_concat_iter_t :
   void __forward__ (unsigned n)
   {
     if (n > a.len ()) {
-      a.__forward__ (a.len ());
       n -= a.len ();
+      a.__forward__ (a.len ());
+      b.__forward__ (n);
```
behdad, line 632: I don't like that you are calling `a.len()` three times. It might be `O(n)`.

garretrieger: Makes sense. I add an alternate implementation when the iterator is not random access.

PR #3315, opened 2021-12-02
Consider continuations before reversing buffer in Chain::apply

### LINKED ISSUE #3314: Directionality of regional-indicator flags in for AAT font
This is a follow up of issue #2265 

The problem has been fixed by pull #3018  but still happens for some fonts on macos.

See https://bugs.chromium.org/p/chromium/issues/detail?id=1272436#c8 for the bug report on chromium.

I have investigated, and I believe that the root cause of the issue in the following code reversing the buffer without considering continuations: https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-aat-layout-morx-table.hh#L1040

Because of that, the regional indicators are reversed before substitution.

arenevier:
Do not reverse continuations.

Introduce a helper method reverse_preserve_graphemes()

Fixes #3314

### DIFF
--- a/src/hb-aat-layout-morx-table.hh
+++ b/src/hb-aat-layout-morx-table.hh
@@ -1037,13 +1037,15 @@ struct Chain
       if (!c->buffer->message (c->font, "start chainsubtable %d", c->lookup_index))
 	goto skip;
 
-      if (reverse)
-	c->buffer->reverse ();
+      if (reverse) {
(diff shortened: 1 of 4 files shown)


behdad:
Thanks. I need more time analyzing this. But yeah, I was going for `reverse_graphemes` as well.

behdad:
Fixed in main. Please test.

PR #3318, opened 2021-12-02
[subset] (Chain)ContextSubst glyph_closure fix

qxliu76:
- When pos_glyphs is empty, use current full glyphs set as input for
subsequent recursive closure process
See also: 
https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/subset/__init__.py#L1069
https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/subset/__init__.py#L1274
- Also increase max_lookup_indices to 35000 cause a real font file hit
previous limit 20000 and some lookups are dropped unexpectedly


behdad:
> https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/subset/__init__.py#L1199
> https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/subset/__init__.py#L992

These lines don't look relevant to me. Did you mean those?

behdad:
> * Also increase max_lookup_indices to 25000 cause a real font file hit
>   previous limit 20000 and some lookups are dropped unexpectedly

This means this kind of counting is wrong. What are we counting exactly?

qxliu76:
> > https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/subset/__init__.py#L1199
> > https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/subset/__init__.py#L992
> 
> These lines don't look relevant to me. Did you mean those?

oops sorry, should be these 2 lines:
https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/subset/__init__.py#L1069
https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/subset/__init__.py#L1274

qxliu76:
> > * Also increase max_lookup_indices to 25000 cause a real font file hit
> >   previous limit 20000 and some lookups are dropped unexpectedly
> 
> This means this kind of counting is wrong. What are we counting exactly?

IIRC this counting is just to avoid an out-of-memory fuzzer crash. Lookup closure is a recursive process, and that fuzzer case might have too many nested lookup references.

qxliu76:
> > * Also increase max_lookup_indices to 25000 cause a real font file hit
> >   previous limit 20000 and some lookups are dropped unexpectedly
> 
> This means this kind of counting is wrong. What are we counting exactly?

Uh, maybe setting the maximum nesting level instead is better? But that still needs to set a certain limit

behdad:
> Uh, maybe setting the maximum nesting level instead is better? But that still needs to set a certain limit

Yes, that's what I mean. We are counting the wrong quantity.

garretrieger:
> > Uh, maybe setting the maximum nesting level instead is better? But that still needs to set a certain limit
> 
> Yes, that's what I mean. We are counting the wrong quantity.

https://github.com/harfbuzz/harfbuzz/commit/834a224a507bc8b971b280483de3f067a709611f added the lookup limit. This is effectively an operation count which so far I've found to be the most effective way to mitigate the fuzzer timeouts. Recursion depth checks on their own aren't usually sufficient.

Normally the sanitizer operation count check would be enough to catch these issues, but this case is special since the recursion happens via lookup indices instead of offsets. Thus the sanitizer can't catch it.

garretrieger:
> > > Uh, maybe setting the maximum nesting level instead is better? But that still needs to set a certain limit
> > 
> > 
> > Yes, that's what I mean. We are counting the wrong quantity.
> 
> [834a224](https://github.com/harfbuzz/harfbuzz/commit/834a224a507bc8b971b280483de3f067a709611f) added the lookup limit. This is effectively an operation count which so far I've found to be the most effective way to mitigate the fuzzer timeouts. Recursion depth checks on their own aren't usually sufficient.
> 
> Normally the sanitizer operation count check would be enough to catch these issues, but this case is special since the recursion happens via lookup indices instead of offsets. Thus the sanitizer can't catch it.

We should probably rename the constant at minimum to better describe what the limit is. This is not  a limit of the number of lookups in a font like it implies.

behdad:
> oops sorry, should be these 2 lines:
> https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/subset/__init__.py#L1069
> https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/subset/__init__.py#L1274

Why are we not using `pos_glyphs == nullptr` like the Python version? Also, I don't see the equivalent `chaos` tracking.

qxliu76:
> > oops sorry, should be these 2 lines:
> > https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/subset/__init__.py#L1069
> > https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/subset/__init__.py#L1274
> 
> Why are we not using `pos_glyphs == nullptr` like the Python version? Also, I don't see the equivalent `chaos` tracking.

pos_glyphs==nullptr is special case which happens in ContextSubst tables only.
If we check for pos_glyphs == nullptr,  that will happen at the beginning of glyph closure for each Lookup. In harfbuzz we could just use dispatch() method in "struct Lookup" for glyph closure, so I did not add another closure() method which 
basically does the same thing as dispatch() with additional one line checking for nullptr. 
And 'chaos' is tracked here: https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-ot-layout-gsubgpos.hh#L1266

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -1305,15 +1313,17 @@ static void context_closure_recurse_lookups (hb_closure_context_t *c,
     hb_set_add (covered_seq_indicies, seqIndex);
     if (pos_glyphs)
       c->push_cur_active_glyphs (pos_glyphs);
+    else
+      c->push_cur_active_glyphs (c->glyphs);
```
behdad, line 1317:
```
  c->push_cur_active_glyphs (pos_glyphs ? pos_glyphs : c->glyphs);
```

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -1305,15 +1313,17 @@ static void context_closure_recurse_lookups (hb_closure_context_t *c,
...
     if (context_format == ContextFormat::CoverageBasedContext)
       endIndex += 1;
 
     c->recurse (lookupRecord[i].lookupListIndex, covered_seq_indicies, seqIndex, endIndex);
 
+    c->pop_cur_done_glyphs ();
     if (pos_glyphs) {
-      c->pop_cur_done_glyphs ();
       hb_set_destroy (pos_glyphs);
     }
```
behdad, line 1328: Remove unneeded brackets.

behdad:
Looks much better!

Issue #3316, opened 2021-12-02
hb-subset fails on empty glyphs


dscorbett:
Subsetting certain single characters, such as U+1BC11 DUPLOYAN LETTER TH, from [this font](https://github.com/dscorbett/duployan-test/blob/4449d6351da8a32219556b2f44f8fa2d38aaa6de/assets/fonts/Duployan-Regular.otf) fails.
```
$ curl -LOs https://github.com/dscorbett/duployan-test/raw/4449d6351da8a32219556b2f44f8fa2d38aaa6de/assets/fonts/Duployan-Regular.otf
$ hb-subset --output-file=subset.otf --unicodes=1bc11 Duployan-Regular.otf
error: Operation failed. Probably a bug. File github issue.
```
U+1BC11 is derived from U+1BC03 plus a dot, like how in Arabic ⟨ذ⟩ is derived from ⟨د⟩. In this font, derived letters get glyphs with no contours and 0 advance width. A lookup then decomposes each one into a plain letter plus one or more dots. The subsetter seems to fail when given a range containing any such letter.

garretrieger:
Checked this and it's a failure of the offset overflow resolution for GSUB. I'll look into it and see what we can do.

garretrieger:
Ok, so what's happening is there's ~3700 extension lookups which are all connected by shared nodes and so end up all in a single space. The resolver is splitting them off from the shared space one at a time but gives up after the current 10 iteration limit.

When there's this many lookups sharing a single space splitting them off one at a time is likely going to be a very inefficient way of finding a working resolution. I see two possible approaches:

1. If after the max iterations have been reached there's still an overflow within a space that can be further split, go nuclear and split each lookup into it's own space.
2. When the number of lookups in a shared space is very large, don't peel off lookups one at a time, instead repeatedly split the space(s) in half.

behdad:
> The resolver is splitting them off from the shared space one at a time

What's the current algorithm?

I wonder if there's a good heuristic there... I can ask my theory friends.

PR #3322, opened 2021-12-06
[repacker] Move all overflowing roots to a new space simultaneously.

garretrieger:
Previous algorithm was to split off roots to their own space one at a time. Instead on each iteration grab the set of all roots that contain at least one overflow and move them to their own space.

Helps with fonts with large numbers of lookups sharing a single space such as in #3316

In the fonts I've tested with the resolver now concludes in a single iteration instead of multiple iterations and produces a final packed size even smaller than the previous implementation.

Additionally this fixes two other bugs I encountered:
- Reduce the number of bits used for 'order' in the modified distance. Under the previous 22 bits the  41 bit distance number was overflowing with fonts that had more than ~500 spaces.
- Have the repacker compute the final buffer size and allocate it, instead of relying on the caller to provide one. In cases where the repacked font expands significantly it was possible to overflow the provided buffer.

### DIFF
--- a/src/hb-repacker.hh
+++ b/src/hb-repacker.hh
@@ -204,27 +204,46 @@ struct graph_t
   /*
    * serialize graph into the provided serialization buffer.
    */
-  void serialize (hb_serialize_context_t* c) const
+  hb_blob_t* serialize () const
   {
-    c->start_serialize<void> ();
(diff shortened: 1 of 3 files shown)


behdad:
Work out art. I'll review soon. Thanks Garrett.

Issue #3325, opened 2021-12-09
[metrics] Sanitize values


behdad:
Currently `hb-ot-metrics` API returns false if table providing a metric is not present. But returns true even if the fetched values are nonsensical. See:

  https://gitlab.gnome.org/GNOME/pango/-/issues/616

I suggest we add some minimal sanity-check to it.

behdad:
Classic question is: what does Windows do with such values?

behdad:
What @matthiasclasen and I are thinking currently is a version of this API named `_with_fallback`, that doesn't return a boolean, and synthesizes missing values internally, including rejecting bad values. Many clients just need a number to plug into their algorithms and cannot accept a missing value. Instead of each client dealing with a missing value, we can centralize that logic. I'll sketch that.

matthiasclasen:
https://github.com/harfbuzz/harfbuzz/pull/3442 has an initial version of that api

PR #3326, opened 2021-12-09
[metrics] Check some values for non-zero

behdad:
Part of https://github.com/harfbuzz/harfbuzz/issues/3325

### DIFF
--- a/src/hb-ot-metrics.cc
+++ b/src/hb-ot-metrics.cc
@@ -138,6 +138,7 @@ hb_ot_metrics_get_position (hb_font_t           *font,
 			    hb_ot_metrics_tag_t  metrics_tag,
 			    hb_position_t       *position     /* OUT.  May be NULL. */)
 {
+  hb_position_t position_tmp;
   hb_face_t *face = font->face;
   switch ((unsigned) metrics_tag)
   {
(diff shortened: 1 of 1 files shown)


behdad:
Something like this, but for more values. @matthiasclasen @khaledhosny @jfkthame

khaledhosny:
The problem with this kind of sanitization is it prevents the font from using said value if it has a legitimate use of doing so. A Y (or X) position of 0 is valid, and suppose for some reason the font wants the underline at the baseline, we will be overriding it. A width of 0 can be less sensible, but it can be a way to say “underlines make no sense for this design, don’t draw them.”

I’d rather go hand wavy here and say this is what the font wants and if the font had these values by mistake, the font should be fixed.

As type designer, these kind of decisions done by software always surprise you when you least expect them and are always almost impossible to workaround (for example, Firefox clamps the underline position to font descender, so I can’t have under line that goes below the descender and keep the line spacing I think is sensible, even though there should be no relationship between underline position and font descender).

behdad:
@khaledhosny I'm with you generally. But I'm asking for a more nuanced decision here.

Pango has been rejecting such zero values for years (before Matthias changed it recently hence the bug). If clients do it, there's no point in us not making it easy for them.

If you think zero underline position is legit, we can allow it. I'm not suggesting that we return true and synthesize values. So, a zero underline thickness is still detectable as "not present". Just would be more obviously distinguishable from "thickness is present" and handled with the "value not present" branch, making it more likely to be handled right, for whatever right is for the client.

jfkthame:
I'm with Khaled here (AIUI) ... if the font explicitly has zero for a value such as underline position or thickness, that's what the designer is asking for and that's what they should get.

> So, a zero underline thickness is still detectable as "not present".

I don't think I understand what you mean here. "Not present" is a different state from "Present, with value zero", but this appears to be collapsing that distinction.

matthiasclasen:
> I'm with Khaled here (AIUI) ... if the font explicitly has zero for a value such as underline position or thickness, that's what the designer is asking for and that's what they should get.

And then we get bugs filed against pango that say "underlines are broken with font foo". They don't get filed against the font with the opinionated designer...

jfkthame:
> > I'm with Khaled here (AIUI) ... if the font explicitly has zero for a value such as underline position or thickness, that's what the designer is asking for and that's what they should get.
> 
> And then we get bugs filed against pango that say "underlines are broken with font foo". They don't get filed against the font with the opinionated designer...

If pango is responsible for drawing underlines, and wants to override font metrics to force them to always be visible, that's something it can consider. (Maybe even let the pango client opt in/out of that behavior via a flag somewhere?) But I don't think it's appropriate for low-level harfbuzz API that exposes font metrics to be imposing that decision on behalf of all clients.

behdad:
Sure, I'll add as new API.

behdad:
So, I like to add a separate API that fetches font metrics, with fallback.

Can I please get help seeing what Pango, Firefox, and Chrome do to synthesize some of these values? @matthiasclasen @jfkthame @drott

matthiasclasen:
Here is pango:

```
  if (hb_ot_metrics_get_position (hb_font, HB_OT_METRICS_TAG_UNDERLINE_SIZE, &position) &&
      position != 0)
    metrics->underline_thickness = position;
  else
    metrics->underline_thickness = PANGO_SCALE;

  if (hb_ot_metrics_get_position (hb_font, HB_OT_METRICS_TAG_UNDERLINE_OFFSET, &position) &&
      position != 0)
    metrics->underline_position = position;
  else
    metrics->underline_position = - PANGO_SCALE;

  if (hb_ot_metrics_get_position (hb_font, HB_OT_METRICS_TAG_STRIKEOUT_SIZE, &position) &&
      position != 0)
    metrics->strikethrough_thickness = position;
  else
    metrics->strikethrough_thickness = PANGO_SCALE;

  if (hb_ot_metrics_get_position (hb_font, HB_OT_METRICS_TAG_STRIKEOUT_OFFSET, &position) &&
      position != 0)
    metrics->strikethrough_position = position;
  else
    metrics->strikethrough_position = metrics->ascent / 2;
```

matthiasclasen:
I've added code to check some values for being non-zero in https://github.com/harfbuzz/harfbuzz/pull/3442

PR #3339, opened 2021-12-20
Inherit slant in sub fonts

matthiasclasen:
Without this, slant will be lost in the sub fonts
that pango creates for passing to hb_shape.


behdad:
Thanks. I folded it into my PR.

Issue #3340, opened 2021-12-21
Consider using a dedicated replacement glyph for missing spaces


matthiasclasen:
Currently, harfbuzz will replace all sorts of missing spaces by proper amounts of nothing. 
But it will __only__ do so if the font has an ASCII space glyph. I figure this is because decomposition can only happen if you have something to decompose into. But it forces pango to fake a 0x20 space glyph for icon fonts in order to benefits from hb' space smarts. 
If you would use a dedicated replacement glyph for missing spaces, you could probably provide smart space replacements for all spaces including 0x20.

behdad:
We should use the buffer invisible-glyph for these, and we can do so for ASCII space itself as well.

PR #3341, opened 2021-12-21
Use invisible-glyph for spaces if font has no ASCII space

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3340

Should add tests ideally.

### DIFF
--- a/src/hb-ot-shape-fallback.cc
+++ b/src/hb-ot-shape-fallback.cc
@@ -525,6 +525,15 @@ _hb_ot_shape_fallback_spaces (const hb_ot_shape_plan_t *plan HB_UNUSED,
   for (unsigned int i = 0; i < count; i++)
     if (_hb_glyph_info_is_unicode_space (&info[i]) && !_hb_glyph_info_ligated (&info[i]))
     {
+      /* If font had no ASCII space and we used the invisible glyph, give it a 1/4 EM default advance. */
+      if (buffer->invisible && info[i].codepoint == buffer->invisible)
+      {
+        if (horizontal)
+	  pos[i].x_advance = +font->x_scale / 4;
+        else
+	  pos[i].y_advance = -font->y_scale / 4;
+      }
+
       hb_unicode_funcs_t::space_t space_type = _hb_glyph_info_get_unicode_space_fallback_type (&info[i]);
       hb_codepoint_t glyph;
       typedef hb_unicode_funcs_t t;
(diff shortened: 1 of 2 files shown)


behdad:
This changes the behavior in that non-ASCII spaces will prefer the invisible glyph over ASCII space if invisible is set. I can change it to retain old behavior. Maybe I should. @jfkthame any input?

behdad:
> This changes the behavior in that non-ASCII spaces will prefer the invisible glyph over ASCII space if invisible is set. I can change it to retain old behavior. Maybe I should. @jfkthame any input?

I changed the logic to keep the current behavior.

behdad:
What this still doesn't do is deleting spaces if there's no ASCII space in the font. Not sure if I want to bother with that. Client should just set an invisible glyph IMO.

src/hb-ot-shape-fallback.cc
```diff
@@ -525,6 +525,13 @@ _hb_ot_shape_fallback_spaces (const hb_ot_shape_plan_t *plan HB_UNUSED,
   for (unsigned int i = 0; i < count; i++)
     if (_hb_glyph_info_is_unicode_space (&info[i]) && !_hb_glyph_info_ligated (&info[i]))
     {
+      /* If font had no ASCII space and we used the invisible glyph, give it a 1/4 EM default advance. */
+      if (buffer->invisible && info[i].codepoint == buffer->invisible)
```
matthiasclasen, line 529: This gets a warning here:

../src/hb-ot-shape-fallback.cc:529:10: warning: suggest explicit braces to avoid ambiguous ‘else’ [-Wdangling-else]
  529 |       if (buffer->invisible && info[i].codepoint == buffer->invisible)

src/hb-ot-shape-fallback.cc
```diff
@@ -525,6 +525,13 @@ _hb_ot_shape_fallback_spaces (const hb_ot_shape_plan_t *plan HB_UNUSED,
...
+        if (horizontal)
+	  pos[i].x_advance = +font->x_scale / 4;
+        else
+	  pos[i].y_advance = -font->y_scale / 4;
+
```
matthiasclasen, line 534: If I read this correctly, you are giving  0x20 a width of em/4 here, while https://docs.microsoft.com/en-us/typography/develop/character-design-standards/whitespace says it should be em/2.

Indeed, in testing, the result of this code is half as wide as what my current pango code produces.

jfkthame: IMO em/2 sounds unusually large compared to what most fonts use.

Indeed, the MS page mentioned actually says

> For an *average* width font a good value is ~1/4 the em.

matthiasclasen: ah, ok. Lets go with that then.

behdad:
@matthiasclasen does it work for pango?

behdad:
To recap, missing-spaces-shaping was added in harfbuzz 1.1; the invisible-glyph property was added in harfbuzz 2.0. At the time I forgot to wire up invisible-glyph to missing-spaces shaping. This change does that.

matthiasclasen:
> @matthiasclasen does it work for pango?

Yes, works fine

PR #3350, opened 2022-01-02
docs: Add some details about coordinates

matthiasclasen:
Mention what the effective values are when no coordinates have been set.

### DIFF
--- a/src/hb-font.cc
+++ b/src/hb-font.cc
@@ -631,7 +631,7 @@ hb_font_funcs_destroy (hb_font_funcs_t *ffuncs)
  * @destroy: (nullable): A callback to call when @data is not needed anymore
  * @replace: Whether to replace an existing data with the same key
  *
- * Attaches a user-data key/data pair to the specified font-functions structure. 
+ * Attaches a user-data key/data pair to the specified font-functions structure.
  *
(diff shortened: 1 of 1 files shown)


src/hb-font.cc
```diff
@@ -2236,10 +2236,10 @@ hb_font_get_var_coords_normalized (hb_font_t    *font,
  *
  * Return value: coordinates array
  *
- * Since: REPLACEME
+ * Since: 3.3.0
```
behdad, line 2247: This gets filled in at release time based on the chosen version.

behdad:
May I suggest an improvement? It's possible that the caller of the set_design_coords / set_normalized_coords sets partial coords. In that case, any unset coords get the 0 / default coords. Can you rewrite to say "any unset coords get 0 / default value"?

src/hb-font.cc
matthiasclasen: oh, nice. I should copy that trick

matthiasclasen:
Still not quite right. The function just returns what has been set by the previous set_var_coords call, which may be a partial array, and must be interpreted as zero-padded in the normalized case and as default-value-padded in the design case.

Issue #1555, opened 2019-01-24
Add `hb_buffer_create_similar()`


behdad:
Right now one can create new buffer and append existing buffer to it, but in utils I also have this piece:
```
  static void copy_buffer_properties (hb_buffer_t *dst, hb_buffer_t *src) 
  { 
    hb_segment_properties_t props; 
    hb_buffer_get_segment_properties (src, &props); 
    hb_buffer_set_segment_properties (dst, &props); 
    hb_buffer_set_flags (dst, hb_buffer_get_flags (src)); 
    hb_buffer_set_cluster_level (dst, hb_buffer_get_cluster_level (src)); 
  } 
```

behdad:
The usecase in utils needs to copy just a range of the buffer.  That's done using `hb_buffer_append` which takes a start/end.  So, maybe we just need a function that copies the non-content properties like above.  Just need to pick a name for it.

ebraminio:
I like something like Java's `.clone()` wording, `hb_buffer_clone` to return a new and modifiable buffer. My usecase was something like to copy a non-shaped buffer, shape it, add Kashida glyph to places I like, which I know and render it using `hb-render`, which I know, is supercilious

behdad:
Clone wouldn't address the need in util.

behdad:
> The usecase in utils needs to copy just a range of the buffer. That's done using `hb_buffer_append` which takes a start/end. So, maybe we just need a function that copies the non-content properties like above. Just need to pick a name for it.

What is also not address by `hb_buffer_append` is the effect on pre/post-context.

behdad:
> What is also not address by `hb_buffer_append` is the effect on pre/post-context.

And that's https://github.com/harfbuzz/harfbuzz/issues/1843

behdad:
Getting ideas from cairo: `hb_buffer_create_similar()`. I like that.

behdad:
> hb_buffer_create_similar()

One problem with it: since, unfortunately, `hb_buffer_clear_contents()` also clears segment-properties, there's a need for two different two different functions. Maybe one doesn't need to be a function:

- `hb_buffer_create_similar()` should copy everything other than contents, content-type, and segment-properties.

- contents and content-type are handled by `hb_buffer_append`. Maybe segment-properties, if not set (individually, script, language, direction), should also be set by `hb_buffer_append`? Then the code in util will become automatic.

behdad:
[...]

PR #3353, opened 2022-01-03
Add `hb_buffer_create_similar()`; use it in util

behdad:
What I like about this is that it clarifies what's buffer setup and what's buffer content.

Fixes https://github.com/harfbuzz/harfbuzz/issues/1555


behdad:
I did some more cleaning on this branch. I really like it. Am merging it so I can rebase UNSAFE_TO_CONCAT on top.

behdad:
Still interested in review though. Thank you.

PR #3357, opened 2022-01-06
Avoid unintentional locale dependency

matthiasclasen:
Not sure what you do about threadsafe initialization and singletons
in harfbuzz, but this is the gist of the fix.

---

hb_variation_to_string uses sprintf with %g,
which will produce a locale-dependent decimal
point, which is not desired here.

The output is supposed to be compatible with
CSS syntax, and that always uses '.' for the
decimal point.

Fix this by changing the per-thread locale
to "C" around sprintf call.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -61,6 +61,7 @@ check_headers = [
   ['unistd.h'],
   ['sys/mman.h'],
   ['stdbool.h'],
+  ['xlocale.h'],
 ]
 
 check_funcs = [
@@ -70,6 +71,8 @@ check_funcs = [
   ['getpagesize'],
   ['mmap'],
   ['isatty'],
+  ['uselocale'],
+  ['newlocale'],
 ]
 
 m_dep = cpp.find_library('m', required: false)
--- a/src/hb-common.cc
+++ b/src/hb-common.cc
@@ -1047,6 +1047,10 @@ hb_variation_from_string (const char *str, int len,
 #if !defined(HARFBUZZ_NO_SETLOCALE) && defined(HAVE_NEWLOCALE) && defined(HAVE_USELOCALE)
 static hb_atomic_ptr_t <void> C_locale;
 
+#ifdef WIN32
+using locale_t = _locale_t;
+#endif
+
 static void
 free_C_locale (void)
 {
@@ -1074,6 +1078,9 @@ get_C_locale (void)
   return (locale_t) locale;
 }
 #else
+#ifdef WIN32
+#define locale_t void*
+#endif
 #define uselocale(Locale) ((locale_t)0)
 #endif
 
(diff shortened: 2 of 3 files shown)


matthiasclasen:
It seems I don't know how to make this compile on Windows :(

src/hb-common.cc
```diff
@@ -1055,6 +1074,8 @@ void
 hb_variation_to_string (hb_variation_t *variation,
 			char *buf, unsigned int size)
 {
+  locale_t oldlocale;
```
behdad, line 1077: Since we're C++, move this to first use site.

meson.build
```diff
@@ -70,6 +71,8 @@ check_funcs = [
   ['getpagesize'],
   ['mmap'],
   ['isatty'],
+  ['uselocale'],
+  ['newlocale'],
```
behdad, line 75: Please add these, and the header check, to `configure.ac` as well. I know.

src/hb-common.cc
```diff
@@ -1039,6 +1043,21 @@ hb_variation_from_string (const char *str, int len,
...
 }
 
+#if !defined(HARFBUZZ_NO_SETLOCALE) && defined(HAVE_NEWLOCALE) && defined(HAVE_USELOCALE)
+static locale_t
+get_C_locale (void)
+{
+  static locale_t C_locale = NULL;
+
+  if (!C_locale)
+    C_locale = newlocale (LC_ALL_MASK, "C", NULL);
```
behdad, line 1053: Assuming `locale_t` is of pointer size, we would use cmpexch for this initialization. And we have `hb_lazy_loader_t` to do the cruft. Check `hb-ft.cc` for how to use it for a type we think is a pointer underneath:
```
tatic struct hb_ft_library_lazy_loader_t : hb_lazy_loader_t<hb_remove_pointer<FT_Library>,
                                                             hb_ft_library_lazy_loader_t>
{
...
}...;
```

behdad:
Isn't `xlocale` deprecated and removed?

behdad:
Oh is the xlocale for windows?

matthiasclasen:
> Oh is the xlocale for windows?

I added the xlocale include back because it was needed to build on macos

src/hb-common.cc
matthiasclasen: will have a look, thanks

matthiasclasen: I did not find any guarantees in POSIX about the size or or characteristics of locale_t, but at least in glibc, its a pointer to a struct.

i gave the cmpexch approach a try.

matthiasclasen:
not sure what to do about the compiler warnings that break the various Windows builds

jfkthame:
> not sure what to do about the compiler warnings that break the various Windows builds

For the error here:
```
../src/hb-common.cc: In function ‘void hb_variation_to_string(hb_variation_t*, char*, unsigned int)’:
../src/hb-common.cc:1116:24: error: statement has no effect [-Werror=unused-value]
 1116 |   uselocale (oldlocale);
```
does it help if you explicitly cast the result of `uselocale` to `void`?

matthiasclasen:
> does it help if you explicitly cast the result of `uselocale` to `void`?

That did the trick. Thanks!

behdad:
The initialization still is not bullet-proof.  Using `hb_lazy_loader_t` saves you the cmpexch gymnastics. I'm doing it for you, hold on.

PR #3358, opened 2022-01-08
Fix unintentional locale dependency

behdad:
Supersedes https://github.com/harfbuzz/harfbuzz/pull/3357

### DIFF
--- a/src/hb-common.cc
+++ b/src/hb-common.cc
@@ -1050,32 +1050,43 @@ static hb_atomic_ptr_t <void> C_locale;
 using locale_t = _locale_t;
 #endif
 
-static void
-free_C_locale (void)
+static inline void free_static_C_locale ();
+
+static struct hb_C_locale_lazy_loader_t : hb_lazy_loader_t<hb_remove_pointer<locale_t>,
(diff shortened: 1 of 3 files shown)


matthiasclasen:
Looks fine to me. Didn't know about this lazy loader thing

matthiasclasen (approved):

src/hb-common.cc
```diff
@@ -1039,6 +1043,56 @@ hb_variation_from_string (const char *str, int len,
   return false;
 }
 
+#if !defined(HARFBUZZ_NO_SETLOCALE) && defined(HAVE_NEWLOCALE) && defined(HAVE_USELOCALE)
```
drott, line 1046: Should this be `HB_NO_SETLOCALE`?

IIUC, if I set, `HB_NO_SETLOCALE`, it looks like I still need to include a `locale.h` or `xlocale.h` on Linux and Mac respectively in order to find a `locale_t` even though the calls are a no-op? Can that be avoided?

behdad: Fixed. Thanks.

drott:
Thank you!

PR #3361, opened 2022-01-11
Use freetype from CMake target when present

tksuoran:
Add check for Freetype CMake target. If it is found:
- enable HB_HAVE_FREETYPE option
- add HAVE_FREETYPE=1 define
- bypass finding Freetype as cmake package
- add Freetype CMake target as dependency to harfbuzz


tksuoran:
The main purpose of this change is to enable using freetype and harfbuzz from a CMake project using CMake `FetchContent()`. In this case., there is no need for (dev) packages to be installed on the system, which is great for example on Windows.

tksuoran:
Thanks!

Issue #2140, opened 2020-01-29
Reordering fails when GDEF table is absent


simoncozens:
(See this [typedrawers](https://typedrawers.com/discussion/3490/prebase-consonants-in-the-use#latest) thread)

I can see there's code in the USE shaper to note down which glyphs are involved in pref substitutions and then reorder them, but I can't get it to fire. I created a font by adding this feature code:

```
feature pref {
    script java;
    sub uniA9BF by glyph01;
} pref;
```

to my [Fallback Plus](https://github.com/simoncozens/test-fonts/blob/master/FallbackPlus-Regular.otf) font, and tried to reorder using the pref feature, but this was the output:

```
$ hb-shape Fallback-Javanese.otf -u 'A995 A9BF'
[uniA995=0+600|glyph01=0+600]
```

I would have expected `glyph 01 uniA995`. Adding some instrumentation (OK, print statements) into `_set_glyph_props` and `record_pref_use` I got this:

```
$ ./util/hb-shape --script='java' --shapers=ot ~/hacks/typography/test-fonts/Fallback-Javanese.otf -u 'A995 A9BF'
Choosing...
Choose USE.
Setting glyph props
Checking pref use...
Checking glyph 591 (2)...
Checking glyph 101 (2)...
[uniA995=0+600|glyph01=0+600]
```

So when the substitution is done, the `HB_OT_LAYOUT_GLYPH_PROPS_SUBSTITUTED` is added to the second glyph, but by the time `record_pref_use` is called, the flag has been cleared.

I feel like this is a bug, but I don't have enough USE experience to be confident.

punchcutter:
I downloaded that font and there's no GSUB in it. Which means there's also no script/lang tags to trigger USE shaping.

simoncozens:
There’s no features in that font. I added the feature above to it.

simoncozens:
Font with added feature:
[Test-Javanese.zip](https://github.com/harfbuzz/harfbuzz/files/4129962/Test-Javanese.zip)

punchcutter:
That zip gives me a 0kb file, but I added the feature to the other font and everything works fine here.
hb-shape (HarfBuzz) 2.6.4
[glyph01=0+600|uniA995=0+600]
![java_test](https://user-images.githubusercontent.com/11257204/73387314-3e358580-4285-11ea-8fe8-d1ae02651c11.png)

simoncozens:
[...]

PR #3365, opened 2022-01-12
GDEF fix

behdad:
This is another try to fix https://github.com/harfbuzz/harfbuzz/issues/2140

Similar to commit f4cd99f28eefe695b86876736166fd119e22e04e
Which was reverted in 4a1eef737ac528228a736144a644f43f80bd0c8a

Previous attempt had introduced https://github.com/harfbuzz/harfbuzz/issues/2516, so was reverted. I'm not sure why that was.

@simoncozens any chance you can test?

cc @khaledhosny


khaledhosny:
I committed the tests from #2655 and they seem to pass.

khaledhosny (approved):

PR #3366, opened 2022-01-13
Convert fallback kwargs to [provide] entries.

jpakkane:
This converts the "old style" `fallback` keyword arguments to use `[provide]` entries in wrap files instead. This makes it more convenient to use Harfbuzz as a subproject in other projects. Due to backwards compatibility issues having those `fallback` entries may cause unwanted subproject promotions. This is also more compatible with getting dependencies from WrapDB instead.

We are currently discussing how wrap files should be handled in release tarballs. Maybe they should be removed altogether, especially those that do Git checkouts. But that is a decision for another time.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -90,15 +90,12 @@ if not get_option('freetype').disabled()
   if not freetype_dep.found()
     # https://github.com/harfbuzz/harfbuzz/pull/2498
     freetype_dep = dependency('freetype2', required: get_option('freetype'),
-                              fallback: ['freetype2', 'freetype_dep'],
                               default_options: ['harfbuzz=disabled'])
   endif
 endif
 
-glib_dep = dependency('glib-2.0', required: get_option('glib'),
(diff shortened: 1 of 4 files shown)


behdad:
Thanks Jussi. I assume this doesn't require a newer version of meson?

jpakkane:
All of these features have been in for a long time so probably no. Though you might want to test that your own dev workflows still work, this might change how and when dep lookups and downloads work.

behdad:
Our bots seem happy. @khaledhosny wdyt?

eli-schwartz:
> Thanks Jussi. I assume this doesn't require a newer version of meson?

It is available since 0.55.0, and older versions of meson would still be able to build the project -- they would just lose the ability to recognize the subproject fallback.

However, it's irrelevant since you already require that version of meson either way:
https://github.com/harfbuzz/harfbuzz/blob/bc0a5fdf43c7932c9ca39fa64a5cd405ebb43da5/meson.build#L1-L2

PR #3369, opened 2022-01-13
Fix docs build

matthiasclasen:
Fix the docs build.


behdad:
Fixed manually. Thanks.

PR #3371, opened 2022-01-13
[vector] Move semantics when resizing

behdad:
Compiler catches cases where types are not safe to do so. We should fix these. Most are in CFF code but a few are not.

We should fix this in main branch and rebase this on top till it's happy, then this branch can continue to make vector follow C++ memory semantics.

### DIFF
--- a/src/hb-ot-color-cbdt-table.hh
+++ b/src/hb-ot-color-cbdt-table.hh
@@ -360,6 +360,16 @@ struct IndexSubtable
 
 struct IndexSubtableRecord
 {
+  /* XXX Remove this and fix by not inserting it into vector. */
+  IndexSubtableRecord& operator = (const IndexSubtableRecord &o)
+  {
+    firstGlyphIndex = o.firstGlyphIndex;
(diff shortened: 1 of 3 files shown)


behdad:
Fortunately my code was wrong. When I fixed it, the errors are much fewer.

behdad:
I think this is ready to be merged as is. The rest (construction / destruction) needs a LOT more work, and I'll open a new PR for that.

behdad:
I'm working on the fuzzer failures.

garretrieger:
> I'm working on the fuzzer failures.

UBSAN failure is happening because fini() is being called twice here: https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-subset-cff1.cc#L399 and then by the maps destructor.

fini_shallow() sets item to a nullptr but doesn't clear mask so when fini_shallow() runs a second time it tries to destruct elements the nullptr items in https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-map.hh#L136

behdad:
I think I rebased on main but fuzzers still unhappy. Help appreciated. Thanks Garret!

PR #3377, opened 2022-01-16
Fix various typos

luzpaz:
Found via `codespell -q 3 -S ./perf/texts -L actualy,ba,beng,fo,gir,inout,nd,ot,pres,ro,te,teh,timne`


behdad:
Thank you.

PR #3383, opened 2022-01-18
Handle ragel better with meson

### LINKED ISSUE #3378: Harfbuzz is broken with --fatal-meson-warnings
https://github.com/harfbuzz/harfbuzz/blob/6c81cd9543f35cbc54e938b402355990a3b0859c/src/meson.build#L294-L302

This isn't a "warning", IMO. If you want to emit alarming warnings at people who are building harfbuzz releases, rather than developing on harfbuzz, can you please add an option to disable this `warning()`?

This is currently an inconvenience when trying to add harfbuzz to the Meson WrapDB: https://github.com/mesonbuild/wrapdb/pull/285

CI fails on Windows, where you cannot install packages, and fails on Linux too by default, and the only workaround is to compile a custom copy of ragel which seems like it should not be necessary.

eli-schwartz:
Just like autotools, ragel should only be required when rl sources are actually out of date.

Fixes #3378

### DIFF
--- a/.github/workflows/linux-ci.yml
+++ b/.github/workflows/linux-ci.yml
@@ -50,7 +50,8 @@ jobs:
           -Dgraphite=enabled \
           -Doptimization=2 \
           -Ddoc_tests=true \
-          -Dragel_subproject=true
+          -Dragel=enabled \
+          -Dwrap_mode=default
     - name: Build
       run: meson compile -Cbuild
     - name: Test
--- a/src/gen-ragel-artifacts.py
+++ b/src/gen-ragel-artifacts.py
@@ -4,22 +4,32 @@
 
 import os, os.path, sys, subprocess, shutil
 
-ragel = sys.argv[1]
-if not ragel:
-	sys.exit ('You have to install ragel if you are going to develop HarfBuzz itself')
+uninstalled_error = """
+'ragel' is missing on your system. In order to develop HarfBuzz itself,
+specifically, by editing the *.rl source files, you have to install ragel
+in order to transpile them to C++ headers.
+""".strip()
 
-if len (sys.argv) < 4:
+ragel = sys.argv[1]
+
+if len (sys.argv) != 5:
 	sys.exit (__doc__)
 
-OUTPUT = sys.argv[2]
+hh = sys.argv[2]
 CURRENT_SOURCE_DIR = sys.argv[3]
-INPUT = sys.argv[4]
+rl = sys.argv[4]
 
-outdir = os.path.dirname (OUTPUT)
-shutil.copy (INPUT, outdir)
-rl = os.path.basename (INPUT)
-hh = rl.replace ('.rl', '.hh')
-subprocess.Popen (ragel.split() + ['-e', '-F1', '-o', hh, rl], cwd=outdir).wait ()
+outdir = os.path.dirname (hh)
+hh_in_src = os.path.join (CURRENT_SOURCE_DIR, os.path.basename (hh))
 
-# copy it also to src/
-shutil.copyfile (os.path.join (outdir, hh), os.path.join (CURRENT_SOURCE_DIR, hh))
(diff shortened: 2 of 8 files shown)


src/gen-ragel-artifacts.py
```diff
@@ -4,22 +4,29 @@
...
+""".strip()
+
 ragel = sys.argv[1]
-if not ragel:
-	sys.exit ('You have to install ragel if you are going to develop HarfBuzz itself')
 
-if len (sys.argv) < 4:
-	sys.exit (__doc__)
+if len(sys.argv) != 5:
+	sys.exit(__doc__)
```
khaledhosny, line 17: Please keep the existing code style, we are using it consistently in this project.

.github/workflows/linux-ci.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       run: sudo apt-get install pkg-config gcc gcovr gtk-doc-tools libfreetype6-dev libglib2.0-dev libcairo2-dev libicu-dev libgraphite2-dev python3 python3-setuptools ninja-build gobject-introspection libgirepository1.0-dev
     - run: sudo pip3 install fonttools meson==0.56.0
     - name: run
-      run: meson build -Db_coverage=true --auto-features=enabled -Dgraphite=enabled -Dchafa=disabled -Dragel_subproject=true -Doptimization=2
+      run: meson build -Db_coverage=true --auto-features=enabled -Dgraphite=enabled -Dchafa=disabled -Dragel=disabled -Doptimization=2
```
khaledhosny, line 19: We no longer test the ragel subproject on ci builds, or am I missing something?

eli-schwartz: The project options specify `wrap_mode=nofallback`

behdad:
Why is the fedora bot failing? Should we instruct it to install ragel then?

khaledhosny:
> Why is the fedora bot failing? Should we instruct it to install ragel then?

We shouldn’t need ragel on any CI jobs (unless we are specifically testing ragel generation, which i snot the case here). That being said, we do actually install ragel on this job but Fedara has the new ragel that we don’t want. This IMHO means the changes in this PR isn’t quite working yet because users shouldn’t need to do anything if they are simply building HarfBuzz and suitable ragel is not present.

src/meson.build
```diff
@@ -291,26 +291,22 @@ hb_gobject_headers = files(
...
-    hh = rl.split('.')[0] + '.hh'
-    custom_target('@0@'.format(hh),
-      build_by_default: true,
-      input: rl,
-      output: hh,
-      command: [ragel_helper, ragel, '@OUTPUT@', meson.current_source_dir(), '@INPUT@'],
-    )
-  endforeach
+  message('You have to install ragel if you are going to develop HarfBuzz itself')
+  ragel = 'error'
```
xclaesse, line 298: Why this?

src/meson.build
```diff
@@ -291,26 +291,22 @@ hb_gobject_headers = files(
...
 endif
+ragel_helper = find_program('gen-ragel-artifacts.py')
+foreach rl : hb_base_ragel_sources
```
xclaesse, line 301: I think this part should remain in the `else` case of `if not has_ragel`, that's what make the CI fail AFAIK.

behdad:
Is someone going to finish this?

eli-schwartz:
I lost track of this PR but I am still interested in getting it to work.

behdad:
> I lost track of this PR (sorry!) but I am still interested in getting it to work.

Any chance you can finish this?

behdad:
What's the status of this?

eli-schwartz:
Sorry I forgot about this again. :( I have just rebased it including some tweaks for the CI.

behdad:
Let's decide one way or another here.

behdad:
Closing unless someone wants to finish this.

PR #3385, opened 2022-01-19
meson: Enable big objects support

chouquette:
Fix cross compilation to win64

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -57,6 +57,12 @@ if host_machine.cpu_family() == 'arm' and cpp.alignment('struct { char c; }') !=
   endif
 endif
 
+if host_machine.system() == 'windows'
+  add_project_arguments(cpp.get_supported_arguments([
+    '-Wa,-mbig-obj'
+  ]), language : 'cpp')
+endif
+
 check_headers = [
   ['unistd.h'],
   ['sys/mman.h'],


chouquette:
For reference, this is what this MR fixes: https://code.videolan.org/videolan/vlc/-/jobs/903124

meson.build
```diff
@@ -57,6 +57,10 @@ if host_machine.cpu_family() == 'arm' and cpp.alignment('struct { char c; }') !=
   endif
 endif
 
+if host_machine.system() == 'windows' and cpp.get_id() != 'msvc'
+  add_project_arguments(['-Wa,-mbig-obj'], language : 'cpp')
+endif
```
khaledhosny, line 62: Why is this only for compilers other than MSVC? Also the check seems to assume any compiler that is not MSVC is GCC-like, not sure if this is a right assumption to make.

chouquette: Different than MSVC, but only when compiling for Windows. In my experience that usually means gcc/llvm, but that's a valid point.

I can refactor the change to add it whenever host_machine is windows, and if the compiler supports it, if that's ok with you

khaledhosny: Yes, please.

chouquette: Done

Issue #1944, opened 2019-08-29
Merge hb_array_t / hb_sorted_array_t?


behdad:
Ala https://github.com/harfbuzz/harfbuzz/issues/1937
Ie. make the difference a boolean in template arg?

Or maybe not worth it.  Can't figure out.  Some methods need duplicating currently, and that's annoying.

PR #3389, opened 2022-01-19
Unify sorted array

behdad:
A failed attempt at https://github.com/harfbuzz/harfbuzz/issues/1944


behdad:
I don't believe this is possible. I'll merge some of the code but it's not completely possible.

Issue #2647, opened 2020-08-19
Lookup flags in fonts with no GDEF table


khaledhosny:
Someone reports [an issue](https://bugs.documentfoundation.org/show_bug.cgi?id=135778) in LibreOffice about some Iranian fonts like B Nazanin (fonts attached in the LIbreOffice issue).

Basically the font has mark ligatures for shadda and the lookups has ignore base and ignore ligatures set (which makes no sense), which causes HarfBuzz to correctly skip bases and make ligature between shadda and fathatan in strings like:
```
$ hb-view "B Nazanin.TTF"  اتّفاقاً
```
![HarfBuzz](https://user-images.githubusercontent.com/93914/90577567-9ac8a580-e1c1-11ea-807e-767504e88006.png)
Oddly enough, both Uniscribe and Core Text ignore the lookup flags and don’t for the ligatures.
```
$ hb-view "B Nazanin.TTF"  اتّفاقاً --shaper=coretext
```
![Core Text](https://user-images.githubusercontent.com/93914/90577550-871d3f00-e1c1-11ea-84f2-f5866deb706e.png)

As it turns out, the font has no `GDEF` table, so HarfBuzz synthesizes one and use it here.

It is a rather odd font, and I HarfBuzz synthesis of a GDEF table actually matches what [the spec](https://docs.microsoft.com/en-us/typography/opentype/spec/gdef#glyph-class-definition-table-overview) is saying (emphasis mine):
> In addition, a client uses class definitions to apply GSUB and GPOS LookupFlag data correctly. For example, a LookupFlag may specify ignoring ligatures and marks during a glyph operation. **If the font does not include a GlyphClassDef table, the client must define and maintain this information when using the GSUB and GPOS tables**.

But it might be an interoperability issue that we might want to consider (my 2¢ is to 🤷🏽  and declare it working as designed).

behdad:
Sounds related to the GDEF change we made and reverted. The carrying-forward of the flags is wrong...
Anyway. I'll check later. This was just me speculating based on the title.

khaledhosny:
That is with 2.7.1 where the offending commit is reverted, unless you mean the old behavior was also wrong.

khaledhosny:
Hmm, if I added a `GDEF` table with glyph classes, HarfBuzz stops making the ligatures: [BNazanin#2.zip](https://github.com/harfbuzz/harfbuzz/files/5093421/BNazanin.2.zip)

[...]

PR #3390, opened 2022-01-19
[layout] Don't skip over bases

behdad:
Do not respect IgnoreBaseGlyphs and IgnoreLigatures when matching
input, lookahead, and backtrack.

Makes five tests fail; four of them in aots. One in our mark-attachment.
I think this matches what other shapers do though. We should test more.

Fixes https://github.com/harfbuzz/harfbuzz/issues/2647

### DIFF
--- a/src/hb-ot-layout-gsubgpos.hh
+++ b/src/hb-ot-layout-gsubgpos.hh
@@ -419,7 +419,12 @@ struct hb_ot_apply_context_t :
 
     void set_ignore_zwnj (bool ignore_zwnj_) { ignore_zwnj = ignore_zwnj_; }
     void set_ignore_zwj (bool ignore_zwj_) { ignore_zwj = ignore_zwj_; }
-    void set_lookup_props (unsigned int lookup_props_) { lookup_props = lookup_props_; }
+    void set_lookup_props (unsigned int lookup_props_)
+    {
+      /* The following two flags are not respected by the matcher.
+       * https://github.com/harfbuzz/harfbuzz/pull/3390 */
+      lookup_props = lookup_props_ & ~(LookupFlag::IgnoreBaseGlyphs | LookupFlag::IgnoreLigatures);
+    }
     void set_mask (hb_mask_t mask_) { mask = mask_; }
     void set_syllable (uint8_t syllable_)  { syllable = syllable_; }
     void set_match_func (match_func_t match_func_,


behdad:
The aots tests I don't worry about. The other test we can check on other shapers to decide if this change sounds right or not.

behdad:
@jfkthame Does this sound reasonable to you?

jfkthame:
I guess this is reasonable, if it matches what uniscribe does.... the spec description of what these flags really do seems woefully inadequate, though!

behdad:
@khaledhosny Do you have access to test this on Uniscribe by any chance?

khaledhosny:
I’ll try to check the failing tests on Windows.

dscorbett:
My Duployan font relies on `IgnoreLigatures`. Does this PR mean that these two lookup flags would no longer have any effect anywhere?

behdad:
> My Duployan font relies on `IgnoreLigatures`. Does this PR mean that these two lookup flags would no longer have any effect anywhere?

I believe so. Can you test your font with/without this, and on other shapers?

dscorbett:
It seems to now be broken in the same way it is broken with Core Text. Well, I can probably recategorize all the ligature glyphs as mark glyphs and make it work somehow.

With the current PR, each call to `set_lookup_props` needs to clear the two flags. That means that any future code that sets lookups properties also needs to clear them, which someone might forget to do. Would it be better to, instead of clearing the flags in a few places, keep the flags but make them do nothing?

behdad:
> With the current PR, each call to `set_lookup_props` needs to clear the two flags. That means that any future code that sets lookups properties also needs to clear them, which someone might forget to do. Would it be better to, instead of clearing the flags in a few places, keep the flags but make them do nothing?

The flags speed up the matching of the initial glyph. They only do nothing when matching subsequent glyphs of a sequence.

behdad:
> The flags speed up the matching of the initial glyph. They only do nothing when matching subsequent glyphs of a sequence.

That is the intention of whoever designed it in uniscribe. Ie. the logic intends it to be implemented this way. shurg.

behdad:
You are right that it's better to do ignore those in the matcher then. I'll try that.

behdad:
Alternate fix pushed.

khaledhosny:
For the failing `mark-attachment` test, both `directwrite` and `uniscribe` HarfBuzz shapers give the same old output (I couldn’t install the font to test  with some other Windows program). The `coretext` shaper give the same glyph string but with different advances.

HarfBuzz without this change:
```
$ hb-shape ../test/shape/data/in-house/fonts/98b7887cff91f722b92a8ff800120954606354f9.ttf -u U+100F,U+103C,U+102F,U+1036
[uni103C102F=0+150|uni100F=0+550|uni1036=0@-150,0+0]
```

`directwrite`:
```
$ hb-shape.exe ../test/shape/data/in-house/fonts/98b7887cff91f722b92a8ff800120954606354f9.ttf -u U+100F,U+103C,U+102F,U+1036 --shaper=directwrite
[uni103C102F=0+150|uni100F=1+550|uni1036=2@-150,0+0]
```

`coretext`:
```
$ hb-shape ../test/shape/data/in-house/fonts/98b7887cff91f722b92a8ff800120954606354f9.ttf -u U+100F,U+103C,U+102F,U+1036 --shaper=coretext
[uni103C102F=0+150|uni100F=0+400|uni1036=0+150]
```

behdad:
Humm. I'm inclined to close this and leave things as is.

behdad:
Okay so if I'm reading the failing test correctly, the IgnoreBase and IgnoreLigature are respected in ligation. Which is weird because the case I was trying to fix was also ligation. Perhaps, as is the case in MS shapers, they behave differently across different shapers. More reason to just leave things as is...

Issue #3010, opened 2021-06-06
[fuzzer] Verify buffer output


behdad:
The buffer verification code in `util/options.hh` should be moved into its own file, and invoked from our fuzzer test as well:

  https://github.com/harfbuzz/harfbuzz/blob/5585ea02eb90c013ddba73290d3a9ae435df9f15/util/options.hh#L273-L421

That would prevent issues like https://github.com/harfbuzz/harfbuzz/issues/2997

behdad:
This is easy task if someone wants to help.

behdad:
I think I want to move the verification code into the library and have it be enabled by abusing a buffer flag, eg. `HB_BUFFER_FLAG_VERIFY`. Any objections? @jfkthame

behdad:
On that note, I'm thinking of also adding a config version that is not as aggressive as `HB_LEAN`, to be used by default by Firefox/Chrome builds. That would eg. also disable the code for this proposed flag, as well as buffer message and other unused features.

jfkthame:
Yeah, that sounds reasonable - my first thought re:

> I think I want to move the verification code into the library and have it be enabled by abusing a buffer flag, eg. HB_BUFFER_FLAG_VERIFY. Any objections? @jfkthame

was that it sounds OK, but I'd want to be able to disable that code in a Firefox build; assuming we wouldn't actually be using it in the browser, we don't want to spend time building it or have it bloating the product.

So a config option to disable things like this would be good - thanks.

behdad:
> Yeah, that sounds reasonable - my first thought re:
> 
> > I think I want to move the verification code into the library and have it be enabled by abusing a buffer flag, eg. HB_BUFFER_FLAG_VERIFY. Any objections? @jfkthame
> 
> was that it sounds OK, but I'd want to be able to disable that code in a Firefox build; assuming we wouldn't actually be using it in the browser, we don't want to spend time building it or have it bloating the product.
> 
> So a config option to disable things like this would be good - thanks.

Absolutely. I also want to add more detailed messaging, to allow building a stepping-level debugger. But obviously don't want all that code to ship in browsers. But I still want it enabled in default build such that hb-shape shipped by Linux distros will have it..

drott:
[...]

PR #3398, opened 2022-01-28
Add HB_BUFFER_FLAG_VERIFY

behdad:
Part of fixing https://github.com/harfbuzz/harfbuzz/issues/3010


behdad:
@jfkthame @drott can we work on defining a subset of HB_LEAN config that is good enough to be default for browsers?

behdad:
> @jfkthame @drott can we work on defining a subset of HB_LEAN config that is good enough to be default for browsers?

It feels like other than `HB_NO_VAR`, the rest might be good enough?

PR #3403, opened 2022-01-31
[subset] Don't hold references to members of the active_glyph_stack.

garretrieger:
These references may get invalidated after the vector for the stack is resized. Fixes: https://oss-fuzz.com/testcase-detail/5422577634377728


behdad (approved):
@khaledhosny can we get a follow up release with this please?

khaledhosny:
https://github.com/harfbuzz/harfbuzz/releases/tag/3.3.1

PR #3410, opened 2022-02-02
[Boring Expansion] >64k loca & hmtx tables

behdad:
These implement https://github.com/be-fonts/boring-expansion-spec/issues/

Starts with the >64k glyph limit.

### DIFF
--- a/src/hb-config.hh
+++ b/src/hb-config.hh
@@ -85,6 +85,7 @@
 #ifdef HB_MINI
 #define HB_NO_AAT
 #define HB_NO_LEGACY
+#define HB_NO_BORING_EXPANSION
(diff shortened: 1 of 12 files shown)


behdad:
I like to merge this now. It does two things:

- The num-glyphs reported by the face now is the maximum reported by the `maxp` and that deduced from the length of the `loca` table; I think this is the right thing to do anyway;  According to OpenType such `loca` tables are invalid.

- The interpretation `hmtx` tables that have excessive bytes at the end, again, invalid according to OpenType, has changed. Previously we were interpreting those excessive bytes as extra lsb values. Now we interpret them as extra advance values.

Issue #2675, opened 2020-09-09
hb-view doesn't render anything with this font


olivierberten:
it seems to work anywhere else (hb-shape included) but hb-view renders an empty image
[SN Kh SONY Moul.ttf.zip](https://github.com/harfbuzz/harfbuzz/files/5196735/SN.Kh.SONY.Moul.ttf.zip)
[source](https://akbalthom.blogspot.com/2020/04/khim-sony.html)

khaledhosny:
The font has `COLR` table that seems to be breaking something, if I drop that table things get fixed.

I though the table is bogus, but it seems that hb-view will not show anything for glyphs not covered by `COLR`, not just in this font which makes no sense. I wounder where is this breaking, pango-view seems fine, so I guess it is not Cairo or FreeType.

behdad:
Is definitely Cairo or FreeType issue since hb-view by  itself doesn't do any magic. It might be that `FT_HAS_COLOR` is broken for this font.

What happens if you pass `--background=01`?

behdad:
When reporting bugs please also specify sample text to reproduce the bug. Thanks.

behdad:
> What happens if you pass `--background=01`?

Doesn't do anything.

I can reproduce. Will look into later.

khaledhosny:
The issue happens with other fonts too, for glyphs not covered by `COLRz` table (e.g. a م in Amiri Quran Colored).

behdad:
So, the color glyphs work?

khaledhosny:
Yes.

simoncozens:
Removing the `0 &&` from this line https://github.com/harfbuzz/harfbuzz/blob/ea4476d7389b8b5d30f1d969787c752856c32088/util/view-cairo.cc#L123

makes it display. That might not be the fix because I presume it's if'd out for a reason. But it should be a good place to start looking.

behdad:
> makes it display. That might not be the fix because I presume it's if'd out for a reason. But it should be a good place to start looking.

Basically what that suggests is that it's a bug in cairo's color-glyph processing. Which is what I knew anyway since we don't do any of that.

matthiasclasen:
Here is what I see with that font:

![Screenshot from 2021-07-31 08-37-08](https://user-images.githubusercontent.com/4834845/127740084-c9567153-d8f8-4d73-ac15-1a496d966a53.png)

It is not that nothing is there, it is that you get white-on-white text.

[...]

PR #3411, opened 2022-02-02
[draw] Finish and release draw API

behdad:
Okay I'm determined to finish and release draw API in the next three weeks of hacking. @ebraminio are you available to help? @khaledhosny can I get some help from you?

I know what I like done, and I'll enumerate that in the next comments. Before that, I like to know how @ebraminio was testing this. I guess with JS? UPDATE: Using src/main tool.


behdad:
So far what I like to change:

[ ] Use `float` instead of `hb_position_t` in the API. If nothing else, saves a `roundf()` per position.
[ ] The callbacks should also get the `hb_draw_funcs_t` pointer itself. It was a lazy move to not do that to begin with.
[ ] Expose the `hb_draw_helper_t` object as a public struct, called `hb_draw_state_t` or similar.
[ ] Add public API called `hb_draw_move_to`, hb_draw_line_t`, etc, that operate on a `hb_draw_funcs_t` and a `hb_draw_state_t`.
[ ] (hardest probably) make `hb-view` use these and plug into cairo user-font API to render without cairo-ft.

ebraminio:
> Okay I'm determined to finish and release draw API in the next three weeks of hacking.

Great to hear!

Guess now needs coordination with the brand new _synthetic_slant also.

behdad:
> Guess now needs coordination with the brand new _synthetic_slant also.

Interesting question!

behdad:
The `test-draw.c` is clearly not doing a good job of testing, since I change the signature of the draw calls from `hb_position_t` to `float` and forgot to update that test, and it's still passing. Ouch...

behdad:
> The `test-draw.c` is clearly not doing a good job of testing, since I change the signature of the draw calls from `hb_position_t` to `float` and forgot to update that test, and it's still passing. Ouch... @ebraminio

I had forgotten to remove `HB_EXPERIMENTAL_API`...

test/api/test-draw.c
```diff
@@ -926,8 +927,8 @@ main (int argc, char **argv)
   hb_test_add (test_hb_draw_cff1_rline);
   hb_test_add (test_hb_draw_cff2);
   hb_test_add (test_hb_draw_ttf_parser_tests);
-  hb_test_add (test_hb_draw_font_kit_glyphs_tests);
-  hb_test_add (test_hb_draw_font_kit_variations_tests);
+if(0)  hb_test_add (test_hb_draw_font_kit_glyphs_tests);
+if(0)  hb_test_add (test_hb_draw_font_kit_variations_tests);
```
behdad, line 931: @ebraminio any chance you can debug why these two tests fail?

behdad:
Humm. draw-fuzzer tests fail as well.

behdad:
> Humm. draw-fuzzer tests fail as well.

That's my fault. Fixing.

meson.build
```diff
@@ -208,10 +208,6 @@ if get_option('icu_builtin')
   conf.set('HAVE_ICU_BUILTIN', 1)
 endif
 
-if get_option('experimental_api')
-  conf.set('HB_EXPERIMENTAL_API', 1)
-endif
```
ebraminio, line 213: Great that alll are stable now, just make sure https://github.com/google/oss-fuzz/blob/master/projects/harfbuzz/build.sh#L32 will be removed while the merge

behdad: Thanks for the heads-up. That would be a good problem to have when we merge this. :)

src/hb-draw.hh
```diff
@@ -27,76 +27,134 @@
...
       return;
     if (!path_open) start_path ();
-    if (funcs->is_quadratic_to_set)
-      funcs->quadratic_to (control_x, control_y, to_x, to_y, user_data);
+    if (funcs->quadratic_to_is_set ())
+      funcs->quadratic_to (draw_data, control_x, control_y, to_x, to_y);
     else
-      funcs->cubic_to (roundf ((current_x + 2.f * control_x) / 3.f),
+      funcs->cubic_to (draw_data,
+		       roundf ((current_x + 2.f * control_x) / 3.f),
```
ebraminio, line 138: guess these can go now

test/api/test-draw.c
ebraminio: most likely libc's roundf used in _hb_iota is different with internal _hb_roundf this code used to enjoy.

ebraminio: which

```
diff --git a/test/api/test-draw.c b/test/api/test-draw.c
index 3204acee0..fdf437f44 100644
--- a/test/api/test-draw.c
+++ b/test/api/test-draw.c
@@ -49,7 +49,7 @@ static void _hb_reverse (char *buf, unsigned int len)
 }
 static unsigned _hb_itoa (float fnum, char *buf)
 {
-  int32_t num = (int32_t) roundf (fnum);
+  int32_t num = (int32_t) floorf (fnum + .5f);
   unsigned int i = 0;
   hb_bool_t is_negative = num < 0;
   if (is_negative) num = -num;
@@ -934,8 +934,8 @@ main (int argc, char **argv)
   hb_test_add (test_hb_draw_cff1_rline);
   hb_test_add (test_hb_draw_cff2);
   hb_test_add (test_hb_draw_ttf_parser_tests);
-if(0)  hb_test_add (test_hb_draw_font_kit_glyphs_tests);
-if(0)  hb_test_add (test_hb_draw_font_kit_variations_tests);
+  hb_test_add (test_hb_draw_font_kit_glyphs_tests);
+  hb_test_add (test_hb_draw_font_kit_variations_tests);
   hb_test_add (test_hb_draw_estedad_vf);
   hb_test_add (test_hb_draw_stroking);
   hb_test_add (test_hb_draw_immutable);
```

makes it work

src/gen-def.py
```diff
@@ -19,23 +19,7 @@
...
-hb_draw_funcs_destroy
-hb_draw_funcs_is_immutable
-hb_draw_funcs_make_immutable
-hb_draw_funcs_reference
-hb_draw_funcs_set_close_path_func
-hb_draw_funcs_set_cubic_to_func
-hb_draw_funcs_set_line_to_func
-hb_draw_funcs_set_move_to_func
-hb_draw_funcs_set_quadratic_to_func""".splitlines ()
+"""""".splitlines ()
```
ebraminio, line 22: guess you can now remove the section

behdad:
> * [ ]  Implement above in `hb-ft`.

@khaledhosny How about this?

ebraminio:
>> Implement above in hb-ft.

> How about this?

Guess you can use these,
* FreeType: https://github.com/harfbuzz/harfbuzz/pull/2185/files#diff-5f9da06928ddb39791ce9bf3ccf6f7840b4fa310d7c45bb03a636a6ca7e16af2
  or this one: https://github.com/harfbuzz/harfbuzz/pull/2183/files#diff-5f9da06928ddb39791ce9bf3ccf6f7840b4fa310d7c45bb03a636a6ca7e16af2R983
* GDI: https://github.com/harfbuzz/harfbuzz/pull/2183/files#diff-68914fbcf819bd13f4411baab71b1d71b530bb520bb13ef2f9074622b0221825R1038
* CoreText: https://github.com/harfbuzz/harfbuzz/pull/2183/files#diff-762f8007a49dafcbe4ff5a2f7077b4576edd6047c7a75686a6ab06ae868e47f5R1166
* DirectWrite: https://github.com/harfbuzz/harfbuzz/pull/2183/files#diff-0ed175486367eab57a67582c28e91f0f01412c0b1686f717427c940aa18ac560R979

behdad:
> * [ ]  Write tests for the `hb_draw_move_to/etc`.

@ebraminio Can I get help from you on this one?

behdad:
image
First rendering from hb-draw in hb-view. Not perfect, but not nothing. I believe the shortcomings are cairo-user-font shortcomings, not ours...

behdad:
> I believe the shortcomings are cairo-user-font shortcomings, not ours...

And as I suspected, fixed in cairo main... Just not in a release yet :(.

behdad:
image
In its full glory with cairo main branch.

behdad:
To summarize, what's left:
* [ ] hb-ft implementation
* [ ] chaining up to parent font in hb-font's default implementation
* [ ] Figure out interaction with synthetic_slant.
* [ ] Document draw callbacks and other new API.
* [ ] Write tests for the hb_draw_move_to/etc.

I *think* this can go in then, with the hb-view implementation as well, but will depend on very recent cairo. So initially I suggest we push in everything but that last bit.

behdad:
> I _think_ this can go in then, with the hb-view implementation as well, but will depend on very recent cairo. So initially I suggest we push in everything but that last bit.

Okay, rewrote the logic to keep the cairo-ft implementation as well, but enable the hb-draw logic if cairo is recent-enough (1.17.5 version-check) AND font-funcs are NOT `ft`. The latter check can be removed when we implement the callbacks for `hb-ft`.

behdad:
Ouch. Lost a file... :(

behdad:
@khaledhosny dumb question. How do I make meson pick up my cairo from ~/.local? pkg-config sees it, but meson refuses to pick it up. Thanks.

khaledhosny:
It should work, but may be you need to rerun meson (e.g. neson setup --wipe).

behdad:
Question: Should I add back the return value of the `hb_font_get_glyph_shape()` call? Such that callers can know whether implementation exists?

behdad:
> Question: Should I add back the return value of the `hb_font_get_glyph_shape()` call? Such that callers can know whether implementation exists?

I personally don't see the point of that. You are writing the code, you know what's on the other side...

behdad:
image
Rendering PNG emoji using recent cairo-user-font render-color-glyph callback and hb-ot-color PNG api now... as well.

behdad:
Implemented fetching shape from parent font if current font's funcs don't implement, with resizing. Took a while to get it right.
image

behdad:
> Implemented fetching shape from parent font if current font's funcs don't implement, with resizing. Took a while to get it right.

Test code for above was this:
```patch
diff --git a/util/helper-cairo-user.hh b/util/helper-cairo-user.hh
index 075db0a15..c8ac90e97 100644
--- a/util/helper-cairo-user.hh
+++ b/util/helper-cairo-user.hh
@@ -108,7 +108,12 @@ render_glyph (cairo_scaled_font_t  *scaled_font,
   hb_font_get_scale (font, &x_scale, &y_scale);
   cairo_scale (cr, +1./x_scale, -1./y_scale);
 
-  hb_font_get_glyph_shape (font, glyph, get_cairo_draw_funcs (), cr);
+  hb_font_t *font2 = hb_font_create_sub_font (font);
+  signed x, y;
+  hb_font_get_scale (font, &x, &y);
+//  printf ("%d %d\n", x, y);
+  hb_font_set_scale (font2, 11520/2, 11520/2);
+  hb_font_get_glyph_shape (font2, glyph, get_cairo_draw_funcs (), cr);
   cairo_fill (cr);
 
   return CAIRO_STATUS_SUCCESS;
```

behdad:
@ebraminio I'm happy with the API now that I've implemented parent scaling fallback I wanted to implement. Can you update your harfbuzzjs to use it and see how it goes?

behdad:
> Test code for above was this:

Before I got it right:
image

behdad:
> Before I got it right:

FWIW, this kind of rescaling is why `hb_draw_state_t` is exposed to the draw callbacks, so they can transform the drawing state.

behdad:
> FWIW, this kind of rescaling is why `hb_draw_state_t` is exposed to the draw callbacks, so they can transform the drawing state, as seen in `hb_font_get_glyph_shape_default_adaptor_t`, which is really a scaling-adaptor: [b9ef217](https://github.com/harfbuzz/harfbuzz/commit/b9ef217cb45d7beab24d57dbda3e9613f28e6951)

To clarify, this is needed because of the quadratic-to-cubic conversion. Otherwise it was not necessary to keep state.

behdad:
> * [ ]  Figure out interaction with `synthetic_slant`.

I'm going to make `synthetic_slant` modify the shape.

src/hb-font.cc
```diff
@@ -1168,6 +1188,24 @@ hb_font_get_glyph_from_name (hb_font_t      *font,
...
 
+/**
+ * hb_font_get_glyph_shape:
+ * @font: #hb_font_t to work upon
+ * @glyph: : The glyph ID
+ * @dfuncs:
+ * @draw_data:
+ *
+ * Fetches the glyph shape that corresponds to a glyph in the specified @font.
+ *
```
matthiasclasen, line 1313: Would be good to say that the glyph shape is 'returned' in the form of calls to the hb_draw_func_t callbacks. That is what is going to happen, right?

src/hb-draw.h
```diff
@@ -33,6 +33,25 @@
...
+  float path_start_y;
+
+  float current_x;
+  float current_y;
+
+  /*< private >*/
+  hb_var_num_t   reserved1;
+  hb_var_num_t   reserved2;
+  hb_var_num_t   reserved3;
+} hb_draw_state_t;
```
matthiasclasen, line 50: You could be more generous with padding here, imo

behdad:
Implemented slanting as well!

behdad:
Can I get some help with documentation and testing please? That's all that is left.

behdad:
Did docs as well. Branch is ready as far as I'm concerned...

khaledhosny:
I messaged the documentation a bit. Question: is `HB_DRAW_STATE_DEFAULT` meant to be public API? If so, we need to add some description to it.

behdad:
> Question: is `HB_DRAW_STATE_DEFAULT` meant to be public API?

Yes. Since `hb_draw_state_t` is a public struct, it needs some initializer.

khaledhosny:
I have updated cairo subproject, so if to automatically build a newer cairo with the `cairo_user_font_face_set_render_color_glyph_func()` API, meson can be configured with `-Dcairo=enabled --force-fallback-for=cairo`.

khaledhosny:
Looks like we lost the ability to use COLRv0 layers and now always show the BW glyphs.

behdad:
> Looks like hb-view lost the ability to use COLRv0 layers and now always show the BW glyphs.

I can implement that too since we have API exposing them and cairo now has API to implement them (and with proper foreground color). Can you email me a sample font?

behdad:
I feel like still having a way to tell hb-view to use the old way to render things would come handy to compare. Let me think.

khaledhosny:
For test fonts there is https://github.com/aliftype/amiri/blob/main/AmiriQuranColored.otf

behdad:
Emoji with font-funcs ft are still broken (always have been). We should switch default font funcs. Will llook into it.

behdad:
Here we go:
```
$ ninja && util/hb-view AmiriQuranColored.otf  -u 623,624,625,626,627,631 --font-funcs ot  --foreground=000080 | display
```
image

khaledhosny:
Excellent, but I’m getting a bunch of warnings now:
```
[1/2] Compiling C++ object util/hb-view.p/hb-view.cc.o
In file included from ../util/hb-view.cc:33:
In file included from ../util/view-cairo.hh:32:
In file included from ../util/helper-cairo.hh:33:
../util/helper-cairo-user.hh:44:28: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_move_to (cr, to_x, to_y);
  ~~~~~~~~~~~~~            ^~~~
../util/helper-cairo-user.hh:44:22: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_move_to (cr, to_x, to_y);
  ~~~~~~~~~~~~~      ^~~~
../util/helper-cairo-user.hh:54:28: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_line_to (cr, to_x, to_y);
  ~~~~~~~~~~~~~            ^~~~
../util/helper-cairo-user.hh:54:22: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_line_to (cr, to_x, to_y);
  ~~~~~~~~~~~~~      ^~~~
../util/helper-cairo-user.hh:66:77: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_curve_to (cr, control1_x, control1_y, control2_x, control2_y, to_x, to_y);
  ~~~~~~~~~~~~~~                                                            ^~~~
../util/helper-cairo-user.hh:66:71: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_curve_to (cr, control1_x, control1_y, control2_x, control2_y, to_x, to_y);
  ~~~~~~~~~~~~~~                                                      ^~~~
../util/helper-cairo-user.hh:66:59: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_curve_to (cr, control1_x, control1_y, control2_x, control2_y, to_x, to_y);
  ~~~~~~~~~~~~~~                                          ^~~~~~~~~~
../util/helper-cairo-user.hh:66:47: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_curve_to (cr, control1_x, control1_y, control2_x, control2_y, to_x, to_y);
  ~~~~~~~~~~~~~~                              ^~~~~~~~~~
../util/helper-cairo-user.hh:66:35: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_curve_to (cr, control1_x, control1_y, control2_x, control2_y, to_x, to_y);
  ~~~~~~~~~~~~~~                  ^~~~~~~~~~
../util/helper-cairo-user.hh:66:23: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_curve_to (cr, control1_x, control1_y, control2_x, control2_y, to_x, to_y);
  ~~~~~~~~~~~~~~      ^~~~~~~~~~
10 warnings generated.
```

khaledhosny:
I also wonder where is the joining mismatch in ئا coming from, I don’t see it elsewhere.

behdad:
> I also wonder where is the joining mismatch in ئا coming from, I don’t see it elsewhere.

I see it pronounced in this branch. It's certainly there with main, just not as much.

behdad:
> I see it pronounced in this branch. It's certainly there with main, just not as much.

Not really:
image

This is hard to debug on my side. Help from @ebraminio would be appreciated.

First step is to see if is reproduced without color stuff. Ie.

behdad:
> First step is to see if is reproduced without color stuff. Ie.

Doesn't happen with the color stuff; which is rather weird:
image

khaledhosny:
Another issue:


```
./util/hb-view AmiriQuranColored.otf  -u 66e,627 --font-funcs=ot
```


But:
```
./util/hb-view AmiriQuranColored.otf  -u 66e,627,626,627 --font-funcs=ot
```

khaledhosny:
```
$ ./util/hb-view AmiriQuranColored.otf  -u 66e,627,626 --font-funcs=ot
Assertion failed: (!"reached"), function _cairo_no_compositor_glyphs, file cairo-no-compositor.c, line 96.
Abort trap: 6
```

behdad:
> Assertion failed: (!"reached"), function _cairo_no_compositor_glyphs, file cairo-no-compositor.c, line 96.

Yeah. If you remove this assertion in cairo it works fine for me:
```
diff --git a/src/cairo-no-compositor.c b/src/cairo-no-compositor.c
index 1602a12f6..e5d1036d5 100644
--- a/src/cairo-no-compositor.c
+++ b/src/cairo-no-compositor.c
@@ -93,7 +93,7 @@ _cairo_no_compositor_glyphs (const cairo_compositor_t *compositor,
                             int                         num_glyphs,
                             cairo_bool_t overlap)
 {
-    ASSERT_NOT_REACHED;
+/*    ASSERT_NOT_REACHED; */
     return CAIRO_INT_STATUS_NOTHING_TO_DO;
 }
 
```

and then, for U+0629 I get no color with old code, but color dots with new code, and looking into the font, looks like colored rendering is correct?

behdad:
> Another issue:

It's the same, whatever it is. Looks like error introduced by intense scaling we do...

behdad:
> Another issue:

Nah, this is really weird...

khaledhosny:
> > Another issue:
> 
> It's the same, whatever it is. Looks like error introduced by intense scaling we do...

In the second screenshot the dotless beh (U+066E) disappears.

behdad:
> In the second screenshot the dotless beh (U+066E) disappears.

Oh. I first thought your talking about the misalignment. Then thought the jagged alef on the left you are talking about...
The missing glyph is a whole other issue...

Anyway, quite possibly we are hitting issues in cairo's bowels now, which is a bummer. Time to report there and live the hb-view rendering to the old backend for a while until cairo is investigated...

behdad:
We need help reporting to cairo list...

behdad:
Looks like the weirdness was there already, and even more inconsistent:
```
$ ninja && HB_DRAW=0 ./util/hb-view AmiriQuranColored.otf  -u 626,627,626,627,626,627 --font-funcs=ot | display
```
image

```
$ ninja && HB_DRAW=1 ./util/hb-view AmiriQuranColored.otf  -u 626,627,626,627,626,627 --font-funcs=ot | display
```
image

khaledhosny:
I think this is #2675

behdad:
> I think this is #2675

The latest case I posted shows fairly convincingly that there's a serious bug in cairo's cache somewhere...

behdad:
So what you're saying is that the fallback from color to non-color isn't working... Is possible. I can work around that by chaining from render-color-glyph to render-glyph myself. But it only sometimes doesn't work, which suggest some serious problem going on...

behdad:
> So what you're saying is that the fallback from color to non-color isn't working... Is possible. I can work around that by chaining from render-color-glyph to render-glyph myself. But it only sometimes doesn't work, which suggest some serious problem going on...

That did help:
image

khaledhosny:
Yes, looks like `HB_DRAW=1` works more reliably now than `HB_DRAW=0`, so I think we should switch the default back.

behdad:
I reported to cairo list. Hopefully we get some help..

> Yes, looks like `HB_DRAW=1` works more reliably now than `HB_DRAW=0`, so I think we should switch the default back.

So, if we also change default font-funcs from `ft` to `ot`, our emoji rendering with the new code will be fairly solid, compared to default-broken that it was.

In return, we lose hinting. But no one uses `hb-view` for that feature, and we default to no hinting. So, should be fine switching. And it would help discover brokenness in hb-draw. So I'm fine doing it...

khaledhosny:
Agree, no hinting by default shouldn’t be an issue for `hb-draw`.

behdad:
> Agree, no hinting by default shouldn’t be an issue for `hb-draw`.

On older cairo that doesn't have `render_color_glyph`, we should use old rendering though, agree?

khaledhosny:
I also like how more of `hb-view` functionality is now implemented by HarfBuzz.

khaledhosny:
> On older cairo that doesn't have `render_color_glyph`, we should use old rendering though, agree?

Yes, makes sense.

behdad:
> On older cairo that doesn't have `render_color_glyph`, we should use old rendering though, agree?

So: if `HB_DRAW=0`, don't use it, if  `HB_DRAW=1`, use it, if unset, choose depending on cairo version.

khaledhosny:
> So: if `HB_DRAW=0`, don't use it, if `HB_DRAW=1`, use it, if unset, choose depending on cairo version.

Yes, better default and explicit control for debugging.

test/api/test-draw.c
```diff
@@ -819,7 +832,7 @@ test_hb_draw_stroking (void)
   /* https://skia-review.googlesource.com/c/skia/+/266945
      https://savannah.nongnu.org/bugs/index.php?57701 */
```
ebraminio, line 833: The reason `test_hb_draw_stroking` which is now disabled existed was the above discussions.

behdad: > The reason `test_hb_draw_stroking` which is now disabled existed was the above discussions.
> 
> Update: So this is done intentionally [0f131f5](https://github.com/harfbuzz/harfbuzz/commit/0f131f594bc4aa0e2a81c897b192cd90e3bb759e) just that the reason for this was helping future Skia integration https://skia-review.googlesource.com/c/skia/+/268166/11/src/ports/SkFontHost_FreeType_common.cpp which apparently doesn't matter, maybe you can have Ben's input also.

Can you elaborate?

If a rasterizer cannot accept no-op operations, it's a broken rasterizer. I highly doubt that's what's at fault here, other than cairo asserting if the glyph is fully empty...

@bungeman does Skia have any problem with line-to, conic-to, cubic-to that have all points equal?

behdad: > If a rasterizer cannot accept no-op operations, it's a broken rasterizer.

Stroking.. I see now how that can be a problem. Still, I think that's best left to such an implementation.

bungeman: Yeah, the Skia thing was mostly about how to handle fonts which use degenerate contours for attachment points (Skia itself handles degenerate segments more or less how SVG specifies, since that's what most of Skia's users expect). These take up no area so are usually harmless, but if you want to stroke the outline (like SVG wants to do) then there are decisions to be made. SVG says degenerate contours get stroked (so draw endcaps) but the reality is that degenerate contours in simple paths in OpenType fonts are used for other reasons. Also, it turns out that FreeType, CoreText, DirectWrite, and even GDI handle the reporting of degenerate contours slightly differently (depends on number of points and if they're on or off curve). So for consistency, Skia's font handling just filters them all out.

That being said, I'm aware of two users of paths. One wants the raw paths with all the original point numbers and whatnot. The other is the more common user who wants to manipulate the path or draw it in some novel way. For these more common users (and this includes Skia) the reporting of any degenerate contours (which can only be meaningful internally to the font) is just noise and needs to be filtered out.

khaledhosny: I think we want to give as unfiltered data as sensible here, because if we do the filtering at the lower level, higher level code that wants the unfiltered data is left helpless, but the other-way around the higher level code will be able to do the filtering as it fits its needs.

behdad: > I think we want to give as unfiltered data as sensible here, because if we do the filtering at the lower level, higher level code that wants the unfiltered data is left helpless, but the other-way around the higher level code will be able to do the filtering as it fits its needs.

We're still filtering out the love `move-to`, and repeat move-to's in CFF, so unfortunately that argument doesn't stand. This however, at least takes care of not stroking a single point used as attachment point...

behdad: > We can add a mode to hb_draw_state_t to choose whether to clean up or not...

Would just add complexity...

khaledhosny: > We can add a mode to `hb_draw_state_t` to choose whether to clean up or not...

That would be ideal, IMHO.

behdad: The mode as I see it, would be `allow_empty_path`...

ebraminio: In response https://github.com/harfbuzz/harfbuzz/commit/0f131f594bc4aa0e2a81c897b192cd90e3bb759e

> Plus, this throws off point-numbers in uses that rely on it.

I remember CFF/CFF2 didn't have guaranteed point numbers also if I'm not mistaken about what you want to refer, am not insisting on either case anyway.

khaledhosny:
The checklist in https://github.com/harfbuzz/harfbuzz/pull/3411#issuecomment-1028379441 is all addressed now.

test/api/test-draw.c
behdad: > I think we want to give as unfiltered data as sensible here, because if we do the filtering at the lower level, higher level code that wants the unfiltered data is left helpless, but the other-way around the higher level code will be able to do the filtering as it fits its needs.

So the points we return are not useful for applying native hints... so it's just for things like master-level compatibility-checking, etc.

We are just eliding empty contours, and I'm leaning towards not doing that and leaving it to the consumer. I believe FreeType / cairo do the same. I don't think it belongs to our layer. Otherwise we need to add a new flags parameter to `hb_font_get_glyph_shape` and add that to the `hb_draw_state_t` and change the state implementation.

khaledhosny: That works for me as well.

behdad: Humm. There's still this bit of code:
```
  close_path (void *draw_data, hb_draw_state_t &st) 
  { 
    if (st.path_open) 
    { 
      if ((st.path_start_x != st.current_x) || (st.path_start_y != st.current_y)) 
↦       emit_line_to (draw_data, st, st.path_start_x, st.path_start_y); 
      emit_close_path (draw_data, st); 
    } 
```

Can't remove it without leaving spurious `line_to`'s all over the place; if I *don't* remove it, breaks the master-compatibility of outlines. So I feel like a "keep outline compatibility" flag is required in the API somewhere.

behdad: And really that's the only way I can think of this flag... No further promise I can think of...

behdad: At any rate, a flags_t in the API seems justified.

behdad: I'll think some more and make up my mind...

behdad:
The fuzzer failing case is beyond my mind right now...

behdad:
> The fuzzer failing case is beyond my mind right now...

Working on it.

behdad:
I gave up on the immediate move-to...

khaledhosny:
> @khaled, can you add a test, testing that making a subfont with a different scale

You mentioned the wrong khaled but I'm working on the test.

behdad:
> I think this is #2675

Related. https://gitlab.freedesktop.org/cairo/cairo/-/issues/533

khaledhosny:
> > can you add a test, testing that making a subfont with a different scale
> 
> You mentioned the wrong khaled but I'm working on the test.

Done. This branch now has net positive increase in coverage compared to main (by a tiny bit) according to codecov. It previously showed net negative change in coverage, so I guess we are on good shape with test coverage.

behdad:
> Done. This branch now has net positive increase in coverage compared to main (by a tiny bit) according to codecov. It previously showed net negative change in coverage, so I guess we are on good shape with test coverage.

Nice. Thanks.

And thanks to Matthias, we're close to having fixes for the cairo issues discovered, so we can get a fixed cairo released hopefully next week...

behdad:
@jfkthame any concerns with merging this?

behdad:
With cairo master now all issues are fixed, except for that off-by-one pixel rendering, which I worked around by always rendering glyphs using `render-color-glyph` callback... We'll try to debug that one as well next week.

test/api/test-draw.c
ebraminio: So now maybe you also wish to revisit "/* Skip empty contours */" of `hb-ot-glyf-table.hh` and allow emit empty contours emit, it was added here https://github.com/harfbuzz/harfbuzz/commit/3a98c7fae196fd149078af4553b966dc1c8d763b but now the correspond line is https://github.com/harfbuzz/harfbuzz/blob/e59da28/src/hb-ot-glyf-table.hh#L1188

behdad: Done.  Thanks.

test/api/test-draw.c
```diff
@@ -330,43 +343,43 @@ test_hb_draw_ttf_parser_tests (void)
...
+      g_assert_cmpmem (str, draw_data.consumed, expected, sizeof (expected) - 1);
     }
     {
-      user_data.consumed = 0;
-      g_assert (hb_font_draw_glyph (font, 5, funcs, &user_data));
+      draw_data.consumed = 0;
+      hb_font_get_glyph_shape (font, 5, funcs, &draw_data);
       char expected[] = "";
-      g_assert_cmpmem (str, user_data.consumed, expected, sizeof (expected) - 1);
+      g_assert_cmpmem (str, draw_data.consumed, expected, sizeof (expected) - 1);
```
ebraminio, line 373: Allowing empty contours emit should have changed expectation of this and the next tests also like what original effect of that https://github.com/harfbuzz/harfbuzz/commit/3a98c7fae196fd149078af4553b966dc1c8d763b weird :/ I should have a closer look at it, if you can also please consider having a look at it, thanks

behdad: Empty can have many meanings. We still don't emit a moveto until a lineto/quadritcto/curveto... I gave up on that... So I think that's why you don't see a drastic change.

ebraminio: Okay debugged this a bit,

5th glyph of the test font glyphs.ttf according to fonttools is:

```
    <TTGlyph name="space" xMin="0" yMin="0" xMax="0" yMax="0">
      <contour>
        <pt x="0" y="0" on="0"/>
      </contour>
      <instructions/>
    </TTGlyph>
```

And this can make it print the empty contour apparently,

```diff
diff --git a/src/hb-ot-glyf-table.hh b/src/hb-ot-glyf-table.hh
index dee465a6b..34170ab36 100644
--- a/src/hb-ot-glyf-table.hh
+++ b/src/hb-ot-glyf-table.hh
@@ -1202,7 +1202,10 @@ struct glyf
              draw_session->move_to (font->em_fscalef_x (mid.x), font->em_fscalef_y (mid.y));
            }
            else
+           {
              first_offcurve = p;
+             last_offcurve = p;
+           }
          }
        }
        else
diff --git a/test/api/test-draw.c b/test/api/test-draw.c
index 72b1aadfb..7ac616884 100644
--- a/test/api/test-draw.c
+++ b/test/api/test-draw.c
@@ -369,7 +369,7 @@ test_hb_draw_ttf_parser_tests (void)
     {
       draw_data.consumed = 0;
       hb_font_get_glyph_shape (font, 5, funcs, &draw_data);
-      char expected[] = "";
+      char expected[] = "M0,0Q15,0 15,0L0,0Z";
       g_assert_cmpmem (str, draw_data.consumed, expected, sizeof (expected) - 1);
     }
     {
@@ -378,7 +378,8 @@ test_hb_draw_ttf_parser_tests (void)
       char expected[] = "M346,468L211,468L211,0L123,0L123,468L29,468L29,509L123,539"
                        "L123,570Q123,674 169,720Q215,765 297,765Q329,765 356,760"
                        "Q382,754 401,747L378,678Q362,683 341,688Q320,693 298,693"
-                       "Q254,693 233,664Q211,634 211,571L211,536L346,536L346,468Z";
+                       "Q254,693 233,664Q211,634 211,571L211,536L346,536L346,468Z"
+                       "M0,0Q15,0 15,0L0,0Z";
       g_assert_cmpmem (str, draw_data.consumed, expected, sizeof (expected) - 1);
     }
 
 ```

I'd suggest now that the expectation of the code is going to be changed it would be a good idea to revisit the logic of `consume_point` and see if it matches trusted codes you have somewhere else. Touching it probably changes expectation of edge cases tests but guess updating those worth the hassle.

Comment (behdad) [reply test/api/test-draw.c:373]: @ebraminio Curious. Where that 15 come from?!! A lone point contour in a TTF outline.... I have no idea how that is to be interpretted!

Comment (behdad) [reply test/api/test-draw.c:373]: I think it all the numbers involved should be 0.

Comment (behdad) [reply test/api/test-draw.c:373]: @ebraminio I suggest you bring it up with @RazrFalcon and we commit whatever fix he comes up with.

Comment (behdad) [reply test/api/test-draw.c:373]: My reading of the code suggests that your suggested change is correct.

Comment (behdad) [reply test/api/test-draw.c:373]: The 15 is coming from a lack of prior `move_to`... so.. to me this change is not working correctly yet :(.

**Comment by behdad** [comment]
> I also wonder where is the joining mismatch in ئا coming from, I don’t see it elsewhere.

It's a mismatch of expectations inside cairo, caused by the subpixel-positioning code, expecting the rasterizer (cairo-ft) to do something, whereas now it's my code in the cairo user-font rasterizing, not cairo-ft...

**Comment by behdad** [comment]
> > I also wonder where is the joining mismatch in ئا coming from, I don’t see it elsewhere.
> 
> It's a mismatch of expectations inside cairo, caused by the subpixel-positioning code, expecting the rasterizer (cairo-ft) to do something, whereas now it's my code in the cairo user-font rasterizing, not cairo-ft...

Fix here:
https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/274

With this patch to HB to expose the cairo bug:
```patch

diff --git a/util/helper-cairo-user.hh b/util/helper-cairo-user.hh
index 8c1f9f222..80030d03c 100644
--- a/util/helper-cairo-user.hh
+++ b/util/helper-cairo-user.hh
@@ -272,6 +272,7 @@ render_color_glyph (cairo_scaled_font_t  *scaled_font,
   if (ret != CAIRO_STATUS_USER_FONT_NOT_IMPLEMENTED)
     return ret;
 
+  return ret;
   return render_glyph (scaled_font, glyph, cr, extents);
 }
 
```

Before:
before
After:
after

test/api/test-draw.c
ebraminio: My proposed change is now:

```diff
diff --git a/src/hb-ot-glyf-table.hh b/src/hb-ot-glyf-table.hh
index dee465a6b..2c26a8df1 100644
--- a/src/hb-ot-glyf-table.hh
+++ b/src/hb-ot-glyf-table.hh
@@ -1251,6 +1251,12 @@ struct glyf
                                       font->em_fscalef_x (first_oncurve.x), font->em_fscalef_y (first_oncurve.y));
          else if (first_oncurve.has_data)
            draw_session->line_to (font->em_fscalef_x (first_oncurve.x), font->em_fscalef_y (first_oncurve.y));
+         else if (first_offcurve.has_data)
+         {
+           float x = font->em_fscalef_x (first_offcurve.x), y = font->em_fscalef_x (first_offcurve.y);
+           draw_session->move_to (x, y);
+           draw_session->quadratic_to (x, y, x, y);
+         }
 
          /* Getting ready for the next contour */
          first_oncurve = first_offcurve = last_offcurve = optional_point_t ();
diff --git a/test/api/test-draw.c b/test/api/test-draw.c
index 72b1aadfb..023e82188 100644
--- a/test/api/test-draw.c
+++ b/test/api/test-draw.c
@@ -369,7 +369,7 @@ test_hb_draw_ttf_parser_tests (void)
     {
       draw_data.consumed = 0;
       hb_font_get_glyph_shape (font, 5, funcs, &draw_data);
-      char expected[] = "";
+      char expected[] = "M15,0Q15,0 15,0Z";
       g_assert_cmpmem (str, draw_data.consumed, expected, sizeof (expected) - 1);
     }
     {
@@ -378,7 +378,8 @@ test_hb_draw_ttf_parser_tests (void)
       char expected[] = "M346,468L211,468L211,0L123,0L123,468L29,468L29,509L123,539"
                        "L123,570Q123,674 169,720Q215,765 297,765Q329,765 356,760"
                        "Q382,754 401,747L378,678Q362,683 341,688Q320,693 298,693"
-                       "Q254,693 233,664Q211,634 211,571L211,536L346,536L346,468Z";
+                       "Q254,693 233,664Q211,634 211,571L211,536L346,536L346,468Z"
+                       "M15,0Q15,0 15,0Z";
       g_assert_cmpmem (str, draw_data.consumed, expected, sizeof (expected) - 1);
     }
 
```

which makes it match freetype result on empty contours:

```diff
diff --git a/test/api/test-draw.c b/test/api/test-draw.c
index 72b1aadfb..2f129f48d 100644
--- a/test/api/test-draw.c
+++ b/test/api/test-draw.c
@@ -329,6 +329,8 @@ test_hb_draw_cff2 (void)
   hb_font_destroy (font);
 }
 
+#include "hb-ft.h"
+
 static void
 test_hb_draw_ttf_parser_tests (void)
 {
@@ -341,6 +343,7 @@ test_hb_draw_ttf_parser_tests (void)
   {
     hb_face_t *face = hb_test_open_font_file ("fonts/glyphs.ttf");
     hb_font_t *font = hb_font_create (face);
+    hb_ft_font_set_funcs (font);
     hb_face_destroy (face);
     {
       draw_data.consumed = 0;
@@ -358,27 +361,28 @@ test_hb_draw_ttf_parser_tests (void)
       draw_data.consumed = 0;
       hb_font_get_glyph_shape (font, 4, funcs, &draw_data);
       char expected[] = "M332,468L197,468L197,0L109,0L109,468L15,468L15,509L109,539"
-                       "L109,570Q109,674 155,720Q201,765 283,765Q315,765 342,760"
+                       "L109,570Q109,674 155,719Q201,765 283,765Q315,765 341,759"
                        "Q368,754 387,747L364,678Q348,683 327,688Q306,693 284,693"
-                       "Q240,693 219,664Q197,634 197,571L197,536L332,536L332,468Z"
-                       "M474,737Q494,737 510,724Q525,710 525,681Q525,653 510,639"
+                       "Q240,693 218,663Q197,634 197,571L197,536L332,536L332,468Z"
+                       "M474,737Q494,737 509,723Q525,710 525,681Q525,653 509,639"
                        "Q494,625 474,625Q452,625 437,639Q422,653 422,681"
-                       "Q422,710 437,724Q452,737 474,737ZM517,536L517,0L429,0L429,536L517,536Z";
+                       "Q422,710 437,723Q452,737 474,737ZM517,536L517,0L429,0L429,536L517,536Z";
       g_assert_cmpmem (str, draw_data.consumed, expected, sizeof (expected) - 1);
     }
     {
       draw_data.consumed = 0;
       hb_font_get_glyph_shape (font, 5, funcs, &draw_data);
-      char expected[] = "";
+      char expected[] = "M15,0Q15,0 15,0Z";
       g_assert_cmpmem (str, draw_data.consumed, expected, sizeof (expected) - 1);
     }
     {
       draw_data.consumed = 0;
       hb_font_get_glyph_shape (font, 6, funcs, &draw_data);
       char expected[] = "M346,468L211,468L211,0L123,0L123,468L29,468L29,509L123,539"
-                       "L123,570Q123,674 169,720Q215,765 297,765Q329,765 356,760"
+                       "L123,570Q123,674 169,719Q215,765 297,765Q329,765 355,759"
                        "Q382,754 401,747L378,678Q362,683 341,688Q320,693 298,693"
-                       "Q254,693 233,664Q211,634 211,571L211,536L346,536L346,468Z";
+                       "Q254,693 232,663Q211,634 211,571L211,536L346,536L346,468Z"
+                       "M15,0Q15,0 15,0Z";
       g_assert_cmpmem (str, draw_data.consumed, expected, sizeof (expected) - 1);
     }
 
```

behdad: Thanks. Can you push that as a commit please?

ebraminio: Done.

test/api/test-draw.c
```diff
@@ -870,26 +884,183 @@ test_hb_draw_stroking (void)
...
 		       "C731,558 582,708 397,708C212,708 62,558 62,373Z";
-    g_assert_cmpmem (str, user_data.consumed, expected2, sizeof (expected2) - 1);
+    g_assert_cmpmem (str, draw_data.consumed, expected2, sizeof (expected2) - 1);
 
     hb_font_destroy (font);
   }
 }
 
+static void
+test_hb_draw_drawing_fo_funcs (void)
```
ebraminio, line 908: drawing_fo? misspell I'd guess

ebraminio:
Also added a test for ft callback https://github.com/harfbuzz/harfbuzz/pull/3411/commits/0429921c109a24205ed19e21bc5f0d1cd570400d feel free to just remove the commit from history if you don't see fit.

Also any interest on having binding to coretext/gdi/directwrite https://github.com/harfbuzz/harfbuzz/pull/3411#issuecomment-1029240565 somewhere somehow? Just though it could be nice for the sake of debugging and so later.

behdad:
> Also any interest on having binding to coretext/gdi/directwrite [#3411 (comment)](https://github.com/harfbuzz/harfbuzz/pull/3411#issuecomment-1029240565) somewhere somehow? Just though it could be nice for the sake of debugging and so later.

If you are willing to implement the rest of the font-funcs callbacks for them... In 10+ years no one bothered / cared / requested.

behdad:
@ebraminio why did the linux-ci bot fail?

ebraminio:
> why did the linux-ci bot fail?

My local freetype version is 24 and the CI one is 21 and that apparently means different results...

I swapped the test with a simpler one case just to test line_to/quadratic_to commands are emitted correctly.

behdad:
Would be nice to add a `hb-draw` commandline tool also, that spews SVG-like draw commands given glyphnames...

ebraminio:
> Would be nice to add a hb-draw commandline tool also, that spews SVG-like draw commands given glyphnames...

I wish it could create a SVG from input text without Cairo, say like https://harfbuzz.github.io/harfbuzzjs/ demo which creates SVG from input text, I needed accurate rendering of some text in SVG to upload to Wikipedia and shaped the tool for the purpose which creates relative SVG path also to have more optimized yet accurate result.

Just to note main.cc also creates SVG for individual glyphs of a font when hb-draw is enabled and supports COLRv0/SVG/png dumping also.

behdad:
> > Would be nice to add a hb-draw commandline tool also, that spews SVG-like draw commands given glyphnames...
> 
> I wish it could create a SVG from input text without Cairo, say like https://harfbuzz.github.io/harfbuzzjs/ demo which creates SVG from input text, I needed accurate rendering of some text in SVG to upload to Wikipedia and shaped the tool for the purpose which creates relative SVG path also to have more optimized yet accurate result.

Maybe we can spec and develop that separately then. I like to merge this soon.

behdad:
https://github.com/behdad/glyphy/pull/39

I ported GLyphy to use hb-draw:
image

yisibl:
> I wish it could create a SVG from input text without Cairo, say like https://harfbuzz.github.io/harfbuzzjs/ demo which creates SVG from input text, I needed accurate rendering of some text in SVG to upload to Wikipedia and shaped the tool for the purpose which creates relative SVG path also to have more optimized yet accurate result.
>> Maybe we can spec and develop that separately then. I like to merge this soon.

@behdad Any progress since then?

behdad:
> @behdad Any progress since then?

No. Can you use the hb-view SVG output and massage it through `svgopt` or other tools?

PR #3413, opened 2022-02-04
[ci] Try to fix MSVC build failure

khaledhosny:


behdad:
Life used to be lighter...

khaledhosny:
> Life used to be lighter...

Indeed. We are supposedly building specific versions of everything to avoid stuff like this, yet the ground keeps changing under our feet!

khaledhosny:
OK, this started failing when GitHub deployed a new virtual environment image, whatever this is seems no have changed something and some meson tool is now silently failing.

khaledhosny:
I give up, the command `meson --internal symbolextractor` is always failing on these CI machines and it does not seem to produce any error messages.

khaledhosny:
I guess we have to wait for this one to fix itself.

PR #3414, opened 2022-02-04
[repacker] expose hb_subset_repack() API, hb_object_t and hb_link_t structs

qxliu76:


src/hb-serialize.hh
```diff
@@ -70,6 +70,40 @@ struct hb_serialize_context_t
...
+
+    object_t (const cy_object_t &o)
+    {
+      head = o.head;
+      tail = o.tail;
+      next = nullptr;
+      real_links.alloc (o.num_real_links);
+      for (unsigned i = 0 ; i < o.num_real_links; i++)
+      {
+        link_t l;
```
garretrieger, line 83: You could add a copy constructor to link_t that takes a cy_link_t then this copy code doesn't need to be duplicated here and in the next loop

src/hb-subset-repacker.cc
```diff
@@ -0,0 +1,37 @@
...
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ * FITNESS FOR A PARTICULAR PURPOSE.  THE SOFTWARE PROVIDED HEREUNDER IS
+ * ON AN "AS IS" BASIS, AND THE COPYRIGHT HOLDER HAS NO OBLIGATION TO
+ * PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
+ *
+ */
+
+#include "hb-repacker.hh"
+
+hb_blob_t* hb_subset_repack (cy_object_t* cy_objects, unsigned num_cy_objs)
```
garretrieger, line 28: Since this is a public method it should have a comment explaining what it does. See hb_subset_or_fail in hb_subset.cc for an example of how to format the comment.

src/hb-subset-repacker.h
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+#ifndef HB_SUBSET_REPACKER_H
+#define HB_SUBSET_REPACKER_H
+
+#include "hb.h"
+
+HB_BEGIN_DECLS
+
+struct cy_link_t
```
garretrieger, line 33: I would drop the "cy_" prefix and most likely you should used a "hb_" prefix to be consistent with the other public harfbuzz types. So hb_link_t and hb_object_t.

garretrieger: Also add a comment describing what the fields of these structs are.

src/hb-subset-repacker.cc
```diff
@@ -0,0 +1,47 @@
...
+ * hb_subset_repack:
+ * @hb_objects: raw array of struct hb_object_t, which provides
+ * object graph info
+ * @num_hb_objs: number of hb_object_t in the hb_objects array.
+ *
+ * Given the input object graph info, repack a table to eliminate
+ * offset overflows. A nullptr is returned if the repacking attempt fails.
+ *
+ **/
+hb_blob_t* hb_subset_repack (hb_object_t* hb_objects, unsigned num_hb_objs)
```
garretrieger, line 38: Change the method name to hb_subset_repack_or_fail. That's the convention we use to indicate a method may return nullptr.

garretrieger (approved):

src/hb-subset-repacker.cc
```diff
@@ -0,0 +1,47 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE.  THE SOFTWARE PROVIDED HEREUNDER IS
+ * ON AN "AS IS" BASIS, AND THE COPYRIGHT HOLDER HAS NO OBLIGATION TO
+ * PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
+ *
+ */
+
+#include "hb-repacker.hh"
+
+/**
+ * hb_subset_repack:
```
khaledhosny, line 29: This need to match the function name, and be added to the relevant section in `docs/harfbuzz-sections.txt`.

behdad:
I'm not comfortable always exposing this. A conditional way is preferred.

garretrieger:
> I'm not comfortable always exposing this. A conditional way is preferred.

 Would you prefer wrapping in a #ifdef or should we set up something in the build configuration to not even include hb-subset-repacker.h/cc in the build?

behdad:
> Would you prefer wrapping in a #ifdef or should we set up something in the build configuration to not even include hb-subset-repacker.h/cc in the build?

Ifdef can only take care of the .h, not the symbol in the binary. So we need a build configuration, since anything that goes in the ABI we have to support indefinitely (although we can call it deprecated).

behdad:
The alternative is to create a new library for the repacker and call it API-unstable and do whatever you want. But I feel like a build switch is easiest.

behdad:
We can resurrect the "experimental-api" build config that I was removing in https://github.com/harfbuzz/harfbuzz/pull/3411

behdad:
Yeah I think you can flag it behind HB_EXPERIMENTAL_API.

garretrieger:
Using experimental api sounds good.

src/hb-subset-repacker.h
```diff
@@ -0,0 +1,68 @@
...
+#define HB_SUBSET_REPACKER_H
+
+#include "hb.h"
+
+HB_BEGIN_DECLS
+
+#ifdef HB_EXPERIMENTAL_API
+struct hb_link_t
+{
+  unsigned width;       //offsetSize in bytes
```
behdad, line 36: No C++-style comments in .h file.

behdad:
Thanks

Issue #3396, opened 2022-01-27
ot-math: Provide some way to get all defined MathKern correction heights


lexi-lambda:
The [`hb_ot_math_get_glyph_kerning`](https://harfbuzz.github.io/harfbuzz-hb-ot-math.html#hb-ot-math-get-glyph-kerning) function allows querying a font for a math kern value at a particular correction height. However, it does not provide any way to determine which correction heights at actually defined, and therefore provides no way to efficiently determine where the discontinuities are.

I am working on an OpenType math renderer, and I would like to be able to get access to this information so that my renderer can provide a debugging visualization of the defined cut-in regions for a particular glyph. As far as I can tell, the only way to do this using the current API is to evaluate some very large number of correction heights by calling `hb_ot_math_get_glyph_kerning` once per pixel along the rendered glyph’s y axis, which is obviously not particularly efficient.

Ideally, an API to request this information would look something like [`hb_ot_math_get_glyph_variants`](https://harfbuzz.github.io/harfbuzz-hb-ot-math.html#hb-ot-math-get-glyph-variants) and would allow extracting the entire set of correction heights and corresponding kerning values at once. The interface would look something like this:

```c
unsigned int
hb_ot_math_get_glyph_kernings (hb_font_t *font,
                               hb_codepoint_t glyph,
                               hb_ot_math_kern_t kern,
                               unsigned int *kern_entry_count,
                               hb_ot_math_kern_entry_t *kern_entries);

typedef struct {
  hb_position_t correction_height;
  hb_position_t kern_value;
} hb_ot_math_kern_entry_t;
```

I would be happy to put together a PR to add such a function if this proposal sounds reasonable.

behdad:
Sgtm. Thank you in advance. :)

khaledhosny:
Would it be simpler to have to `hb_position_t` arrays for output instead of the new struct? I don't think we have any API that takes caller-allocated structs for output.

behdad:
> I don't think we have any API that takes caller-allocated structs for output.

`hb_ot_var_get_axis_infos`

behdad:
> ```c
> unsigned int
> hb_ot_math_get_glyph_kernings (hb_font_t *font,
>                                hb_codepoint_t glyph,
>                                hb_ot_math_kern_t kern,
Add `unsigned int start_offset` at this position like other API that fill an array.
>                                unsigned int *kern_entry_count,
>                                hb_ot_math_kern_entry_t *kern_entries);
[...]

PR #3416, opened 2022-02-05
[ot-math] Add hb_ot_math_get_glyph_kernings

lexi-lambda:
This PR implements `hb_ot_math_get_glyph_kernings`, as discussed in #3396.

As described in https://github.com/harfbuzz/harfbuzz/issues/3396#issuecomment-1028299055, designing an API to get this information is awkward, as there is one more kern value than there are correction heights. However, I ended up going with an API closer to the one I initially proposed, which looks like this:

```c
typedef struct {
  hb_position_t max_correction_height;
  hb_position_t kern_value;
} hb_ot_math_kern_entry_t;

unsigned int
hb_ot_math_get_glyph_kernings (hb_font_t *font,
                               hb_codepoint_t glyph,
                               hb_ot_math_kern_t kern,
                               unsigned int start_offset,
                               unsigned int *entries_count, /* IN/OUT */
                               hb_ot_math_kern_entry_t *kern_entries /* OUT */)
```
Since the last entry’s `max_correction_height` field does not correspond to any value provided by the font, it will always be set to `INT32_MAX`. This is a little aesthetically unsatisfying, but after experimenting with other options, I think this choice is the least headache-inducing.

### DIFF
--- a/src/hb-ot-math-table.hh
+++ b/src/hb-ot-math-table.hh
@@ -369,6 +369,37 @@ struct MathKern
     return kernValue[i].get_x_value (font, this);
   }
 
+  unsigned int get_entries (unsigned int start_offset,
+			    unsigned int *entries_count, /* IN/OUT */
+			    hb_ot_math_kern_entry_t *kern_entries, /* OUT */
+			    hb_font_t *font) const
+  {
+    const MathValueRecord* correctionHeight = mathValueRecordsZ.arrayZ;
+    const MathValueRecord* kernValue = mathValueRecordsZ.arrayZ + heightCount;
+    const unsigned int entriesCount = heightCount + 1;
+
+    if (entries_count)
+    {
+      unsigned int start = hb_min (start_offset, entriesCount);
+      unsigned int end = hb_min (start + *entries_count, entriesCount);
+      *entries_count = end - start;
+
+      for (unsigned int i = 0; i < *entries_count; i++) {
+	unsigned int j = start + i;
+
+	hb_position_t max_height;
+	if (j == heightCount) {
+	  max_height = INT32_MAX;
+	} else {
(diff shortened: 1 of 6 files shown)


behdad:
This is good and how I would have done it. Thank you.

src/hb-ot-math.h
```diff
@@ -204,6 +204,20 @@ typedef enum {
...
 } hb_ot_math_kern_t;
 
+/**
+ * hb_ot_math_kern_entry_t:
+ * @max_correction_height: The maximum height at which this entry should be used
+ * @kern_value: The kern value of the entry
+ *
+ * Data type to hold math kerning (cut-in) information for a glyph.
+ *
+ * Since: 3.4.0
```
khaledhosny, line 214: Use, `REPLACEME` for the version number, we will update it when cutting the release with the actual release number.

lexi-lambda: Thanks, now fixed.

src/hb-ot-math-table.hh
```diff
@@ -423,6 +454,25 @@ struct MathKernInfoRecord
...
 
+  unsigned int get_kernings (hb_ot_math_kern_t kern,
+			     unsigned int start_offset,
+			     unsigned int *entries_count, /* IN/OUT */
+			     hb_ot_math_kern_entry_t *kern_entries, /* OUT */
+			     hb_font_t *font,
+			     const void *base) const
+  {
+    unsigned int idx = kern;
+    if (unlikely (idx >= ARRAY_LENGTH (mathKern))) return 0;
```
behdad, line 465: Doesn't this also have to set `entries_count` before leaving?

behdad:
Thanks!

PR #3417, opened 2022-02-06
[ot-tag] Add HB_SCRIPT_MATH (Zmth) and map it to OT ‘math’ tag

lexi-lambda:
The [ISO 15924 code](https://unicode.org/iso15924/iso15924-codes.html) for mathematical notation is `Zmth`, but the corresponding [OpenType script tag](https://docs.microsoft.com/en-us/typography/opentype/spec/scripttags) is `math`. HarfBuzz needs to know about the mapping, or else OpenType features that depend on the `math` script won’t be applied. This PR adds `HB_SCRIPT_MATH` to `hb_script_t` (for convenience) and changes `hb_ot_tags_from_script_and_language` and `hb_ot_tag_to_script` to do the necessary conversion.

### DIFF
--- a/src/hb-common.h
+++ b/src/hb-common.h
@@ -481,6 +481,7 @@ hb_language_get_default (void);
  * @HB_SCRIPT_TANGSA: `Tnsa`, Since: 3.0.0
  * @HB_SCRIPT_TOTO: `Toto`, Since: 3.0.0
  * @HB_SCRIPT_VITHKUQI: `Vith`, Since: 3.0.0
+ * @HB_SCRIPT_MATH: `Zmth`, Since: REPLACEME
  * @HB_SCRIPT_INVALID: No script set
  *
  * Data type for scripts. Each #hb_script_t's value is an #hb_tag_t corresponding
@@ -696,6 +697,11 @@ typedef enum
   HB_SCRIPT_TANGSA			= HB_TAG ('T','n','s','a'), /*14.0*/
   HB_SCRIPT_TOTO			= HB_TAG ('T','o','t','o'), /*14.0*/
   HB_SCRIPT_VITHKUQI			= HB_TAG ('V','i','t','h'), /*14.0*/
+
+  /*
+   * Since REPLACEME
+   */
+  HB_SCRIPT_MATH			= HB_TAG ('Z','m','t','h'),
 
   /* No script set. */
   HB_SCRIPT_INVALID			= HB_TAG_NONE,
--- a/src/hb-ot-tag.cc
+++ b/src/hb-ot-tag.cc
@@ -41,6 +41,7 @@ hb_ot_old_tag_from_script (hb_script_t script)
   switch ((hb_tag_t) script)
   {
     case HB_SCRIPT_INVALID:		return HB_OT_TAG_DEFAULT_SCRIPT;
+    case HB_SCRIPT_MATH:		return HB_OT_TAG_MATH_SCRIPT;
 
     /* KATAKANA and HIRAGANA both map to 'kana' */
     case HB_SCRIPT_HIRAGANA:		return HB_TAG('k','a','n','a');
@@ -63,6 +64,8 @@ hb_ot_old_tag_to_script (hb_tag_t tag)
 {
   if (unlikely (tag == HB_OT_TAG_DEFAULT_SCRIPT))
     return HB_SCRIPT_INVALID;
+  if (unlikely (tag == HB_OT_TAG_MATH_SCRIPT))
+    return HB_SCRIPT_MATH;
 
   /* This side of the conversion is fully algorithmic. */
 
--- a/docs/harfbuzz-sections.txt
+++ b/docs/harfbuzz-sections.txt
@@ -191,6 +191,7 @@ HB_SCRIPT_CANADIAN_ABORIGINAL
 hb_font_funcs_set_glyph_func
 hb_font_get_glyph_func_t
 HB_MATH_GLYPH_PART_FLAG_EXTENDER
+HB_OT_MATH_SCRIPT
 hb_ot_layout_table_choose_script
 hb_ot_layout_table_find_script
 hb_ot_tag_from_language
@@ -542,7 +543,7 @@ hb_ot_layout_glyph_sequence_func_t
 <SECTION>
 <FILE>hb-ot-math</FILE>
 HB_OT_TAG_MATH
-HB_OT_MATH_SCRIPT
+HB_OT_TAG_MATH_SCRIPT
 hb_ot_math_constant_t
 hb_ot_math_kern_t
 hb_ot_math_glyph_variant_t
--- a/src/hb-ot-deprecated.h
+++ b/src/hb-ot-deprecated.h
@@ -50,6 +50,21 @@ HB_BEGIN_DECLS
  */
 #define HB_MATH_GLYPH_PART_FLAG_EXTENDER HB_OT_MATH_GLYPH_PART_FLAG_EXTENDER
 
+/* https://github.com/harfbuzz/harfbuzz/pull/3417 */
+/**
+ * HB_OT_MATH_SCRIPT:
+ *
+ * Use #HB_SCRIPT_MATH or #HB_OT_TAG_MATH_SCRIPT instead.
+ *
+ * <note>Previous versions of this documentation recommended passing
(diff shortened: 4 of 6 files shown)


src/hb-common.h
```diff
@@ -481,6 +481,7 @@ hb_language_get_default (void);
  * @HB_SCRIPT_TANGSA: `Tnsa`, Since: 3.0.0
  * @HB_SCRIPT_TOTO: `Toto`, Since: 3.0.0
  * @HB_SCRIPT_VITHKUQI: `Vith`, Since: 3.0.0
+ * @HB_SCRIPT_MATH: `Zmth`, Since 3.4.0
```
khaledhosny, line 484: Use `REPLACEME` for the version number.

behdad:
Looks good except for what Khaled said. Thanks.

src/hb-common.h
```diff
@@ -545,6 +546,7 @@ typedef enum
   HB_SCRIPT_HANUNOO			= HB_TAG ('H','a','n','o'), /*3.2*/
   HB_SCRIPT_TAGALOG			= HB_TAG ('T','g','l','g'), /*3.2*/
   HB_SCRIPT_TAGBANWA			= HB_TAG ('T','a','g','b'), /*3.2*/
+  HB_SCRIPT_MATH			= HB_TAG ('Z','m','t','h'), /*3.2*/
```
dscorbett, line 549: These constants are grouped by which HarfBuzz version they were released in, so `HB_SCRIPT_MATH` should be added in a new section, right before `HB_SCRIPT_INVALID`, with a “Since” comment like the other sections have.

dscorbett (changes requested):

src/hb-common.h
behdad: They're grouped by the Unicode version they are added in. In this case, let's group it before UNKNOWN, with an empty Unicode version number.

behdad: Ignore my comment. What David said.

dscorbett:
As I understand it, 'math' is unlike other script tags in that it does not correspond to any specific characters. Instead, it is used to identify features relevant to the MATH table. Is 'math' ever used for non-MATH shaping? If not, is the existing constant `HB_OT_MATH_SCRIPT` sufficient?

behdad:
I guess specifying the `HB_OT_MATH_SCRIPT` does find the right OT script tag after all, but if `hb_shape()` is involved, I don't mind adding what is proposed here.

lexi-lambda:
@dscorbett I had missed the existence of `HB_OT_MATH_SCRIPT`, so thank you for pointing me to it—I’ve updated the definition of `hb_ot_old_tag_to_script` to use that constant. And yes, you are correct that `Zmth` is not a “real” Unicode script and is only relevant for OpenType shaping (though it *is* a real ISO 15924 code).

However, I’d still like this change. In Pango, there’s no way to arrange for an arbitrary `hb_script_t` to be passed to `hb_buffer_set_script`, so there’s no way to easily use `HB_OT_MATH_SCRIPT` via that interface. Instead, Pango uses glib’s `GUnicodeScript` enumeration, which contains a fixed set of values that are mapped to ISO 15924 codes via `g_unicode_script_to_iso15924`. Since `HB_OT_MATH_SCRIPT` is not a valid ISO 15924 code, it would not make sense for an entry of `GUnicodeScript` to ever produce it.

However, one potential alternative approach to this PR would be to make the change to `hb_script_from_iso15924_tag` and `hb_script_to_iso15924_tag` instead of `hb_ot_old_tag_from_script` and `hb_ot_old_tag_to_script`, in which case the internal `hb_script_t` representation of the ISO 15924 `Zmth` code could be `HB_OT_MATH_SCRIPT`. That seemed less preferable to me, since in general I got the sense that `hb_script_t` values adhere more closely to ISO 15924 codes than OpenType script tags, but if you think that way would be better, I’d be happy to do it, instead.

dscorbett:
That sounds like a good reason to merge this. The reason does not apply to 'byzm' and 'musc', the other tags for non-Unicode scripts; I was wondering if we should add constants for them too, but that does not seem as necessary.

Did you mean to squash all those commits with “(squash)” in the message?

src/hb-ot-tag.cc
```diff
@@ -45,6 +45,8 @@ hb_ot_old_tag_from_script (hb_script_t script)
     /* KATAKANA and HIRAGANA both map to 'kana' */
     case HB_SCRIPT_HIRAGANA:		return HB_TAG('k','a','n','a');
 
+    case HB_SCRIPT_MATH:		return HB_TAG('m','a','t','h');
```
dscorbett, line 48: `hb_ot_old_tag_to_script` uses `HB_OT_MATH_SCRIPT` so this should too.

dscorbett (changes requested):

dscorbett:
The documentation of `HB_OT_MATH_SCRIPT` will be wrong once this is merged.

https://github.com/harfbuzz/harfbuzz/blob/ac46c3248e8b0316235943175c4d4a11c24dd4a9/src/hb-ot-math.h#L53-L56

Users should pass `HB_SCRIPT_MATH` to `hb_buffer_set_script` instead of `HB_OT_MATH_SCRIPT`. The latter might still work, but I agree it is better to use the value that matches ISO 15924 like all the other `hb_script_t` values. @behdad, should `HB_OT_MATH_SCRIPT` therefore be deprecated?

src/hb-ot-tag.cc
lexi-lambda: Thanks, corrected.

behdad:
> . @behdad, should `HB_OT_MATH_SCRIPT` therefore be deprecated?

It feels to me that way, yes.

lexi-lambda:
> Did you mean to squash all those commits with “(squash)” in the message?

No, I decided to leave them unsquashed until merging to make it easier for reviewers to see what changed. But based on your comment, it seems like maybe that was more confusing than helpful, so I’ve just gone and squashed them now. :)

lexi-lambda:
> should `HB_OT_MATH_SCRIPT` therefore be deprecated?

I’m not sure that it needs to be deprecated given `HB_OT_TAG_DEFAULT_SCRIPT` exists and is non-deprecated, even though it isn’t a valid `hb_script_t`. The name is perhaps a little unfortunate (it would be better to have been named `HB_OT_TAG_MATH_SCRIPT` for consistency), but I don’t think that’s a big deal.

I agree that the documentation should be updated, though, so I’ll do that.

lexi-lambda:
Alright, I’ve updated the documentation to the following:

https://github.com/harfbuzz/harfbuzz/blob/ac4de6b5145f25d2c20909f135a86e78980c328c/src/hb-ot-math.h#L52-L65

dscorbett (approved):

khaledhosny (approved):

behdad:
Please move `HB_OT_MATH_SCRIPT` to `hb-ot-deprecated.h` and add a `Deprecated: REPLACEME` tag. The wording of "Previous versions of this doccumentation" is good and can remain but can be made more concise. Thanks.

lexi-lambda:
I think deprecating `HB_OT_MATH_SCRIPT` and suggesting unconditionally replacing it with `HB_SCRIPT_MATH` is unhelpful, as it is still useful in situations where a raw OpenType script tag is expected, such as in the `scripts` argument to `hb_ot_layout_collect_features`. Therefore, I’ve pushed a revision that makes two changes:

1. I’ve added a new constant to `hb-ot-math.h`, `HB_OT_TAG_MATH_SCRIPT`, which has the old value of `HB_OT_MATH_SCRIPT`. The documentation notes that `HB_SCRIPT_MATH` should be used where an `hb_script_t` is expected.

2. I’ve moved `HB_OT_MATH_SCRIPT` to `hb-ot-deprecated.h` and made it an alias for `HB_OT_TAG_MATH_SCRIPT`, and I’ve updated its documentation to recommend `HB_SCRIPT_MATH` or `HB_OT_TAG_MATH_SCRIPT`, instead. I’ve also pared down the historical `<note>` to be more succinct.

Let me know if you think this is reasonable. I suspect, in practice, it would be no great loss if `HB_OT_TAG_MATH_SCRIPT` were not included, but it provides nice parallelism with both `HB_OT_TAG_DEFAULT_SCRIPT` and `HB_OT_TAG_MATH`, so I think it makes sense to include it.

PR #3424, opened 2022-02-10
hb-algs.hh: Fix build on Visual Studio 2015

fanc999:
Hi,

This attempts to fix the build for the tip of the main branch when Visual Studio 2015 is being used, as a `static_assert` (error C2338) is triggered during the build, where the compiler does not recognize the `std::hash` call on `hb_decay<decltype (hb_deref (v)>`, so give it a shove so that it is recognized.

With blessings, thank you!

### DIFF
--- a/src/hb-algs.hh
+++ b/src/hb-algs.hh
@@ -226,8 +226,14 @@ struct
   template <typename T> constexpr auto
   impl (const T& v, hb_priority<2>) const HB_RETURN (uint32_t, hb_deref (v).hash ())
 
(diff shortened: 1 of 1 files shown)


fanc999:
Hi @behdad,

Yup, I agree it sucks, so I added specializations in PR 3448 that reverts the things here and added the needed specializations for `std::hash`.

With blessings, thank you!

behdad:
But that looks wrong. The std::hash is not expected to match any types in the library in fact.

Issue #3140, opened 2021-08-12
hb-subset feature request: ability to retrieve glyph id remapping


bungeman:
As things currently stand, the public API for subsetting allows one to specify which codepoints and glyphs to keep. If `HB_SUBSET_FLAGS_RETAIN_GIDS` is set the glyph ids are maintained, otherwise the glyph ids will be re-assigned somewhat arbitrarily in the subset font. The main reason for the existence of `HB_SUBSET_FLAGS_RETAIN_GIDS` is that a user may need to write out many glyph ids before subsetting and cannot currently know in advance what the subset glyph ids may be without setting this flag. (Related, the is no API for getting the `reverse_glyph_map` after mapping, so it isn't straight forward to do the mapping without setting this flag even after subsetting.) However, if the user could control the remapping of the glyph ids then it could make the assignments itself as it went along (mapping the first glyph id it sees from an input font to glyph 4 in the output font, the next to 5, etc).

This is a request for something like `hb_map* hb_subset_input_glyph_map()` which works something like `hb_subset_input_glyph_set` in that the keys of this map are input glyph ids to retain, but the values are the associated subset font glyph ids (if value is repeated at any point the result is failure or one or the other wins). This would then influence `_create_old_gid_to_new_gid_map` when `hb_subset_plan_create` is called, which appears to be the only actual code which would need to be modified.

This would allow the user to output `subset` glyph ids before actually creating the subset font while simply maintaining the hb_map as it goes. Then the subset font can be smaller (especially with large CJK fonts) since the overhead of `loca` and such can be elided. If for some reason the subset failed the user would need fail (or go back and start over again) but this is unlikely since in general the font was just used by HarfBuzz to shape the glyphs.

behdad:
Very good point. Yes, we can do that.

Some tables depend on the relative order of gid's. We should make sure such tables are correctly typeset. The ones I have in mind involve OT Layout Coverage and ClassDef structures.

And yes providing a mapping sounds good.

garretrieger:
Internally the subset plan maintains a glyph map from new to old which the subsetting code then uses to rewrite gids, so on the surface this could be pretty easy to support. However, I suspect that we may have instances of code that were written with the assumption that the glyph map always retains the relative order of the glyphs between the original and the subset.
So if we were to do this we may want to add some tests that attempt to locate parts of the subsetter operating under this assumption.

bungeman:
[...]

PR #3426, opened 2022-02-11
subset: add get_new_gid_for_old_gid

njzhangyifei:
My naive attempt of exposing old->new gid mapping.

For https://github.com/harfbuzz/harfbuzz/issues/3140 
@bungeman @garretrieger


behdad:
We can't do this before subset is planned...

garretrieger:
> We can't do this before subset is planned...

This is grabbing the mapping from after plan generation, but attaching it to the input for access after the subset operation. So it will work however the preferred approach is to expose the subset plan object. I'm going to make a first pass at doing that today.

njzhangyifei:
Cool. I was trying to reuse the input struct because plan is still an internal API but exposing that sounds good.
Thanks Garret for taking a look!

PR #3429, opened 2022-02-11
[subset] expose subset plan in public subsetting API

garretrieger:
For #3140 

Provides access to the subset plan so callers can find information on things such as the glyph mapping from the subset.

### DIFF
--- a/src/hb-subset-plan.cc
+++ b/src/hb-subset-plan.cc
@@ -458,7 +458,7 @@ _nameid_closure (hb_face_t *face,
 }
 
 /**
- * hb_subset_plan_create:
+ * hb_subset_plan_create_or_fail:
  * @face: font face to create the plan for.
  * @input: a #hb_subset_input_t input.
  *
@@ -467,17 +467,18 @@ _nameid_closure (hb_face_t *face,
  * which tables and glyphs should be retained.
  *
  * Return value: (transfer full): New subset plan. Destroy with
- * hb_subset_plan_destroy().
+ * hb_subset_plan_destroy(). If there is a failure creating the plan
+ * nullptr will be returned.
  *
  * Since: REPLACEME
  **/
 hb_subset_plan_t *
-hb_subset_plan_create (hb_face_t	 *face,
-		       const hb_subset_input_t *input)
+hb_subset_plan_create_or_fail (hb_face_t	 *face,
+                               const hb_subset_input_t *input)
 {
   hb_subset_plan_t *plan;
   if (unlikely (!(plan = hb_object_create<hb_subset_plan_t> ())))
-    return const_cast<hb_subset_plan_t *> (&Null (hb_subset_plan_t));
+    return nullptr;
 
   plan->successful = true;
   plan->flags = input->flags;
@@ -514,8 +515,8 @@ hb_subset_plan_create (hb_face_t	 *face,
   plan->layout_variation_indices = hb_set_create ();
   plan->layout_variation_idx_map = hb_map_create ();
 
-  if (plan->in_error ()) {
-    return plan;
+  if (unlikely (plan->in_error ())) {
+    return nullptr;
   }
 
   _populate_unicodes_to_retain (input->sets.unicodes, input->sets.glyphs, plan);
@@ -532,6 +533,9 @@ hb_subset_plan_create (hb_face_t	 *face,
 				  plan->reverse_glyph_map,
 				  &plan->_num_output_glyphs);
 
+  if (unlikely (plan->in_error ())) {
+    return nullptr;
+  }
   return plan;
 }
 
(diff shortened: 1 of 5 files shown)


behdad:
To me not reusing a hb-map in the API would be a mistake. We should just return a hb-map and let that API be reused.

garretrieger:
> To me not reusing a hb-map in the API would be a mistake. We should just return a hb-map and let that API be reused.

Sounds good. I'll update this.

src/hb-subset.h
```diff
@@ -145,6 +154,28 @@ hb_subset_input_set_flags (hb_subset_input_t *input,
 HB_EXTERN hb_face_t *
 hb_subset_or_fail (hb_face_t *source, const hb_subset_input_t *input);
 
+HB_EXTERN hb_face_t *
+hb_subset_from_plan_or_fail (hb_subset_plan_t *plan);
```
behdad, line 158: Borrowing name from shape_plan would suggest:
```
hb_subset_plan_execute_or_fail
```

src/hb-subset.h
```diff
@@ -145,6 +154,28 @@ hb_subset_input_set_flags (hb_subset_input_t *input,
...
+
+HB_EXTERN hb_subset_plan_t *
+hb_subset_plan_create (hb_face_t           *face,
+		       const hb_subset_input_t   *input);
+
+HB_EXTERN void
+hb_subset_plan_destroy (hb_subset_plan_t *plan);
```
behdad, line 165: Please `_reference`, `_set_user_data`, `_get_user_data`.

src/hb-subset.h
```diff
@@ -145,6 +154,25 @@ hb_subset_input_set_flags (hb_subset_input_t *input,
 HB_EXTERN hb_face_t *
 hb_subset_or_fail (hb_face_t *source, const hb_subset_input_t *input);
 
+HB_EXTERN hb_face_t *
+hb_subset_from_plan_or_fail (hb_subset_plan_t *plan);
+
+HB_EXTERN hb_subset_plan_t *
+hb_subset_plan_create (hb_face_t           *face,
```
behdad, line 161: Probably also `_create_or_fail` instead of `_create`.

garretrieger: plan_create always returns a valid pointer. It uses an in_error() to check for failure.

behdad: Would it be useful for our purposes though?

garretrieger: Yeah I could see that (nullptr) being more intuitive for end users. I'll change create to be or_fail.

src/hb-subset.cc
```diff
@@ -343,17 +343,38 @@ hb_subset_or_fail (hb_face_t *source, const hb_subset_input_t *input)
...
+  }
+
+  hb_face_t * result = hb_subset_plan_execute_or_fail (plan);
+  hb_subset_plan_destroy (plan);
+  return result;
+}
+
+
+/**
+ * hb_subset_from_plan_or_fail:
```
khaledhosny, line 358: This needs to match the actual function name.

src/hb-subset.cc
```diff
@@ -343,17 +343,38 @@ hb_subset_or_fail (hb_face_t *source, const hb_subset_input_t *input)
...
+ * @plan: a subsetting plan.
+ *
+ * Subsets a font according to provided plan. Returns nullptr
+ * if the subset operation fails.
+ *
```
khaledhosny, line 363: The return value needs to be documented as well.

src/hb-subset-plan.cc
```diff
@@ -596,3 +602,107 @@ hb_subset_plan_destroy (hb_subset_plan_t *plan)
...
 }
+
+/**
+ * hb_subset_plan_old_to_new_glyph_mapping:
+ * @plan: a subsetting plan.
+ *
+ * Returns the mapping between glyphs in the original font to glyphs
+ * in the subset that will be produced by @plan.
+ *
+ * Since: REPLACEME
```
khaledhosny, line 613: The return value needs to be documented (with `Return value: ` syntax)

behdad:
Lgtm

src/hb-subset-plan.cc
```diff
@@ -596,3 +602,108 @@ hb_subset_plan_destroy (hb_subset_plan_t *plan)
...
+ * hb_subset_plan_old_to_new_glyph_mapping:
+ * @plan: a subsetting plan.
+ *
+ * Returns the mapping between codepoints in the original font and the
+ * associated glyph id in the original font.
+ *
+ * Since: REPLACEME
+ **/
+const hb_map_t*
+hb_subset_plan_codepoint_to_old_glyph_mapping (const hb_subset_plan_t *plan)
```
behdad, line 647: Should we even return this? This is available from the original face readily, right?

garretrieger: In the case where the subset was specified using only glyph id's this provides a method to determine which codepoints are going to be in the final subset.

behdad: I see. Sure.

PR #3432, opened 2022-02-13
[fuzz-shape] Verify shape output

behdad:
Let the fuzzers loose on shape verify.


behdad:
This is the current failing (of one of the in-tree fuzzing fonts), and I have not made any progress as to why / where:
```
 ninja && util/hb-shape  ../test/fuzzing/fonts/clusterfuzz-testcase-minimized-hb-fuzzer-5303930168803328 -u 'U+0041|U+0042|U+0043|U+0044|U+0045|U+0058|U+0059|U+005A|U+0031|U+0032|U+0033|U+0040|U+005F' --show-flags --verify --trace | less
```

behdad:
Going to merge this, although I cannot get the fuzzer tests to pass all verify modes...

PR #3435, opened 2022-02-14
[ot-layout] Add hb_ot_layout_get_horizontal_baseline_tag_for_script()

behdad:
Just a starting point. Many more Brahmi-derived scripts need the hanging baseline I think, and many more the ideographic baseline?


src/hb-ot-layout.cc
```diff
@@ -1959,6 +1959,36 @@ hb_ot_layout_substitute_lookup (OT::hb_ot_apply_context_t *c,
...
+    case HB_SCRIPT_GURMUKHI:
+    case HB_SCRIPT_KANNADA:
+    case HB_SCRIPT_MALAYALAM:
+    case HB_SCRIPT_ORIYA:
+    case HB_SCRIPT_TAMIL:
+    case HB_SCRIPT_TELUGU:
+    /* Unicode-2.0 additions */
+    case HB_SCRIPT_TIBETAN:
+    /* Unicode-3.0 additions */
+    case HB_SCRIPT_SINHALA:
```
behdad, line 1981: Not sure that Sinhala, nor all the main Indic, are hanging either.

behdad:
cc @kojiishi

behdad:
Also, I think the default for ideographic should be face-bottom, not the embox-bottom. Right?

behdad:
> Right?

I'm not sure.

dscorbett:
[The baseline tag registry says](https://docs.microsoft.com/en-us/typography/opentype/spec/baselinetags#ideographic-character-face):

> ICF metrics are not typically used as a text baseline, though they can be used in this way. Rather, ICF metrics allow an application to provide improved visual alignment of page elements to text.

But it also says:

> Applications can use the ICF box as an alignment tool, to ensure that glyphs touch the edges of the text frame and that page objects are visually aligned to text edges. It is also useful for aligning glyphs of different sizes on the same line. In Japanese traditional paper-based workflow, the ICF box was often used for these purposes. It provides optically aligned results that are superior to using the ideographic em-box.

So it sounds like 'icfb' is a better choice than 'ideo'.

src/hb-ot-layout.cc
```diff
@@ -1959,6 +1959,42 @@ hb_ot_layout_substitute_lookup (OT::hb_ot_apply_context_t *c,
...
+    case HB_SCRIPT_DEVANAGARI:
+    case HB_SCRIPT_GURMUKHI:
+    /* Unicode-2.0 additions */
+    case HB_SCRIPT_TIBETAN:
+    /* Unicode-4.1 additions */
+    case HB_SCRIPT_SYLOTI_NAGRI:
+    /* Unicode-5.2 additions */
+    case HB_SCRIPT_MEETEI_MAYEK:
+      return HB_OT_LAYOUT_BASELINE_TAG_HANGING;
+
```
jfkthame, line 1988: Some possible additions: limbu, newa, sharada, tirhuta, siddham, modi, takri, marchen, gunjala gondi, ... probably others, but there are a bunch that I'm not even vaguely familiar with, and don't have fonts readily to hand. (Looking over the unicode charts would be a start...)

And there are a number where it seems hard to know which way to go: e.g. gujarati, where although the letters lack a top line, they're still very reminiscent of the devanagari/bengali/gurmukhi characters, and might expect to be similarly aligned.

Might be worth reaching out to the [w3c i18n group](https://www.w3.org/International/) for input here? My own knowledge, at least, is very fragmentary, and I haven't gone digging for examples of the lesser-known scripts where expectations aren't at all obvious.

src/hb-ot-layout.cc
```diff
@@ -1959,6 +1959,56 @@ hb_ot_layout_substitute_lookup (OT::hb_ot_apply_context_t *c,
...
+    /* Unicode-9.0 additions */
+    case HB_SCRIPT_MARCHEN:
+    case HB_SCRIPT_NEWA:
+    /* Unicode-11.0 additions */
+    case HB_SCRIPT_GUNJALA_GONDI:
+      return HB_OT_LAYOUT_BASELINE_TAG_HANGING;
+
+    case HB_SCRIPT_HAN:
+    case HB_SCRIPT_HANGUL:
+      return HB_OT_LAYOUT_BASELINE_TAG_IDEO_EMBOX_BOTTOM_OR_LEFT;
```
dscorbett, line 2005: I’d guess this also applies to Bopomofo, Hiragana, Katakana, Khitan Small Script, Nüshu, and Tangut.

behdad:
> > Applications can use the ICF box as an alignment tool, to ensure that glyphs touch the edges of the text frame and that page objects are visually aligned to text edges. It is also useful for aligning glyphs of different sizes on the same line. In Japanese traditional paper-based workflow, the ICF box was often used for these purposes. It provides optically aligned results that are superior to using the ideographic em-box.
> 
> So it sounds like 'icfb' is a better choice than 'ideo'.

This is rather a big deal about usability of this API...

behdad:
> > > Applications can use the ICF box as an alignment tool, to ensure that glyphs touch the edges of the text frame and that page objects are visually aligned to text edges. It is also useful for aligning glyphs of different sizes on the same line. In Japanese traditional paper-based workflow, the ICF box was often used for these purposes. It provides optically aligned results that are superior to using the ideographic em-box.
> > 
> > 
> > So it sounds like 'icfb' is a better choice than 'ideo'.
> 
> This is rather a big deal about usability of this API...

In https://github.com/harfbuzz/harfbuzz/pull/3437, we are discussing whether `icfb` should or should not be synthesized. I don't think it can / should be synthesized. Which means, if most fonts don't ship with it, then we're doomed choosing it as default.

dscorbett:
Can this API return multiple baseline tags, so the caller can pick the first the font supports?

src/hb-ot-layout.cc
behdad: I don't know. @dscorbett can you make a decision here please?

dscorbett: [W3C’s latest draft Indic Layout Requirements](https://www.w3.org/TR/2020/WD-ilreq-20200529/) says that Gujarati has a hanging baseline. Let’s follow that advice.

It also says that Kannada, Telugu, and Odia don’t have a hanging baseline, but its description of “median/mean line” sounds essentially like 'hang' except without a visible headstroke (exactly like Gujarati). The important difference is that, apparently, the bigger font’s median/mean line aligns with the smaller font’s ascent. That’s different enough that these scripts should fall back to 'romn', at least for this first draft. I could definitely believe that many more Brahmic scripts ought to use 'hang' though. I’m just guessing here.

For future reference, Ogham and Mongolian use a central baseline, but there’s not really a tag for that yet. Also, some hieroglyphic and cuneiform scripts should maybe use the ideographic baseline.

dscorbett:
Now that #3437 synthesizes 'icfb', how about using that here instead of 'ideo'?

src/hb-ot-layout.cc
n8willis: Interesting examples in there for which characters to synthesize from as well; for those where ka has a cap with an upturn on it that's useful.

n8willis:
> > Can this API return multiple baseline tags, so the caller can pick the first the font supports?
> 
> Maybe. I thought about that. The problem is, to use them meaningfully, is not up to the caller to use whatever the font supports; The caller might need to synthesize whatever the dominant baseline for the line is even if the font doesn't have it, and use that for alignment...

If the question is whether application code would ever want more than one baseline, I believe Inkscape and GIMP have expressed a use-case for that in the past, in order to have options for text-along-a-path features.

behdad:
Proposing that we rename this to `hb_ot_layout_get_horizontal_baseline_tag_for_script()`. As we are noticing, vertical writing uses center baseline.

That works parallel to `hb_script_get_horizontal_direction()`.

behdad:
> Now that #3437 synthesizes 'icfb', how about using that here instead of 'ideo'?

Sounds good. I'll do that.

src/hb-ot-layout.cc
behdad: > Interesting examples in there for which characters to synthesize from as well; for those where ka has a cap with an upturn on it that's useful.

Yeah we tried to do that.

PR #3437, opened 2022-02-14
[BASE] Synthesize missing baselines

matthiasclasen:
Implement some of the heuristics that are outlined
in appendix A2 of the CSS Inline Layout Module, Level 3.


behdad:
Thanks! Let's align the exemplar character-per-scripts with my list in https://github.com/harfbuzz/harfbuzz/pull/3435, which itself is awaiting review from @dscorbett and @jfkthame

src/hb-ot-layout.cc
```diff
@@ -1989,6 +1989,94 @@ hb_ot_layout_get_baseline (hb_font_t                   *font,
...
+      hb_glyph_extents_t extents;
+
+      *coord = 0.6 * font->y_scale;
+
+      switch ((int) script_tag)
+      {
+      case HB_SCRIPT_DEVANAGARI:  ch = 0x915; break;
+      case HB_SCRIPT_BENGALI:     ch = 0x995; break;
+      case HB_SCRIPT_GURMUKHI:    ch = 0xa15; break;
+      case HB_SCRIPT_HEBREW:      ch = 0x5d4; break;
```
behdad, line 2063: Really weird that they have Hebrew in there!

ebraminio:
Guess you'll https://github.com/harfbuzz/harfbuzz/commit/861900c7db2a5adf16d4a208ec5e7cf84f52741a from https://github.com/harfbuzz/harfbuzz/pull/1877 also useful as it tried to implement spec suggested one http://web.archive.org/web/20170714012736/https://www.microsoft.com/typography/otspec/baselinetags.htm

src/hb-ot-layout.cc
```diff
@@ -1989,6 +1989,116 @@ hb_ot_layout_get_baseline (hb_font_t                   *font,
...
+    {
+      hb_codepoint_t glyph;
+      hb_glyph_extents_t extents;
+
+      if (hb_font_get_nominal_glyph (font, 0x638, &glyph) &&
+          hb_font_get_glyph_extents (font, glyph, &extents))
+      {
+        *coord = extents.y_bearing;
+        if (baseline_tag == HB_OT_LAYOUT_BASELINE_TAG_IDEO_FACE_BOTTOM_OR_LEFT)
+          *coord += extents.height;
```
behdad, line 2031: Should be height or width depending on direction.

behdad: Err. Ignore me. Humm.

behdad: 0x638 doesn't look right.

matthiasclasen: I've removed the ideo-face baselines from this for now.

matthiasclasen:
> Thanks! Let's align the exemplar character-per-scripts with my list in #3435, which itself is awaiting review from @dscorbett and @jfkthame

We've fleshed out the switch to pick a character for every script using a hanging baseline.

matthiasclasen:
The function now synthesizes all baselines, as best as we can, and returns void.

behdad:
Thanks @matthiasclasen. If you add your new function to `docs/harfbuzz-sections.txt` I think this is almost read to be merged.

behdad:
@dscorbett Matthias is using this in Pango. It seems to all fit. I think this LGTM to go in. WDYT?

behdad:
cc @jfkthame @drott @kojiishi AFAIK none of the browsers implement baselines, am I correct? Any interest?

matthiasclasen:
The pango code using the new api is here:

https://gitlab.gnome.org/GNOME/pango/-/merge_requests/589

Sample rendering of text with hanging baseline:

Screenshot from 2022-02-14 17-24-02

src/hb-ot-layout.cc
```diff
@@ -1989,6 +2058,184 @@ hb_ot_layout_get_baseline (hb_font_t                   *font,
...
+                                               HB_OT_LAYOUT_BASELINE_TAG_IDEO_EMBOX_BOTTOM_OR_LEFT,
+                                               direction,
+                                               script_tag,
+                                               language_tag,
+                                               &embox_bottom);
+
+      if (baseline_tag == HB_OT_LAYOUT_BASELINE_TAG_IDEO_FACE_TOP_OR_RIGHT)
+        *coord = embox_top + (embox_bottom - embox_top) * 2 / 10;
+      else
+        *coord = embox_bottom + (embox_top - embox_bottom) * 2 / 10;
```
dscorbett, line 2142: Why this particular constant of 20%?

matthiasclasen: I think we agreed here that 20% is too much. Its now 10%

matthiasclasen:
Here is pango-view drawing all the baselines (the dashed ones are synthesized, values from the font
would be solid lines).

Screenshot from 2022-02-15 17-24-32

PR #3442, opened 2022-02-15
[ot-metrics] Synthesize missing metrics

matthiasclasen:
Add a variant of hb_ot_metrics_get_position that
synthesizes missing values.

New api: hb_ot_metrics_get_position_with_fallback


behdad:
Thanks!

Some tests would be nice!

Issue #3168, opened 2021-08-25
docs: details about hb_ot_layout_get_ligature_carets


matthiasclasen:
I spent quite a while trying to find relevant details, it would be useful if the docs included some of this info:

- a ligature from n characters will have n-1 ligature positions
- the first character in the cluster isn't in represented in the list, since its caret position equals the glyph position
- returned positions are 'unshaped' and need to have kerning fixups applied when using them in a shaped glyph string

PR #3443, opened 2022-02-16
Improve docs for hb_ot_layout_get_ligature_carets

matthiasclasen:
Add some relevant details to the documentation
for this function.

Fixes: #3168


src/hb-ot-layout.cc
```diff
@@ -361,6 +361,13 @@ hb_ot_layout_get_attach_points (hb_face_t      *face,
  * Fetches a list of the caret positions defined for a ligature glyph in the GDEF
  * table of the font. The list returned will begin at the offset provided.
  *
+ * Note that a ligature that is formed from n characters will have n-1
+ * caret positions. The first character is not represented in the array,
```
khaledhosny, line 365: Instead of talking about characters, we should be talking about grapheme clusters.

PR #3449, opened 2022-02-17
Meson: Use CMake more to find dependencies on Windows

fanc999:
Hi,

This attempts to clean up the Meson build files a bit for finding dependencies on Visual Studio-like builds, by using CMake more, as

*  The CMake support in Meson is now much better integrated, and we require a Meson version that has the features we need to find a component of a given dependency, namely ICU-UC.
*  CMake is an optionally installed item on Visual Studio 2017 and later, and is something that is quite common for Windows build environments.
* CMake has built-in support to help find ICU and FreeType for Windows, although they use different semantics for CMake for Meson's `dependency()` method.

This PR moves the ICU-UC and FreeType dependency detection to CMake for Visual Studio-like builds, but keeps the Cairo dependency detection unchanged as CMake does not have built-in dependency search for Cairo, and Cairo only supported Meson builds in the latest unstable branch.

With blessings, thank you!


fanc999:
Hi,

For records, I could not get the `modules: ['ICU::uc']` in ICU's `dependency()` call to work, so I used the `component: 'uc'` instead-so we have to live with a Meson warning.  Oh well...

fanc999:
Hi,

Seems that the MSVC CI had trouble finding FreeType (and possibly GLib)...

khaledhosny (approved):
LGTM, thanks.

As you have found out, we build all the dependencies as submodules in MSVC ci jobs, but if you know a way we can reliability get pre built binaries for them, that would be better.

fanc999:
Hi @khaledhosny,

Thanks for the ack!

Personally, I built my copy of HarfBuzz with my own copy of Cairo (sans FontConfig), GLib, G-I, FreeType and ICU with all their dependencies with Visual Studio, along with Uniscribe(GDI) and DirectWrite support, so I do use /EHsc on my own builds as a result.

I think the best bet for reliable binaries for them other than I providing the binaries since this is a CI env is to get them from vcpkg.  Sorry, I am a CI novice, so I can't have a good say about how exactly one integrates obtaining vcpkg items into the CI-please note that for the CMake items here to work properly, they do depend on `%INCLUDE%` and `%LIB%` to be set to find the headers and .lib's respectively on Visual Studio--for FreeType and ICU CMake actually does what we have been doing previously to manually find the FreeType/ICU headers and .lib's, but in a much more complete manner.

My take at this.

With blessings, and cheers!

xclaesse:
This created a super hard to find issue when freetype is a subproject: https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/2507#note_1541409. The fallback is named `freetype2` but when using MSVC, the PR changed the dependency name it lookup to `freetype` which will not use the subproject fallback.

xclaesse:
FYI, with Meson >= 0.60 this could be changed to `dependency('freetype', 'freetype2', ...)`, Meson will try every names provided.

PR #3464, opened 2022-03-02
Fixed various misc. typos

luzpaz:
Found via `codespell -q 3 -S ./perf/texts -L actualy,als,ba,beng,clen,crasher,dependant,eachother,fo,gir,inout,ist,nd,ned,ot,pres,ro,statics,te,teh,timne`

### DIFF
--- a/docs/serializer.md
+++ b/docs/serializer.md
@@ -33,7 +33,7 @@ Here Obj N, is the object currently being written.
 ## Complex Tables
 
 Complex tables are made up of graphs of objects, where offset's are used
-to form the edges of the graphs. Each object is a continous slice of bytes
+to form the edges of the graphs. Each object is a continuous slice of bytes
 that contains zero or more offsets pointing to more objects.
 
 In this case the serialization buffer has a different layout:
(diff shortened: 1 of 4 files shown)


src/hb-ot-shape-complex-use-table.hh
```diff
@@ -29,7 +29,7 @@
  * # Not derivable
  * # Initial version based on Unicode 7.0 by Andrew Glass 2014-03-17
  * # Updated for Unicode 10.0 by Andrew Glass 2017-07-25
- * # Ammended for Unicode 10.0 by Andrew Glass 2018-09-21
+ * # Amended for Unicode 10.0 by Andrew Glass 2018-09-21
```
khaledhosny, line 32: This is a generated file, so the source need to be fixed. The source is https://github.com/harfbuzz/harfbuzz/blob/main/src/ms-use/IndicPositionalCategory-Additional.txt, but this is just a local copy and any changes to it will be eventually overridden. The upstream is https://github.com/microsoft/font-tools/blob/master/USE/IndicPositionalCategory-Additional.txt.

PR #3480, opened 2022-03-14
Make load_num_glyphs_from_loca conditional on HB_NO_BORING_EXPANSION

drott:
Fixes build errors complaining about this function being unused.


drott:
If you don't mind, I'll go ahead and merge this build fix as it unblocks rolling HB in Blink. Please shout if there's any issue.

drott:
> I thought `static inline` functions are immune to that warning / error.

That would indeed make sense, but unfortunately I had some builds complaining.

PR #3481, opened 2022-03-14
[subset] bug fix in prune_langsys

qxliu76:
we should not cache visited langsys cause 2 different RecordOfLangsys could have different Tag while pointing to the same Langsys, a langsys is redundant in Script A does not mean it's redundant in Script B as well.Same thing for visited script.

-Also adding the number of features in the LangSys's feature list to the visited langsys count so it's more accurate.
-Plus some improvement in langsys compare()


src/hb-ot-layout-common.hh
```diff
@@ -105,34 +109,14 @@ struct hb_prune_langsys_context_t
...
-  template <typename T>
-  bool visited (const T *p, hb_set_t &visited_set)
-  {
-    hb_codepoint_t delta = (hb_codepoint_t) ((uintptr_t) p - (uintptr_t) table);
-    if (visited_set.in_error () || visited_set.has (delta))
-      return true;
-
-    visited_set.add (delta);
-    return false;
+    return langsys_feature_count++ < HB_MAX_LANGSYS_FEATURE_COUNT;
```
garretrieger, line 119: You should add the number of features in the particular langsys to the count. So probably need to have this take a reference to the LangSys as an argument.

garretrieger (approved):

PR #3485, opened 2022-03-18
Add methods to import and export sets to arrays.

andyjgf:


src/hb-set.h
```diff
@@ -60,126 +60,147 @@ typedef struct hb_set_t hb_set_t;
 HB_EXTERN hb_set_t *
 hb_set_create (void);
 
+/**
+ * Creates a set initially populated with num_codepoint codepoints, which must
+ * be in increasing order.
+ *
+ * Since: REPLACEME
+ */
```
garretrieger, line 68: Public api method comments should be in the .c file.

src/hb-set.h
```diff
@@ -60,126 +60,147 @@ typedef struct hb_set_t hb_set_t;
...
+ * Exports the contents of the set to an array of hb_codepoint_t.
+ * codepoints[0] .. codepoints[hb_set_get_population(set) - 1] will be
+ * written to; you must ensure the size of the memory codepoints points to.
+ * The results will be in ascending order.
+ *
+ * Since: REPLACEME
+ */
+HB_EXTERN void
+hb_set_export_array (const hb_set_t  *set,
+                     hb_codepoint_t  *codepoints);
```
garretrieger, line 83: For safety the caller should tell us how big the array is so we don't overrun the buffer. You can have this just fill the buffer up to the maximum size if the provided buffer is undersized, maybe with a return code to indicate it was too small?

src/hb-set.h
```diff
@@ -60,126 +60,147 @@ typedef struct hb_set_t hb_set_t;
...
+hb_set_destroy (hb_set_t  *set);
 
 HB_EXTERN hb_bool_t
-hb_set_set_user_data (hb_set_t           *set,
-		      hb_user_data_key_t *key,
-		      void *              data,
-		      hb_destroy_func_t   destroy,
-		      hb_bool_t           replace);
+hb_set_set_user_data (hb_set_t            *set,
+                      hb_user_data_key_t  *key,
```
garretrieger, line 96: Can you undo all the formatting changes on unrelated code. Notably the default style for harfbuzz is to use mixed tabs and spaces. It looks like things have been converted to just spaces.

src/hb-set.h
```diff
@@ -60,126 +60,147 @@ typedef struct hb_set_t hb_set_t;
...
 hb_set_create (void);
 
+/**
+ * Creates a set initially populated with num_codepoint codepoints, which must
+ * be in increasing order.
+ *
+ * Since: REPLACEME
+ */
+HB_EXTERN hb_set_t *
+hb_set_create_from_sorted (const hb_codepoint_t  *sorted_codepoints,
```
garretrieger, line 70: Change this to hb_set_add_sorted_array(...) to mirror the underlying method. That has more utility because it can be used outside of set creation as well.

src/hb-bit-set-invertible.hh
```diff
@@ -117,6 +117,13 @@ struct hb_bit_set_invertible_t
   void del_range (hb_codepoint_t a, hb_codepoint_t b)
   { unlikely (inverted) ? (void) s.add_range (a, b) : s.del_range (a, b); }
 
+  /* Warning: always exports the non-inverted set contents. */
+  template <typename T>
+  void export_array (T *array) const
+  {
+    s.template export_array (array);
```
garretrieger, line 124: Is template a typo here?

test/api/test-set.c
```diff
@@ -1067,6 +1067,40 @@ test_set_inverted_operations (void)
...
+  g_assert (hb_set_has (set, 3));
+  g_assert (hb_set_has (set, 1000));
+  g_assert (hb_set_has (set, 2000));
+  g_assert (hb_set_has (set, 2001));
+  g_assert (hb_set_has (set, 2002));
+}
+
+static void
+test_export_to_array (void)
+{
```
garretrieger, line 1087: Once you add a size limit to the export call, add a test with too small of a buffer.

src/hb-set.h
andyjgf: Done

test/api/test-set.c
andyjgf: Done

src/hb-set.h
andyjgf: Done

src/hb-set.h
andyjgf: Done

src/hb-bit-set-invertible.hh
andyjgf: What looks wrong?

andyjgf: Got it.

src/hb-set.cc
```diff
@@ -61,6 +61,41 @@ hb_set_create ()
   return set;
 }
 
+/**
+ * Creates a set initially populated with num_codepoint codepoints, which must
```
garretrieger, line 65: Comment needs to be updated to match that this is now and  add and not create.

src/hb-bit-set-invertible.hh
```diff
@@ -117,6 +117,13 @@ struct hb_bit_set_invertible_t
   void del_range (hb_codepoint_t a, hb_codepoint_t b)
   { unlikely (inverted) ? (void) s.add_range (a, b) : s.del_range (a, b); }
 
+  /* Warning: always exports the non-inverted set contents. */
```
behdad, line 120: This needs to be fixed. Shouldn't be too hard.

src/hb-bit-page.hh
```diff
@@ -179,13 +199,18 @@ struct hb_bit_page_t
   typedef unsigned long long elt_t;
   static constexpr unsigned PAGE_BITS = 512;
   static_assert ((PAGE_BITS & ((PAGE_BITS) - 1)) == 0, "");
+  static constexpr unsigned PAGE_BITS_LG_2 = 9;
```
behdad, line 202: Just use `LOG` instead of `LG`. Also more readable if you add this and and the other one in a separate commit.

src/hb-bit-page.hh
```diff
@@ -86,6 +86,26 @@ struct hb_bit_page_t
...
+  // Writes out page values to the array p. Returns a pointer to where the next
+  // value should be written, after the values we wrote (if any).
+  // At most size codepoints will be written. Returns the number of codepoints written.
+  unsigned int export_array (uint32_t base, hb_codepoint_t *p, unsigned int size) const
+  {
+    unsigned int count = 0;
+    for (unsigned i = 0; i < len () && count < size; i++)
+    {
+      elt_t bits = v[i];
+      uint32_t v_offset = i << ELT_BITS_LG_2;
```
behdad, line 98: Since this is the only use of `ELT_BITS_LG_2` you could as well just multiply by `ELT_BITS`. That would avoid hardcoding sizeof `elt_t` as you have essentially done now.

src/hb-bit-set.hh
```diff
@@ -809,8 +822,8 @@ struct hb_bit_set_t
   }
   page_t &page_at (unsigned int i) { return pages[page_map[i].index]; }
   const page_t &page_at (unsigned int i) const { return pages[page_map[i].index]; }
-  unsigned int get_major (hb_codepoint_t g) const { return g / page_t::PAGE_BITS; }
-  hb_codepoint_t major_start (unsigned int major) const { return major * page_t::PAGE_BITS; }
+  unsigned int get_major (hb_codepoint_t g) const { return g >> page_t::PAGE_BITS_LG_2; }
```
behdad, line 825: I had assumed the compiler takes care of replacing this with bit ops...

src/hb-set.cc
```diff
@@ -61,6 +61,41 @@ hb_set_create ()
...
+ * be in increasing order.
+ *
+ * Since: REPLACEME
+ */
+HB_EXTERN void
+hb_set_add_sorted_array (hb_set_t             *set,
+		         const hb_codepoint_t *sorted_codepoints,
+		         unsigned int          num_codepoints)
+{
+  if (unlikely(!set || !sorted_codepoints || !num_codepoints)) return;
```
behdad, line 75: We don't support `nullptr` in our functions. The rest need to be pushed down inside `add_sorted_array`.

src/hb-set.cc
```diff
@@ -61,6 +61,41 @@ hb_set_create ()
...
+ * written out. Returns the number of codepoints written.
+ *
+ * Since: REPLACEME
+ */
+HB_EXTERN unsigned int
+hb_set_export_array (const hb_set_t *set,
+		     hb_codepoint_t *codepoints,
+		     unsigned int    size)
+{
+  if (unlikely(!set || !codepoints || !size)) return 0;
```
behdad, line 94: Same here. Also, no need to check for nullptr `codepoints`.

PR #3491, opened 2022-03-21
Use bit shifting for major/minor computation.

andyjgf:


behdad:
Thanks Andy. In the future please start the commit message with "[set] " for set-related changed.

PR #3492, opened 2022-03-21
Add option to add an entire array of values to a set at once.

andyjgf:


src/hb-set.cc
```diff
@@ -61,6 +61,22 @@ hb_set_create ()
...
+HB_EXTERN void
+hb_set_add_sorted_array (hb_set_t             *set,
+		         const hb_codepoint_t *sorted_codepoints,
+		         unsigned int          num_codepoints)
+{
+  set->add_sorted_array (sorted_codepoints,
+		         num_codepoints,
+		         sizeof(hb_codepoint_t));
+}
+
```
khaledhosny, line 79: This should probably go after `hb_set_add` below. The documentation syntax needs to be fixed as well, and the new method need to be added to the relevant section in `docs/harfbuzz-sections.txt`.

behdad:
Fixed up in https://github.com/harfbuzz/harfbuzz/commit/4289684cdbbe5fbfa394bc3ce568eb8f27f404f4

Issue #2910, opened 2021-03-23
Give crowbar more bite


davelab6:
https://typedrawers.com/discussion/comment/48949/#Comment_48949 :

> "Ah, yes, this is probably where I am trying to be too clever. Crowbar uses Harfbuzz to shape the text, and Harfbuzz (correctly) uses a list of features as part of selecting the lookups to execute in a stage. But once it's chosen the lookup IDs, it throws the feature information away. So Crowbar tries to look back into the feature-lookup mapping to try to understand why a lookup got called and where it came from, but if a lookup appears in more than one feature then I can see that it might guess wrong."

— @simoncozens 

I was discussing this with @rsheeter and he observed that HB could offer the correct answer up,  and whether that fits into HB nicely might be another question :)

@behdad wdyt?

simoncozens:
Remember that Crowbar supports and reads the fontTools DEBG table, and so any font built with feature debugging turned on gets annotated with what feature, lookup name and source line/column the lookup came from.

behdad:
It's not super straightforward; will think about it.

behdad:
Tracking in https://github.com/harfbuzz/harfbuzz/pull/3495

PR #3495, opened 2022-03-22
Towards adding more buffer messages

behdad:
Eventually will fix https://github.com/harfbuzz/harfbuzz/issues/2910

@simoncozens how's the proposed message format?

Eg:
```
trace: start lookup 106	buffer: [BehxIni=0|OneDotBelowNS=0|HehMed=1|DalFin=2|AlefSep=3|DalSep=4]
trace: replacing glyph at 2: gid625 -> gid798	buffer: [BehxIni.outD5=0|OneDotBelowNS=0|HehMed=1|DalFin=2|AlefSep=3|DalSep=4]
trace: end lookup 106	buffer: [BehxIni.outD5=0|OneDotBelowNS=0|HehMed.inD5outT1=1|DalFin=2|AlefSep=3|DalSep=4]
```

I opted for leaving out the lookup-type from the message. WDYT?

cc @khaledhosny

### DIFF
--- a/src/OT/Layout/GPOS/CursivePosFormat1.hh
+++ b/src/OT/Layout/GPOS/CursivePosFormat1.hh
@@ -140,6 +140,13 @@ struct CursivePosFormat1
     unsigned int i = skippy_iter.idx;
     unsigned int j = buffer->idx;
 
+    if (HB_BUFFER_MESSAGE_MORE && c->buffer->messaging ())
+    {
+      c->buffer->message (c->font,
+			  "cursive attaching glyph at %d to glyph at %d",
+			  i, j);
+    }
+
     buffer->unsafe_to_break (i, j + 1);
     float entry_x, entry_y, exit_x, exit_y;
     (this+prev_record.exitAnchor).get_anchor (c, buffer->info[i].codepoint, &exit_x, &exit_y);
@@ -229,6 +236,13 @@ struct CursivePosFormat1
 	pos[parent].y_offset = 0;
       else
 	pos[parent].x_offset = 0;
+    }
+
+    if (HB_BUFFER_MESSAGE_MORE && c->buffer->messaging ())
+    {
+      c->buffer->message (c->font,
+			  "cursive attached glyph at %d to glyph at %d",
+			  i, j);
     }
 
     buffer->idx++;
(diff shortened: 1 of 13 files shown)


behdad:
Currently the message is emitted *before* the change. Maybe it should be emitted *after* the change. I think it's better that way. Again, WDYT? If emitted after, then likely that I won't write out what it replaced.

What would be a good message? Just that "something was replaced at this index"?

cc @mhosken

behdad:
cc @dscorbett

simoncozens:
What was replaced can be derived from the before/after. The problem we had before was that there were no consistent messages when lookups were called from other lookups. (Contextual chaining.) You get a message at the start of the top-level lookup, one at the end of it, but the reported buffer state inside is often incorrect.

simoncozens:
Actually, the problem is that right now there are *no* messages inside contextual rules, unless you apply my [horrible hack](https://github.com/harfbuzz/harfbuzz/issues/2468#issuecomment-643488169), and that is the part that is inconsistent.

mhosken:
You can't have too much information in debug, which is why we used json for graphite. That way you can say precisely what is doing what where. So I would include the lookup in the message. I don't think it makes much difference whether one reports the buffer before or after replacement, although if you do report it before for some other reason, it might help to report it after in the substitution. Although I would report it using the same glyph name as you have in the buffer (what is gid625 in the example?)
For a contextual lookup, it might be nice to show the substring matched that has the replacement occurring in it. As in something like:
```
trace: start lookup 106	buffer: [BehxIni=0|OneDotBelowNS=0|HehMed=1|DalFin=2|AlefSep=3|DalSep=4]
trace: in [BehxIni=0|OneDotBelowNS=0|HehMed=1]
trace: lookup 27 replacing glyph at 2: gid625 -> gid798	buffer: [BehxIni.outD5=0|OneDotBelowNS=0|HehMed=1|DalFin=2|AlefSep=3|DalSep=4]
trace: end lookup 106	buffer: [BehxIni.outD5=0|OneDotBelowNS=0|HehMed.inD5outT1=1|DalFin=2|AlefSep=3|DalSep=4]
```

behdad:
I redid this, I think I like it better. I'm adding more.

behdad:
I'm merging this and continuing contextuals in https://github.com/harfbuzz/harfbuzz/issues/3753

nico:
This adds two new `sprintf` calls, which are marked deprecated in the macOS 14 SDK, which was released today. Any chance you could replace those with `snprintf`? (`snprintf (p, sizeof buf - (p - buf), "%u", i);`, I suppose?)

behdad:
> This adds two new `sprintf` calls, which are marked deprecated in the macOS 14 SDK, which was released today. Any chance you could replace those with `snprintf`? (`snprintf (p, sizeof buf - (p - buf), "%u", i);`, I suppose?)

I had already swapped those in 294b1c9f6eda839d22c1509199887e9650b377fe, but it was wrong (ouch), though not exploitable because we know buffer is large enough!  I just fixed them to use your version. Thanks.

@drott for Chrome roll.

nico:
Ah, racing comments. Thanks much for the fast reply :)

Issue #537, opened 2017-09-10
Different vertical glyph positions between ot and ft font functions


khaledhosny:
right (ot):
```
hb-view DejaVuSans.ttf "GgXx" --direction=t --font-funcs=ot
```
left (ft):
```
hb-view DejaVuSans.ttf "GgXx" --direction=t --font-funcs=ft
```

![ot funcs](https://user-images.githubusercontent.com/93914/30253233-7686afd6-9689-11e7-8778-0748d50f03b1.png) ![ft funcs](https://user-images.githubusercontent.com/93914/30253221-56c3047e-9689-11e7-8081-86f6268db804.png)

behdad:
FT doesn't implement VORG.  It should some day.  HB too.  But that only affects CFF.

For TrueType vertical, we should fetch tsb from glyf table and use it for to implement `v_origin` func.  That's currently missing.

behdad:
VORG is https://github.com/behdad/harfbuzz/issues/544

HinTak:
https://github.com/harfbuzz/harfbuzz/issues/1248 is not about glyph positioning, but font extents. FWIW. One of the values has a completely opposite sign!

HinTak:
The font concerned - msmincho - does not have a VORG table. It has a vhea table.

HinTak:
Also, my arriving at this issue is not that the two sets of functions behave differently, but that different set is the default, depending on **how** you open your file.

behdad:
> FWIW. One of the values has a completely opposite sign!

You're amazing at writing useless bug reports. ;)

HinTak:
FWIW, the descender changed sign from -36 to +36 (hence the 216 to 288 difference) between the ot/ft font funcs.

(1) writing/editing from mobile phone, don't always have all the details, (2) msmincho is common enough that it is easy for most people to see it themselves.

behdad:
Still different.  Reopening.

behdad:
For the rest, it looks like if `vmtx` and `vhea` tables are missing, FreeType uses horizontal ascent+descent as vertical advance, whereas we use upem.  That I'm not sure I want to fix, as it's tricky in the code.

HinTak:
Also I don't really think the issue I reported is same as khaley's. His concerns glyph extents, mine concerns font extents. Though I agree it is likely related. Anyway, now you know the **font** descender changes sign between ft and ot.

behdad:
[...]

PR #3497, opened 2022-03-22
[ot-font] Fix vertical-origin fallback to match FreeType

behdad:
Thanks to Tav for analysis in https://github.com/harfbuzz/harfbuzz/issues/537#issuecomment-803589962

Fixes https://github.com/harfbuzz/harfbuzz/issues/537


behdad:
Before:
image

After:
image

PR #3500, opened 2022-03-22
[set] Add call to export set contents to an array.

andyjgf:


src/hb-bit-page.hh
```diff
@@ -86,6 +86,58 @@ struct hb_bit_page_t
...
+  {
+    unsigned int count = 0;
+    for (unsigned i = 0; i < len () && count < size; i++)
+    {
+      elt_t bits = v[i];
+      uint32_t v_offset = i << ELT_BITS_LOG_2;
+      for (unsigned int j = 0; j < ELT_BITS && count < size; j++)
+	if ((elt_t (1) << j) & bits)
+	{
+	  *p++ = base | v_offset | j;
```
behdad, line 102: Any reason you don't add `base` to `v_offset` outside the loop?

src/hb-set.cc
```diff
@@ -239,6 +239,23 @@ hb_set_has (const hb_set_t *set,
...
+ * written out. Returns the number of codepoints written.
+ *
+ * Since: REPLACEME
+ */
+HB_EXTERN unsigned int
+hb_set_export_array (const hb_set_t *set,
+		      hb_codepoint_t *codepoints,
+		      unsigned int    size)
+{
+  if (unlikely(!set || !codepoints || !size)) return 0;
```
behdad, line 254: Remove these please.

src/hb-bit-page.hh
andyjgf: Good idea. Done.

src/hb-set.cc
andyjgf: Done

Issue #3502, opened 2022-03-23
USE machine explosion


behdad:
I'm seeing a fuzzer-found test-case where the USE machine's `machine_index_t` is showing an O(n^2) behavior because of its `operator =`.   The pattern seems to be that there's an assignment of `0` happening every other assignment, and that's very expensive:
```
index 0 n 16432
index 16432 n 16433
index 16432 n 0
index 0 n 16433
index 16433 n 16434
index 16433 n 0
index 0 n 16434
index 16434 n 16435
index 16434 n 0
index 0 n 16435
index 16435 n 16436
index 16435 n 0
```

Can we make assignment of 0 cheap by remembering the start?

behdad:
I'm out of my depth...

behdad:
The problem is the code that ragel generates:
```
 auto pe = p + p.len (); 
  auto eof = +pe; 
  auto ts = +p; 
  auto te = +p; 
  unsigned int act HB_UNUSED; 
  int cs; 
»» 
#line 688 "hb-ot-shape-complex-use-machine.hh" 
↦       { 
↦       cs = use_syllable_machine_start; 
↦       ts = 0; 
↦       te = 0; 
↦       act = 0; 
↦       } 
```

It first sets those variables to `p`, then to `0`...

behdad:
> The problem is the code that ragel generates:
> 
> ```
>  auto pe = p + p.len (); 
>   auto eof = +pe; 
>   auto ts = +p; 
>   auto te = +p; 
>   unsigned int act HB_UNUSED; 
>   int cs; 
> »» 
> #line 688 "hb-ot-shape-complex-use-machine.hh" 
> ↦       { 
> ↦       cs = use_syllable_machine_start; 
> ↦       ts = 0; 
> ↦       te = 0; 
> ↦       act = 0; 
> ↦       } 
> ```
> 
> It first sets those variables to `p`, then to `0`...

Am not sure those are the culprit.

behdad:
In general when this code sets one of those variables to 0, it's trying to set it to null. I'm not sure it relies on the value later on. I'll see if not acting on a 0 matters...

behdad:
Or we can keep a separate nullness...

behdad:
Humm. But if I don't act on 0, how to differentiate that from beginning-of-buffer? Ouch.

behdad:
Oh, we never go to beginning-of-buffer by setting to 0 I guess.

behdad:
[...]

PR #3503, opened 2022-03-23
[use] Avoid O(n^2) in the machine

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3502

I couldn't generate the ragel-generated machine file. Can someone do that and update please? cc @khaledhosny


behdad:
@dscorbett review?

dscorbett:
It does seem to work, but I’m not versed in Ragel enough to be certain this does not introduce an obscure bug. If Ragel is setting a variable to 0, there must be a reason for it, but I don’t know whether this machine uses any feature that relies on it.

behdad:
It sets it for hygiene but has no way to query for it currently.

behdad:
I explicitly deleted `operator bool` now, to ensure the code does not try to check for nullptr... 

Please regenerate the .hh file. Thanks.

dscorbett:
To regenerate, I’m just rebuilding with `-Dragel_subproject=true`. Does that not work for you?

behdad:
> It does seem to work, but I’m not versed in Ragel enough to be certain this does not introduce an obscure bug. If Ragel is setting a variable to 0, there must be a reason for it, but I don’t know whether this machine uses any feature that relies on it.

`ts` and `te` record the start and end of the last found match. Then it clears them when it starts looking for the next match.

We can make the machine record null via a boolean. But then that boolean will never be used.

behdad:
> But then that boolean will never be used.

Unless we use it in the `==` implementation. But since the machine has been working so far, that would be extra overhead...

dscorbett:
Should the copy constructor copy `o.is_null`?

behdad:
This is weird. The `hb-ot-font`'s `get_glyph_path` is not called at all for this particular glyph. I'm suspecting a cairo bug :(.

dscorbett:
I built and tested `main` with this PR’s test file and it passed. Can you show that there was a problem and that this fixes it?

behdad:
> Can you show that there was a problem and that this fixes it?

Running `test/fuzzing/hb-shape-fuzzer` on the font file that this commit adds was taking tens of seconds before this commit...

behdad:
> I built and tested `main` with this PR’s test file and it passed.

It passes, just takes too long, which would timeout the fuzzers.

behdad:
before:
```
$ time test/fuzzing/hb-shape-fuzzer ~/Downloads/clusterfuzz-testcase-hb-shape-fuzzer-5446125635633152 
/home/behdad/Downloads/clusterfuzz-testcase-hb-shape-fuzzer-5446125635633152 (655 bytes)

real	1m5.016s
user	1m4.852s
sys	0m0.003s
```

after:
```
$ time test/fuzzing/hb-shape-fuzzer ~/Downloads/clusterfuzz-testcase-hb-shape-fuzzer-5446125635633152 
/home/behdad/Downloads/clusterfuzz-testcase-hb-shape-fuzzer-5446125635633152 (655 bytes)

real	0m0.045s
user	0m0.042s
sys	0m0.002s
```

dscorbett:
It would be good if fuzzed speed tests could have individual timeouts, so that without this change to the machine this test would fail the build, but I guess that is not necessary as part of this PR.

dscorbett (approved):

behdad:
We used to have that I thought....

behdad:
> We used to have that I thought....

Maybe we have for shape tests, but not the test/fuzzing....

PR #3504, opened 2022-03-24
test: Dist the platform shaper test data

fanc999:
Hi,

From the commit message--I presume the release tarballs are generated from the autotools builds?

<i>For builds from release tarballs, the tests fail in the DirectWrite and Uniscribe tests when these platform shapers are enabled, since the data files were not found in the source tree, when building with Meson at least.</i>

<i>Fix this by dist'ing the platform shaper test data files.</i>

This will fix the failing tests when building from the release tarballs when platform shapers are enabled.

Sorry, I am not that good in autotools to attempt to enable the platform shaper tests in the autotools builds, so I am not attempting to do that here in this PR.

With blessings, thank you!

### DIFF
--- a/test/shape/data/in-house/Makefile.sources
+++ b/test/shape/data/in-house/Makefile.sources
@@ -71,5 +71,15 @@ TESTS = \
 	tests/zero-width-marks.tests \
 	$(NULL)
 
+if HAVE_CORETEXT
+TESTS += tests/coretext.tests
+endif
+if HAVE_DIRECTWRITE
+TESTS += tests/directwrite.tests
+endif
+if HAVE_UNISCRIBE
+TESTS += tests/uniscribe.tests
+endif
+
 DISABLED_TESTS = \
 	$(NULL)


khaledhosny:
> I presume the release tarballs are generated from the autotools builds?

Yes.

fanc999:
(argh..., I forgot the `tests/` prefix...)

test/shape/data/in-house/Makefile.sources
```diff
@@ -71,5 +71,11 @@ TESTS = \
 	tests/zero-width-marks.tests \
 	$(NULL)
 
+PLAT_SHAPER_TESTS = \
+	tests/coretext.tests \
+	tests/uniscribe.tests \
+	tests/directwrite.tests \
+	$(NULL)
+
```
khaledhosny, line 79: This should be (I think):
```
if HAVE_CORETEXT
 	TESTS += tests/coretext.tests
endif
```

etc.

fanc999: Here we go... let's hope this does not trip up CMake :).

khaledhosny: We don’t enable tests with CMake build IIRC.

khaledhosny:
Hmm, I wonder where did circleci jobs (which include the discheck one) go, none seems to be running!

khaledhosny:
> Hmm, I wonder where did circleci jobs (which include the discheck one) go, none seems to be running!

It wouldn’t even trigger manually for this PR, weird!

khaledhosny:
I think my suggestion was wrong. The tests will be added to dist tarball only when the respective platform shaper is enable during configuration, which is not what we want.

khaledhosny:
I have a better fix for this.

khaledhosny:
https://github.com/harfbuzz/harfbuzz/pull/3505

Issue #3507, opened 2022-03-24
Meson build writes to the source directory


aneejit1:
The meson build process is writing to the source directory. This is a problem if there is only read only access to the source and causes the build to fail in one of two places:

- in "src/gen-hb-version.py" while running meson
- in "src/gen-harfbuzzcc.py" once ninja has been invoked

The scripts src/gen-harfbuzzcc.py and src/gen-hb-version.py need to be modified so a copy back to the source tree is only performed if the generated files differ from the baselined files (this should therefore only happen when something does
actually change as part of the development process).

khaledhosny:
Fixed in https://github.com/harfbuzz/harfbuzz/pull/3508

PR #3508, opened 2022-03-24
 Meson build writes to the source directory (issue #3507 )

aneejit1:
The meson build is writing files into the source directory which causes a build failure if the source is read-only. The Python scripts "src/gen-hb-version.py" and "src/gen-harfbuzzcc.py" have been amended so a copy back to the source tree is only performed if the generated files differ from the baselined files (this should therefore only happen when something does
actually change as part of the development process).


khaledhosny:
As you can see, this copying is intentional and we depend on it. Basically, we want to be able to build HarfBuzz from git checkout without using the build system at all, and some large HarfBuzz consumers depend on this.

khaledhosny:
What we can do is to check that the files are the same and skip the copying since the copying is only needed when one modifies HarfBuzz source.

aneejit1:
OK, maybe I'm overthinking this but, would this do? (it's lost the formatting)

`# copy it also to the source tree, but only if it has changed`
`with open(os.path.join (CURRENT_SOURCE_DIR, os.path.basename (OUTPUT)), "r", encoding='utf-8') as baseline:`
`        fromlines = baseline.readlines()`
`with open(OUTPUT) as generated:`
`        tolines = generated.readlines()`
` `
`diff = difflib.unified_diff(fromlines, tolines)`
`if next(diff, None) is not None:`
`        shutil.copyfile (OUTPUT, os.path.join (CURRENT_SOURCE_DIR, os.path.basename (OUTPUT)))`

khaledhosny:
`baseline.read() == generated.read()` should be enough.

aneejit1:
Hmm! Well the checks seem to be a bit happier about that.

src/gen-harfbuzzcc.py
```diff
@@ -14,5 +14,8 @@
 with open (OUTPUT, "wb") as f:
 	f.write ("".join ('#include "{}"\n'.format (os.path.basename (x)) for x in sources if x.endswith (".cc")).encode ())
 
-# copy it also to src/
-shutil.copyfile (OUTPUT, os.path.join (CURRENT_SOURCE_DIR, os.path.basename (OUTPUT)))
+# copy it also to the source tree, but only if it has changed
+with open(os.path.join (CURRENT_SOURCE_DIR, os.path.basename (OUTPUT)), "r") as baseline:
```
khaledhosny, line 18: This long `os.path.join()` is repeated twice, can you make it a variable. Also the the `"r"` option is redundant, that is the default anyway.

src/gen-hb-version.py
```diff
@@ -32,5 +32,8 @@
 			.replace ("@HB_VERSION@", version)
 			.encode ())
 
-# copy it also to src/
-shutil.copyfile (OUTPUT, os.path.join (CURRENT_SOURCE_DIR, os.path.basename (OUTPUT)))
+# copy it also to the source tree, but only if it has changed
+with open(os.path.join (CURRENT_SOURCE_DIR, os.path.basename (OUTPUT)), "r", encoding='utf-8') as baseline:
```
khaledhosny, line 36: You are opening one file with `encoding='utf-8'` and the other not, this will cause trouble on Windows if a non-ASCII character appeared in these files. Better use the default, or even open the files in binary mode so Python does not temper with the content.

aneejit1: The "encoding" parameter was lifted from the open earlier in the code. Is opening in binary mode going to cause a problem with crlf/lf mismatches?

khaledhosny: We open the file for writing in binary mode above, so it should be using the lf on all platforms already.

src/gen-harfbuzzcc.py
aneejit1: The filename is now a variable. I've kept the access mode and changed it to "rb" in line with the other script.

src/gen-hb-version.py
aneejit1: Mode is now "rb", the encoding is gone, and the filename is now in a variable as per the other script.

khaledhosny (approved):
Please squash all the commits into one, thanks.

behdad:
> Please squash all the commits into one, thanks.

Github merge can do that.

Issue #1611, opened 2019-03-10
Context lookup broken if moving back after a glyph delete


behdad:
This was reported to me privately two years ago by Doug.  I tried fixing it (including a test case) in b9b005f3a44ccf78a45b212b126287b69b9f4b40.  That fix was not good and caused memory issues discovered by address-sanitizer fuzzers, so I reverted it in 47e7a1800fba9b8bf042a1f4976a15ab012ebfc7.  I forgot to file an issue for it then.

I *think* that's what @Litherum is also hitting in https://twitter.com/Litherum/status/1104439589719072769

PR #3509, opened 2022-03-24
[apply-lookup] Try to fix the logic for contextual lookups

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/1611

Notably, this fixes Myles's AdditionFont:
https://litherum.blogspot.com/2019/03/addition-font.html

Test with AdditionFont, eg.:
$ util/hb-view AdditionFont.otf =1112112+1113134=


behdad:
I don't know how to debug the failures further currently...

behdad:
Okay. Got this right I thinkk.

Issue #3506, opened 2022-03-24
Bug with Recursive font


khaledhosny:
HarfBuzz seems to have a problem applying variations to some glyphs of Recursive font downloaded from https://fonts.google.com/specimen/Recursive:
```
$ hb-shape Recursive-VariableFont_CASL,CRSV,MONO,slnt,wght.ttf -u c5 --variations="wght=300" --show-extents
[Aring=0+650<40,896,570,-896>]
$ hb-shape Recursive-VariableFont_CASL,CRSV,MONO,slnt,wght.ttf -u c5 --variations="wght=600" --show-extents
[Aring=0+650<0,0,0,0>]
```

`hb-view` also produces a blank image. FreeType font functions work fine:
```
$ hb-shape Recursive-VariableFont_CASL,CRSV,MONO,slnt,wght.ttf -u c5 --variations="wght=600" --font-funcs=ft --show-extents
[Aring=0+650<36,898,578,-898>]
```
[Recursive.zip](https://github.com/harfbuzz/harfbuzz/files/8340177/Recursive.zip)

khaledhosny:
Interestingly, upstream files or even file downloaded directly [from Google Fonts git repository](https://github.com/google/fonts/blob/main/ofl/recursive/Recursive%5BCASL%2CCRSV%2CMONO%2Cslnt%2Cwght%5D.ttf) do not show this problem:
```
$ hb-shape "Recursive[CASL,CRSV,MONO,slnt,wght].ttf" -u c5 --variations="wght=600" --show-extents
[Aring=0+650<36,898,578,-898>]
```

cc @rsheeter in case the download zip from the website is modified by some tool.

behdad:
Reproduces for me.

behdad:
There's a bug inside gvar's `apply_deltas_to_points`. glyf is bailing if that function returns false. That's wrong. But also if I make it not do that, the ring doesn't get variations applied whereas freetype does apply. Debugging further.

behdad:
> There's a bug inside gvar's `apply_deltas_to_points`

Looks to me like `unpack_deltas` is getting wrong bytes to unpack. This condition is being hit:
```
      if (j < run_count) 
```

behdad:
I think we should apply this change regardless:
```diff
diff --git a/src/hb-ot-glyf-table.hh b/src/hb-ot-glyf-table.hh
index 066e152da..b4ac68834 100644
--- a/src/hb-ot-glyf-table.hh
[...]

PR #3510, opened 2022-03-25
[gvar] Fix decoding of private points

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3506

@khaledhosny can you please add test case with that glyph and --show-extents?

### DIFF
--- a/src/hb-ot-glyf-table.hh
+++ b/src/hb-ot-glyf-table.hh
@@ -820,8 +820,7 @@ struct glyf
       }
 
 #ifndef HB_NO_VAR
-      if (unlikely (!glyf_accelerator.gvar->apply_deltas_to_points (gid, font, points.as_array ())))
-	return false;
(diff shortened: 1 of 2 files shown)


khaledhosny (approved):
I can’t comment on the code, but with this change `hb_font_get_glyph_extents` works for all glyphs in the font and seems to return sensible glyph extents.

khaledhosny:
The glyph extents from ot functions seems to be often quite different from ft functions, but that is probably independent to this issue.

behdad:
Humm

behdad:
> The glyph extents from ot functions seems to be often quite different from ft functions, but that is probably independent to this issue.

File an issue with a sample glyph please.

khaledhosny:
> > The glyph extents from ot functions seems to be often quite different from ft functions, but that is probably independent to this issue.
> 
> File an issue with a sample glyph please.

I’m seeing the difference in a test C program I wrote, but not in hb-shape, so I’m guessing something is broken on my end.

Issue #3513, opened 2022-03-28
Regression in Indic shaping of दिसंबर after #3490


drott:
Bisected to 90f09b1e877dc

```
$ build/util/hb-view ~/dev/blink/src/third_party/test_fonts/test_fonts/Lohit-Devanagari.ttf "दिसंबर"
```

<img width="646" alt="image" src="https://user-images.githubusercontent.com/1021784/160410574-bc22c593-5ebf-44df-b89c-200c1604d4a3.png">

Expected, at least that's what it was before:

<img width="644" alt="image" src="https://user-images.githubusercontent.com/1021784/160410880-9e93c832-a4a6-448c-8018-d7d52a0b9c70.png">

drott:
This change discussed in #3488 - so perhaps this is the kind of font that is expecting the previous order?

drott:
```
    <namerecord nameID="5" platformID="3" platEncID="1" langID="0x409">
      2.95.4
    </namerecord>
```
[lohit_devanagari.zip](https://github.com/harfbuzz/harfbuzz/files/8363359/lohit_devanagari.zip)

That seems to be the "newest" version of the font, according to https://pagure.io/lohit/tree/master

behdad:
We should test what other shapers do with this. cc @dscorbett

jfkthame:
Looks like this is a result of GSUB lookup 25, which ligates the i-matra with the anusvara, as part of the `pres` feature:
```
      <Lookup index="25">
        <LookupType value="4"/>
        <LookupFlag value="2"/><!-- ignoreBaseGlyphs -->
        <!-- SubTableCount=1 -->
        <LigatureSubst index="0" Format="1">
          <LigatureSet glyph="isigndeva">
            <Ligature components="radeva_viramadeva,anusvaradeva" glyph="isign_ra_virama_anusvara"/>
            <Ligature components="radeva_viramadeva" glyph="isign_ra_virama"/>
            <Ligature components="anusvaradeva" glyph="isigndeva_anusvaradeva"/>
          </LigatureSet>
          <LigatureSet glyph="isigndeva.alt1">
            <Ligature components="radeva_viramadeva,anusvaradeva" glyph="isign_ra_virama_anusvara.alt1"/>
            <Ligature components="radeva_viramadeva" glyph="isign_ra_virama.alt1"/>
            <Ligature components="anusvaradeva" glyph="isigndeva_anusvara.alt1"/>
          </LigatureSet>
[...]

PR #3516, opened 2022-03-28
[indic] Don't clear syllables until end of GSUB

behdad:
Alternate fix for https://github.com/harfbuzz/harfbuzz/issues/3513

This matches what Uniscribe does, but differs from what we chose
to do before, which was to allow user features to be applied across
syllable boundaries.


behdad:
Decided not to pursue this approach.

PR #3522, opened 2022-04-01
[blob] Return nullptr from create_from_file_or_fail

khaledhosny:
Fix a couple of cases where it would return empty blob, possibly missed in bdfed8f113431a2156e13d59a4b21e19feb7efd9.


behdad:
lgtm.

PR #3527, opened 2022-04-06
CMakeLists.txt: also match 'AppleClang' compiler to not link with libc++

anthrotype:
As of CMake 3.0.0, the `CMAKE_*_COMPILER_ID` for the Apple-provided Clang is no longer "Clang" but "AppleClang".

https://cmake.org/cmake/help/v3.0/release/3.0.0.html#other-changes

This PR ensures that we match either "Clang" or "AppleClang" when adding the flags which prevent from linking with the libc++.
Without this, we currently end up with a shared library on macOS that links with libc++, which is undersirable for pure-C projects like https://github.com/rougier/freetype-py/pull/148

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -462,7 +462,9 @@ if (UNIX OR MINGW)
     link_libraries(-Bsymbolic-functions)
   endif ()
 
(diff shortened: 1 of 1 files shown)


anthrotype:
I know that the preferred build system is meson and we want to eventually deprecate and drop support for cmake, but at least this small fix let us (freetype-py) continue use the cmake build system for a little while. We will port our build script to meson at some point. cc @madig

Issue #3528, opened 2022-04-06
[aat] Regression in Bengali AAT shaping


jfkthame:
Shaping of some AAT fonts was recently broken, as reported at https://bugzilla.mozilla.org/show_bug.cgi?id=1763368.

Good (21c4fc1011854547edf564573fdbec4687e76e04):
```
$ hb-shape --font-file "/System/Library/Fonts/Supplemental/Bangla MN.ttc" --unicodes "U+09AC U+09BF"
[bn_ikaar=0+474|bn_ba=0+998]
```

Bad (23159084b43c1ce429d9e98035bf845919fd8a89, and same with current trunk code):
```
$ hb-shape --font-file "/System/Library/Fonts/Supplemental/Bangla MN.ttc" --unicodes "U+09AC U+09BF"
[bn_ba=0+998|bn_ikaar=0+474]
```

Note the inverted order of the ikaar and the base consonant. This was regressed by #3314.

This breaks at least the Bangla MN and Bangla Sangam MN fonts shipped with macOS; I have not exhaustively tested other AAT Indic fonts but it seems likely others may be similarly affected.

jfkthame:
@behdad I checked that 23159084b43c1ce429d9e98035bf845919fd8a89 reverts cleanly on current trunk, and restores the expected result here. Given that Firefox apparently wasn't affected by #3314, we could do that locally as a band-aid, but obviously this will be less attractive to Chrome as #3314 was a problem for them. So you may want to try and find an alternative?

behdad:
@jfkthame Right. I'm guessing that CoreText is treating regional-indicators as atomic units only? Let me find an alternative fix.

behdad:
I'm out of bits :(.

behdad:
> @jfkthame Right. I'm guessing that CoreText is treating regional-indicators as atomic units only? Let me find an alternative fix.

Nah.

The distinction can be made by looking at logical and reverse flags separately... Let me patch.

behdad:
I'm confused. I think the original RI change was wrong. Let me go back there and discuss.

behdad:
Okay I think I have a proper fix now: https://github.com/harfbuzz/harfbuzz/pull/3539

PR #3529, opened 2022-04-06
[morx] Distinguish logical vs backwards bits reversing

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3528

### DIFF
--- a/src/hb-aat-layout-morx-table.hh
+++ b/src/hb-aat-layout-morx-table.hh
@@ -992,6 +992,8 @@ struct Chain
     unsigned int count = subtableCount;
     for (unsigned int i = 0; i < count; i++)
     {
+      bool logical;
+      bool backwards;
       bool reverse;
 
       if (!(subtable->subFeatureFlags & flags))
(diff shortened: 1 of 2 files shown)


behdad:
Actually there's a problem now that I think about it...

When there's RTL text in the buffer, the buffer is in logical order, which means the glyphs are in LTR at this point, even though the buffer direction says LTR. That tricks the logic we had in morx if `logical` is false. Because that logic thinks the buffer content is LTR and flips it. That's what caused the flag bug to begin with...

behdad:
That is to say, I'm not sure what the effect of the `logical` bit should be.

behdad:
Humm. If the `logical` bit is 0, and buffer is `RTL`, and `backward` is 0, *then* we have to flip the buffer. That is opposite of what we are doing now.  @jfkthame do you agree?

behdad:
And with this new logic I think we should always just buffer-reverse; no grapheme business here.

behdad:
> When there's RTL text in the buffer, the buffer is in logical order, which means the glyphs are in LTR at this point, even though the buffer direction says RTL. That tricks the logic we had in morx if `logical` is false. Because that logic thinks the buffer content is LTR and flips it. That's what caused the flag bug to begin with...

Ignore this. I'm wrong.

behdad:
Humm. I think the core of the problem actually is this:
https://github.com/harfbuzz/harfbuzz/issues/3314#issuecomment-1092136064

behdad:
Okay I need to think more about this. More findings:
https://github.com/harfbuzz/harfbuzz/issues/3314#issuecomment-1103195816

PR #3536, opened 2022-04-19
Add cheat sheet

simoncozens:
This adds the GraphViz source to the [Harfbuzz Cheat Sheet](https://twitter.com/simoncozens/status/1422159944732839942/photo/1) I made. It doesn't hook it up to anywhere in the docs yet; I just don't want the code to get lost.


behdad:
Did you write this manually?

There's been a few changes recently...

simoncozens:
Yes, this is manually written. I'll check over the order of features again. I also thought it would be useful to annotate each feature with the context (some are whole-buffer and some are per-cluster).

PR #3552, opened 2022-04-25
Move freedesktop.org/wiki/HarfBuzz to README.md

khaledhosny:
I copied the text verbatim, but it probably needs to be edited. I’m not sure if there is more to move here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,21 @@ For a comparison of old vs new HarfBuzz memory consumption see [this][10].
 
 <!--See past and upcoming [HarfBuzz Hackfests](https://freedesktop.org/wiki/Software/HarfBuzz/Hackfests/)!-->
 
+## Name
+
+HarfBuzz (حرف‌باز) is my Persian translation of “[OpenType][1]”,
+transliterated using the Latin script. It sports a second meaning, but that
+ain’t translatable.
+
+> Background: Originally there was this font format called TrueType. People and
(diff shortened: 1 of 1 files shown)


khaledhosny:
I think I mixed things and this page is not what harfbuzz.org used to point to (now I see why it felt so thin in details).

Yup, this one is more like it https://web.archive.org/web/20190406161058/https://www.freedesktop.org/wiki/Software/HarfBuzz/

README.md
```diff
@@ -6,26 +6,94 @@
...
+
+Both development and user support discussion around HarfBuzz happens on the
+[github][4].
+
+To report bugs or submit patches please use [github][4] issues and
+pull-requests.
+
+For a comparison of old vs new HarfBuzz memory consumption see [this][10].
+
+<!--See past and upcoming [HarfBuzz Hackfests](https://freedesktop.org/wiki/Software/HarfBuzz/Hackfests/)!-->
```
khaledhosny, line 66: @behdad Do we want the contents of this page as well?

behdad:
Thanks. Some of this info is obsolete or not of interest anymore. Like the hackfests, the pango-vs-harfbuzz. Let's shorten it a bit.

README.md
behdad: Don't think so.

behdad:
Let's merge and cleanup after.

PR #3567, opened 2022-05-05
[subset] Fix cpal subsetting when there are partial palette overlaps.

garretrieger:
The existing code doesn't correctly handle the case where palettes partially overlap in the color record array. This changes the subsetting to only share entries in the color record array when palettes have the same first color index. Partially overlapping palettes will be converted to disjoint segments in the color record array.

Updates one of the color tests to use multiple palettes.

Also fixes fuzzer: https://oss-fuzz.com/testcase-detail/5568200165687296.


src/hb-ot-color-cpal-table.hh
```diff
@@ -197,37 +197,48 @@ struct CPAL
...
+      new_idx = layer_index * retained_color_indices.get_population ();
       if (!c->copy<HBUINT16> (new_idx)) return_trace (false);
     }
 
     c->push ();
-    for (const auto _ : retained_color_record_indices.iter ())
+    for (unsigned layer = 0; layer < first_color_index_for_layer.length; layer++)
     {
-      if (!c->copy<BGRAColor> (color_records[_]))
+      unsigned first_color_index = first_color_index_for_layer[layer];
```
behdad, line 222: You could iterate over enumerated vector here...

garretrieger: Fixed.

PR #3579, opened 2022-05-11
Subset retain buffer

behdad:


behdad:
My benchmark numbers are weird, so @garretrieger if you can measure this that would be great.

garretrieger (approved):

Issue #3591, opened 2022-05-17
[perf] hb_ot_tags_from_language() too slow


behdad:
[This was flagged by @matthiasclasen in https://gitlab.gnome.org/GNOME/gtk/-/issues/3334 https://gitlab.gnome.org/GNOME/gtk/uploads/a6da390ef2fd85d37dd594d88c4229ff/image.png]

The new `perf/benchmark-ot` shows this:
```
----------------------------------------------------------------------------------------------
Benchmark                                                    Time             CPU   Iterations
----------------------------------------------------------------------------------------------
BM_hb_ot_tags_from_script_and_language/COMMON zh_CN        172 ns          171 ns      4083155
BM_hb_ot_tags_from_script_and_language/COMMON en_US        120 ns          119 ns      5849947
BM_hb_ot_tags_from_script_and_language/LATIN en_US         113 ns          112 ns      5840326
BM_hb_ot_tags_from_script_and_language/COMMON none        4.66 ns         4.64 ns    151396224
BM_hb_ot_tags_from_script_and_language/LATIN none         4.66 ns         4.64 ns    149019593
```

Half of the time is spent in `hb_ot_tags_from_complex_language()`; in there, most of the time is spent in `subtag_matches` cases outside the `switch`.  Ideas to speed that up:
1. Make `subtag_matches` use `strchr` for `-` instead of `strstr` as main loop. Might or might not have any effect.
2. All the `subtag_matches` outside the switch match long strings (>= 6 or so). As such, check the tag for such length before going into any of them.

I implementing the second.

After that, bulk of the time I suppose is spent in binary-searching the language table. I suggest we split the language table in 2-letter and 3-letter tags, to speed-up the vast majority of cases that are 2-letter.

matthiasclasen:
♥

behdad:
> 2\. All the `subtag_matches` outside the switch match long strings (>= 6 or so). As such, check the tag for such length before going into any of them.

I did this now. New numbers:
```
----------------------------------------------------------------------------------------------
Benchmark                                                    Time             CPU   Iterations
----------------------------------------------------------------------------------------------
BM_hb_ot_tags_from_script_and_language/COMMON zh_CN        112 ns          112 ns      6357763
BM_hb_ot_tags_from_script_and_language/COMMON en_US       60.5 ns         60.3 ns     11475091
BM_hb_ot_tags_from_script_and_language/LATIN en_US        54.9 ns         54.8 ns     12575690
[...]

PR #3592, opened 2022-05-17
[ot-tags] Avoid strcmp in hb_ot_tags_from_language

dscorbett:
I thought of this while reviewing #3591. I don’t know whether the speed improvement is worth the extra space from `LangTag`’s two new fields.

### DIFF
--- a/src/hb-ot-tag.cc
+++ b/src/hb-ot-tag.cc
@@ -217,6 +217,8 @@ lang_matches (const char *lang_str, const char *limit, const char *spec, unsigne
 struct LangTag
 {
   hb_tag_t language;
+  unsigned int idx_for_same_tag;
+  unsigned int total_for_same_tag;
   hb_tag_t tag;
 
   int cmp (hb_tag_t a) const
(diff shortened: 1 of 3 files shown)


dscorbett:
And this is already redundant. Oh well. I think I will stop thinking about performance till I am sure you are done making changes.

behdad:
Can you check again with master? I removed the strcmp's already.

behdad:
Namely: https://github.com/harfbuzz/harfbuzz/commit/909f00ac6e6b3eb459f0553b84fe508bb697e9af

behdad:
I believe I'm done. :)

src/hb-ot-tag.cc
```diff
@@ -217,6 +217,8 @@ lang_matches (const char *lang_str, const char *limit, const char *spec, unsigne
 struct LangTag
 {
   hb_tag_t language;
+  unsigned int idx_for_same_tag;
+  unsigned int total_for_same_tag;
```
behdad, line 221: `uint16_t` please.

behdad:
A bit wasteful. Let me see the numbers and decide. Trying it now. Thanks.

behdad:
Size increase is 5.5kb and there's no measurable time improvement. I'd abandon.

PR #3595, opened 2022-05-19
[ot-tags] Fix `min_subtag_len` calculations

dscorbett:
`hb_ot_tags_from_complex_language` had two conditions guarding the sequence of checks that don’t check the language subtag.

The first condition was `limit - lang_str > 5`. I modified it to `limit - lang_str >= 7`. This is because 5 is the minimum length of a relevant language tag excluding the language subtag. The minimum language subtag length is 2. Therefore, the minimum total language tag length is 7. The practical effect is minimal: 6-character tags like `zh-cmn` no longer pass into that block.

The second condition was `(limit - lang_str == 5 && *lang_str == '-')`. That accepts language tags like `-geok`. However, language tags can’t begin with hyphens, so no valid input would ever pass this condition, so I removed it.

I also fixed gen-tag-table.py to add 1 for each subtag’s hyphen, instead 1 for just the first subtag’s hyphen. It currently makes no difference because none of these special cases check multiple subtags.

### DIFF
--- a/src/gen-tag-table.py
+++ b/src/gen-tag-table.py
@@ -1019,14 +1019,12 @@ for initial, items in sorted (complex_tags.items ()):
 		if not tags:
 			continue
 		subtag_len = 0
-		subtag_len += len(lt.script) if lt.script is not None else 0
-		subtag_len += len(lt.region) if lt.region is not None else 0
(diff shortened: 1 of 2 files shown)


behdad:
Thanks. I wasn't making assumption on the language subtag length.

PR #3596, opened 2022-05-19
[subset-perf] Signficiantly speed up ClassDef*::subset.

garretrieger:
Eliminates the usage of a glyph -> klass hash map and replaces it with a vector storing the mapping. This allows us to use the vector directly as the iterator driving the serialize. Approximately 1% speedup for Noto Nastaliq.


src/hb-ot-layout-common.hh
```diff
@@ -91,12 +91,12 @@ template<typename Iterator>
...
 
-static void ClassDef_remap_and_serialize (hb_serialize_context_t *c,
-					  const hb_map_t &gid_klass_map,
-					  hb_sorted_vector_t<HBGlyphID16> &glyphs,
-					  const hb_set_t &klasses,
-					  bool use_class_zero,
-					  hb_map_t *klass_map /*INOUT*/);
+static void ClassDef_remap_and_serialize (
+    hb_serialize_context_t *c,
+    hb_sorted_vector_t<hb_pair_t<hb_codepoint_t, hb_codepoint_t>> &glyph_and_klass,
```
behdad, line 96: Can you rewrite the signature of these as templates taking sorted iterator?

garretrieger: I'll update the argument order to put glyph_and_klass with the other IN/OUT arg

behdad: Sounds good.

PR #3603, opened 2022-05-20
Add font serial API

behdad:


behdad:
The hb-ft part is optional and just a convenience for users of that API.

The main reason I'm implementing this is such that I can implement scalar-caching for variable-font calculations, to fix the performance issues raised in https://github.com/harfbuzz/harfbuzz/pull/2878

behdad:
I used the new font-serial internally to implement advance-width caching on hb-ot; used the (internal) coords-serial for invalidation. This enables performance almost equal to non-variable shaping for variable fonts.  The cache is only used for variable-font non-default coordinates shaping.

Issue #2965, opened 2021-05-01
Implement 32bit VariationStore


behdad:
ala https://github.com/fonttools/fonttools/issues/2279

PR #3606, opened 2022-05-23
32bit varstore

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/2965

### DIFF
--- a/src/hb-ot-layout-common.hh
+++ b/src/hb-ot-layout-common.hh
@@ -2771,44 +2771,65 @@ struct VarData
     if (unlikely (!c->extend_min (this))) return_trace (false);
     itemCount = inner_map.get_next_value ();
 
-    /* Optimize short count */
-    unsigned short ri_count = src->regionIndices.len;
-    enum delta_size_t { kZero=0, kByte, kShort };
+    /* Optimize word count */
+    unsigned ri_count = src->regionIndices.len;
+    enum delta_size_t { kZero=0, kNonWord, kWord };
     hb_vector_t<delta_size_t> delta_sz;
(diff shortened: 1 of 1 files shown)


behdad:
This is ready. Would be nice to add tests, including for serializing / subsetting it...

src/hb-ot-layout-common.hh
```diff
@@ -2762,43 +2771,87 @@ struct VarData
...
+	  if (delta < -65536 || 65535 < delta)
+	  {
+	    has_long = true;
+	    break;
+	  }
+        }
+      }
+    }
+
+    if (has_long)
```
garretrieger, line 2802: You could reduce the code duplication between this and the !has_long branch by using a helper method that you pass the word size too.

garretrieger (approved):

garretrieger:
> This is ready. Would be nice to add tests, including for serializing / subsetting it...

Does fontTools currently support 32 bit var stores? I should be able to pretty easily add some subsetting tests which would cover serialization as well.

src/hb-ot-layout-common.hh
behdad: Thanks. I removed it by using a couple variables for the bounds check.

behdad:
> Does fontTools currently support 32 bit var stores?

It does.

> If so I should be able to pretty easily add some subsetting tests which would cover serialization as well.

GSUB/GPOS normally don't use 32bit values. Doesn't mean we cannot artificially construct such values though.

behdad:
Okay I manually tested against fonttools. Caught a bug and fixed.

PR #3608, opened 2022-05-24
[ci] Fix fedora-valgrind job

khaledhosny:
Fedora 33 is EOL since 2021-11-30, try the latest Fedora release (36).


behdad:
Thanks.

PR #3610, opened 2022-05-25
[subset] fix subsetting of faces created via hb_face_create_for_tables. 

### LINKED ISSUE #3609: [subset] Doesn't work with hb_face_create_for_tables.
If an `hb_face` is created with `hb_face_create_for_tables` and passed to `hb_subset_or_fail` then a valid but very empty `hb_face` is returned (0 tables). At least one should get nullptr to indicate an error, or better yet an actual subset. This appears to be due to `hb_subset_plan_execute_or_fail` assuming `success` is true by default and `hb_face_get_table_tags` always returns 0 for faces created with `hb_face_create_for_tables` since that interface currently doesn't allow the user to specify the tables. Therefore `hb_subset_plan_execute_or_fail` returns that is has "successfully" subset the 0 tables it could find.

This is especially interesting for cases when using CoreText which does not provide a general API for the full flat data of the original font, the API is table based where one can query which tables are available and then get a copy of each table on request. It would be nice not to be required to flatten all the tables and re-create a ttf header to do the subset, which is more or less going to make another (partial) copy of the data anyway.

For the short term just making the check to ensure that the `hb_face` is capable of listing its tables should be made in `hb_subset_plan_execute_or_fail` and it should return nullptr if it is not. Longer term it would be nice to add a new API like `hb_face_create_for_tables` which takes a both a pointer to a `hb_reference_table_func_t` and a pointer to a new `hb_list_tables_func_t` which could fill out a list of table tags available. Then the subsetter could successfully subset table based fonts.

garretrieger:
Fixes #3609 

If subsetting a face create by hb_create_for_tables then hb_face_get_table_tags doesn't work. So fallback processing all non-empty tables that are in the list of tables the subsetter can subset and the 'no subset tables' list.

### DIFF
--- a/src/hb-subset.cc
+++ b/src/hb-subset.cc
@@ -153,7 +153,7 @@ _get_table_tags (const hb_subset_plan_t* plan,
   if (num_tables)
     return hb_face_get_table_tags (plan->source, start_offset, table_count, table_tags);
 
-  // If face has 0 tables associated with it, assume that it has built from
+  // If face has 0 tables associated with it, assume that it was built from
   // hb_face_create_tables and thus is unable to list its tables. Fallback to
   // checking each table type we can handle for existence instead.
   auto it =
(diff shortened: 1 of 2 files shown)


behdad:
The iterator is a bit ugly but does the job.  Thanks.

Issue #3612, opened 2022-05-26
Race in hb_ot_tags_from_language


bungeman:
While rolling HarfBuzz in Chromium at https://chromium-review.googlesource.com/c/chromium/src/+/3668368 ThreadSanitizer detected a race which appears to be around `hb_ot_tags_from_language`. Still investigating.

bungeman:
The race appears to be on `last_tag_idx` introduced in dfca47f419b6ef5c6df813822e4b10a7cec92434 `[ot-tag] Cache last bsearch result`

bungeman:
I see that the intent here is to read from `last_tag_idx` exactly once to get a guess (and that guess may be anything, even unitialized is fine) and write to it exactly once to store the next guess. However, in the absence of any barriers or volatile, there is nothing currently preventing the compiler from optimizing away `last_tag_idx` completely, or worse read from it multiple times, since in the absence of any annotation the compiler may assume single threaded execution and optimize as it wishes.

behdad:
Thanks. I'll make it an atomic and use get/set_relaxed on it, which should satisfy thread-sanitizer.

behdad:
That said, that optimization was a bit misguided. I'll revert it.

behdad:
> I see that the intent here is to read from `last_tag_idx` exactly once to get a guess (and that guess may be anything, even uninitialized is fine) and write to it exactly once to store the next guess. However, in the absence of any barriers or volatile, there is nothing currently preventing the compiler from optimizing away `last_tag_idx` completely, or worse read from it multiple times, since in the absence of any annotation the compiler may assume single threaded execution and optimize as it wishes.

It's a static variable. It doesn't make any assumption on multiple threaded execution for effectiveness.

behdad:
@bungeman Can you see if the commit I just pished fixes the sanitizer issue? Thanks.

behdad:
Weird, I cannot reproduce this locally with our multithreaded test no matter how I try.

bungeman:
At https://chromium-review.googlesource.com/c/chromium/src/+/3668368/3 attempting to roll HarfBuzz in Chromium to the commit to use atomic ints (before the change was reverted) to see if this alleviates the issue. If this lands it will be interesting to attempt to roll to the revert and see if it re-introduces the issue.

behdad:
The revert was on a PR just to test.

behdad:
> The revert was on a PR just to test.

[...]

PR #3613, opened 2022-05-27
Threads test

behdad:


behdad:
cc @bungeman

behdad:
We should make sure thread sanitizer is hooked up to at least one bot.

behdad:
This test currently fails since it fails to find the fonts; it needs to look for fonts in the source tree.

behdad:
And we should add a similar subset test. This test is vastly based on the benchmark-shape test. cc @garretrieger

behdad:
> And we should add a similar subset test. This test is vastly based on the benchmark-shape test. cc @garretrieger

So much that maybe they should share some code.

behdad:
Test passes now.

behdad:
Apparently we have a tsan bot. So now just have to wait and see.

behdad:
I reverted  d61b2074915cf5f8044dcb8e3dafc04b5b58c6b8 but could not reproduce the issue @bungeman was facing in https://github.com/harfbuzz/harfbuzz/issues/3612 :(.

The tsan is apparently working though, since it caught an unprotected variable in the test itself.

behdad:
Humm. I rebuilt the library with HB_NO_MT and tsan still doesn't catch anything. Something's not set up correctly. (testing it locally). Investigating.

behdad:
Humm. Actually we already had `test/api/test-multithread` that did the same, and it fails with tsan if I build the library with `HB_NO_MT`. But my new test doesn't fail. No idea why.

behdad:
> Humm. Actually we already had `test/api/test-multithread` that did the same, and it fails with tsan if I build the library with `HB_NO_MT`. But my new test doesn't fail. No idea why.

Closing this PR as such.

behdad:
> Humm. Actually we already had `test/api/test-multithread` that did the same, and it fails with tsan if I build the library with `HB_NO_MT`. But my new test doesn't fail. No idea why.

My bad. There was a lock facing consecutive execution.

behdad:
Okay, I've fixed my test and confirmed that it catches things.

PR #3614, opened 2022-05-29
Coverage cache

behdad:
A PR to use a cache in coverage checking of GPOS (GSUB can be done as well).

Slows things down, so abandoning. Posting for posterity.


behdad:
I like to try this some more.

behdad:
> Slows things down, so abandoning. Posting for posterity.

I fixed up the slowdown. But doesn't speed things up. Mainly because only applies to lookups that have one subtable only...

PR #3617, opened 2022-05-30
Skip warning when building as subproject and  ragel is missing

xclaesse:
It is unlikely to be a developer build in that case.


xclaesse:
For context, this is blocking adding harfbuzz in Meson's WrapDB: https://github.com/mesonbuild/wrapdb/pull/285.

khaledhosny:
Thanks

PR #3625, opened 2022-06-01
[threads-test] Add a threads test against hb-subset.

garretrieger:


test/threads/hb-subset-threads.cc
```diff
@@ -0,0 +1,180 @@
...
+int main(int argc, char** argv)
+{
+  if (argc > 1)
+    num_threads = atoi (argv[1]);
+  if (argc > 2)
+    num_repetitions = atoi (argv[2]);
+
+  if (argc > 4)
+  {
+    num_tests = (argc - 3) / 2;
```
behdad, line 162: The `/ 2` is wrong. Remove.

PR #3628, opened 2022-06-03
Reduce complexity

behdad:
I know this is controversial since just renaming files. This change basically renames 'ot-shape-complex' to just 'ot-shaper' for brevity.


khaledhosny (approved):

khaledhosny:
My only concern is that commits that fail to build interfere badly with bisecting. Otherwise it is a welcome change, and I personally always disliked the stigma of “complex”.

behdad:
I'll rewrite those commits so it all builds. I agree re complex. That term was from a different era.

behdad:
> I'll rewrite those commits so it all builds.

Done.

behdad:
> My only concern is that commits that fail to build interfere badly with bisecting.

For changes like this I merge with a merge commit, such that bisect sees them as all-or-nothing.

behdad:
> > My only concern is that commits that fail to build interfere badly with bisecting.
> 
> For changes like this I merge with a merge commit, such that bisect sees them as all-or-nothing.

Though maybe this one is better without a merge commit. Higher chance that git can follow the logic.

khaledhosny:
> > My only concern is that commits that fail to build interfere badly with bisecting.
> 
> For changes like this I merge with a merge commit, such that bisect sees them as all-or-nothing.

Interesting, I didn’t know about this behavior of git bisect.

khaledhosny:
We still have `hb_ot_substitute_complex`, `hb_ot_position_complex`, do we want to rename them as well (I’m not sure what complex refers to here). There is also `//Complex shapers` in `hb-subset-input.cc`.

khaledhosny:
I’m updating the documentation to speak less about “complex” shapers/scripts.

behdad:
> We still have `hb_ot_substitute_complex`, `hb_ot_position_complex`, do we want to rename them as well (I’m not sure what complex refers to here). There is also `//Complex shapers` in `hb-subset-input.cc`.

Fixed.

behdad:
> > > My only concern is that commits that fail to build interfere badly with bisecting.
> > 
> > 
> > For changes like this I merge with a merge commit, such that bisect sees them as all-or-nothing.
> 
> Interesting, I didn’t know about this behavior of git bisect.

git-bisect takes the first parent of each commit. Merge commits have multiple parents, typically two. This typically becomes frustrating, when bisect tells you the first bad commit is a merge commit, you are left to your own to then do a separate bisect to find out where in the merged branch the regression was introduced, perhaps by another bisect with a different start point, which would route the bisect path to the merged branch.

behdad:
> git-bisect takes the first parent of each commit. Merge commits have multiple parents, typically two. This typically becomes frustrating, when bisect tells you the first bad commit is a merge commit, you are left to your own to then do a separate bisect to find out where in the merged branch the regression was introduced, perhaps by another bisect with a different start point, which would route the bisect path to the merged branch.

Looks like I'm wrong, and that's the behavior only if you specify `--first-parent`.

https://git-scm.com/docs/git-bisect#Documentation/git-bisect.txt---first-parent

What I said is still valid, I think, if the merged branch started its life before the starting point of the bisect (which is not the case for short-lived branches).

khaledhosny:
I see, thanks!

Issue #3354, opened 2022-01-03
Add minimally-sufficient BCP-47 language matching implementation


behdad:
Currently using our hb-ot-name API is cumbersome because eg. asking for "en_US" language name fails because font has "en" language name only. Fix this by adding minimally-sufficient BCP-47 language matching implementation.

cc @matthiasclasen

dscorbett:
It sounds like you want `get_index` to find an entry with a close enough language instead of necessarily exactly the same language. Is that right?

It would also make sense to change the language maps in hb-ot-name-language-static.hh to use more precise subtags, e.g. 0x0409 is `"en-US"` and 0x0809 is `"en-GB"`. The Microsoft language IDs all include regions; for each language we’ll have to pick one region as the default.

behdad:
Sounds about right.

matthiasclasen:
That would be great to have. I'd love to use it in pango too, if you expose some hb_language api for it

behdad:
Also see the rudimentary `pango_language_matches`:
https://www.manpagez.com/html/pango/pango-1.36.8/pango-Scripts-and-Languages.php#pango-language-matches

behdad:
> close enough language instead of necessarily exactly the same language. Is that right?

Not just close-enough. They might be exact language, but not exact language *tag*. Eg. `zh_CN` vs `zh-Hant-CN`, etc.

behdad:
> What about macrolanguages: do you want e.g. the requested language tag `pes` to match a font’s 0x0429 (`fa-IR`)?

I'd say no because `pes` is not in common use.

behdad:
> What about macrolanguages: do you want e.g. the requested language tag `pes` to match a font’s 0x0429 (`fa-IR`)?

But if you are automatically generating tables and it generates that, then sure, I have no problem with that.

behdad:
This is actually duplicate of https://github.com/harfbuzz/harfbuzz/issues/1372

behdad:
@matthiasclasen Maybe we can flesh out the API for this.

behdad:
So, the logic in `hb-ot-tag` applies here, right @dscorbett? We just need to take that out of `hb-ot-tag` and move it somewhere more generic.

behdad:
I added `hb_language_matches()` with a trivial prefix-matching implementation.

PR #3629, opened 2022-06-03
[name] Implement approximate language matching

behdad:
Very rudimentary.

Fixes https://github.com/harfbuzz/harfbuzz/issues/3354

### DIFF
--- a/src/hb-ot-name-table.hh
+++ b/src/hb-ot-name-table.hh
@@ -156,7 +156,7 @@ struct NameRecord
 };
 
 static int
-_hb_ot_name_entry_cmp_key (const void *pa, const void *pb)
+_hb_ot_name_entry_cmp_key (const void *pa, const void *pb, bool exact)
 {
   const hb_ot_name_entry_t *a = (const hb_ot_name_entry_t *) pa;
   const hb_ot_name_entry_t *b = (const hb_ot_name_entry_t *) pb;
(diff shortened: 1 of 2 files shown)


src/hb-ot-name-table.hh
```diff
@@ -169,16 +169,31 @@ _hb_ot_name_entry_cmp_key (const void *pa, const void *pb)
...
+
+  signed c = strcmp (astr, bstr);
+
+  if (!exact && c)
+  {
+    unsigned la = strlen (astr);
+    unsigned lb = strlen (bstr);
+    // 'a' is the user request, and 'b' is string in the font.
+    // If eg. user asks for "en-us" and font has "en", approve.
+    if (la > lb && astr[lb] == '-')
```
matthiasclasen, line 184: You don't actually check that the prefixes match here ? Needs a strncmp (astr, bstr, lb) == 0

It doesn't matter for these short strings, but you could write that check without entirely iterating both strings twice,

test/api/test-ot-name.c
```diff
@@ -93,10 +93,15 @@ test_ot_name (void)
   name_id = entries[3].name_id;
   g_assert_cmpuint (3, ==, name_id);
   lang = entries[3].language;
+
   g_assert_cmpstr (hb_language_to_string (lang), ==, "en");
   g_assert_cmpuint (27, ==, hb_ot_name_get_utf8 (face, name_id, lang, &text_size, text));
   g_assert_cmpuint (9, ==, text_size);
   g_assert_cmpstr (text, ==, "FontForge");
+
+  g_assert_cmpuint (27, ==, hb_ot_name_get_utf8 (face, name_id, hb_language_from_string ("en_US", -1), &text_size, text));
```
matthiasclasen, line 102: your fuzzy check above looks for '-', this uses '_'. Does hb_language_from_string normalize ?

behdad: It does normalize.

src/hb-ot-name-table.hh
behdad: Oops right. Fixed using the slow strncmp version.

matthiasclasen (commented):
Looks good now.

PR #3394, opened 2022-01-26
Code reorg for HarfBuzz-next

behdad:
Eventually fixes https://github.com/harfbuzz/harfbuzz/issues/3372
Supersedes https://github.com/harfbuzz/harfbuzz/pull/3370

PR #3630, opened 2022-06-03
Fix check-* scripts when harfbuzz is a subproject

xclaesse:
When harfbuzz is a subproject paths are in the form
"subprojects/harfbuzz/src/...". Instead of removing "src/" prefix, take
the absolute path and make it relative to current source dir.

This fix regression introduced in
https://github.com/harfbuzz/harfbuzz/pull/3394.


src/meson.build
```diff
@@ -718,6 +718,7 @@ if get_option('tests').enabled()
 
   env = environment()
   env.set('srcdir', meson.current_source_dir())
+  env.set('top_srcdir', meson.source_root())
```
khaledhosny, line 721: This probably should be added to `TESTS_ENVIRONMENT` in `Makefile.am` as well.

xclaesse: Done. Out of curiosity, is there any plan to drop autotools in favor of Meson in the future?

khaledhosny: There is an vague plan to drop it at some point in the future and I think all HarfBuzz developers are using meson now, but supporting autotools is low effort and we are familiar with it (unlike, say, CMake), so we are probably going to keep it around for a little while yet. After all, we are still keeping CMake support even though we are not actively maintaining it.

src/Makefile.am
```diff
@@ -437,6 +437,7 @@ endif
 
 TESTS_ENVIRONMENT = \
 	srcdir="$(srcdir)" \
+	top_srcdir="$(srcdir)" \
```
behdad, line 440: This has to be `top_srcdir=top_srcdir`, or `top_srcdir=srcdir` is correct?

xclaesse: I think you're right, fixed.

PR #3635, opened 2022-06-05
Dedup khmer/myanmar categories

behdad:


dscorbett:
The Khmer and Myanmar shapers still use hb-ot-shaper-indic-table.cc so their categories still have to match. This PR obscures that relationship.

behdad:
Well this reduces one set of duplications. I'll add a comment.

We should re-evaluate that relationship. Can we separate the Khmer / Myanmar tables and generate them separately, with their own overrides ala USE table?

dscorbett:
That sounds like a fine idea. It might be more space-efficient to keep all three shapers’ data in the same table; maybe it doesn’t matter.

I think we should change everything about their generating scripts, tables, and overrides to match the simpler way USE does it. For USE, the overrides are done at build time in the generating script, whereas for the others the overrides are applied at runtime. For USE, there is one kind of category, but for the others, there are two which are mapped to each other, like `INDIC_SYLLABIC_CATEGORY_GEMINATION_MARK` and `OT_SM`.

behdad:
> That sounds like a fine idea. It might be more space-efficient to keep all three shapers’ data in the same table; maybe it doesn’t matter.

It doesn't matter. And we would exclude Myanmar & Khmer from Indic table when the move happens.


> I think we should change everything about their generating scripts, tables, and overrides to match the simpler way USE does it. For USE, the overrides are done at build time in the generating script, whereas for the others the overrides are applied at runtime.

Right. We should go even further and put the overrides in a text file with the same format as the Indic data files and read those, instead of hardcoding them.

> For USE, there is one kind of category, but for the others, there are two which are mapped to each other, like `INDIC_SYLLABIC_CATEGORY_GEMINATION_MARK` and `OT_SM`.

For USE, the Unicode data is converted to USE categories in the generator. For Indic, the conversion happens in the shaper currently. It should be moved to the generator.

behdad:
Superseded by https://github.com/harfbuzz/harfbuzz/pull/3648

PR #3636, opened 2022-06-05
Classdef cache

behdad:
[layout] Use a cache for main input ClassDef of ChainContextLookups
    
    I expected this to speed up NotoNastaliqUrdu, but didn't make a dent.
    Need to profile on Linux to understand why. But it's mostly because
    the cache is only effective when we enter the lookup from the main
    loop (apply_forward), but NotoNastaliq mostly recurses...

### DIFF
--- a/src/hb-buffer.hh
+++ b/src/hb-buffer.hh
@@ -132,9 +132,7 @@ struct hb_buffer_t
    * Managed by enter / leave
    */
 
-#ifndef HB_NDEBUG
   uint8_t allocated_var_bits;
-#endif
   uint8_t serial;
   hb_buffer_scratch_flags_t scratch_flags; /* Have space-fallback, etc. */
   unsigned int max_len; /* Maximum allowed len. */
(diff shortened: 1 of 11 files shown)


behdad:
The cache has a hit rate of 10% on NotoNastaliqUrdu shaping. So I'm baffled that I'm not seeing any speedup whatsoever.

behdad:
I see ~2 percent speedup and ~1 percent overhead, adding up to net ~1 percent speedup...

behdad:
Phew! Finally some progress!

Looks like NotoNastaliqUrdu spends most of its time in GPOS ContextFormat2 (dot positioning, etc...), and I was only applying the cache to ChainContextFormat2. Applying the cache to ContextFormat2 as well shows 16% speedup in NotoNastaliqUrdu.  I observe a slowdown of less than 3% in Amiri as the overhead of new infrastructure, and much less than that for other tests.

Interestingly, the Gulzar font does a lot more lookups than NotoNastaliqUrdu, but is 7 times faster. Need to investigate when I have a profiler (Linux machine) around.

behdad:
cc @jfkthame @drott

simoncozens:
I didn’t use format 2 to format 3 converter in Gulzar. I should try it.

behdad:
> I didn’t use format 3 to format 2 converter in Gulzar. I should try it.

I hope you mean format 2 to format 3.

behdad:
> > I didn’t use format 3 to format 2 converter in Gulzar. I should try it.
> 
> I hope you mean format 2 to format 3.

Nvmind. I see you edited your comment. format 3 to format 2 converter is correct.

behdad:
> I observe a slowdown of less than 3% in Amiri as the overhead of new infrastructure

Upon further measurement the slowdown is nothing like that. That one should have been noise. I don't see any measurable slowdown whatsoever.

behdad:
From last commit I pushed:

[layout-cache] Cache lookahead, not input, classdef in ChainContextFo…
…rmat2

From the commit:

+    /* For ChainContextFormat2 we cache the LookaheadClassDef instead of InputClassDef.
+     * The reason is that most heavy fonts want to identify a glyph in context and apply
+     * a lookup to it. In this scenario, the length of the input sequence is one, whereas
+     * the lookahead / backtrack are typically longer.  The one glyph in input sequence is
+     * looked-up below and no input glyph is looked up in individual rules, whereas the
+     * lookahead and backtrack glyphs are tried.  Since we match lookahead before backtrack,
+     * we should cache lookahead.  This decisions showed a 20% improvement in shaping of
+     * the Gulzar font.

PR #3637, opened 2022-06-05
[layout] Use NEON SIMD instructions for apply_forward()

behdad:
[layout] Use NEON SIMD instructions for apply_forward()
    
    To skip four glyphs at a time.
    
    Unfortunately slows down benchmark-shape by up to 25% for Amiri, instead of
    speeding anything up. This is due to the overhead of switching back and forth
    between the ARM and NEON processors.
    
    To be experimented with: Same architecture but with AVX on x86_64 to skip 8
    glyphs at a time.
    
    https://github.com/harfbuzz/harfbuzz/issues/566

Posting for posterity.

### DIFF
--- a/src/hb-ot-layout-gsubgpos.hh
+++ b/src/hb-ot-layout-gsubgpos.hh
@@ -3824,7 +3824,8 @@ struct hb_ot_layout_lookup_accelerator_t
   }
   void fini () { subtables.fini (); }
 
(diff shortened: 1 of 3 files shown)


behdad:
Reopening to try with AVX.

behdad:
AVX2 is in https://github.com/harfbuzz/harfbuzz/pull/3657

PR #3650, opened 2022-06-11
[subset] Fix fuzzer issue.

garretrieger:
Fixes https://oss-fuzz.com/testcase-detail/5693568490012672. new_index should be set from new_index2 when the entry is present in the map.


src/hb-ot-post-table-v2subset.hh
```diff
@@ -61,7 +61,8 @@ HB_INTERNAL bool postV2Tail::serialize (hb_serialize_context_t *c,
     HBUINT8 *o = c->allocate_size<HBUINT8> (HBUINT8::static_size * (s.length + 1));
     if (unlikely (!o)) return_trace (false);
     if (!c->check_assign (o[0], s.length, HB_SERIALIZE_ERROR_INT_OVERFLOW)) return_trace (false);
-    memcpy (o+1, s.arrayZ, HBUINT8::static_size * s.length);
+    if (s.length)
+      memcpy (o+1, s.arrayZ, HBUINT8::static_size * s.length);
```
behdad, line 65: We have `hb_memcpy` for this.

PR #3653, opened 2022-06-11
Indic cleanup2

behdad:
Here's the promised duplication-removal branch. I'm actually happy with it.


behdad:
@dscorbett This is ready as well, and addresses all the remaining comments of yours.

behdad:
@dscorbett friendly ping here.

src/gen-indic-table.py
```diff
@@ -82,6 +82,58 @@
...
+  'myanmar' : [
+    'VAbv',
+    'VBlw',
+    'VPre',
+    'VPst',
+
+    'IV',
+    'As',
+    'DB',
+    'GB	',
```
dscorbett, line 126: There’s a stray tab here.

dscorbett (approved):
There is a typo, but it doesn’t end up breaking anything. I approve.

PR #3661, opened 2022-06-16
[repack] always run the sort in repack.

garretrieger:
This is needed to ensure virtual link ordering constraints are respected when repack is being called from fontTools. For subset usage, repack won't be called if the graph doesn't already overflow so this change doesn't cause any extra work.


behdad:
Tests need updating?

garretrieger:
> > Tests need updating?
> 
> Hmm weird, the comparison is only failing under the asan/tsan bots. I'll see if I can track down why.

Ah, it's because the repack api is experimental API only, so the tests only run w/ experimental api enabled. I'll updated the test output.

garretrieger:
> > > Tests need updating?
> > 
> > 
> > Hmm weird, the comparison is only failing under the asan/tsan bots. I'll see if I can track down why.
> 
> Ah, it's because the repack api is experimental API only, so the tests only run w/ experimental api enabled. I'll updated the test output.

Tests are fixed now.

PR #3662, opened 2022-06-16
[subset] Fix potential out of bounds write setting overlap flag on composite glyphs.

garretrieger:


garretrieger:
> Is there a fuzzer issue this fixes?

The external fuzzers can't currently reach this issue because they don't run subsets with the set_overlap_bit option. It would be good to update the fuzzers to take a bit field that sets the subsetting options from the input to get more complete coverage.

PR #3665, opened 2022-06-17
No complex languages

behdad:
This saves around 7% size in TINY build so I'm inclined to take it.


dscorbett (changes requested):
“Complex” in the sense of `hb_ot_tags_from_complex_language` approximately means “containing multiple subtags”. A three-letter tag is not complex. How about calling it `HB_NO_LANGUAGE_LONG` or `HB_ONLY_LANGUAGE_TINY` or something like that instead?

The block conditioned on `(s && limit - lang_str >= 6)` can be ifdeffed out too, because extended language subtags are always three letters long.

With this change, `hb_ot_ambiguous_tag_to_language` can return language tags that `hb_ot_tags_from_script_and_language` would not interpret correctly. I’m not sure the API explicitly guarantees any relationship between the functions but it seems weird for them to be inconsistent.

behdad:
> HB_NO_LANGUAGE_LONG

Sounds good. Thanks.

dscorbett:
Do you want to do anything about `hb_ot_ambiguous_tag_to_language`?

behdad:
> Do you want to do anything about `hb_ot_ambiguous_tag_to_language`?

My hunch is no. What do you suggest we do?

behdad:
> > Do you want to do anything about `hb_ot_ambiguous_tag_to_language`?

Or just disable it?

behdad:
The tag-to-script/language mapping should probably also be disabled under HB_TINY.

dscorbett:
I think it would be a good idea to disable this under `HB_NO_LANGUAGE_LONG`:
https://github.com/harfbuzz/harfbuzz/blob/a15ad778fede9e94428f6811293c71a63216234a/src/hb-ot-tag.cc#L475-L487

That leaves `hb_ot_tag_to_language` with just the algorithmic “x-hbot” part, which is tiny.

behdad:
> I think it would be a good idea to disable this under `HB_NO_LANGUAGE_LONG`:
> 
> https://github.com/harfbuzz/harfbuzz/blob/a15ad778fede9e94428f6811293c71a63216234a/src/hb-ot-tag.cc#L475-L487
> 
> That leaves `hb_ot_tag_to_language` with just the algorithmic “x-hbot” part, which is tiny.

So, disable even the 2-letter language table you mean?

behdad:
Disabled disambiguate_tags as well.

dscorbett:
> So, disable even the 2-letter language table you mean?

That is what I meant, but it is okay to leave it enabled too.

behdad:
> > So, disable even the 2-letter language table you mean?
> 
> That is what I meant, but it is okay to leave it enabled too.

That wouldn't save us much since the data still has to be there.

dscorbett:
https://github.com/harfbuzz/harfbuzz/blob/a4d5c3ca08145998aef774c30458f79025d2987b/src/hb-ot-tag.cc#L331-L336
That assumption is only safe if `ot_languages3` has already been checked; see #2735. This block should be ifdeffed out too.

behdad:
Done.

dscorbett (approved):

Issue #3478, opened 2022-03-12
Question: Is directWrtite required by default dll builds?


EyeAndTea:
Greetings.

The dlls for Harfbuzz distributed for the 4.0.1 release appear to require DirectWrite. It is my understanding from elsewhere that by default this would not be the case. I got an error for a missing "dwrite.dll" under Windows XP.

behdad:
No DirectWrite is not required by HarfBuzz. The Windows binaries we distribute might have it enabled for testing reasons. It's unfortunate what you are facing. cc @khaledhosny

behdad:
I'm actually confused as to how it was being linked, given that I just added it to the link line here:
https://github.com/harfbuzz/harfbuzz/commit/ea5131507a723b4858f6a90584351ac14a990ecb

because without it it was failing to link using mingw32 for me. Any ideas?

behdad:
I see the meson build is linking, but not the autotools build. That should explain I guess.

lb90:
Hi! I can help on this. The problem is that https://github.com/harfbuzz/harfbuzz/pull/3670 causes a segfault on MSVC builds?

behdad:
> Hi! I can help on this. The problem is that #3670 causes a segfault on MSVC builds?

Yes.

> How can one launch the failing test manually?

Just invoking hb-view with any font and text, and `--shaper=directwrite` should do.

PR #3670, opened 2022-06-20
Revert "Revert "Revert "[hb-directwrite] Don't load dwrit.dll dynamically"""

khaledhosny:
This reverts commit 361a438658dcddea29d7c8b9c68bf2bc88109bde.

Lets see if DirectWrite is still crashing.

Fixes https://github.com/harfbuzz/harfbuzz/issues/3478

### DIFF
--- a/src/hb-directwrite.cc
+++ b/src/hb-directwrite.cc
@@ -43,6 +43,14 @@
  * Functions for using HarfBuzz with DirectWrite fonts.
  **/
 
+/* Declare object creator for dynamic support of DWRITE */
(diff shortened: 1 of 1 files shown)


behdad:
What was the conclusion?

khaledhosny:
This revert causes DirectWrite shaper to sigfault (at least on CI), https://github.com/harfbuzz/harfbuzz/runs/6969868027?check_suite_focus=true#step:8:186. Apparently the original commit that this revert restores broke something.

PR #3674, opened 2022-06-25
[use] Switch Sinhala to USE

### LINKED ISSUE #1044: Move Sinhala to USE
See https://github.com/harfbuzz/harfbuzz/issues/967 for some context.  Sinhala shaping apparently changed between Win7 and Win10. We used to match Win7 behavior. Then hacked our shaper to be closer to Win10. Moving to USE should be a better long-term fix.

### LINKED ISSUE #1246: Fallback for Sinhala Conjuncts
There is the problem with rendering a request for a Sinhala conjunct followed by a left-reordering matra if the conjunct is not in the font.  For a fairly sure example of such a lack, consider the following encoding for the rare Pali syllable thve ථ්‍වෙ <U+0DAE MAHAAPRAANA TAYANNA, U+0DCA AL-LAKUNA, U+200D ZWJ, U+0DC0 VAYANNA, U+0DD9 KOMBUVA>.

With Windows 7 Uniscribe and with HarfBuzz, it renders like thēva ථේව <U+0DAE, U+0DDA DIGA KOMBUVA, U+0DC0> .

On Notepad on Windows 10 and on email on iPhone, it renders like Sinhalese thve ථ්වෙ  <U+0DAE, U+0DCA, U+0DC0, U+0DD9>, and I endorse this fallback.  As the al-lakuna is an unchanged virama glyph, it should force a syllable break, so the left reordering matra should be written before the second consonant.

An unsatisfied request for touching consonants, e.g. Pali thve as ථ‍්වෙ <U+0DAE, U+200D, U+0DCA,  U+0DC0, U+0DD9>, is currently rendered the same more or less as <U+0DAE, U+0DCA, U+0DC, U+0DD9>.  Some fonts, e.g. LKLUG, fail to achieve the proper interaction between the glyphs of the initial consonant and al-lakuna; allowing the glyph of U+200D to block it appears to be a font error.  (A satisfied request results in the matra appearing at the left of the cluster and the lack of an explicit al-lakuna glyph.)

It is arguable that touching glyphs would be a better fallback for a missing conjunct, but I think that should be a font issue rather than a renderer issue, unless user control were to be included somehow.

dscorbett:
Fixes #1044 and therefore also fixes #1246.

I implemented it without any compatibility hacks. Microsoft’s USE has been shaping Sinhala for years without acquiring any script-specific workarounds, so it is safe to assume HarfBuzz can do the same.

The Indic machine allowed `forced_rakar` after a vowel sign. This was added in 422ecd2d3c198a36d07d409341cb82ea57c7ad6b to match Uniscribe. Sinhala was never supposed to use that sequence, but it worked in Uniscribe (probably by accident) so people used it. However, in Windows Server 2019, Uniscribe inserts a dotted circle before the U+0DCA, so this hack doesn’t work in Windows anymore, so it needn’t in HarfBuzz.

Iskoola Pota 6.96 no longer has the extra ZWJ-less rule in 'rphf'. Using a font that has both rules in 'rphf', as in older versions of Iskoola Pota, Uniscribe can apply either rule. There is no special case in USE for Sinhala to require ZWJ in 'rphf'.

Previously, HarfBuzz could decompose vowel signs “Khmer-style” or using the Unicode decompositions, depending on what the font expected. With this pull request, it only uses Unicode decompositions.
https://github.com/harfbuzz/harfbuzz/blob/1555b3008197b77d142587322c0aec3fa83bc99f/src/hb-ot-shaper-indic.cc#L1579-L1581
What fonts break if we unconditionally use Unicode decompositions? Hopefully they don’t matter anymore.


behdad:
> What fonts break if we unconditionally use Unicode decompositions? Hopefully they don’t matter anymore.

The commit suggests lklug.ttf. Commit is from 2012. I'm happy to go with this.

PR #3678, opened 2022-06-25
Split glyf

behdad:
A lot more can be done. Just a start.


behdad:
Err. I pushed this to main accidentally as well.

PR #3681, opened 2022-06-27
[lao] Decompose and reorder U+0EB3 around U+0EBB

dscorbett:
U+0EBB LAO VOWEL SIGN MAI KON behaves like other above-base marks in the context of U+0EB3 LAO VOWEL SIGN AM in Notepad in Windows Server 2019. I changed HarfBuzz to match.

U+0EBB has no Thai counterpart. There are multiple marks that only exist in one block or the other, and the existing code did not try to detect reserved code points, so I didn’t do anything to prevent U+0E33 THAI CHARACTER SARA AM from reordering around the reserved U+0E3B. This does not match Uniscribe, but I guess reserved code points don’t matter.

### DIFF
--- a/src/hb-algs.hh
+++ b/src/hb-algs.hh
@@ -857,6 +857,11 @@ hb_in_ranges (T u, T lo1, T hi1, T lo2, T hi2, T lo3, T hi3)
 {
   return hb_in_range (u, lo1, hi1) || hb_in_range (u, lo2, hi2) || hb_in_range (u, lo3, hi3);
 }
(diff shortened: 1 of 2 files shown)


behdad:
Thanks. Any tests maybe?

dscorbett:
There aren’t any tests for decomposing sara am. I can write some tomorrow.

Oh, a fuzz test just failed. Is that because of my change?

behdad:
No that's my fault...

PR #3682, opened 2022-06-27
Fix the annotation for hb_blob_get_data

matthiasclasen:
This function will return NULL for the the
empty blob. That is important information for
bindings that treat nullability as a type trait.


matthiasclasen:
I don't think the fuzzer failure is my fault

Issue #3683, opened 2022-06-27
objects returned by hb_directwrite_face_create aren't usable


matthiasclasen:
What I mean is that I have a hard time creating a cairo scaled font for such a hb_font_t.

We first tried to use cairo_dwrite_font_face_create_for_dwrite_fontface, but that fails because hb_directwrite_face_get_font_face doesn't return the object that the hb_font_t was created from.

Then we tried to use cairo-ft. That fails because hb_face_reference_blob doesn't give me the font data for such fonts.

The situation for coretext fonts on macOS may be a little better because hb_coretext_font_create seems to keep around the native object. Can the directwrite api be fixed to do the same?

matthiasclasen:
Or maybe there is an alternative scenario where I should use hb_face_builder in pango to create hb_face_t objects that have a blob?

behdad:
> The nicest (and for me, easiest) would be if we had
> 
> ```
> cairo_font_face_t *cairo_hb_font_face_create_for_hb_face (hb_font_face_t *face)
> ```
> 
> it looks like that would work nicely, by using hb_face_reference_table for the load_truetype_table implementation

Right. Or cairo had a load-table constructor for font-face...

matthiasclasen:
https://gitlab.gnome.org/GNOME/pango/-/merge_requests/593/diffs?commit_id=6a21d3456ee61b0b1adb8f1ce8b30c3772fcc0d7

is the workaround I came up with, using the face builder

behdad:
@matthiasclasen see if #3686 helps?

behdad:
Another question is: in the `_get_` API: if the font was NOT created with a native object, but we already have created a native object internally, should we return that? That happens if user has shaped something with the respective native shaper already.

behdad:
> Another question is: in the `_get_` API: if the font was NOT created with a native object, but we already have created a native object internally, should we return that? That happens if user has shaped something with the respective native shaper already.

I'm leaning toward no. But for CoreText shaper this becomes an issue the way we currently do things...

matthiasclasen:
> I'm leaning toward no.

Yeah, separating that out might be cleaner. 

What will happen in cross-api scenarios ? Create a face with hb_coretext_face_create, then call hb_ft_font_lock_face

behdad:
> What will happen in cross-api scenarios ? Create a face with hb_coretext_face_create, then call hb_ft_font_lock_face

Right now it returns null.

[...]

PR #3686, opened 2022-06-27
[dwrite] Return existing fontface if any

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3683

### DIFF
--- a/src/hb-directwrite.cc
+++ b/src/hb-directwrite.cc
@@ -854,6 +854,9 @@ hb_directwrite_face_create (IDWriteFontFace *font_face)
 IDWriteFontFace *
 hb_directwrite_face_get_font_face (hb_face_t *face)
 {
+  if (face->destroy == _hb_directwrite_font_release)
+    return (IDWriteFontFace *) face->user_data;
+
   return face->data.directwrite->fontFace;
 }
 


src/hb-directwrite.cc
```diff
@@ -854,6 +854,9 @@ hb_directwrite_face_create (IDWriteFontFace *font_face)
 IDWriteFontFace *
 hb_directwrite_face_get_font_face (hb_face_t *face)
 {
+  if (face->destroy == _hb_directwrite_font_release)
+    return (IDWriteFontFace *) face->user_data;
+
   return face->data.directwrite->fontFace;
```
matthiasclasen, line 860: Should the final return just be NULL?

Under what circumstances would you fail the if condition, but have a non-NULL face->data.directwrite->fontFace ?

behdad: If the font was NOT constructed from a directwrite object... In which case we try to create a dwrite object. Maybe we should separate that as a different API? Or add new API for just returning the original object?

behdad: CoreText has the same issue currently.

behdad:
I suggest we add a `_or_create` variation of the getter API.

src/hb-directwrite.cc
behdad: > Comparing to the coretext code, I don't know - that code doesn't do any checks, and seems to just assume that data.coretext exists ?

data.* is automatically and lazily created...

behdad: As I said, CoreText is also broken in various ways.

PR #3693, opened 2022-06-28
avar2

behdad:
Implements https://github.com/be-fonts/boring-expansion-spec/issues/14


rsheeter (approved):
Nit: it would be nice if code comments directed the reader to the docs on where to learn more, particularly given you won't find this in normal places you go to look at the spec

behdad:
Will do. Thanks.

Issue #3688, opened 2022-06-28
OpenType MATH fuzzing coverage


fred-wang:
cc @behdad @drott @kcc 

Hello, Igalia sent an [intent to ship for MathML in Chromium](https://www.igalia.com/2022/06/22/Intent-to-Ship-MathML.html) and there were question about support by fuzzing tools. After discussion with Dominik I understand one thing we can do is to try to improve coverage for the OpenType MATH table.

There is already some coverage reported at https://storage.googleapis.com/oss-fuzz-coverage/harfbuzz/reports/20220627/linux/src/harfbuzz/src/report.html ; but surprisingly some parts of MathConstants's get_constant from hb-ot-math-table.hh are not covered.

Can someone please provide some guidance about how I can improve that?

Thanks!

behdad:
Hi @fred-wang,

All you need to do is to add calls to that API in `test/api/test-ot-face.c:test_font()` function.

fred-wang:
@behdad Thanks! Will give a try tomorrow...

PR #3695, opened 2022-06-29
[math] Improve fuzzing coverage

fred-wang:
Extend testing to cover parts that are missing according to the recent
oss-fuzz-coverage report:
- Retriving all constants from MathConstants.
- Retrieving entries from MathKern, MathGlyphPartRecord and
  MathGlyphAssembly.
- Retrieving italic correction from MathGlyphAssembly.
- Choosing between horizontal/vertical offset in MathVariants.

https://storage.googleapis.com/oss-fuzz-coverage/harfbuzz/reports/20220627/linux/src/harfbuzz/src/hb-ot-math-table.hh.html
https://github.com/harfbuzz/harfbuzz/issues/3688


fred-wang:
@kcc @behdad Do we need a math font in test/fuzzing/fonts/ to "help" the oss-fuzz to find new branches? I believe test/api/fonts/MathTestFontFull.otf would do the work.

PR #3697, opened 2022-06-29
[reorg] Move GSUB.hh up one level and change GSUB namespace to GSUB_impl.

### LINKED ISSUE #3677: Adjust new namespace layout
The new internal namespace layout needs some adjustment. For GSUB we are doing:
```
namespace OT::Layout::GSUB {
struct GSUB;
}

namespace OT {
struct GSUB_accelerator_t;
}
```

This scheme fails for non-accelerated tables. Eg.:
```
namespace OT::glyf {
...glyf's impl...
}
namespace OT {
struct glyf;
}
```
this errs because `OT::glyf` is defined both as a namespace and struct. Moving `glyf` inside `OT::glyf` doesn't work because the face object needs to find it at `OT_glyf`. I suggest we change things to:
```
namespace OT::glyf_impl {
...
}
namespace OT {
struct glyf;
}
```

cc @garretrieger

garretrieger:
Fixes #3677

### DIFF
--- a/src/Makefile.sources
+++ b/src/Makefile.sources
@@ -118,7 +118,7 @@ HB_BASE_sources = \
 	OT/Layout/GSUB/ExtensionSubst.hh \
 	OT/Layout/GSUB/SubstLookupSubTable.hh \
 	OT/Layout/GSUB/SubstLookup.hh \
-	OT/Layout/GSUB/GSUB.hh \
+	OT/Layout/GSUB.hh \
 	OT/Layout/GPOS.hh \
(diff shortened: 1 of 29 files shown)


behdad:
Thanks. I suggest we still keep GSUB.hh and GPOS.hh in their respective directories. WDYT?

garretrieger:
> Thanks. I suggest we still keep GSUB.hh and GPOS.hh in their respective directories. WDYT?

Works for me. I'll update to move them back.

PR #3699, opened 2022-06-30
[subset] Add support for --layout-scripts

### LINKED ISSUE #2815: hb-subset feature request: features and language selection
Similar to `fontTools.subset`'s `--layout-features[+|-]=<feature>[,<feature>...]` and `--layout-scripts[+|-]=<script>[,<script>...]` switches, but with a twist: `-=` means "everything except what I specify henceforth".

garretrieger:
Allows a set of script tags to be specified which filters the subset by script tag. Fixes #2815


src/hb-subset-plan.cc
```diff
@@ -91,94 +91,102 @@ _remap_indexes (const hb_set_t *indexes,
 typedef void (*layout_collect_func_t) (hb_face_t *face, hb_tag_t table_tag, const hb_tag_t *scripts, const hb_tag_t *languages, const hb_tag_t *features, hb_set_t *lookup_indexes /* OUT */);
 
 
-template <typename T>
-static void _collect_layout_indices (hb_face_t		  *face,
-                                     const T&              table,
-                                     const hb_set_t	  *layout_features_to_retain,
-                                     layout_collect_func_t layout_collect_func,
-                                     hb_set_t		  *indices /* OUT */)
+/**
```
khaledhosny, line 94: Comments starting with `/**` are interpreted by gtk-doc, which I think not the intention here.

khaledhosny (commented):
FontTools supports also specifying language tags using, e.g., `arab.URD`, in case you want to also support this (https://github.com/fonttools/fonttools/pull/2112).

garretrieger:
> FontTools supports also specifying language tags using, e.g., `arab.URD`, in case you want to also support this ([fonttools/fonttools#2112](https://github.com/fonttools/fonttools/pull/2112)).

Ah got it, I'll plan add that in a second PR after this one then.

src/hb-subset.h
```diff
@@ -100,6 +100,8 @@ typedef enum { /*< flags >*/
  * @HB_SUBSET_SETS_NAME_LANG_ID: the set of name lang ids that will be retained.
  * @HB_SUBSET_SETS_LAYOUT_FEATURE_TAG: the set of layout feature tags that will be retained
  * in the subset.
+ * @HB_SUBSET_SETS_LAYOUT_SCRIPT_TAG: the set of layout script tags that will be retained
+ * in the subset. Defaults to all tags.
```
khaledhosny, line 104: I thought no we need to add `Since: REPLACEME` here.

src/hb-subset-plan.cc
```diff
@@ -91,94 +91,102 @@ _remap_indexes (const hb_set_t *indexes,
...
+                                     hb_set_t		  *indices /* OUT */)
+{
+  unsigned num_features = table.get_feature_count () + 1;
+  hb_vector_t<hb_tag_t> features;
+  if (!plan->check_success (features.resize (num_features))) return;
+  table.get_feature_tags (0, &num_features, features.arrayZ);
+  features.resize (num_features);
+  bool retain_all_features = !_filter_tag_list (&features, plan->layout_features);
+
+  unsigned num_scripts = table.get_script_count () + 1;
```
behdad, line 139: What's the +1 for? And why a second resize below?

garretrieger: That was meant to add extra space for the null terminating entry, but this was the wrong place for it. I've moved the +1 into the filter method which produces the vector that actually has the null entry added to the end.

src/hb-subset-plan.cc
```diff
@@ -99,7 +99,7 @@ static bool _filter_tag_list(hb_vector_t<hb_tag_t>* tags, /* IN/OUT */
                              const hb_set_t* filter)
 {
   hb_vector_t<hb_tag_t> out;
-  out.alloc (tags->get_size());
+  out.alloc (tags->get_size() + 1); // +1 is to allocate room for the null terminator.
```
behdad, line 102: `alloc` just allocates. Do you need to do a push at the end?

behdad: Please use `HB_TAG_NONE`. For features as well.

PR #3704, opened 2022-07-05
Towards breaking the 64k in GSUB/GPOS

behdad:


behdad:
This seems complete now, and I've tested it and it decodes with no sanitize issue, and shapes correctly in my quick testing, font generated by https://github.com/fonttools/fonttools/pull/2678.

garretrieger (approved):

PR #3706, opened 2022-07-06
Load DirectWrite dynamically

lb90:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3478

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -308,9 +308,12 @@ if (WIN32 AND HB_HAVE_UNISCRIBE)
 endif ()
 
 if (WIN32 AND HB_HAVE_DIRECTWRITE)
+  check_include_file_cxx(dwrite_1.h, HAVE_DWRITE_1_H)
+  if (NOT ${HAVE_DWRITE_1_H})
+    message(FATAL_ERROR "DirectWrite was enabled explicitly, but required header is missing")
+  endif()
   add_definitions(-DHAVE_DIRECTWRITE)
   list(APPEND project_headers ${PROJECT_SOURCE_DIR}/src/hb-directwrite.h)
-  list(APPEND THIRD_PARTY_LIBS dwrite rpcrt4)
 endif ()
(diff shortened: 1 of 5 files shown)


lb90:
extern "C" shouldn't really matter, do you prefer to have it removed?

behdad:
Oh, because it's just a typedef.

lb90:
> We should also remove the -ldwrite from Makefile.am and meson.build.

Right! Should be OK now

lb90:
Looks like [UuicCreate](https://docs.microsoft.com/en-us/windows/win32/api/rpcdce/nf-rpcdce-uuidcreate) is only used in `hb-uniscribe.cc`:

    $ grep -r --exclude-dir=".git" UuidCreate
    perf/texts/en-words.txt:UuidCreate
    src/hb-uniscribe.cc:  UuidCreate ((UUID*) &id);

It's also the only function imported from `rpcrt4.lib`:

    RPCRT4.dll
             2AE6627E8 Import Address Table
             2AE662360 Import Name Table
                     0 time date stamp
                     0 Index of first forwarder reference

                         22F UuidCreate

So I took the liberty to remove `rpcrt4.lib` from `CMakeLists.txt`, right where it adds dependencies needed for `hb-directwrite.cc`. `rpcrt4` is still linked if Uniscribe/GDI is enabled. Hope it's ok 😄

behdad:
> Looks like [UuidCreate](https://docs.microsoft.com/en-us/windows/win32/api/rpcdce/nf-rpcdce-uuidcreate) is only used in `hb-uniscribe.cc`:

We should actually test if the same code is needed in the directwrite backend, but for now it LGTM.

The issue that solves is this: if someone tries to shape a font with the uniscribe backend and the font has the same name as a font installed on the system, uniscribe will prefer the font installed on the system instead of the font provided. For that reason, we rewrite the font to change the name to an arbitrary UUID so the name is unique.

Issue #3705, opened 2022-07-06
build: `struct std::hash' in different namespace` error


kleisauke:
It looks like PR https://github.com/harfbuzz/harfbuzz/pull/3623 broke the build with a somewhat old GCC version (6.5.0).
<details>
  <summary>Details</summary>
  
```
[2/65] Compiling C++ object src/libharfbuzz-subset.a.p/hb-subset-input.cc.o
FAILED: src/libharfbuzz-subset.a.p/hb-subset-input.cc.o 
arm-rpi-linux-gnueabihf-g++ -Isrc/libharfbuzz-subset.a.p -Isrc -I../src -I. -I.. -I/target/include -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -std=c++11 -fno-rtti -O3 -fno-exceptions -fno-rtti -fno-threadsafe-statics -fvisibility-inlines-hidden -DHAVE_CONFIG_H -marm -mcpu=arm1176jzf-s -mfpu=vfp -mfloat-abi=hard -Os -fPIC -D_GLIBCXX_USE_CXX11_ABI=1 -fno-asynchronous-unwind-tables -ffunction-sections -fdata-sections -fPIC -Wno-non-virtual-dtor -MD -MQ src/libharfbuzz-subset.a.p/hb-subset-input.cc.o -MF src/libharfbuzz-subset.a.p/hb-subset-input.cc.o.d -o src/libharfbuzz-subset.a.p/hb-subset-input.cc.o -c ../src/hb-subset-input.cc
In file included from ../src/hb.hh:473:0,
                 from ../src/hb-subset.hh:31,
                 from ../src/hb-subset-input.cc:27:
../src/hb-cplusplus.hh:170:13: error: specialization of 'template<class _Tp> struct std::hash' in different namespace [-fpermissive]
 struct std::hash<hb::shared_ptr<T>>
             ^~~~~~~~~~~~~~~~~~~~~~~
In file included from /root/tools/x64-gcc-6.5.0/arm-rpi-linux-gnueabihf/arm-rpi-linux-gnueabihf/include/c++/6.5.0/bits/basic_string.h:5655:0,
                 from /root/tools/x64-gcc-6.5.0/arm-rpi-linux-gnueabihf/arm-rpi-linux-gnueabihf/include/c++/6.5.0/string:52,
[...]

PR #3707, opened 2022-07-06
Fix build on GCC < 7

kleisauke:
Fixes: https://github.com/harfbuzz/harfbuzz/issues/3705.

(See https://github.com/harfbuzz/harfbuzz/compare/4.4.1...kleisauke:harfbuzz:workaround-gcc-bug-4.4.1 for a rebased version on top of 4.4.1)


mojca:
@kleisauke: thank you very much (in the name of another project that depends on Harfbuzz whose build broke on Solaris 10 and Debian 9).

PR #3709, opened 2022-07-06
Fixes for DWrite header checks

lb90:
Here are a few fixes, mainly targeting CMake build definitions. I have now tested all the three build systems on Windows and everything is working as intended \o/

Sorry that didn't make it in the previous PR!


behdad:
Thanks a lot!

Issue #3716, opened 2022-07-13
multiple warnings in hb-algs.hh and hb-ot-metrics.cc, possible misuse of comma operator, clang


kaie:
Building with clang 14.0.5 I get multiple warnings about possible misuse of comma operator

One example is:
mozilla/gfx/harfbuzz/src/hb-algs.hh:983:43: error: possible misuse of comma operator here [-Werror,-Wcomma]
for(; a < end; a++, b++) { ((tmp) = (*a), (*a) = (*b), (*b) = (tmp)); }

See https://phabricator.services.mozilla.com/D151590 for the places where I had to disable the warning.

behdad:
Does this work? https://github.com/harfbuzz/harfbuzz/pull/3719

behdad:
clang 14.0.0 doesn't catch those. We have -Wcomma in our error list...

PR #3719, opened 2022-07-13
More -Wcomma fixes

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3716


behdad:
Can you try again? Thanks.

Yeah, I'm on Fedora and don't see these.

kaie:
Will try your patch in a moment.
I've just compared the build commands, they're identical on both machines.

Only difference is that the old system ran through sccache, and the one through ccache.

kaie:
Yes! This builds fine. Thanks!

PR #3720, opened 2022-07-13
[repacker] Increase max_rounds when called via public api.

garretrieger:
Fix for https://github.com/fonttools/fonttools/issues/2691


behdad:
Magic lies ahead...

Issue #1573, opened 2019-02-09
Arabic shaper: pause after rclt/calt breaks Sakkal Kitab


emuller-amazon:
In the font Sakkal Kitab, the rclt=calt and liga/dlig/ss01 features have interleaved lookups:

rclt and calt : 17-18, 20-42, 65-67
liga: 15-16, 43-63, 69
dlig: 64
ss01: 68

The formation of ligatures depends on that interleaving. For example, for the sequence <0646, 062C, 0627>, a noon-jeem ligature is intended and is formed if the lookups are interleaved:

![interleaved](https://user-images.githubusercontent.com/16089970/52519460-6d845a00-2c10-11e9-8961-16b5db76a1d4.png)

but is not formed if the lookups are not interleaved:

![notinterleaved](https://user-images.githubusercontent.com/16089970/52519466-7d9c3980-2c10-11e9-986a-5c48d17f36bb.png)

The shaper currently has a pause after rclt/calt, which prevents the interleaving.

This pause was introduced by commit c1432bc, on July 14, 2017, in order to fix issue #505, 'Incorrect shaping with of <0644, 064E, 0670, 0653, ...> with "KFGQPC Uthmanic Script HAFS".'  What's ironic is that I created that issue! There are other problems with that font (starting with the hijacking of code points, handling of end of ayah), so it cannot really be used for anything but the (non strictly Unicode) text of the Qu'ran provided by KFGQPC. 

Unfortunately, I cannot provide a copy of Sakkal Kitab here.

behdad:
@emuller-amazon can you share the font with me for testing?

We need to find a pause order that satisfies all fonts we know of, instead of breaking KFGQPC in favor of this one... Would have been easier if someone in MS could share what they actually do.

emuller-amazon:
I am afraid I don't have a license that allows me to do that. Sorry.

khaledhosny:
I can contact Mamoun Sakkal and ask for a copy for testing, if @behdad still wants to look into this.

behdad:
@khaledhosny That would help indeed. I''m sure you can look into it yourself as well.

behdad:
> @khaledhosny That would help indeed. I''m sure you can look into it yourself as well.

Nevermind. There's an email thread that Eric started with me and Mamoun / Aida around the same time. (it was my last week at Google.) But I never received the font file. I'll CC you there so we can debug and make a resolution.

behdad:
[...]

PR #3724, opened 2022-07-14
[arabic] Disable calt if rclt is present

behdad:
And apply rclt without a pause.

Fixes https://github.com/harfbuzz/harfbuzz/issues/1573

### DIFF
--- a/src/hb-ot-shaper-arabic.cc
+++ b/src/hb-ot-shaper-arabic.cc
@@ -201,9 +201,6 @@ collect_features_arabic (hb_ot_shape_planner_t *plan)
    * work.  However, testing shows that rlig and calt are applied
    * together for Mongolian in Uniscribe.  As such, we only add a
    * pause for Arabic, not other scripts.
-   *
-   * A pause after calt is required to make KFGQPC Uthmanic Script HAFS
-   * work correctly.  See https://github.com/harfbuzz/harfbuzz/issues/505
    */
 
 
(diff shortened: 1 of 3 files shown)


khaledhosny:
I don’t understand this change. I thought it means `calt` will be disabled if `rclt` is present, but I have a font with both features and both are applied even with this PR. Is it because `calt` is still one of the default features enabled for all shapers?

behdad:
> I don’t understand this change. I thought it means `calt` will be disabled if `rclt` is present, but I have a font with both features and both are applied even with this PR. Is it because `calt` is still one of the default features enabled for all shapers?

Ah. Right. In other words what it ends up doing; and can be written in that way, is that if `rclt` is present, it won't pause after `calt`.

src/hb-ot-shaper-arabic.cc
```diff
@@ -232,10 +229,13 @@ collect_features_arabic (hb_ot_shape_planner_t *plan)
   if (plan->props.script == HB_SCRIPT_ARABIC)
     map->add_gsub_pause (arabic_fallback_shape);
 
-  /* No pause after rclt.  See 98460779bae19e4d64d29461ff154b3527bf8420. */
-  map->enable_feature (HB_TAG('r','c','l','t'), F_MANUAL_ZWJ);
-  map->enable_feature (HB_TAG('c','a','l','t'), F_MANUAL_ZWJ);
-  map->add_gsub_pause (nullptr);
+   map->enable_feature (HB_TAG('c','a','l','t'), F_MANUAL_ZWJ);
+   /* https://github.com/harfbuzz/harfbuzz/issues/1573 */
+   if (!map->has_feature (HB_TAG('r','c','l','t')))
```
behdad, line 234: This should check for `calt`, NOT `rclt`, humm? The code as written doesn't make sense: it only adds rclt if the font does not have rclt!! I think this essentially removed rclt from Arabic? @khaledhosny

khaledhosny: `rclt` is still added by `hb_ot_shape_collect_features()` since it is included in `horizontal_features`:
https://github.com/harfbuzz/harfbuzz/blob/a070f9ebbe88dc71b248af9731dd49ec93f4e6e6/src/hb-ot-shape.cc#L298-L308

I think the effect of this change is adding a pause after `calt` only if there is no `rclt` in the font. Which is essentially what you said in: https://github.com/harfbuzz/harfbuzz/pull/3724#issuecomment-1184801987

I think the `map->enable_feature (HB_TAG('r','c','l','t'), F_MANUAL_ZWJ);` is superflous and should be dropped, it only confuses things.

behdad: Thanks. Dropped it.

PR #3726, opened 2022-07-15
Ft bitmap

behdad:


behdad:
> * Build when `FT_Get_Transform` is not available.

@khaledhosny any chance you can work on this?

khaledhosny:
> > > * Build when `FT_Get_Transform` is not available.
> > 
> > 
> > @khaledhosny any chance you can work on this?
> 
> Sure, I’ll do right now.

I guess I need to restore parts of the old code and make it conditional, since now tests are failing on CI with old FreeType.

behdad:
I don't understand the failure.

khaledhosny:
Neither I. But I can reproduce it locally by removing the FT_Get_Transform check from meson.build.

behdad:
Seems like a bitmap font that one.

behdad:
Worked around it.

behdad:
Honestly I don't know how it was working, and how it works now, for a bitmap font!

behdad:
Looks like `FT_Set_Char_Size` does set the size variables even if it fails?

behdad:
I'm very confused now.

behdad:
Nope. I was wrong.

behdad:
> Looks like `FT_Set_Char_Size` does set the size variables even if it fails?

Correct. So I made the second `FT_Set_Char_Size` be conditional on having `FT_Get_Transform` as well and test is fixed.

behdad:
> * Figure out the vertical metrics issue;

`ft_face->ascender` et al are zero in the bitmap case. Sigh...

behdad:
> > * Figure out the vertical metrics issue;
> 
> `ft_face->ascender` et al are zero in the bitmap case. Sigh...

Fixed.

behdad:
Still doesn't work with negative font size. But I'm willing to let that go for now.

behdad:
@khaledhosny I think this is ready.

behdad:
Wait. This broke the normal negative font sizes (non-bitmap). Sigh.

behdad:
> Still doesn't work with negative font size. But I'm willing to let that go for now.

Fixed.

Issue #2786, opened 2020-12-27
Deploy documentation only in release builds


khaledhosny:
We were doing this, but switched to always deploying new documentation so improvements get to users fast as we had little documentation initially. But I think it is weird to have documentation on the site for APIs that are not yet part of the release.

alerque:
Agree. The release cadence is currently such that the published docs only need to match the last stable release.

behdad:
Sgtm.

PR #3730, opened 2022-07-18
[ci] Deploy docs only on tagged builds

khaledhosny:
Fixes https://github.com/harfbuzz/harfbuzz/issues/2786

### DIFF
--- a/.ci/deploy-docs.sh
+++ b/.ci/deploy-docs.sh
@@ -3,7 +3,10 @@
 set -x
 set -o errexit -o nounset
 
-TAG="$(git describe --exact-match --match "[0-9]*" HEAD 2>/dev/null || true)"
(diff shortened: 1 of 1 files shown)


.ci/deploy-docs.sh
```diff
@@ -17,8 +20,8 @@ cp ../build/docs/html/* .
 
 git init
 git branch -m main
-git config user.name "Travis CI"
-git config user.email "travis@harfbuzz.org"
+git config user.name "CI"
+git config user.email "ci@harfbuzz.org"
```
behdad, line 24: Maybe change to `harfbuzz-admin@googlegroups.com` while at it?

behdad (approved):

.ci/deploy-docs.sh
khaledhosny: Done.

Issue #3733, opened 2022-07-19
Add harfbuzz-subset.cc similar to harfbuzz.cc


khaledhosny:
We can make use of it in harfbuzzjs.

behdad:
Sounds good. And sounds straightforward in both build systems.

PR #3734, opened 2022-07-19
[subset] Add amalgam harfbuzz-subset.cc

khaledhosny:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3733


src/Makefile.am
```diff
@@ -160,6 +160,23 @@ pkginclude_HEADERS += $(HB_SUBSET_headers)
...
+	$(AM_V_GEN) \
+	for f in \
+		$(HB_BASE_sources) \
+		$(HB_GLIB_sources) \
+		$(HB_FT_sources) \
+		$(HB_GRAPHITE2_sources) \
+		$(HB_UNISCRIBE_sources) \
+		$(HB_GDI_sources) \
+		$(HB_DIRECTWRITE_sources) \
+		$(HB_CORETEXT_sources) \
```
behdad, line 173: You *just* need `HB_SUBSET_sources` here to build harfbuzz-subset library.

behdad:
Maybe add a sort command to the autotools path as well?

src/Makefile.am
khaledhosny: The idea is to build a (static) standalone hb-subset, but may be that is too WASM-specific. I guess one can always `c++ harfbuzz.cc harfbuzz-subset.cc` to achieve the same.

behdad: > I guess one can always c++ harfbuzz.cc harfbuzz-subset.cc to achieve the same.

Yes, if that works, I prefer that. Then `harfbuzz-subset.cc` will be equivalent to `libharfbuzz-subset`.

behdad: Okay, then at least just include the base HB sources. No need for any auxiliary sources; unless you need those as well.

khaledhosny: Done.

khaledhosny:
> Maybe add a sort command to the autotools path as well?

Done.

PR #3740, opened 2022-07-20
Simplify 'using' statements in namespace 'OT::Layout'.

ascherer:
Just an observation.

### DIFF
--- a/src/OT/Layout/GPOS/GPOS.hh
+++ b/src/OT/Layout/GPOS/GPOS.hh
@@ -7,10 +7,9 @@
 #include "PosLookup.hh"
 
 namespace OT {
-
-using Layout::GPOS_impl::PosLookup;
-
 namespace Layout {
+
+using GPOS_impl::PosLookup;
 
 static void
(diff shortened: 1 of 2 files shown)


ascherer:
Just another observation.

behdad:
This was changed to current way to fix gcc <= 7 build apparently.

kleisauke:
FWIW, this change should probably be fine with GCC < 7. I also tried this initially, but some files expect `PosLookup` and `SubstLookup` to be present within the `OT` namespace (hence the CI failures on this PR).

ascherer:
That was my second observation above. The use of `OT::SubstLookup` appears in files that seem to date prior to sub-namespaces `Layout` and `GSUB`(`_impl`). In the new version **5.0.1** the code base is somewhat cleaner in this respect. Thanks!

(Apart from observing compilation and use of Harfbuzz in the TeX Live code base – which broke hard four weeks ago with **GCC 5** and **GCC 6** –, I'm not invested in this software.)

PR #3742, opened 2022-07-20
[subset] WIP experimental new closure algorithm

garretrieger:
Utilizes information from the shaper to guide GSUB glyph closure.


behdad:
This seems like a failed experiment.

Issue #3739, opened 2022-07-20
[subset] COLR table closure need not repeat multiple times


behdad:
I don't think it does, because glyphs produced by it cannot be consumed by the table itself.

behdad:
I might be wrong though.

qxliu76:
So V0 and V1 closures are separate. 
V0 closure method is straightforward, just adding all referenced gids for base gids(through firstLayerIndex and numLayers). I think we would need to repeat this multiple times for V0 until we reach a stable gid set.

V1 closure is graph traversing so there's no need to repeat when V0 tables do not exist. I'm thinking I could add some code to break the loop when V0 tables do not exist.

When both V0 and V1 tables exist, we would need to repeat both. Let me know what you think

behdad:
> V0 closure method is straightforward, just adding all referenced gids for base gids(through firstLayerIndex and numLayers). I think we would need to repeat this multiple times for V0 until we reach a stable gid set.

I don't think so, no. No need to repeat. Just a one-time map.

behdad:
That's because, at least in V0, when a glyph maps to a set of layers, those layers are non-color layers, so no need to recurse on them.

qxliu76:
> That's because, at least in V0, when a glyph maps to a set of layers, those layers are non-color layers, so no need to recurse on them.

oh OK, if these are non-color layers we don't need to repeat, I'll come up with a PR for this

khaledhosny:
Fixed with 0f800769379d05a3086cd3fc56f0c4c1a19076f8

PR #3743, opened 2022-07-20
[subset] Do not repeat COLR table closure

qxliu76:
This is for #3739


src/hb-subset-plan.cc
```diff
@@ -314,22 +314,16 @@ static void _colr_closure (hb_face_t *face,
...
-
-    glyphs_colred->union_ (glyphset_colrv0);
-
-    //closure for COLRv1
-    colr.closure_forV1 (glyphs_colred, &layer_indices, &palette_indices);
-  } while (iteration_count++ <= HB_CLOSURE_MAX_STAGES &&
-           glyphs_num != glyphs_colred->get_population ());
+  // Collect all glyphs referenced by COLRv0
+  hb_set_t glyphset_colrv0;
+  for (hb_codepoint_t gid : glyphs_colred->iter ())
```
behdad, line 320: I think you don't need the `->iter ()` here.

qxliu76: there's an error without `->iter()`: error: no matching function for call to ‘begin(hb_set_t*&)’
Seems either `+hb_iter ()` or `->iter () `would work.

behdad: Ah it's a pointer. Then just `*glyphs_colred`.

qxliu76: Done. Thank you!

PR #3745, opened 2022-07-21
[repacker] Extension promotion mechanism in the repacker

garretrieger:
When enabled analyzes the non-extension lookups within a GSUB/GPOS table and decides if any need to be promoted into extensions prior to space assignemnt.

This is v1 of the algorithm which doesn't factor in shared nodes between lookups.

### DIFF
--- a/src/hb-ot-layout-gsubgpos.hh
+++ b/src/hb-ot-layout-gsubgpos.hh
@@ -3850,7 +3850,7 @@ struct ExtensionFormat1
     return_trace (dest_offset.serialize_subset (c, src_offset, this, get_type ()));
   }
 
-  public: // TODO
+  protected:
   HBUINT16	format;			/* Format identifier. Set to 1. */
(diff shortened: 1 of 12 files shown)


src/hb-ot-layout-gsubgpos.hh
```diff
@@ -3988,7 +3988,7 @@ struct GSUBGPOSVersion1_2
 {
   friend struct GSUBGPOS;
 
-  protected:
+  public: // TODO
```
behdad: Add a friend statement if you can. Otherwise fine with me.

behdad:
I'm not very well-versed in this code. Looks good to me in general.

src/hb-ot-layout-gsubgpos.hh
garretrieger: Ended up just adding a method to access the field I needed.

behdad:
Is this ready now?

garretrieger:
Yeah

PR #3759, opened 2022-07-27
[instance] update glyf/hmtx/vmtx/OS2/post tables

qxliu76:


src/hb-ot-os2-table.hh
```diff
@@ -183,6 +198,26 @@ struct OS2
 
     _update_unicode_ranges (c->plan->unicodes, os2_prime->ulUnicodeRange);
 
+    if (c->plan->user_axes_location->has (HB_TAG ('w','g','h','t')) &&
+        !c->plan->pinned_at_default)
+    {
+      float weight_class = c->plan->user_axes_location->get (HB_TAG ('w','g','h','t'));
+      if (!c->serializer->check_assign (os2_prime->usWeightClass,
+                                        roundf (hb_max (1.0f, hb_min (weight_class, 1000.0f))),
```
behdad, line 206: There's `hb_clamp`.

src/OT/glyf/glyf.hh
```diff
@@ -389,6 +403,30 @@ glyf::_populate_subset_glyphs (const hb_subset_plan_t   *plan,
   ;
 }
 
+inline void
+glyf::_compile_subset_glyphs_with_deltas (const hb_subset_plan_t *plan,
```
garretrieger, line 407: This compilation allocates an individual buffer for each glyph. It's probably better to create a second serialization buffer the same size as the final serialization buffer and then write the compiled glyph byte blobs into the serialization buffer. Then the final serialization will transfer them into the final positions with appropriate padding in the final serialization context.

src/OT/glyf/Glyph.hh
```diff
@@ -72,12 +72,107 @@ struct Glyph
...
+
+    int yMin, yMax;
+    yMin = yMax = roundf (all_points[0].y);
+
+    for (unsigned i = 0; i < all_points.length - 4; i++)
+    {
+      xMin = hb_min (xMin, roundf (all_points[i].x));
+      xMax = hb_max (xMax, roundf (all_points[i].x));
+      yMin = hb_min (yMin, roundf (all_points[i].y));
+      yMax = hb_max (yMax, roundf (all_points[i].y));
```
behdad, line 124: You're rounding each point twice. Do it once.

src/OT/glyf/SimpleGlyph.hh
```diff
@@ -206,6 +206,132 @@ struct SimpleGlyph
 	&& read_points (p, points_, end, &contour_point_t::y,
 			FLAG_Y_SHORT, FLAG_Y_SAME);
   }
+
+  static void compile_points (int value,
```
behdad, line 210: This looks like should be named `encode_coord`.

src/OT/glyf/SimpleGlyph.hh
```diff
@@ -206,6 +206,132 @@ struct SimpleGlyph
...
+    }
+    else
+    {
+      int16_t val = value;
+      coords.push (val >> 8);
+      coords.push (val & 0xff);
+    }
+  }
+
+  static void compile_flags (uint8_t &flag,
```
behdad, line 236: Also this should be named `encode_flag`?

src/OT/glyf/SubsetGlyph.hh
```diff
@@ -55,6 +58,20 @@ struct SubsetGlyph
...
   }
 
+  bool compile_bytes_with_deltas (const hb_subset_plan_t *plan,
+                                  hb_font_t *font,
+                                  const glyf_accelerator_t &glyf)
+  { return source_glyph.compile_bytes_with_deltas (plan, font, glyf, dest_start, dest_end); }
+
+  void free_compiled_bytes ()
+  {
+    if (dest_start.arrayZ)
```
behdad, line 68: I believe it's safe to call `fini()` without checking `arrayZ`.

src/OT/glyf/SubsetGlyph.hh
```diff
@@ -55,6 +58,20 @@ struct SubsetGlyph
...
+      dest_start.fini ();
+
+    if (dest_end.arrayZ)
```
behdad, line 71: Same.

src/hb-ot-hmtx-table.hh
```diff
@@ -130,33 +132,59 @@ struct hmtxvmtx
...
+		    return hb_pair (0u, 0);
+		  int lsb = 0;
+		  (void) _mtx.get_leading_bearing_without_var_unscaled (old_gid, &lsb);
+		  return hb_pair (_mtx.get_advance_without_var_unscaled (old_gid), +lsb);
+		})
+      ;
+
+      table_prime->serialize (c->serializer, it, num_long_metrics);
+    }
+    else
```
behdad, line 166: Why this code duplication? Can we avoid it?

src/hb-ot-hmtx-table.hh
```diff
@@ -330,6 +358,22 @@ struct hmtxvmtx
     hb_blob_ptr_t<V> var_table;
   };
 
+  unsigned get_advance (const hb_subset_plan_t *plan,
```
behdad, line 361: I don't like this subset-specific `get_advance` here. If anything, name it properly. You notice I named all `get_advance` functions like `with_variations_unscaled` etc.

src/hb-ot-os2-table.hh
```diff
@@ -183,6 +198,26 @@ struct OS2
 
     _update_unicode_ranges (c->plan->unicodes, os2_prime->ulUnicodeRange);
 
+    if (c->plan->user_axes_location->has (HB_TAG ('w','g','h','t')) &&
```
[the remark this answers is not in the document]
qxliu76: yes this is handled by `hb_subset_input_pin_axis_location` already at command line level

src/OT/glyf/SubsetGlyph.hh
qxliu76: I think we would need to check `arrayZ` cause the `fini()` for hb_bytes_t() doesn't check it.
In case of an empty glyph(e.g .notdef), the dest_start and dest_end would be empty hb_bytes_t(). And we do not want to free nullptr.

src/hb-ot-hmtx-table.hh
qxliu76: The code is basically like:
if (a)
{
  auto it = {iterator generation code}
  serialize (it); 
}
else
{
  auto it = {iterator generation code}
  serialize (it);
}

I tried to avoid duplicated code for serialize() call, but it seems there's not an easy way. Because the type of iterator will be different in either branch. So the serialize call also uses different argument types, thus I'll need two distinct calls

src/hb-ot-hmtx-table.hh
qxliu76: I changed it to `get_new_gid_advance_unscaled`. It gets the advance from mtx_map in subset_plan when there're variations, otherwise it calls get_advance_without_var_unscaled

src/OT/glyf/glyf.hh
qxliu76: Uh, I think maybe we don't need a second buffer cause top level glyf table serializes glyphs one by one. See:  https://github.com/harfbuzz/harfbuzz/blob/main/src/OT/glyf/glyf.hh#L50, and padding is handled during each SubsetGlyph serialization: https://github.com/harfbuzz/harfbuzz/blob/main/src/OT/glyf/SubsetGlyph.hh#L20

src/hb-ot-hmtx-table.hh
qxliu76: done.

src/OT/glyf/SubsetGlyph.hh
behdad: Freeing nullptr is safe I believe.

src/OT/glyf/Glyph.hh
```diff
@@ -72,12 +72,109 @@ struct Glyph
...
+    return compile_header_bytes (plan, all_points, dest_start);
+  }
+
   /* Note: Recursively calls itself.
    * all_points includes phantom points
    */
   template <typename accelerator_t>
   bool get_points (hb_font_t *font, const accelerator_t &glyf_accelerator,
 		   contour_point_vector_t &all_points /* OUT */,
+		   contour_point_vector_t &deltas, /* OUT */
```
behdad, line 177: I don't understand what the deltas here is. This function is very performance critical.

src/OT/glyf/SimpleGlyph.hh
```diff
@@ -206,6 +206,132 @@ struct SimpleGlyph
 	&& read_points (p, points_, end, &contour_point_t::y,
 			FLAG_Y_SHORT, FLAG_Y_SAME);
   }
+
+  static void encode_coords (int value,
```
behdad, line 210: `encode_coord` please. Singular, since it encodes a single value.

src/OT/glyf/SimpleGlyph.hh
```diff
@@ -206,6 +206,132 @@ struct SimpleGlyph
...
+    }
+    else
+    {
+      int16_t val = value;
+      coords.push (val >> 8);
+      coords.push (val & 0xff);
+    }
+  }
+
+  static void encode_flags (uint8_t &flag,
```
behdad, line 236: `encode_flag`, singular.

behdad:
This slows down glyph-drawing by up to 10%. I'd rather not do that. Test with:
```
build/perf/benchmark-font --benchmark_filter=glyph_shape.*hb
```

src/OT/glyf/Glyph.hh
behdad: This slows down glyph-drawing by up to 10%. I'd rather not do that. Test with:
```
build/perf/benchmark-font --benchmark_filter=glyph_shape.*hb
```

qxliu76: get_points() gets a flattened list of all contour points. When instancing composite glyphs, we actually only need the deltas for x/y offset, no need to do recursive call on child glyphs. So I added code to fill these deltas only when level is 0. If that slows down glyph-drawing, I can make another function to get points  for Composite glyphs only.

src/OT/glyf/SubsetGlyph.hh
qxliu76: uh,...thank you! It's nullptr is not NULL

src/OT/glyf/Glyph.hh
qxliu76: uh... I think for recalculating boundaries, we still need to expand child glyphs' points. So I would probably just add a new function for subset use, which is mainly a copy of get_points()

garretrieger: get_points() is a pretty sizable method so we should try and find a way to avoid having to duplicate the code in it.

To avoid duplicating all of the code from get_points() while maintaining performance you could try templating the method (ie. maybe like a boolean template parameter that says wether or not to get the deltas). That way in the case where you don't need to get deltas the templated function should compile to equivalent as the current version.

src/OT/glyf/glyf.hh
garretrieger: The most important thing is it would be nice to use a serialize context throughout the compilation could instead of memcopying into raw buffers.

src/OT/glyf/Glyph.hh
behdad: I agree. Either template or just passing a nullptr for deltas.

src/OT/glyf/glyf.hh
behdad: Also agree.

src/OT/glyf/Glyph.hh
qxliu76: Done

src/OT/glyf/glyf.hh
qxliu76: Here's my understanding for the original subset code: we need to know if `use_short_loca` is true before we could write each glyph into serializer buffer(padding depends on this) https://github.com/harfbuzz/harfbuzz/blob/main/src/OT/glyf/glyf.hh#L84. 
And in order to know if we use short loca, we need to compile each glyph first and then compute the sum of each glyph's (length+padding) size. Thus we can not directly use serializer at first.  Not sure if there's another better way to do this.

behdad:
I see a warning:
```
../src/OT/glyf/SimpleGlyph.hh:241:5: warning: ‘lastflag’ may be used uninitialized [-Wmaybe-uninitialized]
```

behdad:
Performance looks good now.

behdad:
> I see a warning:

I see it's compiler being stupid.

behdad:
> > I see a warning:
> 
> I see it's compiler being stupid.

Just initialize lastflag to 0 and this is good to go. Thanks.

src/OT/glyf/glyf.hh
garretrieger: Right, I was thinking you could create a buffer + a new serialize context in this function then pass that down to all of the encoding/compiling operations to write into. Each glyph could be packed as an object in the serialization resulting in an object id. Then with the object id's you could extract the head and tail positions for each glyph in the buffer.

PR #3760, opened 2022-07-28
[meson] Don’t tamper with paths in amalgam files

khaledhosny:
Using `os.path.basename()` breaks files in subdirectories (`gsubgpos-graph.cc` instead of `graph/gsubgpos-graph.cc`). The relative paths that meson returns look a bit ugly but they still work.

### DIFF
--- a/src/gen-harfbuzzcc.py
+++ b/src/gen-harfbuzzcc.py
@@ -14,7 +14,7 @@ CURRENT_SOURCE_DIR = sys.argv[2]
 sources = sorted(set(sys.argv[3:]))
 
 with open (OUTPUT, "wb") as f:
-	f.write ("".join ('#include "{}"\n'.format (os.path.basename (x)) for x in sources if x.endswith (".cc")).encode ())
+	f.write ("".join ('#include "{}"\n'.format (os.path.relpath (os.path.abspath (x), CURRENT_SOURCE_DIR)) for x in sources if x.endswith (".cc")).encode ())
 
 # copy it also to the source tree, but only if it has changed
 baseline_filename = os.path.join (CURRENT_SOURCE_DIR, os.path.basename (OUTPUT))
--- a/src/gen-harfbuzzcc.py
+++ b/src/gen-harfbuzzcc.py
@@ -2,23 +2,24 @@
 
 "This tool is intended to be used from meson"
 
-import os, sys, shutil
+import shutil
+import sys
+
+from pathlib import Path
 
 if len (sys.argv) < 3:
 	sys.exit (__doc__)
 
-OUTPUT = sys.argv[1]
-CURRENT_SOURCE_DIR = sys.argv[2]
+OUTPUT = Path (sys.argv[1])
+CURRENT_SOURCE_DIR = Path (sys.argv[2])
 
 # make sure input files are unique
-sources = sorted(set(sys.argv[3:]))
+sources = [Path(x) for x in sorted(set(sys.argv[3:]))]
 
 with open (OUTPUT, "wb") as f:
-	f.write ("".join ('#include "{}"\n'.format (os.path.relpath (os.path.abspath (x), CURRENT_SOURCE_DIR)) for x in sources if x.endswith (".cc")).encode ())
+	f.write ("".join (f'#include "{p.resolve ().relative_to (CURRENT_SOURCE_DIR)}"\n' for p in sources if p.suffix == ".cc").encode ())
 
 # copy it also to the source tree, but only if it has changed
-baseline_filename = os.path.join (CURRENT_SOURCE_DIR, os.path.basename (OUTPUT))
-with open(baseline_filename, "rb") as baseline:
-	with open(OUTPUT, "rb") as generated:
-		if baseline.read() != generated.read():
-			shutil.copyfile (OUTPUT, baseline_filename)
+baseline = CURRENT_SOURCE_DIR / OUTPUT.name
+if baseline.read_bytes() != OUTPUT.read_bytes():
+	shutil.copyfile (OUTPUT, baseline)
--- a/src/harfbuzz-subset.cc
+++ b/src/harfbuzz-subset.cc
@@ -1,4 +1,4 @@
-#include "gsubgpos-graph.cc"
+#include "graph/gsubgpos-graph.cc"
 #include "hb-aat-layout.cc"
 #include "hb-aat-map.cc"
 #include "hb-blob.cc"


khaledhosny:
If someone knows a better way to fix this, please help.

behdad:
Can't we chop the extra part?

Also wonder if the Makefile.am path generates the same. Or let's remove that one at least?

behdad:
What's the way to quickly set CXXFLAGS in a meson build? That's the only thing I use the autotools build for these days.

eli-schwartz:
Meson respects the `$CXXFLAGS` environment variable for user-defined flags. :)

behdad:
> Meson respects the `$CXXFLAGS` environment variable for user-defined flags. :)

At meson configure time, right? Noted.

I'm used to quick `make CXXFLAGS=...` as a lazy way. :)

eli-schwartz:
Meson uses ninja, which doesn't support overriding variables after the build file is generated. :P

eli-schwartz (changes requested):
> Using `os.path.basename()` breaks files in subdirectories (`gsubgpos-graph.cc` instead of `graph/gsubgpos-graph.cc`). The relative paths that meson returns look a bit ugly

Meson passes inputs as arguments in such a manner as to permit them to be found by the program, which means they need to be relative to the cwd when `gen-harfbuzzcc.py` is run.

You could add another argument to the script for the base directory, which via `meson.current_source_dir()` will be an absolute path. Use os.path.abspath on each file argument, then relpath compared to the current source dir, then write that path out.

> but they still work.

They definitively do not.

```
meson setup ~/git/harfbuzz /tmp/hb-builddir
ninja -C /tmp/hb-builddir
git -C ~/git/harbuzz diff
```

Want to guess the output of that?

```diff
diff --git a/src/harfbuzz.cc b/src/harfbuzz.cc
index d6089712b..65da0c252 100644
--- a/src/harfbuzz.cc
+++ b/src/harfbuzz.cc
@@ -1,56 +1,56 @@
-#include "../src/hb-aat-layout.cc"
-#include "../src/hb-aat-map.cc"
-#include "../src/hb-blob.cc"
-#include "../src/hb-buffer-serialize.cc"
-#include "../src/hb-buffer-verify.cc"
-#include "../src/hb-buffer.cc"
-#include "../src/hb-common.cc"
-#include "../src/hb-coretext.cc"
-#include "../src/hb-directwrite.cc"
-#include "../src/hb-draw.cc"
-#include "../src/hb-face.cc"
-#include "../src/hb-fallback-shape.cc"
-#include "../src/hb-font.cc"
-#include "../src/hb-ft.cc"
-#include "../src/hb-gdi.cc"
-#include "../src/hb-glib.cc"
-#include "../src/hb-graphite2.cc"
-#include "../src/hb-map.cc"
-#include "../src/hb-number.cc"
-#include "../src/hb-ot-cff1-table.cc"
-#include "../src/hb-ot-cff2-table.cc"
-#include "../src/hb-ot-color.cc"
-#include "../src/hb-ot-face.cc"
-#include "../src/hb-ot-font.cc"
-#include "../src/hb-ot-layout.cc"
-#include "../src/hb-ot-map.cc"
-#include "../src/hb-ot-math.cc"
-#include "../src/hb-ot-meta.cc"
-#include "../src/hb-ot-metrics.cc"
-#include "../src/hb-ot-name.cc"
-#include "../src/hb-ot-shape-fallback.cc"
-#include "../src/hb-ot-shape-normalize.cc"
-#include "../src/hb-ot-shape.cc"
-#include "../src/hb-ot-shaper-arabic.cc"
-#include "../src/hb-ot-shaper-default.cc"
-#include "../src/hb-ot-shaper-hangul.cc"
-#include "../src/hb-ot-shaper-hebrew.cc"
-#include "../src/hb-ot-shaper-indic-table.cc"
-#include "../src/hb-ot-shaper-indic.cc"
-#include "../src/hb-ot-shaper-khmer.cc"
-#include "../src/hb-ot-shaper-myanmar.cc"
-#include "../src/hb-ot-shaper-syllabic.cc"
-#include "../src/hb-ot-shaper-thai.cc"
-#include "../src/hb-ot-shaper-use.cc"
-#include "../src/hb-ot-shaper-vowel-constraints.cc"
-#include "../src/hb-ot-tag.cc"
-#include "../src/hb-ot-var.cc"
-#include "../src/hb-set.cc"
-#include "../src/hb-shape-plan.cc"
-#include "../src/hb-shape.cc"
-#include "../src/hb-shaper.cc"
-#include "../src/hb-static.cc"
-#include "../src/hb-style.cc"
-#include "../src/hb-ucd.cc"
-#include "../src/hb-unicode.cc"
-#include "../src/hb-uniscribe.cc"
\ No newline at end of file
+#include "../../home/eschwartz/git/harfbuzz/src/hb-aat-layout.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-aat-map.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-blob.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-buffer-serialize.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-buffer-verify.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-buffer.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-common.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-coretext.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-directwrite.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-draw.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-face.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-fallback-shape.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-font.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ft.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-gdi.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-glib.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-graphite2.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-map.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-number.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-cff1-table.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-cff2-table.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-color.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-face.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-font.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-layout.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-map.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-math.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-meta.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-metrics.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-name.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-shape-fallback.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-shape-normalize.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-shape.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-shaper-arabic.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-shaper-default.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-shaper-hangul.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-shaper-hebrew.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-shaper-indic-table.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-shaper-indic.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-shaper-khmer.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-shaper-myanmar.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-shaper-syllabic.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-shaper-thai.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-shaper-use.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-shaper-vowel-constraints.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-tag.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ot-var.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-set.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-shape-plan.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-shape.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-shaper.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-static.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-style.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-ucd.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-unicode.cc"
+#include "../../home/eschwartz/git/harfbuzz/src/hb-uniscribe.cc"
\ No newline at end of file
```

khaledhosny:
> > Using `os.path.basename()` breaks files in subdirectories (`gsubgpos-graph.cc` instead of `graph/gsubgpos-graph.cc`). The relative paths that meson returns look a bit ugly
> 
> Meson passes inputs as arguments in such a manner as to permit them to be found by the program, which means they need to be relative to the cwd when `gen-harfbuzzcc.py` is run.
> 
> You could add another argument to the script for the base directory, which via `meson.current_source_dir()` will be an absolute path. Use os.path.abspath on each file argument, then relpath compared to the current source dir, then write that path out.

I tried something like this first but couldn’t get it to work. I think using os.path.abspath was the missing peice.

> > but they still work.
> 
> They definitively do not.

I’m ony concerned about the file we ship.

khaledhosny:
> Can't we chop the extra part?

I did that now based on @eli-schwartz suggestion.

> Also wonder if the Makefile.am path generates the same. Or let's remove that one at least?

The Makefile.am is not doing anything with the paths, so it should be fine. The meson issue is because it always gives as relative paths not the plain file names in the meson.build file.

linsui:
Hi, it seems that Python 3.6 is dropped in this MR since f-string is introduced in Python 3.7. Is this intended? Is there any related docs?

khaledhosny:
Python 3.6 is EOL already, but if this is causing issues I can revert the f-string use.

linsui:
F-Droid still builds apps on Debian stretch so Python 3.5 is used by default. Not a big problem since we can build new versions of Python and we are going to upgrade it soon. It may be useful though if you can support python 3.5 or add a note somewhere mentioning the Python version requirement.

khaledhosny:
We don’t have an explicit policy for supported Python versions, but practically it should be at least the minimum version supported by the minimum meson version we require. The autotools build does not use this script.

linsui:
Thanks for your help!

Issue #3721, opened 2022-07-13
Glyph flag for kashida insertion


khaledhosny:
Kashida insertion as it is currently implemented in many software is incompatible with fonts that make heavy use of contextual alternates (e.g. Amiri), since kashida insertion is done after shaping and applications have no way to know if inserting a kashida after two glyphs will break their connection or not.

I considered checking for `HB_GLYPH_FLAG_UNSAFE_TO_BREAK`, but it is not usable for this as it is set between any two connected Arabic letters.

I’m wondering if it is possible to have a flag similar to `HB_GLYPH_FLAG_UNSAFE_TO_BREAK` but it would be set only if there is a substitution or positioning other than the `init`/`medi`/`fina` features?

For example in Amiri, in the following sequence the first and last glyphs are the same regardless of the presence of kashida, so it would be safe to insert it post shaping:
```
$ hb-shape Amiri-Regular.ttf "به بـه" --show-flags --no-positions --no-clusters
[heh-ar.fina#3|kashida-ar.1#3|beh-ar.init|space|heh-ar.fina#3|beh-ar.init]
```

But not in this sequence (since there is a contextual substitution involved):
```
$ hb-shape Amiri-Regular.ttf "له لـه" --show-flags --no-positions --no-clusters
[heh-ar.fina#3|kashida-ar.1#3|lam-ar.init|space|heh-ar.fina.LamHeh#3|lam-ar.init.LamHeh]
```

Both sequences currently get the same glyph flags. It is just an idea though, and might not be that useful since the only HarfBuzz client that does any kashida insertion at all is LibreOffice (it should be able to make use of such a glyph flag, though).

khaledhosny:
I was thinking of something like this:
```
diff --git a/src/hb-ot-shaper-arabic.cc b/src/hb-ot-shaper-arabic.cc
index ea1d49daa..0261039ec 100644
--- a/src/hb-ot-shaper-arabic.cc
+++ b/src/hb-ot-shaper-arabic.cc
@@ -328,7 +328,7 @@ arabic_joining (hb_buffer_t *buffer)
     if (entry->prev_action != NONE && prev != UINT_MAX)
     {
       info[prev].arabic_shaping_action() = entry->prev_action;
-      buffer->unsafe_to_break (prev, i + 1);
+      //buffer->unsafe_to_break (prev, i + 1);
     }
     else
     {
@@ -362,7 +362,7 @@ arabic_joining (hb_buffer_t *buffer)
[...]

PR #3762, opened 2022-07-29
Prototype glyph flag safe-to-kashida

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3721


behdad:
Seems to work.

behdad:
Tests are failing badly. No idea why.

behdad:
Oh, I know why.

khaledhosny:
Excellent. Let me try to use this in LibreOffice and report back.

behdad:
@jfkthame is this something you are possibly interested in?

khaledhosny:
Initial testing seems to work fine.

It seems I need to check the flag on the left-side glyph of the kashida insertion point (I was initially checking the right side glyph).

image

khaledhosny:
Before 
Screen Shot 2022-07-30 at 1 37 14 PM

khaledhosny (approved):

behdad:
> It seems I need to check the flag on the left-side glyph of the kashida insertion point (I was initially checking the right side glyph).

It gives you the flag where the kashida can be inserted, right?

khaledhosny:
> > It seems I need to check the flag on the left-side glyph of the kashida insertion point (I was initially checking the right side glyph).
> 
> It gives you the flag where the kashida can be inserted, right?

Yes. I was checking the other side of the insertion point. e.g. if a kashida is to be inserted in the middle of سر,  I have to check the flag on the ر glyph to know if it is safe or not, but I was checking the flag on the س glyph.

behdad:
Okay. I'll probably clean this up this way:
- [ ] Rename to `SAFE_TO_INSERT_KASHIDA`
- [ ] Documentation

Humm. That's about it I think.

behdad:
The only issue with adding more glyph flags is that we are eating from the font-feature glyph mask bits... :(  For that reason, it's better to use a HB_BUFFER_PRODUCE_SAFE_TO_INSERT_KASHIDA flag as well, so we can make it not consume the bit at least. We currently don't do that for @HB_BUFFER_FLAG_PRODUCE_UNSAFE_TO_CONCAT, but we can. Since most clients would have no use for this flag...

On the other hand, since the position of these flags is fixed in the glyph mask, it's only the highest one the client request that matters. I still think we should add a FLAG_PRODUCE_... for this.

behdad:
I think this is ready.

khaledhosny:
cc @jbezos as I think the LaTeX babel package is doing some kind of Arabic kashida justification and might benefit from this flag.

PeterCon:
What is the flag that's being used to control this?

behdad:
> What is the flag that's being used to control this?

The user communicates their need for this flag via the buffer flag `HB_BUFFER_FLAG_PRODUCE_SAFE_TO_INSERT_TATWEEL`. In which case the output per-cluster flag `~HB_GLYPH_FLAG_SAFE_TO_INSERT_TATWEEL` is produced.

behdad:
> What is the flag that's being used to control this?

Oh if you mean where in the font, then there's no flag. The result is automatically produced by disabling insertion opportunities where cursive-attachment happens.

khaledhosny:
It is basically the reverse of `HB_GLYPH_FLAG_UNSAFE_TO_BREAK` but additionally it is not set for substitutions performed during basic Arabic shaping.

PeterConstable:
>The result is automatically produced by disabling insertion opportunities where cursive-attachment happens.

Does that mean that, if the connection between two glyphs is horizontal and amenable to insertion of a tatweel but the font also happened to have an anchor positioning lookup for that pair of glyphs (probably not needed, but suppose), then the insertion would be blocked?

khaledhosny:
Yes. We wouldn’t know if the connection is straight or not and whether the anchor is superfluous or not.

jbezos:
@khaledhosny I’ve just realized you mentioned me. Yes, there is some work in `babel` to deal with kashida in `luatex`. How can I take advantage of this feature?

khaledhosny:
> @khaledhosny I’ve just realized you mentioned me. Yes, there is some work in `babel` to deal with kashida in `luatex`. How can I take advantage of this feature?

It would solve the issue of needing per-font kashida exception list. For example, in Amiri لا is two glyphs so kashida insertion algorithm might want to insert a kashida between them (which is wrong), also سحـ use an alternate س and ح but code might want to insert a kashida between them (which also wrong). The flag helps with that. e.g.:
```
$ ./util/hb-shape ../../../aliftype/amiri/fonts/Amiri-Regular.ttf 'لا سحاب' --no-clusters --no-positions --show-flags --safe-to-insert-tatweel
[beh-ar|alef-ar.fina#5|hah-ar.medi.SadHaaInit#1|seen-ar.init.SenHaaInit|space|alef-ar.fina.LamAlfIsol#1|lam-ar.init.LamAlfIsol]
```

(getting this flag requires setting a buffer flag during shaping, so support at luaotfload side is also needed).

Here the flag is set only on the final ا so that is the only place in this string where a kashida can be inserted (before the ا). The flag does not determine kashida insertion opportunities, babel code will need to that then check this flag to filter-out places where inserting a kashida will break the shape of the text.

An extreme example is fonts like Noto Nastaliq or Aref Ruqaa where virtually the flag will never be set meaning there is no where a kashida can be inserted without breaking the text shape.

jbezos:
@khaledhosny After investigating a little bit, I think the current code can be adapted, because the basic idea is similar. I’ve managed to find the glyph names in the `luahbtex` font subtable `resources.unicodes` (with the default renderer), but they don’t quite match those shown by `hb-shape` in your example (they have the form `uni0684.medi_AynHaaInit`).

However, from your answer I infer the exception list can’t be currently obtained and that `luaotfload` has to be extended. I’m right?

jbezos:
@khaledhosny To be more precise, what I need is:

- The exception list
- A table mapping from the internal glyph code to the glyph name (or the other way around).

jbezos:
Then `luaotfload` must be extended, and therefore from `babel` very little can be done currently. The justification algorithm is already done (with programmable rules, although it must be fine tuned), and it works when the fonts just use the basic contextual forms.

PR #3763, opened 2022-07-29
[repacker] Add ability for repacker to pre split PairPosFormat1 subtables.

garretrieger:
When recalculate extensions is enabled the repacker will analyze all PairPosFormat1 subtables and split them into more subtables if needed to guarantee that each subtable is <= than 64kb total size. This guarantees successful packing of all PairPosFormat1 subtables.


src/graph/pairpos-graph.hh
```diff
@@ -0,0 +1,110 @@
...
+    unsigned format = u.format;
+    printf("PairPos::format = %u\n", format);
+    switch (u.format) {
+    case 1:
+      return ((PairPosFormat1*)(&u.format1))->split_subtables (c, this_index);
+    case 2:
+      return ((PairPosFormat2*)(&u.format2))->split_subtables (c, this_index);
+#ifndef HB_NO_BORING_EXPANSION
+    case 3:
+    case 4:
```
behdad, line 269: Add `HB_FALLTHROUGH` in these cases.

behdad:
LGTM in that it doesn't touch non-repacker code. :) Added one comment re FALLTHROUGH.

behdad:
CIFuzz failure is my fault. Though I can't understand how it can fail.
https://github.com/harfbuzz/harfbuzz/commit/28c02b37684b41bf584ed81d910a448bdf9f706a

behdad:
> CIFuzz failure is my fault. Though I can't understand how it can fail.
> [28c02b3](https://github.com/harfbuzz/harfbuzz/commit/28c02b37684b41bf584ed81d910a448bdf9f706a)

Oh, I see. Memory failure...

PR #3764, opened 2022-07-29
[repacker] Add a second repacker api method which takes a table tag.

garretrieger:
This is needed to allow table specific optimizations to be performed during repacking.

### DIFF
--- a/src/hb-subset-repacker.cc
+++ b/src/hb-subset-repacker.cc
@@ -25,18 +25,24 @@
 #include "hb-repacker.hh"
 
 #ifdef HB_EXPERIMENTAL_API
+
 /**
  * hb_subset_repack_or_fail:
+ * @table_tag: tag of the table being packed, needed to allow table specific optimizations.
  * @hb_objects: raw array of struct hb_object_t, which provides
  * object graph info
  * @num_hb_objs: number of hb_object_t in the hb_objects array.
  *
  * Given the input object graph info, repack a table to eliminate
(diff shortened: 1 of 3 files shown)


behdad:
This is C api so we can't have duplicate names. I suggest you just renamed the existing one since this is experimental API. Or deprecate the old one and come up with a new name.

khaledhosny:
Changing the existing API should be fine since this is hidden behind a build flag.

garretrieger:
> Changing the existing API should be fine since this is hidden behind a build flag.

Ok, I'll just remove the old one, not really any reason you'd want to use it.

garretrieger:
> > Changing the existing API should be fine since this is hidden behind a build flag.
> 
> Ok, I'll just remove the old one, not really any reason you'd want to use it.

Alright fixed, ready for review.

behdad:
Repacker tests need updating.

garretrieger:
> Repacker tests need updating.

Fixed.

behdad:
Going  to merge this if @khaledhosny is okay fixing the uharfbuzz side.

garretrieger:
> Going to merge this if @khaledhosny is okay fixing the uharfbuzz side.

I have a branch ready to fix the uharfbuzz side: https://github.com/harfbuzz/uharfbuzz/compare/tag_in_repack?expand=1

behdad:
Go ahead as you see fit.

PR #3777, opened 2022-08-04
bug fix in VarData get_delta ()

qxliu76:
I think this is a bug that needs a fix when LONG_WORDS flag is set. Correct me if I'm wrong cause I tried to understand the code according to the spec.

Given below variables defined in the function:
lcount: num of INT 32 deltas
scount: num of INT16 deltas
count: total num of deltas

when LONG_WORDS flag is set, this get_delta() was not working correctly
- the "row" is not computed correctly
- lcursor should be interpreted as INT32

I don't have a testcase for this, but existing tests don't seem to catch this.

### DIFF
--- a/src/hb-ot-layout-common.hh
+++ b/src/hb-ot-layout-common.hh
@@ -2284,19 +2284,19 @@ struct VarData
    unsigned int lcount = is_long ? word_count : 0;
 
    const HBUINT8 *bytes = get_delta_bytes ();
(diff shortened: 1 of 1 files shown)


behdad:
Good catch. Thanks.

behdad:
I made https://github.com/harfbuzz/harfbuzz/commit/c67c1b745bb8306aa6a948f0337a98dc010042a5

Issue #2650, opened 2020-08-21
Chinese Phonetic ("ZHP ") not converted to BCP 47?


RoelN:
I'm looking at what [gen-tag-table.py](https://github.com/harfbuzz/harfbuzz/blob/master/src/gen-tag-table.py) does, and noticed Chinese Phonetic ("Pinyin romanization") might be dropped from the LangSys-tag-to-BCP-47 conversion.

Around [this line](https://github.com/harfbuzz/harfbuzz/blob/master/src/gen-tag-table.py#L648) the language list is tweaked and amended, mostly by adding or substituting BCP 47 codes to LangSys tags. On [line 747](https://github.com/harfbuzz/harfbuzz/blob/master/src/gen-tag-table.py#L757) Chinese Phonetic ("ZHP ") is removed, but nothing is returned in place. If I understand [hb_ot_tag_to_language](https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ot-tag.cc#L434) correctly, this means a "x-hbot-AABBCCDD" tag will be returned for "ZHP ", while I think it should be "zh-Latn" according to the [IANA](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry).

This was one of my test tags while researching for Wakamai Fondue, so I might be wrong as I only just begun grasping the extent of languages/scripts/fonts 😅

behdad:
> It would be better if HarfBuzz could convert zh to 'ZHO ' or 'ZHP ' based on the buffer’s script, without requiring an explicit script subtag.

Which we now can because our new tag-choosing logic takes both script and language, right?

dscorbett:
`hb_ot_tags_from_language` only takes the language. You can override the script in the language tag using `x-hbsc`, but the choice of language tag does not depend at all on the buffer’s script. I did consider adding this feature, but it would complicate gen-tag-table.py to support a single language system tag that probably no one uses, so I didn’t. That’s not to say we shouldn’t do it someday.

RoelN:
For completeness sake, [here](https://github.com/Wakamai-Fondue/wakamai-fondue-engine/blob/master/tools/lang-tags-from-hb.py) is the script I use to convert HarfBuzz' `hb-ot-tag-table.hh` to a [JSON list](https://github.com/Wakamai-Fondue/wakamai-fondue-engine/blob/master/src/tools/ot-to-html-lang.js). I hope you don't mind me using HarfBuzz data like this!

behdad:
[...]

PR #3783, opened 2022-08-07
[ot-tags] Support the language system tag 'ZHP '

dscorbett:
This is the practical part of the solution for #2650. 'ZHP ' now corresponds to the three romanizations of Chinese that have BCP 47 variant subtags: Pinyin, Tongyong Pinyin, and Wade–Giles. The full solution would be to recognize any romanization of Chinese, even without an explicitly tagged variant, based on the script of the buffer, but that would be too big an architectural change for the benefit of a single language system tag whose original intent no one remembers and which almost no fonts use.


behdad:
LGTM if you are happy with it.

behdad:
Honestly I'm not sure. It adds quite a few cases. Happy to close it as wontfix.

dscorbett:
That would be fine. The implementation will still be here if we change our minds.

PR #3789, opened 2022-08-18
[subset] add destructor to subset plan so it's members destruct before the plan object.

garretrieger:
Now that object destroy calls the destructor of plan, that can cause problems when destructing its members afterwards.

### DIFF
--- a/src/hb-subset-plan.cc
+++ b/src/hb-subset-plan.cc
@@ -620,7 +620,7 @@ _normalize_axes_location (hb_face_t *face,
     }
     if (has_avar)
       seg_maps = &StructAfter<OT::SegmentMaps> (*seg_maps);
-    
+
     axis_count++;
   }
   all_axes_pinned = !axis_not_pinned;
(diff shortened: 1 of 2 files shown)


behdad:
No, this is wrong. The `hb_object_destroy` also takes care of reference-counting.

The correct fix is to move the destructions to plan's destructor.

behdad:
The existing code should work. How did this become an issue? It's a pattern we use in other places in the code.

behdad:
I like to commit this change:
```diff
diff --git a/src/hb-object.hh b/src/hb-object.hh
index 8cbf7a8a9..bc37cc5ae 100644
--- a/src/hb-object.hh
+++ b/src/hb-object.hh
@@ -273,7 +273,8 @@ static inline bool hb_object_destroy (Type *obj)
 
   hb_object_fini (obj);
 
-  obj->~Type ();
+  if (!std::is_trivially_destructible<Type>::value)
+    obj->~Type ();
 
   return true;
 }
```
such that we're technically not distructing POD structs. But that wouldn't help the `hb_subset_plan_t` case because it has a bare vector in it.

behdad:
Thanks. At some point would be nice to move initialization to the constructor as well.

Issue #3458, opened 2022-02-19
Expose optical bounds


matthiasclasen:
As discussed in San Miguel, this information needs to be applied outside of shaping individual runs (when post-processing a line), so pango would need a getter for the boundary adjustments for a glyph. Something like

    hb_bool_t hb_ot_layout_get_optical_bounds (hb_font_t *font, 
                                               hb_direction_t direction,
                                               hb_codepoint_t glyph,
                                               hb_position_t *left_bound,
                                               hb_position_t *right_bound);

behdad:
See https://docs.microsoft.com/en-us/typography/opentype/spec/features_ko#opbd

behdad:
What's direction here for then?

behdad:
A much nicer API would read:
```
hb_position_t hb_ot_layout_get_optical_bounds (hb_font_t *font,
                                               hb_direction_t direction,
                                               hb_codepoint_t glyph);
```

behdad:
But this being `ot-layout`, we need script/language to find the respective feature. Ugh.

behdad:
> But this being `ot-layout`, we need script/language to find the respective feature. Ugh.

The feature registry says "Script/language sensitivity: None.". So it's one of those that we just need to find any feature with the tag and run with it. We just need to cache the feature in the accelerator...

matthiasclasen:
> What's direction here for then?

I was thinking for deciding between left/right and top/bottom

behdad:
> I was thinking for deciding between left/right and top/bottom

Right. But since hb_direction_t has four values, the separate right/left is redundant.

Also, OpenType currently has no features for top/bottom.

Moreover, OpenType is unclear how such feature should be implemented.

matthiasclasen:
My understanding of how this would be implemented is that pango would shape runs until it has a line assembled, then it asks hb for the left margin for the first char and the right margin of the last char, and adjusts the line width and positioning by those values, before justification (and it would only do that for justified lines...).

I have some renewed interest in this since I found that the linux libertine actually has the lfbd and rtbd features.

behdad:
> first char and the right margin of the last char

Glyphs.



> and it would only do that for justified lines

One side should be done for non-justified lines too I would think.

I'll look into this again.

n8willis:
Would you also need to check that the first/last glyph isn't a non-spacing mark?

behdad:
> Would you also need to check that the first/last glyph isn't a non-spacing mark?

[...]

PR #3799, opened 2022-08-31
Optical bounds

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3458

Untested.


src/hb-ot-layout.cc
```diff
@@ -1703,6 +1703,63 @@ hb_ot_layout_get_size_params (hb_face_t       *face,
...
+  dispatch (const T &obj, Ts&&... ds) HB_AUTO_RETURN
+  ( _dispatch (obj, hb_prioritize, std::forward<Ts> (ds)...) )
+};
+
+hb_position_t
+hb_ot_layout_get_optical_bound (hb_font_t      *font,
+				unsigned        lookup_index,
+				hb_direction_t  direction,
+				hb_codepoint_t  glyph)
+{
```
matthiasclasen, line 1731: My immediate questions, trying to use this api:
- How do I interpret the value that is returned ? Is it 'how far the glyph should extend into the margin' or something else?
- Where do I get this lookup_index you want?
- Can we add some boolean for 'has optical bound information' ? Or maybe that would be answered by trying to find that lookup_index?

behdad: > * How do I interpret the value that is returned ? Is it 'how far the glyph should extend into the margin' or something else?

Right. Though, it's *negative* for extending into the margin. That's just how the numbers in the font were and I kept them that way. So basically that's how much you add to the x_offset and advance_width, for the left side, and you add to advance_width, for the right side.

> * Where do I get this lookup_index you want?

Ultimately from `hb_ot_layout_feature_get_lookups`. You need to get feature index first.

> * Can we add some boolean for 'has optical bound information' ? Or maybe that would be answered by trying to find that lookup_index?

Yeah, if there was no lookup_index you know...  Or even earlier, there would be no feature_index.

matthiasclasen: I've got lucky with hb_ot_layout_collect_lookups

src/hb-ot-layout.cc
```diff
@@ -1703,6 +1703,63 @@ hb_ot_layout_get_size_params (hb_face_t       *face,
...
+  /* This is ugly... */
+  hb_buffer_t buffer;
+  buffer.props.direction = direction;
+  OT::hb_ot_apply_context_t c_apply (1, font, &buffer);
+  const OT::PosLookup &lookup = font->face->table.GPOS->table->get_lookup (lookup_index);
+  hb_glyph_position_t pos = {0};
+  hb_position_single_dispatch_t c;
+  lookup.dispatch (&c, &c_apply, glyph, pos);
+  hb_position_t ret = 0;
+  switch (direction)
```
matthiasclasen, line 1737: Do you think this is a misuse of direction? What we really want here is a side (ie left/right/top/bottom), not a direction. 
The right bounds will be used in ltr context too...

behdad: I know... But another enum with exactly four values sounds a bit overkill...

behdad: Think of it as "left-to-right edge of the glyph".

matthiasclasen: but then, there are no features for top/bottom bounds anyway

behdad: In theory there could; just needs a feature tag (which could be private) to be named.

matthiasclasen:
I can confirm that this api works, tested with this pango code: https://gitlab.gnome.org/GNOME/pango/-/commit/3fedb89e7b2301d4de0f3885bd5e2ed9cfe38105
and Linux Libertine fonts.

src/hb-ot-layout.cc
matthiasclasen: We should probably turn this discussion into a doc comment

behdad:
Thanks. I'll add docs, double-check, and merge.

src/hb-ot-layout.cc
matthiasclasen: Theoretical question: what would I do if ltbd returned more than one lookup ? try all until I find the glyph ?

behdad: > Theoretical question: what would I do if ltbd returned more than one lookup ? try all until I find the glyph ?

Correct.

behdad: > Theoretical question: what would I do if ltbd returned more than one lookup ? try all until I find the glyph ?

The loop you have ignores current script/lang. You might want to adjust it but is more work of course. You might as well keep it and loop over the lookups.

src/hb-ot-layout.cc
```diff
@@ -1703,6 +1703,59 @@ hb_ot_layout_get_size_params (hb_face_t       *face,
...
+  template <typename T, typename ...Ts> auto
+  dispatch (const T &obj, Ts&&... ds) HB_AUTO_RETURN
+  ( _dispatch (obj, hb_prioritize, std::forward<Ts> (ds)...) )
+};
+
+hb_position_t
+hb_ot_layout_get_optical_bound (hb_font_t      *font,
+				unsigned        lookup_index,
+				hb_direction_t  direction,
+				hb_codepoint_t  glyph)
```
khaledhosny, line 1730: Applications like LuaTeX want all the optical bound values in advance (it has a mode where line breaking takes it into account). I guess one can use `hb_ot_layout_lookup_collect_glyphs()` to get all affected glyphs and iterate on them one by one, so would it be faster if the API took an array of glyphs (or to have a different API that return optical bounds for all glyphs in the lookup?)

behdad: Humm. That complicates things. Yeah I'm leaning towards suggesting collect_glyphs first.

behdad: > Applications like LuaTeX want all the optical bound values in advance

We can add a version that works on an array later if that fetching one by one seems to be too slow.

PR #3800, opened 2022-08-31
docs: Improve docs for some apis

matthiasclasen:
Various functions take hb_tag_t arrays. Those
arrays are expected to be 0-terminated. Document
that.


behdad:
Care to mention `HB_TAG_NONE` instead.

matthiasclasen:
Sure. Should also mention that these are nullable, at the same time

behdad:
> Should also mention that these are nullable, at the same time

Yes.

PR #3801, opened 2022-08-31
[instance] add benchmarks for instancing

qxliu76:
Added instance operation which does both subsetting and instancing.
Currently it's wrapped by experimental_api.
This PR is dependent on #3795 which contains a fix for a crash in the benchmark runs.


garretrieger (approved):

behdad:
This commit added over 6MB of font files to the repo / distribution. I'm not sure if it's a good idea.

PR #3807, opened 2022-09-03
hb-view: Detect WezTerm and use iterm2 image protocol

wez:
Similar to the logic that detects iterm2, but look for
TERM_PROGRAM=WezTerm which identifies wezterm is present.

This allows hb-view to output an image directly to the terminal.

### DIFF
--- a/util/helper-cairo.hh
+++ b/util/helper-cairo.hh
@@ -410,7 +410,7 @@ helper_cairo_create_context (double w, double h,
 	extension = "png";
 	protocol = image_protocol_t::ITERM2;
       }
-      if ((name = getenv ("TERM_PROGRAM")) != nullptr &&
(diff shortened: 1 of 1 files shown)


util/helper-cairo.hh
```diff
@@ -410,6 +410,12 @@ helper_cairo_create_context (double w, double h,
 	extension = "png";
 	protocol = image_protocol_t::ITERM2;
       }
+      if ((name = getenv ("TERM_PROGRAM")) != nullptr &&
```
khaledhosny, line 413: Shouldn’t this be `else if`?

wez: doh, sorry, yeah, fixed!

khaledhosny:
Thanks!

PR #3810, opened 2022-09-06
[cmake] Generate pkg-config .pc files

### LINKED ISSUE #3809: Installing harfbuzz with cmake does not install any pkg-config pc files
Hello!

I'm trying to package freetype along with harfbuzz, and I'm using cmake on linux (which is a centos7 docker image running under Ubuntu 22.04 WSL, which in turn is running under Windows 11; this info is probably not relevant)

Before the install, freetype (without harfbuzz) is already installed, and none of the other dependencies are installed.

I'm running
```
mkdir build && cd build

cmake .. -DCMAKE_INSTALL_PREFIX=/usr/local/ \
  -DCMAKE_INSTALL_LIBDIR=lib \
  -DCMAKE_BUILD_TYPE=Release \
  -DBUILD_SHARED_LIBS=true \
  -DHB_HAVE_FREETYPE=true 
make
make install
```

The install runs without any fail, and the second pass of freetype compilation (with harfbuzz support) also runs: but I don't have a `harfbuzz.pc` in `/usr/local/lib/pkgconfig` as I'd expect.

This means I can't call `pkg-config` with `freetype2` (which errors)
```
sh-4.2# pkg-config --cflags freetype2
Package harfbuzz was not found in the pkg-config search path.
Perhaps you should add the directory containing `harfbuzz.pc'
to the PKG_CONFIG_PATH environment variable
Package 'harfbuzz', required by 'FreeType 2', not found
```

I did a `find` for `harfbuzz.pc` and it's not installed on any directory on the container.
I know that the meson build system is recommended by harfbuzz now, but I thought this bug was still worth reporting

ankith26:
fixes #3809

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.12)
 project(harfbuzz)
 
 message(WARN "HarfBuzz has a Meson port and tries to migrate all the other build systems to it, please consider using it as we might remove our cmake port soon.")
@@ -107,12 +107,17 @@ if (${HAVE_STDBOOL_H})
   add_definitions(-DHAVE_STDBOOL_H)
 endif ()
 
+# These will be used while making pkg-config .pc files
+set(PC_REQUIRES_PRIV "")
+set(PC_LIBS_PRIV "")
+
 if (NOT MSVC)
   set(THREADS_PREFER_PTHREAD_FLAG ON)
   find_package(Threads)
   if (CMAKE_USE_PTHREADS_INIT)
     add_definitions("-DHAVE_PTHREAD")
     list(APPEND THIRD_PARTY_LIBS Threads::Threads)
+    list(APPEND PC_LIBS_PRIV -pthread)
   endif ()
 endif ()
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.12)
```
alerque, line 1: Is this bump *necessary* for the new functionality or just routine or because it's what you happened to test with?

ankith26: I'm using the latest CMake version (at the time of writing of this comment, 3.24.1)

Cmake 3.1 is rather old, and I suspect there are already many things in this file relying on 3.2+ features. I just bumped it up to 3.12 because I'm adding the usage of the list `REMOVE_DUPLICATES` feature (that was added in CMake 3.12)

But I could be wrong too, I've not tested this with 3.12, maybe this number would need further bumping

alerque: Thatks okay you you're adding usage of a feature you know what added in 3.12 then the bump makes sense. I just wanted to make sure it was *required*. Perhaps one would need an even newer version, but since you know this at *least* is the minimum then we're good. Thanks for answering.

PR #3811, opened 2022-09-07
meson: Fix freetype and icu dependency lookup

xclaesse:
It is wrong to search for a different name depending on the compiler. If anything, cmake name could be available on systems that uses GCC too. Meson >=0.60 can lookup multiple names at once and fallback to a subproject only if none of them are found on the system.


xclaesse:
CC @fanc999

tp-m:
Thanks @xclaesse 

This is needed to un-break meson subproject builds with recent harfbuzz versions in projects such as GStreamer/cairo/gtk fwiw.

fanc999:
> CC @fanc999

Hi,

Sorry, I was not aware of the subproject fallback items at that time.  Seems ok to me.

tp-m:
Thanks behdad!

PR #3816, opened 2022-09-15
Instantiate feature variations/featurelist/fvar 

qxliu76:
- For featurevariations, I added 2 more hash_maps in the plan:
1. retained feature variation record idx->set of retained condition idxes for this record
2. feature index->pointer to the variation for this feature

- For fvar, an axis old index->new index map is added
- tests generation script is updated because for harfbuzz instancing, we collect lookups after feature variations are applied, and then do glyph closure,etc. This seems generate a smaller glyph set compared to collecting lookups/glyph closure first. So the test script is updated to do instancing first and then subsetting with fonttools.

Still all changes support full instancing only.


src/hb-ot-layout-common.hh
```diff
@@ -3002,24 +3254,42 @@ struct FeatureVariations
     return (this+record.substitutions).find_substitute (feature_index);
   }
 
+  void collect_feature_substitutes_with_variations (hb_collect_feature_substitutes_with_var_context_t *c) const
```
garretrieger, line 3257: FYI the behaviour in fontTools maybe changing in the near future: https://github.com/fonttools/fonttools/issues/2737

Don't need to do anything here yet, but we may need to update this in the future.

src/hb-ot-layout-common.hh
```diff
@@ -2778,15 +2907,104 @@ struct ConditionSet
...
+    }
+
+    // all conditions met
+    if (num_kept_cond == 0) return 1;
+   
+    if (c->num_cond_to_record_map->has (num_kept_cond))
+    {
+      //check if condition_set is unique with variations
+      hb_set_t *record_set = c->num_cond_to_record_map->get (num_kept_cond);
+      for (unsigned record_idx : *record_set)
```
garretrieger, line 2950: This is potentially O(N^2) here's an alternate algorithm that should be  O(N) instead:
- Add a hash() and operator== method to ConditionSet
- Then you should be able to have a hb_hashmap<ConditionSet*, int> (note: the value isn't used so can be whatever type).
- In each call to keep_with_variations call has() on the map with a pointer to the current condition set.  Internally map will dereference the condition set pointer so if there's already an equivalent condition set it will return true. If true drop this condition set.
- If keeping the current set add it to the map.

See how "packed_map" in hb_serialize_context_t works, the idea is basically the same.

garretrieger (approved):

PR #3821, opened 2022-09-20
Map 'rlig' to AAT required-ligatures selector

jfkthame:


drott (approved):
LGTM, big thanks!

behdad:
Ah interesting. Looks like an omission. Thanks.

Issue #3824, opened 2022-09-21
an odd HB_GLYPH_FLAG_UNSAFE_TO_BREAK case


caolanm:
In LibreOffice we're making use of HB_GLYPH_FLAG_UNSAFE_TO_BREAK to cache layout results but have found an odd edge case in one test document (of 250,000) which I've attempted to boil down to the test case here where on verifying that we could reuse the old layout I see different results on shaping one side of the original string which had no HB_GLYPH_FLAG_UNSAFE_TO_BREAK flags.

The output I get is:

full sequence: advance is 1235
full sequence: advance is 460
is unsafe to break set on any glyph? no, (expect no)
shape again without space at edge, expect same results as previous iteration
sub sequence: advance is 1135
sub sequence: advance is 460

which I expect is:

full sequence: advance is 1235
full sequence: advance is 460
is unsafe to break set on any glyph? no, (expect no)
shape again without space at edge, expect same results as previous iteration
sub sequence: advance is 1235
sub sequence: advance is 460

[harfbuzz-demo.zip](https://github.com/harfbuzz/harfbuzz/files/9617359/harfbuzz-demo.zip)

khaledhosny:
I’m not sure I follow what the C code is comparing, but if I try to shape the text with `hb-shape`, I get the same glyph advances:
```
$ hb-shape XBRoya34.ttf -u "20,D8,A7,D9,84,D8,A3" --direction=r --script=arab
[sterling=6+1134|Oslash=5+1588|.notdef=4+1239|Ugrave=3+1474|section=2+1134|Oslash=1+1588|space=0+600]
$ hb-shape XBRoya34.ttf -u "D8,A7,D9,84,D8,A3" --direction=r --script=arab
[sterling=5+1134|Oslash=4+1588|.notdef=3+1239|Ugrave=2+1474|section=1+1134|Oslash=0+1588]
$ hb-shape XBRoya34.ttf -u "20" --direction=r --script=arab
[space=0+600]
```

Note that HarfBuzz advances are relative not absolute position, to measure the full with of a sequence of glyphs, you need to add the advances.

khaledhosny:
Scratch that, I mis-decoded the UTF-8 string.

khaledhosny:
More relevant output (I skipped the morx output for brevity):
```
$ hb-shape XBRoya34.ttf " الأ" --direction=r --script=arab --language=ar --show-flags -V
[...]

PR #3825, opened 2022-09-21
[gpos] in PairPos, always advance next-glyph by one only

behdad:
This goes against the spec.

WIP. Needs comments at least.

https://github.com/harfbuzz/harfbuzz/issues/3824

### DIFF
--- a/src/OT/Layout/GPOS/PairPosFormat2.hh
+++ b/src/OT/Layout/GPOS/PairPosFormat2.hh
@@ -243,8 +243,6 @@ struct PairPosFormat2_4
 
 
     buffer->idx = skippy_iter.idx;
-    if (len2)
-      buffer->idx++;
 
     return_trace (true);
   }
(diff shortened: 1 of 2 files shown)


behdad:
@jfkthame comments?

jfkthame:
While I think this makes much better sense than the current spec behavior, I'm a bit uneasy about diverging from the other main implementations here. Can we try to get some feedback from Microsoft and Apple to see if they would be willing to consider an actual spec change, and updating their engines to match?

If there's no prospect of the spec and other implementations changing, then I'm leaning (reluctantly) towards saying that interoperability is more important than having improved-but-incompatible behavior for this edge case.

behdad:
Thanks Jonthan. I reported against OpenType. Closing this since I now have an alternate fix for the original issue.

behdad:
Reopening, since in my testing, CoreText implements this behavior.

For test font and further discussion see https://github.com/harfbuzz/harfbuzz/issues/3876

PR #3829, opened 2022-09-22
[subset] Allow table ordering on the face builder to be overriden.

garretrieger:


src/hb-face.cc
```diff
@@ -787,3 +802,35 @@ hb_face_builder_add_table (hb_face_t *face, hb_tag_t tag, hb_blob_t *blob)
   hb_blob_destroy (previous);
   return true;
 }
+
+/**
+ * hb_face_builder_sort_tables:
+ * @face: A face object created with hb_face_builder_create()
+ * @tags: Null terminated list of table tags.
```
behdad, line 809: Add the zero-terminated annotation (not the nullable) as well as mentioning HB_TAG_NONE instead -f Null:

https://github.com/harfbuzz/harfbuzz/blob/b70032eddadb5488514d60c54f6254834f3d750d/src/hb-ot-layout.cc#L1189

PR #3833, opened 2022-09-27
[subset] optimize glyf subsetting w/ retain gids.

garretrieger:
When retain gids is enabled the subset plan may require the output of many empty glyphs. This change optimizes the glyf subsetting code when the number of retained glyphs << number of output glyphs. Unnessecary lookups to the glyph map are reduced by iterating through the glyph map instead of the output glyph set.


behdad:
LGTM.

PR #3841, opened 2022-10-13
[glyf] VariableComposites

behdad:
Draft prototype of https://github.com/harfbuzz/boring-expansion-spec/blob/glyf1/glyf1.md


behdad:
This is untested but complete now.

PR #3851, opened 2022-10-20
Make build work for arm-none-eabi

JoelOtter:
I've been working with arm-none-eabi recently and trying to build HarfBuzz for it. I was getting a few errors about mismatches between integer types - this makes it build cleanly.

If we want to ensure this doesn't regress I'd be happy to add a CI job for this, but it might end up being a little overspecific based on the toolchain used.

### DIFF
--- a/.github/workflows/arm-ci.yml
+++ b/.github/workflows/arm-ci.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Configure CMake
         run: |
           cmake -S . -B build \
-            -DCMAKE_CXX_FLAGS="-w -DHB_NO_MT" \
             -DCMAKE_TOOLCHAIN_FILE=${DEVKITPRO}/cmake/3DS.cmake
       - name: Build
-        run: cmake --build build
+        run: make CXXFLAGS="-w -DHB_NO_MT"
+        working-directory: /src/build
--- a/.github/workflows/arm-ci.yml
+++ b/.github/workflows/arm-ci.yml
@@ -21,5 +21,5 @@ jobs:
           cmake -S . -B build \
             -DCMAKE_TOOLCHAIN_FILE=${DEVKITPRO}/cmake/3DS.cmake
       - name: Build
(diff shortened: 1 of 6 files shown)


behdad:
LGTM. Thanks. Curious, what's the size of unsigned on this platform?

behdad:
> If we want to ensure this doesn't regress I'd be happy to add a CI job for this, but it might end up being a little overspecific based on the toolchain used.

If it doesn't randomly break, we're fine having it. Thanks

JoelOtter:
> LGTM. Thanks. Curious, what's the size of unsigned on this platform?

So sizeof(unsigned) is 4, I think, but the issue stems from the fact that on ARM Cortex `int` and `long` are the same. So `unsigned long` is assumed to be 8, but it's still 4. https://developer.arm.com/documentation/dui0472/k/C-and-C---Implementation-Details/Basic-data-types-in-ARM-C-and-C--

JoelOtter:
I'm afraid I don't know enough about MinGW to understand why that build is failing - I see a couple of warnings but nothing that looks fatal?

.github/workflows/arm-ci.yml
```diff
@@ -21,5 +21,5 @@ jobs:
           cmake -S . -B build \
             -DCMAKE_TOOLCHAIN_FILE=${DEVKITPRO}/cmake/3DS.cmake
       - name: Build
-        run: make CXXFLAGS="-w -DHB_NO_MT"
+        run: make CXX_FLAGS="-w -DHB_NO_MT"
```
behdad, line 24: Humm. Isn't `CXXFLAGS` correct?

behdad: Or is cmake's different?

JoelOtter: Yeah this confused me too - CMake templates the Makefile with its own parameters. Weirdly setting these build flags at CMake configure time causes its compiler check to fail, which is why I'm doing it like this and not with `cmake --build`.

Issue #3853, opened 2022-10-21
subset font fails OTS with Bad lookup index


anthrotype:

anthrotype:
Pressed green button too fast, hold on while I write up the bug report..

anthrotype:
Take the font from https://github.com/google/fonts/blob/main/ofl/notosansmarchen/NotoSansMarchen-Regular.ttf

and run hb-subset (current HEAD, --version reports 5.3.1) on it with this input:

```sh
$ hb-subset ofl/notosansmarchen/NotoSansMarchen-Regular.ttf --unicodes=U+11C74,U+11C78,U+11C7A,U+11C7C,U+11C7D,U+11C7E,U+11C81,U+11C85,U+11C89,U+11C8A,U+11C8C,U+11C8D,U+11C8E,U+11C8F,U+11C9A,U+11CA5,U+11CA9,U+11CAA,U+11CAE,U+11CB0,U+11CB1,U+11CB3,U+11CB5,U+0020 --output-file=ofl/notosansmarchen/NotoSansMarchen-Regular.subset.ttf
```

Then run [ots](https://github.com/khaledhosny/ots) on the subset font and see the reported error:


```
ERROR: Layout: Bad lookup index 8 for lookup 0
ERROR: Layout: Failed to parse feature table 1
ERROR: GSUB: Failed to parse feature list table
ERROR: GSUB: Failed to parse table
Failed to sanitize file!
```

The same font subsetted with `fonttools subset` passes OTS without errors.

Original bug https://bugzilla.mozilla.org/show_bug.cgi?id=1796704#c2

garretrieger:
Qunxin tracked this down to https://github.com/harfbuzz/harfbuzz/commit/f7677213d7e2637796eb61df3bce89e163821cf4. In this font there's a single substitution which does something like this: Coverage: {..., 601, 608, ...}, delta: +1

Which gets caught by the degenerate check. However this example isn't degenerate since none of the possible new glyphs intersect with any of the existing glyphs so you won't have a case where glyphs keep being added. We'll likely want to change that check to be more conservative.

behdad:
> Qunxin tracked this down to [f767721](https://github.com/harfbuzz/harfbuzz/commit/f7677213d7e2637796eb61df3bce89e163821cf4). In this font there's a single substitution which does something like this: Coverage: {..., 601, 608, ...}, delta: +1
> 
[...]

PR #3854, opened 2022-10-21
[subset] Always output Lookup's even if they are empty.

garretrieger:
The rest of layout subsetting depends on lookup indices being consistent with those computed during planning. So if an empty lookup is discarded during the subset phase it will invalidate all subsequent lookup indices. Generally we shouldn't end up with an empty lookup as we pre-prune them during the planning phase, but it can happen in rare cases such as when a subtable is considered degenerate (eg. #3853)


qxliu76 (approved):
LGTM

PR #3859, opened 2022-10-28
freetype: Fix function signatures to match without casts

mstorsjo:
Clang 16 has got a new stricter warning for casts of function types (see https://github.com/llvm/llvm-project/commit/1aad641c793090b4d036c03e737df2ebe2c32c57).

This new warning gets included as part of the existing error diagnostic setting of -Wcast-function-type.

This fixes errors like these:

```
../src/hb-ft.cc:1011:34: error: cast from 'void (*)(FT_Face)' (aka 'void (*)(FT_FaceRec_ *)') to 'FT_Generic_Finalizer' (aka 'void (*)(void *)') converts to incompatible function type [-Werror,-Wcast-function-type-strict]
    ft_face->generic.finalizer = (FT_Generic_Finalizer) hb_ft_face_finalize;
                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/hb-ft.cc
+++ b/src/hb-ft.cc
@@ -732,16 +732,18 @@ hb_ft_get_font_h_extents (hb_font_t *font HB_UNUSED,
 
 static int
 _hb_ft_move_to (const FT_Vector *to,
-		hb_draw_session_t *drawing)
+		void *arg)
 {
+  hb_draw_session_t *drawing = (hb_draw_session_t *) arg;
   drawing->move_to (to->x, to->y);
   return FT_Err_Ok;
 }
 
 static int
 _hb_ft_line_to (const FT_Vector *to,
-		hb_draw_session_t *drawing)
+		void *arg)
 {
+  hb_draw_session_t *drawing = (hb_draw_session_t *) arg;
   drawing->line_to (to->x, to->y);
   return FT_Err_Ok;
 }
@@ -749,8 +751,9 @@ _hb_ft_line_to (const FT_Vector *to,
 static int
 _hb_ft_conic_to (const FT_Vector *control,
 		 const FT_Vector *to,
-		 hb_draw_session_t *drawing)
+		 void *arg)
 {
+  hb_draw_session_t *drawing = (hb_draw_session_t *) arg;
   drawing->quadratic_to (control->x, control->y,
 			 to->x, to->y);
   return FT_Err_Ok;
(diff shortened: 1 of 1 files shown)


mstorsjo:
Also FWIW, the same warning (which gets upgraded to an error by the pragmas) fires a lot in the glib/gobject code - I didn't set out to try to fix those cases.

behdad:
> Also FWIW, the same warning (which gets upgraded to an error by the pragmas) fires a lot in the glib/gobject code - I didn't set out to try to fix those cases.

If you can also send PR for that it would be appreciated. Thanks!

mstorsjo:
> > Also FWIW, the same warning (which gets upgraded to an error by the pragmas) fires a lot in the glib/gobject code - I didn't set out to try to fix those cases.
> 
> If you can also send PR for that it would be appreciated. Thanks!

Sorry, those were out of scope for what I originally encountered - I ran into these when testing cross compilation for mingw targets, but for making the patch for upstreaming, I tested it in a linux build. These cases seemed straightforward enough to fix (where there actually was a real function signature difference that could be fixed), but for the glib errors, it seems to be things from plain C headers, which ends up turned into errors by the harfbuzz pragmas:

```
In file included from ../src/hb-glib.cc:33: 
In file included from ../src/hb-glib.h:34:
In file included from /usr/include/glib-2.0/glib.h:113:
/usr/include/glib-2.0/glib/glib-autocleanups.h:48:1: error: cast from 'void (*)(GChecksum *)' (aka 'void (*)(_GChecksum *)') to 'void (*)()' converts to incompatible function type [-Werror,-Wcast-function-type-strict]
G_DEFINE_AUTOPTR_CLEANUP_FUNC(GChecksum, g_checksum_free)
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/glib-2.0/glib/gmacros.h:1061:3: note: expanded from macro 'G_DEFINE_AUTOPTR_CLEANUP_FUNC'
  _GLIB_DEFINE_AUTOPTR_CLEANUP_FUNCS(TypeName, TypeName, func) 
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/glib-2.0/glib/gmacros.h:1049:47: note: expanded from macro '_GLIB_DEFINE_AUTOPTR_CLEANUP_FUNCS'
    { g_list_free_full (*_l, (GDestroyNotify) (void(*)(void)) cleanup); }                                       \
                                              ^~~~~~~~~~~~~~~~~~~~~~~ 
```

Not sure if there's much else to do here than to try to disable the warning?

behdad:
> Not sure if there's much else to do here than to try to disable the warning?

Okay I disabled it. Can you test please?

mstorsjo:
> > Not sure if there's much else to do here than to try to disable the warning?
> 
> Okay I disabled it. Can you test please?

That seems to have done the trick, thanks!

PR #3860, opened 2022-10-28
[subset] add an experimental API that can override name strings for specified name_id

qxliu76:


src/hb-ot-name-table.hh
```diff
@@ -97,12 +97,30 @@ struct NameRecord
...
+      if (!c->check_assign (out->length, len, HB_SERIALIZE_ERROR_INT_OVERFLOW)) {
+        c->revert (snap);
+        return_trace (nullptr);
+      }
+
+      hb_bytes_t name_bytes (name_str, len);
+
+      out->offset = 0;
+      c->push ();
+      name_bytes.copy (c);
```
garretrieger, line 119: You'll need to convert the strings into UTF-16BE before storage, the spec says the strings are:

"Strings for the Unicode platform must be encoded in UTF-16BE."

@behdad is there an existing method in harfbuzz to encode the string into UTF-16BE?

src/hb-subset-input.cc
```diff
@@ -478,4 +480,26 @@ hb_subset_preprocess (hb_face_t *source)
 
   return new_source;
 }
+
+/**
+ * hb_subset_input_override_name_table:
+ * @input: a #hb_subset_input_t object.
+ * @name_id: name_id of a nameRecord
+ * @name_str: pointer to name string new value or null to indicate should remove
```
garretrieger, line 488: Can you clarify that a null will remove the string from the override map, not from the name table.

src/hb-ot-name-table.hh
qxliu76: Can we assume the input string is encoded in UTF-16BE already? We can do this conversion in JAVA

jfkthame: I don't think a harfbuzz API should require that. It should take UTF-16 (platform-endian) or UTF-8.

behdad: It encodes one character at a time.

garretrieger: My preference is to have the API take UTF-8 and then do encoding to UTF-16BE as that is most likely what callers will be using (also fine with having a second version of the method that can take a UTF-16)

src/hb-subset-input.cc
garretrieger: Oh sorry I misread the code I thought you were deleting the entry from the map when null, but instead you're removing it from the name id set which is correct. So nothing to do here.

src/hb-ot-name-table.hh
behdad: Don't expose it. Just use it internally.

src/hb-ot-name-table.hh
```diff
@@ -97,12 +97,34 @@ struct NameRecord
...
   {
     TRACE_SERIALIZE (this);
+    auto snap = c->snapshot ();
     auto *out = c->embed (this);
     if (unlikely (!out)) return_trace (nullptr);
-    out->offset.serialize_copy (c, offset, base, 0, hb_serialize_context_t::Tail, length);
+    if (name_table_overrides->has (nameID)) {
+      const char *name_str = name_table_overrides->get (nameID);
+      unsigned text_size = 4 * strlen (name_str);
+      char *name_str_utf16_be = (char *) hb_calloc (text_size + 1, HBUINT16::static_size);
```
garretrieger, line 110: text_size is the number of bytes, so divide it by two since we're allocating a number of HBUINT16's here.

qxliu76: Done, I made it = text_size/2 +1 to save room for NULL terminator.

src/hb-ot-name-table.hh
qxliu76: Done. thanks!

garretrieger (approved):

src/hb-ot-name-table.hh
```diff
@@ -97,12 +142,39 @@ struct NameRecord
...
     TRACE_SERIALIZE (this);
+    auto snap = c->snapshot ();
     auto *out = c->embed (this);
     if (unlikely (!out)) return_trace (nullptr);
-    out->offset.serialize_copy (c, offset, base, 0, hb_serialize_context_t::Tail, length);
+    if (name_table_overrides->has (nameID)) {
+      const char *name_str = name_table_overrides->get (nameID);
+      unsigned len = strlen (name_str);
+
+      unsigned text_size = 4 * len + 1;
```
behdad, line 156: Don't do this. Get `encoded_len` by calling `hb_ot_name_convert_utf` one time first with nullptr as text_size and output.

src/hb-subset-input.cc
```diff
@@ -478,4 +480,26 @@ hb_subset_preprocess (hb_face_t *source)
...
+ **/
+HB_EXTERN void
+hb_subset_input_override_name_table (hb_subset_input_t  *input,
+                                     unsigned            name_id,
+                                     const char         *name_str)
+{
+  if (!name_str || strlen (name_str) == 0) {
+    hb_set_del (hb_subset_input_set(input, HB_SUBSET_SETS_NAME_ID), name_id);
+  } else {
+    input->name_table_overrides->set (name_id, name_str);
```
behdad, line 501: We should copy the string, and free it properly.

src/hb-subset-input.cc
```diff
@@ -478,4 +480,26 @@ hb_subset_preprocess (hb_face_t *source)
...
+ * @name_id: name_id of a nameRecord
+ * @name_str: pointer to name string new value or null to indicate should remove
+ *
+ * Override the name string of a nameRecord with specified name_id
+ * Since: EXPERIMENTAL
+ **/
+HB_EXTERN void
+hb_subset_input_override_name_table (hb_subset_input_t  *input,
+                                     unsigned            name_id,
+                                     const char         *name_str)
```
behdad, line 496: Everywhere we take a string we also take its length (where -1 means "use strlen").

behdad: Also use `hb_ot_name_id_t`

src/hb-ot-name-table.hh
```diff
@@ -97,12 +142,39 @@ struct NameRecord
...
-  NameRecord* copy (hb_serialize_context_t *c, const void *base) const
+  NameRecord* copy (hb_serialize_context_t *c, const void *base,
+                    const hb_hashmap_t<unsigned, const char*> *name_table_overrides) const
   {
     TRACE_SERIALIZE (this);
+    auto snap = c->snapshot ();
     auto *out = c->embed (this);
     if (unlikely (!out)) return_trace (nullptr);
-    out->offset.serialize_copy (c, offset, base, 0, hb_serialize_context_t::Tail, length);
+    if (name_table_overrides->has (nameID)) {
```
behdad, line 152: This branch should also set platform, etc, to make sure they are what we expect. But then the sort order might go wrong :(.. Humm... Maybe is okay to assume they are correct. Ouch.

Also, currently, names corresponding to the nameID, in ALL languages, will be overriden.

In general this API is very brittle and I don't like it. I'm happy to leave it in experimental for now though.

src/hb-subset-input.cc
qxliu76: Done.

src/hb-subset-plan.cc
```diff
@@ -848,6 +848,10 @@ hb_subset_plan_create_or_fail (hb_face_t	 *face,
   plan->check_success (plan->vmtx_map = hb_hashmap_create<unsigned, hb_pair_t<unsigned, int>> ());
   plan->check_success (plan->hmtx_map = hb_hashmap_create<unsigned, hb_pair_t<unsigned, int>> ());
 
+  plan->check_success (plan->name_table_overrides = hb_hashmap_create<unsigned, hb_bytes_t> ());
+  if (plan->name_table_overrides && input->name_table_overrides)
+    *plan->name_table_overrides = *input->name_table_overrides;
```
garretrieger, line 853: As this is currently set up input->name_table_overrides is the owner of these strings  and plan->name_table_overrides is just holding references to those same strings. However, it's not safe to assume that the subset input object will live as long as the subset plan object (since the plan can be created separately from plan execution).

So it will be necessary to reallocate memory for each string and have the plan own it's own copy of each.

qxliu76: Done.

garretrieger (approved):

Issue #3863, opened 2022-11-01
Indic reordering of pre-base reordering characters is backwards


dscorbett:
<i>The Unicode Standard, Version 15.0</i> [clarifies](https://www.unicode.org/versions/Unicode15.0.0/ch02.pdf#page=50) that “For combining characters placed to the left of a base character, the combining characters start from the base character and are arranged leftward.” HarfBuzz’s USE shaper already does this.

The Indic shaper needs updating. Here is ⟨किॎ⟩ <U+0915, U+093F, U+094E>:
![किॎ](https://user-images.githubusercontent.com/1124347/199331645-46e53b58-90f6-4b4b-bd79-93f49c72b625.png)

The Myanmar shaper is a modification of the Indic shaper so it also needs updating. Here is ⟨ကေႄ⟩ <U+1000, U+1031, U+1084>:
![ကေႄ](https://user-images.githubusercontent.com/1124347/199332722-35759131-0052-4d38-8755-4c63bca7f2d7.png)

The Khmer shaper does not allow multiple pre-base vowels so this issue does not apply. It probably uses a copy of the Indic code for reordering too, though, so it still ought to be fixed.

behdad:
> The Myanmar shaper is a modification of the Indic shaper so it also needs updating. Here is ⟨ကေႄ⟩ <U+1000, U+1031, U+1084>:

I thought for that one we tested and the order was intentional. @jfkthame @mhosken

PR #3866, opened 2022-11-04
[indic] Order left-matras inside-out

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3863

How's this for implementation? One test fails currently.  I haven't studied it but is probably expected.
```
hb-shape /home/behdad/harfbuzz/test/shape/data/in-house/fonts/3493e92eaded2661cadde752a39f9d58b11f0326.ttf --unicodes U+0BA4,U+0BC6,U+1133C,U+0BAA,U+1133C,U+0BC6,U+1133C
Actual:   [u1133C=0+0|u0BC6=0+2093|u0BA4=0+1863|u1133C=3+0|u0BC6=3+2093|u0BAA=3+1706|u1133C=3+0]
Expected: [u0BC6=0+2093|u1133C=0+0|u0BA4=0+1863|u0BC6=3+2093|u1133C=3+0|u0BAA=3+1706|u1133C=3+0]
```


dscorbett (changes requested):
When a reordering spacing mark has a non-spacing mark applied to it, like a nukta or a variation selector, the non-spacing mark should follow the spacing mark, even after the call to `reverse_range`. That is why the test is failing.

Once the code is fixed, it should be copied to the Myanmar shaper. I can’t find any evidence that the current behavior was once tested and found intentional.

Because the Khmer shaper doesn’t support multiple pre-base vowels, it doesn’t need changing, but I think it would be good to add a comment to hb-ot-shaper-khmer-machine.rl explaining that if that restriction is ever lifted, the shaper will need updating.

behdad:
> When a reordering spacing mark has a non-spacing mark applied to it, like a nukta or a variation selector, the non-spacing mark should follow the spacing mark, even after the call to `reverse_range`. That is why the test is failing.

Fixed.



> Once the code is fixed, it should be copied to the Myanmar shaper. I can’t find any evidence that the current behavior was once tested and found intentional.

Working on that.

behdad:
Added Myanmar. I tested the Devanagari case, but not Myanmar. Will add Deva test.

What's a good sequence to test Myanmar?

behdad:
> Because the Khmer shaper doesn’t support multiple pre-base vowels, it doesn’t need changing, but I think it would be good to add a comment to hb-ot-shaper-khmer-machine.rl explaining that if that restriction is ever lifted, the shaper will need updating.

Khmer encoding is visual, so there's no sorting and the issue doesn't occur.

behdad:
PTAL.

dscorbett (approved):
This looks good now.

Khmer vowels are encoded in phonetic order, not visual order, so they are reordered, but it doesn’t really matter for this issue anyway.

PR #3870, opened 2022-11-08
meson: fix regression in detecting freetype2/icu-uc when explicitly disabled

eli-schwartz:
In #3811 / commit 53a194aa3f5f7de0b40e879e41fcbe0de6e9fefe a broken and half-implemented approach to kind of sort of handling the detection of both pkg-config and cmake names for dependencies, was implemented. It just checked for both versions with required: false, but when the build was configured with *disabled* options, it was still found because it was treated as auto.

Really, the problem here is trying to outsmart Meson, which handles a lot of edge cases correctly. But it's possible, albeit very wordy, to manually implement Meson's internal logic via if/else fallbacks. Do so here.

Fixes failure to build on Gentoo, /cc @thesamesam

Fixes:
https://bugs.gentoo.org/880405
https://bugs.gentoo.org/880479

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -83,20 +83,35 @@ check_funcs = [
 
 m_dep = cpp.find_library('m', required: false)
 
-
-# Try pkgconfig name
-freetype_dep = dependency('freetype2', required: false)
-if not freetype_dep.found()
-  # Try cmake name
-  freetype_dep = dependency('freetype', required: false)
-endif
-if not freetype_dep.found()
-  # Subproject fallback, `allow_fallback: true` means the fallback will be
-  # tried even if the freetype option is set to `auto`.
-  freetype_dep = dependency('freetype2',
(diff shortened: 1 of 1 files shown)


behdad:
The CIFuzz build is failing with:
```
meson.build:145:6: ERROR: Fallback wrap or subproject not found for dependency 'icu-uc'
```

eli-schwartz:
> > The CIFuzz build is failing with:
> > ```
> > meson.build:145:6: ERROR: Fallback wrap or subproject not found for dependency 'icu-uc'
> > ```
> 
> Same with some other bots.

This is a pointless error which Meson complains about since ICU doesn't have a wrap file installed in the harfbuzz source tree, but specifies that it is allowed to fallback anyway. I tested against newer versions of Meson and didn't notice the issue.

We don't need to specify allow_fallback -- before this PR, it was only specified for freetype2 anyway.

PR #3873, opened 2022-11-11
[subset] Update hb_subset_input_override_name_table API

qxliu76:
Override the name string for the NameRecord identified by name_id, platform_id, encoding_id and language_ids specified by the user. If a record with specified name_id does not exist, this API will create a new NameRecord with provided info and insert it to the name table.

Hopefully this also makes this API less brittle. I'm not sure yet if we would need check to make sure the input platform_id, encoding_id and language_id are valid.


src/hb-subset-input.hh
```diff
@@ -62,7 +62,7 @@ struct hb_subset_input_t
   bool attach_accelerator_data = false;
   hb_hashmap_t<hb_tag_t, float> *axes_location;
 #ifdef HB_EXPERIMENTAL_API
-  hb_hashmap_t<unsigned, hb_bytes_t> *name_table_overrides;
+  hb_hashmap_t<hb_vector_t<unsigned>, hb_bytes_t> *name_table_overrides;
```
garretrieger, line 65: Suggest having a custom struct to hold the id values.

src/hb-ot-name-table.hh
```diff
@@ -153,10 +153,23 @@ struct NameRecord
...
-      unsigned text_size = hb_ot_name_convert_utf<hb_utf8_t, hb_utf16_be_t> (name_bytes, nullptr, nullptr);
+      if (unlikely (!record_ids.alloc (4)))
+      {
+        c->revert (snap);
+        return_trace (nullptr);
+      }
+      record_ids.push (platformID);
+      record_ids.push (encodingID);
+      record_ids.push (languageID);
+      record_ids.push (nameID);
```
behdad, line 167: Why use a vector instead of a struct?!

qxliu76: oh yes it was a struct initially and I don't remember why I changed it.... I'll make another try.

qxliu76: oh... I remember now, after I added struct definition into `hb-ot-name-table.hh` and then included it in `hb-subset-plan.hh`, build failed because there's circular dependency: `hb-subset-plan.hh->hb-ot-name-table.hh->hb-open-type.hh->hb-subset.hh`, and `hb-subset.hh->hb-subset-plan.hh`, would you have any suggestion on where should I declare this struct? Thanks!

qxliu76: or maybe I just create a new .hh file

behdad: I would say in hb-subset-plan.hh should work. Just before the API definition. No?

qxliu76: Can I put it in `hb-subset-input.hh`? The struct needs to be visible to hb-subset-input as well, and hb-subset-plan.hh depends on hb-subset-input.hh, I tried that it's

behdad: Sure.

src/hb-subset-input.hh
qxliu76: Done.

src/hb-subset-input.cc
```diff
@@ -501,38 +501,47 @@ hb_subset_preprocess (hb_face_t *source)
...
-
-  if (str_len == -1)
-    str_len = strlen (name_str);
-
-  if (!str_len)
+  else
   {
-    hb_set_del (hb_subset_input_set(input, HB_SUBSET_SETS_NAME_ID), name_id);
-    return;
+    if (str_len == -1)
```
garretrieger, line 531: this can be an "else if"

garretrieger (approved):

src/hb-subset-input.cc
qxliu76: Done

src/hb-subset-input.hh
```diff
@@ -36,6 +36,52 @@
...
+      language_id (language_id_),
+      name_id (name_id_) {}
+
+  bool operator != (const hb_ot_name_record_ids_t o) const
+  {
+    return platform_id != o.platform_id ||
+           encoding_id != o.encoding_id ||
+           language_id != o.language_id ||
+           name_id != o.name_id;
+  }
```
behdad, line 57: Implement this as `!(*this == o)`.

src/hb-subset-input.hh
```diff
@@ -36,6 +36,48 @@
...
+    return platform_id == o.platform_id &&
+           encoding_id == o.encoding_id &&
+           language_id == o.language_id &&
+           name_id == o.name_id;
+  }
+
+  inline uint32_t hash () const
+  {
+    uint32_t current = 0;
+    current = current * 31 + (uint32_t) (platform_id * 2654435761u);
```
behdad, line 65: Instead of that multiplication, use `hb_hash()`.

qxliu76: Done

qxliu76:
Added some code for mac platform,  we'll check if the name string to be set is all characters in 0-128, if that's true just pass the name string through as is, no need to encode it with UTF16-BE. Other wise don't include the name record.

Could you help review again? Thanks!

src/hb-ot-name-table.hh
```diff
@@ -477,6 +563,38 @@ struct name
...
+      if (record.platformID == record_ids.platform_id &&
+          record.encodingID == record_ids.encoding_id &&
+          record.languageID == record_ids.language_id &&
+          record.nameID == record_ids.name_id)
+        return true;
+    }
+    return false;
+  }
+
+  bool all_ascii_characters (hb_bytes_t bytes) const
```
garretrieger, line 582: I would move this check into hb-subset-input.cc when the entry is first being added to the override map. Then if the string is not valid you can skip adding it to the map and return false from that method to signal to the caller it was not accepted.

qxliu76: The problem is hb_utf8_t is not visible to hb-subset-input.cc and `hb-utf.hh` can not be included in `hb-subset-input.hh` because there's circular dependency issue. Any suggestions?

qxliu76: Oh my mistake, I can include it in `.cc` not `.hh` file. I'll fix it.

qxliu76: Done.

PR #3874, opened 2022-11-11
[GPOS.PairPos] Adjust unsafe-to-break for non-zero ValueFormat2

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3824

### DIFF
--- a/src/OT/Layout/GPOS/PairPosFormat2.hh
+++ b/src/OT/Layout/GPOS/PairPosFormat2.hh
@@ -244,7 +244,11 @@ struct PairPosFormat2_4
 
     buffer->idx = skippy_iter.idx;
     if (len2)
+    {
       buffer->idx++;
+      // https://github.com/harfbuzz/harfbuzz/issues/3824
+      buffer->unsafe_to_break (buffer->idx - 1, buffer->idx + 1);
+    }
 
     return_trace (true);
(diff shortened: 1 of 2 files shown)


behdad:
@caolanm can you test this?

behdad:
There's a further adjustment that can be done here. Currently we skip second glyph even if the kerning did not apply any adjustment whatsoever. We can instead, in a rather non-conforming reading of the spec, only skip the second glyph if there was actually any adjustment. @jfkthame

jfkthame:
> There's a further adjustment that can be done here. Currently we skip second glyph even if the kerning did not apply any adjustment whatsoever. We can instead, in a rather non-conforming reading of the spec, only skip the second glyph if there was actually any adjustment. @jfkthame

Do the other engines do that? While I think it'd make sense to modify the spec (and everyone's behavior) -- thanks for opening that issue -- if we can't get agreement there, then I think having consistent behavior across engines takes priority over a non-conforming improvement in an obscure edge case.

jfkthame (approved):

behdad:
> Do the other engines do that?

Seems to me like CoreText does that, but not Uniscribe.

caolanm:
@behdad yes this solves our problem, thanks

PR #3887, opened 2022-11-22
[subset] Don't trim glyf's again if preprocessed

behdad:
Speeds up M1/10000 benchmark by 30%!


behdad:
This patch makes me uneasy, accessing accelerator from the glyf table subsetting code. But it's so simple and elegant at the same time!

garretrieger (approved):

PR #3896, opened 2022-11-25
[subset] use charstrings directly from accelerator cache if mutability isn't needed.

garretrieger:
~18% improvement for large subset sizes.

BM_subset/subset_glyphs/SourceHanSans-Regular_subset.otf/10000_median                    -0.1794         -0.1796            54            44            54            44
BM_subset/subset_codepoints/SourceHanSans-Regular_subset.otf/10000_median                -0.1807         -0.1809            52            42            52            42

### DIFF
--- a/src/hb-subset-cff-common.hh
+++ b/src/hb-subset-cff-common.hh
@@ -583,7 +583,8 @@ template <typename SUBSETTER, typename SUBRS, typename ACC, typename ENV, typena
 struct subr_subsetter_t
 {
   subr_subsetter_t (ACC &acc_, const hb_subset_plan_t *plan_)
-      : acc (acc_), plan (plan_), closures(acc_.fdCount), remaps(acc_.fdCount)
+      : acc (acc_), plan (plan_), closures(acc_.fdCount), cached_charstrings(),
+        remaps(acc_.fdCount)
   {}
 
   /* Subroutine subsetting with --no-desubroutinize runs in phases:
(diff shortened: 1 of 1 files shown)


src/hb-subset-cff-common.hh
```diff
@@ -1038,12 +1049,25 @@ struct subr_subsetter_t
 
   }
 
+  parsed_cs_str_t& get_parsed_charstring (unsigned i)
+  {
+    if (cached_charstrings.length) return *(cached_charstrings[i]);
```
behdad, line 1054: vector has operator bool so you can just say `if (cached_charstrings)`.

behdad:
You are using a vector of pointers. I was thinking of not even using. Is that too much effort, to store a pointer to the vector-of-charstrings in the accel?

behdad:
Shouldn't really matter for performance.

garretrieger:
> You are using a vector of pointers. I was thinking of not even using. Is that too much effort, to store a pointer to the vector-of-charstrings in the accel?

The parsed_charstrings vector uses different indexing (the new gids) vs the cached one (old gids)

src/hb-subset-cff-common.hh
garretrieger: Fixed

behdad:
> > You are using a vector of pointers. I was thinking of not even using. Is that too much effort, to store a pointer to the vector-of-charstrings in the accel?
> 
> The parsed_charstrings vector uses different indexing (the new gids) vs the cached one (old gids)

I see. Sounds good.

PR #3901, opened 2022-11-29
[subset] Cache per subtable cmap unicode mappings.

garretrieger:
22-27% improvement for the 10 codepoint CFF subsets.

### DIFF
--- a/src/hb-ot-cmap-table.hh
+++ b/src/hb-ot-cmap-table.hh
@@ -1474,15 +1474,54 @@ struct EncodingRecord
   DEFINE_SIZE_STATIC (8);
 };
 
+struct cmap;
+
 struct SubtableUnicodesCache {
 
  private:
+  hb_blob_ptr_t<cmap> base_blob;
   const char* base;
   hb_hashmap_t<unsigned, hb::unique_ptr<hb_set_t>> cached_unicodes;
 
  public:
+
+  static SubtableUnicodesCache* create (hb_blob_ptr_t<cmap> source_table)
+  {
+    SubtableUnicodesCache* cache =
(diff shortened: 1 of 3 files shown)


src/hb-ot-cmap-table.hh
```diff
@@ -1538,7 +1594,14 @@ struct cmap
     unsigned format4objidx = 0, format12objidx = 0, format14objidx = 0;
     auto snap = c->snapshot ();
 
-    SubtableUnicodesCache unicodes_cache (base);
+    SubtableUnicodesCache local_unicodes_cache (base);
+    SubtableUnicodesCache* unicodes_cache = &local_unicodes_cache;
+
+    if (plan->accelerator &&
+        plan->accelerator->cmap_cache &&
+        plan->accelerator->cmap_cache->same_base (base))
```
behdad, line 1602: Why do you need the `same_base` check here?

behdad: I can see it as a good idea but curious why you added.

behdad: Actually, unless you always retain the old blob, this is racy since old and new blob might have the same base pointer address.

behdad: It looks to me like it should always be true and its a safety check. I''m happy with that. Or maybe change it to an assert.

behdad: That might fail indeed if the face loader returns new copies of tables each time...

garretrieger: This is mostly a safety check in the case of the situation such as where reference table references a different memory address each time it's called. In that case we just avoid using the cache.

PR #3912, opened 2022-12-02
[subset-cff] Compact parsed strings if using accelerator

behdad:
Saves 32% on SourceHanSans/10000 benchmark!

Also, use memcmp now for writing out strings since now that our ops are not super short, that's faster.

This makes cff-japanese test takes super long though; that needs inspection.

### DIFF
--- a/src/hb-array.hh
+++ b/src/hb-array.hh
@@ -267,7 +267,7 @@ struct hb_array_t : hb_iter_with_fallback_t<hb_array_t<Type>, Type&>
 
   template <typename hb_serialize_context_t,
 	    typename U = Type,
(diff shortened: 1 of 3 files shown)


behdad:
> This makes cff-japanese test takes super long though; that needs inspection.

I'm guessing that something in the binary changes, and we end up ttx'ing and that passes... Curious though.

behdad:
> > This makes cff-japanese test takes super long though; that needs inspection.
> 
> I'm guessing that something in the binary changes, and we end up ttx'ing and that passes... Curious though.

Humm. Not it seems like preprocessing just got much slower; which is unexpected.

behdad:
At 50% speedup now! Some of the later changes are at the expense of non-preprocessed case.

garretrieger (approved):

PR #3914, opened 2022-12-02
[multimap] Add a multimap datastructure & use for gid-to-unicodes subset accelerator

behdad:
To be used for glyph-to-unicodes mapping in the subset accelerator.


src/test-multimap.cc
```diff
@@ -0,0 +1,50 @@
...
+  assert (m.get (10).length == 1);
+
+  m.add (10, 12);
+  assert (m.get (10).length == 2);
+
+  m.add (10, 13);
+  assert (m.get (10).length == 3);
+  assert (m.get (10)[0] == 11);
+  assert (m.get (10)[1] == 12);
+  assert (m.get (10)[2] == 13);
```
garretrieger, line 45: Maybe also test adding some values under a different key.

behdad:
37% speedup on BM_subset/subset_glyphs/SourceHanSans-Regular_subset.otf/nohinting/10 benchmark.

garretrieger (approved):

PR #3919, opened 2022-12-05
[repacker] fix two fuzzer issues.

garretrieger:
https://oss-fuzz.com/testcase-detail/5390364397928448
https://oss-fuzz.com/testcase-detail/6714085985353728


behdad:
asan unhappy.

garretrieger:
> asan unhappy.

Fixed.

Issue #3352, opened 2022-01-03
[docs] Fix sorting


behdad:
Some sections have all `_get_` before all `_set_`. Others have them interleaved. Decide which is better and adhere. Ie. alphabetical or curated.

behdad:
> I can't say I have a _strong_ feeling about the options, but perhaps one could say it's slightly preferable to keep each "get_x" next to "set_x", just so that a casual reader doesn't wonder momentarily if they're looking at the strange getter that doesn't have a matching setter or vice versa...

I'm fine with that.

PR #3924, opened 2022-12-08
[doc] Fix sorting

khaledhosny:
* Keep setters and getters together, with setters first.
* Keep common functions at the top and in a predictable order.
* Put callback functions right above their setters.

Fixes https://github.com/harfbuzz/harfbuzz/issues/3352


behdad:
Thanks Khaled!

PR #3928, opened 2022-12-11
COLRv1: use ClipBoxes for extents

matthiasclasen:
This is a first step; ultimatively, we should compute the extents is ClipBoxes are missing.

### DIFF
--- a/src/hb-ot-color-colr-table.hh
+++ b/src/hb-ot-color-colr-table.hh
@@ -930,9 +930,7 @@ struct ClipBox
   }
 
   bool get_extents (hb_glyph_extents_t *extents,
-		    const DeltaSetIndexMap &varIdxMap,
-		    const VariationStore &varStore,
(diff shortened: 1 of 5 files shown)


matthiasclasen:
This also needs support for the variable clipbox format.

behdad:
Would be nice to add a small test.

behdad:
This now just need variations to be hooked up. I'll see what I can do.

matthiasclasen:
I tried with your variation hookup, but I can't observe any clipbox changes, despite dragging the right axis sliders, and seeing get_extents() getting called in response - your code always adds 0s. Printing out the coords shows that my slider changes are applied.

matthiasclasen:
This works in quick testing (at least I can confirm that variations are applied to these extents now).

behdad:
Let's open new issue for v0

PR #3933, opened 2022-12-12
[subset] Fix infinite loop when instancing CFF fonts

### LINKED ISSUE #3930: [subset] L1 instancing of fonts with CFF2 table seems to result in an infinite loop
This works (I think 200 is the default instance)
```
$ hb-subset SourceCodeVariable-Roman.otf --instance="wght=200" -o out.otf
```

But this never finishes:
```
$ hb-subset SourceCodeVariable-Roman.otf --instance="wght=400" -o out.otf
```

Font from [VAR-source-code-var-1.018R.zip](https://github.com/adobe-fonts/source-code-pro/releases/download/2.038R-ro%2F1.058R-it%2F1.018R-VAR/VAR-source-code-var-1.018R.zip)

garretrieger:
Fixes #3930

### DIFF
--- a/test/subset/generate-expected-outputs.py
+++ b/test/subset/generate-expected-outputs.py
@@ -35,7 +35,6 @@ def generate_expected_output(input_file, unicodes, profile_flags, instance_flags
 			"--no-overlap-flag",
 			"--no-recalc-bounds",
 			"--no-recalc-timestamp",
-			"--no-harfbuzz-repacker", # disable harfbuzz repacker so we aren't comparing to ourself.
 			"--output=%s" % instance_path,
 			input_file]
 		args.extend(instance_flags)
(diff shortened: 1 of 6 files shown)


test/subset/generate-expected-outputs.py
```diff
@@ -35,7 +35,6 @@ def generate_expected_output(input_file, unicodes, profile_flags, instance_flags
 			"--no-overlap-flag",
 			"--no-recalc-bounds",
 			"--no-recalc-timestamp",
-			"--no-harfbuzz-repacker", # disable harfbuzz repacker so we aren't comparing to ourself.
```
khaledhosny, line 38: Did you mean to remove this line?

garretrieger: Yes, the fonttools instancer doesn't support that flag.

PR #3934, opened 2022-12-13
VarC: properly cast int

alatiera:
msvc is rightfully complaining that the types on the sides of the ternary are not matching:

```
C:\pango\subprojects\harfbuzz\src\OT\glyf\VarCompositeGlyph.hh(317): error C2446: ':': no conversion from 'const OT::HBUINT16' to 'const OT::HBUINT8'
```

### DIFF
--- a/src/OT/glyf/VarCompositeGlyph.hh
+++ b/src/OT/glyf/VarCompositeGlyph.hh
@@ -314,7 +314,7 @@ struct VarCompositeGlyphRecord
     unsigned count = numAxes;
     for (unsigned i = 0; i < count; i++)
     {
-      unsigned axis_index = axis_width == 1 ? *p++ : *q++;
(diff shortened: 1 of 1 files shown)


alatiera:
It's been a minute since I touched C, let alone cpp, very possible this is just wrong.

(Still testing this works on all platforms)

alatiera:
Seems to be working with msvc2019 and fedora

jfkthame:
I don't think it can be right to cast between HBUINT8 and HBUINT16 pointers. That'll cause the pointed-at data to be misinterpreted.

If you need to cast things here, cast the integer value after dereferencing the pointer. Maybe something like:
```
      unsigned axis_index = axis_width == 1 ? (unsigned) *p++ : (unsigned) *q++;
```

alatiera:
Thanks! wasn't sure at first if the types were int aliases or not. Updated the patch.

PR #3936, opened 2022-12-13
Colrv0 glyph extents

matthiasclasen:
Untested


src/hb-ot-color-colr-table.hh
```diff
@@ -932,27 +932,21 @@ struct ClipBox
...
-	extents->height    += _hb_roundf (instancer (varIdx+3));
-      }
+    extents->x_bearing = u.format1.xMin;
+    extents->y_bearing = u.format1.yMax;
+    extents->width = u.format1.xMax - u.format1.xMin;
+    extents->height = u.format1.yMin - u.format1.yMax;
 
-      return true;
-    default:
-      return false;
```
behdad, line 954: What about the unknown-format case?

Issue #3931, opened 2022-12-11
COLRv1 api


matthiasclasen:
I want to use harfbuzz instead of freetype whereever possible. 
Currently, I can't get at the COLRv1 data without using freetype.

Here is a draft for a callback-style api for COLRv1, tentatively called hb-paint:

```
typedef struct hb_paint_funcs_t hb_paint_funcs_t;

typedef void (*hb_paint_push_transform_func_t) (hb_paint_funcs_t *funcs,
                                                void *draw_data,
                                                float xx, float xy,
                                                float yx, float yy,
                                                float x0, float y0,
                                                void *user_data);
typedef void (*hb_paint_pop_transform_func_t)  (hb_paint_funcs_t *funcs,
                                                void *draw_data,
                                                void *user_data);

typedef void (*hb_paint_push_clip_func_t)      (hb_paint_funcs_t *funcs,
                                                void *draw_data,
                                                hb_codepoint_t glyph,
                                                void *user_data); 
typedef void (*hb_paint_pop_clip_func_t)       (hb_paint_funcs_t *funcs,
                                                void *draw_data,
                                                void *user_data);

typedef void (*hb_paint_solid_func_t)          (hb_paint_funcs_t *funcs,
                                                void *draw_data,
                                                unsigned int color_index,
                                                void *user_data);

typedef struct hb_color_line_t hb_color_line_t;

typedef struct {
  float offset;
  unsigned int color_index;
} hb_color_stop_t;

unsigned int hb_color_line_get_color_stops (hb_color_line_t color_line,
                                            unsigned int start,
                                            unsigned int count,
                                            hb_color_stop_t *color_stops);

typedef enum {
  HB_PAINT_EXTEND_PAD,
  HB_PAINT_EXTEND_REPEAT,
  HB_PAINT_EXTEND_REFLECT
} hb_paint_extend_t;

hb_paint_extend_t hb_color_line_get_extend (hb_color_line_t color_line);

typedef void (*hb_paint_linear_gradient_func_t) (hb_paint_funcs_t *funcs,
                                                 void *draw_data,
                                                 hb_color_line_t *color_line,
                                                 hb_position_t x0, hb_position_t y0,
                                                 hb_position_t x1, hb_position_t y1,
                                                 hb_position_t x2, hb_position_t y2,
                                                 void *user_data);
                                             
typedef void (*hb_paint_radial_gradient_func_t) (hb_paint_funcs_t *funcs,
                                                 void *draw_data,
[...]

PR #3938, opened 2022-12-15
hb-paint API

matthiasclasen:
This PR adds a callback-based paint api.

It started out as a way to expose COLRv1 data, but it has been extended to
allow painting all kinds of glyphs.

Fixes https://github.com/harfbuzz/harfbuzz/issues/3931


src/hb-paint.cc
```diff
@@ -30,9 +30,9 @@
 
 static void
 hb_paint_push_transform_nil (hb_paint_funcs_t *funcs, void *paint_data,
-                             float xx, float xy,
```
behdad: Humm. I know this is controvertial. But I'm used to "xx, xy, yx, yy" order. I think cairo's one of the few to go against that.

behdad: Nevermind.

behdad:
Matthias and I are happy with this now.  The hb-paint API is a unified API that draws COLRv1, COLRv0, sbix, CBDT, as well as regular glyf, CFF, CFF2 outlinels.

matthiasclasen:
Two suggestions I have lingering in my head worth discussion about this api:

- Drop the 'direct call' apis (ie hb_paint_push_transform() etc). If you need to write tests that simulate painting, you can just call the callbacks themselves
-  Add a state struct to the callbacks, like hb-draw has. We can use that to report errors without overloading the boolean return

behdad:
> * Drop the 'direct call' apis (ie hb_paint_push_transform() etc). If you need to write tests that simulate painting, you can just call the callbacks themselves. Maybe one use that could be worthwhile is implementing svg glyphs by translating them to COLRv1

Yeah I like to keep them for exactly what you said, client code driving HB.

> * Add a state struct to the callbacks, like hb-draw has. We could use that to report errors without overloading the boolean return

But *we* don't have error to return really. Fonts with bad tables are not considered error in HarfBuzz. Client can track their own error status (out of memory, etc) in the closure.

> * Should this api be marked as experimental, initially ?

That might be a good idea, though not necessarily needed, given the level of care that has gone into this.

behdad:
> [draw] Add hb_font_draw_glyph

Should we deprecate the old stuff? I think we can.

matthiasclasen:
> Should we deprecate the old stuff? I think we can.

Yes, we can. I'll cut over more completely to the new naming, and leave the old apis as deprecated wrappers.

src/hb-font.cc
```diff
@@ -1366,6 +1398,51 @@ hb_font_get_glyph_shape (hb_font_t *font,
...
+ *
+ * Since: Replaceme
+ */
+void
+hb_font_draw_glyph (hb_font_t *font,
+		    hb_codepoint_t glyph,
+		    hb_draw_funcs_t *dfuncs, void *draw_data)
+{
+  hb_font_get_glyph_shape (font, glyph, dfuncs, draw_data);
+}
```
khaledhosny, line 1418: I’m a bit uneasy about this naming, since the function is not doing any drawing, but calling client callbacks that will do the actual drawing. If we grow a rasterization API at some point, this naming is going to be confusing.

src/hb-font.h
```diff
@@ -530,6 +513,42 @@ typedef void (*hb_font_get_glyph_shape_func_t) (hb_font_t *font, void *font_data
...
+ * Alternative name for #hb_font_get_glyph_shape_func_t.
+ *
+ * Since: REPLACEME
+ *
+ **/
+typedef void (*hb_font_draw_glyph_func_t) (hb_font_t *font, void *font_data,
+                                           hb_codepoint_t glyph,
+                                           hb_draw_funcs_t *draw_funcs, void *draw_data,
+                                           void *user_data);
+
```
khaledhosny, line 534: IMHO again, I don’t believe the new names (even if we decide they are better, which I’m not convinced about) are so better it is worth the duplication and the confusing and the complications of API deprecation.

khaledhosny: I mean the client will have to check for HarfBuzz versions and call different functions for different versions and all these complications, but it is the exact same API only renamed.

matthiasclasen: Yes, its a tradeoff. Between short-term inconvenience and long-term inconsistency

matthiasclasen: Nobody is forced to do anything anyway. The old api is not going away

matthiasclasen: The api is called hb-draw anyway, so...

matthiasclasen: The only argument I see here is whether the new names are better or not.

And yes, at the very beginning, it was called hb_font_get_glyph_paint. But I find that a pretty awkward name, even worse than hb_font_get_glyph_shape, which should probably have been called hb_font_glyph_outline, with hb_outline_funcs_t. Too late now.

behdad: I understand the hesitation but I like the consistency of the new API.

khaledhosny: The experimental API had hb_font_draw_glyph, and it wad renamed when it got stabalized last February, now we are renaming it back...

behdad: I feel like there's few enough users that we can update all.

matthiasclasen: I'll try remain neutral here and leave this decision up to you. My input is that it should be consistent.

Either get_glyph_shape+get_glyph_paint or draw_glyph+paint_glyph. A mixture is worth either of these alternatives.

behdad: @khaledhosny Would not deprecating the old `hb_font_get_glyph_shape` name change anything for you?

khaledhosny: It would at least keep existing clients undisturbed. Having two sets of identical but differently named functions will be confusing, but if you feel strongly about the new names, we cn probably hide the old ones from the documentation.

behdad: I want an API that has draw and paint in parallel in it.  I don't feel as strongly as you do about deprecation. Specially since we control all users of this API currently. I doubt anyone else is using it other than FontGoggles and uharfbuzz.

behdad: Plus, we keep deprecated API forever...

khaledhosny: > I doubt anyone else is using it other than FontGoggles and uharfbuzz.

I’m using it in (to be released) LibreOffice 7.5. If the old API gets deprecated, I’ll have to do some cpp dance around it since packagers don’t like deprecation warnings, not a show stopper but feel unnecessary. Anyway, I don’t want to block this on a naming issue, so your call.

behdad: I like the new naming so if you don't mind I like to go ahead with it.

behdad:
One issue is I think we should pass the font down to all callbacks. The ones needing color need to access the face for palette access, and currently they have to carry the font/face in the paint_data themselves. clip_glyph also needs the font to get the outline (and unscale it).

src/hb-paint.h
```diff
@@ -0,0 +1,785 @@
...
+
+/**
+ * hb_paint_composite_mode_t:
+ *
+ * The values of this enumeration describe the compositing modes
+ * that can be used when combining temporary redirected drawing
+ * with the backdrop.
+ *
+ * See the OpenType spec [COLR](https://learn.microsoft.com/en-us/typography/opentype/spec/colr)
+ * section for details.
```
khaledhosny, line 445: Shouldn’t there be a `Since:` here?

src/hb-paint.h
```diff
@@ -0,0 +1,785 @@
...
+/**
+ * hb_paint_extend_t:
+ *
+ * The values of this enumeration determine how color values
+ * outside the minimum and maximum defined offset on a #hb_color_line_t
+ * are determined.
+ *
+ * See the OpenType spec [COLR](https://learn.microsoft.com/en-us/typography/opentype/spec/colr)
+ * section for details.
+ */
```
khaledhosny, line 331: Shouldn’t there be a `Since:` here?

src/hb-paint.h
```diff
@@ -0,0 +1,785 @@
...
+/**
+ * hb_paint_push_group_func_t:
+ * @funcs: paint functions object
+ * @paint_data: The data accompanying the paint functions in hb_font_paint_glyph()
+ * @user_data: User data pointer passed to hb_paint_funcs_set_push_group_func()
+ *
+ * A virtual method for the #hb_paint_funcs_t to use
+ * an intermediate surface for subsequent paint calls.
+ *
+ * The drawing will be redirected to an intermediate surface
```
khaledhosny, line 487: Who will be doing this redirection, HarfBuzz (as implied by the wording) or the callback function (which I think what is meant here)?

util/hb-cairo-utils.c
```diff
@@ -0,0 +1,759 @@
...
+    case HB_PAINT_EXTEND_REPEAT: return CAIRO_EXTEND_REPEAT;
+    case HB_PAINT_EXTEND_REFLECT: return CAIRO_EXTEND_REFLECT;
+    default:;
+    }
+
+  return CAIRO_EXTEND_PAD;
+}
+
+void
+hb_face_get_color (hb_face_t *face,
```
khaledhosny, line 69: Since this is an internal method, shouldn’t it be using a prefix different from public API?

util/helper-cairo-user.hh
```diff
@@ -82,7 +87,7 @@ close_path (hb_draw_funcs_t *dfuncs,
 
 
 static hb_draw_funcs_t *
-get_cairo_draw_funcs ()
+get_cairo_draw_funcs (void)
```
khaledhosny, line 90: This is C++, `void` is redundant here.

util/hb-cairo-utils.c
khaledhosny: Or since thus code is included only once, turn it into a `.hh` like the rest and use unprefixed names like other `.hh` files here.

src/OT/glyf/glyf.hh
```diff
@@ -337,6 +338,19 @@ struct glyf_accelerator_t
     return glyph_for_gid (gid).get_extents_without_var_scaled (font, *this, extents);
   }
 
+  bool paint_glyph (hb_font_t *font, hb_codepoint_t gid, hb_paint_funcs_t *funcs, void *data) const
+  {
+    funcs->push_root_transform (data, font);
+
+    funcs->push_clip_glyph (data, gid);
```
khaledhosny, line 345: Shouldn’t there be a plain `paint_glyph()` for uses like this where there is no clipping is needed to avoid the unnecessary overhead of clipping?

util/helper-cairo-user.hh
```diff
@@ -98,181 +103,287 @@ get_cairo_draw_funcs ()
...
+                 void *paint_data,
+                 hb_codepoint_t glyph,
+                 void *user_data)
+{
+  paint_data_t *data = (paint_data_t *) paint_data;
+  cairo_t *cr = data->cr;
+
+  cairo_save (cr);
+  cairo_new_path (cr);
+  hb_font_get_glyph_shape (data->unscaled_font, glyph, get_cairo_draw_funcs (), cr);
```
khaledhosny, line 153: Should be using the new name here.

util/helper-cairo-user.hh
```diff
@@ -98,181 +103,287 @@ get_cairo_draw_funcs ()
...
-      }
-      cairo_restore (cr);
-    }
-  }
-  while (len == ARRAY_LENGTH (layers));
+  hb_position_t x_scale, y_scale;
+  hb_font_get_scale (font, &x_scale, &y_scale);
+  cairo_scale (cr, +1./x_scale, -1./y_scale);
+
+  hb_font_get_glyph_shape (font, glyph, get_cairo_draw_funcs (), cr);
```
khaledhosny, line 344: Should be using the new name here as well.

src/hb-font.h
khaledhosny: The new naming then.

behdad:
Thanks Khaled for the extensive review.

src/OT/glyf/glyf.hh
behdad: Such a function would have to do analysis of the paint graph. I don't think we want to do that.

khaledhosny: May be I’m missing something, but I was thinking of something that would be used when painting COLRv0 or non-COLR tables where we know it is a simple painting with no clipping involved.

khaledhosny:
I get the following warnings building this branch:
```
[142/301] Compiling C++ object util/hb-view.p/hb-view.cc.o
In file included from ../util/hb-view.cc:33:
In file included from ../util/view-cairo.hh:32:
In file included from ../util/helper-cairo.hh:35:
../util/helper-cairo-user.hh:127:46: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_matrix_init (&m, xx, yx, xy, yy, dx, dy);
  ~~~~~~~~~~~~~~~~~                          ^~
../util/helper-cairo-user.hh:127:42: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_matrix_init (&m, xx, yx, xy, yy, dx, dy);
  ~~~~~~~~~~~~~~~~~                      ^~
../util/helper-cairo-user.hh:127:38: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_matrix_init (&m, xx, yx, xy, yy, dx, dy);
  ~~~~~~~~~~~~~~~~~                  ^~
../util/helper-cairo-user.hh:127:34: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_matrix_init (&m, xx, yx, xy, yy, dx, dy);
  ~~~~~~~~~~~~~~~~~              ^~
../util/helper-cairo-user.hh:127:30: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_matrix_init (&m, xx, yx, xy, yy, dx, dy);
  ~~~~~~~~~~~~~~~~~          ^~
../util/helper-cairo-user.hh:127:26: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_matrix_init (&m, xx, yx, xy, yy, dx, dy);
  ~~~~~~~~~~~~~~~~~      ^~
../util/helper-cairo-user.hh:168:54: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_rectangle (cr, xmin, ymin, xmax - xmin, ymax - ymin);
  ~~~~~~~~~~~~~~~                               ~~~~~^~~~~~
../util/helper-cairo-user.hh:168:41: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_rectangle (cr, xmin, ymin, xmax - xmin, ymax - ymin);
  ~~~~~~~~~~~~~~~                  ~~~~~^~~~~~
../util/helper-cairo-user.hh:168:30: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_rectangle (cr, xmin, ymin, xmax - xmin, ymax - ymin);
  ~~~~~~~~~~~~~~~            ^~~~
../util/helper-cairo-user.hh:168:24: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_rectangle (cr, xmin, ymin, xmax - xmin, ymax - ymin);
  ~~~~~~~~~~~~~~~      ^~~~
../util/helper-cairo-user.hh:224:39: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_set_source_rgba (cr, r, g, b, a);
  ~~~~~~~~~~~~~~~~~~~~~               ^
../util/helper-cairo-user.hh:224:36: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_set_source_rgba (cr, r, g, b, a);
  ~~~~~~~~~~~~~~~~~~~~~            ^
../util/helper-cairo-user.hh:224:33: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_set_source_rgba (cr, r, g, b, a);
  ~~~~~~~~~~~~~~~~~~~~~         ^
../util/helper-cairo-user.hh:224:30: warning: implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]
  cairo_set_source_rgba (cr, r, g, b, a);
  ~~~~~~~~~~~~~~~~~~~~~      ^
14 warnings generated.
```

src/OT/glyf/glyf.hh
behdad: You're not missing anything. That's certainly doable to add a specialization call for painting a glyph with a solid color. I suggest doing that later if necessary. What I like currently is how few functions we have and that it unifies COLRv0 and non-COLR tables.

behdad: Or if Matthias wants to add it that sounds good to me.

docs/harfbuzz-sections.txt
```diff
@@ -215,6 +218,66 @@ hb_draw_funcs_t
...
+hb_paint_extend_t
+hb_color_line_get_extend
+hb_paint_linear_gradient_func_t
+hb_paint_radial_gradient_func_t
+hb_paint_sweep_gradient_func_t
+hb_paint_composite_mode_t
+hb_paint_push_group_func_t
+hb_paint_pop_group_func_t
+
+hb_paint_funcs_set_push_transform_func
```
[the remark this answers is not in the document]
matthiasclasen: oh, right. I noticed that and forgot to reorder the paint section

src/hb-common.h
```diff
@@ -897,6 +897,24 @@ HB_EXTERN uint8_t
...
+ *
+ * Note that @height is negative, in coordinate systems that grow up.
+ **/
+typedef struct hb_glyph_extents_t {
+  hb_position_t x_bearing;
+  hb_position_t y_bearing;
+  hb_position_t width;
+  hb_position_t height;
+} hb_glyph_extents_t;
+
```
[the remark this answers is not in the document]
matthiasclasen: did I leave it behind? we should

util/hb-cairo-utils.c
matthiasclasen: I'll have a look

src/OT/glyf/glyf.hh
matthiasclasen: I think the caller can keep track of the current path, do a fill, then add the path to the clip?
Or something like that

matthiasclasen:
> One issue is I think we should pass the font down to all callbacks. The ones needing color need to access the face for palette access, and currently they have to carry the font/face in the paint_data themselves. clip_glyph also needs the font to get the outline (and unscale it).

Passing the font now, still need to to do the other changes

src/OT/glyf/glyf.hh
```diff
@@ -349,7 +349,7 @@ struct glyf_accelerator_t
     funcs->push_root_transform (data, &ctx);
 
     funcs->push_clip_glyph (data, gid, &ctx);
-    funcs->color (data, 0xffff, 1., &ctx);
+    funcs->color (data, hb_paint_get_color (&ctx, 0xffff, 1.), &ctx);
     funcs->pop_clip (data, &ctx);
```
khaledhosny: Hmm, this is not going to work for PDF for me. I output color glyphs to PDF Type 3 glyphs and parts with color index 0xFFFF don’t get any color set and they get the text color when used in PDF text operator i.e. I don’t know the foreground color at the time of writing the layers and the same glyph can be used with different foreground colors.

khaledhosny: I don’t mind resolving the colors internally, but I need a way to tell when foreground color is to be used without having to set it beforehand.

matthiasclasen: How do you do this for gradients?

matthiasclasen: I've added an is_foreground boolean that you can use. If you pass 0,0,0,1 as the foreground color to hb_font_paint_glyph, you can read off the alpha value in the color.

matthiasclasen: I've added an is_foreground boolean that you can use. If you set the foreground color to 0,0,0,255, then you can read the alpha value off the color.

behdad:
One tiny issue left is that painting slanted images is broken currently.

behdad:
> One tiny issue left is that painting slanted images is broken currently.

Fixed.

behdad:
I implemented paint_glyph for hb-ft for outline glyphs and COLRv0. PNG/SVG don't have fetch API in freetype. COLRv1 to be done by a brave soul.

behdad:
Color bitmaps also added for hb-ft using +HB_PAINT_IMAGE_FORMAT_BGRA.

Issue #3925, opened 2022-12-08
[hb-view] Relax build requirements on cairo-ft


behdad:
Currently we don't build hb-view if cairo-ft is not available. With hb-draw, that's not required anymore I believe. Can we test and relax that?

behdad:
I do believe we still require freetype.

khaledhosny:
I think I tried a while ago, it seems that freetype is still needed but probably can be worked out.

behdad:
I'm trying to help @justvanrossum build hb-view on Mac, where he sees cairo as available but hb-view is not built. At least our messaging in meson configure is off as well. My guess is that his cairo lacks cairo-ft.

PR #3941, opened 2022-12-16
[hb-view] Allow building without cairo-ft

khaledhosny:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3925


behdad:
Nice. Thanks.

khaledhosny:
Should we simplify the Windows binary builds and drop cairo-ft dependency there as well (which will allow us to also stop building freetype, fontconfig, and their dependencies on the crossbuild jobs)?

behdad:
> Should we simplify the Windows binary builds and drop cairo-ft dependency there as well (which will allow us to also stop building freetype, fontconfig, and their dependencies on the crossbuild jobs)?

I'm fine with that; since our default is hb-draw anyway.

PR #3942, opened 2022-12-16
COLRv1: Revamp extents variation

matthiasclasen:
Try to do this a cleaner way.

### DIFF
--- a/src/hb-ot-color-colr-table.hh
+++ b/src/hb-ot-color-colr-table.hh
@@ -900,7 +900,7 @@ struct ClipBoxFormat1
     return_trace (c->check_struct (this));
   }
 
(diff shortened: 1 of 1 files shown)


src/hb-ot-color-colr-table.hh
```diff
@@ -895,6 +900,14 @@ struct ClipBoxFormat1
     return_trace (c->check_struct (this));
   }
 
+  void get_clip_box (ClipBoxData &clip_box, const VarStoreInstancer &instancer) const
```
behdad, line 903: This doesn't use the instancer.

behdad: If you need it for API signature, mark it HB_UNUSED.

matthiasclasen: Yes, thats the non-variable implementation of the api. I'll add the unused

PR #3947, opened 2022-12-16
[build] Drop autotools

khaledhosny:


khaledhosny:
LibreOffice is still using autotools to build HarfBuzz and since I’m likely to end up having to deal with dropping autotools on LibreOffice side, it is going to be more work dropping autotools than keeping it.

PR #3971, opened 2022-12-23
Make hb-features.h usable standalone

matthiasclasen:
The intended use for hb-features.h is to
be included standalone, so we can't put
the single-include guards in here.


behdad:
Thanks. Pushed it to main already.

PR #3976, opened 2022-12-25
Add libharfbuzz-cairo

matthiasclasen:
This library will provide integration with cairo for font rendering.


src/hb-cairo.c
```diff
@@ -126,10 +85,10 @@ get_cairo_draw_funcs (void)
   if (!funcs)
   {
     funcs = hb_draw_funcs_create ();
-    hb_draw_funcs_set_move_to_func (funcs, (hb_draw_move_to_func_t) move_to, nullptr, nullptr);
-    hb_draw_funcs_set_line_to_func (funcs, (hb_draw_line_to_func_t) line_to, nullptr, nullptr);
-    hb_draw_funcs_set_cubic_to_func (funcs, (hb_draw_cubic_to_func_t) cubic_to, nullptr, nullptr);
-    hb_draw_funcs_set_close_path_func (funcs, (hb_draw_close_path_func_t) close_path, nullptr, nullptr);
+    hb_draw_funcs_set_move_to_func (funcs, (hb_draw_move_to_func_t) move_to, NULL, NULL);
```
khaledhosny, line 88: Is there a particular reason this a C not C++ file like the rest of HarfBuzz sources?

behdad: Agreed. I'd need to pull hb-machinery.hh at least to make the funcs creations threadsafe for example.

behdad:
This looks good. The only issue we need to figure out is that currently the font scale maps directly to cairo user units. I think we need to add a scale factor there.

src/meson.build
```diff
@@ -576,6 +576,25 @@ libharfbuzz_subset_dep = declare_dependency(
...
 meson.override_dependency('harfbuzz-subset', libharfbuzz_subset_dep)
 
+hb_cairo_sources = [
+  'hb-cairo.c',
+  'hb-cairo-utils.c'
+]
+
+cairo_dep = dependency('cairo')
+
+libharfbuzz_cairo = library('harfbuzz-cairo', hb_cairo_sources,
```
behdad, line 586: Where does it install header files?

matthiasclasen: It was just a quick draft, not a fully worked out branch. If we want this, I'll fix it up

behdad: Yes we definitely want this, given how complicated writing a COLRv1 renderer is.

matthiasclasen:
Not sure whats going on with the non-installed hb-cairo-utils.h header in the autotools build. My autofoo is a bit rusty. @khaledhosny can you take a look?

behdad:
> Not sure whats going on with the non-installed hb-cairo-utils.h header in the autotools build. My autofoo is a bit rusty. @khaledhosny can you take a look?

Humm? I renamed that to .hh

behdad:
I'm feeling good about this now. @khaledhosny what do you think?

src/hb-cairo.cc
```diff
@@ -0,0 +1,867 @@
...
+}
+
+/**
+ * hb_cairo_font_face_create_for_face:
+ * @face: a `hb_face_t`
+ *
+ * Creates a `cairo_font_face_t` for rendering text according
+ * to @face.
+ *
+ * Returns: a newly created `cairo_font_face_t`
```
khaledhosny, line 556: The ownership of the return value is missing.

src/hb-cairo.cc
```diff
@@ -0,0 +1,867 @@
...
+  return user_font_face_create (face);
+}
+
+/**
+ * hb_cairo_font_face_get_face:
+ * @font_face: a `cairo_font_face_t`
+ *
+ * Gets the `hb_face_t` associated with @font_face.
+ *
+ * Returns: (nullable): the `hb_face_t` associated with @font_face
```
khaledhosny, line 574: The ownership of the return value is missing.

src/hb-cairo.cc
```diff
@@ -0,0 +1,867 @@
...
+ * hb_cairo_font_face_create_for_font:
+ * @font: a `hb_font_t`
+ *
+ * Creates a `cairo_font_face_t` for rendering text according
+ * to @font.
+ *
+ * Note that the scale of @font does not affect the rendering,
+ * but the variations and slant that are set on @font do.
+ *
+ * Returns: a newly created `cairo_font_face_t`
```
khaledhosny, line 594: The ownership of the return value is missing.

src/hb-cairo.cc
```diff
@@ -0,0 +1,867 @@
...
+  return cairo_face;
+}
+
+/**
+ * hb_cairo_font_face_get_font:
+ * @font_face: a `cairo_font_face_t`
+ *
+ * Gets the `hb_font_t` that @font_face was created from.
+ *
+ * Returns: (nullable): the `hb_font_t` that @font_face was created from
```
khaledhosny, line 619: The ownership of the return value is missing.

src/hb-cairo.cc
```diff
@@ -0,0 +1,867 @@
...
+  return (hb_font_t *) cairo_font_face_get_user_data (font_face, &hb_cairo_font_user_data_key);
+}
+
+/**
+ * hb_cairo_scaled_font_get_font:
+ * @scaled_font: a `cairo_scaled_font_t`
+ *
+ * Gets the `hb_font_t` associated with @scaled_font.
+ *
+ * Returns: (nullable): the `hb_font_t` associated with @scaled_font
```
khaledhosny, line 635: The ownership of the return value is missing.

src/hb-cairo.cc
```diff
@@ -0,0 +1,867 @@
...
+ * @scale_factor: scale factor to divide hb_positions_t values by
+ * @x: X position to place first glyph
+ * @y: Y position to place first glyph
+ * @utf8: (nullable): the text that was shaped in @buffer
+ * @utf8_len: the length of @utf8 in bytes
+ * @glyphs: (out): return location for an array of `cairo_glyph_t`
+ * @num_glyphs: (out): return location for the length of @glyphs
+ * @clusters: (out) (nullable): return location for an array of cluster positions
+ * @num_clusters: (out) (nullable): return location for the length of @clusters
+ * @cluster_flags: (out) (nullable): return location for cluster flags
```
khaledhosny, line 729: The out arrays are allocated using cairo function, do they need to be deallocated with similar functions as well, if so, then this should be documented.

Issue #3977, opened 2022-12-26
Provide per-glyph info about color


matthiasclasen:
Pango marks glyphs in shaping results as 'color' or not, since GTK wants to treat color glyphs differently.

We can check if a glyph has a png or svg blob, or if it has COLRv0 layers, but for COLRv1 paint, there's no
suitable hb api to check, short or painting the glyph.

Something like

```
hb_bool_t hb_ot_color_glyph_has_paint (hb_face_t *face, hb_codepoint_t glyph)
```
would be sufficient

PR #3978, opened 2022-12-26
Add hb_ot_color_glyph_has_paint

matthiasclasen:
See https://github.com/harfbuzz/harfbuzz/issues/3977

There is an unrelated fix for hb_ot_color_has_paint here too, with test.
I can put that in a separate PR if you prefer.


behdad:
Both this and `hb_ot_color_has_paint` are slightly misleading given that our paint API paints everything...

PR #3979, opened 2022-12-26
[paint] Limit the size of the graph we follow

matthiasclasen:
Instead of just checking the depth, just count
the number of edges in the graph we've followed,
and give up after 256.

### DIFF
--- a/src/hb-ft-colr.hh
+++ b/src/hb-ft-colr.hh
@@ -35,6 +35,14 @@
 
 #ifdef TT_SUPPORT_COLRV1
 
+#ifndef HB_COLRV1_MAX_NESTING_LEVEL
+#define HB_COLRV1_MAX_NESTING_LEVEL     128
+#endif
+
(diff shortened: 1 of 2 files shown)


behdad:
I feel like we should aim for a higher number, like 1024;
Then also keep the depth counter to avoid stack overflows;
Also needs counterpart in hb-ft-colr.

matthiasclasen:
Sure, we can keep the depth check.

matthiasclasen:
Is there a good header to collect these 'arbitrary limit' defines in?

behdad:
> Is there a good header to collect these 'arbitrary limit' defines in?

Let's add hb-limits.hh

behdad:
> > Is there a good header to collect these 'arbitrary limit' defines in?
> 
> Let's add hb-limits.hh

I can do that. I'll merge this and add that after. Thanks!

Issue #3984, opened 2022-12-28
Issues in hb-ft.cc


matthiasclasen:
hb-ft.cc has these checks all over the place:

```
  if (font->destroy != (hb_destroy_func_t) _hb_ft_font_destroy)
    ...
```

But from what I can make out, _hb_ft_font_destroy is never used as font->destroy.

In gdb, I see the following for a font created via hb_ft_face_create_referenced+hb_font_create:

```
(gdb) print font->destroy
$1 = (hb_destroy_func_t) 0x7ffff7cbe232 <_hb_ot_font_destroy(void*)>
```

This is what the tests in https://github.com/harfbuzz/harfbuzz/pull/3696 fail.

Following the trail of _hb_ft_font_destroy, it seems to be passed as destroy in hb_blob_create.

It would probably be better to mark 'ft' fonts and faces more explicitly.

khaledhosny:
This is set by `hb_ft_font_create()` or `hb_ft_font_set_funcs()`, but you are not using either in in #3696.

matthiasclasen:
The documentation explicitly recommends to use hb_ft_font_create_referenced over hb_ft_font_create, and does not mention any side-effects of that choice on hb_ft_font_get_face.

matthiasclasen:
The documentation of hb_ft_font_get_face should be explicit about the fact that it only works with fonts created in just the right way.

khaledhosny:
`hb_ft_font_create_referenced()` also sets it. I think it is a documentation failure. If you are using `hb_font_create()`, then this is not an ft-based font and `hb_ft_font_get_face()` rightly can’t return an `FT_Face` from it.

matthiasclasen:
I see.

hb_ft_face_create_* is really just convenience api to pull a blob out of an FT_Face. The docs could make that clearer.

khaledhosny:
This applies to all `hb_ft_font_*()` functions, they only work on fonts created with `hb_ft_font_create*()`

khaledhosny:
> I see.
> 
> hb_ft_face_create_* is really just convenience api to pull a blob out of an FT_Face. The docs could make that clearer.

Basically, yes.

PR #3985, opened 2022-12-28
[ft] Clarify docs

matthiasclasen:
Spell out what fonts hb_ft_font_get_face() works with.

See https://github.com/harfbuzz/harfbuzz/issues/3984

### DIFF
--- a/src/hb-ft.cc
+++ b/src/hb-ft.cc
@@ -226,6 +226,9 @@ _hb_ft_hb_font_check_changed (hb_font_t *font,
  * For more information, see 
  * https://www.freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_load_xxx
  *
+ * This function works with #hb_font_t objects created by
+ * hb_ft_font_create() or hb_ft_font_create_referenced().
+ *
  * Since: 1.0.5
  **/
 void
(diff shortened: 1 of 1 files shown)


src/hb-ft.cc
```diff
@@ -273,6 +273,9 @@ hb_ft_font_get_load_flags (hb_font_t *font)
  * Fetches the FT_Face associated with the specified #hb_font_t
  * font object.
  *
+ * This function works with #hb_font_t objects created by
+ * hb_ft_font_create() or hb_ft_font_create_referenced().
+ *
```
khaledhosny, line 278: This probably should be copied to the rest of the hb_ft_font* functions, otherwise it might give the feeling that these two are an exception.

matthiasclasen: Yes, I just noticed that. Will do

src/hb-ft.cc
```diff
@@ -1018,6 +1032,10 @@ _hb_ft_reference_table (hb_face_t *face HB_UNUSED, hb_tag_t tag, void *user_data
  *
  * Creates an #hb_face_t face object from the specified FT_Face.
  *
+ * Note that this is using the FT_Face object just to get at the underlying
+ * font data, and fonts created from the returned #hb_face_t will use the native
+ * HarfBuzz font implementation.
```
behdad, line 1037: Can you mention calling `hb_ft_font_set_funcs()` on them?

PR #3986, opened 2022-12-28
Test native api

matthiasclasen:
Here are a few tests for what I would consider the mvp level for the native api.


matthiasclasen:
I've left out the directwrite api, since the implementation looks much less baked than ft or coretext.

PR #3992, opened 2022-12-31
[static] Remove HB_NO_VISIBILITY conditional

behdad:
That was just wrong. No idea how it worked. Let's see what bots break...

### DIFF
--- a/src/hb-static.cc
+++ b/src/hb-static.cc
@@ -38,7 +38,6 @@
 #include "hb-ot-head-table.hh"
 #include "hb-ot-maxp-table.hh"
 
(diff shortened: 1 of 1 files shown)


behdad:
Ah okay, there's this in hb-common.cc:
```
/* If there is no visibility control, then hb-static.cc will NOT 
 * define anything.  Instead, we get it to define one set in here 
 * only, so only libharfbuzz.so defines them, not other libs. */ 
#ifdef HB_NO_VISIBILITY 
#undef HB_NO_VISIBILITY 
#include "hb-static.cc" 
#define HB_NO_VISIBILITY 1 
#endif 
```

Issue #3999, opened 2023-01-02
Consider adding hb_language_get_scripts


matthiasclasen:
I am considering replacing PangoLanguage with hb_language_t.

We've already replaced PangoScript with GUnicodeScript, so dropping PangoLanguage
would get pango out of the Unicode data game for good.

There's a few apis that we need, though. One is a language <> script mapping. Pango has:

```
const PangoScript *  pango_language_get_scripts  (PangoLanguage  *language,
                                                  int            *num_scripts);
```
A HarfBuzz equivalent could be
```
unsigned int
hb_language_get_scripts (hb_language_t language,
                         unsigned int *n_scripts,
                         hb_script_t *scripts);
```

Another API that pango has is sample texts for languages:

```
const char *
pango_language_get_sample_string (PangoLanguage *language);
```
But I'm not asking for that to be included in HarfBuzz, currently.

PR #4001, opened 2023-01-02
Add hb_language_get_scripts

matthiasclasen:
See https://github.com/harfbuzz/harfbuzz/issues/3999

This is a copy of the Pango implementation, minus caching.

Tests included.


src/hb-script-lang-table.h
```diff
@@ -0,0 +1,253 @@
...
+  { "it",       { HB_SCRIPT_LATIN } },
+  { "iu",       { HB_SCRIPT_CANADIAN_SYLLABICS } },
+  { "ja",       { HB_SCRIPT_HAN, HB_SCRIPT_KATAKANA, HB_SCRIPT_HIRAGANA } },
+  { "jv",       { HB_SCRIPT_LATIN } },
+  { "ka",       { HB_SCRIPT_GEORGIAN } },
+  { "kaa",      { HB_SCRIPT_CYRILLIC } },
+  { "kab",      { HB_SCRIPT_LATIN } },
+  { "ki",       { HB_SCRIPT_LATIN } },
+  { "kj",       { HB_SCRIPT_LATIN } },
+  { "kk",       { HB_SCRIPT_CYRILLIC } },
```
khaledhosny, line 107: What is the source of truth for this list and how do we keep it up to date (changes are rare but do happen, for example [Kazakh is witching to Latin, presumably staring this year](https://en.wikipedia.org/wiki/Kazakh_alphabets#Latest_developments).

src/hb-script-lang-table.h
```diff
@@ -0,0 +1,253 @@
...
+  { "mh",       { HB_SCRIPT_LATIN } },
+  { "mi",       { HB_SCRIPT_LATIN } },
+  { "mk",       { HB_SCRIPT_CYRILLIC } },
+  { "ml",       { HB_SCRIPT_MALAYALAM } },
+  { "mn-cn",    { HB_SCRIPT_MONGOLIAN } },
+  { "mn-mn",    { HB_SCRIPT_CYRILLIC } },
+  { "mni",      { HB_SCRIPT_BENGALI } },
+  { "mo",       { HB_SCRIPT_CYRILLIC, HB_SCRIPT_LATIN } },
+  { "mr",       { HB_SCRIPT_DEVANAGARI } },
+  { "ms",       { HB_SCRIPT_LATIN } },
```
khaledhosny, line 145: Malay is often written in Arabic script (AKA [Jawi](https://en.wikipedia.org/wiki/Jawi_script), but I’m not sure what language tag is used for this.

src/hb-common.cc
```diff
@@ -665,6 +666,125 @@ hb_script_get_horizontal_direction (hb_script_t script)
...
+}
+
+/**
+ * hb_language_get_scripts:
+ * @language: a #hb_language_t
+ * @script_count: (inout) (optional): Input = the maximum number of scripts to return;
+ *      Output = the actual number of scripts returned (may be zero)
+ * @scripts: (out) (array length=script_count): the array of #hb_script_t found
+ *
+ * Fetches the scripts that can be used with @language.
```
khaledhosny, line 743: This is a bit vague, since languages can practically be written in any any given script. Also I’m not sure what this API is useful for.

matthiasclasen: > This is a bit vague, since languages can practically be written in any any given script. Also I’m not sure what this API is useful for.

My immediate use is finding a direction from a language.

src/hb-script-lang-table.h
matthiasclasen: > What is the source of truth for this list and how do we keep it up to date 

Pango ships a C program that uses fontconfigs charset information for each language, and then looks at the script for each Unicode character in the charset. I didn't include that here since, fontconfig is not a dependency of HarfBuzz, and, as you say, this rarely changes.

src/hb-common.cc
khaledhosny: I don’t follow. Direction is a property of the characters (i.e. from UCD and BiDi algorithm), when does language play a role?

behdad: Pango goes from language tag -> script -> direction with this API.

matthiasclasen: When you show a UI in a given locale, and you need to decide whether to put the sidebar on the left or on the right.

behdad: For the main layout direction of a language. We have `hb_script_get_horizontal_direction`.

matthiasclasen: you could of course cut out the script here, and just provide a hb_language_get_horizontal_direction()

khaledhosny: That sounds like a rather odd way to get the direction of a locale (and locales != languages). IIRC, GTK had a special string you translate to LTR or RTL to set the UI direction for a given translation. But also it feels very pango-speficic and not of much value to anyone else and I don’t see why Pango can’t keep using the code it has even after switching to `hb_language_t` as it does seem to be using any insider HarfBuzz knowledge about the langues here.

matthiasclasen: We want to get out of abusing translations for this. Not the least because it turns out that random other projects open gtk.mo just to pull out the translation for default:LTR

matthiasclasen: And I want to get rid of the code in pango, because I am the only one who ever touches pango, whereas there are multiple contributors here.

behdad: Yeah I find language<->script to be generally useful, plus also script_get_prominent_language(). These, of course as Khaled says, don't have any use in HarfBuzz itself, but are useful to higher level specially to itemization and font selection layers.

khaledhosny: I don’t know, but language feels like the wrong level for this information to me. Besides, what does this API return for, say, `ms-Arab` or `sr-Latn`?

behdad: Our language tags *are* BCP 47. So we can make it understand those.

khaledhosny: I still don’t like this, but at least give it some less-confident name, like `hb_langauge_guess_script()` or some such and make it clear in the documentation this is a best guess (not that it is going to stop any one, people still use `hb_buffer_guess_segemet_properties()` in lieu of proper segmentation).

khaledhosny: There is also [uppress-Script Field](https://www.rfc-editor.org/rfc/rfc5646.html#section-3.1.9) of the [IANA Language Subtag Registry](http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry). At this point, one might as well use https://github.com/tagoh/liblangtag.

matthiasclasen: Other use cases for language <> script relation that I find: Pango gets text that is marked up with a language, and wants to ensure that the language matches the script of the text, to avoid nonsensical situations.

matthiasclasen: > I don’t know, but language feels like the wrong level for this information to me.

You mean the direction, here?

khaledhosny: > Other use cases for language <> script relation that I find: Pango gets text that is marked up with a language, and wants to ensure that the language matches the script of the text, to avoid nonsensical situations.

But is this necessary? Many languages are written in scripts they are not predominantly written in (either historically or by minority or diaspora communities), what languages should these texts be tagged in. What kind of issues Pango is preventing by filtering out unlikely languages? Shaping depends on the script and language influence is usually minimal.

khaledhosny: > > I don’t know, but language feels like the wrong level for this information to me.
> 
> You mean the direction, here?

I mean getting direction from the language is the wrong level.

matthiasclasen: > But is this necessary? 

I don't know. This is old code that I did not write:

```
static PangoLanguage *
compute_derived_language (PangoLanguage *lang,
                          PangoScript    script)
{
  PangoLanguage *derived_lang;

  /* Make sure the language tag is consistent with the derived
   * script. There is no point in marking up a section of
   * Arabic text with the "en" language tag.
   */
  if (lang && pango_language_includes_script (lang, script))
    derived_lang = lang;
  else
    {
      derived_lang = pango_script_get_sample_language (script);
      /* If we don't find a sample language for the script, we
       * use a language tag that shouldn't actually be used
       * anywhere. This keeps fontconfig (for the PangoFc*
       * backend) from using the language tag to affect the
       * sort order. I don't have a reference for 'xx' being
       * safe here, though Keith Packard claims it is.
       */
      if (!derived_lang)
        derived_lang = pango_language_from_string ("xx");
    }

  return derived_lang;
}
```

khaledhosny: IMHO, that feels like a solution in search of a problem.

khaledhosny: Reading https://bugzilla.gnome.org/show_bug.cgi?id=123581, it seems the root of the issue is that Pango defaults to the locale language which might be pointless but harmless, except I guess for FontConfig, but I don’t know what FontConfig does with the language so I don’t know how bad is it. Also it is not making a distinction between explicitly set language (which should be unconditionally taken as is) and the locale-driven one. Still it feels like a very Pango-speficic problem.

matthiasclasen: I could be convinced that pango should just take languages as-is. That will probably be less of a concern if language does not affect font selection ordering anymore.

behdad: > But is this necessary?

For using fontconfig properly it's necessary to use this API, because of fonts like DejaVu which eg. have Arabic glyphs but are not deemed suitable for Persian (low quality). So I made this complex machinery in Pango, using this API, such that if you set LANGUAGES=en:fa, then any Arabic-script text will use the preferred font for 'fa'.

Without the API proposed here that would not be possible, because I need to know that Arabic script is not used to write en, to pass fa to fontconfig.

If I pass both en,fa in that order to fontconfig, it will return DejaVu, which is what I want to avoid.

khaledhosny: I still feel it is a shoot yourself in the foot-kind API but I understand there is a need and I don’t want to block you, so feel free to finish and merge this as you see fit.

src/hb-common.cc
```diff
@@ -665,6 +666,125 @@ hb_script_get_horizontal_direction (hb_script_t script)
...
+
+  if (language == nullptr)
+    return nullptr;
+
+  lang_str = language->s;
+
+  records = (const char *) hb_script_for_lang;
+  record_size = sizeof (HbScriptForLang);
+  num_records = sizeof (hb_script_for_lang) / record_size;
+  record = (const char *) bsearch (lang_str,
```
ebraminio, line 704: I think we were done with libc's bsearch once since #1602, guess hb_bsearch can be used here?

src/hb-common.cc
```diff
@@ -665,6 +666,125 @@ hb_script_get_horizontal_direction (hb_script_t script)
...
+  /* find the best match among all those that have the same first-component */
+
+  /* go to the final one matching in the first component */
+  while (record < end - record_size &&
+         lang_compare_first_component (lang_str, record + record_size) == 0)
+
+  /* go back, find which one matches completely */
+  while (start <= record &&
+         lang_compare_first_component (lang_str, record) == 0)
+    {
```
ebraminio, line 723: Inconsistent indention vs rest of the project it seems

src/hb-common.cc
```diff
@@ -665,6 +666,125 @@ hb_script_get_horizontal_direction (hb_script_t script)
...
+find_best_lang_match (hb_language_t language)
+{
+  const char *lang_str;
+  const char *record, *start, *end;
+  const char *records;
+  unsigned int record_size;
+  unsigned int num_records;
+
+  if (language == nullptr)
+    return nullptr;
```
ebraminio, line 697: guess this can be moved to top so assignments can be merged with declarations as we are in C++?

src/hb-common.cc
```diff
@@ -665,6 +666,125 @@ hb_script_get_horizontal_direction (hb_script_t script)
...
+  for (i = 0; i < sizeof (script_for_lang->scripts) / sizeof (hb_script_t); i++)
+    if (script_for_lang->scripts[i] == 0)
+      break;
+
+  count = i;
+
+  if (script_count)
+  {
+    for (i = 0; i < count; i++)
+      {
```
ebraminio, line 777: inconsistent indention vs rest of the project it seems

khaledhosny:
Inspired by this discussion, I think I learnt about a Pango limitation: language setting is ignored if the script is not one of what it considers language scripts. Compare `hb-view` output:
```
$ hb-view Amiri-Regular.ttf "ملایو، جاوی" --language=ms
```
hb

with `pango-view` output:
```
$ pango-view --font="Amiri 200" --text="ملایو، جاوی" --language=ms
```

pango

(The font has substitutions rules for OpenType `MLY` langauge tag in `arab` script, but Pango foes not apply it)

behdad:
> Inspired by this discussion, I think I learnt about a Pango limitation: language setting is ignored if the script is not one of what it considers language scripts.

Sure. It is made to work for cases that eg. you set LANGUAGE=en,fa as well, so it was to drop `en` and set lang for Arabic script to `fa`. So it has to rely on that script-per-language knowledge.

khaledhosny:
> > Inspired by this discussion, I think I learnt about a Pango limitation: language setting is ignored if the script is not one of what it considers language scripts.
> 
> Sure. It is made to work for cases that eg. you set LANGUAGE=en,fa as well, so it was to drop `en` and set lang for Arabic script to `fa`. So it has to rely on that script-per-language knowledge.

Yes, I think the issue here is that it does not distinguish between explicitly set language (which should be assumed to be correct and intentional) and the one(s) deduced from the locale.

khaledhosny:
> > Inspired by this discussion, I think I learnt about a Pango limitation: language setting is ignored if the script is not one of what it considers language scripts.
> 
> Interesting. Maybe worth a pango issue

https://gitlab.gnome.org/GNOME/pango/-/issues/723

ilius:
I have also extracted a list of languages with script info, mostly based on wikipedia, and it seems to contain more languages (but it has no country codes).
https://github.com/ilius/pyglossary/blob/master/pyglossary/langs/langs.json

simotek:
I'd find this API useful, mostly because in the language I am using we can use optional parameters and in cases where people are lazy and the guess is right they can then pass less parameters to the function.

behdad:
Lets close for now. When and if we add an itemizer API, we'll add this back if necessary.

PR #4009, opened 2023-01-05
docs: Fix a typo

matthiasclasen:
Here's a typo noticed in passing.

And a question: Should you not empty out that static_unicodes set to avoid accumulating memory there?


behdad:
> And a question: Should you not empty out that static_unicodes set to avoid accumulating memory there?

C++ destructor takes care of it.

behdad:
Sorry, should be called stack_unicodes.

PR #4015, opened 2023-01-08
Add more paint tests

matthiasclasen:
These tests compare the output of the ft and ot
implementations for all the glyphs in the test_glyphs font.


behdad:
This doesn't use hb-ft-funcs. See my recent change to test-paint.

matthiasclasen:
> This doesn't use hb-ft-funcs. See my recent change to test-paint.

Made that change. Now need to fix tests to work again, like all the others.

PR #4017, opened 2023-01-08
test-paint: Limit the precision of clip boxes

matthiasclasen:
This is an attempt to work around rounding differences seen in some of the ci runs, for clipboxes.


matthiasclasen:
Looking into the test failure in ci, I notice that the glyph that fails is one that has skew transforms and no clipbox, so we get the computed extents, and for some reason, on the circleci alpine runner, they differ from what I get locally, in the 4the decimal.

PR #4018, opened 2023-01-09
Draft: Add a draw test

matthiasclasen:
This test compares output between ft and ot font funcs.

It fails. I don't know if that is expected or not, but this is the core of the differences
between paint extent computation for ft and ot.

I believe the difference is due to FreeType scaling the outlines.


matthiasclasen:
Removing the `hb_font_set_scale (font, 20, 20)` makes the comparison test succeed.

Issue #4020, opened 2023-01-10
Range selectors for ligatures does not work properly with aat fonts applying ligatures via morx table


behdad:
### Discussed in https://github.com/harfbuzz/harfbuzz/discussions/4019

<div type='discussions-op-text'>

<sup>Originally posted by **DeependraKhetwal** January  9, 2023</sup>
Hi , I am facing an issue with aat fonts where on disabling ligatures for a particular range the ligatures gets disabled for the whole line. The issue does not happen with fonts that apply ligature substitutions via opentype 'liga' feature but which apply ligature substitutions via 'morx' table. In the attached example I am disabling ligatures for 'ffi' (using -liga[3:])  but ligature is getting turned off for 'ff' also.
<img width="571" alt="Screenshot 2022-12-05 at 2 02 42 PM (1)" src="https://user-images.githubusercontent.com/122281418/211316367-7a7ceb8f-e6cf-4078-937d-650481fad787.png">
</div>

behdad:
To implement this properly we need to treat runs with different features as independent runs for running the state machine, as the subtable flags might be different. That would be a significant change to our internal implementation.

I have a hacky implementation instead, which seems to work for simple cases. I'm going to push that instead.

behdad:
That broke tests. Reverting while I investigate whether it can be fixed.

behdad:
Okay. "Fixed" it. I really don't like this hack, but I think it will do the job for now. It doesn't have any adverse effect if user doesn't specify any features.  For fully specified features I might have broken things. To be determined.

behdad:
Humm, actually my change probably breaks some fonts. I'm uncomfortable with it. Let me think more. Might revert it.

behdad:
Okay. I fixed it properly. Please test @DeependraKhetwal.

behdad:
Humm. It broke more things. Debugging still :(

PR #4023, opened 2023-01-10
[aat] Support feature ranges

behdad:


behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/4020

behdad:
Hi @jfkthame. I'm confident in this implementation finally. Any chance you can take a look and test it a bit?

What it does basically is that it moves creating an aat-map from planning time to shaping time, and if any user-features are specified, then the subtable matching is moved to the state-machine driving time and if the subtable flags don't match the flags for the range, we just reset the state machine state and skip the glyphs for the range. It doesn't try to flush other state-machine states though. Seems to work.

PR #4027, opened 2023-01-11
Plan simplify

behdad:


behdad:
I've done most of these now. A few are left. I'm happy to merge as is.

garretrieger (approved):

PR #4028, opened 2023-01-11
[instancer] Add MVAR values to OS/2

behdad:
Start of applying MVAR during instancing. Untested, and our test suite doesn't seem to hit it either.


behdad:
cc @madig

behdad:
Needs to be done: hhea and vhea.

behdad:
I tested this with AdobeVFPrototype and the OS/2 part seems to work.

behdad:
This is complete sans the gasp table, which I don't think we subset much.

behdad:
We need to add tests.

garretrieger (approved):

PR #4033, opened 2023-01-14
[instance] drop and don't collect lookups from feature variations when fully instancing.

garretrieger:
This previously incorrectly collected lookups that could be reached via feature variations that are dropped and not activated by the current instance position.


qxliu76:
LGTM, thanks!

PR #4034, opened 2023-01-16
hb-info: Print table version where relevant

matthiasclasen:
When listing tables with --list-tables, show
the version for CPAL and COLR tables.


behdad:
This feels crude. Let me think about it.

behdad:
I feel like closing this, if you don't mind.

PR #4040, opened 2023-01-17
[ci] Try using ccache

khaledhosny:


khaledhosny:
> Is the cache retained across runs?

It should be.

khaledhosny:
This shaves-off a few minutes from each GHA job, not bad. Might do CircleCI jobs later.

behdad:
Thanks Khaled.

PR #4041, opened 2023-01-17
Revert "[ci] Use sccache with msys2"

khaledhosny:


khaledhosny:
OK, so this seems to be the same old mysterious msys2 Python crash (https://github.com/harfbuzz/harfbuzz/pull/3903), except that the workaround no longer works. I don’t know what relation it has with using sccache, but I guess I’ll revert b9646dfd6290dbf7819cc042bb6f541b80ef8b68.

PR #4046, opened 2023-01-19
[doc] Try to fix generating hb-features docs

khaledhosny:
Move it to a separate section, since hb-common says include hb.h, while we want to include hb-features.h here.

This still does not fix generating documentation of undefined macros (e.g. HB_HAS_GDI since we build docs on Linux).


khaledhosny:
> This still does not fix generating documentation of undefined macros (e.g. HB_HAS_GDI since we build docs on Linux).

No idea how Cairo manages this. When I build its docs manually it behaves like this, but its onlline documentation has all `CAIRO_HAS_*` macros documented.

khaledhosny:
I tried explicitly outputting `#undef HB_HAS_*`, but this is not convincing gtk-doc to generate the documentation either.

behdad:
> I tried explicitly outputting `#undef HB_HAS_*`, but this is not convincing gtk-doc to generate the documentation either.

I like explicitly outputting those as it makes it easier for clients to see what symbols are there.

khaledhosny:
> > I tried explicitly outputting `#undef HB_HAS_*`, but this is not convincing gtk-doc to generate the documentation either.
> 
> I like explicitly outputting those as it makes it easier for clients to see what symbols are there.

They are currently written to the file, but commented out.

matthiasclasen:
> No idea how Cairo manages this. When I build its docs manually it behaves like this, but its onlline documentation has all `CAIRO_HAS_*` macros documented.

Are you sure cairo actually manages? I see the online docs have all defines, but when I locally generate cairo docs, I don't get the WIN32 defines in the docs. Maybe some secret sauce was lost in the transition to meson?

One way I could see going about this is to somehow make gtk-doc see a doctored hb-features.h file that has all the defines in it.

khaledhosny:
> > No idea how Cairo manages this. When I build its docs manually it behaves like this, but its onlline documentation has all `CAIRO_HAS_*` macros documented.
> 
> Are you sure cairo actually manages? I see the online docs have all defines, but when I locally generate cairo docs, I don't get the WIN32 defines in the docs. Maybe some secret sauce was lost in the transition to meson?
> 
> One way I could see going about this is to somehow make gtk-doc see a doctored hb-features.h file that has all the defines in it.

Seems so. [Makefile.sources](https://gitlab.freedesktop.org/cairo/cairo/-/blob/b43e7c6f3cf7855e16170a06d3a9c7234c60ca94/src/Makefile.sources#L30-35) (gone with autotools), suggests that a special `cairo-supported-features.h` was generated for gtk-doc.

behdad:
Should we add Since: tags here?

khaledhosny:
> Should we add Since: tags here?

I thought about this before, but since one checks for them using ifdef it probably doesn't matter and might even be misleading. A version check begore checking the macro would be superfluous.

behdad:
> A version check begore checking the macro would be superfluous.

I wouldn't expect people to do that. But it does help them know that these are very recent and won't work with older HB.

Doesn't matter really one way or the other I suppose.

khaledhosny:
I think this works now.

PR #4048, opened 2023-01-19
hb-info

behdad:


behdad:
TODO:
- [ ] Add `--list-scripts` and separate that part from `--list-features`.
- [ ] Add `--script` and `--language` and if those are specified, make `--list-features` respect those.
- [ ] Add `--list-names`.
- [ ] Add `--family`, `--style`, `--full-name`, `--postscript-name`, `--font-version`.
- [ ] Add `--dump-table`.
- [ ] Add `--optical-size`?
- [ ] Add something for `cvXX` info?

behdad:
I've added a `--list-all` as well. However, that's too verbose to be useful. Maybe that should be `--show-all` and skip showing charset info.

There's generally a tension between making the output machine-parsable, versus human-readable. Ie. if you ask for upem, I like to just print a number, and not eg. `upem: 2048`. With the former, then in a `--show-all`, values would have no description. `otfinfo` gets around this by having `--info` that generates a human-readable output of select fields.

khaledhosny:
May be have to modes, one human readable and the other is machine readable? It might be too much work, but if we go there, we might as well go all the way and make the machine readable output json or something.

behdad:
> May be have to modes, one human readable and the other is machine readable? It might be too much work, but if we go there, we might as well go all the way and make the machine readable output json or something.

I like the two modes. I can add `--output-format` and a json can come later. Too much work though.

Or for now just add `--quiet`. Add `--output-format` later.

khaledhosny:
I’d go with `--output-format` and leave the door open for a `json` format to be added later, but IMHO the `text` (human readable) format is more important.

behdad:
> I’d go with `--output-format` and leave the door open for a `json` format to be added later, but IMHO the `text` (human readable) format is more important.

That can always be added. I added `--quiet`, which can become an alias for `--output-format=bare` or something in the future.

matthiasclasen:
Looks like a very useful too.

Small suggestions:

For --list-all, it would be good to have clear separation between the lists. Maybe -----------, or just two empty lines
> Should we prefix --upem etc with --show-upem?

We could try to be systematic and have all the output-requesting options be either --show-foo or --list-bar, depending on whether it is a single item or a list of things

behdad:
> Looks like a very useful tool.
> 
> Small suggestions:
> 
> For --list-all, it would be good to have clear separation between the lists. Maybe -----------, or just two empty lines

Two empty lines is easier. I added one empty line and a description. Will make it two.

> > Should we prefix --upem etc with --show-upem?
> 
> We could try to be systematic and have all the output-requesting options be either --show-foo or --list-bar, depending on whether it is a single item or a list of things

Yeah that's what I was leaning towards.

behdad:
I'll go with `---` separator so it's searchable.

behdad:
> * [ ]  Add `--list-scripts` and separate that part from `--list-features`.
> * [ ]  Add `--script` and `--language` and if those are specified, make `--list-features` respect those.

For these there's a question of whether to take and report OpenType tags, or ISO / BCP. I'm leaning towards the latter though I know Khaled probably likes the former.

khaledhosny:
Going with OpenType tags is tempting and also because that is what `--list-features` (future `--list-scripts`) reports. Or we can have both 😄

behdad:
> Going with OpenType tags is tempting and also because that is what `--list-features` (future `--list-scripts`) reports. Or we can have both smile

Both :D.

behdad:
> Going with OpenType tags is tempting and also because that is what `--list-features` (future `--list-scripts`) reports. Or we can have both smile

I was going to change the `--list-features` / `--list-scripts` to convert back as well.

khaledhosny:
On the other hand, people might (reasonably) expect `--script` and `--language` to take the same input as the other tools, so indeed both 😄

khaledhosny:
> > Going with OpenType tags is tempting and also because that is what `--list-features` (future `--list-scripts`) reports. Or we can have both smile
> 
> I was going to change the `--list-features` / `--list-scripts` to convert back as well.

It should show both then, one might need to know if this is `deva`, `dev2`, or `dev3`.

khaledhosny:
```
$ ./util/hb-info ../../../aliftype/mada/fonts/Mada.ttf --list-variations
Face variations information:

Varitation axes:

Tag:	Minimum	Default	Maximum	Name

wght:	200	400	900	Weight
Named instances:

0. ExtraLight                      200,
1. Light                           300,
2. Regular                         400,
3. Medium                          500,
4. SemiBold                        600,
5. Bold                            700,
6. ExtraBold                       800,
7. Black                           900,
```

May be remove the new line after `Tag:	Minimum	Default	Maximum	Name` and add a new line before `Named instances:`?

behdad:
> ```
> 
> ```
> 
> 
>     
>   
> 
> May be remove the new line after `Tag:	Minimum	Default	Maximum	Name` and add a new line before `Named instances:`?

Done.

behdad:
> May be remove the new line after `Tag: Minimum Default Maximum Name`

Do you suggest I remove the same newline for `--list-tables` and others?

behdad:
Or I can add `-----------------------------------`

behdad:
I think the tool should default to `--show-all` (which just shows the short info). I'm not sure how to do that with GOptions. Will poke around.

behdad:
> I think the tool should default to `--show-all` (which just shows the short info). I'm not sure how to do that with GOptions. Will poke around.

Done.

khaledhosny:
> > May be remove the new line after `Tag: Minimum Default Maximum Name`
> 
> Do you suggest I remove the same newline for `--list-tables` and others?

That would be my preference, yes. `-----------------------------------` would be fine, too.

util/hb-info.cc
```diff
@@ -46,6 +46,7 @@ struct info_t
       {"show-glyph-count",0, 0, G_OPTION_ARG_NONE,	&this->show_glyph_count,	"Show glyph count",		nullptr},
 
       {"list-all",	0, 0, G_OPTION_ARG_NONE,	&this->list_all,		"List all long information",	nullptr},
+      {"list-names",	'l', 0, G_OPTION_ARG_NONE,	&this->list_names,		"List names",			nullptr},
```
khaledhosny, line 49: I think `'l'` was not meant to be here.

util/hb-info.cc
```diff
@@ -190,12 +194,37 @@ struct info_t
...
+    const auto *entries = hb_ot_name_list_names (face, &count);
+    for (unsigned i = 0; i < count; i++)
+    {
+      char name[128];
+      unsigned name_len = sizeof name;
+      hb_ot_name_get_utf8 (face, entries[i].name_id,
+			   language,
+			   &name_len, name);
+
+      printf ("%u	%s\n", entries[i].name_id, name);
```
khaledhosny, line 256: It would be nice to separate-out the pre-defined name IDs and add a human name next to the ID.

behdad: That's too much work. I'm adding separate `--show-family` etc, which will be in the default output.

behdad: Same thing with feature names, script names, etc. We don't have that information accessible.

khaledhosny: It can be something like (assuming this is valid C code 😄 :
```C
char* pre_defined_names[] = {
  "Copyright",
  "Family",
   ...
};

if (entries[i].name_id < sizeof (pre_defined_names))
  printf ("%s (%u):	%s\n", entries[i].name_id, name);
else
  printf ("%u	%s\n", entries[i].name_id, name);
```

util/hb-info.cc
```diff
@@ -50,8 +50,8 @@ struct info_t
       {"show-glyph-count",0, 0, G_OPTION_ARG_NONE,	&this->show_glyph_count,	"Show glyph count",		nullptr},
 
       {"list-all",	0, 0, G_OPTION_ARG_NONE,	&this->list_all,		"List all long information",	nullptr},
-      {"list-names",	'l', 0, G_OPTION_ARG_NONE,	&this->list_names,		"List names",			nullptr},
-      {"list-tables",	'l', 0, G_OPTION_ARG_NONE,	&this->list_tables,		"List tables",			nullptr},
+      {"list-names",	0, 0, G_OPTION_ARG_NONE,	&this->list_names,		"List names",			nullptr},
+      {"list-tables",	0, 0, G_OPTION_ARG_NONE,	&this->list_tables,		"List tables",			nullptr},
```
khaledhosny, line 54: I thought you want to keep it for `--list-tables`

behdad: Err. RIght. Sorry.

util/hb-info.cc
khaledhosny: > Same thing with feature names, script names, etc. We don't have that information accessible.

This list is shorter and tags are memorable unlike numeric IDs, but I won’t insist.

behdad:
`--list-shapers` or something to list shaping technologies supported. Although it sucks that we don't have `--shaper=aat` in our other tools.

behdad:
Bikeshed: `--show-font-version` or `--show-version`? So far I've avoided `font` in much of options or their help strings.

behdad:
I've added `--show-*` commands for name IDs 0 to 6. Any more requests?

khaledhosny:
Wild idea, support `--variations` and make it alter the relevant data.

May be it already does this. I was thinking of its effect on `--show-postscript-name`, not sure what other data currently shown would be affected.

behdad:
> Wild idea, support `--variations` and make it alter the relevant data.

Already in there. Will vary the font extents and font metrics.

behdad:
> May be it already does this. I was thinking of its effect on `--show-postscript-name`, not sure what other data currently shown would be affected.

Right. I should adjust `--show-postscript-name` and --show-style` to take it from `--named-instance` if specified. And `--show-postscript-name` to generate name according to Adobe tech report when we have that.

behdad:
> Right. I should adjust `--show-postscript-name` and `--show-style` to take it from `--named-instance` if specified.

Done.

PR #4063, opened 2023-01-23
hb-info: Only build if we have hb-gobject

matthiasclasen:
This was using the wrong condition.


khaledhosny:
I think we want to build without gobject, as the parts that depend on it are ifdef'd.

PR #4072, opened 2023-01-25
Bug fix

qxliu76:
1. avoid freeing compile glyph bytes if serialization failed.
2.  increase HB_GLYF_MAX_POINTS limit to 20000 cause the test file includes a .notdef glyph which is composite and has 10,176 points after get_points() call


behdad:
> 2\. increase HB_GLYF_MAX_POINTS limit to 20000 cause the test file includes a .notdef glyph which is composite and has 10,176 points after get_points() call

is the test file legit though?

qxliu76:
> > 2. increase HB_GLYF_MAX_POINTS limit to 20000 cause the test file includes a .notdef glyph which is composite and has 10,176 points after get_points() call
> 
> is the test file legit though?

yes it's legit, full font file is here: https://github.com/google/fonts/tree/main/ofl/handjet, I subsetted it down to a few characters

behdad:
> > > 2. increase HB_GLYF_MAX_POINTS limit to 20000 cause the test file includes a .notdef glyph which is composite and has 10,176 points after get_points() call
> > 
> > 
> > is the test file legit though?
> 
> yes it's legit, full font file is here: https://github.com/google/fonts/tree/main/ofl/handjet, I subsetted it down to a few characters

I see a big black box. Not sure why it needs over 10,000 points for that.

behdad:
I see. It's a pixel font...

behdad:
Weird that the `pixel` glyph in that font has so many points...

behdad:
> Weird that the `pixel` glyph in that font has so many points...

Ah, it's because of variations.

src/OT/glyf/glyf.hh
```diff
@@ -98,7 +98,9 @@ struct glyf
       use_short_loca = max_offset < 0x1FFFF;
     }
 
-    glyf_prime->serialize (c->serializer, glyphs.writer (), use_short_loca, c->plan, font);
+    if (!glyf_prime->serialize (c->serializer, glyphs.writer (), use_short_loca, c->plan, font))
```
garretrieger, line 101: serialization could fail part way through which means some of the glyphs will have allocated memory but not all, so we still need to run the loop to free memory, but will need a way to track which SubsetGlyphs have allocated memory and which are references.

qxliu76: yes I missed that. I'll work on it.

PR #4074, opened 2023-01-25
Glyph name docs

matthiasclasen:


src/hb-font.cc
```diff
@@ -1343,6 +1343,9 @@ hb_font_get_glyph_contour_point (hb_font_t      *font,
  *
  * Fetches the glyph-name string for a glyph ID in the specified @font.
  *
+ * According to the OpenType specification, glyph names are limited to 63
```
khaledhosny, line 1346: This could make use of a link to the relevant spec page.

PR #4077, opened 2023-01-25
[fonts] move OFL-1.1 license to a higher directory

dilinger:
There's a bunch of font directories inside of test/ for which the vast majority of fonts are licensed under the SIL open font license. We currently have a COPYING file in test/shape/data/in-house/COPYING that says that most of the fonts are OFL-1.1, but that doesn't apply to the fonts in, say, test/api/fonts/ or test/fuzzing/fonts/. Since there are so many OFL-1.1 fonts all over test, let's move the COPYING file to the top-level test/ directory.


behdad:
This needs to update the Makefile.am of the respective directories.

@khaledhosny look like meson is not distributing this COPYING file?

dilinger:
Oops, sorry about that!

PR #4081, opened 2023-01-25
[COPYING] Another update

dilinger:
This brings the COPYRIGHT file up to date, looking more like what I ended up filling out in Debian's debian/copyright file (the main difference being that the decopy script put names into alphabetical order, which I'm happy to do for harfbuzz if you'd like).

Adobe, Inc has copyright in src/hb-subset-cff*, test/api/test-subset*, and misc other places.

Ebrahim Byagowi has copyright as far back as 2015 in places like src/hb-directwrite.cc.

Google, Inc has newer copyright into 2022 in places like src/graph/test-classdef-graph.cc. Also, listing every year was getting a bit unwieldy, so just do 2010-2022.

Igalia S.L. contributed the stuff in src/hb-ot-math*.

The only references I could find to Martin Hosken & SIL were in src/hb-graphite2*, and they were 2011, not 2009.

Mozilla's got a bunch of 2015 code in src/hb-ot-shaper-*.

Red Hat has copyright up to 2023 (eg, test/api/test-glyph-names.c).

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -2,21 +2,23 @@ HarfBuzz is licensed under the so-called "Old MIT" license.  Details follow.
 For parts of HarfBuzz that are licensed under different licenses see individual
 files names COPYING in subdirectories where applicable.
 
-Copyright © 2010,2011,2012,2013,2014,2015,2016,2017,2018,2019,2020  Google, Inc.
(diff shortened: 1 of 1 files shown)


behdad:
Thank you. A lot of contributions have not updated header files. But good to at least update the ones we are aware of.

PR #4087, opened 2023-01-27
fix meson install tag for `harfbuzz-config.cmake`

benoit-pierre:
So it does not get installed when filtering out `devel` stuff, e.g. by using `meson install --tags runtime`.

Note: [installation tags](https://mesonbuild.com/Installing.html#installation-tags) support is available starting with meson version 0.60.0, released Nov 2, 2021.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,5 +1,5 @@
 project('harfbuzz', 'c', 'cpp',
-  meson_version: '>= 0.55.0',
+  meson_version: '>= 0.60.0',
   version: '6.0.0',
   default_options: [
     'cpp_eh=none',          # Just to support msvc, we are passing -fno-exceptions also anyway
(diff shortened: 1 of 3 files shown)


meson.build
```diff
@@ -1,5 +1,5 @@
 project('harfbuzz', 'c', 'cpp',
-  meson_version: '>= 0.55.0',
+  meson_version: '>= 0.60.0',
```
khaledhosny, line 2: We intentionally don't depend on a too new version of meson.

benoit-pierre: Then feel free to close.

khaledhosny: If @behdad is OK with raising the minimum version we can do it, we can also keep the minimum version of older meson versions will ignore the new option.

benoit-pierre: Does it? From my limited testing, the build will fail if an unknown keyword is used (e.g. after renaming to `install_tags`).

benoit-pierre:
I've updated the Linux CI workflow version of meson.

meson.build
behdad: I prefer not. It's just over a year old.

PR #4089, opened 2023-01-28
Move GDEF and name tables into OT/

matthiasclasen:


behdad:
Thanks. Makefile needs update.

matthiasclasen:
Arrr. I always forget!

PR #4092, opened 2023-01-29
[ot-font] Use the cmap cache more

matthiasclasen:
Use the cmap cache for get_nominal_glyph and
get_variation_glyph as well. The first of these
is used a lot in pango.

### DIFF
--- a/src/hb-ot-cmap-table.hh
+++ b/src/hb-ot-cmap-table.hh
@@ -31,6 +31,7 @@
 #include "hb-ot-shaper-arabic-pua.hh"
 #include "hb-open-type.hh"
 #include "hb-set.hh"
+#include "hb-cache.hh"
 
 /*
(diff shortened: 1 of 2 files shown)


src/hb-ot-cmap-table.hh
```diff
@@ -1896,9 +1896,8 @@ struct cmap
     ~accelerator_t () { this->table.destroy (); }
 
     bool get_nominal_glyph (hb_codepoint_t  unicode,
-			    hb_codepoint_t *glyph) const
+                            hb_codepoint_t *glyph) const
     {
-      if (unlikely (!this->get_glyph_funcZ)) return false;
```
behdad, line 1901: I just need to verify this line.

behdad: Currently it can be null sometimes, so let's add that line back.

PR #4093, opened 2023-01-30
[coverage] Add a static cache

behdad:
Speeds up Roboto by around 5%; costs 2kb globally.


behdad:
Going to close since it makes tsan unhappy even though it's safe. The alternative, to make tsan happy, consumes twice the memory. In general it's a bit ugly.

behdad:
tsan issue was addressed.

PR #4095, opened 2023-01-31
[subset] for keep everything, don't drop any tables.

garretrieger:


behdad:
Should we passthru unrecognized as well?

behdad:
> Should we passthru unrecognized as well?

I'll push that out.

behdad:
> > Should we passthru unrecognized as well?
> 
> I'll push that out.

Oh it's already there. Nevermind.

Issue #3883, opened 2022-11-21
[draw] embolden outlines?


khaledhosny:
I’m trying to replace a code that uses GDI/FreeType/CoreText to get glyph outlines with hb-draw, and it works fine except that I need to also handle synthetic bold (synthetic slant works already), not sure if this is in scope, but I thought to ask.

behdad:
I'd say it's in scope, just not very easy to implement...

PR #4097, opened 2023-02-01
Embolden

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/3883

### DIFF
--- a/src/hb-font.cc
+++ b/src/hb-font.cc
@@ -2448,6 +2448,27 @@ hb_font_get_ptem (hb_font_t *font)
   return font->ptem;
 }
 
+/**
+ * hb_font_set_synthetic_bold:
+ * @font: #hb_font_t to work upon
+ * @x_embolden: the amount to embolden horizontally
+ * @y_embolden: the amount to embolden vertically
+ *
+ * Sets the "synthetic boldness" of a font.
+ *
+ * Positive values make a font bolder, negative values
+ * thinner. Typical values are in the 0.01 to 0.05 range.
+ * The default value is zero.
+ *
+ * Synthetic boldness is applied by offsetting the contour
+ * points of the glyph shape.
+ *
+ * HarfBuzz applies synthetic boldness when rendering a glyph
+ * via hb_font_draw_glyph(), and also uses the values to adjust
+ * the advance width of emboldened glyphs.
+ *
+ * Since: REPLACEME
+ **/
 void
 hb_font_set_synthetic_bold (hb_font_t *font, float x_embolden, float y_embolden)
 {
@@ -2465,6 +2486,16 @@ hb_font_set_synthetic_bold (hb_font_t *font, float x_embolden, float y_embolden)
   font->mults_changed ();
 }
 
+/**
+ * hb_font_get_synthetic_bold:
+ * @font: #hb_font_t to work upon
+ * @x_embolden: return location for horizontal value
+ * @y_embolden: return location for vertical value
+ *
+ * Fetches the "synthetic boldness" of a font.
+ *
+ * Since: REPLACEME
+ **/
 void
 hb_font_get_synthetic_bold (hb_font_t *font, float *x_embolden, float *y_embolden)
 {
--- a/src/hb-font.cc
+++ b/src/hb-font.cc
@@ -1772,8 +1772,12 @@ DEFINE_NULL_INSTANCE (hb_font_t) =
 
   1000, /* x_scale */
   1000, /* y_scale */
-  0., /* slant */
-  0., /* slant_xy; */
+  0.f, /* x_embolden */
+  0.f, /* y_embolden */
+  0, /* x_shift */
+  0, /* y_shift */
+  0.f, /* slant */
+  0.f, /* slant_xy; */
   1.f, /* x_multf */
   1.f, /* y_multf */
   1<<16, /* x_mult */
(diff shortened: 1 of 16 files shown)


behdad:
Todo:
- [ ] Relicense the code from FreeType authors (contacted),
- [ ] Add `hb_font_t` API and hook it up. Currently embolden amount hardcoded.
- [ ] Adapt advances and shaping to it.

matthiasclasen:
I've added some docs.

behdad:
@khaledhosny rightly pointed out that adjusting the advance-widths unfortunately breaks Arabic fonts that use rounded edges for their connecting strokes. We can't ignore that unfortunately. So I'm stuck as to how to proceed. Khaled also mentioned that in LibreOffice he uses unadjusted advance-widths and so far no one complained.

If using unadjusted advance-widths though, I think we should NOT shift the glyph shape horizontally then. Which I can do. But then we should also unshift the FreeType outlines in hb-ft.  I think we should still shift them vertically such that they sit on the baseline.

All of this ends up who knows where in vertical writing....

behdad:
No idea why on mac-os two shape tests timeout. Is it just that bot is so underpowered?!

khaledhosny:
> @khaledhosny rightly pointed out that adjusting the advance-widths unfortunately breaks Arabic fonts that use rounded edges for their connecting strokes. We can't ignore that unfortunately. So I'm stuck as to how to proceed. Khaled also mentioned that in LibreOffice he uses unadjusted advance-widths and so far no one complained.
> 
> If using unadjusted advance-widths though, I think we should NOT shift the glyph shape horizontally then. Which I can do. But then we should also unshift the FreeType outlines in hb-ft. I think we should still shift them vertically such that they sit on the baseline.
> 
> All of this ends up who knows where in vertical writing....

There are other cases where keeping advance width is desired like monospace fonts or using emboldening to achieve something like [grade](https://fonts.google.com/knowledge/glossary/grade). I think it might be worth while to add an option or a flag to control advance adjustment and shift in both directions.

behdad:
> There are other cases where keeping advance width is desired like monospace fonts or using emboldening to achieve something like [grade](https://fonts.google.com/knowledge/glossary/grade). I think it might be worth while to add an option or a flag to control advance adjustment and shift in both directions.

Matthias had the same idea. Okay I'm convinced.

behdad:
Okay so then to hb-view I'll add `--font-bold` and `--font-grade`?

behdad:
@khaledhosny @matthiasclasen How does this look?

behdad:
> * [ ]  Glyph extents.
> * [ ]  Font extents. The rest of metrics we're not going to try.

Working on these.

behdad:
Should grade shift glyphs up or not? Currently it does. @khaledhosny @davelab6

behdad:
> > * [ ]   Glyph extents.
> > * [ ]   Font extents. The rest of metrics we're not going to try.
> 
> Working on these.

Done. I think this is good to land now. I can work on `hb-ot-metrics` later.

src/hb-font.cc
```diff
@@ -2442,6 +2451,74 @@ hb_font_get_ptem (hb_font_t *font)
...
+ * 0.01 to 0.05 range. The default value is zero.
+ *
+ * Synthetic boldness is applied by offsetting the contour
+ * points of the glyph shape.
+ *
+ * Synthetic boldness is applied when rendering a glyph via
+ * hb_font_draw_glyph().
+ *
+ * If @in_place is %FALSE, then glyph advance-widths are also
+ * adjusted, otherwise they are not.
```
matthiasclasen, line 2474: It might be nice to give a hint about the use case for @in_place here.

util/font-options.hh
```diff
@@ -286,7 +332,11 @@ font_options_t::add_options (option_parser_t *parser)
 			      G_OPTION_ARG_CALLBACK,	(gpointer) &parse_font_ppem,	"Set x,y pixels per EM (default: 0; disabled)",	"1/2 integers"},
     {"font-ptem",	0, font_size_flags,
 			      G_OPTION_ARG_DOUBLE,	&this->ptem,			"Set font point-size (default: 0; disabled)",	"point-size"},
-    {"font-slant",	0, 0,
+    {"font-bold",	0, font_size_flags,
+			      G_OPTION_ARG_CALLBACK,	(gpointer) &parse_font_bold,	"Set synthetic bold (default: 0)",		"1/2 numbers; eg. 0.05"},
+    {"font-grade",	0, font_size_flags,
```
matthiasclasen, line 337: Is that a well-known term? I would not know what --font-grade does.

matthiasclasen:
Looks good to me apart from those two comments.

util/font-options.hh
behdad: It is. I just don't know where to explain it more.

src/hb-font.cc
behdad: I'll mention grading here and link to https://fonts.google.com/knowledge/glossary/grade

behdad: @khaledhosny What's the docbook syntax for links? https://developer-old.gnome.org/gtk-doc-manual/stable/documenting_syntax.html.en only has the markdown syntax. Is that kosher in our docs?

behdad: > Is that kosher in our docs?

Seems to work.

PR #4110, opened 2023-02-10
Clarify that those two test fonts are CC0-licensed

jenskutilek:
As I received a question about the license for two test fonts, I updated the internal license fields in the name tables.

```
This subset of the Normschrift font is licensed under https://creativecommons.org/publicdomain/zero/1.0/deed.en. The full Normschrift font is copyright 2017 by Jens Kutilek.
```


behdad:
Thank you Jens.

PR #4116, opened 2023-02-12
Update hb-info.cc

pg83:
shut up -Werror + release build + musl libc + "unused variable" warning


behdad:
I tried a fix in https://github.com/harfbuzz/harfbuzz/commit/40fa046cf311718665496d0516495e9c139221ba
Does that work for you?

pg83:
> I tried a fix in [40fa046](https://github.com/harfbuzz/harfbuzz/commit/40fa046cf311718665496d0516495e9c139221ba)
> Does that work for you?

Why not, basicaly it is the same.

PR #4121, opened 2023-02-14
[instancer] Apply cvar deltas 

qxliu76:


behdad:
Any idea why one bot is failing?

src/hb-ot-var-cvar-table.hh
```diff
@@ -0,0 +1,151 @@
...
+        if (unlikely (idx >= num_cvt_item)) continue;
+        if (scalar != 1.0f) cvt_deltas[idx] += unpacked_deltas[i] * scalar ;
+        else cvt_deltas[idx] += unpacked_deltas[i];
+      }
+    } while (iterator.move_to_next ());
+
+    return true;
+  }
+
+  bool add_cvt_and_apply_deltas (hb_subset_plan_t *plan) const
```
behdad, line 104: Should this be a static function?

qxliu76: Done

qxliu76:
> Any idea why one bot is failing?

No..I'm trying to get access and debug on windows machine. Would you have any thoughts on what might be the issue on windows machine?

behdad:
> Would you have any thoughts on what might be the issue on windows machine?

I don't.

PR #4126, opened 2023-02-19
CFF2 instancer make ots-sanitize happy

behdad:
Trying to make CFF2 instancer pass ots-sanitize


behdad:
What is left is processing blends in the Private dict. Not easy.

behdad:
We should add a subset-test, not API test, for CFF2 instancing, such that the ots-sanitize is exercised, which would fail currently.

behdad:
cc @qxliu76 @garretrieger

behdad:
cff2_private_dict_values_subset_t is relevant.

behdad:
The problem is that the Private dict subsetter doesn't parse the values into memory, but writes them out from the original dict. So we can't easily process blends there. :(

behdad:
This passes OTS now.

Issue #4066, opened 2023-01-23
error: use of undeclared identifier '__builtin_mul_overflow'


ryandesign:
Hello, I'm the maintainer of harfbuzz in MacPorts where I just updated it to version 6.0.0 and it is failing to build on just one system:
```
In file included from main.cc:350:
In file included from ./hb-static.cc:27:
In file included from ./hb.hh:480:
./hb-algs.hh:882:10: error: use of undeclared identifier '__builtin_mul_overflow'
  return __builtin_mul_overflow (count, size, result);
         ^
```
Here's the [full log](https://build.macports.org/builders/ports-10.10_x86_64-builder/builds/211885/steps/install-port/logs/stdio).

This system where the problem occurs runs:

* OS X Yosemite v10.10.5 (14F2511)
* Xcode v7.2.1 (7C1002)
* Apple LLVM version 7.0.2 (clang-700.1.81) 

But there is no problem on newer systems including:

* OS X El Capitan v10.11.6 (15G22010)
* Xcode v8.2.1 (8C1002)
* Apple LLVM version 8.0.0 (clang-800.0.42.1)

There is also no problem on older systems including:

* OS X Mavericks v10.9.5 (13F1911)
* Xcode v6.2 (6C131e)
* Apple LLVM version 6.0 (clang-600.0.57) (based on LLVM 3.5svn)

because when the Xcode Clang version is < 700 we already switch to a newer Clang from MacPorts because of https://github.com/harfbuzz/harfbuzz/issues/1651#issuecomment-484051395.

I have modified our build to use MacPorts Clang now when the Xcode Clang version is < 800 and that works around the problem however if it is possible to reinstate support for older Clangs perhaps that is desirable.

behdad:
Thanks. I tried a fix. Please test.

ryandesign:
Thank you. I haven't had a chance to try it yet but a problem I see is that Apple clang and open source clang use different version numbering schemes. This feature you're testing for apparently appeared in Apple clang 8, but it will have appeared in a different version of open source clang. You can determine whether the compiler is Apple clang by checking whether `__apple_build_version__` is defined.

ryandesign:
When I updated the MacPorts harfbuzz-devel port (intended for tracking development versions and verifying updates before updating the main harfbuzz port) to 7.0.1 I re-allowed the use of Apple Clang 700, and the build failed again the same way, so the fix was not effective:

[...]

PR #4130, opened 2023-02-22
Use __has_builtin for builtin checks instead of compiler versions

behdad:
https://github.com/harfbuzz/harfbuzz/issues/4066#issuecomment-1439510188


behdad:
Patch fully tested now.

PR #4131, opened 2023-02-22
Wasm

behdad:


behdad:
cc @ebraminio

meson_options.txt
```diff
@@ -21,6 +21,8 @@ option('directwrite', type: 'feature', value: 'disabled',
   description: 'Enable DirectWrite shaper backend on Windows (experimental)')
 option('coretext', type: 'feature', value: 'disabled',
   description: 'Enable CoreText shaper backend on macOS')
+option('wasm', type: 'feature', value: 'disabled',
+  description: 'Enable WebAssembly shaper backend')
```
ebraminio, line 25: Awesome work. Since this apparently will use a generic 'Wasm' table tag from a font maybe it can be named a 'font backend' in hope someday it gains outline drawing / painting capabilities.

behdad:
cc @jfkthame

behdad:
I compiled graphite to a wasm library and and integrated it with hb-wasm-api, then embedded it in CharilSIL. The CharisSIL font is 1.7MB. The Wasm table with entire graphite library in it is 110kb. It renders.

cc @mhosken

behdad:
I think with a bit of polish this can actually be merged as an experimental feature, off by default.

behdad:
@dscorbett Is Duployan shorthand something that can benefit from custom processing? I know your font pushes the limits of OpenType.

dscorbett:
> Is Duployan shorthand something that can benefit from custom processing? I know your font pushes the limits of OpenType.

Yes, it could benefit from custom processing. Most of its complexity is because it lacks runtime access to cursively connected glyph sequences’ overall bounding boxes. I have [a workaround](https://github.com/dscorbett/duployan-font/blob/c7f0ccb428b5fb08083f00796c9855186f2c3944/docs/width-system.md), but with this new shaper I could finally do math with numbers instead of glyphs. And there are other features that haven’t been realistic to consider because there wasn’t enough space for them in GSUB.

docs/wasm-shaper.md
```diff
@@ -2,9 +2,375 @@
...
 
+## What you can and can't do: the WASM shaper's role in shaping
+
+The Harfbuzz shaping engine, unlike its counterparts CoreText and DirectWrite, is only responsible for a small part of the text rendering process. Specifically, Harfbuzz is purely responsible for *shaping*; although Harfbuzz does have APIs for accessing glyph outlines, typically other libraries in the free software text rendering stack are responsible for text segmentation into runs, outline scaling and rasterizing, setting text on lines, and so on.
+
+Harfbuzz is therefore restricted to turning a buffer of codepoints for a segmented run of the same script, language, font, and variation settings, into glyphs and positioning them. This is also all that you can do with the WASM shaper; you can influence the process of mapping a string of characters into n array of glyphs, you can determine how those glyphs are positioned and their advance widths, but you cannot manipulate outlines, variations, line breaks, or affect text layout between texts of different font, variation, language, script or OpenType feature selection.
```
behdad, line 9: Typo "into a array of"

behdad:
> I've no objection to calling them pointers if you prefer.

I think I prefer that since that's what's in the API. Up to you though.


> > Maybe better to add API docs to hb-wasm-api.h and point people to that instead?
>
> I wrote this because I received complaints that "read the header file" was not adequate documentation. :-)

I meant documenting the API and having gtkdoc generate API docs like our normal API documentation.

Thanks for this.

BaGsn:
Hi,
I see that this has been release but I could not find any example/documentation/mention of it anywhere (beside in the code of harfbuzz itself).
Is there any open source source implementation of wasm font example somewhere in the wild? As for example the demo fonts that were shown at ATIPY? That would be very helpfull as a starting point to experiment with wasm. Thanks!

simoncozens:
Please see https://github.com/harfbuzz/harfbuzz-wasm-examples

Issue #4125, opened 2023-02-18
Cantarell CFF2 subsetting corrupts outlines


madig:
Tested with https://github.com/harfbuzz/harfbuzz/tree/adccc5355b695cfaaf6403a3187c67c7fcae623d.

1. Get https://cantarell.gnome.org/releases/cantarell-fonts-0.303.tar.xz and pull out the `prebuilt/*.otf` file inside
2. Run `hb-subset --glyphs="*" --layout-features="*" --layout-scripts="*" --drop-tables-="*" --instance="wght=100" Cantarell-VF.otf -o /tmp/a.otf`

![image](https://user-images.githubusercontent.com/380829/219903680-46e983b2-82e4-40e9-a54e-35486ecb3832.png)

It also fails to load in sites like http://cyreal.org/Font-Testing-Page/, with these error messages in Firefox:

* downloadable font: Variations: Required fvar table is missing (font-family: "a" style:normal weight:400 stretch:100 src index:0)
* downloadable font: Variations: Failed to parse variation region list (font-family: "a" style:normal weight:400 stretch:100 src index:0)
* downloadable font: CFF2: Failed to parse Top DICT Data (font-family: "a" style:normal weight:400 stretch:100 src index:0)
* downloadable font: CFF2: Failed to parse table (font-family: "a" style:normal weight:400 stretch:100 src index:0)
* downloadable font: rejected by sanitizer (font-family: "a" style:normal weight:400 stretch:100 src index:0)
* downloadable font: font load failed (font-family: "a" style:normal weight:400 stretch:100 src index:0)

behdad:
Oh boy...

behdad:
> * downloadable font: Variations: Required fvar table is missing (font-family: "a" style:normal weight:400 stretch:100 src index:0)

Humm. Is OTS requiring fvar table for CFF2 fonts? If that's the case, that seems wrong. cc @jfkthame @khaledhosny

behdad:
Oh, it's probably because we didn't null the VarStore in the CFF2 table. Nevermind.

madig:
Random thought: should subsetting tests employ OTS to sniff-check resulting files?

behdad:
> Random thought: should subsetting tests employ OTS to sniff-check resulting files?

We already do that:
https://github.com/harfbuzz/harfbuzz/blob/main/test/subset/run-tests.py#L22

We just don't have many (any?) tests for CFF instancing.

cc @qxliu76

[...]

PR #4132, opened 2023-02-22
[subset] Add a test for CFF2 instancing.

garretrieger:
Adds option to disable the fonttools comparison check in the test. This is needed since CFF2 instancing is not yet supported in fonttools.

For #4125


khaledhosny:
FontTools varLib.mutator supports CFF2 instancing

garretrieger:
I'm running into this error: https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/varLib/instancer/__init__.py#L1085 when trying to use fonttools. Should that check be removed on the fonttools side then?

khaledhosny:
[varLib.mutator](https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/varLib/mutator.py) is a different one (older, and officially “deprecated”, and does not support L3 and L4 instancing, but it supports CFF2). Just in case you want to check CFF2 instancing against it.

behdad:
I don't think mutator supports blending the Private dict etc. So it needs more work to be in par with HarfBuzz now. Not sure it's worth trying. Or maybe it is.

PR #4135, opened 2023-02-24
[instancer] bug fix in post table applying mvar deltas

qxliu76:


behdad:
So if no glyph names are requested, what do we do? Generate no post table at all?

qxliu76:
> So if no glyph names are requested, what do we do? Generate no post table at all?



> So if no glyph names are requested, what do we do? Generate no post table at all?
In this case, we will generate a post table with version 3. This fix changes the code to apply mvar deltas after serialization. Previously we just did start_embed() before applying mvar deltas.

behdad:
Thanks.

PR #4144, opened 2023-02-28
Justify

behdad:


ebraminio:
Awesome work, just that once I was curious whether this can somehow help on math uses also https://github.com/chromium/chromium/blob/main/third_party/blink/renderer/platform/fonts/shaping/stretchy_operator_shaper.cc#L49-L55 or perhaps guess not.

behdad:
Currently hb-view only works for one-line text.

PR #4154, opened 2023-03-08
Ensure glyf_accelerator's hmtx/vmtx/gvar accelerators are always popu…

garretrieger:
…lated.

Fixes fuzzer issue: https://oss-fuzz.com/testcase-detail/4875306193518592


behdad:
I thought about this. I think the current code is intentional. I see two reasons why not initializing those tables is ideal:

- It better imitates the Null() object, which is something we need to be able to handle,
- In the case of wrong table format, we should never access those fields. So nullptr catches bugs, as it did in this case.

So you are essentially papering over the actual bug here.

PR #4169, opened 2023-03-15
[subset] Don't add invalid gids (-1) to the glyphset when loading glyph map from the accelerator.

garretrieger:


src/test-set.cc
```diff
@@ -137,5 +137,29 @@ main (int argc, char **argv)
...
+
+    hb_codepoint_t array[] = {(unsigned) HB_SET_VALUE_INVALID, 0, 2};
+    s.clear();
+    s.add_array(array, 3);
+    assert(!s.has(HB_SET_VALUE_INVALID));
+    assert(s.has(2));
+
+    hb_codepoint_t sorted_array[] = {0, 2, (unsigned) HB_SET_VALUE_INVALID};
+    s.clear();
+    s.add_array(sorted_array, 3);
```
behdad, line 159: Did you mean add_sorted_array here?

garretrieger: Ah yes, thanks. I'll fix.

PR #4180, opened 2023-03-20
[instancer] COLRv1 full instancing support

qxliu76:


garretrieger (approved):

src/OT/Color/COLR/COLR.hh
```diff
@@ -945,13 +1059,26 @@ struct PaintScaleAroundCenter
...
-    return_trace (out->src.serialize_subset (c, src, this));
+    if (instancer && !c->plan->pinned_at_default && varIdxBase != VarIdx::NO_VARIATION)
+    {
+      out->scaleX.set_float (scaleX.to_float (instancer (varIdxBase, 0)));
+      out->scaleY.set_float (scaleY.to_float (instancer (varIdxBase, 1)));
+      out->centerX = centerX + (int) roundf (instancer (varIdxBase, 2));
+      out->centerY = centerY + (int) roundf (instancer (varIdxBase, 3));
+    }
+
+    if (format == 19 && c->plan->all_axes_pinned)
```
behdad, line 1078: While you lower the format, you still are serializing the bigger type I believe. No?

qxliu76: In case of bigger type: e.g Variable<PaintScaleAroundCenter>, the call is dispatched to subset() in struct Variable<> first, and I made code changes in it to make sure it will not serialize the varIdxBase if all axes are pinned.

behdad: Ah okay I see. A bit sketchy but does the job I suppose.

PR #4189, opened 2023-04-04
[instancer] update bound metrics for CFF2 instancing

### LINKED ISSUE #4149: CFF2 instancing: does not instantiate advance widths?
Follow-up to https://github.com/harfbuzz/harfbuzz/issues/4125.

Same font, but e.g. the instance of wght=400 and wght=800 have the same advance width, as if the `hmtx` table isn't being variated.

qxliu76:
trying to fix #4149


src/hb-ot-var-hvar-table.hh
```diff
@@ -323,7 +326,7 @@ struct HVARVVAR
 				    const int *coords, unsigned int coord_count,
 				    VariationStore::cache_t *store_cache = nullptr) const
   {
-    uint32_t varidx = (this+advMap).map (glyph);
+    uint32_t varidx = advMap ? (this+advMap).map (glyph) : glyph;
```
behdad, line 329: Why is this line needed? The null `DeltaSetIndexMap` is designed to handle that I think.

qxliu76: Oh yes, I have reverted my change. Thanks!

src/hb-subset-plan.cc
```diff
@@ -853,6 +854,80 @@ _normalize_axes_location (hb_face_t *face, hb_subset_plan_t *plan)
...
+
+    if (_hmtx.has_data ())
+    {
+      int hori_aw = _hmtx.get_advance_without_var_unscaled (old_gid);
+      if (_hmtx.var_table.get_length ())
+        hori_aw += (int) roundf (_hmtx.var_table->get_advance_delta_unscaled (old_gid, font->coords, font->num_coords,
+                                                                              hvar_store_cache));
+      int lsb = extents.x_bearing;
+      if (!has_bounds_info)
+        _hmtx.get_leading_bearing_without_var_unscaled (old_gid, &lsb);
```
behdad, line 906: This might return false.

src/hb-subset-plan.cc
```diff
@@ -853,6 +854,80 @@ _normalize_axes_location (hb_face_t *face, hb_subset_plan_t *plan)
...
+{
+  if (!plan->normalized_coords) return;
+  OT::cff2::accelerator_t cff2 (plan->source);
+  if (!cff2.is_valid ()) return;
+
+  hb_font_t *font = nullptr;
+  if (unlikely (!plan->check_success (font = _get_hb_font_with_variations (plan))))
+    return;
+
+  hb_glyph_extents_t extents = {0};
```
behdad, line 869: This always takes 0,0 into the bounds calculation. Maybe that's okay, but it's not always correct.

src/hb-subset-plan.cc
```diff
@@ -853,6 +854,86 @@ _normalize_axes_location (hb_face_t *face, hb_subset_plan_t *plan)
...
+
+static void
+_update_instance_metrics_map_from_cff2 (hb_subset_plan_t *plan)
+{
+  if (!plan->normalized_coords) return;
+  OT::cff2::accelerator_t cff2 (plan->source);
+  if (!cff2.is_valid ()) return;
+
+  hb_font_t *font = nullptr;
+  if (unlikely (!plan->check_success (font = _get_hb_font_with_variations (plan))))
```
garretrieger, line 866: add a hb_font_destroy (font) before returning, since check_success can return false even if font creation failed in the event the plan was already in error. See: https://github.com/harfbuzz/harfbuzz/pull/4187/files

src/hb-subset-plan.cc
```diff
@@ -853,6 +854,86 @@ _normalize_axes_location (hb_face_t *face, hb_subset_plan_t *plan)
...
+  float *hvar_store_cache = nullptr;
+  if (_hmtx.has_data () && _hmtx.var_table.get_length ())
+    hvar_store_cache = _hmtx.var_table->get_var_store ().create_cache ();
+  
+  OT::vmtx_accelerator_t _vmtx (plan->source);
+  float *vvar_store_cache = nullptr;
+  if (_vmtx.has_data () && _vmtx.var_table.get_length ())
+    vvar_store_cache = _vmtx.var_table->get_var_store ().create_cache ();
+
+  for (unsigned new_gid = 0; new_gid < plan->num_output_glyphs (); new_gid++)
```
garretrieger, line 880: It's preferable to iterate over the entries of the glyph map, since when retain_gids is on there can be a large number of empty glyphs and the glyph map only contains the non-empty glyphs.

test/subset/data/tests/instantiate_cff2_update_metrics.tests
```diff
@@ -0,0 +1,14 @@
+FONTS:
+Cantarell-VF-ABC.otf
+
+PROFILES:
+default.txt
```
garretrieger, line 5: also add a retain gids profile

garretrieger (approved):

PR #4199, opened 2023-04-17
[subset] Fix fuzzer issue https://oss-fuzz.com/testcase-detail/6521393809588224

garretrieger:


behdad:
ubsan is failing for something unrelated that we should probably fix. @qxliu76

garretrieger:
> ubsan is failing for something unrelated that we should probably fix. @qxliu76

I was able to fix it, was a missing (int) cast.

PR #4202, opened 2023-04-21
[instancer] compile composite glyphs directly with shifted component instead of deltas

qxliu76:


behdad:
Ah, it's because the component points are fetched late and are not in the points vector by the time you copy them. They are fetched in `comp_points` later in Glyph.hh:compile_bytes_with_deltas. Might be able to move it earlier.

behdad:
What we probably want is a mode to get_points to NOT recurse, and use that instead of the `points_with_deltas`.

qxliu76:
> Ah, it's because the component points are fetched late and are not in the points vector by the time you copy them. They are fetched in `comp_points` later in `Glyph.hh:get_points`. Might be able to move it earlier.

but I don't want comp_points, which are contour points of child glyphs. I just need x/y offset component points, I thought they're already processed by `gvar->apply_deltas_to_points`?

behdad:
> > Ah, it's because the component points are fetched late and are not in the points vector by the time you copy them. They are fetched in `comp_points` later in `Glyph.hh:get_points`. Might be able to move it earlier.
> 
> but I don't want comp_points, which are contour points of child glyphs. I just need x/y offset component points, I thought they're already processed by `gvar->apply_deltas_to_points`?

You are right.

behdad:
You need the equivalent of this line in the block before it:
https://github.com/harfbuzz/harfbuzz/blob/main/src/OT/glyf/Glyph.hh#L332

behdad:
I'm trying to make it work. Might be too tricky.

qxliu76:
> What we probably want is a mode to get_points to NOT recurse, and use that instead of the `points_with_deltas`.

Agreed.

behdad:
I pushed a change onto main. Rebasing your change onto that passes tests now.

behdad:
I pushed this to main manually. Thanks.

Let's see if we can make get_points have a recurse parameter separately.

qxliu76:
Thank you! I was still trying to see how to do get_points for comp glyph and you fixed already:)
Seems we need to have both contour points and component points for composite glyphs. Contour points will be used to calculate xMin/xMax/yMin/yMax

behdad:
I see. Maybe not worth it to touch then.

PR #4205, opened 2023-04-24
Gvar optimize

behdad:


behdad:
@qxliu76 One test fails here, can you please update the expectation? It's just one point has one unit difference. I think that's fine.

behdad:
Or do we deviate from fonttools here?

behdad:
I measured more. I still see 6% speedup.  Let's consider.

behdad:
I found some other optimization opportunities. So this is at combined 14% now.

PR #4206, opened 2023-04-24
[subset] Fix inefficient ItemVariationStore subsetting w/ retain_gids.

garretrieger:
ItemVariationStore is relying on the assumption that the inner_map is populated for all output glyphs, this is not true for subsetting operations with retain gids enabled. Fixes fuzzer timeout: https://oss-fuzz.com/testcase-detail/4575222591520768.


behdad:
Good catch.

PR #4207, opened 2023-04-26
[unicode] Fix typo

arHSM:
I believe the `hb-unicode.h` has a typo where `HB_UNICODE_COMBINING_CLASS_CCC133	= 132,` is supposed to be `HB_UNICODE_COMBINING_CLASS_CCC132`


behdad:
Thanks!

dscorbett:
Should `HB_UNICODE_COMBINING_CLASS_CCC133` exist for backwards compatibility?

behdad:
Thinking more, probably yes.

behdad:
Should I add it with a value of 132 or 133?

behdad:
hb_unicode_combining_class_t is rather unused publicly. It was a mistake to expose it IMO.

dscorbett:
I think it should have the value 133. I would treat this as a combination of a bug fix (`HB_UNICODE_COMBINING_CLASS_CCC133` should always have had the value 133) and a new feature (`HB_UNICODE_COMBINING_CLASS_CCC132` exists).

behdad:
Agreed. I'll add it to `hb-deprecated.h` as a macro.

behdad:
> Agreed. I'll add it to `hb-deprecated.h` as a macro.

Done.

dscorbett:
See [L2/12-286](https://www.unicode.org/L2/L2012/12286-ccc133.pdf) for how this typo appeared in the first place.

PR #4209, opened 2023-04-27
[instancer] no need to create hb_font_t object when fetching delta from varstore

qxliu76:


behdad:
Thanks. Valgrind not happy.

behdad:
Humm. That *might* be my fault.

behdad:
Yeah, it is.

PR #4216, opened 2023-04-29
[buffer] fix `hb_buffer_t::similar` setting `replacement` to `src.invisible`

arHSM:
```cpp
void
hb_buffer_t::similar (const hb_buffer_t &src)
{
  hb_unicode_funcs_destroy (unicode);
  unicode = hb_unicode_funcs_reference (src.unicode);
  flags = src.flags;
  cluster_level = src.cluster_level;
  replacement = src.invisible; // <- this should be src.replacement
  invisible = src.invisible;
  not_found = src.not_found;
}
```


behdad:
Thank you!

PR #4218, opened 2023-05-02
[Ligature] Speed up

behdad:
Match the first component of the ligature in the LigatureSet loop.

Speeds up Roboto shaping by 25%. I don't think it breaks anything. The test suite seems happy.

### DIFF
--- a/src/OT/Layout/GSUB/Ligature.hh
+++ b/src/OT/Layout/GSUB/Ligature.hh
@@ -10,7 +10,7 @@ namespace GSUB_impl {
 template <typename Types>
 struct Ligature
 {
-  protected:
+  public:
   typename Types::HBGlyphID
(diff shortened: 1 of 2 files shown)


behdad:
Specially with fonts with ligatures, marks in between ligating components, and unsafe-to-concat produced...

khaledhosny:
I tested with a couple of ligature heavy fonts and a large corpus of Arabic text (with all sorts of marks), using `--verify --unsafe-to-concat --safe-to-insert-tatweel` and I got no verification issue nor any change of output after this change.

behdad:
> I tested with a couple of ligature heavy fonts and a large corpus of Arabic text (with all sorts of marks), using `--verify --unsafe-to-concat --safe-to-insert-tatweel` and I got no verification issue nor any change of output after this change.

Thank you!

PR #4221, opened 2023-05-04
[subset] Add API method to allow a custom glyph map to be specified.

garretrieger:


garretrieger:
> Let's fail subsetting if the mapping is not unique, or if retain-gid is set.

Done

behdad:
LGTM!

behdad:
@khaledhosny Is the linux-ci doc check going to fail all the time that we have new API till a release? It wasn't that way. What changed this time?

khaledhosny:
> @khaledhosny Is the linux-ci doc check going to fail all the time that we have new API till a release? It wasn't that way. What changed this time?

Use XSince instead of Since, gtk-doc will ignore it and I fix it when cutting the release and adding version number.

khaledhosny:
Seems to work fine from my end, thanks!

Issue #4231, opened 2023-05-09
[subset] Possibly use vector instead of map for a few items


behdad:
These: `bounds_width_map`, `bounds_height_map`, `hmtx_map`, `vmtx_map`.

behdad:
Not sure if feasible. The main problem is retain-gid.

behdad:
I'm working on it.

qxliu76:
Oh sorry, just saw this. My understanding is we need to empty the glyphs that are not retained. During instancing, we call get_points() on empty glyphs to get phantom points so we can calculate adv width etc. I wonder if we need to do that for all emptied glyphs or just on one glyph and set the same metric for all emptied glyphs?

behdad:
@qxliu76 
Thanks. I did the bounds maps in https://github.com/harfbuzz/harfbuzz/pull/4232
One test is failing though. Can you take a look?

qxliu76:
sure, I'll take a look

behdad:
Half of it is done. The other half maybe not worth it.

PR #4232, opened 2023-05-09
[subset] Use vector instead of map for glyph bounds

behdad:
Is faster.

Part of https://github.com/harfbuzz/harfbuzz/issues/4231

### DIFF
--- a/src/hb-ot-hmtx-table.hh
+++ b/src/hb-ot-hmtx-table.hh
@@ -83,7 +83,7 @@ struct hmtxvmtx
   bool subset_update_header (hb_subset_context_t *c,
 			     unsigned int num_hmetrics,
 			     const hb_hashmap_t<hb_codepoint_t, hb_pair_t<unsigned, int>> *mtx_map,
-			     const hb_map_t *bounds_map) const
+			     const hb_vector_t<unsigned> &bounds_vec) const
   {
     hb_blob_t *src_blob = hb_sanitize_context_t ().reference_table<H> (c->plan->source, H::tableTag);
     hb_blob_t *dest_blob = hb_blob_copy_writable_or_fail (src_blob);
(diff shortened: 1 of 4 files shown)


src/hb-ot-hmtx-table.hh
```diff
@@ -125,9 +126,10 @@ struct hmtxvmtx
         int lsb = _.second.second;
         max_adv = hb_max (max_adv, adv);
 
-        if (bounds_map->has (gid))
+        if (bounds_vec[gid])
         {
-          unsigned bound_width = bounds_map->get (gid);
+	  empty = false;
```
qxliu76, line 131: oh there's a composite glyph which contains only one child glyph that is EMPTY, so bounds_vec value for parent glyph is 0 and min_lsb/min_rsb/max_extent are thus not updated accordingly.

qxliu76: maybe changing the vector to store another flag indicating whether this glyph is empty or not?

qxliu76: I tried to initialize the bounds_vec value to be 0xFFFFFFFF which seems working:

```
for (unsigned i = 0; i < _num_output_glyphs; i++)
    bounds_width_vec[i] = 0xFFFFFFFF;

```

not sure if that is efficient.

behdad: Can you please push a commit to this branch for that?

qxliu76: I got this error when trying to push to upstream: 
remote: Permission to harfbuzz/harfbuzz.git denied to qxliu76.
fatal: unable to access 'https://github.com/harfbuzz/harfbuzz.git/': The requested URL returned error: 403

I pushed to googlefonts instead: https://github.com/googlefonts/harfbuzz/tree/fix-vectors

PR #4233, opened 2023-05-10
Robinhood

behdad:
This implements Robinhood hashing. While it significantly improves lookup probing, it doesn't improve subset times.


behdad:
A couple tests timeout, so there's definitely a bug in there...

behdad:
In further testing, this seems to be win to me. Asking @garretrieger for confirmation.

Issue #3290, opened 2021-11-10
hb_font_extents_for_script/language API


behdad:
Which we would be able to hook into BASE table then.

behdad:
To look into this week? @matthiasclasen

matthiasclasen:
One suggestion:

```
HB_EXTERN void
hb_ot_layout_get_font_extents (hb_font_t         *font,
                               hb_direction_t     direction,
                               hb_tag_t           script_tag,
                               hb_tag_t           language_tag,
                               hb_font_extents_t *extents);
```

behdad:
Or maybe this is not a good idea at all...

behdad:
cc @rsheeter

behdad:
I'll go ahead and implement this.

PR #4237, opened 2023-05-19
[layout] Add hb_ot_layout_get_font_extents()

behdad:
Untested.

Fixes https://github.com/harfbuzz/harfbuzz/issues/3290


behdad:
Checking whether the NotoCJK fonts have the BASE table values for these...

behdad:
The BASE table also has versions of min/max with a feature-tag, which it says must match the GSUB/GPOS feature tag. That feels non-sense to me. How is one supposed to use it with multiple features?

behdad:
One more thing. Matthias and I discussed adding a "2" version of this, and baseline, API that take `hb_script_t` / `hb_language_t` instead of the OT tags.

behdad:
> Checking whether the NotoCJK fonts have the BASE table values for these...

Nope.

@blueshade7 do you know of any fonts that have BASE table MinMax records?

behdad:
> The BASE table also has versions of min/max with a feature-tag, which it says must match the GSUB/GPOS feature tag. That feels non-sense to me. How is one supposed to use it with multiple features?

Should I expose the feature-tag as well? @khaledhosny

blueshade7:
> @blueshade7 do you know of any fonts that have BASE table MinMax records?

not aware of

behdad:
I did some testing and fixups on this. Seems to work now.

behdad:
Should probably add a test.

behdad:
> Should probably add a test.

Done.

behdad:
> > The BASE table also has versions of min/max with a feature-tag, which it says must match the GSUB/GPOS feature tag. That feels non-sense to me. How is one supposed to use it with multiple features?
> 
> Should I expose the feature-tag as well? @khaledhosny

Just this outstanding.

khaledhosny:
Just a thought, since a line/paragraph can have multiple scripts/language, would it make since for these functions to take an array of scripts/languages and return the maximum extent?

I’m just think to think how can this API be used in multiscript/multilingual text layout. Right now, when multiple fonts are involved, some application will use the maximum of the line metrics of all fonts involved, others I think designate a font as the main font and use its metrics. I probably need how 

This would also be used for feature tags as well (i.e. accepting array of feature tags, not that I can imagine a use case for this, well may be if a stylistic set swishes between accent styles and one of them gives accents that are compressed into the capitals as seen is some signage).

cc @tiroj who have always been interested in this functionality.

behdad:
I think multiple scripts/languages would be higher-level's responsibility.

khaledhosny:
> I think multiple scripts/languages would be higher-level's responsibility.

 Makes sense.

src/hb-ot-layout.cc
```diff
@@ -2036,6 +2036,109 @@ hb_ot_layout_substitute_lookup (OT::hb_ot_apply_context_t *c,
...
+ *
+ * Fetches script/language-specific font extents.  These values are
+ * looked up in the `BASE` table's `MinMax` records.
+ *
+ * If no such extents are found, the default extents for the font are
+ * fetched. As such, the return value of this function can for the
+ * most part be ignored.  Note that the per-script/language extents
+ * do not have a line-gap value, and the line-gap is set to zero in
+ * that case.
+ *
```
khaledhosny, line 2121: Might be worth adding a note about the difference between the 2 and the other API.

behdad: Done.

khaledhosny (approved):

PR #4245, opened 2023-05-29
[subset] Fix fuzzer crash.

garretrieger:
https://oss-fuzz.com/testcase-detail/6608005089853440


behdad:
Similar issue exists in `MultipleSubstFormat1` I think.

behdad:
I like to better understand the crash reason though. This shouldn't be needed necessarily as the serializer should carry the error status.

garretrieger:
> I like to better understand the crash reason though. This shouldn't be needed necessarily as the serializer should carry the error status.

we've set up the serializer such that it's allowed to discard/revert while in error status if the error was an int overflow (or related) type variety. This is useful in some cases where the int overflow can be recovered (ie. discarding a cmap4 subtable).

So in this case, we get an int overflow but since that particular call wasn't checked we don't bail out and then subsequently attempt to do a pop_pack() which is a noop if any error type is set on the serializer. This then gets the serializer into a state where there is a unpacked object when revert is called which we have an assert to check for

PR #4251, opened 2023-06-01
[sorted-map] Add hb_sorted_map_t<V>

behdad:


behdad:
Looks like we don't need this after all.

Ended up with a simpler solution:
https://github.com/harfbuzz/harfbuzz/pull/4254

PR #4258, opened 2023-06-02
[BEInt] Use packed-int trick for writing as well

behdad:


behdad:
Might have small perf gain, but also makes binary smaller.

PR #4264, opened 2023-06-04
Glyph to sid cache

behdad:
This speeds up small CFF CJK retaingids subsets, but mysteriously slows down large CFF CJK retaingids subsets by 20%. My profiling doesn't show why. @garretrieger can you take a look please?


behdad:
The perf gain from this is rather minimal, so I'm fine abandoning it. But I still want to understand why I'm seeing a slowdown with /10000 benchmark.

PR #4267, opened 2023-06-05
[instancer] decompile tuple varstore

qxliu76:
- Added custom structs: tuple_delta_t and tuple_variations_t to hold tuple variation data and change axis limits
- Added function to decompile cvar tuple variation data into above structs
- Added a unit test for decompiling cvar tuple variations


src/hb-ot-var-common.hh
```diff
@@ -364,6 +364,132 @@ struct TupleVariationHeader
   DEFINE_SIZE_MIN (4);
 };
 
+struct tuple_delta_t
+{
+  public:
+  hb_hashmap_t<hb_tag_t, Triple> axis_tuples;
+
+  /* indices_length = point_count, indice[i] = 1 means point i is referenced */
+  hb_vector_t<uint8_t> indices;
```
behdad, line 420: Use `bool` maybe?

src/hb-ot-var-common.hh
```diff
@@ -364,6 +364,132 @@ struct TupleVariationHeader
...
+
+  bool set_tent (hb_tag_t axis_tag, Triple tent)
+  { return axis_tuples.set (axis_tag, tent); }
+
+  tuple_delta_t& operator += (const tuple_delta_t& o)
+  {
+    unsigned num = indices.length;
+    for (unsigned i = 0; i < num; i++)
+    {
+      if (indices[i])
```
behdad, line 466: In these loops you can access vectors' `.arrayZ` for faster access.

src/hb-ot-var-common.hh
```diff
@@ -364,6 +364,132 @@ struct TupleVariationHeader
...
+
+  tuple_delta_t& operator *= (float scalar)
+  {
+    if (scalar == 1.0f)
+      return *this;
+
+    unsigned num = indices.length;
+    for (unsigned i = 0; i < num; i++)
+    {
+      if (!indices[i]) continue;
```
behdad, line 494: Same here.

src/hb-ot-var-common.hh
behdad: Well, only if arrays are not in error. Maybe just do for `indices` since we're sure.

src/hb-ot-var-common.hh
```diff
@@ -249,6 +249,53 @@ struct TupleVariationHeader
   const TupleVariationHeader &get_next (unsigned axis_count) const
   { return StructAtOffset<TupleVariationHeader> (this, get_size (axis_count)); }
 
+  bool unpack_axis_tuples (unsigned axis_count,
```
behdad, line 252: You can also return the hashmap directly. Seems like false is never returned.

src/hb-ot-var-cvar-table.hh
```diff
@@ -51,6 +51,25 @@ struct cvar
...
+                                   bool is_gvar,
+                                   TupleVariationData::tuple_variations_t& tuple_variations /* OUT */) const
+  {
+    hb_vector_t<unsigned> shared_indices;
+    TupleVariationData::tuple_iterator_t iterator;
+    unsigned var_data_length = tupleVariationData.get_size (axis_count);
+    hb_bytes_t var_data_bytes = hb_bytes_t (reinterpret_cast<const char*> (get_tuple_var_data ()), var_data_length);
+    if (!TupleVariationData::get_tuple_iterator (var_data_bytes, axis_count, this,
+                                                 shared_indices, &iterator))
+      return true;
```
behdad, line 65: So, this is not an error condition?

src/hb-ot-var-common.hh
qxliu76: I'd still return a bool: there's a check on input shared_tuples and axis_count, if it's invalid, it'll return false

src/Makefile.am
```diff
@@ -457,6 +457,7 @@ COMPILED_TESTS = \
 	test-repacker \
 	test-classdef-graph \
 	test-instancer-solver \
+	test-decompile-tuple-varstore \
```
behdad, line 460: Please rename the test to `test-tuple-varstore`

src/hb-ot-var-common.hh
```diff
@@ -364,6 +411,132 @@ struct TupleVariationHeader
...
+  /* indices_length = point_count, indice[i] = 1 means point i is referenced */
+  hb_vector_t<bool> indices;
+  
+  hb_vector_t<float> deltas_x;
+  /* empty for cvar tuples */
+  hb_vector_t<float> deltas_y;
+
+  tuple_delta_t () = default;
+  tuple_delta_t (const tuple_delta_t& o) : tuple_delta_t ()
+  { copy (o); }
```
behdad, line 428: Wouldn't the default copy-constructor / copy-assignment work?

behdad:
Thank you.

PR #4269, opened 2023-06-06
[shape] Add dedicated OpenType and AAT shapers

### LINKED ISSUE #4108: Blocklist dysfunctional `morx` table
Currently we use some heuristics to decide whether to use AAT or OT layout table when fonts provide both and our heuristics try to match what CoreText does. All good. However, some fonts have AAT tables for historical reasons (since OSX initially didn’t support OT layout) and on non-mac platforms people were getting OT layout, now they get AAT and it can be unexpected (sometimes the AAT table are even there by accident, FontForge sometimes ”translates” OT to AAT and does it badly).

So, I’m asking for a way to always prefer OT. Currently our `ot` shaper does both OT and AAT layout, so changing shaper list order does not help.

One way, though probably will be ugly internally, is to introduce and new dummy `aat` shaper that simply prefers AAT, and may be make `ot` do the reverse and add a new `default` (or some other name) that is equivalent to current `ot` shaper (which is misnomer already).

Or may be some other way, I don’t really mind, though shaper list is an already existing mechanism and some clients already expose it to users (I know at least luaotfload and SILE).

khaledhosny:
Fixes #4108


khaledhosny:
@behdad is this the direction to take for #4108?

khaledhosny:
@behdad does any of what I wrote above make sense?

behdad:
> @behdad does any of what I wrote above make sense?

I've been travelling. Will respond soon.

behdad:
Okay the main thing I'm uncomfortable with is suddenly making "ot" shaper fail at all. This wasn't the case before. I'm fine if it starts ignoring "aat" tables, but failing might be bad. I know there are a client or more that explicitly pass "ot"...

behdad:
The way I see it is rather:
- "ot" would never apply morx/kerx,
- "aat" would never apply GSUB,
- "hb" is as is today.

shaper order will have "hb" at the top.

behdad:
My reasoning is that "no layout table" is valid OpenType / AAT shaping. So we shouldn't fail.

behdad:
I can see two options for "ot", does any of these satisfy your requirements:
- Ignore AAT tables,
- Prefer OpenType tables.

khaledhosny:
OK, so another idea. Make the new `ot` and `aat` shapers use only the respective layout tables but not fail, and instead of depending on them failing and HarfBuzz falling back to the next shaper, the client code would do something like:
```c++
bool has_aat = hb_aat_layout_has_substitution(face) || hb_aat_layout_has_positioning(face);
bool has_ot = hb_ot_layout_has_substitution(face) || hb_ot_layout_has_positioning(face);

if (has_aat && has_ot)
  // hybrid font, we want OT
  shapers = {"ot", "fallback"};
else
  // Keep HarfBuzz defaults
  shapers = default_shapers;
```

behdad:
> ```c++
> 
> if (has_aat && has_ot)
>   // hybrid font, we want OT
>   shapers = {"ot", "fallback"};
> ```

But if "ot" always tries ot before aat, then you won't even need the conditional, right?

behdad:
> > ```c++
> > if (has_aat && has_ot)
> >   // hybrid font, we want OT
> >   shapers = {"ot", "fallback"};
> > ```
> 
> But if "ot" always tries ot before aat, then you won't even need the conditional, right?

Or is that what you want to avoid?

behdad:
@jfkthame can we get some direction here?

khaledhosny:
> > > ```c++
> > > if (has_aat && has_ot)
> > >   // hybrid font, we want OT
> > >   shapers = {"ot", "fallback"};
> > > ```
> > 
> > 
> > But if "ot" always tries ot before aat, then you won't even need the conditional, right?
> 
> Or is that what you want to avoid?

Right, I wouldn’t need it indeed. Not for my use case here.

khaledhosny:
I don’t need this anymore, if someone else needs it in the future, it will be their problem to handle.

PR #4270, opened 2023-06-07
Add HB_EXTERN to hb_ot_layout_lookup_get_optical_bound

lb90:


lb90:
I thought this would have fixed https://gitlab.gnome.org/lb90/pango/-/jobs/2871856, but actually the error is in Pango. Sorry for the noise!

PR #4271, opened 2023-06-08
[instancer] change tuple variations' axis limits

qxliu76:
- add merge_tuple_variations (), this is not covered by unit test yet.
- fix the issue in hashmap usage: use axis tag as the hashmap key instead of axis index, this makes remove_axis and set_tent faster
- add unit test for change_tuple_variation_axis_limits (). Test is still cvar only,  currently gvar is not supported cause we don't have code to unpack shared_tuples yet. will work on it later.


behdad:
Thanks. Some bot failure; typically mixing unsigned and hb_codepoint_t.

qxliu76:
> Thanks. Some bot failure; typically mixing unsigned and hb_codepoint_t.

thanks! I pushed a fix. The fuzzing failure seems not related to my code change.

Issue #4273, opened 2023-06-09
Hash-pin workflow dependencies


pnacht:
Hey, it's Pedro and I'm back (see #4191 and #4194) with another security suggestion!

When developing with CI workflows, it's common to version-pin dependencies (i.e. `actions/checkout@v3`). However, version tags are mutable, so a malicious attacker could overwrite a version tag to point to a malicious or vulnerable commit instead.

Pinning workflow dependencies by hash (or commit SHA) ensures the dependency is immutable and its behavior is guaranteed.

Your CI workflows have two types of dependencies:
- GitHub Actions (`actions/checkout`, `msys2/setup-msys2`, etc)
- Python packages installed via `pip install ...`

These dependencies can be kept up-to-date with dependabot, which the project has already set up. So your routine won't change at all and workflows will simply be safer. (For example, see the PRs dependabot sent to [my fork](https://github.com/pnacht/harfbuzz/pulls))

I'll send PRs to handle each type of dependency along with this issue.

PR #4275, opened 2023-06-09
Hash-pin Python dependencies in CI

pnacht:
See #4273.

As described in the issue above, this PR ensures the Python dependencies used in GitHub workflows are immutable.

This is done by creating a `requirements.in` file within the `.ci` directory. This file contains all the dependencies used across the workflows, as defined there. From this file, `pip-compile --generate-hashes .ci/requirements.in` creates a `requirements.txt` file that defines the specific hashes expected for each dependency.

The PR also sets up dependabot to keep these dependencies up-to-date. For an example of how that works, see this dependabot PR on my fork: https://github.com/pnacht/harfbuzz/pull/1.


behdad:
Thanks!  Looks like glib requires meson >= 0.60.

behdad:
> I just tested a change on my fork and setting meson == 0.60.0 passes all checks. However, some of the workflows are currently running with meson == 0.56.0.
> 
> Would it be okay to upgrade everything to 0.60.0 or would you rather keep the 0.56.0 ones as they are?

Thanks. I don't know the answer. I leave to @khaledhosny to decide.

khaledhosny:
We currently require meson 0.56.0 because it is the earliest version of meson that can build HarfBuzz (earlier versions have bugs that prevented building HarfBuzz) and we try not to use too versions of meson. 0.60.0 was released less than two years ago. I don’t mind myself what version of meson we use.

behdad:
I think it's fine to hashpin to meson 0.60 while leaving the required versions as they are.

khaledhosny:
> I think it's fine to hashpin to meson 0.60 while leaving the required versions as they are.

One reason we pin required version on (some) CI jobs is too avoid accidentally depending on newer meson version (which does happen and gets caught by CI failures).

behdad:
Sigh. Now pip is failing in installing ninja...

pnacht:
Yep, I saw that, have been investigating it. It's some issue with msys2 or mingw32, dunno if those are supported. I'm creating a separate requirements file for fonttools, since that's the only dependency actually used in that workflow. I'm just making sure dependabot is well-behaved with multiple requirements files before pushing the commit here.

behdad:
Thank you. I'm gone using newer meson in CI.

pnacht:
I assume you meant "fine", so just sent a new commit with that change.

behdad:
Thank you!!

PR #4287, opened 2023-06-20
[instancer] compile tuple variations

qxliu76:
1. Added code to compile tuple_variations_t into bytes, also added unit testing code for it.
    - compile all points_set and store the bytes data in a hashmap in tuple_variations_t, cause some points_set could be reused
    - compile deltas and tuple variation header for each tuple_delta_t, bytes stored within each tuple
    - shared_tuples not supported yet
2. Added subset()/serialize() method in TupleVariationData and cvar for instantiating cvar table, this is not covered by test yet, I'll add a full test once gvar is supported.

### DIFF
--- a/src/hb-ot-var-common.hh
+++ b/src/hb-ot-var-common.hh
@@ -421,32 +421,25 @@ struct TupleVariationHeader
 /* not using hb_bytes_t: avoid potential build issues with some compilers */
 struct byte_data_t
 {
-  const char *p = nullptr; //allocated by hb_calloc () or similar
-  unsigned length = 0;
+  hb_bytes_t bytes;
 
   byte_data_t () = default;
-  byte_data_t (const byte_data_t&) = default;
-  byte_data_t (const char *p_, unsigned len_) : p (p_), length (len_) {}
-  ~byte_data_t () = default;
-  byte_data_t& operator= (const byte_data_t&) = default;
-  byte_data_t& operator= (byte_data_t&&) = default;
+  byte_data_t (const char *p_, unsigned len_) : bytes (hb_bytes_t (p_, len_)) {}
 
-  void fini () { hb_free ((char *) p); p = nullptr; length = 0; }
+  void fini () { bytes.fini (); }
 
   bool operator == (const byte_data_t& o) const
-  { return p == o.p && length == o.length; }
+  { return bytes.arrayZ == o.bytes.arrayZ && bytes.length == o.bytes.length; }
 
-  explicit operator bool () const { return length; }
+  explicit operator bool () const { return bytes.length; }
 
-  byte_data_t* copy (hb_serialize_context_t *c) const
+  void copy (hb_serialize_context_t *c) const
   {
-    if (unlikely (!length || !p)) return nullptr;
+    if (unlikely (!bytes.length || !bytes.arrayZ)) return;
 
-    char* ret = c->allocate_size<char> (length);
-    if (unlikely (!ret)) return nullptr;
-    hb_memcpy (ret, p, length);
-    // cast returning pointer, required by serializer copy()
-    return reinterpret_cast<byte_data_t *> (ret);
+    char* ret = c->allocate_size<char> (bytes.length);
+    if (unlikely (!ret)) return;
+    hb_memcpy (ret, bytes.arrayZ, bytes.length);
   }
 };
 
@@ -472,20 +465,11 @@ struct tuple_delta_t
   /* compiled data: header and deltas
    * compiled point data is saved in a hashmap within tuple_variations_t cause
    * some point sets might be reused by different tuple variations */
-  byte_data_t compiled_tuple_header;
-  byte_data_t compiled_deltas;
+  hb_vector_t<char> compiled_tuple_header;
+  hb_vector_t<char> compiled_deltas;
 
   tuple_delta_t () = default;
   tuple_delta_t (const tuple_delta_t& o) = default;
-  ~tuple_delta_t ()
-  {
-    compiled_deltas.fini ();
-    compiled_tuple_header.fini ();
-    deltas_y.fini ();
-    deltas_x.fini ();
-    indices.fini ();
-    axis_tuples.fini ();
-  }
 
   tuple_delta_t (tuple_delta_t&& o) : tuple_delta_t ()
   {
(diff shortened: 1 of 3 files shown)


behdad:
Humm. Issue persists even if I nuke the avar table.

src/hb-ot-var-common.hh
```diff
@@ -416,6 +418,45 @@ struct TupleVariationHeader
   DEFINE_SIZE_MIN (4);
 };
 
+/* not using hb_bytes_t: avoid potential build issues with some compilers */
+struct byte_data_t
+{
+  const char *p = nullptr; //allocated by hb_calloc () or similar
+  unsigned length = 0;
```
garretrieger, line 425: Could this instead just hold an instance of hb_bytes_t and forward the needed methods to it? That way we don't need to duplicate the functionality of hb_bytes_t. I think that shouldn't trigger the compilation issues since it won't have an overloaded & operator.

src/hb-ot-var-common.hh
```diff
@@ -416,6 +418,45 @@ struct TupleVariationHeader
...
+  byte_data_t (const byte_data_t&) = default;
+  byte_data_t (const char *p_, unsigned len_) : p (p_), length (len_) {}
+  ~byte_data_t () = default;
+  byte_data_t& operator= (const byte_data_t&) = default;
+  byte_data_t& operator= (byte_data_t&&) = default;
+
+  void fini () { hb_free ((char *) p); p = nullptr; length = 0; }
+
+  bool operator == (const byte_data_t& o) const
+  { return p == o.p && length == o.length; }
```
garretrieger, line 437: Should equals compare the bytes instead of the pointer? If you switch to using a hb_bytes_t instance then you can use the == method from hb_bytes_t which should do this. Also I believe we'll need a hash() method if this is to be used as a hashmap key.

src/hb-ot-var-common.hh
```diff
@@ -528,6 +584,326 @@ struct tuple_delta_t
...
+    return true;
+  }
+
+  F2DOT14* encode_peak_coords (char* p, F2DOT14* end,
+                               unsigned& flag,
+                               const hb_map_t& axes_index_map,
+                               const hb_map_t& axes_old_index_tag_map) const
+  {
+    unsigned orig_axis_count = axes_old_index_tag_map.get_population ();
+    F2DOT14* out = reinterpret_cast<F2DOT14 *> (p);
```
garretrieger, line 636: Instead of passing around and accessing the raw pointers directly you could use hb_array_t<Type> which will provide automatic bounds checking. Since arrays are iterators they also support things like ++, so most of the code interacting with them shouldn't really need to change.

src/hb-ot-var-common.hh
```diff
@@ -428,8 +469,23 @@ struct tuple_delta_t
   /* empty for cvar tuples */
   hb_vector_t<float> deltas_y;
 
+  /* compiled data: header and deltas
+   * compiled point data is saved in a hashmap within tuple_variations_t cause
+   * some point sets might be reused by different tuple variations */
+  byte_data_t compiled_tuple_header;
+  byte_data_t compiled_deltas;
```
garretrieger, line 476: Can you use hb_vector_t<char>'s here? They provide automatic management/ownership of the memory. The vectors could be easily converted to byte_data_t when needed for inclusion in hashmaps.

src/hb-ot-var-common.hh
```diff
@@ -528,6 +584,326 @@ struct tuple_delta_t
...
+  bool compile_tuple_var_header (const hb_map_t& axes_index_map,
+                                 unsigned points_data_length,
+                                 const hb_map_t& axes_old_index_tag_map)
+  {
+    if (!compiled_deltas) return false;
+
+    unsigned cur_axis_count = axes_index_map.get_population ();
+    /* allocate enough memory: 1 peak + 2 intermediate coords + fixed header size */
+    unsigned alloc_len = 3 * cur_axis_count * (F2DOT14::static_size) + 4;
+    char *p = (char *) hb_calloc (alloc_len, sizeof (char));
```
garretrieger, line 601: Likewise here, you could use a vector_t<char> + a call to resize() to allocate the memory. That way the vector would own the memory. At the end you would move it to compiled_tuple_header. This would then ensure memory is automatically cleaned up so there's no need to call hb_free(p) before each return.

Same in the other compile functions that allocate memory.

src/hb-ot-var-common.hh
qxliu76: We need to reinterpret_cast between char* and F2DOT14* or HBINT16* when compiling deltas, it's easier if we could just use raw pointers. I'd like to avoid hb_vector_t here if you agree.

src/hb-ot-var-common.hh
qxliu76: My thought is to allocate bytes once, and free them only during destruction. Copy and assignment operations of the struct just copy the pointer value, so I think equals doesn't need to compare the bytes, maybe just check the pointer is enough in this case?

src/hb-ot-var-common.hh
qxliu76: uh..maybe I can still try to use hb_vector_t, just pre-allocate enough memory and then operate on its raw arrayZ. Would that be better?

src/hb-ot-var-common.hh
garretrieger: Are you planning to use these as hash map keys in future changes? If you are then we'll need to the value comparison equals. If they are only planned to be used as hashmap values then this equals is fine.

src/hb-ot-var-common.hh
garretrieger: Yes, that would be better. To handle the casting case you could add a templated helper method which takes hb_vector_t<char>& and returns a hb_array_t<T> (it would cast the arrayZ and create a hb_array_t with it). Then you could easily convert to hb_array_t's of the appropriate type and operate on those instead of the raw pointers.

src/hb-ot-var-common.hh
```diff
@@ -681,6 +702,88 @@ struct TupleVariationData
...
+          while (i < indices_length && !point_indices[i])
+            i++;
+
+          if (i >= indices_length) break;
+
+          unsigned cur_value = i;
+          unsigned delta = cur_value - last_value;
+
+          if (!use_byte_encoding)
+            use_byte_encoding = (delta >= 0 && delta <= 0xFF);
```
behdad, line 1136: I'm suspicious of this line.

src/hb-ot-var-common.hh
```diff
@@ -681,6 +702,88 @@ struct TupleVariationData
...
+
+          if (use_byte_encoding && (delta <0 || delta > 0xFF))
```
behdad, line 1138: delta is unsigned, so cannot be negative.

src/hb-ot-var-common.hh
```diff
@@ -456,8 +456,23 @@ struct tuple_delta_t
...
+  byte_data_t compiled_deltas;
+
   tuple_delta_t () = default;
   tuple_delta_t (const tuple_delta_t& o) = default;
+  ~tuple_delta_t ()
+  {
+    compiled_deltas.fini ();
+    compiled_tuple_header.fini ();
+    deltas_y.fini ();
+    deltas_x.fini ();
```
behdad, line 485: You don't need the vector fini's...

src/hb-ot-var-common.hh
behdad: hb_vector_t has operators for hb_array_t if I remember correctly.

src/hb-ot-var-common.hh
```diff
@@ -437,6 +437,17 @@ struct byte_data_t
...
+
+  byte_data_t* copy (hb_serialize_context_t *c) const
+  {
+    if (unlikely (!length || !p)) return nullptr;
+
+    char* ret = c->allocate_size<char> (length);
+    if (unlikely (!ret)) return nullptr;
+    hb_memcpy (ret, p, length);
+    // cast returning pointer, required by serializer copy()
+    return reinterpret_cast<byte_data_t *> (ret);
```
behdad, line 449: But this line is wrong! Better to return nullptr. No?

behdad: Can you return void here and call the copy() directly, instead of calling it through the serializer?

src/hb-ot-var-cvar-table.hh
```diff
@@ -125,6 +125,44 @@ struct cvar
...
+    if (unlikely (!c->embed (version))) return_trace (false);
+
+    return_trace (tupleVariationData.serialize (c, false, tuple_variations));
+  }
+
+  bool subset (hb_subset_context_t *c) const
+  {
+    TRACE_SUBSET (this);
+    /* subset() for cvar is called by partial instancing only, we always pass
+     * through cvar table in other cases */
```
behdad, line 142: Humm, I expect that returning false would drop the table, no?

src/hb-ot-var-cvar-table.hh
```diff
@@ -125,6 +125,44 @@ struct cvar
...
+    if (unlikely (!c->plan->normalized_coords ||
```
behdad, line 143: Also why `unlikely`? Suggest removing.

src/hb-ot-var-common.hh
qxliu76: no, they're just used as hashmap values

src/hb-ot-var-common.hh
```diff
@@ -528,6 +568,322 @@ struct tuple_delta_t
...
+                             hb_array_t<char> encoded_bytes,
+                             const hb_vector_t<int>& deltas) const
+  {
+    unsigned num_deltas = deltas.length;
+    unsigned encoded_len = 0;
+    while (i < num_deltas)
+    {
+      int val = deltas[i];
+      if (val == 0)
+        encoded_len += encode_delta_run_as_zeroes (i, encoded_bytes.sub_array (encoded_len), deltas);
```
behdad, line 736: While the user of `hb_array_t` here is cleaner and Garret suggested it. It's typically slower in a measurable way for such functions. I can fix it later when we profile things.

behdad (approved):

behdad:
This looks good to me. @qxliu76 Are you done with it?

qxliu76:
> This looks good to me. @qxliu76 Are you done with it?

yes

src/hb-ot-var-common.hh
garretrieger: Yeah, my preference is to start with the memory safe versions and then switch to raw access where it pops up in profiles.

Issue #4309, opened 2023-07-05
Graphite returns incorrect advance widths for Annapurna SIL font


khaledhosny:
Testing with https://software.sil.org/downloads/r/annapurna/AnnapurnaSIL-1.204.zip, in the following sequence, the last glyph has a very small advance width when using `graphite2` shaper (which is the default since the font has Graphite tables):

```
$ hb-shape AnnapurnaSIL-Regular.ttf -u 0930,094D,0935,093E,0923
[uni0935=0+1886|uni093E=0@-555,0+0|uni0930094D.reph=0@-46,0+0|uni0923=4+36]
```
![graphite2](https://github.com/harfbuzz/harfbuzz/assets/93914/b786703d-2962-42d7-8c44-c51fca5afd1a)

With `ot` shaper it has a more sensible advance:
```
$ hb-shape AnnapurnaSIL-Regular.ttf -u 0930,094D,0935,093E,0923 --shaper=ot
[uni0935=0+1331|uni093E=0+555|uni0930094D.reph=0+2|uni0923=4+1367]
```
![ot](https://github.com/harfbuzz/harfbuzz/assets/93914/20cf561d-747a-464b-a1dd-95bdfdb37ddf)

I’m not sure if it is a font, Graphite, or HarfBuzz bug, though.

khaledhosny:
It seems to be a bug in the `graphite2` shaper, namely in the code that maps Graphite advances to HarfBuzz one. I don’t know Graphite 2 to API to grok that code, though.

cc @mhosken

khaledhosny:
Another similar issue with [Linux Libertine G](http://www.numbertext.org/linux/e7a384790b13c29113e22e596ade9687-LinLibertineG-20120116.zip) font. When a tt ligature is followed by space the positioning becomes odd:
```
$ hb-shape LinLibertine_R_G.ttf "tt"
[t_t=0+1257]
$ hb-shape LinLibertine_R_G.ttf "tt "
[t_t=0+1769|space=0@-512,0+0]
```

khaledhosny:
This diff is almost certainly wrong, but without I get saner glyph advances:

```diff
diff --git a/src/hb-graphite2.cc b/src/hb-graphite2.cc
index 9bb9136c4..ce324c46d 100644
--- a/src/hb-graphite2.cc
+++ b/src/hb-graphite2.cc
@@ -223,7 +223,6 @@ struct hb_graphite2_cluster_t {
[...]

PR #4310, opened 2023-07-05
[graphite] Simplify and fix glyph position handling

khaledhosny:
Instead of trying to accumulate the advance width of all the glyphs in a cluster, assign it to the first glyph then adjust the advance widths of the rest of the glyphs in the cluster, assign each glyph its own advance width since this accumulation is not necessary and seems to result in wrong text advance in many cases.

Vertical advance for RTL text was also being subtracted while in LTR it was added, which seems wrong (it probably does not matter, since vertical advances are only used for vertical text).

### DIFF
--- a/src/hb-graphite2.cc
+++ b/src/hb-graphite2.cc
@@ -263,9 +263,6 @@ _hb_graphite2_shape (hb_shape_plan_t    *shape_plan HB_UNUSED,
   }
 
   gr_segment *seg = nullptr;
-  const gr_slot *is;
-  unsigned int ci = 0, ic = 0;
-  int curradvx = 0, curradvy = 0;
 
   unsigned int scratch_size;
   hb_buffer_t::scratch_buffer_t *scratch = buffer->get_scratch_buffer (&scratch_size);
@@ -330,7 +327,9 @@ _hb_graphite2_shape (hb_shape_plan_t    *shape_plan HB_UNUSED,
   float xscale = (float) font->x_scale / upem;
   float yscale = (float) font->y_scale / upem;
   yscale *= yscale / xscale;
-  for (is = gr_seg_first_slot (seg), ic = 0; is; is = gr_slot_next_in_segment (is), ic++)
+  unsigned int ci = 0, ic = 0;
+  const gr_slot *is;
+  for (is = gr_seg_first_slot (seg); is; is = gr_slot_next_in_segment (is), ic++)
   {
     unsigned int before = gr_slot_before (is);
     unsigned int after = gr_slot_after (is);
@@ -372,10 +371,10 @@ _hb_graphite2_shape (hb_shape_plan_t    *shape_plan HB_UNUSED,
   buffer->len = glyph_count;
 
   /* Positioning. */
+  int curradvx = 0, curradvy = 0;
   hb_glyph_position_t *pPos = hb_buffer_get_glyph_positions (buffer, nullptr);
   if (!HB_DIRECTION_IS_BACKWARD (direction))
   {
-    curradvx = 0;
     for (is = gr_seg_first_slot (seg); is; pPos++, is = gr_slot_next_in_segment (is))
     {
       pPos->x_offset = gr_slot_origin_X (is) * xscale - curradvx;
--- a/src/hb-graphite2.cc
+++ b/src/hb-graphite2.cc
@@ -223,7 +223,6 @@ struct hb_graphite2_cluster_t {
   unsigned int base_glyph;
   unsigned int num_glyphs;
   unsigned int cluster;
-  int advance;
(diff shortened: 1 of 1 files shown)


khaledhosny:
I’m not sure what is the best way to test change like this. The usual shaping tests won’t be helpful as we will be getting different advances and offsets even if the final positions are unchanged.

khaledhosny:
I’m seeing improved handling of advanced in many multi-glyph clusters over several graphite fonts, so I’m positive this change is a good one.

behdad:
Thanks. I'll take a look.

Any chance you can apply similar changes to the wasm graphite shaper?

mhosken:
I've built this and run it in libreoffice and I see this for Awami Nastaliq:
image

The characters covered by the selected range are: 06D4 0020 0646 06CC 06A9 0020 0645 0650 0646 06A9 06BE 0020 0631 06CC at the start of line. There seems to be some misalignment between what is rendered and what is thought to be rendered.

khaledhosny:
> The characters covered by the selected range are: 06D4 0020 0646 06CC 06A9 0020 0645 0650 0646 06A9 06BE 0020 0631 06CC at the start of line. There seems to be some misalignment between what is rendered and what is thought to be rendered.

Can you share with me the full document?

There have been other changes to the `hb-graphite2.cc` file since the version of HarfBuzz currently use by LibreOffice, so I prepared a new tarball to test there, [harfbuzz-7.4.0.tar.xz.zip](https://github.com/harfbuzz/harfbuzz/files/11979410/harfbuzz-7.4.0.tar.xz.zip) (remove the `.zip` extension it is here to fool GitHub). Still waiting for my build to finish.

mhosken:
Can you email me a request so I can reply.

mhosken:
[GEN1_urdArab.odt](https://github.com/harfbuzz/harfbuzz/files/11979712/GEN1_urdArab.odt)

khaledhosny:
Thanks for the document. I can see the issue now. Debugging...

khaledhosny:
I think I’m probably misunderstanding graphite positioning data. In sequences like `سے`, I get the following:
```
seg_advance_X=2494
555: slot_advance_X=1345, gr_slot_origin_X=134
1400: slot_advance_X=2494, gr_slot_origin_X=0
```
My change is assuming that `seg_advance_X` is equal to the sum of `slot_advance_X`’s, but apparently this is not true.

mhosken:
From memory, the advance of a glyph is it's hmtx width as modified by
graphite rules. So. No the width of a segment is not the sum of its
constituent glyph advances.

There is some complexity when it comes to clusters and their advances too.
I'm sorry, I don't have time atm to give the full story. Iirc correctly
(sorry it's been a while but the start of an Arabic cluster, being the last
base in the sequence will have the width of the whole cluster, if there is
no insertion points in the cluster.)

The easiest approach is to work from glyph positions.

If you want, I can work to give a more complete picture.



GB,
Martin

mhosken:
Graphite was designed before OpenType had taken off and has much more flexibility built in. Thus it doesn't think in terms of irreducible clusters like OT does. Instead it thinks in terms of cursor insertion points. So, for example it even allows a cursor point before an attached diacritic. It was designed to allow for split cursors, diacritic cursors and all that kind of funky stuff that OpenType just dropped on the floor. The real need probably lies somewhere between the two.

This means that the advance of a glyph is nothing more than an internal metric that is used to help arrive at the cluster width. Then clusters are assembled left to right based on their width. The base of a cluster returns NULL to gr_slot_attached_to(). There is no easy way with the current API to get the width of a cluster but one can 'simply' take the position of the start of the next cluster (or segment width) and subtract the position of the start of this cluster.

So as to the fixing of the harfbuzz adaptor, the current algorithm works. If we want to be a little more subtle, we could break down the advances within the OT cluster based on whether a glyph is a base or not. This would involve a more subtle approach to the use of info->var1.i32. But given the, hopeful, transition to WASM, we may get to refactor it there and optimise the whole thing towards whatever the agreed WASM model is.

khaledhosny:
Thanks for the explanation. So the change here is a dead end and I’m closing this pull request.

Can you open a pull request for the change you proposed in https://github.com/harfbuzz/harfbuzz/issues/4309#issuecomment-1624730406, or should I do it?

khaledhosny:
> Thanks for the explanation. So the change here is a dead end and I’m closing this pull request.
> 
> Can you open a pull request for the change you proposed in [#4309 (comment)](https://github.com/harfbuzz/harfbuzz/issues/4309#issuecomment-1624730406), or should I do it?

I submitted https://github.com/harfbuzz/harfbuzz/pull/4314

mhosken:
Thanks

PR #4318, opened 2023-07-10
Add -Wconstant-conversion to errors

behdad:


behdad:
No idea why this doesn't fail for me locally. But I got report that this fails for others. Let's see if bots fail.

PR #4323, opened 2023-07-14
[subset] begin implementing omit-glyf mode.

garretrieger:
When omit glyf is enabled the subsetter won't actually output glyf table bytes. However, it will still have an entry for the glyf table in the table directory as if the glyf table was there, including the correct checksum and length. This is intended for use with patch subset incxfer where a custom diffing implementation will source the glyph bytes as needed from the original font.

Currently this computes the checksum by inspecting the glyph bytes. However, we could popuate a data structure in the accelerator to significantly accelerate the checksum computation (may require forcing 4 byte alignments of glyphs).

Not yet fully implemented.

### DIFF
--- a/src/OT/glyf/SubsetGlyph.hh
+++ b/src/OT/glyf/SubsetGlyph.hh
@@ -26,19 +26,17 @@ struct SubsetGlyph
   {
     TRACE_SERIALIZE (this);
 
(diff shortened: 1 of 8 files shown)


garretrieger:
Going to abandon this, showing only modest performance improvements (~5%) so not likely worth the added complexity.

PR #4337, opened 2023-07-21
[gsubgpos] Try at caching check_glyph_properties() result

behdad:
I was hoping this would (significantly) speed up Duployan, but it doesn't. Apparently the recursed-to lookups do the heavy lifting and they have different lookup-props, so they cannot share the cache :(.

Recording here for posterity.

### DIFF
--- a/src/hb-ot-layout-gpos-table.hh
+++ b/src/hb-ot-layout-gpos-table.hh
@@ -63,12 +63,17 @@ inline bool PosLookup::dispatch_recurse_func<hb_ot_apply_context_t> (hb_ot_apply
   c->set_lookup_index (lookup_index);
   c->set_lookup_props (l.get_props ());
 
+  bool saved_props_cached = c->props_cached;
+  if (saved_lookup_props != c->lookup_props)
+    c->props_cached = false;
+
   bool ret = false;
   auto *accel = gpos->get_accel (lookup_index);
   ret = accel && accel->apply (c, l.get_subtable_count (), false);
 
   c->set_lookup_index (saved_lookup_index);
   c->set_lookup_props (saved_lookup_props);
+  c->props_cached = saved_props_cached;
   return ret;
 }
 #endif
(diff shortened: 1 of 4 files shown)


dscorbett:
Which are the lookups with different properties that would otherwise have benefited from this optimization? Maybe I can refactor them to have the same properties.

behdad:
Thanks. I need to study the font more...

Maybe you can illuminate us on how the font works to begin with? Where I see the time is spent is, I think, skipping over lots of the phantom marks.

dscorbett:
The time-consuming part is, I suspect, [the width system](https://github.com/dscorbett/duployan-font/blob/master/docs/width-system.md). It involves adding lots of numbers, where every digit is represented as a glyph. Multiple sums are calculated concurrently. To avoid one sum interfering with another sum, their lookups use different mark filtering sets to specify which digit glyphs are relevant to each sum.

behdad:
I see! Thanks.

behdad:
> Which are the lookups with different properties that would otherwise have benefited from this optimization? Maybe I can refactor them to have the same properties.

I don't think it's possible. The main lookup has a different mark-filter-set than the recursed lookups, which has to be intentional... (hex numbers, top-level lookup then the recursed-to):
```
620010 70014
620010 100014
620010 190014
620010 220014
620010 2b0014
620010 340014
620010 3b0014
```
The top byte is the mark-filtering-set index.

dscorbett:
Would it help if the recursed-to lookups all shared the same mark filtering set, different from the main lookup’s mark filtering set? I could probably make that work. Or does the cache only help specifically when a recursed-to lookup shares its main lookup’s mark filtering set?

behdad:
> Would it help if the recursed-to lookups all shared the same mark filtering set, different from the main lookup’s mark filtering set? I could probably make that work. Or does the cache only help specifically when a recursed-to lookup shares its main lookup’s mark filtering set?

With my current implementation idea it wouldn't help. I have room for storing (caching) one lookup's matching result. So for now I do that for the main lookup. Any recursed-to lookups that share the same lookup-props can keep using the cache.

PR #4345, opened 2023-07-27
cosmetic change

qxliu76:


src/hb-ot-var-common.hh
```diff
@@ -1222,9 +1222,8 @@ struct TupleVariationData
       TRACE_SERIALIZE (this);
       for (const auto& tuple: tuple_vars)
       {
-        hb_bytes_t compiled_bytes {tuple.compiled_tuple_header.arrayZ, tuple.compiled_tuple_header.length};
-        compiled_bytes.copy (c);
-        if (c->in_error ()) return_trace (false);
+        if (!c->embed (tuple.compiled_tuple_header.arrayZ, tuple.compiled_tuple_header.length))
```
behdad, line 1225: Why not use `compiled_tuple_header.copy` here?

src/hb-ot-var-common.hh
```diff
@@ -1240,10 +1239,10 @@ struct TupleVariationData
         if (!point_data_map.has (points_set, &point_data))
           return_trace (false);
 
-        point_data->copy (c);
-        hb_bytes_t compiled_bytes {tuple.compiled_deltas.arrayZ, tuple.compiled_deltas.length};
-        compiled_bytes.copy (c);
-        if (c->in_error ()) return_trace (false);
+        if (!c->embed (point_data->arrayZ, point_data->length))
```
behdad, line 1242: Same.

src/hb-ot-var-common.hh
qxliu76: oh I did not find a copy in hb_vector_t. If we use serializer's copy, it copies the length of the vector as well, and I'd like to avoid that

behdad: Yeah I think if you do `as_array().copy()` that works.

qxliu76: oh...I see! Then it's just a cosmetic change now. no functional change.

PR #4349, opened 2023-07-28
[atomic] Don't disable memory-barrier for MSVC

behdad:


behdad:
Still internal-compiler-error from msvc2019. Closing.

PR #4350, opened 2023-07-30
Require numerators and denominators in automatic fractions

dscorbett:
HarfBuzz automatically applies fraction features to any instance of U+2044 plus any decimal digits that precede or follow it. However, [_The Unicode Standard_, version 15.0, chapter 6](https://www.unicode.org/versions/Unicode15.0.0/ch06.pdf#page=21) says “The standard form of a fraction built using the fraction slash is defined as follows: any sequence of one or more decimal digits (General Category = Nd), followed by the fraction slash, followed by any sequence of one or more decimal digits.” That is, there must be at least one digit in the numerator and at least one digit in the denominator.

### DIFF
--- a/src/hb-ot-shape.cc
+++ b/src/hb-ot-shape.cc
@@ -765,6 +765,8 @@ hb_ot_shape_setup_masks_fraction (const hb_ot_shape_context_t *c)
 	     _hb_glyph_info_get_general_category (&info[end]) ==
 	     HB_UNICODE_GENERAL_CATEGORY_DECIMAL_NUMBER)
 	end++;
+      if (start == i || end == i + 1)
+	continue;
 
       buffer->unsafe_to_break (start, end);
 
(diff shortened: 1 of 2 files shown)


behdad:
Oh I merged this while mingw bots failing. @dscorbett any idea why?

behdad:
I think we need some unsafe_to_concat around this as well.

dscorbett:
I don’t know why the msys2 tests failed. On `main`, only one of them failed. Could there be a caching problem?

behdad:
Humm. I don't know. I reverted this. Will try to fix the unsafe_to_concat and then try merging again.

behdad:
Reverted the revert. Fingers crossed.

behdad:
Okay bots happy now.

behdad:
Humm. It failed again. Not sure what's up.
https://github.com/harfbuzz/harfbuzz/actions/runs/5707741152

khaledhosny:
```
Actual:   [one.numr=0+600|two.numr=1+600|three.numr=2+600|fraction=3+252]
Expected: [one=0+1090|two=1+1090|three=2+1090|fraction=3+252]
```

So for some reason the new code is not working on this build.

khaledhosny:
I wanted to cut a release today, should we revert this for the release and fix later on main?

behdad:
> I wanted to cut a release today, should we revert this for the release and fix later on main?

Sounds good to me. I have no idea what's possibly wrong though.

behdad:
The bot issue should be fixed.

PR #4361, opened 2023-08-03
[COLR] Implement paint templates

behdad:
Based on googlefonts/colr-gradients-spec#371

See https://github.com/behdad/colr-templates for usage.


behdad:
FontTools PR that can generate this format.
https://github.com/fonttools/fonttools/pull/3242

PR #4363, opened 2023-08-06
COLRv2

behdad:
Counterpart to FontTools PR:
https://github.com/fonttools/fonttools/pull/3244


behdad:
Subsetting `PaintGlyphDelta` is tricky. To be determined.

PR #4368, opened 2023-08-07
Initial implementation for gvar partial instancing

qxliu76:
Sorry for this large PR, I didn't realize it's so big... 

- This PR is just an initial implementation, for correctness. Currently in glyf table, we update glyph outlines using the result from get_points (), which iterates all tuples, infer deltas and adds up all deltas on all points. To speed up, we might try to modify the glyph outline points using the solver output.

- This implementation doesn't do iup delta optimization, so I generate tests with `"--no-optimize"` option in fonttools.

- When merging tuples with the same tent, if a point is not referenced by any tuple, it will not be referenced/saved for this tent in the output file. My understanding is this would save some bytes. Thus we see some differences in the ttx files, because fonttools will reference all points if --no-optimize option is used.

- There're some rounding differences on some points in the ttx file when compared to fonttools result


garretrieger (approved):

test/subset/meson.build
```diff
@@ -65,6 +65,7 @@ tests = [
   'collect_name_ids',
   'instantiate_colrv1',
   'instantiate_cff2_update_metrics',
+  #'glyf_partial_instancing',
```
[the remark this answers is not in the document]
qxliu76: The API hb_subset_input_set_axis_range () and according command parsing code are currently wrapped by experimental_api, so I disabled this test for now.

behdad:
> * When merging tuples with the same tent, if a point is not referenced by any tuple, it will not be referenced/saved for this tent in the output file. My understanding is this would save some bytes. Thus we see some differences in the ttx files, because fonttools will reference all points if --no-optimize option is used.

I'm not sure this optimization is correct.

qxliu76:
> > * When merging tuples with the same tent, if a point is not referenced by any tuple, it will not be referenced/saved for this tent in the output file. My understanding is this would save some bytes. Thus we see some differences in the ttx files, because fonttools will reference all points if --no-optimize option is used.
> 
> I'm not sure this optimization is correct.

I'll make changes to do the same thing as fonttools --no-optimize

qxliu76:
I have added changes to reference all points for gvar. 
Previous behavior is: if points are not referenced by any tuple and deltas are not inferred as well, I skip these points in the output file. Fonttools code sets them to 0: https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/varLib/iup.py#L110

Now harfbuzz is doing the same thing as fonttools.

src/hb-ot-var-common.hh
```diff
@@ -914,6 +914,99 @@ struct tuple_delta_t
...
+    hb_vector_t<unsigned> end_points;
+
+    for (unsigned i = 0; i < point_count; i++)
+    {
+      if (indices.arrayZ[i])
+        ref_count++;
+      if (orig_points.arrayZ[i].is_end_point)
+        end_points.push (i);
+    }
+    /* all points are referened, nothing to do */
```
behdad, line 940: Typo.

src/hb-ot-var-common.hh
```diff
@@ -914,6 +914,99 @@ struct tuple_delta_t
...
+    if (ref_count == point_count)
+      return true;
+    if (unlikely (end_points.in_error ())) return false;
+
+    hb_set_t inferred_idxes;
```
behdad, line 945: Any chance the IUP code can be shared with gvar table?

behdad:
Looks generally good to me. Great work!

src/hb-ot-var-common.hh
qxliu76: yes, once initial implementation is done, I will try to work on sharing iup code as much as possible.

behdad: > yes, once initial implementation is done, I will try to work on sharing iup code as much as possible.

Sounds good.

behdad:
Is this ready now?

qxliu76:
> Is this ready now?

yes

PR #4370, opened 2023-08-08
Re-do and clean up CMake support, making it use relocatable paths

### LINKED ISSUE #4025: absolute paths in cmake config
Change db292f6f0238581a489aa8cddc585129b6e920cd makes cmake config use absolute paths and this breaks things for cross-compilations when harfbuzz is in a sysroot and not in the absolute path used at configure time.

Compilations will then try to use the lib and includes from the host system paths, when they should use everything only from the sysroot.

2xsaiko:
Please test (@sapiippo), it looks to me like it now generates relocatable paths in all cases but I only have NixOS so I can't test your sysroot problem :)

Closes #4025.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -385,6 +385,26 @@ foreach check : check_funcs
   endif
 endforeach
 
+# CMake support (package install dir)
+
+# Equivalent to configure_package_config_file(INSTALL_DESTINATION ...), see
+# https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#command:configure_package_config_file.
+# In certain unusual packaging layouts such as Nixpkgs, the Harfbuzz package
+# is installed into two Nix store paths, "out" and "dev", where "out" contains
+# libraries only (i.e. lib/libharfbuzz.so) and "dev" contains development
+# files, i.e. include and lib/cmake. If CMake package files are installed to
+# "out", Nixpkgs will move them to "dev", which breaks assumptions about
+# our file paths. Since we need to figure out relative install paths here
+# to make a relocatable package, we do need to know the final path of our
+# CMake files to calculate the correct relative paths.
+# Of course, this still defaults to $libdir/cmake if unset, which works for
+# most packaging layouts.
+cmake_package_install_dir = get_option('cmakepackagedir')
+
+if cmake_package_install_dir == ''
+  cmake_package_install_dir = get_option('libdir') / 'cmake'
+endif
+
 subdir('src')
 
 if not get_option('utilities').disabled()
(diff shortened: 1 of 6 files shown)


src/meson.build
```diff
@@ -789,15 +791,97 @@ endif
...
+  cmake_package_install_dir = get_option('libdir') / 'cmake'
+endif
+
+cmake_config_dir = cmake_package_install_dir / 'harfbuzz'
+
+# TODO: Replace these with
+# fs.relative_to(get_option('prefix') / cmake_config_dir, get_option('prefix'))
+# etc. once implemented
+# See https://github.com/mesonbuild/meson/pull/12087
+relative_to = find_program('relative_to.py')
```
LuxxxLucy, line 820: PR#12087 of meson is merged, so now this should be replaced with `fs.relative_to). And the temporary solution `relative_to.py` can be pruned from the commit as it becomes depreciated.

2xsaiko: Ah, nice! It'll be there in Meson 1.3.0 which isn't released yet, but I can update the PR already.

meson.build
```diff
@@ -1,5 +1,5 @@
 project('harfbuzz', 'c', 'cpp',
-  meson_version: '>= 0.55.0',
+  meson_version: '>= 1.3.0',
```
alerque, line 2: This is a pretty major jump. It might be justifiable (and I haven't looked into specifics) but I do recall a while back (~1.5 years?) there was some blocker downstream that needed old Meson support. Perhaps that was an Ubuntu LTS release? I don't actually remember, but just want to check that we're sure moving to 1.3.0 isn't going to leave any distros that switched from autotools to meson out in the cold.

LuxxxLucy:
> Does this change make us require newer meson? That wouldn't fly since we have to support very old systems as well.

In this case perhaps we need to revert to the previous way of using a python script (os.path.relpath). @2xsaiko Sorry my bad.

eli-schwartz:
Maybe just do `if meson.version().version_compare('>=1.3.0')` and use the new optimized function, or fall back to an approach available on older mesons otherwise.

2xsaiko:
> Maybe just do `if meson.version().version_compare('>=1.3.0')` and use the new optimized function, or fall back to an approach available on older mesons otherwise.

What I just pushed should do that, I hope.

behdad:
Please add `relative_to.py` to Makefile.am's `EXTRA_DIST`.

2xsaiko:
Hmm, despite the other python files run from Meson not being in there? I assume it should be written as `src/relative_to.py`.

behdad:
> Hmm, despite the other python files run from Meson not being in there?

Like which one? This is why the distcheck bot is failing.

> I assume it should be written as `src/relative_to.py`.

No. Add it to `src/Makefile.am` and just `relative_to.py`.

2xsaiko:
Ahh, I was looking at the Makefile.am in the root. My bad.

PR #4384, opened 2023-08-22
WIP [COLR] Add PaintLocation

behdad:
Doesn't work...

### DIFF
--- a/src/OT/Color/COLR/COLR.hh
+++ b/src/OT/Color/COLR/COLR.hh
@@ -34,6 +34,8 @@
 #include "../../../hb-paint.hh"
 #include "../../../hb-paint-extents.hh"
 
+#include "../../../OT/glyf/coord-setter.hh"
+
 /*
  * COLR -- Color
  * https://docs.microsoft.com/en-us/typography/opentype/spec/colr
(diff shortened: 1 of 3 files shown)


behdad:
This is near impossible to make work. Definitely not with our current abstractions and APIs.

The problem is that `PaintLocation` cannot just set new coordinates on `hb_font_t`, for two reasons:
- It would break the threadsafety of `hb_font_t`, which is a non-starter.
- Even if we install new coordinates on `hb_font_t`, then `hb_font_t` attachments (a cairo scaled font, etc) won't notice that the `hb_font_t` changed underneath them.

Another approach would be to create a new `hb_font_t` and install the locations on it. But in general we don't have a `hb_font_create_similar()`. The `hb_font_create_sub_font()` is not sufficient, since it won't have proper font-funcs installed on it.

So to do this, we would need to add a `hb_font_draw_at_location()` which would be very ugly.

Yet another approach, would be to add `push_location` / `pop_location` to the hb-paint API, but that would have its own problems I believe.

behdad:
FontTools counterpart. https://github.com/fonttools/fonttools/pull/3262

behdad:
> FontTools counterpart. [fonttools/fonttools#3262](https://github.com/fonttools/fonttools/pull/3262)

The patch in this PR doesn't quite match the fonttools code. In particular `AxisList` has a Format field in the fonttools impl, that this patch doesn't account for. But as I said in the previous comment, it's near impossible to make it work currently anyway. So I'm abandoning it.

PR #4385, opened 2023-08-23
[repacker] fix case of failing repacker.

garretrieger:
See: https://github.com/fonttools/fonttools/issues/3260

This font has huge number of subtables, but many are de-dupped. This fixes the repacking failure by ensuring that created extension subtables are shared where possible.

### DIFF
--- a/src/graph/graph.hh
+++ b/src/graph/graph.hh
@@ -474,6 +474,18 @@ struct graph_t
     return root ().equals (other.root (), *this, other, 0);
   }
 
+  void print () const {
+    for (int i = vertices_.length - 1; i >= 0; i--)
+    {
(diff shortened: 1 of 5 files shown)


behdad:
Looks good. Thanks.

dscorbett:
[u0-Regular.otf and u1-Regular.otf](https://github.com/harfbuzz/harfbuzz/files/12432135/repacker-variants.zip) are two variants of the same font built with uharfbuzz before and after this change. GSUB went from 738108 bytes to 740136 bytes. The difference is that u1-Regular.otf uses many more extension lookups. It’s not a big deal, but it is surprising, since this PR is about reducing the number of extensions. Is this change expected?

PR #4386, opened 2023-08-24
1 fuzzer fix

qxliu76:


src/hb-ot-var-common.hh
```diff
@@ -1031,10 +1031,13 @@ struct TupleVariationData
...
-    TRACE_SANITIZE (this);
-    // here check on min_size only, TupleVariationHeader and var data will be
+    // here check on min_size and TupleVariationHeader array only, var data will be
     // checked while accessing through iterator.
-    return_trace (c->check_struct (this));
+    TRACE_SANITIZE (this);
+    if (!(c->check_struct (this)))
+        return_trace (false);
+
+    return_trace (c->check_range (this, min_size + TupleVariationHeader::min_size * (tupleVarCount.get_count ())));
```
behdad, line 1040: Should that be `static_size` instead of `min_size`?

behdad: Humm. I'm not sure about this change. Why is it needed again?

src/hb-ot-var-common.hh
```diff
@@ -1076,6 +1079,28 @@ struct TupleVariationData
     unsigned compiled_byte_size = 4;
 
     public:
+    friend void swap (tuple_variations_t& a, tuple_variations_t& b)
+    {
+      hb_swap (a.tuple_vars, b.tuple_vars);
+      hb_swap (a.point_data_map, b.point_data_map);
+      hb_swap (a.point_set_count_map, b.point_set_count_map);
+      hb_swap (a.shared_points_bytes.arrayZ, b.shared_points_bytes.arrayZ);
+      hb_swap (a.shared_points_bytes.length, b.shared_points_bytes.length);
```
behdad, line 1088: Can you just swap `shared_points_bytes` and add the swap for that to `hb_array_t`?

src/hb-ot-var-common.hh
qxliu76: when you init tuple iterator here: https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-ot-var-common.hh#L1525, this pointer is out of range already: https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-ot-var-gvar-table.hh#L526,  I thought we should at least check TupleVariationHeader arrays' min_size.
TupleVariationHeader has no static_size, since we might or might not have peak/intermediate tuples.

qxliu76: or  I can add a sanitize() for each TupleVariationHeader so we check the exact size of each header

qxliu76: Oh you're right, we don't need this. I should get var_data from hb_blob, which has total length info so we won't access any pointer beyond that. I'll revert this commit and make another fix

garretrieger (approved):

behdad:
The first commit was already done similarly by Ben. Can you rebase please, and add your swaps if they make sense still?

qxliu76:
> The first commit was already done similarly by Ben. Can you rebase please, and add your swaps if they make sense still?

Thanks! I've rebased my code, and swap is not needed now.

PR #4391, opened 2023-08-31
[meson] add source fallback URL to freetype2 wrap

benoit-pierre:
Since downloading from savannah can be pretty flaky.


khaledhosny:
Thanks!

PR #4392, opened 2023-08-31
Specify tuple_variations_t special member functions

bungeman:
Building with clang complains about the use of `tuple_variations_t`'s implicit copy constructor and copy assignment operator, since automatic generation of these is deprecated when declaring a non-default destructor. This is a good warning because it isn't obvious that copies were being made instead of the object being moved and this struct should be moved and not copied. Declare all the special member functions, defaulting the moves and deleting the copies.

After making `tuple_variations_t` move only, an issue is revealed in `hb_vector_t::push` which has been requiring that objects be copyable. Remove the old non-emplacing `push` now that this works with all existing objects and make a single `push` which is more like `std::vector::emplace_back` since that is somewhat what the newer `push` is attempting to do.

### DIFF
--- a/src/hb-ot-var-common.hh
+++ b/src/hb-ot-var-common.hh
@@ -1076,6 +1076,11 @@ struct TupleVariationData
     unsigned compiled_byte_size = 4;
 
     public:
+    tuple_variations_t () = default;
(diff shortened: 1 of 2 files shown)


behdad:
LGTM. Thanks Ben.

PR #4393, opened 2023-09-01
[instancer] instantiate item varstore

qxliu76:
- instantiate item var_store
- there's a diff in the final ttx file, I feel this might be a fonttools' issue and created a PR for this: https://github.com/fonttools/fonttools/pull/3268
- tests are identical when compared against fonttools' result (with above fix applied)


src/hb-ot-var-common.hh
```diff
@@ -1909,7 +1910,222 @@ struct item_variations_t
...
+            if (delta != 0)
+            {
+              all_zeros = false;
+              break;
+            }
+          }
+          if (all_zeros)
+            continue;
+        }
+        unsigned chars = delta_row_encoding_t::get_row_chars (row);
```
behdad, line 1978: The Python code supports bigint. Using unsigned will silently miscompile if there's more than 8 regions (each take four bits). As a minimum use uint64_t. But we should detect when there's more rows than we support and bail or do something.

Alternatively, use a vector<uint8> and for each region use three bits of a byte. Is not as efficient to do ops on them, but will be correct.

behdad (changes requested):
Just one comment for now. I'll review more.

src/hb-ot-var-common.hh
behdad: Or use `hb_bit_page_t`...

qxliu76: As suggested I have changed `unsigned chars` into a `a vector of uint8_t`

garretrieger (approved):

src/hb-ot-layout-common.hh
```diff
@@ -2300,6 +2300,147 @@ static inline bool ClassDef_serialize (hb_serialize_context_t *c,
...
+
+  inline unsigned get_width ()
+  {
+    unsigned ret = + hb_iter (chars)
+                   | hb_reduce (hb_add, 0u)
+                   ;
+    return ret;
+  }
+
+  unsigned get_columns ()
```
behdad, line 2378: Isn't this unsigned wrong as well? Only supports up to 32 columns.

src/hb-ot-layout-common.hh
```diff
@@ -2300,6 +2300,147 @@ static inline bool ClassDef_serialize (hb_serialize_context_t *c,
...
+    for (auto v : chars)
+    {
+      if (v)
+        cols |= i;
+      i <<= 1;
+    }
+    return cols;
+  }
+
+  static inline unsigned get_chars_overhead (unsigned cols)
```
behdad, line 2391: Same here?

src/hb-ot-layout-common.hh
qxliu76: Thank you! I just pushed a fix

behdad:
Can you add some tests? The CodeCov entries make it hard to review the code.

qxliu76:
I pushed 1 more bug fix and added a unit test. Hopefully this helps reducing the Codecov entries

PR #4399, opened 2023-09-06
[repacker] prevent possible double counting of incoming edges during …

garretrieger:
…graph isolation.

If a single node has multiple incoming links then the links to it's descendants will be double counted due to the lack of a visited set in find_subgraph. Found by Colin in the rust port of the repacker: https://github.com/googlefonts/fontations/pull/591.

Leaving as draft PR until I add a test case.


garretrieger:
Actually, this isn't needed we're using the subgraph map as a visited set already.

PR #3736, opened 2022-07-19
build fix for HB_NO_VAR in hb-subset-plan.cc

qxliu76:
So this just fixed issues in subset_plan.cc, the build is not fine yet. 

I see there're still build issues in hb-shape:
```
FAILED: util/hb-ot-shape-closure 
c++  -o util/hb-ot-shape-closure util/hb-ot-shape-closure.p/hb-ot-shape-closure.cc.o -Wl,--as-needed -Wl,--no-undefined -DHB_NO_VAR=1 '-Wl,-rpath,$ORIGIN/../src' -Wl,-rpath-link,/usr/local/google/home/qxliu/harfbuzz/build/src -Wl,--start-group src/libharfbuzz.so.0.40401.0 /usr/lib/x86_64-linux-gnu/libfreetype.so /usr/lib/x86_64-linux-gnu/libcairo.so /usr/lib/x86_64-linux-gnu/libglib-2.0.so -Wl,--end-group
/usr/bin/ld: util/hb-ot-shape-closure.p/hb-ot-shape-closure.cc.o: in function `font_options_t::post_parse(_GError**)':
/usr/local/google/home/qxliu/harfbuzz/build/../util/font-options.hh:105: undefined reference to `hb_font_set_variations'
collect2: error: ld returned 1 exit status
[208/279] Linking target util/hb-view
FAILED: util/hb-view 
c++  -o util/hb-view util/hb-view.p/hb-view.cc.o -Wl,--as-needed -Wl,--no-undefined -DHB_NO_VAR=1 '-Wl,-rpath,$ORIGIN/../src' -Wl,-rpath-link,/usr/local/google/home/qxliu/harfbuzz/build/src -Wl,--start-group src/libharfbuzz.so.0.40401.0 /usr/lib/x86_64-linux-gnu/libfreetype.so /usr/lib/x86_64-linux-gnu/libcairo.so /usr/lib/x86_64-linux-gnu/libglib-2.0.so -Wl,--end-group
/usr/bin/ld: util/hb-view.p/hb-view.cc.o: in function `font_options_t::post_parse(_GError**)':
/usr/local/google/home/qxliu/harfbuzz/build/../util/font-options.hh:105: undefined reference to `hb_font_set_variations'
/usr/bin/ld: util/hb-view.p/hb-view.cc.o: in function `helper_cairo_create_ft_font_face(font_options_t const*)':
[...]

PR #4408, opened 2023-09-19
HB_NO_VAR: Fix build, drop fvar, avar, cvar, MVAR tables from subsets

tpetillon:
In the same vein as #3736, this fixes compilation when in `HB_NO_VAR` mode.

I’m not familiar with the code, so I hope I didn’t `#ifndef` out too much or too little code.

### DIFF
--- a/src/hb-ot-var-cvar-table.hh
+++ b/src/hb-ot-var-cvar-table.hh
@@ -143,19 +143,6 @@ struct cvar
     if (c->plan->all_axes_pinned)
       return_trace (false);
 
-    /* subset() for cvar is called by partial instancing only, we always pass
-     * through cvar table in other cases */
-    if (!c->plan->normalized_coords)
-    {
-      unsigned axis_count = c->plan->source->table.fvar->get_axis_count ();
-      unsigned total_size = min_size + tupleVariationData.get_size (axis_count);
-      char *out = c->serializer->allocate_size<char> (total_size);
-      if (unlikely (!out)) return_trace (false);
-
-      hb_memcpy (out, this, total_size);
-      return_trace (true);
-    }
-
     OT::TupleVariationData::tuple_variations_t tuple_variations;
     unsigned axis_count = c->plan->axes_old_index_tag_map.get_population ();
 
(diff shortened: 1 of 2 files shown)


qxliu76:
I'd suggest add this `ifndef` wrapper here: https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-subset.cc#L518
fvar, avar,cvar and MVAR should be all wrapped.

```
#ifndef HB_NO_VAR
case HB_OT_TAG_fvar:
    if (plan->user_axes_location.is_empty ()) return _passthrough (plan, tag);
    return _subset<const OT::fvar> (plan, buf);
  case HB_OT_TAG_avar:
    if (plan->user_axes_location.is_empty ()) return _passthrough (plan, tag);
    return _subset<const OT::avar> (plan, buf);
  case HB_OT_TAG_cvar:
    if (plan->user_axes_location.is_empty ()) return _passthrough (plan, tag);
    return _subset<const OT::cvar> (plan, buf);
  case HB_OT_TAG_MVAR:
    if (plan->user_axes_location.is_empty ()) return _passthrough (plan, tag);
    return _subset<const OT::MVAR> (plan, buf);
#endif
```

tpetillon:
@qxliu76 I’ve pushed the suggested `#ifndef` wrapper in hb-subset.

qxliu76:
> @qxliu76 I’ve pushed the suggested `#ifndef` wrapper in hb-subset.

Thanks! Can you also delete the first commit? Since cvar is dropped, the first commit is not needed now.

behdad:
> > @qxliu76 I’ve pushed the suggested `#ifndef` wrapper in hb-subset.
> 
> Thanks! Can you also delete the first commit? Since cvar is dropped, the first commit is not needed now.

I think that's needed, otherwise the `face->ot.tables.fvar` will not compile.

qxliu76:
> > > @qxliu76 I’ve pushed the suggested `#ifndef` wrapper in hb-subset.
> > 
> > 
> > Thanks! Can you also delete the first commit? Since cvar is dropped, the first commit is not needed now.
> 
> I think that's needed, otherwise the `face->ot.tables.fvar` will not compile.

Oh I think we can get rid of that code block because it's covered by below passthrough: 
```
case HB_OT_TAG_fvar:
    if (plan->user_axes_location.is_empty ()) return _passthrough (plan, tag);
    return _subset<const OT::fvar> (plan, buf);
```
What do you think?

behdad:
> What do you think?

I believe so, yes.

qxliu76:
@tpetillon can you help remove the whole code block wrapped by your 1st commit? After that, your 1st commit can be deleted. Thanks!

behdad:
Thank you!

tpetillon:
All right, I deleted the block altogether.

PR #4410, opened 2023-09-20
[instancer] instantiate HVAR/VVAR

qxliu76:
There're differences on varstore inner indices only when compared to fonttools' result. 
This is because we sort rows differently. I've verified if we change harfbuzz sorting method then we would get identical result as fonttools. 
Here's python code to sort items(based on values in tuples): https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/varLib/varStore.py#L627

While harfbuzz uses hb_memcmp to sort rows: https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-ot-var-common.hh#L2151,  https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-ot-var-common.hh#L2123

I feel we might accept this difference since harfbuzz's result is valid and optimized. I'd be happy to update sorting method if that's recommended from review comments.


behdad:
I think `memcmp` is not acceptable because it produces different results depending on endianness.

behdad:
> I think `memcmp` is not acceptable because it produces different results depending on endianness.

Also, there's value in matching fonttools for testing purposes.

behdad:
Please fix the Arm failure.

qxliu76:
> > I think `memcmp` is not acceptable because it produces different results depending on endianness.
> 
> Also, there's value in matching fonttools for testing purposes.

Thanks! I pushed a fix for the sorting method.

src/hb-ot-var-common.hh
```diff
@@ -2174,7 +2174,14 @@ struct item_variations_t
     const hb_vector_t<int>** a = (const hb_vector_t<int>**) pa;
     const hb_vector_t<int>** b = (const hb_vector_t<int>**) pb;
 
-    return ((*b)->as_array ()).cmp ((*a)->as_array ());
+    for (unsigned i = 0; i < (*b)->length; i++)
+    {
+      int va = (*a)->arrayZ[i];
+      int vb = (*b)->arrayZ[i];
+      if (va != vb)
+        return va - vb;
```
behdad, line 2182: In rare cases this can underflow. Just use a `va < vb ? -1 : +1`.

PR #4416, opened 2023-09-22
[instancer] try to simplify item_variation_t API calls

qxliu76:
1. It's possible that region lists become empty after instantiation(e.g deltas become all zeros). Return true in this case. And varStore serialize() call will return false if regions are empty.
2.  add an API which does create (), instantiate tuple vars(), and as_item_varstore() together, and update MVAR and HVAR code to call this API instead.

### DIFF
--- a/src/hb-ot-var-common.hh
+++ b/src/hb-ot-var-common.hh
@@ -1812,6 +1812,20 @@ struct item_variations_t
   const hb_map_t& get_varidx_map () const
   { return varidx_map; }
 
+  bool instantiate (const VariationStore& varStore,
+                    const hb_subset_plan_t *plan,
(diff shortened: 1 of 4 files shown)


behdad:
> 1. And varStore serialize() call will return false if regions are empty.

`serialize` returning false typically means an error; `subset` returning false means "not needed" or "empty".

qxliu76:
> > 1. And varStore serialize() call will return false if regions are empty.
> 
> `serialize` returning false typically means an error; `subset` returning false means "not needed" or "empty".

All right, I deleted that commit. When region list is empty, `instantiate` returns false and then `subset` also returns false, meaning varStore is empty. So nothing needs to be changed.

PR #4422, opened 2023-09-28
fuzzer fix https://oss-fuzz.com/testcase-detail/5842152921628672

qxliu76:
Access TupleVariationData through blob, because we don't sanitize var_data


garretrieger (approved):

behdad:
@qxliu76 Test needs to be updated. Bots failing.

qxliu76:
> @qxliu76 Test needs to be updated. Bots failing.

Oops my fault that I forgot to run normal tests, now it's updated.

PR #4438, opened 2023-10-12
[instancer] templatize the priority queue, use custom type for varstore

qxliu76:
when instantiating varstore, we need to pop a tuple like (combined_gain, i, j), if combined gain is the same then we compare the value of i, and then j. So we'd like to use custom type as the key when popping from the queue. This would match fonttool's algorithm cause it uses heappop(): https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/varLib/varStore.py#L586

I'll add a test for this later(in another PR that I'm still working on)

### DIFF
--- a/src/hb-ot-var-common.hh
+++ b/src/hb-ot-var-common.hh
@@ -1938,6 +1938,35 @@ struct item_variations_t
 
   /* main algorithm ported from fonttools VarStore_optimize() method, optimize
    * varstore by default */
+
+  struct combined_gain_idx_tuple_t
+  {
+    int gain;
+    unsigned idx_1;
+    unsigned idx_2;
+
+    combined_gain_idx_tuple_t () = default;
+    combined_gain_idx_tuple_t (int gain_, unsigned i, unsigned j)
+        :gain (gain_), idx_1 (i), idx_2 (j) {}
+
+    bool operator < (const combined_gain_idx_tuple_t& o)
+    {
+      if (gain != o.gain)
+        return gain < o.gain;
+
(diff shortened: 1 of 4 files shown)


src/hb-ot-var-common.hh
```diff
@@ -1938,6 +1938,35 @@ struct item_variations_t
 
   /* main algorithm ported from fonttools VarStore_optimize() method, optimize
    * varstore by default */
+
+  struct combined_gain_idx_tuple_t
+  {
+    int gain;
+    unsigned idx_i;
+    unsigned idx_j;
```
garretrieger, line 1946: Is there a better more descriptive name for these indices than i and j?

garretrieger (approved):

src/hb-ot-var-common.hh
qxliu76: they represent the indices of 2 encodings in an array that we want to combine. Would idx_1, idx_2 be better?

garretrieger: Got it, yeah idx_1/2 seems fine.

garretrieger (approved):

src/hb-priority-queue.hh
```diff
@@ -42,10 +42,11 @@
  * priority of its children. The heap is stored in an array, with the
  * children of node i stored at indices 2i + 1 and 2i + 2.
  */
+template <typename Type>
```
behdad, line 45: Can you call `Type` just `K` or `Key`? We might want to templatize the value one day as well.

PR #4448, opened 2023-10-17
[docs] Fixed links to freetype FT_LOAD_XXX flags

biodranik:
What is the right way to automatically generate clickable URLs in gtk-doc? Will this work: `<https://some.url/>`?

### DIFF
--- a/src/hb-ft.cc
+++ b/src/hb-ft.cc
@@ -225,7 +225,7 @@ _hb_ft_hb_font_check_changed (hb_font_t *font,
  * Sets the FT_Load_Glyph load flags for the specified #hb_font_t.
  *
  * For more information, see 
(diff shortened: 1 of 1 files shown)


behdad:
Maybe @matthiasclasen knows. Or @khaledhosny

biodranik:
Thanks! How to build Harfbuzz docs locally? I tried ninja but it didn't create html even with gtk-doc installed.

khaledhosny:
You can use markdown syntax in the doc comments.

behdad:
> Thanks! How to build Harfbuzz docs locally? I tried ninja but it didn't create html even with gtk-doc installed.

Try `ninja -Cbuild harfbuzz-doc

biodranik:
Thanks for the hint! I've tested my changes, and they work (and the link is valid):

behdad:
Thanks.

PR #4469, opened 2023-10-31
Typed offsets

behdad:
Playing with adding type information to `base` arguments when it's non-trivial.


behdad:
Not sure it's worth it, but this is a sketch that seems to work.

behdad:
I think this covers most of our indirect bases. Going to merge.

PR #4470, opened 2023-11-01
[instancer] Update help message for --instance.

### LINKED ISSUE #4462: Improve `hb-subset --help-all` output
Figure out how to better format it.

Also, now that we do partial instancing the `--instance` help message needs to be adjusted.

I also suggest renaming (with a back-compat alias) to `--variations` to match other tools of us.

garretrieger:
Remove note about partial instancing not being supported. Add --variations flag as an alternate name for --instance. This matches other hb utils. Fixes #4462


behdad:
You can make `--instance` hidden from the help by using `G_OPTION_FLAG_HIDDEN` in the third argument.

util/hb-subset.cc
```diff
@@ -1024,15 +1024,14 @@ subset_main_t::add_options ()
...
-     "number or the literal string 'drop'\n"
-     "                                                        "
-     "For example: --instance=\"wdth=100 wght=200\" or --instance=\"wdth=drop\"\n"
-     "                                                        "
-     "Note: currently only fully instancing is supported\n",
+#ifndef HB_NO_VAR    
+     {"variations",	0, 0, G_OPTION_ARG_CALLBACK, (gpointer) &parse_instance,
+     "(Partially|Fully) Instantiate a variable font. A location consists of the tag "
+     "of a variation axis, followed by '=', followed by a number or the literal "
+     "string 'drop'. For example: --instance=\"wdth=100 wght=200\" or --instance=\"wdth=drop\"",
```
behdad, line 1031: Update `--instance` to `--variations` here.

garretrieger:
> You can make `--instance` hidden from the help by using `G_OPTION_FLAG_HIDDEN` in the third argument.

Done.

PR #4473, opened 2023-11-01
[instancer] add the option to leave blanks in the min:def:max syntax.

garretrieger:
When parsing axis positions in the --variations flag recognize empty values as meaning the existing value for that part. For example:

:300:500

Specifies min = existing, def = 300, max = 500.

See: https://github.com/fonttools/fonttools/issues/3322

Need to add tests before merging this.

### DIFF
--- a/src/hb-subset-input.cc
+++ b/src/hb-subset-input.cc
@@ -529,10 +529,12 @@ hb_subset_input_set_axis_range (hb_subset_input_t  *input,
 /**
  * hb_subset_input_get_axis_range: (skip)
  * @input: a #hb_subset_input_t object.
-  * @axis_tag: Tag of the axis
+ * @axis_tag: Tag of the axis
  * @axis_min_value: Set to the previously configured minimum value of the axis variation range.
  * @axis_max_value: Set to the previously configured maximum value of the axis variation range.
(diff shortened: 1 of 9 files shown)


qxliu76 (approved):

behdad:
Can you finish this? I don't quite remember all we discussed changing about it.

garretrieger:
Yes, I'll get this finished this week. All that remains is adding some tests.

src/hb-subset-input.cc
```diff
@@ -509,21 +507,55 @@ hb_subset_input_set_axis_range (hb_subset_input_t  *input,
...
+  float new_min_val = hb_clamp(min, axis_info.min_value, axis_info.max_value);
+  float new_max_val = hb_clamp(max, axis_info.min_value, axis_info.max_value);
+  float new_default_val = hb_clamp(def, new_min_val, new_max_val);
   return input->axes_location.set (axis_tag, Triple (new_min_val, new_default_val, new_max_val));
 }
+
+/**
+ * hb_subset_input_get_axis_range: (skip)
+ * @input: a #hb_subset_input_t object.
+  * @axis_tag: Tag of the axis
```
behdad, line 532: Extra space here. Maybe that's throwing it off?

PR #4474, opened 2023-11-02
[instancer] fix value format for SinglePos/PairPos when strip_hints is on

qxliu76:
- support strip_hints for SinglePos
- SinglePos/PairPos: do not strip hints on var devices for VF and partial instancing


behdad:
Can you rebase the conflict please?

qxliu76:
> Can you rebase the conflict please?

Done.

PR #4476, opened 2023-11-02
[instancer] Fix partial instancing for LigCaretList

qxliu76:
It needs to happen after varstore instantiation because the varidx needs to be remapped first.


src/OT/Layout/GDEF/GDEF.hh
```diff
@@ -664,7 +664,12 @@ struct GDEFVersion1_2
     out->version.minor = version.minor;
     bool subset_glyphclassdef = out->glyphClassDef.serialize_subset (c, glyphClassDef, this, nullptr, false, true);
     bool subset_attachlist = out->attachList.serialize_subset (c, attachList, this);
-    bool subset_ligcaretlist = out->ligCaretList.serialize_subset (c, ligCaretList, this);
+    bool subset_ligcaretlist = false;
+    /* For partial instancing, if varStore exists, CaretValue need to be after
+     * varstore instantiation cause varidx needs to be remapped first. */
+    if (version.to_int () < 0x00010003u || !varStore
+        || c->plan->all_axes_pinned || !c->plan->normalized_coords)
+      subset_ligcaretlist = out->ligCaretList.serialize_subset (c, ligCaretList, this);
```
behdad, line 672: Why can't we always do it after varstore block?

qxliu76: Yeah you're right, I figured out we need to put it at last, after snapshot revert

Issue #4478, opened 2023-11-03
[sanitize] segfault in sanitize code may need a memory barrier.


garretrieger:
In a usage of harfbuzz in another library we're encountering a segfault when running a fuzzer. Happens here:

https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-ot-layout-common.hh#L3911

I've checked and version.sanitize(c) is returning false as expected. However, the memory access to version.major is still happening which triggers the segfault. I suspect this is a memory barrier issue like: https://github.com/harfbuzz/harfbuzz/commit/f3151b6582a38fb1a377eb6070b8cecedb2ea711

behdad:
Unfortunately this is a major issue in our `sanitize()` code. What's happening is that in code like this:
```
    return_trace (version.sanitize (c) &&
		  likely (version.major == 1) &&
		  varRecords.sanitize (c, this));
```
after inlining `version.sanitize (c)` and realizing that it doesn't change the value of `version.major`, the `version.major` load is moved earlier by the compiler to pipeline with the `version.sanitize(c)` code. The read then becomes illegal if sanitize is to fail...

This probably affects hundreds of our sanitize functions. Even the ones like:
```
  template <typename TLookup> 
  bool sanitize (hb_sanitize_context_t *c) const 
  { 
    TRACE_SANITIZE (this); 
    if (unlikely (!u.version.sanitize (c))) return_trace (false); 
    switch (u.version.major) { 
    case 1: return_trace (u.version1.sanitize<TLookup> (c)); 
#ifndef HB_NO_BEYOND_64K 
    case 2: return_trace (u.version2.sanitize<TLookup> (c)); 
#endif 
    default: return_trace (true); 
    } 
  }   
```
Or even simpler:
```
  bool sanitize (hb_sanitize_context_t *c) const 
  { 
    TRACE_SANITIZE (this); 
    return_trace (c->check_struct (this) && 
↦       ↦         extensionLookupType != T::SubTable::Extension); 
  } 
```

Basically any access to `*this` can possibly be moved before the `c->check_struct (this)` and cause a problem.

I don't think adding a barrier inside the sanitize machinery itself can fix this, just probably make it less likely to happen.

behdad:
To properly fix this we need to audit all `sanitize()` methods and add read barriers between `check_struct` and subsequent reads. And similarly for the original case reported.

behdad:
[...]

PR #4479, opened 2023-11-03
[sanitize] Try a barrier

behdad:
Not sure it works.

To fix https://github.com/harfbuzz/harfbuzz/issues/4478

### DIFF
--- a/src/hb-atomic.hh
+++ b/src/hb-atomic.hh
@@ -118,12 +118,12 @@ _hb_atomic_ptr_impl_cmplexch (const void **P, const void *O_, const void *N)
  */
 #ifndef _hb_compiler_memory_r_barrier
 #if defined(__ATOMIC_ACQUIRE) // gcc-like
-#define _hb_compiler_memory_r_barrier() asm volatile("": : :"memory")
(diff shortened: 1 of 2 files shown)


behdad:
This makes the code size grow by about 1%; which makes me think it's doing its job...

behdad:
Humm, compiler crash in the `asan+ubsan` bug :(
https://app.circleci.com/pipelines/github/harfbuzz/harfbuzz/7735/workflows/43c7f1fc-b0b9-459e-9c4f-763f8027762b/jobs/213155

garretrieger:
Unfortunately this doesn't resolve the crash I'm getting. Where-as adding:

(_hb_memory_r_barrier (), true) &&

In between the sanitize and the data access does work for me.

behdad:
> (_hb_memory_r_barrier (), true) &&
> 
> In between the sanitize and the data access does work for me.

Yeah that's what I was afraid of, that we have to audit all sanitize() methods and adjust.

behdad:
Proper fix in https://github.com/harfbuzz/harfbuzz/pull/4480

PR #4480, opened 2023-11-04
Sanitize barrier properly

behdad:
I'm sure I've forgot some. But this is a good start.

Fixes https://github.com/harfbuzz/harfbuzz/issues/4478

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,13 +3,13 @@ version: 2.1
 executors:
   win32-executor:
     docker:
-      - image: cimg/base:edge-20.04
+      - image: cimg/base:2023.10
   win64-executor:
     docker:
-      - image: cimg/base:edge-20.04
+      - image: cimg/base:2023.10
   autotools-executor:
     docker:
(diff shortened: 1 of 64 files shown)


behdad:
CircleCI bot roll forward was because the `asan-ubsan` bot compiler was crashing.

behdad:
@khaledhosny Can I get some help from you with the bots? I don't know what's going on.

khaledhosny:
> @khaledhosny Can I get some help from you with the bots? I don't know what's going on.

Looks like the job is getting killed because it is stuck (or at least that is what the runnuer thinks). Dropping the `-j9` or using a less number of parallel jobs might help. You can also rerurn the job with ssh access to debug directly on the remote machine.

khaledhosny:
But something is making these jobs much slower than before, may be a compiler bug?

behdad:
> But something is making these jobs much slower than before, may be a compiler bug?

That's my feeling.  Both get stuck linking `test/threads/hb-subset-threads`.

khaledhosny (approved):

PR #4487, opened 2023-11-06
[subset] Add HB_SUBSET_FLAGS_IFTB_REQUIREMENTS flag.

garretrieger:
When enabled the output subset conforms to the requirements for a base font to be patched by IFTB patches. IFTB is a proposed incremental font transfer patch. This new flag is set as experimental. Currently, setting this flag causes the output subset to also use long offsets for outline data in loca/glyf, gvar, CFF, and CFF2.

This change is a version of https://github.com/skef/harfbuzz/commit/3ae2fe2084d96f61d8ad1bbe42a6ef4e62ec830a rebased onto head w/ tests added.


src/hb-ot-var-gvar-table.hh
```diff
@@ -428,7 +428,10 @@ struct gvar
       subset_data_size += get_glyph_var_data_bytes (c->source_blob, glyph_count, old_gid).length;
     }
 
-    bool long_offset = subset_data_size & ~0xFFFFu;
+    bool long_offset = (subset_data_size & ~0xFFFFu);
+ #ifdef HB_EXPERIMENTAL_API
+    long_offset = long_offset || (c->plan->flags & HB_SUBSET_FLAGS_IFTB_REQUIREMENTS);
```
behdad, line 433: I think `||=` is a valid op?

src/hb-ot-cff-common.hh
```diff
@@ -111,11 +112,14 @@ struct CFFIndex
...
   bool serialize_header (hb_serialize_context_t *c,
 			 Iterator it,
-			 unsigned data_size)
+			 unsigned data_size,
+                         unsigned min_off_size = 0)
   {
     TRACE_SERIALIZE (this);
 
     unsigned off_size = (hb_bit_storage (data_size + 1) + 7) / 8;
+    if (min_off_size > off_size)
```
behdad, line 121: Maybe `hb_max`?

src/hb-subset-cff1.cc
```diff
@@ -778,13 +786,42 @@ struct cff1_subset_plan
...
+
+  plan.info.char_strings_link = c->pop_pack (false);
+  return true;
+}
+
 bool
 OT::cff1::accelerator_subset_t::serialize (hb_serialize_context_t *c,
 					   struct OT::cff1_subset_plan &plan) const
 {
+  if (plan.charstrings_last) {
```
behdad, line 820: I don't see any issue always putting the charstrings last. (with a comment obviously).

behdad:
Left a couple of comments. LGTM otherwise.

src/hb-ot-cff-common.hh
garretrieger: Done

src/hb-subset-cff1.cc
garretrieger: Sounds good, I'll do that.

src/hb-ot-var-gvar-table.hh
garretrieger: Doesn't look like it. I think only |= is defined.

PR #4493, opened 2023-11-11
[set] Speed up next_range()

behdad:
This in theory speeds up `next_range()`, which is highly used in fonts like AdobeBlank, used in the subset `cmap` test. But I'm not sure if the gains are worth the extra complexity. It took me quite a while to Get It Right.


behdad:
This actually shows me slowdown in testing.

behdad:
I fixed the slowdown. But I'm not seeing any measurable speedup either.

PR #4497, opened 2023-11-15
[ChainRule] Speed up sanitize

behdad:
Limit backtrack-len and lookahead-len to HB_MAX_CONTEXT_LENGTH. Add a max_size for the struct, and bail early if max_size is within sanitize range.

Speeds up Gulzar sanitization by 20%.


behdad:
The logic here is false. It can only work if we push the array max-len into the array `operator []` itself. Which is not worth it IMO.

PR #4507, opened 2023-11-30
Improve CMake builds

fanc999-1:
Hi,

This attempts to improve CMake builds so that things are more in-line with the Meson builds, where:

* We now build the `harfbuzz-cairo` integration library that was later added, if requested, by passing in `-DHB_HAVE_CAIRO=ON`.  Cairo's presence will be checked for in this case.
* `hb-view` is now only built if `-DHB_HAVE_CAIRO=ON` is passed in, in addition to `-DHB_BUILD_UTILS=ON`, so `-DHB_BUILD_UTILS=ON` alone will not automatically look for Cairo, and also fix its build as it will depend on `harfbuzz-cairo`.
* Build `hb-info` if GObject support is enabled, when we have `-DHB_BUILD_UTILS=ON`.
* Generate (and install, unless requested not to) `hb-features.h`.
* Try to update the cflags in the introspection builds to match the ones used in Meson better.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,7 @@ endif ()
 
 
 ## HarfBuzz build configurations
+option(HB_HAVE_CAIRO "Enable cairo interop helpers" OFF)
 option(HB_HAVE_FREETYPE "Enable freetype interop helpers" OFF)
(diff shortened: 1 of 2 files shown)


behdad:
Thanks!

khaledhosny (approved):

PR #4510, opened 2023-12-02
Per syllable

behdad:
Trying to fix https://github.com/harfbuzz/harfbuzz/discussions/4509
but tests fail.

### DIFF
--- a/src/hb-ot-layout-gsubgpos.hh
+++ b/src/hb-ot-layout-gsubgpos.hh
@@ -518,11 +518,14 @@ struct hb_ot_apply_context_t :
 #ifndef HB_OPTIMIZE_SIZE
     HB_ALWAYS_INLINE
 #endif
-    void reset (unsigned int start_index_)
+    void reset (unsigned int start_index_, bool per_syllable = true)
     {
       idx = start_index_;
       end = c->buffer->len;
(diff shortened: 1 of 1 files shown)


behdad:
The use test expectations might need to be updated.

khaledhosny:
The failing tests were added in e3db84a9d05f40c5281d889a726eb2e4baa12328 from https://github.com/harfbuzz/harfbuzz/pull/1695

behdad:
> The failing tests were added in [e3db84a](https://github.com/harfbuzz/harfbuzz/commit/e3db84a9d05f40c5281d889a726eb2e4baa12328) from #1695

I think it happened to work that way by chance. IIRC we always ignore ZWNJ/ZWJ in the backtrack/lookahead. Let me wrap my head around this.

khaledhosny:
> The failing tests were added in [e3db84a](https://github.com/harfbuzz/harfbuzz/commit/e3db84a9d05f40c5281d889a726eb2e4baa12328) from #1695

CoreText seems to give the old output for both failing tests.

behdad:
Thanks. Yeah, the ZWNJ should do something there...

behdad:
Some help reviewing the failing test would be appreciated. I'll see if I can do it myself.

PR #4515, opened 2023-12-05
CFF negative offset

behdad:
This is a little bit brittle as I don't know if I got all the offsets or not (quite possibly not). But otherwise seems to work.


behdad:
Ugh. A lot of tests fail...

behdad:
@garretrieger Can you help look int the subset-cff fails?

garretrieger:
> @garretrieger Can you help look int the subset-cff fails?

Sure, no problem.

behdad:
Ah I think I found it.

behdad:
Tests pass now. @khaledhosny this is ready for review. Thanks.

PR #4530, opened 2023-12-15
[repacker] Increase repacker max rounds to 30.

garretrieger:
Found an example font that needs the higher limit.


behdad:
LGTM. Make it round, 32. ;)

garretrieger:
> LGTM. Make it round, 32. ;)

Done

PR #4541, opened 2024-01-06
[ot-math-table] fix a bug in MathKern::get_value()

lie-yan:
Fixes harfbuzz/harfbuzz#4542

### DIFF
--- a/src/hb-ot-math-table.hh
+++ b/src/hb-ot-math-table.hh
@@ -354,7 +354,7 @@ struct MathKern
     while (count > 0)
     {
       unsigned int half = count / 2;
-      hb_position_t height = correctionHeight[i + half].get_y_value(font, this);
+      hb_position_t height = correctionHeight[i + half].get_y_value (font, this);
       if (sign * height <= sign * correction_height)
       {
         i += half + 1;
@@ -362,7 +362,7 @@ struct MathKern
       } else
         count = half;
     }
-    return kernValue[i].get_x_value(font, this);
+    return kernValue[i].get_x_value (font, this);
   }
 
   unsigned int get_entries (unsigned int start_offset,
--- a/src/hb-ot-math-table.hh
+++ b/src/hb-ot-math-table.hh
@@ -344,22 +344,14 @@ struct MathKern
     const MathValueRecord* kernValue = mathValueRecordsZ.arrayZ + heightCount;
     int sign = font->y_scale < 0 ? -1 : +1;
 
-    /*
-     * According to OpenType spec (v1.9), except for the boundary cases, the index
+    /* According to OpenType spec (v1.9), except for the boundary cases, the index
      * chosen for kern value should be i such that
-     *  correctionHeight[i-1] <= correction_height < correctionHeight[i]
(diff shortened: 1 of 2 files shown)


lie-yan:
Fixes harfbuzz/harfbuzz#4542

lie-yan:
Fixes harfbuzz/harfbuzz#4542

src/hb-ot-math-table.hh
```diff
@@ -344,26 +344,32 @@ struct MathKern
     const MathValueRecord* kernValue = mathValueRecordsZ.arrayZ + heightCount;
     int sign = font->y_scale < 0 ? -1 : +1;
 
-    /* The description of the MathKern table is a ambiguous, but interpreting
-     * "between the two heights found at those indexes" for 0 < i < len as
-     *
-     *   correctionHeight[i-1] < correction_height <= correctionHeight[i]
-     *
-     * makes the result consistent with the limit cases and we can just use the
-     * binary search algorithm of std::upper_bound:
```
fred-wang, line 353: With your changes, isn't the loop below actually exactly doing the same as `std::upper_bound` (see https://en.cppreference.com/w/cpp/algorithm/upper_bound#Possible_implementation)?

If so, I think it's not necessary to add a comment with the explanation of the algorithm below. Instead, I'd keep the reference to `std::upper_bound` and explain that this matches the spec, including for the boundary cases.

lie-yan: You are right. That's just what `std::upper_bound()` does.
I have changed the comment as you suggested.

fred-wang (commented):
Thanks. This LGTM, but I'm not a maintainer harfbuzz so I'll defer official review to someone else.

behdad:
I wonder if we could use our internal bsearch?

lie-yan:
> I wonder if we could use our internal bsearch?

I believe we can. We can use the following code snippet to compute index `i`. 
```cpp
unsigned pos;
unsigned i = hb_bsearch_impl(&pos, key, base, nmemb, stride, compar) ? pos + 1 : pos;
```

For `key` in the array, `pos` points to the element that is equal to `key`.
For `key` not in the array, we can verify that the algorithm is equivalent to the following code snippet, which is exactly [`std::upper_bound()`](https://en.cppreference.com/w/cpp/algorithm/upper_bound#Version_1) except for the computation of `step`. We can prove that they are equivalent. 

```cpp
int min = 0, count = nmemb;
// Define:
//   max := min + count - 1
// Assume:
//   array[-1] = -infty
//   array[nmemb]= +infty
// Loop invariant:
//   0 <= min + count <= nmemb
//   array[min-1] <= key < array[min + count]
// Termination:
//   count is decreased in each iteration and reaches 0 on loop end
while (count > 0) 
{
  int step = (count - 1) / 2;
  int mid = min + step;
  
  if (!(key < array[mid])) {
    min = mid + 1;
    count -= step + 1;
  }
  else {
    count = step;
  }
}
pos = min;
```

behdad:
Thanks. Would be nice to do it.

lie-yan:
I will create a pull request.

PR #4562, opened 2024-01-26
[autotools] Installing missing hb-features.h header

TingPing:
This is already installed by CMake and Meson.


TingPing:
Actually `hb.h` never includes `hb-features.h`.

The docs say include `hb-features.h` directly except then it warns and tells you to include `hb.h`.

Which is intended?

behdad:
> except then it warns and tells you to include `hb.h`.

I don't see any warning in:
https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-features.h.in

TingPing:
Ah that part was fixed in 30c5402e3d0cc156fd5f04560864a88723173cf2

PR #4574, opened 2024-02-05
Fix coverity defects: Null pointer dereferences

qxliu76:
Make coverity happy, but I think this pointer will not be null when hashmap has() call returns true


behdad:
You are right. I'm happy to ignore the Coverity warning in this case.

PR #4576, opened 2024-02-06
Update hb-serialize.hh

lzsiga:
This version (unlike the original) does compile with g++ 4.8.3 which is the newest available version on Aix5.3

### DIFF
--- a/src/hb-serialize.hh
+++ b/src/hb-serialize.hh
@@ -155,10 +155,10 @@ struct hb_serialize_context_t
     hb_vector_t<link_t> virtual_links;
     object_t *next;
 
-    auto all_links () const HB_AUTO_RETURN
-        (( hb_concat (this->real_links, this->virtual_links) ));
-    auto all_links_writer () HB_AUTO_RETURN
-        (( hb_concat (this->real_links.writer (), this->virtual_links.writer ()) ));
+    hb_vector_t<link_t> all_links ()
+        { return hb_concat (this->real_links, this->virtual_links); }
+    hb_vector_t<link_t> all_links_writer ()
+        { return hb_concat (this->real_links.writer (), this->virtual_links.writer ()); }             
   };
 
   struct snapshot_t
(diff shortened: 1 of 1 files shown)


behdad:
But this is not the same. This creates a vector of all the entries and returns that instead.

Can you paste the error message you get?

lzsiga:
Hi, thank you for your quick reply; well, you got me here, I honestly thought it is the same thing, just with older syntax. Here's the error message from older g++:

    In file included from hb.hh:531:0,
                     from hb-static.cc:27,
                     from main.cc:350:
    hb-serialize.hh:159:27: error: invalid use of incomplete type 'const struct hb_serialize_context_t::object_t'
             (( hb_concat (this->real_links, this->virtual_links) ));
                               ^
    hb-meta.hh:76:41: note: in definition of macro 'HB_AUTO_RETURN'
     #define HB_AUTO_RETURN(E) -> decltype ((E)) { return (E); }
                                             ^
    In file included from hb-subset.hh:36:0,
                     from hb-open-type.hh:37,
                     from hb-static.cc:29,
                     from main.cc:350:
    hb-serialize.hh:69:10: error: forward declaration of 'const struct hb_serialize_context_t::object_t'
       struct object_t
              ^

behdad:
We've had similar issues before. Can you replace `this` with `(+this)` and see if it works?

lzsiga:
I tried something even simpler, which seems to help, removed `this` altogether:

    --- src/hb-serialize.orig       2023-11-11 15:08:45 +0100
    +++ src/hb-serialize.hh 2024-02-06 17:05:58 +0100
    @@ -156,9 +156,9 @@
         object_t *next;
     
         auto all_links () const HB_AUTO_RETURN
    -        (( hb_concat (this->real_links, this->virtual_links) ));
    +        (( hb_concat (real_links, virtual_links) ));
         auto all_links_writer () HB_AUTO_RETURN
    -        (( hb_concat (this->real_links.writer (), this->virtual_links.writer ()) ));
    +        (( hb_concat (real_links.writer (), virtual_links.writer ()) ));
       };

behdad:
Great! Can you update the PR please?

PR #4578, opened 2024-02-06
`VARC` table

behdad:
https://github.com/harfbuzz/boring-expansion-spec/blob/main/VARC.md


behdad:
FontTools counterpart: https://github.com/fonttools/fonttools/pull/3395

behdad:
This renders things now!

Need to clean up a bit: remove the old glyf1 varcomposites, add test, benchmark, etc.

behdad:
> This renders things now!
> 
> Need to clean up a bit: remove the old glyf1 varcomposites, add test, benchmark, etc.

glyf1 varcomposites removed.

Just needs a test and some benchmarking.

behdad:
Main TODO is subsetting closure over VARC, and subsetting VARC itself... Left for future.

behdad:
Now with cycle detection, depth limit, and edge limit.

behdad:
Both this and the fontttools PR are now up to date with the latest changes I made to the VARC table (including conditional components); and are have tests too. I think we can commit them. WDYT @rsheeter @anthrotype @khaledhosny

test/api/test-draw-varc.c
```diff
@@ -0,0 +1,268 @@
...
+{
+  unsigned move_to_count;
+  unsigned line_to_count;
+  unsigned quad_to_count;
+  unsigned cubic_to_count;
+  unsigned close_path_count;
+} draw_data_t;
+
+/* Our modified itoa, why not using libc's? it is going to be used
+   in harfbuzzjs where libc isn't available */
```
khaledhosny, line 40: I don’t know what is the history of this comment, but we have libc when building harfbuzzjs for a while now. This also is a test file, so any harfbuzzjs concerns do not apply.

khaledhosny:
This is way above my head, so I have nothing meaningful to add.

test/api/test-draw-varc.c
behdad: Thanks. I just copy/pasted from `test-draw`. Let's leave.

behdad:
> This is way above my head, so I have nothing meaningful to add.

Thanks for checking.

It should be straightforward translation of the spec:
https://github.com/harfbuzz/boring-expansion-spec/blob/main/VARC.md

PR #4579, opened 2024-02-09
Fixed some errors in STAT table implementation

chrsan:
- Fixed wrong return type for `AxisValue::get_value` (`bool` -> `float`)
- Getting the actual axis value must add `offsetToAxisValueOffsets` to the offset for the axis value table


src/hb-ot-stat-table.hh
```diff
@@ -349,7 +349,7 @@ struct AxisValueFormat4
 
 struct AxisValue
 {
-  bool get_value (unsigned int axis_index) const
```
ebraminio, line 352: 😳😢

Thanks for fixing this

PR #4588, opened 2024-02-16
cmake: Switch to `find_package(Python3)`.

waywardmonkeys:
The existing way of working used in this file was deprecated in cmake 3.12 and a new way supplied.

See https://cmake.org/cmake/help/latest/policy/CMP0148.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,8 +75,8 @@ include_directories(AFTER
   ${PROJECT_BINARY_DIR}/src
 )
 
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
This is partially done blind as the CI doesn't cover this (since it is cmake) and I did this on a macOS machine while the actual usage of the `Python3_EXECUTABLE` is on Windows.

khaledhosny:
I don’t use CMake or know much about it, so I’m going to blindly merge it. If it breaks something, we will soon know.

PR #4589, opened 2024-02-16
Use `noexcept` on swap, move constructors, etc.

waywardmonkeys:


behdad:
Thanks. What tool flags these?

behdad:
In general we build HB with `-fno-exceptions`. What functions in particular do you want marked `noexcept`, so I know in the future? Or is there a compiler flag to flag them?

waywardmonkeys:
I was using `clang-tidy`, looking to clean up a base set of things so that I could propose checking it in CI.

The usual thing is that move constructors, move assignment operators, and swap functions be marked as `noexcept`.

Interesting point about building with `-fno-exceptions` as I don't think the Rust binding of harfbuzz does that when it builds its own copy of harfbuzz.

PR #4599, opened 2024-02-25
[meson] Block uses of hb-static throughout Microsoft compilers

amyspark:
Hi,

This PR fixes a linking error when using `clang-cl` or `clang` as supplied by Visual Studio. PE-COFF cannot export extern, non-`dllexport` symbols in shared libraries, a case which is already handled for MSVC itself, but the check in `compiled_tests` does not consider one of the three Microsoft-supplied compilers.

I've also clarified the `TODO`s so as to list the underlying reason.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -682,7 +682,8 @@ if conf.get('HAVE_CAIRO', 0) == 1
 endif
 
 if get_option('tests').enabled()
-  # TODO: MSVC gives the following,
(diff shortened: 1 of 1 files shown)


behdad:
Thanks. The build fails seem to be in pixman and unrelated, yeah?

amyspark:
@behdad CI looks green now 👍

behdad:
Thanks.

PR #4616, opened 2024-03-08
[repacker] Rework how ClassDef sizes are estimated during splitting. 

garretrieger:
The existing implementation was not correctly computing the final serialized sizes of class def tables. It was computing an approximate incremental size increase based only on the next class to be added. I've update the estimator to keep a running total for the current split. This allows the format to switch back and forth as needed.

To confirm the new implementation is correct I updated the tests to compare to actual classdef serialization sizes.

### DIFF
--- a/src/graph/classdef-graph.hh
+++ b/src/graph/classdef-graph.hh
@@ -26,7 +26,6 @@
 
 #include "graph.hh"
 #include "../hb-ot-layout-common.hh"
-#include <cstdint>
 
 #ifndef GRAPH_CLASSDEF_GRAPH_HH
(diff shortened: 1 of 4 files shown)


src/graph/classdef-graph.hh
```diff
@@ -26,6 +26,7 @@
 
 #include "graph.hh"
 #include "../hb-ot-layout-common.hh"
+#include <cstdint>
```
behdad, line 29: Curious why you added this.

garretrieger: I think a previous iteration of my code for this used that, but it's not needed anymore. I'll remove.

Issue #4161, opened 2023-03-13
Subsets omit BASE table


skef:
I'm not seeing the BASE table in subsets with fairly generic arguments, including when I include `HB_SUBSET_FLAGS_PASSTHROUGH_UNRECOGNIZED` (which shouldn't change things as BASE is in the list of `known_tables`. 

(I understand there may be cases where BASE is tricky to subset, including removing or reducing axes in a variable font, but it seems like there are many safe cases where it could just be moved over as-is.)

source-serif has a BASE table if you want reference material: https://github.com/adobe-fonts/source-serif/tree/release/OTF

garretrieger:
BASE is in the default drop tables list: https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-subset-input.cc#L53 So to get it to passthrough you'll need to remove it from the drop tables list. Since we don't yet have subsetting code for BASE you'll also need to enable retain gids so the gid references in the table remain valid.

Let's leave this issue open to track implementation of BASE subsetting support.

skef:
Thanks, that's helpful.

It sounds like it would be fairly low-effort to start passing it through when there's no axis subsetting and when RETAIN_GIDS is specified.

behdad:
BASE subsetting doesn't look that complicated. Maybe @qxliu76 can work on it after she's done with partial-instancing.

garretrieger:
Instancing support was added in #4697

PR #4618, opened 2024-03-12
[subset] support BASE table subsetting

qxliu76:
This is for #4161 
Currently fonttools doesn't support BASE subsetting yet, so I have no golden output to compare with and did it based on the spec. Subsetting operation mainly affects baseScriptList(retained scripts only) , FeatMinMaxRecord (retained features only) and VarStore.


garretrieger (approved):

behdad:
Any chance you can also implement it for fonttools?

qxliu76:
> Any chance you can also implement it for fonttools?

yeah I can do that, but it pretty low priority

PR #4619, opened 2024-03-12
Fixing compiler warnings

jpcomito:
All warning are for format specifiers (-WFormat) in sprintf().

Test: Compile only

PR #4623, opened 2024-03-15
Use inttypes.h PRI?32 macros in format strings for [u]int32_t args.

behdad:
This is a modified and cleaned-up version of PR #4619 by @jpcomito: hb-common.h now expects 'inttypes.h' presence with MSVC exceptions, defines __STDC_FORMAT_MACROS in there (if not already defined) and undoes c930ae2 (PR #1974) and 902ab86.

### DIFF
--- a/src/hb-common.h
+++ b/src/hb-common.h
@@ -47,14 +47,10 @@
 # endif /* !__cplusplus */
 #endif
 
-#if defined (_SVR4) || defined (SVR4) || defined (__OpenBSD__) || \
-    defined (_sgi) || defined (__sun) || defined (sun) || \
-    defined (__digital__) || defined (__HP_cc)
-#  include <inttypes.h>
-#elif defined (_AIX)
+#if defined (_AIX)
 #  include <sys/inttypes.h>
-#elif defined (_MSC_VER) && _MSC_VER < 1600
-/* VS 2010 (_MSC_VER 1600) has stdint.h */
+#elif defined (_MSC_VER) && _MSC_VER < 1800
+/* VS 2013 (_MSC_VER 1800) has inttypes.h */
 typedef __int8 int8_t;
 typedef unsigned __int8 uint8_t;
 typedef __int16 int16_t;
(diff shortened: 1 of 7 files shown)


behdad:
@jpcomito does this work for you?

src/hb-common.h
```diff
@@ -47,14 +47,10 @@
...
-#if defined (_SVR4) || defined (SVR4) || defined (__OpenBSD__) || \
-    defined (_sgi) || defined (__sun) || defined (sun) || \
-    defined (__digital__) || defined (__HP_cc)
-#  include <inttypes.h>
-#elif defined (_AIX)
+#if defined (_AIX)
 #  include <sys/inttypes.h>
 #elif defined (_MSC_VER) && _MSC_VER < 1600
-/* VS 2010 (_MSC_VER 1600) has stdint.h */
+/* VS 2010 (_MSC_VER 1600) has inttypes.h */
```
sezero, line 53: VS2013 has inttypes.h, not 2010: You should add _MSC_VER 1800 special casings from the previous PR to avoid any possible clashes

behdad: Fixed. Thanks.

sezero: Hm, I hope a stdint.h from _MSC_VER < 1800 won't cause any clashes..

sezero: In any case, I'm OK with this version of the patch to solve #4619

src/hb.hh
```diff
@@ -213,6 +218,12 @@
 #include <winapifamily.h>
 #endif
 
+#ifndef PRId32
```
jpcomito, line 221: Do you think that it is worth adding some of the others (PRId16, PRIu16, PRId8, etc) to make a basic subset of the macros?

behdad: Let's add them as the need arises.

jpcomito: Sounds good. Thanks for doing this.

behdad: Good to merge?

jpcomito: Yes.

PR #4641, opened 2024-03-27
[subset] Update subset test suite to match some changes made to fonttools

garretrieger:
FontTools has added pruning functionality for code pages which we don't yet have in harfbuzz. Update our test runner to disable that for now. Also match fonttools and always recalc usFirst/LastCharIndex even if no prune unicode ranges is set.


behdad:
LGTM.

PR #4650, opened 2024-04-03
Meson: Use actual FreeType version when using CMake

fanc999-1:
Hi,

The Meson build files were updated lately to look for FreeType with a minimum version requirement, but then that broke the dependency search with CMake as FreeType's pkg-config files used a different versioning scheme.  This updates the Meson build files to account for that for CMake.

With blessings, thank you!


khaledhosny:
Do we really need to support CMake-built FreeType? I don’t like how it is special case in every aspect. IMHO, it should be fixed upstream: deprecate the `Freetype` dependency file (whatever it is called in CMake world) and add a `freetype2` one that uses the same versioning scheme as pkg-config file, instead of each project having to workaround this incompatibility.

PR #4672, opened 2024-04-19
[subset] If repacking fails for GSUB/GPOS retry with extension promotion.

garretrieger:
In rare cases a subsetting operation may have expanded the non-extension space data enough to make it unable to be repacked. So if we encounter a case that does not repack successfully retry the operation with extension promotion recalculation enabled.

We don't do extension promotion by default since it's rarely needed and it will slow down repacking + increase the repacking size when enabled and not needed.

### DIFF
--- a/src/hb-repacker.hh
+++ b/src/hb-repacker.hh
@@ -337,9 +337,10 @@ bool _process_overflows (const hb_vector_t<graph::overflow_record_t>& overflows,
 inline bool
 hb_resolve_graph_overflows (hb_tag_t table_tag,
                             unsigned max_rounds ,
-                            bool recalculate_extensions,
+                            bool always_recalculate_extensions,
                             graph_t& sorted_graph /* IN/OUT */)
 {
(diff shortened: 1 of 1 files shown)


behdad:
I dislike the option. I think we should just do what's right. I'm fine with promoting all lookups to extension if packing fails...

garretrieger:
> I dislike the option. I think we should just do what's right. I'm fine with promoting all lookups to extension if packing fails...

Alright, in that case I could do something like this: if repacking fails in space 0, then re-run the entire operation but with extension promotion enabled. How does that sound?

The downside is it will approx double repacking times in the rare cases which hit this scenario.

behdad:
Sounds good. Thanks.

garretrieger:
> Sounds good. Thanks.

Ok, I'll rework this.

garretrieger:
> > Sounds good. Thanks.
> 
> Ok, I'll rework this.



> > Sounds good. Thanks.
> 
> Ok, I'll rework this.

Done, please take another look.

behdad:
I don't fully understand but I trust you know what you are doing. :)

PR #4678, opened 2024-04-23
[subset] Offsets in BaseLangSysRecord are relative to the parent table and not the BaseLangSysRecord.

garretrieger:
https://learn.microsoft.com/en-us/typography/opentype/spec/base#baselangsysrecord-1


behdad:
Good catch.

PR #4679, opened 2024-04-25
[subset] collect variation indices for COLRv1

qxliu76:
Step 1 for subsetting/instantiating COLRv1: collect variation indices and generate inner_maps for subsetting usage(currently COLRv1 subsetting copies varStore as is).

Also merged 2 remap_variation_indices() APIs used by GDEF and BASE into 1, it will be used by COLRv1 as well.

### DIFF
--- a/src/OT/Color/COLR/COLR.hh
+++ b/src/OT/Color/COLR/COLR.hh
@@ -160,7 +160,10 @@ struct hb_colrv1_closure_context_t :
   { palette_indices->add (palette_index); }
 
   void add_var_idxes (unsigned first_var_idx, unsigned num_idxes)
-  { variation_indices->add_range (first_var_idx, first_var_idx + num_idxes - 1); }
+  {
+    if (!num_idxes || first_var_idx == VarIdx::NO_VARIATION) return;
+    variation_indices->add_range (first_var_idx, first_var_idx + num_idxes - 1);
+  }
 
   public:
   const void *base;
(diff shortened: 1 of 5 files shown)


src/OT/Color/COLR/COLR.hh
```diff
@@ -242,7 +251,11 @@ struct Variable
   }
 
   void closurev1 (hb_colrv1_closure_context_t* c) const
-  { value.closurev1 (c); }
+  {
+    // update c->num_var_idxes during value closure
+    value.closurev1 (c);
```
garretrieger, line 256: Can you init c->num_var_idxes to 0 before calling the closure. If the method doesn't happen to set a value for it then add_var_idxes below will just use whatever the previously set value is.

garretrieger (approved):

src/OT/Color/COLR/COLR.hh
```diff
@@ -159,23 +159,35 @@ struct hb_colrv1_closure_context_t :
   void add_palette_index (unsigned palette_index)
   { palette_indices->add (palette_index); }
 
+  void add_var_idxes (unsigned first_var_idx, unsigned num_idxes)
+  {
+    if (!num_idxes) return;
+    variation_indices->add_range (first_var_idx, first_var_idx + num_idxes - 1);
```
behdad, line 165: Also return if `first_var_idx` is `VarIdx::NO_VARIATION`

PR #4693, opened 2024-05-02
[instancer] Promote partial instancing API out of experimental.

garretrieger:


src/OT/name/name.hh
```diff
@@ -143,16 +143,13 @@ struct NameRecord
   }
 
   NameRecord* copy (hb_serialize_context_t *c, const void *base
-#ifdef HB_EXPERIMENTAL_API
                     , const hb_hashmap_t<hb_ot_name_record_ids_t, hb_bytes_t> *name_table_overrides
-#endif
```
khaledhosny, line 148: Please fix also the code formatting after removing the `#ifdef`'s.

behdad:
LGTM

src/hb-subset-instancer-iup.cc
```diff
@@ -443,11 +443,11 @@ static bool _iup_contour_optimize (const hb_array_t<const contour_point_t> conto
     unsigned contour_point_size = hb_static_size (contour_point_t);
     for (unsigned i = 0; i < n; i++)
     {
-      hb_memcpy ((void *) repeat_x_deltas.arrayZ, (const void *) x_deltas.arrayZ, n * sizeof (float));
-      hb_memcpy ((void *) (repeat_x_deltas.arrayZ + n), (const void *) x_deltas.arrayZ, n * sizeof (float));
+      hb_memcpy ((void *) repeat_x_deltas.arrayZ, (const void *) x_deltas.arrayZ, n * sizeof (int));
```
behdad, line 446: `sizeof (int)`?!

Maybe make those just `sizeof (repeat_x_deltas[0])`.

PR #4703, opened 2024-05-07
Fix double to float cast warning.

garretrieger:


behdad:
@garretrieger Can you fix the rest of them as well? I still see quite a few warnings.

Issue #4700, opened 2024-05-07
Partially instancing CFF appears to work, but outlines don't vary


simoncozens:
```
% hb-subset --variations=wght=400:900 --unicodes='*' Sans/Variable/OTF/NotoSansCJKjp-VF.otf -o regbold.otf
% shape regbold.otf abc
[gid66=0+563|gid67=1+618|gid68=2+510]
```
![shape](https://github.com/harfbuzz/harfbuzz/assets/106728/8a8f8ca4-c4f2-4c00-9990-f93972e9e663)

(This is correct at 400 weight, but:)

```
$ shape regbold.otf abc --variations=wght=900
[gid66=0+606|gid67=1+658|gid68=2+537]
```
![shape](https://github.com/harfbuzz/harfbuzz/assets/106728/e6beaeb6-bf3a-497e-a48e-88eb66f4c888)

Horizontal metrics have varied correctly, outlines have not.

I don't mind if it doesn't work yet, but it should give a warning or some indication to the user.

behdad:
I thought I implemented this. Will take a look.

behdad:
Oh nevermind. I implemented full instancing, not partial instancing IIRC. It's really complicated but doable...

@qxliu76 can you add a warning or error?

PR #4705, opened 2024-05-07
[instancer] add a warning for CFF partial instancing

qxliu76:
For #4700

### DIFF
--- a/src/hb-subset-cff2.cc
+++ b/src/hb-subset-cff2.cc
@@ -666,6 +666,9 @@ OT::cff2::accelerator_subset_t::serialize (hb_serialize_context_t *c,
 bool
 OT::cff2::accelerator_subset_t::subset (hb_subset_context_t *c) const
 {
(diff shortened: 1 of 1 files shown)


behdad:
CFF1 has no variations so no need there.

qxliu76:
Updated the PR. Thanks!

ebraminio:
Maybe subsetter can raise some similar warning for unsupported AAT layout tables subsetting, or maybe it could suggest adding using keep gid flag, sorry if that's already considered.

behdad:
Feel free to add l send PR.

khaledhosny:
This change is breaking the WASM port. It seems to cause filesystem support to be needed (it wasn’t before), and this seem to be problematic for the way we build hb-subset to WASM. I also think it is not appropriate for the library to printf stuff like this. I think this should be moved to the command line utility of possible, or be behind some ifdef that I can turn off when building to WASM.

behdad:
I agree library shouldn't print. Let's remove it. hb-subset cmdline can do if it wants.

PR #4708, opened 2024-05-09
usermanual-getting-started.xml Add Simon's order image

davelab6:
Add Simon's order image, from https://twitter.com/simoncozens/status/1422159944732839942/photo/1


dscorbett:
docs/features.dot has a more up-to-date source. The image should be generated when building the documentation, if possible, instead of being static in the repository.

behdad:
Perhaps some text to note that this is just an approximate and not how shaping is fully done.

Maybe also link to https://prezi.com/view/THNPJGFVDUCWoM20syev/ as well

rsheeter:
> Perhaps some text to note that this is just an approximate and not how shaping is fully done.

Is it infeasible to have it be correct? Failing that, to note where it is not?

behdad:
> > Perhaps some text to note that this is just an approximate and not how shaping is fully done.
> 
> Is it feasible to have it be correct? Failing that, to note where it is not?

HB individual shapers do a lot of logic. That's not reflected in the graph. Also, just for example, recent versions of HB, to match Windows, apply `rclt` only if `calt` feature was NOT found in the font.

behdad:
@simoncozens can you share the source dot file of this please? Closing until then.

khaledhosny:
It is already in the repo https://github.com/harfbuzz/harfbuzz/blob/main/docs/features.dot, it is probably outdated though (but the PNG is even more outdated).

behdad:
> It is already in the repo https://github.com/harfbuzz/harfbuzz/blob/main/docs/features.dot, it is probably outdated though (but the PNG is even more outdated).

But we don't consume it, right? maybe worse finishing.

khaledhosny:
No, we don’t use it. It needs to be reviewed against the current code (I think there was some misunderstanding when it was created as well).

PR #4709, opened 2024-05-09
[subset-cff] Shorten output of `encode_num_tp`

dscorbett:
This optimizes `encode_num_tp` to return shorter outputs. It is ported from fonttools/fonttools#3492. The reviewer should be extra careful because, unlike in fontTools, HarfBuzz has no unit tests for this function, and also because I am not fluent in C++.


behdad:
Would you have time to add the tests? Or one of us would need to do that? I think it would be a good idea.

src/hb-subset-cff-common.hh
```diff
@@ -133,13 +133,56 @@ struct str_encoder_t
       (void) hb_uselocale (((void) freelocale (clocale), oldlocale));
 
       char *s = buf;
+      size_t len;
+      char *comma = strchr (s, ',');
+      if (comma)
+	*comma = '.';
```
behdad, line 139: Is this because of locales? That shouldn't happen. What's this for?

dscorbett: A preexisting comment later in the function says “Comma for some European locales in case no uselocale available”, so I assume it can happen. I added this line so that rewrite rules that check for the decimal point would work in that situation.

behdad: Ughh. Okay.

behdad:
> I can try to add tests. Should they go in test/api/test-subset-cff2.c or a new file?

No, keep them easy in `src/test-cff.cc`

PR #4710, opened 2024-05-11
Trying to partial-instance CFF2

behdad:
Honestly I'm lost. But I'll find my way.


behdad:
What needs to be done from what I currently understand:

- The private-dict always processes blends. We need a specialization for the subset case that stores them, like the cs-interpretter does.

- Both need to gain support for partial instancing. Not sure were to put all the region expensions. Perhaps the CFF2 subset accel.

- The varstore needs to be pruned of unused regions.

behdad:
Humm. Maybe easier to do in fonttools first...

behdad:
> I actually implemented this in FontTools. Wasn't too bad there. I hope we can use some of the same design here.

https://github.com/fonttools/fonttools/pull/3506

behdad:
I rebased this but don't think I'd have time any time soon to get back to this. :(

behdad:
This is ready for review.

garretrieger (approved):
Did a high level review, the approach looks good to me.

PR #4711, opened 2024-05-11
[aat] Use set digest to speed up

behdad:
Still not cached in an accelerator, but shows up promising results.

### DIFF
--- a/src/hb-aat-layout-common.hh
+++ b/src/hb-aat-layout-common.hh
@@ -899,10 +899,8 @@ struct StateTableDriver
   using EntryT = Entry<EntryData>;
 
   StateTableDriver (const StateTableT &machine_,
-		    hb_buffer_t *buffer_,
 		    hb_face_t *face_) :
 	      machine (machine_),
-	      buffer (buffer_),
 	      num_glyphs (face_->get_num_glyphs ())
   {
     machine.collect_glyphs (glyph_set, num_glyphs);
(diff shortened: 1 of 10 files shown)


behdad:
And cached now. I see 25% speedup for en-thelittleprince benchmark, and 8% for en-words.

behdad:
This is with morx only. Kerx is not done yet.

behdad:
> This is with morx only. Kerx is not done yet.

Done for kern / kerx as well.

en-thelittleprince benchmark is 30% faster, en-words is 10% faster with HelveticaNeue.

drott:
@kojiishi [kicked off benchmarks](https://pinpoint-dot-chromeperf.appspot.com/job/1422a2be010000) and initial measurements show an improvement in the Editor-TipTap 10% subtest of speedometer3 (long text shaping: french, CJK and code) and an overall speedometer score 1.2% improvement in Chromium. This is great, thank you! LGTU, happy to see this landed. Not sure if the fuzzer findings are caused by this set of changes or from main.

behdad:
Yeah the CIFuzz breakage is real. I'm looking into it.

behdad:
I think I fixed the fuzzer issue. In that case this is ready to be merged.

jfkthame:
Very nice! Maybe worth a new release sometime soon?

behdad:
> Very nice! Maybe worth a new release sometime soon?

Yes, I just asked @khaledhosny for one.

PR #4712, opened 2024-05-13
[meson] build ragel as a native tool

robUx4:
When cross compiling it will build a target (host) binary that can't be run on the building machine.

Also remove the install flag which is not compatible with the native flag, but the custom built ragel is correctly picked.


khaledhosny:
Thanks!

PR #4719, opened 2024-05-13
[aat] For short words, use buffer digest to skip morx / kerx subtables

behdad:
Shows 5% speedup in en-words test case with Helvetica Neue.

### DIFF
--- a/src/hb-aat-layout-common.hh
+++ b/src/hb-aat-layout-common.hh
@@ -39,6 +39,7 @@ namespace AAT {
 
 using namespace OT;
 
+#define HB_AAT_BUFFER_DIGEST_THRESHOLD 32
 
 struct ankr;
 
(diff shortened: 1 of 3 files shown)


behdad:
I don't know if it's worth it. @jfkthame @drott any ideas?

I don't know who still shapes word at a time.

jfkthame:
Seems worthwhile to me. There are lots of text fragments that are short overall, even if not considering word boundaries as such.

behdad:
> Seems worthwhile to me. There are lots of text fragments that are short overall, even if not considering word boundaries as such.

Thanks. Can you do a quick logic check of my work?

jfkthame:
I don't claim to really have a good grasp of this code, but it looks reasonable enough. The one thing I wonder is whether 16 is the appropriate threshold.... have you experimented with increasing it (e.g. to 32) to see what the trade-offs look like? (Obviously at some point the cost of creating the buffer digest will be too great to be worthwhile, on average.)

behdad:
Thanks. I changed it to 32. Can't detect any difference since my two test cases one includes words (which are typically under 16), and the other includes paragraphs. So neither test particularly falls between 16 and 32.

I also implemented it for state-machine kerning.

Issue #4558, opened 2024-01-18
Subsetting fails on certain fonts


jackyjjc-canva:
```
$ hb-subset GlowSansSC-Normal-Bold.otf --text=abc -o glowsansj.subset.otf
error: Operation failed. Probably a bug. File github issue.
```

The font was an open source font downloaded from https://github.com/welai/glow-sans/releases. (`GlowSansSC-Normal-Bold.otf` from "GlowSansSC-Normal-v0.93.zip")

It might be related to this previous issue since the error is the same: https://github.com/harfbuzz/harfbuzz/issues/3959

behdad:
Thanks. Unlikely to be the same issue. The error message is generic.

martyla:
Was digging into this a little bit... It looks like a regression introduced in the 8.0.0 release (hb-subset command succeeds for release <=7.3.0)

Seems to have an issue subsetting the CFF table: https://github.com/harfbuzz/harfbuzz/blob/bc90b29b37fe3809f9e48aa7be08fbf2208e481a/src/hb-subset.cc#L510

And it succeeds if I disable CFF subsetting (via setting cpp flag `-DHB_NO_SUBSET_CFF`)

Debugging with `-DHB_DEBUG_SUBSET=100`:
```
$ ./build/util/hb-subset GlowSansSC-Normal-Bold.otf "abc" -o out.otf
SUBSET                          ├╴: subset BASE
SUBSET                          ├╴: OT::BASE initial estimated table size: 8194 bytes.
... <stripped for brevity>
SUBSET    (     0x10419011c)  1 │├╯OT::BASE::subset: return true (line 779)
SUBSET                          ├╴: OT::BASE final subset table size: 222 bytes.
SUBSET                          ├╴: add table BASE, dest 222 bytes, source 222 bytes
SUBSET                          ├╴: OT::BASE::subset success
SUBSET                          ├╴: subset CFF
SUBSET                          ├╴: OT::CFF ::subset sanitize failed on source table.
error: Operation failed. Probably a bug. File github issue.
```

Unfortunately I'm not familiar enough yet with the code to understand how to fix this

behdad:
Can you share the font with us?

martyla:
[GlowSansSC-Normal-v0.93.zip](https://github.com/welai/glow-sans/releases/download/v0.93/GlowSansSC-Normal-v0.93.zip) -> `GlowSansSC-Normal-Bold.otf`

behdad:
[...]

PR #4725, opened 2024-05-24
[CFF] Ignore unknown operators

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/4558


behdad:
FreeType accepts the font on the issue; so I think it does ignore unknown ops at least in some places.

garretrieger (approved):

PR #4727, opened 2024-05-25
[CMake] Fix C++20

OgreTransporter:
Enable C++20 in CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,7 @@
 cmake_minimum_required(VERSION 3.12)
 project(harfbuzz)
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 message(WARN "HarfBuzz has a Meson port and tries to migrate all the other build systems to it, please consider using it as we might remove our cmake port soon.")
 


behdad:
Why?!

OgreTransporter:
https://github.com/harfbuzz/harfbuzz/blob/484cb2608e0253c28c1a9963e12de2293bf36065/src/hb-geometry.hh#L143
`and` only works with C++20. To ensure that the compilers also use C++20 under CMake, these two lines must be added.

behdad:
> https://github.com/harfbuzz/harfbuzz/blob/484cb2608e0253c28c1a9963e12de2293bf36065/src/hb-geometry.hh#L143
> 
> 
> `and` only works with C++20. To ensure that the compilers also use C++20 under CMake, these two lines must be added.

Huh! That was clearly a mistake. Fixed. I'm surprised none of our bots caught that.

khaledhosny:
> > https://github.com/harfbuzz/harfbuzz/blob/484cb2608e0253c28c1a9963e12de2293bf36065/src/hb-geometry.hh#L143
> > 
> > `and` only works with C++20. To ensure that the compilers also use C++20 under CMake, these two lines must be added.
> 
> Huh! That was clearly a mistake. Fixed. I'm surprised none of our bots caught that.

Because it is C++98 https://stackoverflow.com/a/2419889

OgreTransporter:
But then you must also include the appropriate header for the macros.
https://cplusplus.com/reference/ciso646/

behdad:
> But then you must also include the appropriate header for the macros.
> https://cplusplus.com/reference/ciso646/

That seems to be for C only?

khaledhosny:
```
$ cat a.cc
int
main(int argc, char **argv)
{
  if (true and true)
    return 0;
}
$ c++ -std=c++98 a.cc
```
Compiles without any errors.

OgreTransporter:
The code does not compile for me, but now I know why:
> C++ specifies and as an alternative spelling for &&. In C, the alternative spelling is provided as a macro in the <iso646.h> header. In C++, the alternative spelling is a keyword; use of <iso646.h> or the C++ equivalent <ciso646> is deprecated. In Microsoft C++, the [/permissive-](https://learn.microsoft.com/en-us/cpp/build/reference/permissive-standards-conformance?view=msvc-170) or [/Za](https://learn.microsoft.com/en-us/cpp/build/reference/za-ze-disable-language-extensions?view=msvc-170) compiler option is required to enable the alternative spelling.

OgreTransporter:
To solve the problem, you must either set the compiler to C++20, set the /permissive- flag or omit `and`.
> The /permissive- option is implicitly set by the /std:c++latest option starting in Visual Studio 2019 version 16.8, and in version 16.11 by the /std:c++20 option. /permissive- is required for C++20 Modules support. Perhaps your code doesn't need modules support but requires other features enabled under /std:c++20 or /std:c++latest. You can explicitly enable Microsoft extension support by using the /permissive option without the trailing dash. The /permissive option must come after any option that sets /permissive- implicitly.

behdad:
I omitted the `and` already.

behdad:
We'll get a release out soon.

Issue #4671, opened 2024-04-19
harfbuzz-icu fails to build with ICU 75


dschepler:
The error I'm getting is:
```
FAILED: src/libharfbuzz-icu.so.0.60840.0.p/hb-icu.cc.o 
c++ -Isrc/libharfbuzz-icu.so.0.60840.0.p -Isrc -I../../src -I. -I../.. -fdiagnostics-color=always -D_GLIBCXX_ASSERTIONS=1 -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=c++11 -fno-exceptions -O3 -fno-exceptions -fno-rtti -fno-threadsafe-statics -fvisibility-inlines-hidden -DHAVE_CONFIG_H -fPIC -Wno-non-virtual-dtor -MD -MQ src/libharfbuzz-icu.so.0.60840.0.p/hb-icu.cc.o -MF src/libharfbuzz-icu.so.0.60840.0.p/hb-icu.cc.o.d -o src/libharfbuzz-icu.so.0.60840.0.p/hb-icu.cc.o -c ../../src/hb-icu.cc
In file included from /usr/include/unicode/uset.h:36,
                 from /usr/include/unicode/unorm2.h:35,
                 from ../../src/hb-icu.cc:39:
/usr/include/unicode/localpointer.h:561:26: error: parameter declared 'auto'
  561 | template <typename Type, auto closeFunction>
      |                          ^~~~
/usr/include/unicode/localpointer.h:573:76: error: template argument 2 is invalid
  573 |     explicit LocalOpenPointer(std::unique_ptr<Type, decltype(closeFunction)> &&p)
      |                                                                            ^
/usr/include/unicode/localpointer.h:583:78: error: template argument 2 is invalid
  583 |     LocalOpenPointer &operator=(std::unique_ptr<Type, decltype(closeFunction)> &&p) {
      |                                                                              ^
/usr/include/unicode/localpointer.h:599:59: error: template argument 2 is invalid
  599 |     operator std::unique_ptr<Type, decltype(closeFunction)> () && {
      |                                                           ^
/usr/include/unicode/uset.h:358:1: note: invalid template non-type parameter
  358 | U_DEFINE_LOCAL_OPEN_POINTER(LocalUSetPointer, USet, uset_close);
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/unicode/unorm2.h:286:1: note: invalid template non-type parameter
[...]

PR #4733, opened 2024-05-27
build: bump c++ version to c++17

kanavin:
This is required by icu 75 as reported in
https://github.com/harfbuzz/harfbuzz/issues/4671

and matches GCC 13+ and clang16+ defaults.


behdad:
That would be too recent for us. We want to only do it if ICU is enabled and requires it preferably.

kanavin:
Alright. I opened a replacement pull request:
https://github.com/harfbuzz/harfbuzz/pull/4734

Issue #2283, opened 2020-03-27
[subset] Expand the input unicode set to include decompositions and compositions.


garretrieger:
Since the shaper may automatically apply decompositions or compositions of characters even if there are no GSUB rules in a font for them. For a subset to be fully correct we should automatically expand the input unicode set to include these possible substitutions.

For example say we want to cut a subset for the text 0x41 (A), 0x30A (combining ring). For that text the shaper may choose to use 0xC5 (A with ring) instead of 0x41 + 0x30A. Today the subsetter will only retain glyphs for A and combining ring, which means that it'll be missing the glyph for A with ring.

I wrote a proof of concept implementation for a closure algorithm in java: https://gist.github.com/garretrieger/9b470871ae5c2159d2c083aa072d25b1. I need to port this to C++ to use in harfbuzz.

khaledhosny:
> For that text the shaper may choose to use 0xC5 (A with ring) instead of 0x41 + 0x30A. Today the subsetter will only retain glyphs for A and combining ring, which means that it'll be missing the glyph for A with ring.

HarfBuzz will not apply composition if the composed glyph is missing from the font. So if the font has glyphs for only 0x41 + 0x30A, 0xC5 will be decomposed by 0x41 + 0x30A will never be composed.

garretrieger:
Right, which can mean that the rendering for the subset will end up being different than for the whole font where the single combined glyph is used instead of the two separate glyphs.

A real example of this that we encountered was in Bengali for the font Hind Siliguri, the subset text was 0x09af + 0x09cc. With the whole font the shaper decomposes this to: 0x9c7 + 0x9af + 0x9d7 and everything renders correctly. When the font is subset to {0x9af, 0x9cc} currently the glyphs for 0x9c7 and 0x9d7 get dropped and then the subsetted font incorrectly renders.

khaledhosny:
But isn’t this is a slightly different issue, as it is not canonical decomposition per se. I believe this is script shaping logic and does not always correspond to the normalization rules, and should be handled regardless of how normalization is handled.

behdad:
> But isn’t this is a slightly different issue, as it is not canonical decomposition per se. I believe this is script shaping logic and does not always correspond to the normalization rules, and should be handled regardless of how normalization is handled.

Right. That knowledge needs to be reflected somewhere and used in both subsetters.

behdad:
[...]

PR #4747, opened 2024-06-06
[subset] decompose Unicodes

behdad:
Part of https://github.com/harfbuzz/harfbuzz/issues/2283

### DIFF
--- a/src/hb-subset-plan.cc
+++ b/src/hb-subset-plan.cc
@@ -742,6 +742,7 @@ _fill_unicode_and_glyph_map(hb_subset_plan_t *plan,
     plan->unicode_to_new_gid_list.push (hb_pair (cp, gid));
   }
 
+  return; // disabled for now
   _unicode_closure (plan->unicode_to_new_gid_list,
 		    start,
 		    *plan->codepoint_to_glyph,
--- a/src/hb-subset-plan.cc
+++ b/src/hb-subset-plan.cc
@@ -552,7 +552,7 @@ _unicode_closure (hb_sorted_vector_t<hb_pair_t<hb_codepoint_t, hb_codepoint_t>> 
   hb_script_t last_script = HB_SCRIPT_INVALID;
   hb_direction_t direction = HB_DIRECTION_INVALID;
   const hb_tag_t gsub_script = HB_TAG_NONE;
-  const hb_ot_shaper_t *shaper = hb_ot_shaper_categorize (last_script, direction, gsub_script);
+  const hb_ot_shaper_t *shaper = hb_ot_shaper_categorize (last_script, direction, gsub_script, false);
 
   hb_ot_shape_normalization_mode_t mode = shaper->normalization_preference;
   hb_ot_shape_normalize_context_t c = {
@@ -574,7 +574,7 @@ _unicode_closure (hb_sorted_vector_t<hb_pair_t<hb_codepoint_t, hb_codepoint_t>> 
     {
       last_script = script;
       direction = hb_script_get_horizontal_direction (script);
-      shaper = hb_ot_shaper_categorize (last_script, direction, gsub_script);
+      shaper = hb_ot_shaper_categorize (last_script, direction, gsub_script, false);
       mode = shaper->normalization_preference;
       c.override_decompose_and_compose (shaper->decompose, shaper->compose);
     }
--- a/src/hb-subset-plan.cc
+++ b/src/hb-subset-plan.cc
@@ -26,8 +26,6 @@
 
 #include "hb-subset-plan.hh"
 #include "hb-subset-accelerator.hh"
-#include "hb-ot-shape-normalize.hh"
-#include "hb-ot-shaper.hh"
 #include "hb-map.hh"
 #include "hb-multimap.hh"
 #include "hb-set.hh"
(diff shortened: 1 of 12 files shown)


behdad:
Linking is going to be ugly. I'd have to move the per-shaper decomposition logic to a header, which I don't like, but probably the easiest.

@jfkthame any opinions about the logic here?

behdad:
I've been testing this with:
```
$ hb-subset NotoSansBengali-Regular.ttf -u 995,9cb --output-file out.ttf
$ hb-view out.ttf -u 995,9cb
```

src/hb-subset-plan.cc
```diff
@@ -768,8 +853,10 @@ _populate_unicodes_to_retain (const hb_set_t *unicodes,
   }
 
   auto &arr = plan->unicode_to_new_gid_list;
+
   if (arr.length)
   {
+    arr.qsort (); // Decompositions are added in non-sorted order
```
garretrieger, line 859: Since this is only needed if decompositions were added, maybe we can have the _fill_unicode_and_glyph_map return a bool to indicate if decompositions were added and only sort here when needed.

Also in some cases the list get's qsort'd prior to getting here (eg. line 825)

src/hb-subset-plan.cc
```diff
@@ -656,6 +736,11 @@ _fill_unicode_and_glyph_map(hb_subset_plan_t *plan,
     plan->codepoint_to_glyph->set (cp, gid);
     plan->unicode_to_new_gid_list.push (hb_pair (cp, gid));
   }
+
+  _unicode_closure (plan->unicode_to_new_gid_list,
+		    start,
+		    *plan->codepoint_to_glyph,
+		    unicode_to_gid);
```
garretrieger, line 743: The "unicode_to_gid" function is only valid for codepoints in unicode_iterator on each particular invocation of _fill_unicode_and_glyph_map and can't be used as a general unicode to gid lookup. For example line 812 the function is fixed to just always return a specific gid because this function is invoked only for codepoints that map to that gid. We probably need to add an additional templated arg that provides a general mapper which is valid for all unicodes in the font.

behdad: What an ouch... Yeah I think we might need that then, if we want to pursue this approach.

garretrieger: Here's my pass at it: https://github.com/harfbuzz/harfbuzz/pull/4758

behdad: Thanks. Rebased to use that.

behdad:
I'm still unsure how to address the linking issue here. Perhaps we can add API to libharfbuzz that would return the compose/decompose functions *per* script?

@garretrieger would you be able to address the rest of the issues (the ones you identified)? Thanks.

behdad:
> I'm still unsure how to address the linking issue here. Perhaps we can add API to libharfbuzz that would return the compose/decompose functions _per_ script?

Easiest: `hb_unicode_funcs_create_for_script(hb_unicode_funcs_t *parent, hb_script_t script)`. Should crash if parent is `nullptr` instead of assuming a nil Unicode-funcs. That would be too slippery of a slope. Instead, client should pass `hb_unicode_funcs_get_default()` as parent in most cases.

behdad:
> > I'm still unsure how to address the linking issue here. Perhaps we can add API to libharfbuzz that would return the compose/decompose functions _per_ script?
> 
> Easiest: `hb_unicode_funcs_create_for_script(hb_unicode_funcs_t *parent, hb_script_t script)`. Should crash if parent is `nullptr` instead of assuming a nil Unicode-funcs. That would be too slippery of a slope. Instead, client should pass `hb_unicode_funcs_get_default()` as parent in most cases.

Two issues with the above:

- The chosen `compose`/`decompose` are per-shaper, not per-script. So we also need the font to decide which shaper to use since the same script may be routed to different shapers depending on which script system tags are available in the font (indic2 vd indic3 for example).

- Currently, the per-shaper `compose`/`decompose` also take the shape plan. That is used in the Hebrew decompose function to only do some compatibility composition if the font does NOT have gpos mark feature. My life would be a lot easier if I can remove this hack. @jfkthame Is it time we can remove this, or you don't think so?

behdad:
> * The chosen `compose`/`decompose` are per-shaper, not per-script. So we also need the font to decide which shaper to use since the same script may be routed to different shapers depending on which script system tags are available in the font (indic2 vd indic3 for example).

``
hb_unicode_funcs_t *
hb_face_unicode_funcs_create(
    hb_face_t *face,
    hb_script_t script,
    hb_unicode_funcs_t *parent)`.

> * Currently, the per-shaper `compose`/`decompose` also take the shape plan. That is used in the Hebrew decompose function to only do some compatibility composition if the font does NOT have gpos mark feature. My life would be a lot easier if I can remove this hack.

One way this can be worked around is to define two separate shapers: a `hebrew` and `hebrew_fallback` and choose the different shaper if GPOS doesn't have mark feature. Though now the hardship would be that, that information is not yet available during shaper selection. Shouldn't be too hard to fix though, maybe.

behdad:
> ```c++
> hb_unicode_funcs_t *
> hb_face_unicode_funcs_create(
>     hb_face_t *face,
>     hb_script_t script,
>     hb_unicode_funcs_t *parent);
> ```

Annoyingly, `hb_ot_shaper_categorize()` takes both script, and direction. Because:
```
      /* For Arabic script, use the Arabic shaper even if no OT script tag was found.
       * This is because we do fallback shaping for Arabic script (and not others).
       * But note that Arabic shaping is applicable only to horizontal layout; for
       * vertical text, just use the generic shaper instead. */
      if ((gsub_script != HB_OT_TAG_DEFAULT_SCRIPT || script == HB_SCRIPT_ARABIC) &&
          HB_DIRECTION_IS_HORIZONTAL (direction))
        return &_hb_ot_shaper_arabic;
      else
        return &_hb_ot_shaper_default;
```
I'm tempted to ignore that case in the new API and only pass the default horizontal direction of the script instead. The same way that I want to use `HB_LANGUAGE_INVALID` when resolving the OT script/langsys.

behdad:
> ```c++
> hb_face_unicode_funcs_create
> ```

Oh, or better:
```cpp
hb_unicode_funcs_t *
hb_ot_shape_plan_unicode_funcs_create(
    hb_shaple_plan_t *shape_plan,
    hb_unicode_funcs_t *parent
);
```

behdad:
Okay. I implemented what I came up with, which addresses the linkage issue. However, unlike the previous prototype that only did this extra compose/decompose when the shaper had a non-default normalization mode (that made 3 tests fail), I'm now doing it all the time, and makes 23 tests fail. Needs investigation and adjustment.

PR #4748, opened 2024-06-06
[subset] refactor populate_unicodes_to_retain.

garretrieger:
Pulls out commonly repeated pattern of filling the gid and unicode maps into a templated helper.


behdad:
LGTM.

PR #4753, opened 2024-06-13
Use find_package for ICU

donny-dont:
ICU 75.1 requires C++17 support so use `find_package(ICU)` which determines the versioning. Sets `CMAKE_CXX_STANDARD` when that happens.

Follows along with what Meson build did in 93d58f83.


donny-dont:
Thank you @behdad !

Issue #4774, opened 2024-06-28
Bug in `unicode_to_macroman`?


behdad:
From https://github.com/RazrFalcon/rustybuzz/pull/115#issuecomment-2191329791

Looks like there is a bug in `unicode_to_macroman`. It returns 0 for 0x0131, but 0x0131 is in the table.
Is it expected or is it because you use `hb_bsearch` on an unsorted array?

behdad:
Ideas welcome. One way is to have a two-level switch based on the second byte then first byte. Probably good enough. Another is to create a static hashmap on demand, which then needs to be cleaned up at shutdown (a bit of a pain but doable).

behdad:
Or just a plane switch and let the compiler bsearch it.

behdad:
This line is also very curious for using `0x7F` instead of `0x80`:
```
return c ? (c - mapping) + 0x7F : 0;
```

PR #4775, opened 2024-06-28
[cmap] Fix macroman lookup

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/4774

Untested.


behdad:
Would be good to add a test here. I don't remember when I added this code and what font I was testing with.

PR #4793, opened 2024-07-08
[subset] Use hb_clamp instead of consequent hb_min and hb_max calls

ebraminio:
As just a minor code tweak to clarify the intention better.


behdad:
Thanks @ebraminio. Good to see you.

PR #4803, opened 2024-07-16
meson: Fix builds against ICU >= 75.x on Visual Studio

fanc999-1:
Hi,

This brings PR #4734 to also work with Visual Studio as it expects `/std:c++17` and does not understand `-std=c++17`, so that builds on Visual Studio against ICU >= 75.x will work without passing `-Dcpp_std=c++17` in the Meson setup command line.

With blessings, thank you!


fanc999-1:
Thanks @behdad!

PR #4835, opened 2024-08-20
[subset] bug fix in post table

qxliu76:
Bug fix: insert retain-gid holes when both retain-gids and glyph-names are enabled. 
This fix would use 0 for empty glyphs, which refers to the name .notdef, as the glyphNameIndex entry for that glyph ID.

Unfortunately the result doesn't match fonttools' because fonttools just retains all glyph names, and a fix in fonttools seems tricky because fonttools relies on post compilation.


garretrieger (approved):

yisibl:
Can you file an issue with fonttools recording this problem?

Issue #2953, opened 2021-04-16
Flexible-array hack: Out of bounds issues in hb-open-type.hh


qarmin:
When compiling Godot with ubsan sanitizer, then I got this errors
```
thirdparty/harfbuzz/src/hb-open-type.hh:584:18: runtime error: index 13 out of bounds for type 'TableRecord [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:584:18: runtime error: index 13 out of bounds for type 'TableRecord [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:672:11: runtime error: index 2 out of bounds for type 'Record [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:672:11: runtime error: index 2 out of bounds for type 'Record [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:672:11: runtime error: index 2 out of bounds for type 'Record [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:672:11: runtime error: index 2 out of bounds for type 'Record [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:672:11: runtime error: index 2 out of bounds for type 'OffsetTo [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:672:11: runtime error: index 2 out of bounds for type 'OffsetTo [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:672:11: runtime error: index 2 out of bounds for type 'OffsetTo [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:672:11: runtime error: index 2 out of bounds for type 'OffsetTo [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:672:11: runtime error: index 2 out of bounds for type 'OffsetTo [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:672:11: runtime error: index 2 out of bounds for type 'OffsetTo [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:722:29: runtime error: index 2 out of bounds for type 'OffsetTo [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:722:29: runtime error: index 2 out of bounds for type 'OffsetTo [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:672:11: runtime error: index 2 out of bounds for type 'MarkRecord [1]'
[...]

PR #4836, opened 2024-08-22
Make codebase compatible with -fstrict-flex-array={2,3}

serge-sans-paille:
The implementation of HBUINT32VAR relies on a flexible array member, and wants to enforce a size of at least one. It uses a member of

        unsigned char[1]

field to achieve that role, but this syntax is a GNU extension that is *not* considered as a flexible array member under -fstrict-flex-array=2 and -fstrict-flex-array=3.

Use a more explicit implementation to achieve the intended goal.

### DIFF
--- a/src/hb-open-type.hh
+++ b/src/hb-open-type.hh
@@ -137,7 +137,7 @@ struct HBUINT32VAR
 {
   unsigned get_size () const
   {
-    unsigned b0 = v[0];
+    unsigned b0 = v.head;
     if (b0 < 0x80)
       return 1;
     else if (b0 < 0xC0)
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
This is for both clang and gcc.

I see that `HB_VAR_ARRAY`  is defined to 1 by default, which does not play well with compiler, sizeof etc (see for instance https://www.kernel.org/doc/html/v5.16/process/deprecated.html#zero-length-and-one-element-arrays and also https://developers.redhat.com/articles/2022/09/29/benefits-limitations-flexible-array-members#motivation_for_stricter_standard_conformance). Forcing `HB_VAR_ARRAY` would be an option but this does not interact correctly with the assertion in `DEFINE_SIZE_MIN` (it leads to a sizeof 0). Same foes with defining  `HB_VAR_ARRAY` as empty.

This is why I've enforced a size of one in the patch through an explicit structure.

behdad:
> This is for both clang and gcc.

Ah, there was a typo in the title. It should read `-fstrict-flex-arrays` with an `s` at the end, isn't it? With that it reproduces with with clang++ but not g++.

> I see that `HB_VAR_ARRAY` is defined to 1 by default, which does not play well with compiler, sizeof etc (see for instance https://www.kernel.org/doc/html/v5.16/process/deprecated.html#zero-length-and-one-element-arrays and also https://developers.redhat.com/articles/2022/09/29/benefits-limitations-flexible-array-members#motivation_for_stricter_standard_conformance). Forcing `HB_VAR_ARRAY` would be an option but this does not interact correctly with the assertion in `DEFINE_SIZE_MIN` (it leads to a sizeof 0). Same foes with defining `HB_VAR_ARRAY` as empty.

Right. We've hammered this issue to death before:

  https://github.com/harfbuzz/harfbuzz/issues/2953

I'll read your links and get back to you. I just don't like using flex arrays here whereas we don't use in other parts of the code.

> This is why I've enforced a size of one in the patch through an explicit structure.

behdad:
> I'll read your links and get back to you.

I think you only get the warning for this struct because it explicitly accesses trailing bytes by constants. Otherwise, we have a lot of flexible arrays using 1.

I'm leaning towards using `HB_VAR_ARRAY` and enforcing (in `hb.hh`) `-fstrict-flex-array=1`. Even if needed, taking your patch then adding `HB_VAR_ARRAY` to `tail`.

My idea for `HB_VAR_ARRAY` always was to be able to define it to empty on g++/clang++. But unfortunately that leads to other errors, because some (legitimate) structs become 0-length and that's not allowed in C++ at least.

behdad:
Alternatively just changing `v[1]` to `v[5]` seems fine to me.

behdad:
Humm. Looks like this line doesn't work:
```
#pragma GCC diagnostic error   "-Wstrict-flex-arrays=3"
```

Anyone know how to set it to 3 from the source code?

behdad:
I pushed a fix. Please let me know if you find it inadequate.

serge-sans-paille:
I'm fine with everything you find fine :-)
I liked the idea of a structure using a minimal amount of memory, but that's definitively up to you.

behdad:
> I'm fine with everything you find fine :-) I liked the idea of a structure using a minimal amount of memory, but that's definitively up to you.

That pattern already exists elsewhere. That's why we have `DEFINE_SIZE_MIN(1)`.

PR #4841, opened 2024-08-27
Smalll fix on documentation

leandrosansilva:
there is no `hb_ft_font_has_changed()`. Instead, it should be `hb_ft_font_changed()`.


khaledhosny:
Thanks!

PR #4845, opened 2024-08-30
[subset] remove unnecessary check on name IDs

qxliu76:
From https://github.com/fonttools/fonttools/pull/3617
See: https://learn.microsoft.com/en-us/typography/opentype/spec/features_ae

The spec says they are expected to be in the font-specific name ID range (256–32767), though that is not a requirement in this Feature Parameters specification.

### DIFF
--- a/src/hb-ot-layout-common.hh
+++ b/src/hb-ot-layout-common.hh
@@ -646,8 +646,7 @@ struct FeatureParamsCharacterVariants
       return;
 
     unsigned last_name_id = (unsigned) firstParamUILabelNameID + (unsigned) numNamedParameters - 1;
-    if (last_name_id >= 256 && last_name_id <= 32767)
-      nameids_to_retain->add_range (firstParamUILabelNameID, last_name_id);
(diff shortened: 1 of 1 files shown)


src/hb-ot-layout-common.hh
```diff
@@ -646,8 +646,7 @@ struct FeatureParamsCharacterVariants
       return;
```
khaledhosny, line 646: Do we want to also drop the check for `numNamedParameters >= 0x7FFF` above?

behdad: Looks suspicious indeed.

qxliu76: oh yeah, I think I just ported it from old fonttools' code

qxliu76: I'll remove this check in another PR, I'm going to add a testcase from Cosimo

qxliu76: I found another issue in fonttools related with FeatureParams, so I'll add a testcase later after that got fixed

PR #4863, opened 2024-09-22
Follow up to variation-selector-not-found glyph

behdad:
Addresses https://github.com/harfbuzz/harfbuzz/pull/4529#discussion_r1769638033

I'm not sure if this is an improvement. By leaving the var-selector as default-ignorable, ligatures can form around it, and the resulting cluster won't make it clear *which* base+var-selector could not be resolved... That doesn't quite help font fallback the way we want. Putting up for review.


behdad:
Actually maybe we should make the VS non-default-ignorable if the not-found is set, such that GSUB must resolve it instead of jumping over it.

behdad:
I think I'm happy with this now.

khaledhosny (approved):

PR #4864, opened 2024-09-22
Barrier

behdad:
Add more barriers, around switches on format and version.

There might be some that need adding that use other names, but this is a good start.

It shrinks the library slightly. Hopefully it's also fixing latent bugs.

I haven't done any performance measurements. Doing now.

### DIFF
--- a/src/hb-aat-layout-bsln-table.hh
+++ b/src/hb-aat-layout-bsln-table.hh
@@ -125,7 +125,7 @@ struct bsln
       return_trace (false);
     hb_barrier ();
 
-    switch (format)
+    switch (hb_barrier (format))
     {
     case 0: return_trace (parts.format0.sanitize (c));
     case 1: return_trace (parts.format1.sanitize (c));
--- a/src/hb-aat-layout-common.hh
+++ b/src/hb-aat-layout-common.hh
@@ -467,7 +467,7 @@ struct Lookup
 {
   const T* get_value (hb_codepoint_t glyph_id, unsigned int num_glyphs) const
   {
-    switch (u.format) {
+    switch (hb_barrier (u.format)) {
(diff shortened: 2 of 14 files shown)


behdad:
> I haven't done any performance measurements. Doing now.

In quick testing, shaping performance doesn't seem affected.

behdad:
Humm. Godbolt suggests that this might not be working as I expected: https://godbolt.org/z/Gnxnsd434
```c++
#include <atomic>

struct T
{
    uint16_t a;
    uint16_t b;
};

int f(const T& t, unsigned l)
{
  switch (l)
  {
  case 0: return 0;
  case 1: return t.b;
  }
  return 0;
}

template <typename T>
T barrier (const T& t)
{
  T tt = t;
  std::atomic_signal_fence (std::memory_order_acquire);
  return t;
}

int g(const T& t, unsigned l)
{
  switch (barrier (l))
  {
  case 0: return 0;
  case 1: return t.b;
  }
  return 0;
}
```

Both functions produce a jump-less code that reads `t.b` unconditionally.

behdad:
Doesn't work, so...

behdad:
This, however works: https://godbolt.org/z/sEE5YdPeT
```c++
#include <atomic>

struct T
{
    uint16_t a;
    uint16_t b;
};

int f(const T& t, unsigned l)
{
  switch (l)
  {
  case 0: return 0;
  case 1: return t.b;
  }
  return 0;
}

bool barrier ()
{
  std::atomic_signal_fence (std::memory_order_acquire);
  return true;
}

int g(const T& t, unsigned l)
{
  switch (l)
  {
  case 0: barrier (); return 0;
  case 1: barrier (); return t.b;
  }
  return 0;
}
```
It's much more work to annotate all relevant places. I'll give it a go and open another PR.

PR #4865, opened 2024-09-22
Add hb_barrier() to switches of unions

behdad:
https://github.com/harfbuzz/harfbuzz/pull/4864#issuecomment-2366923736


khaledhosny (commented):
This is way over my head, so there is not much I can review.

behdad:
> This is way over my head, so there is not much I can review.

Sure. But if you review the godbolt link from the linked issue you can probably figure out what this does.

PR #4870, opened 2024-09-23
[USE] Update the data files

dscorbett:
This uses the data files from
<https://github.com/microsoft/font-tools/tree/1ffb9024b755313e661e8f492d87954856cc96a6/USE> to get the change from https://github.com/microsoft/font-tools/pull/54. It does not include the change from https://github.com/microsoft/font-tools/pull/56 because that requires the USE spec to be updated to handle `Reordering_Killer` first.


behdad:
LGTM.

PR #4882, opened 2024-10-01
[meson] Update wrap files

khaledhosny:
They have not been updated in a while.


khaledhosny:
Glib’s wrap file requires meson 1.2. Lets wait a few more years.

PR #4885, opened 2024-10-06
[COLR] Enable COLRv0 support in get_extents()

khaledhosny:
paint_glyph() supports COLRv0 but this code is never reached because of the early return if version is not 1. This early return seems to be from before COLRv0 extents were supported.


khaledhosny:
I don’t understand this fuzzer failure. It is failing in `version == 1` branch, but shouldn’t this be an already supersized code?

behdad:
This might fix it:
https://github.com/harfbuzz/harfbuzz/commit/4587e08a4614368fe002110b6b1e9805f6434c33

khaledhosny:
Now it is failing in a different place...

khaledhosny:
The last commit seems to have made it happy.

PR #4895, opened 2024-10-10
Coretext font

behdad:


behdad:
While this is definitely incomplete, it already does enough to be useful.

@khaledhosny Any ideas about the backwards-incompatible change that new fonts created using the hb_coretext API will switch to these (incomplete) hooks? I think we should back out that change out, or finish the font funcs. Not sure.

khaledhosny:
> While this is definitely incomplete, it already does enough to be useful.
> 
> @khaledhosny Any ideas about the backwards-incompatible change that new fonts created using the hb_coretext API will switch to these (incomplete) hooks? I think we should back out that change out, or finish the font funcs. Not sure.

I think we better not have this. It is already bad that creating fonts with hb_ft_* API can lead to different shaping output (because, say, FreeType chooses a different cmap subtable or handles symbol fonts differently), so we shouldn’t introduced another such potential inconsistency here.

behdad:
> I think we better not have this.

Agreed.

behdad:
Next issue: looks like CoreText refuses to process `hvgl` table for memory fonts. As such, the new `hvgl` PingFang UI fails to load with HarfBuzz currently (creating CGFont fails). However, if we use the `CTFontManagerCreateFontDescriptorsFromURL()` API to load the font using CoreText, and with the path of the system PingFang UI font, then it works.

This suggests adding the following:
- hb_coretext_face_create_from_file_or_fail()
- hb_ft_face_create_from_file_or_fail() [This would allow reading WOFF files.]
- hb_ot_face_create_from_file_or_fail() [Just for symmetry?]

I don't know about the last one. It can also just be called `hb_face_create_from_file_or_fail()`.

Then we add `--loader` to `hb-view`, `benchmark-font`, etc.

Thoughts?

khaledhosny:
Feels a a bit too specific for this use case, but I don’t think it hurts to provide such API. I’d go with `hb_face_create_from_file_or_fail()` for the last one (and though it would be a thin wrapper aloud `hb_blob_create_from_file_or_fail()`, I think it is OK for completeness).

src/hb-coretext-font.cc
```diff
@@ -0,0 +1,355 @@
...
+ * @font: #hb_font_t to work upon
+ *
+ * Configures the font-functions structure of the specified
+ * #hb_font_t font object to use CoreText font functions.
+ *
+ * In particular, you can use this function to configure an
+ * existing #hb_face_t face object for use with CoreText font
+ * functions even if that #hb_face_t face object was initially
+ * created with hb_face_create(), and therefore was not
+ * initially configured to use CoreText font functions.
```
khaledhosny, line 329: Is this actually true? I don’t see where the `CTFont` is created in this case, since `hb_coretext_font_get_ct_font()` does not create one, but only gets the `CTFont` tht was passed to `hb_coretext_font_create()`.

behdad: I was also confused by it, but I *think* it's created on demand, by this code:
```
CTFontRef 
hb_coretext_font_get_ct_font (hb_font_t *font) 
{ 
  CTFontRef ct_font = (CTFontRef) (const void *) font->data.coretext; 
  return ct_font ? (CTFontRef) ct_font : nullptr; 
} 
```
That `data.coretext` is a `hb_shaper_lazy_loader_t` that will call `_hb_coretext_shaper_font_data_create` on demand. No?

khaledhosny: I see. That makes sense!

behdad:
`--face-loader=coretext` is working as expected, loading PingFang `hvgl` font fine.

`--face-loader=ft` is hitting a double-free of an `FT_Face`. Help investigating is appreciated.

`collections` test is failing. Investigating.

behdad:
> `collections` test is failing. Investigating.

Loading face-index 1 / 2 in DFONT.dfont fails now. I don't know if that font has multiple faces. I made the new API `hb_face_create_for_file_or_fail()` to fail if `face_index` is out of range. So we might just need to adjust the test. Not sure.

behdad:
> > `collections` test is failing. Investigating.
> 
> Loading face-index 1 / 2 in DFONT.dfont fails now. I don't know if that font has multiple faces. I made the new API `hb_face_create_for_file_or_fail()` to fail if `face_index` is out of range. So we might just need to adjust the test. Not sure.

Yeah, just needed to remove the invalid tests.

behdad:
All in. Remains: 1. figuring out crash with `--face-loader=ft`, and finishing the fleshing out the CoreText font funcs more.

src/hb-coretext-font.cc
```diff
@@ -0,0 +1,355 @@
...
+{
+  CTFontRef ct_font = (CTFontRef) font_data;
+
+  CGFloat ct_font_size = CTFontGetSize (ct_font);
+  CGFloat x_mult = (CGFloat) font->x_scale / ct_font_size;
+  CGFloat y_mult = (CGFloat) font->y_scale / ct_font_size;
+
+  CGGlyph glyphs[1] = { glyph };
+  CGRect bounds = ::CTFontGetBoundingRectsForGlyphs(ct_font,
+						    kCTFontDefaultOrientation, glyphs, NULL, 1);
```
khaledhosny, line 160: This causes a compiler warning:
```
../src/hb-coretext-font.cc:160:11: warning: 'kCTFontDefaultOrientation' is deprecated: first deprecated in macOS 10.11 - Deprecated [-Wdeprecated-declarations]
```
https://developer.apple.com/documentation/coretext/ctfontorientation/kctfontdefaultorientation?language=objc Suggests using `kCTFontOrientationDefault`.

If we use the new name someone is going to complain that we broke compilation on some ancient macOS version, so may be something like:
```diff
diff --git a/src/hb-coretext-font.cc b/src/hb-coretext-font.cc
index e06049d44..4e4a7dddf 100644
--- a/src/hb-coretext-font.cc
+++ b/src/hb-coretext-font.cc
@@ -142,6 +142,10 @@ hb_coretext_get_glyph_h_advances (hb_font_t* font, void* font_data,
 }


+#if MAC_OS_X_VERSION_MIN_REQUIRED < 101100
+# define kCTFontOrientationDefault kCTFontDefaultOrientation
+#endif
+
 static hb_bool_t
 hb_coretext_get_glyph_extents (hb_font_t *font,
                               void *font_data,
@@ -157,7 +161,7 @@ hb_coretext_get_glyph_extents (hb_font_t *font,

   CGGlyph glyphs[1] = { glyph };
   CGRect bounds = ::CTFontGetBoundingRectsForGlyphs(ct_font,
-                                                   kCTFontDefaultOrientation, glyphs, NULL, 1);
+                                                   kCTFontOrientationDefault, glyphs, NULL, 1);

   extents->x_bearing = round (bounds.origin.x * x_mult);
   extents->y_bearing = round (bounds.origin.y * y_mult);
```

util/test-hb-subset-parsing.c
```diff
@@ -7,11 +7,8 @@
 static
 hb_face_t* open_font(const char* path)
 {
-  hb_blob_t *blob = hb_blob_create_from_file_or_fail (path);
-  g_assert(blob);
-  hb_face_t* face = hb_face_create(blob, 0);
-  hb_blob_destroy(blob);
-
+  hb_face_t *face = hb_face_create_from_file_or_fail (path, 0);
```
khaledhosny, line 10: I don’t know why this didn’t cause (or show) a memory leak before, but now one bot is failing because the face returned from the second `open_font()` call below is not destroyed: https://app.circleci.com/pipelines/github/harfbuzz/harfbuzz/8892/workflows/42c621fe-708a-4f38-bb8d-71097f0123e2/jobs/224684?invite=true#step-104-71803_110

khaledhosny:
> All in. Remains: 1. figuring out crash with `--face-loader=ft`, and finishing the fleshing out the CoreText font funcs more.

It looks like `FT_Done_FreeType()` is being called before `_hb_ft_face_destroy()`'s `FT_Done_Face()`.

behdad:
> > All in. Remains: 1. figuring out crash with `--face-loader=ft`, and finishing the fleshing out the CoreText font funcs more.
> 
> It looks like `hb_ft_library_lazy_loader_t.destroy()`’s `FT_Done_FreeType()` is being called before `_hb_ft_face_destroy()`’s `FT_Done_Face()`.

That's what it looks like to me as well. I'll see if I can get the ref-counted `FT_Library` to work...

behdad:
> > All in. Remains: 1. figuring out crash with `--face-loader=ft`, and finishing the fleshing out the CoreText font funcs more.
> 
> It looks like `hb_ft_library_lazy_loader_t.destroy()`’s `FT_Done_FreeType()` is being called before `_hb_ft_face_destroy()`’s `FT_Done_Face()`.

I implemented using ref-counted `FT_Library`, but still doing wrong somewhere...

behdad:
> > > All in. Remains: 1. figuring out crash with `--face-loader=ft`, and finishing the fleshing out the CoreText font funcs more.
> > 
> > 
> > It looks like `hb_ft_library_lazy_loader_t.destroy()`’s `FT_Done_FreeType()` is being called before `_hb_ft_face_destroy()`’s `FT_Done_Face()`.
> 
> I implemented using ref-counted `FT_Library`, but still doing wrong somewhere

I don't think we can fix this properly using existing FreeType API implementation.

What's happening *currently* is that:
- First, the `atexit()` calls are processed, which destroy the global `FT_Library`, and that in turn destroys all `FT_Face`'s attached to it. That's just how FT API does it. It's horrible and makes no sense, but it is what we have to work with right now.
- *Then*, the destructor of a global object destroys the `FT_Face`. But the face has already been destroyed by `FT_Done_FreeType` from above.

I thought maybe using a reference-counted `FT_Library` would help this. However, it still doesn't... The idea was to create a refcounted `FT_Library`, which can be done using `FT_New_Library`, and when creating an `FT_Face` from it, +1 the refcount on it, and attach a callback to the `FT_Face` to "destroy" (-1 refcount) the `FT_Library` in question. This way, it should NOT matter which destructor is called first: 1. the one for FT_Library, or 2. the one for FT_Face.

However, it fails in the details. I implemented it, but the time `FT_Face` destructor is called, it calls the `FT_Library` destructor as we chained them together, then `FT_Library` destroys all `FT_Face`'s attached to it, *including* the one we are destructing. Then we return to destructing the `FT_Face` and cause a double-free, because now our refcount has been dropped to zero and we are already destructed, but we don't know.

behdad:
> * 
> 
> I thought maybe using a reference-counted `FT_Library` would help this. However, it still doesn't... The idea was to create a refcounted `FT_Library`, which can be done using `FT_New_Library`, and when creating an `FT_Face` from it, +1 the refcount on it, and attach a callback to the `FT_Face` to "destroy" (-1 refcount) the `FT_Library` in question. This way, it should NOT matter which destructor is called first: 1. the one for FT_Library, or 2. the one for FT_Face.
> 
> However, it fails in the details. I implemented it, but the time `FT_Face` destructor is called, it calls the `FT_Library` destructor as we chained them together, then `FT_Library` destroys all `FT_Face`'s attached to it, _including_ the one we are destructing. Then we return to destructing the `FT_Face` and cause a double-free, because now our refcount has been dropped to zero and we are already destructed, but we don't know.

I don't think my analysis is fully correct. I'm still investigating.

util/test-hb-subset-parsing.c
behdad: Fixed.

behdad:
> I don't think my analysis is fully correct. I'm still investigating.

Fixed the crash. It was my fault.

Now just fleshing out more CoreText font-funcs. Trying.

behdad:
@khaledhosny I think this is ready. There's only get_font_v_extents left. Let me see if I can do that as well.

khaledhosny (approved):
I’d prefer if commits that fix other commits get squashed, but LGTM otherwise.

behdad:
> I’d prefer if commits that fix other commits get squashed, but LGTM otherwise.

I try to do that. Lemme give it another round. Thanks.

PR #4910, opened 2024-10-21
bug fix in hashmap get_with_hash()

qxliu76:


garretrieger (approved):

behdad:
Nice find! How did you arrive at it?

qxliu76:
> Nice find! How did you arrive at it?

I was looking at HB serializer pop_pack() code and found it uses this API for efficiency, so I looked into its implementation to see if we could do the same thing in Rust(which seems not possible in std HashMap).

PR #4925, opened 2024-11-04
Check HB_NO_VAR in hb-cairo.cc

len-deuel:


behdad:
Thanks! I already pushed a fix.

PR #4927, opened 2024-11-05
[ci] Pin Python version to 3.12 on macOS

khaledhosny:
It seems that lxml does not have 3.13 binary wheels (for macOS at least), and building it from source is failing.

### DIFF
--- a/.github/workflows/macos-ci.yml
+++ b/.github/workflows/macos-ci.yml
@@ -37,7 +37,7 @@ jobs:
     - name: Setup Python
       uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
       with:
-        python-version: '3.x'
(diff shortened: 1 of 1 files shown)


khaledhosny:
Here is a typical failure log from before this PR: https://github.com/harfbuzz/harfbuzz/actions/runs/11673872819/job/32505442739#step:6:261

We have lxml==4.9.3 in .ci/requirements.txt, though, which comes from gcovr==5.0 in .ci/requirements.in and apparently we reverted from gcovr 7.2 to gcovr in #4604 (no reason mentioned in the PR 🤦🏾, but *I think* it caused a ci failure).

I think we should try updating gcovr and see what happens.

khaledhosny:
Actually updating gcovr still breaks CI https://github.com/harfbuzz/harfbuzz/pull/4900, a failure I don’t really understand.

PR #4934, opened 2024-11-11
[gsubgpos] Guard integer arithmetic against wrapping below zero

alerque:
When either casting a signed integer to an unsigned one or subtracting a value from an unsigned one, add assertions to verify we will not end up below expected bounds.

This tracks the extra assertions added to Rustybuzz in https://github.com/harfbuzz/rustybuzz/pull/146 subsequent to having found a bogus type casting in the Rust port via https://github.com/harfbuzz/rustybuzz/issues/142.


alerque:
@khaledhosny Is Circle just mad that I opened this PR from my fork instead of the org repo?

khaledhosny:
> @khaledhosny Is Circle just mad that I opened this PR from my fork instead of the org repo?

May be you didn’t give it the required permissions on your personal account?

PR #4955, opened 2024-11-25
[instancer] use hb_parse_double() for parsing axis positions

qxliu76:
Added hb_subset_axis_range_from/to_string()


src/hb-subset-input.cc
```diff
@@ -597,6 +596,144 @@ hb_subset_input_get_axis_range (hb_subset_input_t  *input,
...
+ * @a #hb_subset_input_t object.
+ * @axis_tag: an axis to convert
+ * @buf: (array length=size) (out caller-allocates): output string
+ * @size: the allocated size of @buf
+ *
+ * Converts an axis range into a `NULL`-terminated string in the format
+ * understood by hb_subset_axis_range_from_string(). The client in responsible for
+ * allocating big enough size for @buf, 128 bytes is more than enough.
+ *
+ * Since: FIXME
```
behdad, line 702: Please use: `XSince: REPLACEME`

behdad: What's the return value?!

qxliu76: Done. I'm heading out for dinner so will address review comments tomorrow if any:)

qxliu76:
oh yeah I always missed that part. Done. Thank you!

behdad:
> oh yeah I always missed that part. Done. Thank you!

Still not happy:
```
harfbuzz-undocumented.txt:1:E: 2 undocumented or incomplete symbols
harfbuzz-unused.txt:1:E: 1 unused documentation entries
```

Can you build docs locally and inspect those two files?

behdad:
I don't understand why that test is failing, and only on Mac.

qxliu76:
> I don't understand why that test is failing, and only on Mac.

I tried to add an HB_NO_SETLOCALE wrap

qxliu76:
and that doesn't help:)

src/hb-subset.h
```diff
@@ -203,6 +203,20 @@ hb_subset_input_set_axis_range (hb_subset_input_t  *input,
 				float               axis_max_value,
 				float               axis_def_value);
 
+HB_EXTERN hb_bool_t
+hb_subset_axis_range_from_string (const char *str, int len,
+				  float *axis_min_value,
+				  float *axis_max_value,
+				  float *axis_def_value);
+
+#ifndef HB_NO_SETLOCALE
```
behdad, line 212: You can't use these macros in header files. What's the problem exactly, do you know?

behdad:
> and that doesn't help:)

Don't make the symbol conditional on `HB_NO_SETLOCALE`. Worst case, make the implementation work only for ASCII, but keep the symbol there.

src/hb-subset-input.cc
```diff
@@ -597,6 +596,146 @@ hb_subset_input_get_axis_range (hb_subset_input_t  *input,
...
   return true;
 }
+
+/**
+ * hb_subset_axis_range_from_string:
+ * @str: a string to parse
+ * @len: length of @str, or -1 if str is NULL terminated
+ * @axis_min_value:(out)the axis min value to initialize with the parsed value
+ * @axis_max_value:(out)the axis max value to initialize with the parsed value
+ * @axis_def_value:(out)the axis default value to initialize with the parse
```
khaledhosny, line 606:
```suggestion
 * @axis_min_value: (out): the axis min value to initialize with the parsed value
 * @axis_max_value: (out): the axis max value to initialize with the parsed value
 * @axis_def_value: (out): the axis default value to initialize with the parse
```

src/hb-subset.h
qxliu76: oh yeah, I'll remove this wrapper

qxliu76:
> > and that doesn't help:)
> 
> Don't make the symbol conditional on `HB_NO_SETLOCALE`. Worst case, make the implementation work only for ASCII, but keep the symbol there.

Thanks! It seems working now:)

PR #4963, opened 2024-12-02
Sidestep nontrivial-memaccess warning in Crap()

kesyog:
With `-Wnontrivial-memaccess`, recent versions of clang flag the usage of memcpy in hb-null.hh with pointers to nontrivially-copyable types. Sidestep this warning by casting the problematic pointer to `void*`.

### DIFF
--- a/src/hb-null.hh
+++ b/src/hb-null.hh
@@ -176,7 +176,7 @@ template <typename Type>
 static inline Type& Crap () {
   static_assert (hb_null_size (Type) <= HB_NULL_POOL_SIZE, "Increase HB_NULL_POOL_SIZE.");
   Type *obj = reinterpret_cast<Type *> (_hb_CrapPool);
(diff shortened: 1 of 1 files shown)


ilovepi:
Using type erasure to avoid a compiler warning is rather questionable. Why not just use a normal assignment operation?  Clang is literally telling you that there's UB here.

If the requirement is that you must use memcpy, then the API should probably require that `Type` is trivially copyable. If the type is already trivially copyable, then most compilers will emit efficient code for doing the copy. So if memcpy is optimal they'll use it anyway, and, in cases where it isn't, the correct copy constructor/assignment operation will be used. In either case there won't be UB.

behdad:
HB dances around UB indeed... This codepath only is triggered in OOM and some code bugs really. We can't use assignment since many of the types we assign here *are* non-trivially-copyable in general. But we know for a fact that their `Null(Type)` instance *is* trivially-copyable... I know...

Issue #1920, opened 2019-08-22
Expose OS/2 fsSelection


khaledhosny:
I was going to try porting LibreOffice to the new metrics API, when I remembered that I need access to OS/2 `fsSelection` (or at least the `USE_TYPO_METRICS` bit) to make it actually work.

behdad:
Should we add similar API that pull out various bitfields out of the font using (our own custom) tags?

khaledhosny:
Sounds good, there are a few such fields scattered across the tables.

behdad:
Let's start listing them here.

ebraminio:
Probably my comment will lead to flow of other bitfields we can have here but if this is the only thing we need, we can also have this as a separate simple API in hb-ot-metrics.h or by just comparing metrics value of `HB_OT_METRICS_TAG_HORIZONTAL_ASCENDER` with `_HB_OT_METRICS_TAG_HORIZONTAL_ASCENDER_OS2`

behdad:
I think some clients would also like access to codepage / scripts bitfields as well, even though I personally don't believe in them.

behdad:
@khaledhosny can you enumerate which bits you think should be exposed?

khaledhosny:
* `OS/2.fsSelection & USE_TYPO_METRICS`
* `OS/2.ulUnicodeRange*`
* `OS/2.ulCodePageRange*`

I think this is all I need that is left.

khaledhosny:
* `OS/2.fsType`

khaledhosny:
* `OS/2.fsSelection`/`head.macStyle` (at least the bold bit)

behdad:
Okay looks like this is pretty OpenType-specific API, so it should go in `hb-ot-`something. Let's say `hb-ot-bits.h`. I suggest we return bitfields as they are, ie. all of fsSelection in one call and define macros for the various bits.

`uint32_t hb_ot_bits_get(tag)`.

ebraminio:
If `OS2->fsSelection` is going to be exposed directly given existence of some other bits in `head->macStyle` also for example it shouldn't have this generic naming, and OS/2 is an [optional table](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6.html#:~:text=The%20'OS%2F2'%20table,for%20all%20TrueType%20font%20files.) so the API should somehow reflect that also, this maybe still can be proceeded with using

[...]

PR #4971, opened 2024-12-13
add hb-style tag

mcratt:


behdad:
But there's no such registered axis!

mcratt:
> But there's no such registered axis!

delete hb-ot-var.h #define HB_OT_TAG_VAR_AXIS_BOLD	HB_TAG('b','o','l','d')  ?

behdad:
I think the issue is that in the variable-font model, there is no "bold" bit. There's the weight axis. Se we might need to translate that bit into wght=700 instead. @khaledhosny and @ebraminio would know.

mcratt:
Okay, do you think this change is fine? I need the fsSelection & BOLD field, but I can't find a related API, so I can only add a new tag like ITALIC.

behdad:
That's what I like more input from @khaledhosny and @ebraminio on. I'm not well-versed on the style API.

ebraminio:
The API we currently provide to detect a font is bold currently is

`hb_style_get_value (font, HB_STYLE_TAG_WEIGHT) == 700`

Is it not useful for you case?

If not I guess you have a font that has `BOLD` flag in `OS2->fsSelection` but has an incorrect number in `OS2->usWeightClass`, which font is this, is it some well known proprietary can be found in Windows / macOS so we can test also?

mcratt:
A single WEIGHT value cannot directly determine if a font is bold, and it's difficult to accurately guess a range for this because in some font creations, the BOLD mark may be incorrect. For instance, some fonts with a WEIGHT greater than 700 may not be marked as BOLD, while some fonts with a WEIGHT less than 500 may be marked as BOLD. However, the BOLD flag still has some value in determining boldness. It needs to be considered together with the WEIGHT to make a more accurate judgment. This is based on the code and results I ran.

```python
import os
import freetype
import py2cy._harfbuzz as hb
import time
from pathlib import Path

ROOT_PATH = os.path.dirname(os.path.abspath(__file__))
DEFAULT_FONT_PATH = os.path.join(os.path.dirname(ROOT_PATH), r"fonts")
FONT_DIRS = [DEFAULT_FONT_PATH]
FONTS_TYPE = os.environ.get("FONTS_TYPE", ["ttc", "ttf", "otf"])

if os.environ.get("FONT_DIRS"):
    for dirPath in os.environ.get("FONT_DIRS").split(";"):
        if dirPath.strip() != "" and os.path.exists(dirPath):
            FONT_DIRS.append(dirPath.strip())

def getAllFiles(path):
    Filelist = []
    for home, _, files in os.walk(path):
        for filename in files:
            if Path(filename).suffix.lower()[1:] in FONTS_TYPE:
                # 保证所有系统下\\转变成/
                Filelist.append(Path(home, filename).as_posix())
    return Filelist

def fuse_test(fontdir_files):
    start = time.perf_counter_ns()
    for file_path in fontdir_files:

        with open(file_path, "rb") as f:
            fontBytes = f.read()
        face = freetype.Face.from_bytes(fontBytes)
        font_count = face.num_faces

        for index in range(font_count):
            ITALIC_harfbuzz = False
            WEIGHT_harfbuzz = False
            ITALIC_freetype = False
            WEIGHT_freetype = False

            face = hb.Face(fontBytes, index)
            font = hb.Font(face)
            ITALIC = font.get_style_value(hb.StyleTag.ITALIC)
            WEIGHT = font.get_style_value(hb.StyleTag.WEIGHT)
            if ITALIC != 0:
                ITALIC_harfbuzz = True
            if WEIGHT >= 700:
                WEIGHT_harfbuzz = True

            face = freetype.Face.from_bytes(fontBytes, index)
            style_flags = face.style_flags
            is_bold = bool(style_flags & freetype.FT_STYLE_FLAG_BOLD)
            is_italic = bool(style_flags & freetype.FT_STYLE_FLAG_ITALIC)
            if is_italic:
                ITALIC_freetype = True
            if is_bold:
                WEIGHT_freetype = True

            if ITALIC_harfbuzz != ITALIC_freetype:
                print(f" ITALIC_harfbuzz: {ITALIC_harfbuzz}\n ITALIC_freetype: {ITALIC_freetype}\n ITALIC: {ITALIC}\n file_path: {file_path}\n index: {index}\n")

            if WEIGHT_harfbuzz != WEIGHT_freetype:
                print(f" WEIGHT_harfbuzz: {WEIGHT_harfbuzz}\n WEIGHT_freetype: {WEIGHT_freetype}\n WEIGHT: {WEIGHT}\n file_path: {file_path}\n index: {index}\n")

    print(f"fuse耗时： {(time.perf_counter_ns() - start) / 1000000:.2f}ms")

if __name__ == "__main__":
    # 获取目录中的所有字体文件
    fontdir_files = {}
    for dir_path in FONT_DIRS:
        files = getAllFiles(dir_path)
        for file_path in files:
            fontdir_files[file_path] = os.path.getsize(file_path)
    # harfbuzz_test(fontdir_files)
    # print("\n")
    # freetype_test(fontdir_files)
    fuse_test(fontdir_files)

```

file:  
[result.txt](https://github.com/user-attachments/files/18123972/result.txt)

mcratt:
There are many fonts related to this issue. Would you like me to package a few and send them to you?

ebraminio:
So guess this falls into the category of bitfields API we haven't created yet https://github.com/harfbuzz/harfbuzz/issues/1920 guess this is a good opportunity to do that finally

mcratt:
Yes, this issue has been mentioned in several issues. I think it would be better to directly return fsSelection instead of using fsSelection & BOLD, because the bits of fsSelection are generally fixed, which can reduce a lot of new TAG additions, although there may be differences in the table version.

behdad:
> Yes, this issue has been mentioned in several issues. I think it would be better to directly return fsSelection instead of using fsSelection & BOLD, because the bits of fsSelection are generally fixed, which can reduce a lot of new TAG additions, although there may be differences in the table version.

Yeah that's what I'm suggesting in #1920 now.

mcratt:
This is a good thing. I found that hb-ot-metrics has already exposed many of the OS/2 attribute values. I think the issue can be resolved by simply adding the remaining unexposed OS/2 attribute values through an API, similar to how HB_ADD_MVAR_VAR was created. The same applies to the head table.

ebraminio:
If the case is common enough I think we can still add one more heuristic to the style API on when `OS2->usWeightClass` is 400, most likely the default value, but either of `OS2->fsSelection & BOLD` or `OS2->macStyle & BOLD` is set, we override it with 700. Not sure if is the best idea but if that makes it more compatible with ft's result, guess worths the hassle.

ebraminio:
Something like

```
diff --git a/src/hb-ot-os2-table.hh b/src/hb-ot-os2-table.hh
index c00d22b..b3d4d03 100644
--- a/src/hb-ot-os2-table.hh
+++ b/src/hb-ot-os2-table.hh
@@ -137,6 +137,7 @@ struct OS2
   };
 
   bool        is_italic () const { return fsSelection & ITALIC; }
+  bool          is_bold () const { return fsSelection & BOLD; }
   bool       is_oblique () const { return fsSelection & OBLIQUE; }
   bool use_typo_metrics () const { return fsSelection & USE_TYPO_METRICS; }
 
diff --git a/src/hb-style.cc b/src/hb-style.cc
index fbab091..58dbec5 100644
--- a/src/hb-style.cc
+++ b/src/hb-style.cc
@@ -123,9 +123,11 @@ hb_style_get_value (hb_font_t *font, hb_style_tag_t style_tag)
              face->table.head->is_expanded () ? 125 :
              100);
   case HB_STYLE_TAG_WEIGHT:
-    return face->table.OS2->has_data ()
-          ? face->table.OS2->usWeightClass
-          : (face->table.head->is_bold () ? 700 : 400);
+    if (!face->table.OS2->has_data ())
+      return face->table.head->is_bold () ? 700 : 400;
+    unsigned weight = face->table.OS2->usWeightClass;
+    if (weight != 400) return weight;
+    return face->table.OS2->is_bold () || face->table.head->is_bold () ? 700 : 400;
   default:
     return 0;
   }
```

mcratt:
There are some changes, but they only account for a small portion. I guess it's still a range issue.

result:
```python
count: 20745
USE_OLD_BOLD_harfbuzz != UES_BOLD_freetype: 0
USE_OLD_WEIGHT_harfbuzz != UES_BOLD_freetype: 1127
USE_NEW_WEIGHT_harfbuzz != UES_BOLD_freetype: 1086
```


```python
  case HB_STYLE_TAG_WEIGHT:
//    return face->table.OS2->has_data ()
//	   ? face->table.OS2->usWeightClass
//	   : (face->table.head->is_bold () ? 700 : 400);
	{
	    if (!face->table.OS2->has_data ())
	      return face->table.head->is_bold () ? 700 : 400;
	    unsigned weight = face->table.OS2->usWeightClass;
	    if (weight != 400) return weight;
	    // If default weight is set, still check bold flag in OS2 and head tables
	    return face->table.OS2->is_bold () || face->table.head->is_bold () ? 700 : 400;
	}
```

```python
import os
import freetype
import py2cy._harfbuzz as hb
import _harfbuzz as hb_test
import time
from pathlib import Path

def new_test(fontdir_files):
    count = 0
    USE_OLD_BOLD_harfbuzz_count = 0
    USE_OLD_WEIGHT_harfbuzz_count = 0
    USE_NEW_WEIGHT_harfbuzz_count = 0

    for file_path in fontdir_files:
        with open(file_path, "rb") as f:
            fontBytes = f.read()
        face = freetype.Face.from_bytes(fontBytes)
        font_count = face.num_faces

        for index in range(font_count):
            USE_OLD_BOLD_harfbuzz = False
            USE_OLD_WEIGHT_harfbuzz = False

            UES_BOLD_freetype = False

            USE_NEW_WEIGHT_harfbuzz = False

            # now harfbuzz
            blob = hb.Blob.from_file_path(file_path)
            face = hb.Face(blob, index)
            font = hb.Font(face)
            BOLD = font.get_style_value(hb.StyleTag.BOLD)
            WEIGHT = font.get_style_value(hb.StyleTag.WEIGHT)
            if BOLD:
                USE_OLD_BOLD_harfbuzz = True
            if WEIGHT >= 700:
                USE_OLD_WEIGHT_harfbuzz = True

            # freetype
            face = freetype.Face.from_bytes(fontBytes, index)
            style_flags = face.style_flags
            is_bold = bool(style_flags & freetype.FT_STYLE_FLAG_BOLD)
            if is_bold:
                UES_BOLD_freetype = True

            # test harfbuzz
            blob2 = hb_test.Blob.from_file_path(file_path)
            face2 = hb_test.Face(blob2, index)
            font2 = hb_test.Font(face2)
            WEIGHT2 = font2.get_style_value(hb_test.StyleTag.WEIGHT)
            if WEIGHT2 >= 700:
                USE_NEW_WEIGHT_harfbuzz = True

            if USE_OLD_BOLD_harfbuzz != UES_BOLD_freetype:
                USE_OLD_BOLD_harfbuzz_count = USE_OLD_BOLD_harfbuzz_count + 1

            if USE_OLD_WEIGHT_harfbuzz != UES_BOLD_freetype:
                USE_OLD_WEIGHT_harfbuzz_count = USE_OLD_WEIGHT_harfbuzz_count + 1

            if USE_NEW_WEIGHT_harfbuzz != UES_BOLD_freetype:
                USE_NEW_WEIGHT_harfbuzz_count = USE_NEW_WEIGHT_harfbuzz_count + 1
            count = count + 1

    print(f"count: {count}\n"
          f"USE_OLD_BOLD_harfbuzz != UES_BOLD_freetype: {USE_OLD_BOLD_harfbuzz_count}\n"
          f"USE_OLD_WEIGHT_harfbuzz != UES_BOLD_freetype: {USE_OLD_WEIGHT_harfbuzz_count}\n"
          f"USE_NEW_WEIGHT_harfbuzz != UES_BOLD_freetype: {USE_NEW_WEIGHT_harfbuzz_count}")
```

ebraminio:
What if `if WEIGHT2 >= 500:` is used, looking at your results.txt seems 500 being a better cut but not sure how many false positives it would have

mcratt:
The results are not that good.

```python
count: 20745
USE_OLD_BOLD_harfbuzz != UES_BOLD_freetype: 0
USE_OLD_WEIGHT_harfbuzz != UES_BOLD_freetype: 1127
USE_NEW_WEIGHT_harfbuzz != UES_BOLD_freetype: 1956
```

```pyton
if WEIGHT2 >= 500:
    USE_NEW_WEIGHT_harfbuzz = True
```

ebraminio:
> The results are not that good.

How many of them are true negative of freetype, I think the idea isn't to match freetype on where it's wrong but to match it when it's right and we are wrong.

ebraminio:
From results.txt, https://github.com/harfbuzz/harfbuzz/pull/4971#issuecomment-2540957440

```
 WEIGHT_harfbuzz: True
 WEIGHT_freetype: False
 WEIGHT: 700.0
 file_path: C:/Users/Administrator/Desktop/py/fontInAss/fonts/Arphic（文鼎）/简体/文鼎云黑体_B.ttf
 index: 0

 WEIGHT_harfbuzz: False
 WEIGHT_freetype: True
 WEIGHT: 400.0
 file_path: C:/Users/Administrator/Desktop/py/fontInAss/fonts/Arphic（文鼎）/简体/文鼎古印体.ttf
 index: 0
```

Assuming that _B suffix, seems using weight is even better relying on those bits?

ebraminio:
First of all, thank you for all the interest you have on the project, it is very appreciated and at least I've always learnt things from passionate and great works like yours, and thanks for the investigation you had on the fonts also, it's awesome to see someone had been interested on the API, I've sent you an email to the address you've used for the patch also.

About the specific issue, the way I see it there is no decisive solution here and if we want to expose an API, the naming or at least the documentation should reflect the uncertainty, the current PR adds what should be a boolean result to an API that supposed to offer float results and adds a tag to the enum that supposed to have registered tags like https://learn.microsoft.com/en-us/typography/opentype/spec/dvaraxisreg#registered-axis-tags I understand what you already followed is the italic one which that also can have a number between 0 and 1 https://learn.microsoft.com/en-us/typography/opentype/spec/dvaraxistag_ital via querying STAT or fvar tables, and this API uses a hb_font_t object as the input as hb_font_t variations can affect values of it while the bold flag you are interested in isn't supposed to be changed by variation so it needs an API that takes a hb_face_t object. There are caveats to what I said here however, there is a semi-official variation registry https://github.com/googlefonts/axisregistry which even is ok with boolean values but at the last resort and that is also applied to variations. And the current style API already offers unofficial 'Slnt' tags, which the value is derived from the official 'slnt'; please note that only officially registered axes have all lower case characters and not even the ones are accepted at Google https://fonts.google.com/variablefonts#axis-definitions

Given these I suggest addition of a new API for this, `hb_bool_t hb_style_has_bold_flag(hb_face_t)` that can have a check like `(OS2->fsSelection & BOLD) || (head->macStyle & BOLD)`, that somehow tries to reflect this hb_face_t object has flags that it can be bold but not that definitely is a bold font, please note that I'm not making the final decision here but that's just my own suggestion. Similarly we can later add boolean API for other flags we like to expose like `hb_style_is_underline` which guess the naming can reflect we are more certain about it.

Thanks 😊

mcratt:
I have found another method to obtain the data I wanted, so this TAG might no longer be needed.

behdad:
> Given these I suggest addition of a new API for this, `hb_bool_t hb_style_has_bold_flag(hb_face_t)` that can have a check like `(OS2->fsSelection & BOLD) || (head->macStyle & BOLD)`, that somehow tries to reflect this hb_face_t object has flags that it can be bold but not that definitely is a bold font, please note that I'm not making the final decision here but that's just my own suggestion. Similarly we can later add boolean API for other flags we like to expose like `hb_style_is_underline` (or `hb_style_face_is_underline`?) which guess the naming can reflect we are more certain about it.

I think those are way too specific for HB API. I suggest we expose the raw bits for apps that truly want those.

PR #4975, opened 2024-12-15
[myanmar] Allow all tone marks (including visarga) in any order

dscorbett:
Addresses MicrosoftDocs/typography-issues#1187. All tone marks (including visarga) are now allowed in any order. U+1038 MYANMAR SIGN VISARGA is not a tone mark in Asho but this categorization produces the right behavior. This stops dotted circles from being inserted into many unattested character sequences, but none of them are visually ambiguous, so I think this side effect is actually a good thing.

While making this change I noticed that U+AA7C MYANMAR SIGN TAI LAING TONE-2 and U+AA7D MYANMAR SIGN TAI LAING TONE-5 are categorized as `DB` (like U+1037 MYANMAR SIGN DOT BELOW). This was probably not intentional: it is just the default category based on their Unicode properties. This might not make any difference in real Tai Laing words, but it might be worth double-checking (FYI, @ohbendy).


ohbendy:
Thanks. Tai Laing uses 108D AA7C 1038 1089 and AA7D as tonemarks, which are stored after any vowel. I don't think the orthography allows more than one of those at a time.

The only difficulty is in the reordering a font needs to do to get those that attach to the consonant (AA7C and 108D) back past any post-base glyphs. For example, in the sequences 1000 103B 103D 102D 102F AA7C or 1000 103B 103D 102D 102F 108D the font needs to reorder the final character above/below the consonant. I've often wondered if the shaping engine could do that reordering, but actually having the font maker control the shaping is quite helpful.

Issue #227, opened 2016-02-18
hb-blob.cc:107: Undefined Behaviour ?


dcb314:
[hb-blob.cc:107]: (warning) Invalid test for overflow 'data+length<data'. Condition is always false unless there is overflow, and overflow is Undefined Behaviour.

  data + length < data /* overflows */ ||

behdad:
Good point.  Thanks.  Is unsigned-int overflow undefined as well?  Do you have suggested replacement?  How about:

  (uintptr_t)-1 - data > length

dcb314:
>Is unsigned-int overflow undefined as well? 

No. See 

http://stackoverflow.com/questions/18195715/why-is-unsigned-integer-overflow-defined-behavior-but-signed-integer-overflow-is

>Do you have suggested replacement? 

The test for overflow looks mostly pointless to me. Suggest remove.

behdad:
> The test for overflow looks mostly pointless to me. Suggest remove.

Good point.  The input must be valid.  However, this was NOT the case in one incident, and that's how the code was introduced.  But I believe the 2GB check handles that already. See:

https://github.com/behdad/harfbuzz/commit/7d5e7613ced3dd39d05df83ca7e8952cbecd68f6

PR #4987, opened 2025-01-06
[repacker] Promote repacking API to stable.

garretrieger:
For #227.


src/hb-subset-repacker.cc
```diff
@@ -38,7 +37,7 @@
  * Table specific optimizations (eg. extension promotion in GSUB/GPOS) may be performed.
  * Passing HB_TAG_NONE will disable table specific optimizations.
  *
- * XSince: EXPERIMENTAL
+ * Since: REPLACEME
```
behdad, line 40: `XSince` please. That might be docs fail.

behdad:
@khaledhosny to see if he has ideas about bot fails.

behdad:
lgtm otherwise.

khaledhosny:
The linux ci failure seems to be caused by #4985

PR #4998, opened 2025-01-12
[morx] Blocklist dysfunctional morx table of AALMAGHRIBI.ttf font

### LINKED ISSUE #4108: Blocklist dysfunctional `morx` table

khaledhosny:
Fixes #4108


behdad:
lgtm. @jfkthame ?

PR #5028, opened 2025-01-29
Fix hb_buffer deserialization

drott:
IIUC this line intends to check that the last
character closes the buffer array and != should be ==.

### DIFF
--- a/src/hb-buffer-deserialize-json.hh
+++ b/src/hb-buffer-deserialize-json.hh
@@ -789,7 +789,7 @@ _again:
 
   *end_ptr = p;
 
-  return p == pe && *(p-1) != ']';
+  return p == pe && *(p-1) == ']';
 }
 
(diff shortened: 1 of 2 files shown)


drott:
What do those bits mean.. ignore whitespace until the closing delimiter? Not sure what the text-format deserialization looks like but I assume the '>' is some sort of array-like closing delimiter in the text format?

behdad:
> What do those bits mean.. ignore whitespace until the closing delimiter?

I confess I'm not sure what that code does :)). I'll dig into it and report back.

> Not sure what the text-format deserialization looks like but I assume the '>' is some sort of array-like closing delimiter in the text format?

Yes. The text format for Unicodes is like: `<U+0041|U+0042>` or something like that.

behdad:
Superseded by https://github.com/harfbuzz/harfbuzz/pull/5129

PR #5031, opened 2025-02-02
[aat] Add a class-cache to the machine & fix speed regression from 2023!

behdad:
Speeds up Times.ttc benchmark by 27%.

### DIFF
--- a/src/hb-aat-layout-common.hh
+++ b/src/hb-aat-layout-common.hh
@@ -657,13 +657,14 @@ struct StateTable
 
   const Entry<Extra> &get_entry (int state, unsigned int klass) const
   {
-    if (unlikely (klass >= nClasses))
+    unsigned n_classes = nClasses;
+    if (unlikely (klass >= n_classes))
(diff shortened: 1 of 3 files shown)


behdad:
This removes the previous set-digest and adds a 512-byte cache per subtable instead.

khaledhosny:
I fixed the MSVC CI failures on main, you may want to rebase.

behdad:
Thanks.

behdad:
The lambdafication of some of the complicated condiitonal had added, like, 50% overhead to shaping LucidaGrande. It was a regression from 978918c32a66af41df86386510cf73a4c0e8d393 from Sep 2023.

behdad:
I have no idea what the alpine bot failure is about!

khaledhosny:
> I have no idea what the alpine bot failure is about!

Looks the compiler process was killed, probably it was stuck or taking too much resources. Usually it is a transient issue, and re-running the job fixes it.

Issue #5036, opened 2025-02-03
[coretext] Implement `hb_face_reference_blob`


behdad:
CT doesn't have API for that. BUT, it implements `get_table_tags`. So we can fetch all tables and use the face-builder to throw a face together. This will make eg. ft font-funcs work with a coretext face-loader, albeit with lots of wasted memory...

khaledhosny:
Here is some code I wrote for LibreOffice to do this, in case it helps:
```c++
        auto pTags = CTFontCopyAvailableTables(pFont, kCTFontTableOptionNoOptions);
        CFIndex nTags = pTags ? CFArrayGetCount(pTags) : 0;
        if (nTags > 0)
        {
            hb_face_t* pHbFace = hb_face_builder_create();
            for (CFIndex i = 0; i < nTags; i++)
            {
                auto nTable = reinterpret_cast<intptr_t>(CFArrayGetValueAtIndex(pTags, i));
                assert(nTable);
                auto pTable = GetHbTable(nTable);
                assert(pTable);
                hb_face_builder_add_table(pHbFace, nTable, pTable);
            }
            pBlob = hb_face_reference_blob(pHbFace);

            hb_face_destroy(pHbFace);
        }
        if (pTags)
            CFRelease(pTags);
```

behdad:
Exactly what I had in mind, except for, using `hb_face_get_table_tags()`. Would you be able to do it?

khaledhosny:
I can give it a try.

behdad:
I think we should put the code in `hb_face_reference_blob()` itself, if the `reference_table (HB_TAG_NONE)` returned an empty blob.

khaledhosny:
Should we cache the created blob in the face object, or create it a fresh each time?

behdad:
> Should we cache the created blob in the face object, or create it a fresh each time?

I'd say fresh every time. Let's start that way anyway.

PR #5037, opened 2025-02-03
[face] Add fallback implementation to hb_face_reference_blob

khaledhosny:
If referencing the face blob is not possible (e.g. not implemented by the font functions), use face builder to create a blob out of individual table blobs.

Fixes https://github.com/harfbuzz/harfbuzz/issues/5036

### DIFF
--- a/src/hb-face.cc
+++ b/src/hb-face.cc
@@ -504,7 +504,37 @@ hb_face_reference_table (const hb_face_t *face,
 hb_blob_t *
 hb_face_reference_blob (hb_face_t *face)
 {
-  return face->reference_table (HB_TAG_NONE);
+  hb_blob_t *blob = face->reference_table (HB_TAG_NONE);
+
+  if (blob == hb_blob_get_empty ())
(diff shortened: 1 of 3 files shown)


behdad:
LGTM.

Please update documentation. Also, documentation of `get_table` callback to say it should return empty blob in case of null tag if it doesn't have the face blob; and implement `get_table_tags` when possible...

Thanks.

khaledhosny:
> Also, documentation of `get_table` callback to say it should return empty blob in case of null tag if it doesn't have the face blob

`hb_face_t::reference_table()` already returns an empty blob if the callback function returned NULL, so I don’t think we need to request any special handling for unsupported tags.

behdad:
Good point. Thanks. Looks like `hb_reference_table_func_t` doesn't say that NONE tag is used to fetch the entire face... Fix if you can. Otherwise, LGTM, feel free to merge. Thanks!

khaledhosny:
I just did.

PR #5038, opened 2025-02-04
[kern/kerx] Use hb-set-t for left/right, to guard bsearch & remove aat buffer-digest code

behdad:
Use a set of all left glyphs participating in kerning, and all right glyphs participating in kerning, and use these two to weed out non-kerning pairs before bsearching into the kerns.

Speeds up shaping of HelveticaNeue by ~10%.


drott (approved):
LGTM, confirmed additional improvements with SM3 editor dump.

PR #5039, opened 2025-02-04
[aat] Speed up Lucida Grande among others

behdad:
It has a lot of small chains, many of them for Hebrew, or other scripts. But using the buffer-digest, for all buffers, not only short ones, we can skip such subchains.


drott (approved):
LGTM, appears to roughly give 4,6% improvement on the shape calls from SM3.

PR #5040, opened 2025-02-04
[aat] Speed up subchain skipping using exact sets

behdad:
Another 12% speedup with LucidaGrande.


drott:
This brought an improvement for Lucida on longer texts, but slowed shaping of very short strings, for two letter words (Helvetica Neue) of about 20%.

PR #5041, opened 2025-02-05
[morx] Only collect glyphs that can initiate action in the machine

behdad:
And match them against the buffer glyph_set.
    
4% speedup in LucidaGrande. 12% in Times.


drott (approved):
LGTM, even though in the SM3 mixed workloads we do not see the expected improvement, likely due to the regression on #5040.

PR #5042, opened 2025-02-05
[aat] Fix performance regression

behdad:


drott (approved):
LGTM, tested against SM3 workload, and comes out ~1.3% better than tip of tree at 7d17ea49961b16e7b042bebff13cdeb8a9208dd5

PR #5043, opened 2025-02-05
[OT] PairPos cache

behdad:
10% speedup on Roboto-Regular. 5% on SF Pro.


drott (approved):
Shows improvements in local measurements checked against Arial from SM3 dump.

PR #5045, opened 2025-02-06
[set] Tweak allocation

behdad:
Allocation 2-page sets as exact count as well.
Don't do the same for the page_map.

@garretrieger Any chance you can measure if this has any effect on the subset benchmarks? Thanks.

### DIFF
--- a/src/hb-bit-set.hh
+++ b/src/hb-bit-set.hh
@@ -88,10 +88,11 @@ struct hb_bit_set_t
   {
     if (unlikely (!successful)) return false;
 
-    if (pages.length == 0 && count == 1)
+    if (pages.length < count && count <= 2)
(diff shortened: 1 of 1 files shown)


behdad:
> Allocation 2-page sets as exact count as well. Don't do the same for the page_map.
> 
> @garretrieger Any chance you can measure if this has any effect on the subset benchmarks? Thanks.

And the set benchmarks.

garretrieger:
> > Allocation 2-page sets as exact count as well. Don't do the same for the page_map.
> > @garretrieger Any chance you can measure if this has any effect on the subset benchmarks? Thanks.
> 
> And the set benchmarks.

Sounds good I'll start up some benchmark runs.

behdad:
Thanks. That's good enough for me. :)

PR #5047, opened 2025-02-06
[shape-plan] Add hb_ot_shape_plan_get_feature_tags

khaledhosny:
This should get the features on a shape plan after executing it.

Initially I wanted to return an array of tags, but then there can be user features that are not enabled globally, so I thought returning hb_feature_t with value and range would be better. There is a TODO since I couldn’t figure out how to get the value and range from the feature mask. But also it may be overkill and a simple boolean indicating wither it is a global feature or not would be enough.

I wounder also what should happen to non-user features that are applied selectively, like init or medi, does ot make sense to indicate whether they are global or not?

This is inspired by the discussion in:
https://github.com/fontforge/fontforge/pull/5522#pullrequestreview-2574321449, but it might be useful to other HarfBuzz users.


behdad:
The plan doesn't know the ranges: just whether they are global or not.

behdad:
To go with `hb_ot_shape_plan_collect_lookups`, maybe call it similar to that. I make that function returned unordered lookups, because I didn't want to give the user that knowing the lookups they can just shape doing those....

Similarly, maybe returning just a set / list of feature tags is enough. Or with a bool whether they are global or non-global. init/medi etc would fit the non-global then.

src/hb-ot-map.cc
```diff
@@ -390,5 +390,28 @@ hb_ot_map_builder_t::compile (hb_ot_map_t                  &m,
...
+  {
+    auto sub_features = features.as_array ().sub_array (start_offset, *feature_count);
+    *feature_count = sub_features.length;
+
+    if (out_features)
+    {
+      for (unsigned int i = 0; i < sub_features.length; i++)
+      {
+        out_features[i].tag = sub_features[i].tag;
+        // FIXME: I should be using the mask to get the proper value, start, and
```
behdad, line 405: It's a `>>` and an `&` away from you.

behdad:
Upon further thought, I'm fine with a `get` API like you propose actually.

behdad:
Can you please change the name to `hb_ot_shape_plan...` and move to `hb-ot-shape.h`?

src/hb-ot-map.cc
```diff
@@ -390,5 +390,28 @@ hb_ot_map_builder_t::compile (hb_ot_map_t                  &m,
   }
 }
 
+unsigned int hb_ot_map_t::get_features (unsigned int start_offset, unsigned int *feature_count, hb_feature_t *out_features) const
+{
+  if (feature_count)
+  {
+    auto sub_features = features.as_array ().sub_array (start_offset, *feature_count);
+    *feature_count = sub_features.length;
```
behdad, line 398: If you pass the pointer to `sub_array` instead of dereferencing it, it sets the `feature_count` for you automagically.

src/hb-ot-map.cc
```diff
@@ -390,5 +390,27 @@ hb_ot_map_builder_t::compile (hb_ot_map_t                  &m,
...
+    auto sub_features = features.as_array ().sub_array (start_offset, feature_count);
+
+    if (out_features)
+    {
+      for (unsigned int i = 0; i < sub_features.length; i++)
+      {
+        out_features[i].tag = sub_features[i].tag;
+        // FIXME: I should be using the mask to get the proper value, start, and
+        // end, but I'm not sure how to do that.
+        out_features[i].value = 1;
```
behdad, line 406: Maybe we use value=0 if it's a partial feature, and 1 if it's global? Just check if it uses the global mask or not, for now?

behdad: Or we can set it to *max* possible set value.

behdad: lemme think a bit. A rather unintuitively, but:
- value set to *0* if a global feature,
- value set to max possible value for the shape plan. for binary non-global features this will turn to on, suggesting that the UI can expose a on/off for it. You just use the shift value here.

khaledhosny: Thinking about it more, may be it should just return feature tags. Global vs partial features would be interesting only for user features, but the caller can already track that. I’m leaning to renaming it `hb_ot_shape_plan_get_feature_tags()`. If we need the finer details in the future, we can have another API that returns `hb_feature_t`.

behdad: Well that would possibly give the UI's the wrong impression that eg. they should expose `init` etc to the user. I'm fine with doing what you suggest for now.

khaledhosny: It depends on the intended use of the UI, an application like FontForge or FontGoggles will want to show all features anyway. A typesetting application should not show any required feature (i.e. hide rlig but show liga), but the API here won’t help with that. May be we should have an API to list required feature, but it would be a hard-coded list based on OpenType feature registry, so HarfBuzz might not be the best place for it as HarfBuzz does not itself make a distinction between required and not required default features.

behdad: Right... Okay. Let's proceed then.

khaledhosny: Waiting for @iorsh to try the API and see if it works for his FontForge code.

behdad:
LGTM.

iorsh:
The API works in my code - see commit https://github.com/fontforge/fontforge/pull/5522/commits/f60b3eea7665ccf1383ad9e0e21b02e3195b0129

I noticed that for aldhabi.ttf (attached to https://github.com/fontforge/fontforge/issues/3837) `hb_ot_shape_plan_get_feature_tags()` doesn't return the `"locl"` feature for `HB_SCRIPT_ARABIC`. This feature is defined only for `arab{FAR }`, and I'm not sure if this is the expected behaviour.

khaledhosny:
> I noticed that for aldhabi.ttf (attached to [fontforge/fontforge#3837](https://github.com/fontforge/fontforge/issues/3837)) `hb_ot_shape_plan_get_feature_tags()` doesn't return the `"locl"` feature for `HB_SCRIPT_ARABIC`. This feature is defined only for `arab{FAR }`, and I'm not sure if this is the expected behaviour.

I think it is expected. If the feature is not available for the current language system, then it is not applied. But lets discuss this over HarfBuzz PR so Behdad can chime in.

khaledhosny:
*This is the PR* 🤦🏾 

Behdad, what do you think?

iorsh:
> I think it is expected. If the feature is not available for the current language system, then it is not applied. But lets discuss this over HarfBuzz PR so Behdad can chime in.

FYI, I also set Farsi language in buffer `hb_segment_properties_t`, but it still doesn't return `"locl"`.

khaledhosny:
> > I think it is expected. If the feature is not available for the current language system, then it is not applied. But lets discuss this over HarfBuzz PR so Behdad can chime in.
> 
> FYI, I also set Farsi language in buffer `hb_segment_properties_t`, but it still doesn't return `"locl"`.

That is not what happens to me, if I set language to `fa` I get `locl` feature.

behdad:
> > I noticed that for aldhabi.ttf (attached to [fontforge/fontforge#3837](https://github.com/fontforge/fontforge/issues/3837)) `hb_ot_shape_plan_get_feature_tags()` doesn't return the `"locl"` feature for `HB_SCRIPT_ARABIC`. This feature is defined only for `arab{FAR }`, and I'm not sure if this is the expected behaviour.
> 
> I think it is expected. If the feature is not available for the current language system, then it is not applied. But lets discuss this over HarfBuzz PR so Behdad can chime in.

Yeah if you want all features that are *tried* by the shaper, you need a different implementation. Not sure what the intention here is though. If the feature is to show what caps the font has, I think this impl is adequate, no?

iorsh:
Ok, the bug was my fault. I was trying to feed `"FAR "` into `hb_language_from_string()` instead of `"fa"`. Fixed it, now everything works.

Issue #5049, opened 2025-02-06
[trak] Part of font-funcs, not shaping?


behdad:
CoreText, and hence the coretext font funcs, seem to apply trak table to the h_advance (and v_advance?). We apply during shaping. We should change.

Also, not sure, looks like traking is always added to one side of the glyph. We split it to both sides.

behdad:
> Also, not sure, looks like traking is always added to one side of the glyph. We split it to both sides.

This, I fixed already.

PR #5058, opened 2025-02-09
[ot-font/trak] Move trak application to ot-font instead of ot-shape

behdad:
This is what CoreText does.  Before, trak was double-applied with `--shaper-ot --font-funcs=coretext`.
    
Fixes https://github.com/harfbuzz/harfbuzz/issues/5049


behdad:
Let's land this and see. I doubt it will break anything.

Issue #5061, opened 2025-02-09
[test/fuzzer] Run each test with all font files in one process


behdad:
Much much faster, specially under valgrind, than spawning one process per font.

PR #5062, opened 2025-02-09
[test/fuzzing] Run each fuzzer on all fonts in one process

behdad:
Much much faster, specially under valgrind, than spawning one process per font.

Fixes https://github.com/harfbuzz/harfbuzz/issues/5061


behdad:
I think all these four scripts can further be simplified, or at least use a common module.

Possibly we can just remove the Python scripts and pass all the fonts to the binaries from meson? Not sure if possible.

Finally, I removed the valgrind cruft, as there are meson ways to run under valgrind, as indeed our `fedora-valgrind` bot does.

Scripts converted using ChatGPT!

test/fuzzing/run-draw-fuzzer-tests.py
```diff
@@ -1,66 +1,53 @@
...
+from hb_fuzzer_tools import (
+    run_command,
+    chunkify,
+    find_fuzzer_binary,
+    gather_files
+)
 
+def main():
+    srcdir = os.getenv("srcdir", ".")
+    EXEEXT = os.getenv("EXEEXT", "")
```
khaledhosny, line 14: I think this is a leftover from autotools build. We don’t seem to set this envvar anywhere.

test/fuzzing/run-draw-fuzzer-tests.py
```diff
@@ -1,66 +1,53 @@
...
+    top_builddir = os.getenv("top_builddir", ".")
 
-def cmd (command):
-	# https://stackoverflow.com/a/4408409 as we might have huge output sometimes
-	with tempfile.TemporaryFile () as tempf:
-		p = subprocess.Popen (command, stderr=tempf)
+    # Find the fuzzer binary
+    default_bin = os.path.join(top_builddir, "hb-draw-fuzzer" + EXEEXT)
+    hb_draw_fuzzer = find_fuzzer_binary(default_bin, sys.argv)
```
khaledhosny, line 19: May be it is time to simplify this and drop `default_bin` entirely? We always pass the fuzzer binary to meson and I don’t anyone runs these scripts manually.

behdad: Agreed. I don't know meson. Help :D.

behdad:
I asked chatgpt to refactor the common code into the module. But looks like there's a lot more that can be done. The scripts should be like 5 lines really....

khaledhosny:
Yes, I think most of this was inherited from autotools time. Let me give it a try.

behdad:
> Yes, I think most of this was inherited from autotools time. Let me give it a try.

Thanks. I won't touch it then.

behdad:
Oof! I like this.

khaledhosny:
The script can probably be simplified even further, but I think it is good for now.

test/fuzzing/meson.build
```diff
@@ -34,45 +34,42 @@ foreach file_name : tests
...
-  env: env,
   priority: 1,
   suite: ['fuzzing'],
 )
 
-test('repacker-fuzzer', find_program('run-repacker-fuzzer-tests.py'),
+test('repacker-fuzzer', find_program('run-fuzzer-tests.py'),
   args: [
     hb_repacker_fuzzer_exe,
+    meson.current_source_dir() / 'fonts',
```
behdad, line 61: This one should use files in `graph`, no?

behdad: @garretrieger Are those fonts? Or some custom graph encoding?

khaledhosny: Right, I missed that. The subset fuzzer uses to dirs as well, let me fix that.

Issue #4153, opened 2023-03-07
ninja test now runs google-benchmark tests?


behdad:
With benchmarking enabled, it looks like "ninja test" also runs "google-benchmark" tests. That's weird. A minor thing though.

khaledhosny:
I this a subproject? I think it might be caused by the `yield: true` in:
https://github.com/harfbuzz/harfbuzz/blob/ea17c7a81a743d7e319da0ff4111bcf650d2011b/meson_options.txt#L26-L27

We probably should drop it, I don’t think we are interested in running tests for subprojects we build.

behdad:
Yes, it's a subproject.

khaledhosny:
The yield stuff is not related. The latest version of the google-benchmark wrap file (the subproject stuff) now include tests and provide no option to disable them. To skip them locally, you can run the tests with `meson test --no-suite google-benchmark` (which is awkward), but I don’t know if it can be made the default.

behdad:
It's alright. Just a minor thing. Thanks for looking into it.

eli-schwartz:
> To skip them locally, you can run the tests with `meson test --no-suite google-benchmark` (which is awkward), but I don’t know if it can be made the default.

You could also do `meson test --suite harfbuzz` to avoid tests from *any* subproject.

behdad:
> While at it: is there a way to instruct meson to spawn the `slow` test suite earlier, to better parallelize?

That helped a little. Not much.

eli-schwartz:
> While at it: is there a way to instruct meson to spawn the `slow` test suite earlier, to better parallelize?

https://mesonbuild.com/Reference-manual_functions.html#test_priority

> specifies the priority of a test. Tests with a higher priority are started before tests with a lower priority. The starting order of tests with identical priorities is implementation-defined. The default priority is 0, negative numbers are permitted.

eli-schwartz:
> Running `meson test` takes 14s. Running `meson test --no-suite google-benchmark` takes 11.5s.
> 
> Is there any way to make this default, or still no luck?

https://mesonbuild.com/Reference-manual_functions.html#add_test_setup

you can use:

- exclude_suites: A list of test suites that should be excluded when using this setup. Suites specified in the --suite option to meson test will always run, overriding add_test_setup if necessary. Since 0.57.0
- is_default: Set whether this is the default test setup. If true, the setup will be used whenever meson test is run without the --setup option. Since 0.49.0

[...]

PR #5063, opened 2025-02-09
[test] Exclude google-benchmark tests by default if meson "recent" an…

behdad:
…d...

...glib not built internally.

Second try.

See https://github.com/harfbuzz/harfbuzz/issues/4153#issuecomment-2646347531

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -468,7 +468,7 @@ if meson.version().version_compare('>=0.57.0')
   # Re glib, see https://github.com/harfbuzz/harfbuzz/issues/4153#issuecomment-2646347531
   add_test_setup('default',
                  exclude_suites: ['google-benchmark'],
-                 is_default: glib_dep.type_name() != 'internal')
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -464,6 +464,13 @@ configure_file(output: 'config.h', configuration: conf)
 alias_target('lib', libharfbuzz)
 alias_target('libs', libharfbuzz, libharfbuzz_subset)
 
+if meson.version().version_compare('>=0.57.0')
+  # Re glib, see https://github.com/harfbuzz/harfbuzz/issues/4153#issuecomment-2646347531
+  add_test_setup('default',
+                 exclude_suites: ['google-benchmark'],
+                 is_default: glib_dep.type_name() != 'internal')
```
khaledhosny, line 471: May be:
```suggestion
                 is_default: glib_dep.type_name() != 'internal' and not meson.is_subproject())
```

So that we don’t break when HarfBuzz is used as subproject.

behdad: Nice.

PR #5067, opened 2025-02-11
[ci] Update Ubuntu 20.04 images to 24.04

khaledhosny:
GitHub actions will start failing jobs using Ubuntu 20.04 images, so switch to 24.04. Switch also on CircleCI while at it.


khaledhosny:
I don’t know what to do with the gcovr errors. I’m tempted to drop the whole coverage thing as we don’t seem to do much with it.

Issue #3159, opened 2021-08-22
Weird spacing between glyphs when using Apple’s default UI font.


notcome:
I am trying to use Harfbuzz with CoreText, and noticed that when using Apple’s default UI font, the spacing between glyphs are wrong.

Take a look at the below screenshot. They are drawn on macOS Big Sur, in a Catalyst app:

<img width="54" alt="截屏2021-08-22 下午10 20 38" src="https://user-images.githubusercontent.com/2232964/130358507-820f5747-0ff8-4a70-a9f1-2deb52f1c55c.png">

The first row uses a CTFont created by `CTFontCreateUIFontForLanguage(.application, 16, nil)`.

The second row uses a CTFont created by `CTFontCreateWithName("SFProText-Regular" as CFString, 16, nil)`. This looks up a pre-installed SF Pro Text font that I downloaded from Apple’s website.

The third row is a reference drawing using SwiftUI: `Text("World").font(.system(size: 16))`.

In [this thread](https://developer.apple.com/forums/thread/126548) on Apple Developer, an engineered mentioned that Apple put SF Pro Text and Display together since iOS 13. I feel like something might be wrong because of that, since using SF Pro Text explicitly works without any issue.

Below is the same thing drawn under 32px point size. The first row using `CTFontCreateUIFontForLanguage`, the second using SF Pro Text, the third SF Pro Display, and the final line a SwiftUI reference drawing:

<img width="72" alt="截屏2021-08-22 下午10 29 32" src="https://user-images.githubusercontent.com/2232964/130358765-992083d0-efd4-40a6-8553-2569867e095a.png">

It seems that under 32px, the output of the UI font using Harfbuzz matches that of SF Pro Text, but I believe SF Pro Display design should be used when >= 20px?

notcome:
I built Harfbuzz from the master, 34e0b28. I packaged them into a XCFramework and used Swift to call Harfbuzz directly, without any wrapper.

Below is my shaping and drawing code:

```swift
import CoreGraphics
import CoreText
import CHarfbuzz

func shape(_ text: String, font ctFont: CTFont) -> ([CGGlyph], [CGPoint]) {
    var text = text

    let buffer = hb_buffer_create()
    defer { hb_buffer_destroy(buffer) }
    text.withUTF8 { ptr in
        ptr.withMemoryRebound(to: Int8.self) { ptr in
[...]

PR #5068, opened 2025-02-11
[coretext] Copy variations when creating hb-font

behdad:
https://github.com/harfbuzz/harfbuzz/issues/3159#issuecomment-2652262262

Just some Copilot code. Untested.

### DIFF
--- a/src/hb-coretext-shape.cc
+++ b/src/hb-coretext-shape.cc
@@ -524,24 +524,21 @@ hb_coretext_font_create (CTFontRef ct_font)
   if (variations)
   {
     hb_vector_t<hb_variation_t> vars;
+    hb_vector_t<CFTypeRef> keys;
+    hb_vector_t<CFTypeRef> values;
 
     CFIndex count = CFDictionaryGetCount (variations);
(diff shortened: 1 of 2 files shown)


khaledhosny (approved):
My own code is [here](https://git.libreoffice.org/core/+/2e87c644245751aea9f50faf16c4ca4d57b0c2f1/vcl/quartz/CoreTextFontFace.cxx#47), but this is more concise.

khaledhosny:
This does not actually work, let me debug.

behdad:
Also prealloc the vector.

khaledhosny:
Do we we even need to use a vector here?

behdad:
> Do we we even need to use a vector here?

vector with an alloc is my easy way to do a malloc/free. I prefer it, but up to you. Even for the keys/values *I* would have used a vector.

khaledhosny:
> Even for the keys/values _I_ would have used a vector.

That is a good idea actually.

behdad:
> > Even for the keys/values _I_ would have used a vector.
> 
> That is a good idea actually.

If you want to be "fancy", since you checked allocation, you can access `keys.arrayZ` and `values.arrayZ` for direct access. LGTM. Thanks again.

PR #5074, opened 2025-02-16
[decycler] Implement an efficient graph cycle detector

behdad:
This is an algorithm I came up with, based on the Floyd's
    Tortoise-Hare constant-memory linear-time linked-list cycle-detection
    algorithm. It is linear-time and malloc-free. It *eventually* detects
    cycles, not immediately.

    https://en.wikipedia.org/wiki/Cycle_detection#Floyd's_tortoise_and_hare
    
    The main different with Floyd's algorithm is that this algorithm
    detects cycles when one is traversing down a graph, not just a
    linked list.
    
    Our existing cycle-detection algorithms use a set-of-integers,
    either hb_set_t, or more efficient in this case, hb_map_t. Those
    include at least one malloc, and as such show up on profiles.
    
    Port hb-ot-font COLRv1 to use the decycler instead of previous
    hb_map_t usage for cycle detection.
    
    benchmark-font paint_glyph on NotoColorEmoji-Regular.ttf:
    Before: 8ms; After: 5.5ms. No cycle detection: 5.5ms.
    
    FT COLRv1 API is so slow (174ms) it's not worth porting to this.
    Other graphs (VARC, etc) to be ported.
    
For usage example and documentation, see the diff.

### DIFF
--- a/src/OT/Color/COLR/COLR.hh
+++ b/src/OT/Color/COLR/COLR.hh
@@ -29,6 +29,7 @@
 #define OT_COLOR_COLR_COLR_HH
 
 #include "../../../hb.hh"
+#include "../../../hb-decycler.hh"
 #include "../../../hb-open-type.hh"
 #include "../../../hb-ot-var-common.hh"
 #include "../../../hb-paint.hh"
@@ -71,8 +72,8 @@ public:
   hb_array_t<const BGRAColor> palette;
   hb_color_t foreground;
   ItemVarStoreInstancer &instancer;
-  hb_map_t current_glyphs;
-  hb_map_t current_layers;
+  hb_decycler_t glyphs_decycler;
+  hb_decycler_t layers_decycler;
   int depth_left = HB_MAX_NESTING_LEVEL;
   int edge_count = HB_MAX_GRAPH_EDGE_COUNT;
 
@@ -2576,7 +2577,9 @@ struct COLR
 	                         &(get_delta_set_index_map ()),
 	                         hb_array (font->coords, font->num_coords));
     hb_paint_context_t c (this, funcs, data, font, palette_index, foreground, instancer);
-    c.current_glyphs.add (glyph);
+
+    hb_decycler_node_t node (c.glyphs_decycler);
+    node.visit (glyph);
 
     if (version >= 1)
     {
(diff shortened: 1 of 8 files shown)


khaledhosny (approved):
I don’t claim to understand any of this, but it works and faster, so LGTM.

behdad:
> FT COLRv1 API is so slow (174ms) it's not worth porting to this.

https://gitlab.freedesktop.org/freetype/freetype/-/issues/1318

I ported it anyway.

behdad:
I reduced the stack use for the node. Down from my initial 6 pointers, to 3 now. It's optimal I believe. Proof left to the reader.

behdad:
The running-time claim needs some clarification:
- This is _not_ a linear-time graph cycle detection. But:
- This is a linear-time DFS cycle detection.

That is to say:
- If there are no cycles, the running time is linear in the size of the full DFS traversal,
- If there *are* cycles, the full DFS traversal is infinite. This breaks that to return eventually. The time taken however can be exponential in the original graph's size.

behdad:
I should also note that, in all (at least theoretically) our graph traversals, we still enforce a max stack depth as well as max traversed edges budget. So we won't loop infinitely or risk stack-overflow anyway. But this helps break out quicker if there are simple cycles. Since it's an error to have cycles, this is optimal for the normal case, and saving malloc was measurable. So yes, a nicety to have more than a necessity.

behdad:
Merging but more reviews are welcome.

raphlinus:
Sorry to come in late. I've read through the algorithm description (thanks for adding that!) to try to understand it, and fully approve. This handles the common case very efficiently, and will keep cycles from causing bad things to happen. I'm wondering if this should be a blog post or something, I'm not sure this technique is well known. Pretty much everything I could find on a quick search uses some form of node coloring.

behdad:
> Sorry to come in late. I've read through the algorithm description (thanks for adding that!) to try to understand it, and fully approve. This handles the common case very efficiently, and will keep cycles from causing bad things to happen. I'm wondering if this should be a blog post or something, I'm not sure this technique is well known. Pretty much everything I could find on a quick search uses some form of node coloring.

Thank you for the review. I'll write something up.

PR #5079, opened 2025-02-19
[directwrite] Add hb_directwrite_font_create()/font_get_font()

### LINKED ISSUE #5072: [directwrite] Add `hb_directwrite_font_create` and set variations on it?
I think we should have a `hb_directwrite_font_create()` that takes an `IDWriteFont`, and copy variations from it to the created `hb_font_t`. Similar to `hb_coretext_font_create()`.

khaledhosny:
Uses hb_directwrite_face_create() under the hood, but copies variations from directwrite to hb-font.

* Fixes #5072

### DIFF
--- a/src/hb-directwrite.cc
+++ b/src/hb-directwrite.cc
@@ -29,6 +29,7 @@
 #include "hb-shaper-impl.hh"
 
 #include <dwrite_1.h>
+#include <dwrite_3.h>
 
 #include "hb-directwrite.h"
 
@@ -275,6 +276,8 @@ _hb_directwrite_shaper_font_data_create (hb_font_t *font)
 void
 _hb_directwrite_shaper_font_data_destroy (hb_directwrite_font_data_t *data)
 {
+  if (data != HB_SHAPER_DATA_SUCCEEDED)
+    ((IDWriteFont *) data)->Release();
 }
 
 
@@ -839,7 +842,7 @@ _hb_directwrite_reference_table (hb_face_t *face HB_UNUSED, hb_tag_t tag, void *
 }
 
 static void
-_hb_directwrite_font_release (void *data)
+_hb_directwrite_face_release (void *data)
 {
   if (data)
     ((IDWriteFontFace *) data)->Release ();
@@ -847,7 +850,7 @@ _hb_directwrite_font_release (void *data)
 
 /**
  * hb_directwrite_face_create:
- * @font_face: a DirectWrite IDWriteFontFace object.
+ * @dw_face: a DirectWrite IDWriteFontFace object.
  *
  * Constructs a new face object from the specified DirectWrite IDWriteFontFace.
  *
@@ -856,12 +859,12 @@ _hb_directwrite_font_release (void *data)
  * Since: 2.4.0
  **/
 hb_face_t *
-hb_directwrite_face_create (IDWriteFontFace *font_face)
+hb_directwrite_face_create (IDWriteFontFace *dw_face)
 {
-  if (font_face)
-    font_face->AddRef ();
-  return hb_face_create_for_tables (_hb_directwrite_reference_table, font_face,
-				    _hb_directwrite_font_release);
+  if (dw_face)
+    dw_face->AddRef ();
+  return hb_face_create_for_tables (_hb_directwrite_reference_table, dw_face,
+				    _hb_directwrite_face_release);
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/hb-directwrite.h
```diff
@@ -35,6 +35,12 @@ hb_directwrite_face_create (IDWriteFontFace *font_face);
 HB_EXTERN IDWriteFontFace *
 hb_directwrite_face_get_font_face (hb_face_t *face);
 
+HB_EXTERN hb_font_t *
+hb_directwrite_font_create (IDWriteFont *font_face);
+
+HB_EXTERN IDWriteFont *
+hb_directwrite_font_get_font (hb_font_t *face);
```
behdad, line 42: s/face/font/

src/hb-directwrite.h
```diff
@@ -35,6 +35,12 @@ hb_directwrite_face_create (IDWriteFontFace *font_face);
 HB_EXTERN IDWriteFontFace *
 hb_directwrite_face_get_font_face (hb_face_t *face);
 
+HB_EXTERN hb_font_t *
+hb_directwrite_font_create (IDWriteFont *font_face);
```
behdad, line 39: Maybe rename `font_face` here and in other API to `dw_face`.

behdad: Or `dw_font`.

src/hb-directwrite.cc
```diff
@@ -880,5 +881,82 @@ hb_directwrite_face_get_font_face (hb_face_t *face)
...
+*
+* Return value: DirectWrite IDWriteFont object corresponding to the given input
+*
+* XSince: REPLACEME
+**/
+IDWriteFont *
+hb_directwrite_font_get_font (hb_font_t *font)
+{
+  IDWriteFont *dw_font = (IDWriteFont *) (const void *) font->data.directwrite;
+  return dw_font ? dw_font : nullptr;
```
behdad, line 959: This line seems redundant.

khaledhosny: I think I copied it from the coretext equivalent.

src/hb-directwrite.cc
```diff
@@ -880,5 +883,81 @@ hb_directwrite_face_get_font_face (hb_face_t *face)
...
+	if (likely (vars.alloc_exact (count)))
+	{
+	  for (uint32_t i = 0; i < count; ++i)
+	  {
+	    hb_tag_t tag = values[i].axisTag;
+	    float value = values[i].value;
+	    hb_variation_t var = {tag, value};
+	    vars.push (var);
+	  }
+	  hb_font_set_variations (font, vars.arrayZ, count);
```
behdad, line 932: `vars.length` instead of count please, in case of malloc fail.

khaledhosny:
> Can you please add a `hb_coretext_face_create_from_file_or_fail` as well?

Creating DW fonts from paths seems to be rather convoluted but I can try.

src/hb-directwrite.cc
```diff
@@ -880,5 +883,81 @@ hb_directwrite_face_get_font_face (hb_face_t *face)
...
+	}
+      }
+    }
+    dw_face5->Release ();
+  }
+
+  dw_font->AddRef ();
+  font->data.directwrite.cmpexch (nullptr, (hb_directwrite_font_data_t *) dw_font);
```
behdad, line 940: We should release the font if `cmpexch` failed. Please fix coretext as well.

behdad: Ignore me. font was just created. This can't fail.

behdad:
LGTM with the minor comments. Thanks.

src/hb-directwrite.cc
khaledhosny: This is only called after verifying the allocation, but `vars.length` is cleaner anyway.

src/hb-directwrite.cc
khaledhosny: I think `get_dw_font` is better, `font_get_font` looks odd indeed.

lazka:
This leads to a build error with clang:

```
  ../harfbuzz-10.3.0/src/hb-directwrite.cc:280:6: error: cast from 'hb_directwrite_font_data_t *' to 'IDWriteFont *' increases required alignment from 1 to 8 [-Werror,-Wcast-align]
    280 |     ((IDWriteFont *) data)->Release();
        |      ^~~~~~~~~~~~~~~~~~~~
```

khaledhosny:
> This leads to a build error with clang:
> 
> ```
>   ../harfbuzz-10.3.0/src/hb-directwrite.cc:280:6: error: cast from 'hb_directwrite_font_data_t *' to 'IDWriteFont *' increases required alignment from 1 to 8 [-Werror,-Wcast-align]
>     280 |     ((IDWriteFont *) data)->Release();
>         |      ^~~~~~~~~~~~~~~~~~~~
> ```

Does something like `((IDWriteFont *) (const void *))->Release();` fix it?

lazka:
> Does something like `((IDWriteFont *) (const void *) data)->Release();` fix it?

yes, thanks

Issue #5080, opened 2025-02-20
-Wsuggest-attribute=format


khaledhosny:
I see these warnings building with mingw:
```
[360/707] Compiling C++ object src/libharfbuzz-0.dll.p/hb-buffer-verify.cc.obj
../src/hb-buffer-verify.cc: In function 'void buffer_verify_error(hb_buffer_t*, hb_font_t*, const char*, ...)':
../src/hb-buffer-verify.cc:56:14: warning: function 'void buffer_verify_error(hb_buffer_t*, hb_font_t*, const char*, ...)' might be a candidate for 'gnu_printf' format attribute [-Wsuggest-attribute=format]
   56 |     vfprintf (stderr, fmt, ap);
      |     ~~~~~~~~~^~~~~~~~~~~~~~~~~
[362/707] Compiling C++ object src/libharfbuzz-0.dll.p/hb-buffer.cc.obj
../src/hb-buffer.cc: In member function 'bool hb_buffer_t::message_impl(hb_font_t*, const char*, va_list)':
../src/hb-buffer.cc:2302:13: warning: function 'bool hb_buffer_t::message_impl(hb_font_t*, const char*, va_list)' might be a candidate for 'gnu_print ' format attribute [-Wsuggest-attribute=format]
 2302 |   vsnprintf (buf, sizeof (buf), fmt, ap);
      |   ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Is this something we want to address?

behdad:
We already do `HB_PRINTF_FUNC` on those.  That's designed under `#if defined(__GNUC__) && (__GNUC__ >= 3)`. If you know how to enable it for mingw, go for it.

khaledhosny:
It seems that mingw wants `gnu_ptintf` format.

PR #5081, opened 2025-02-20
Annotate printf functions with gnu_printf format under MingW

khaledhosny:
Fixes https://github.com/harfbuzz/harfbuzz/issues/5080


behdad:
LGTM.

PR #5083, opened 2025-02-20
[directwrite] Rename font_get_font to font_get_dw_font

khaledhosny:
https://github.com/harfbuzz/harfbuzz/pull/5079#pullrequestreview-2627407692


behdad:
Thanks

PR #5090, opened 2025-02-25
meson: refactor freetype dependency logic

parona-source:
Simplify and respect -Dfreetype=disabled again.

Bug: https://bugs.gentoo.org/950274
Fixes: 1ad48fddd08654052da4f8a93609e7f4380d5c42
See-Also: 604fe807078ce41d0ac7742547e90b17c066709f

@eli-schwartz

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -102,44 +102,36 @@ check_funcs = [
 
 m_dep = cpp.find_library('m', required: false)
 
-if meson.version().version_compare('>=0.60.0')
+# Painful hack to handle multiple dependencies but also respect options
+if get_option('freetype').disabled()
+  freetype_dep = dependency('', required: false)
+else
   # Sadly, FreeType's versioning schemes are different between pkg-config and CMake
-  # pkg-config: freetype2, cmake: Freetype
+
+  # Try pkg-config name
   freetype_dep = dependency('freetype2',
                             version: freetype_min_version,
                             method: 'pkg-config',
                             required: false,
                             allow_fallback: false)
   if not freetype_dep.found()
-    freetype_dep = dependency('FreeType',
+    # Try cmake name
+    freetype_dep = dependency('Freetype',
                               version: freetype_min_version_actual,
                               method: 'cmake',
-                              required: get_option('freetype'),
-                              default_options: ['harfbuzz=disabled'],
-                              allow_fallback: true)
-  endif
-else
-  # painful hack to handle multiple dependencies but also respect options
-  freetype_opt = get_option('freetype')
-  # we want to handle enabled manually after fallbacks, but also handle disabled normally
-  if freetype_opt.enabled()
-    freetype_opt = false
-  endif
-  # try pkg-config name
-  freetype_dep = dependency('freetype2', version: freetype_min_version, method: 'pkg-config', required: freetype_opt)
-  # when disabled, leave it not-found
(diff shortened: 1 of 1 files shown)


khaledhosny:
cc @fanc999-1

khaledhosny:
Sorry, but I’m no longer able to follow any of this logic, so whatever you, @fanc999-1 and @eli-schwartz agree on I’ll merge (as long as CI passes).

khaledhosny:
Also, I question the need to support cmake-built freetype, so I’m more than happy to see that go if it makes things any simpler.

eli-schwartz (approved):
> Sorry, but I’m no longer able to follow any of this logic, so whatever you, @fanc999-1 and @eli-schwartz agree on I’ll merge (as long as CI passes).

LGTM with my meson upstream maintainer hat on. The current version before this PR, implemented via commit https://github.com/harfbuzz/harfbuzz/commit/1ad48fddd08654052da4f8a93609e7f4380d5c42, is ***incorrect*** and broken usage of meson. It does not function.

To give a bit of background on what's going on here:

Meson has the notion that a dependency is the abstract concept of requiring the Freetype software. There are multiple different ways to find it, and among those are:
- pkg-config
- cmake
- using subproject wraps

When looking up a project dependency, there are different states:
- enabled feature
- disabled feature
- automatically enabled feature, if the dependency can be found, or not enabled otherwise

All of this is concisely plugged in by meson itself:

```meson
foobar_dep = dependency('foobar', required: get_option('foobar'))
```

and will correctly handle all sorts of edge cases for you. If you CANNOT use this, and according to @fanc999-1 cmake doesn't work with this due to the addition of the minimum freetype version requirement, then in order to get feature parity, you need to implement all of this yourself! Here are the steps:

- check if feature is disabled ---> skip it and log the skip
- check if wrap-mode is forcefallback --> use a subproject wrap
- check if force-fallback-for contains "foobar" ---> use a subproject wrap
- check whether pkg-config can find the dependency
- check whether cmake can find the dependency
- use a subproject wrap
- record the dependency as not found. If required, raise a fatal error

The first item on this list that matches is used. Notice that subproject are handled both at the beginning and at the end.

- The change by @fanc999-1 caused step 1 to be completely skipped, which breaks Gentoo and various other people.
- steps 2 and 3 are painful busywork and complexify things even more than @parona-source has already implemented, and I advised as an initial fix to stick with what is currently implemented.

> Also, I question the need to support cmake-built freetype, so I’m more than happy to see that go if it makes things any simpler.

I don't see the value in cmake-built freetype either. It is the source of the complications. Without it, we could go back to:

```meson
freetype_dep = dependency('freetype2',
                          version: freetype_min_version, 
                          required: get_option('freetype'))
```

I assume that @fanc999-1 finds cmake support very valuable for whatever reason (and cannot use the subproject wrap) but I'm not privy to the "why".

eli-schwartz:
@khaledhosny 

Can you confirm whether it's worth supporting cmake detection of freetype2?

khaledhosny:
> @khaledhosny
> 
> Can you confirm whether it's worth supporting cmake detection of freetype2?

I don’t mind dropping it, but it was added by @fanc999-1, so I’d wait for his input.

khaledhosny:
Lets merge this and leave the question for supporting cmake-built FreeType for later.

PR #5093, opened 2025-02-27
Rename and deprecate `hb_directwrite_face_get_font_face`, `hb_ft_font_get_face`

### LINKED ISSUE #5084: Rename and deprecate `hb_directwrite_face_get_font_face`, `hb_ft_font_get_face`
We have `hb_coretext_face_get_cg_font`, `hb_coretext_font_get_ct_font`, and `hb_directwrite_font_get_dw_font`, so these two should be `hb_directwrite_face_get_dw_font_face` and `hb_ft_font_get_ft_face` for consistency and to make it clear what face/font they are returning.

khaledhosny:
* Fixes #5084

### DIFF
--- a/docs/harfbuzz-sections.txt
+++ b/docs/harfbuzz-sections.txt
@@ -350,6 +350,7 @@ hb_coretext_font_set_funcs
 <SECTION>
 <FILE>hb-directwrite</FILE>
 hb_directwrite_face_create
+hb_directwrite_face_get_dw_font_face
 hb_directwrite_face_get_font_face
 hb_directwrite_font_create
 hb_directwrite_font_get_dw_font
--- a/docs/harfbuzz-sections.txt
+++ b/docs/harfbuzz-sections.txt
@@ -332,6 +332,9 @@ hb_font_funcs_set_glyph_shape_func
 hb_font_get_glyph_v_kerning
 hb_font_get_glyph_v_kerning_func_t
 HB_AAT_LAYOUT_FEATURE_TYPE_CURISVE_CONNECTION
+hb_directwrite_face_get_font_face
+hb_ft_font_get_face
+hb_graphite2_font_get_gr_font
 </SECTION>
 
 <SECTION>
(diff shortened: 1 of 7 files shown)


docs/harfbuzz-sections.txt
```diff
@@ -350,6 +350,7 @@ hb_coretext_font_set_funcs
 <SECTION>
 <FILE>hb-directwrite</FILE>
 hb_directwrite_face_create
+hb_directwrite_face_get_dw_font_face
```
behdad, line 353: Should the old symbols be moved to the deprecated section?

behdad:
LGTM w comment.

docs/harfbuzz-sections.txt
khaledhosny: I noticed that we didn’t do this for the deprecated `hb_graphite2_font_get_gr_font()`, so I thought may be we do that only for core API. Should I move the three of them.

khaledhosny: I moved them in the documentation. Do we want me to move the declerations to `hb-deprecated.h` too?

behdad: Since `hb-deprecated.h` is included by `hb.h`, I don't see why not.

khaledhosny: But `hb-ot.h`, `hb-ft.h`, etc. are not included by `hb.h`, which I think why `hb-ot-deprecated.h` exists (to be included from `hb-ot.h`.

khaledhosny: In that sense, I think including the deprecated `hb_ot_*` APIs in the `hb-deprecated` section in the documentation is not strictly accurate, ad that section’s includes has only `hb.h` which does not provide the `hb_ot_*` APIs.

behdad: > But `hb-ot.h`, `hb-ft.h`, etc. are not included by `hb.h`, which I think why `hb-ot-deprecated.h` exists (to be included from `hb-ot.h`.

Oh right. We don't want to pull in graphite ft etc headers in hb-deprecated.h; leave them then. Thanks.

Issue #5007, opened 2025-01-18
[VARC] Doesn't implement `get_extents()`


behdad:
Probably good enough to extract the outline and compute it, but the code isn't currently there.

behdad:
Or maybe get extents of the component glyphs and transform those...

PR #5099, opened 2025-03-02
[`VARC`] Implement get_extents()

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/5007
    
Untested.


behdad:
Output values look wrong. I'm looking into it.

PR #5111, opened 2025-03-06
Two small Rust tidyups

simoncozens:
- **Banish static lifetimes**
- **"where" not needed for trait impl**


behdad:
Awesome. Thanks!

Issue #5113, opened 2025-03-07
[directwrite] Setting features results in "Shaping failed" message


khaledhosny:
Similar results for `liga` 

In case it might help, I have tried changing `kern` in the middle of the text, e.g.:
```
hb-view --font-file GentiumPlus-Regular.ttf --text "WA WA" --features +kern[0:2],-kern[3:] -O svg --shaper ot -o wa-ot.svg
```
and it 
- works great for `ot` shaper, 
- kerns entire string in `uniscribe`, 
- generates a mailformed or empty SVG with `directwrite` (sometimes with a message that "Shaping failed").

_Originally posted by @bobh0303 in https://github.com/harfbuzz/harfbuzz/discussions/4855#discussioncomment-10572307_

PR #5114, opened 2025-03-07
[directwrite] Fix applying features to directwrite shaper

khaledhosny:
The range_features was ending up with random data because hb_ms_make_feature_ranges() keeps pointers to data allocated in feature_records and range_records, but that data was getting freed before range_features was used. Change the variable scope to avoid freeing the data too early.

Fixes https://github.com/harfbuzz/harfbuzz/issues/5113


behdad:
Oof. Thanks.

khaledhosny:
This looks a little hacky. May be `hb_ms_make_feature_ranges()` should be made more robust, but it is an internal API, so probably it is OK to be a bit fragile.

behdad:
Lemme review the code and see if we can do something better.

behdad:
I think this is just how C++ works... In Rust there would be lifetimes and the compiler would validate it. I wouldn't worry more about it than a comment.

PR #5119, opened 2025-03-07
[AAT] Clean up deleted-glyph handling

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/discussions/5118

Fixed a few missing property settings as well while in there.

### DIFF
--- a/src/hb-aat-layout-common.hh
+++ b/src/hb-aat-layout-common.hh
@@ -29,6 +29,8 @@
 
 #include "hb-aat-layout.hh"
 #include "hb-aat-map.hh"
+#include "hb-ot-layout-common.hh"
+#include "hb-ot-layout-gdef-table.hh"
 #include "hb-open-type.hh"
 #include "hb-cache.hh"
 #include "hb-bit-set.hh"
@@ -48,6 +50,8 @@ struct ankr;
 using hb_aat_class_cache_t = hb_cache_t<15, 8, 7>;
 static_assert (sizeof (hb_aat_class_cache_t) == 256, "");
 
+enum { DELETED_GLYPH = 0xFFFF };
+
 struct hb_aat_apply_context_t :
        hb_dispatch_context_t<hb_aat_apply_context_t, bool, HB_DEBUG_APPLY>
 {
@@ -64,7 +68,8 @@ struct hb_aat_apply_context_t :
   hb_buffer_t *buffer;
   hb_sanitize_context_t sanitizer;
   const ankr *ankr_table;
-  const OT::GDEF *gdef_table;
+  const OT::GDEF &gdef;
+  bool has_glyph_classes;
   const hb_sorted_vector_t<hb_aat_map_t::range_flags_t> *range_flags = nullptr;
   bool using_buffer_glyph_set = false;
   hb_bit_set_t buffer_glyph_set;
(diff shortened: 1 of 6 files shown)


behdad:
Of course Mac bots fail now because we don't delete those glyphs anymore, we just hide them. Not sure how best to proceed. Also, we probably should do this hiding thing in hb-coretext as well.

behdad:
Maybe we can add `--delete-default-ignorables` to those tests.

khaledhosny:
While at it, we need to refresh our macOS fonts tests, many fonts have different checksums now and the tests are skipped. Looking into the failing job, I see: “45 tests passed; 13 failed; 112 skipped”, most tests are skipped, that doesn’t sound good.

behdad:
> While at it, we need to refresh our macOS fonts tests, many fonts have different checksums now and the tests are skipped. Looking into the failing job, I see: “45 tests passed; 13 failed; 112 skipped”, most tests are skipped, that doesn’t sound good.

IIRC we have same test for multiple Mac versions. So, two-third skipping might not be all bad. But sure, let's investigate.

behdad:
An alternative for updating the tests is to keep deleting DELETED_GLYPH but only if it's been marked default-ignorable.

jfkthame:
> I'll wait for Jonathan's review before deciding how to fix the tests.

I'll try to take a look at this tomorrow.

src/hb-aat-layout-common.hh
```diff
@@ -106,15 +111,70 @@ struct hb_aat_apply_context_t :
...
+  void replace_glyph_inplace (unsigned i, hb_codepoint_t glyph)
+  {
+    buffer->info[i].codepoint = glyph;
+    buffer_glyph_set.add (glyph);
+#ifndef HB_NO_OT_LAYOUT
+    if (has_glyph_classes)
+      _hb_glyph_info_set_glyph_props (&buffer->info[i],
+				      gdef.get_glyph_props (glyph));
+#endif
+  }
```
jfkthame, line 170: I notice that `replace_glyph_inplace` does not check whether `glyph` is actually `DELETED_GLYPH`; can we be confident that the subtable types that call this will never attempt to "delete" a glyph by replacing it with 0xffff? In theory it seems like a Contextual or Noncontextual substitution table could do that.

Probably we should either handle that case, or add an assertion that `glyph` here must not be `DELETED_GLYPH`.

jfkthame (approved):
Generally this looks OK; I left one comment inline for consideration, though I'm unsure if it would ever be an issue in practice.

behdad:
> Generally this looks OK; I left one comment inline for consideration, though I'm unsure if it would ever be an issue in practice.

Thanks.

Any comment on whether to delete the glyphs from the buffer or leave them as is, hidden?

behdad:
> > Generally this looks OK; I left one comment inline for consideration, though I'm unsure if it would ever be an issue in practice.
> 
> Thanks.
> 
> Any comment on whether to delete the glyphs from the buffer or leave them as is, hidden?

I'm more comfortable NOT deleting them (as current patch is); but the macos.tests need to be regenerated and I can't possibly do it on all versions of mac...

jfkthame:
> Any comment on whether to delete the glyphs from the buffer or leave them as is, hidden?

No strong view offhand. My first inclination would be to follow whichever Core Text does (which I'm not sure offhand -- I think I recall that back when I first was digging into this stuff, they used to be left in place, but I guess that could have changed in more recent versions).

behdad:
> > Any comment on whether to delete the glyphs from the buffer or leave them as is, hidden?
> 
> No strong view offhand. My first inclination would be to follow whichever Core Text does (which I'm not sure offhand -- I think I recall that back when I first was digging into this stuff, they used to be left in place, but I guess that could have changed in more recent versions).

Coretext deletes them indeed. I'll give it a try.

behdad:
Tests seem to pass now. Waiting for @gredler to test tomorrow.

behdad:
So now we delete them if the value is DELETED_GLYPH and a default-ignorable.

Which still can trip OpenJDK in very weird situations.. but the immediate issue should be resolved by this.

behdad:
Merging this since it's a good cleanup anyway. We'll see if there's any remaining issues and deal with those.

gredler:
Thanks for the quick turnaround! I'm still planning to test tomorrow, but I do have a question: with this fix, will users see a change in behavior when they use `hb_buffer_set_invisible_glyph(buffer, 0xFFFF)` and ask HarfBuzz to layout text containing actual default-ignorable characters, like ZWNJ? (i.e. would the ZWNJ now be deleted instead of being sent back as 0xFFFF?) If not, is there an easy way to add an automated test to verify? (I didn't see any test infra support for setting the invisible glyph, but I only had a quick scan.)

behdad:
> Thanks for the quick turnaround! I'm still planning to test tomorrow, but I do have a question: with this fix, will users see a change in behavior when they use `hb_buffer_set_invisible_glyph(buffer, 0xFFFF)` and ask HarfBuzz to layout text containing actual default-ignorable characters, like ZWNJ? (i.e. would the ZWNJ now be deleted instead of being sent back as 0xFFFF?)

Correct. It will be deleted now. But only if the font has morx. That's the weird corner case I was referring to. To fix it beyond what I've done requires finding a new bit flag in the buffer. I'll take another look.

> If not, is there an easy way to add an automated test to verify? (I didn't see any test infra support for setting the invisible glyph, but I only had a quick scan.)

Our test suite allows passing arbitrary arguments to hb-shape.

behdad:
> Correct. It will be deleted now. But only if the font has morx. That's the weird corner case I was referring to. To fix it beyond what I've done requires finding a new bit flag in the buffer. I'll take another look.

I did a follow-up PR: https://github.com/harfbuzz/harfbuzz/pull/5126

behdad:
> Thanks for the quick turnaround! I'm still planning to test tomorrow, but I do have a question: with this fix, will users see a change in behavior when they use `hb_buffer_set_invisible_glyph(buffer, 0xFFFF)` and ask HarfBuzz to layout text containing actual default-ignorable characters, like ZWNJ? (i.e. would the ZWNJ now be deleted instead of being sent back as 0xFFFF?) If not, is there an easy way to add an automated test to verify? (I didn't see any test infra support for setting the invisible glyph, but I only had a quick scan.)

Looks like that would not have happened, since we convert ZWNJ et al to invisible-glyph *after* deleting the AAT deleted-glyphs. It would have happened if get_nominal_glyph in OpenJDK returned 0xFFFF for a missing glyph that also happened to be a default-ignorable.

Anyway, with my new PR that shouldn't happen either.

PR #5120, opened 2025-03-08
Fontations shrink

behdad:
libharfbuzz.so .text size:

- Without fontations: 1.4mb
- With fontations:
  - before: 4.7mb
  - after: 1.7mb


behdad:
How to run `fmt-check` and `clippy-check` in `meson test`?

PR #5121, opened 2025-03-08
[ci] Port minsize job from Alpine to Ubuntu

khaledhosny:
Alpine is frequently failing with compiler crashing, but we don’t really need Alpine. No other job building with --buildtype=minsize, so keep it but use Ubuntu.


alerque:
Wasn't the point of doing that particular build on Alpine that it defaults to musl?

khaledhosny:
> Wasn't the point of doing that particular build on Alpine that it defaults to musl?

Possibly, but is musl really a concern for us? It does not look like we ever got any issues related to musl and this ci job never caught any.

Issue #5117, opened 2025-03-07
font-funcs and create-face versions with "using" from a string


behdad:
In util, benchmark, tests, etc, we have this requirement to map from a string to font funcs or face loader. I suggest we add these to the library.

I want to add

- `hb_font_set_funcs_using(font, name);`
- `hb_font_list_funcs();`

- `hb_face_create_or_fail_using();` 
- `hb_face_create_from_file_or_fail_using();`
- `hb_face_list_loaders();`

Possibly with env vars (`HB_FONT_FUNCS` and `HB_FACE_LOADER`).

This is all similar to how shaper-selection works. I might add the same for unicode_funcs for completeness as well, I'm not sure.

The defaults will pick up whatever is available. If OT is not available, then FT, then Fontations, then Coretext. The default behavior of HB will not change if OT is compiled in.

khaledhosny:
I think only `hb_face_create_or_fail_using()` from the list above is missing, do we still want it?

behdad:
> I think only `hb_face_create_or_fail_using()` from the list above is missing, do we still want it?

The ft/coretext versions were missing, so I was lazy and didn't. Both are possible though.

behdad:
> > I think only `hb_face_create_or_fail_using()` from the list above is missing, do we still want it?
> 
> The ft/coretext versions were missing, so I was lazy and didn't. Both are possible though.

Done in PR now. Just not tested.

PR #5122, opened 2025-03-08
Add API for querying font-funcs / face-loaders and setting them using strings

behdad:
For now:

+ hb_font_set_funcs_using()
+ hb_font_list_funcs()

Part of https://github.com/harfbuzz/harfbuzz/issues/5117

### DIFF
--- a/src/hb-face.cc
+++ b/src/hb-face.cc
@@ -352,6 +352,10 @@ static struct supported_face_loaders_t {
  * A thin wrapper around the face loader functions registered with HarfBuzz.
  * If @loader_name is `NULL`, the first available loader is used.
  *
+ * For example, the FreeType ("ft") loader might be able to load
+ * .woff and .woff2 files if FreeType is built with those features,
+ * whereas the OpenType ("ot") loader will not.
+ *
  * Return value: (transfer full): The new face object, or `NULL` if
  * the file cannot be read or the loader fails to load the face.
  *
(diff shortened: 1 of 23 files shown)


test/api/test-draw.c
```diff
@@ -1105,7 +1088,7 @@ test_hb_draw_funcs (const char* font_funcs_name)
   {
     hb_face_t *face = hb_test_open_font_file ("fonts/glyphs.ttf");
     hb_font_t *font = hb_font_create (face);
-    set_font_funcs (font, font_funcs_name);
+    hb_font_set_funcs_using (font, font_funcs_name);
```
khaledhosny, line 1091: It probably should assert here (and the two other places below) that `hb_font_set_funcs_using()` returned true.

src/hb-font.cc
```diff
@@ -2292,6 +2305,100 @@ hb_font_set_funcs_data (hb_font_t         *font,
...
+{
+  static_font_funcs_list.free_instance ();
+}
+
+/**
+ * hb_font_list_funcs:
+ *
+ * Retrieves the list of font functions supported by HarfBuzz.
+ *
+ * Return value: (transfer none) (array zero-terminated=1): an array of
```
khaledhosny, line 2392: May be explicitly state it is a null-terminated array, in case it is not obvious from the annotation (admittedly, we are not consistent about it in our documentation).

test/api/test-multithread.c
```diff
@@ -93,8 +93,11 @@ thread_func (void *data)
 }
 
 static void
-test_body (void)
+test_body (const char *backend)
 {
+  if (!hb_font_set_funcs_using (font, backend))
+    return;
```
khaledhosny, line 99: Do we really want to silently ignore the return status? I does not look like it should ever happen.

behdad: It's the coretext backend for example wouldn't load a PingFangUI font not at its default location. But yeah in the tests we can assert.

behdad:
Not sure why clippy is recompiling everything. Those should have been done already.

test/api/test-multithread.c
khaledhosny: I mean setting the font functions should not fail, that using the new face can fail is orthogonal to this.

behdad: But setting font funcs do fail. Not with the fonts the test is using though. So I'll make it an assert here.

behdad:
The new `HB_FONT_FUNCS` setting to ft/fontations shows wrong font size. I'm going to change them to automatically pick up changes in the font...

src/hb-shape.cc
```diff
@@ -93,6 +93,10 @@ void free_static_shaper_list ()
  *
  * Return value: (transfer none) (array zero-terminated=1): an array of
  *    constant strings
+ * Return value: (transfer none) (array zero-terminated=1): a
+ *    `NULL`-terminated array of supported shapers constant string.
+ *    The returned array is owned by HarfBuzz and should not be
+ *    modified or freed.
```
khaledhosny, line 99: There are now two `Return value:`.

khaledhosny (approved):

src/hb-face.cc
```diff
@@ -318,7 +325,134 @@ hb_face_create_from_file_or_fail (const char   *file_name,
...
+};
+
+/**
+ * hb_face_create_from_file_or_fail_using:
+ * @file_name: A font filename
+ * @index: The index of the face within the file
+ * @loader_name: (nullable): The name of the loader to use, or `NULL`
+ *
+ * A thin wrapper around the face loader functions registered with HarfBuzz.
+ * If @loader_name is `NULL`, the first available loader is used.
```
khaledhosny, line 352: `NULL` or empty string, right?

khaledhosny (approved):

PR #5125, opened 2025-03-10
[fontations] Check for font serial change automatically

behdad:
Let's see what tsan thinks... I also have not profiled this.

It also doesn't hold the mutex for the whole use time. It probably should or bad things might happen.


behdad:
What I do here is not entirely threadsafe, but should be Good Enough. Going to merge since tsan happy, and it makes setting `HB_FONT_FUNCS=fontations` actually work.

PR #5129, opened 2025-03-10
Fix buffer deserialize

behdad:
I think I got it right.
    
Alternative to https://github.com/harfbuzz/harfbuzz/pull/5028


drott (approved):
I mean... :) A bit tricky to review.
But are there tests JSON actually gets loaded?  You could try with my benchmark CL and see if it loads something. I guess the conclusion from the bug was that the JSON parsing needs a bit more testing. The JSON grammar change looks similar to what I had in mind so that part is good with me.

behdad:
Thanks. I'll add some tests.

behdad:
I think I got it right this time. :)

PR #5130, opened 2025-03-11
[fontations] Make fontations use hb_malloc et al

behdad:
Going to close. Just leaving here for the future.

It requires making `hb_malloc` etc public. Which would have no purpose for us as is.


behdad:
Actually, what are your thoughts @khaledhosny

behdad:
I think I like to merge this. Made hb_malloc() et al public after all. Internal overhead should be minimal.

PR #5131, opened 2025-03-11
Add hb_ft_face_create_from_blob_or_fail() et al

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/5117

Untested.

New API:
+hb_face_create_or_fail_using()
+hb_ft_face_create_from_blob_or_fail()
+hb_coretext_face_create_from_blob_or_fail()

### DIFF
--- a/src/hb-coretext-shape.cc
+++ b/src/hb-coretext-shape.cc
@@ -161,16 +161,42 @@ release_data (void *info, const void *data, size_t size)
 }
 
 static CGFontRef
+create_cg_font (CFArrayRef ct_font_desc_array, unsigned int index)
+{
+  auto ct_font_desc = (CFArrayGetCount (ct_font_desc_array) > index) ?
+		      (CTFontDescriptorRef) CFArrayGetValueAtIndex (ct_font_desc_array, index) : nullptr;
+  if (unlikely (!ct_font_desc))
+  {
+    CFRelease (ct_font_desc_array);
+    return nullptr;
+  }
+  auto ct_font = ct_font_desc ? CTFontCreateWithFontDescriptor (ct_font_desc, 0, nullptr) : nullptr;
+  CFRelease (ct_font_desc_array);
+  if (unlikely (!ct_font))
+    return nullptr;
+
+  auto cg_font = ct_font ? CTFontCopyGraphicsFont (ct_font, nullptr) : nullptr;
+  CFRelease (ct_font);
+
+  return cg_font;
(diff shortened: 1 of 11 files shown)


behdad:
Would be nice to get some testing for this, even if manual.

behdad:
> Would be nice to get some testing for this, even if manual.

Test added.

src/hb-face.cc
```diff
@@ -79,14 +79,14 @@ hb_face_count (hb_blob_t *blob)
   if (unlikely (!blob))
     return 0;
 
-  /* TODO We shouldn't be sanitizing blob.  Port to run sanitizer and return if not sane. */
-  /* Make API signature const after. */
```
khaledhosny, line 83: Do we still want to make the blob const?

src/hb-coretext-shape.cc
```diff
@@ -160,30 +160,42 @@ release_data (void *info, const void *data, size_t size)
   hb_blob_destroy ((hb_blob_t *) info);
 }
 
+static CGFontRef
+create_cg_font (hb_blob_t *blob, unsigned int index)
+{
+  if (unlikely (index != 0))
+    return nullptr; // How?
```
khaledhosny, line 167: There is [`CTFontManagerCreateFontDescriptorsFromData()`](https://developer.apple.com/documentation/coretext/ctfontmanagercreatefontdescriptorsfromdata(_:)?language=objc), then propably use [`CTFontCopyGraphicsFont()`](https://developer.apple.com/documentation/coretext/ctfontcopygraphicsfont(_:_:)?language=objc) to get `CGFontRef` from the `CTFontDescriptorRef`.

src/hb-face.cc
behdad: Guess not. Looks like the only places we use const in the API for our objects is the `get_user_data` APIs.

behdad: Actually I'm not sure. Even `hb_blob_get_length` should be const then. Let's not touch.

src/hb-coretext-shape.cc
behdad: Ah that one. Yeah. It's slow since it loads all the faces at once :(. I'll see if I can use that.

src/hb-face.cc
behdad: New code still calls functions that don't take const blob. Let's leave it as is.

PR #5132, opened 2025-03-11
[test-cplusplus] Add more headers

behdad:


behdad:
I went down a rabbithole...

behdad:
Two weird things I couldn't figure out:
- dwrite doesn't seem to like to be included with a C compiler. Is that accurate?
- hb-gobject.h includes two other hb-gobject headers that can't be found by the test, so I disabled it there.

Also, the dep detection for dwrite is untested. Is the library called dwrite? How were we linking to it before?

khaledhosny:
> dwrite doesn't seem to like to be included with a C compiler. Is that accurate?

I think dwrite*.h files are C++-only.



> Also, the dep detection for dwrite is untested. Is the library called dwrite? How were we linking to it before?

We load it an runtime and don’t link to it. Last time we tried linking to it, people complained: #3232

behdad:
Good to merge?

behdad:
I don't know if changing the signature is safe. I've now convinced myself that I don't mind doing it.

PR #5134, opened 2025-03-12
[kern/kerx] Use a bit-vector instead of bit-set for left/right filtering

behdad:
Shows ~2% speedup shaping of HelveticaNeue.ttc.

UPDATE: I can't verify the speedup / slowdown. It's just too close. Leaving for someone else to benchmark with a heavier workload.


behdad:
Ignore this. Doesn't show much speedup. In the mean time I merged another approach that did speed up GeezaPro and LucidaGrande, and that approach is slightly incompatible with this one.

Issue #5143, opened 2025-03-14
[directwrite] Add font-funcs


behdad:

behdad:
https://searchfox.org/mozilla-central/source/gfx/thebes/gfxDWriteFonts.cpp

behdad:
https://github.com/harfbuzz/harfbuzz/pull/2183/files

PR #5149, opened 2025-03-16
[directwrite] Add fontfuncs

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/5143 eventually.


behdad:
The main thing not implemented is variation-selectors. I can't find that in the API.

PR #5159, opened 2025-03-17
[ci] Cleanup Windows cross-compilation

khaledhosny:


.ci/build-win.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+set -ex
+
+if [ "$#" -lt 2 ]
+then
+  echo "Usage $0 ARCH CPU"
```
behdad, line 6: Can we just take one of ARCH or CPU and deduce the other? Also would be nice if the Usage says what values it accepts.

behdad: Sounds good. It was trivial to turn strip on last I checked. But you probably need to specify the strip script in the cross files.

khaledhosny: This is done now. I’m getting binaries that are slightly bigger than before, no idea what is going on 🤷🏾

behdad: Do you set `-Dbuildtype=release`? That's the default I think. Donno.

behdad: Oh. Didn't we switch to debugoptimized by default?

khaledhosny: Release build gives even bigger binaries.

behdad: Humm... If the differences are not huge, so be it I'd say.

khaledhosny: From 4.1 MB to 5.1 MB compressed. Not a big deal IMO.

behdad: Agreed.

PR #5165, opened 2025-03-18
Fontations glyph name

behdad:
One font is failing. I assume hb rejects the post table completely
while fontations partially uses it.

Also, for another font, fontations is returning Some("") for a all glyphs in the font. So I had special code to return false in that case as well. Probably something that needs fixing in fontations. @dfrg

### DIFF
--- a/src/fontations/lib.rs
+++ b/src/fontations/lib.rs
@@ -22,7 +22,7 @@ use skrifa::metrics::{BoundingBox, GlyphMetrics};
 use skrifa::outline::pen::OutlinePen;
 use skrifa::outline::DrawSettings;
 use skrifa::OutlineGlyphCollection;
-use skrifa::{GlyphId, MetadataProvider};
+use skrifa::{GlyphId, GlyphNames, MetadataProvider};
 
 struct MyAllocator;
 
@@ -60,6 +60,7 @@ struct FontationsData<'a> {
     char_map: Charmap<'a>,
     outline_glyphs: OutlineGlyphCollection<'a>,
     color_glyphs: ColorGlyphCollection<'a>,
+    glyph_names: GlyphNames<'a>,
 
     // Mutex for the below
     mutex: Mutex<()>,
@@ -87,6 +88,8 @@ impl FontationsData<'_> {
 
         let color_glyphs = font_ref.color_glyphs();
 
+        let glyph_names = font_ref.glyph_names();
+
         let mut data = FontationsData {
             face_blob,
             font,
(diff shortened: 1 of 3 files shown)


khaledhosny (approved):

behdad:
Thanks. I'll wait for a new skrifa release such that tests pass.

behdad:
@dfrg Can I get a review please?

src/fontations/lib.rs
```diff
@@ -60,6 +61,8 @@ struct FontationsData<'a> {
     char_map: Charmap<'a>,
     outline_glyphs: OutlineGlyphCollection<'a>,
     color_glyphs: ColorGlyphCollection<'a>,
+    glyph_names: GlyphNames<'a>,
+    glyph_from_names: AtomicPtr<HashMap<String, hb_codepoint_t>>,
```
dfrg, line 65: The stdlib has a nice synchronization primitive for this called [`OnceLock`](https://doc.rust-lang.org/beta/std/sync/struct.OnceLock.html) so you can replace this type with `OnceLock<HashMap<String, hb_codepoint_t>>` and then the code below becomes much cleaner.

src/fontations/lib.rs
```diff
@@ -768,6 +780,83 @@ extern "C" fn _hb_fontations_paint_glyph(
...
+                unsafe {
+                    let _ = Box::from_raw(map);
+                }
+            }
+        }
+    } else {
+        glyph_from_names
+    };
+
+    let glyph_id = unsafe { &*glyph_from_names }.get(name);
```
dfrg, line 848: with the above change, this entire block becomes
```rust
    let glyph_from_names = data.glyph_from_names.get_or_init(|| {
        data.glyph_names
            .iter()
            .map(|(gid, name)| (name.to_string(), gid.to_u32()))
            .collect()
    });
    let glyph_id = glyph_from_names.get(name);
```

dfrg (approved):
Looks good with a suggestion to use a builtin sync primitive. 

Also, this will return synthesized names. I'm not sure if that's what HB expects for this API. You may want to filter on `GlyphName::is_synthesized()`.

src/fontations/lib.rs
behdad: Oof. Thanks.

dfrg:
And skrifa 0.29.1 is published with some glyph name fixes that hopefully fix the failing test.

src/fontations/lib.rs
behdad: I tried to do the same to the other use of Atomic I had but I gave up. I think this is good enough for now.

PR #5170, opened 2025-03-19
[directwrite] Copy font variations when shaping

### LINKED ISSUE #3066: --variations and --shaper=directwrite

khaledhosny:
Fixes #3066

### DIFF
--- a/src/hb-directwrite-shape.cc
+++ b/src/hb-directwrite-shape.cc
@@ -443,6 +443,44 @@ _hb_directwrite_shape (hb_shape_plan_t    *shape_plan,
 	      hb_language_to_string (buffer->props.language), 20);
 
   /*
+   * Set up variations.
+   */
+  IDWriteFontFace5 *fontFaceVariations = nullptr;
+  if (font->num_coords)
+  {
+    IDWriteFontFace5 *fontFace5;
+    if (SUCCEEDED (fontFace->QueryInterface (__uuidof (IDWriteFontFace5), (void **) &fontFace5)))
+    {
+      IDWriteFontResource *fontResource;
+      if (SUCCEEDED (fontFace5->GetFontResource (&fontResource)))
+      {
+	hb_vector_t<DWRITE_FONT_AXIS_VALUE> axis_values;
+	if (likely (axis_values.resize_exact (font->num_coords)))
+	{
+	  for (unsigned int i = 0; i < font->num_coords; i++)
+	  {
+	    hb_ot_var_axis_info_t info;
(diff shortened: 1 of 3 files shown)


src/hb-directwrite.hh
```diff
@@ -237,7 +237,7 @@ HB_INTERNAL hb_directwrite_global_t *
 get_directwrite_global ();
 
 HB_INTERNAL IDWriteFontFace *
-dw_face_create (hb_blob_t *blob, unsigned index);
+dw_face_create (hb_blob_t *blob, unsigned index, hb_font_t *font);
```
behdad, line 240: This is a confusing function signature. blob&index typically mean a face. Can't you get that out from the font already? Why Both?

khaledhosny: It is called in two places, one has a font, the other only has a blob and index.

khaledhosny:
This is currently broken. Setting variations works, but I screwed `hb_directwrite_font_create()`

src/hb-directwrite.hh
behdad: Sounds like two different functions to me :)). What's up?

I'm still digesting, but I think an IDWriteFontFace maps to hb-font-t indeed.

khaledhosny: Yes. I’m starting to think it was a mistake to make `hb_directwrite_font_create()` take an `IDWriteFont`, it should have been `IDWriteFontFace`.

khaledhosny:
I totally confused myself multiple times. This is for the `directwrite` shaper not the font functions.

behdad (approved):
Thanks. I'm fine landing this for now, but it's not how things *should* be done. Instead:
- `IDWriteFontResource` should be put in `face->data.directwrite`,
-  Put `IDWriteFontFace(1/5)` in `font->data.directwrite`. It needs to be serial-checked and updated automagically as needed.
- Use the same in hb-directwrite-font.

PR #5177, opened 2025-03-21
[test] Allow running test suite under wine

khaledhosny:
Set exe_wrapper in the Windows cross files, which will cause unit tests to be run with wine.

When we call the binary ourselves, e.g. in shape run-tests.py, we need to check for MESON_EXE_WRAPPER env var (which meson sets automatically if exe_wrapper is set) and use it.


behdad:
Thanks. So, it's unhappy about glib tests? I don't want to take much of your time on this, happy to close it. I was just curious.

khaledhosny:
It is trying to call wine during the build to check the executables, not only when running the tests and that is not working for some reason. I think Ubuntu’s wine is 64-bit only, and that is why only the win32 job is failing.

It also made the build slower (from ~7 to ~10 minutes), so lets close for now.

khaledhosny:
I kept the Python change and dropped the cross file change.

behdad:
Do you know if meson is supposed to understand `-Dexe_wrapper=wine` on the cmdline?

khaledhosny:
> Do you know if meson is supposed to understand `-Dexe_wrapper=wine` on the cmdline?

Does not seem to be supported outside the cross file. `meson test` has a `--wrapper` option but it seem to work differently.

PR #5180, opened 2025-03-21
[COLRv1] Push font transform trickery to hb-cairo

behdad:
This simplifies the renderers but pushes the burdon to clients of the paint API. I am ambivalent about this change. All paint clients need to be updated. It's a breaking change.

This reduces one vector of allocations from hb-fontations paint implementation however.

### DIFF
--- a/src/OT/Color/COLR/COLR.hh
+++ b/src/OT/Color/COLR/COLR.hh
@@ -938,13 +938,9 @@ struct PaintGlyph
   void paint_glyph (hb_paint_context_t *c) const
   {
     TRACE_PAINT (this);
-    c->funcs->push_inverse_font_transform (c->data, c->font);
     c->funcs->push_clip_glyph (c->data, gid, c->font);
-    c->funcs->push_font_transform (c->data, c->font);
     c->recurse (this+paint);
-    c->funcs->pop_transform (c->data);
     c->funcs->pop_clip (c->data);
-    c->funcs->pop_transform (c->data);
   }
 
   HBUINT8		format; /* format = 10 */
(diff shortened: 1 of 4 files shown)


behdad:
Alternatively we can add a new paint callback for it, `push_clip_glyph_unscaled` and have `push_clip_glyph` do the current behavior as backup. This is more palatable to me.

behdad:
Yeah, I'll do a new callback.

rsheeter:
> I'll do a new callback

+1

> All paint clients need to be updated. It's a breaking change.

Out of curiosity, what paint clients exist? How big of a deal is this?

behdad:
> Out of curiosity, what paint clients exist? How big of a deal is this?

@khaledhosny would know. I know at least one Open Source project implementing this for Unity. There's also uharfbuzz bindings, that might be in use. I'm not sure about LibreOffice.

At any rate, adding a new callback is easy. I'll do it today.

behdad:
Humm. Looks like even a new callback wouldn't fix this, since we still need to keep a vector state of which callback succeeded, and pop accordingly.

The only way to get rid of the allocation is to change the existing callback. And that would only shift the responsibility to hb-cairo. I'm leaning leaving things as are. Or add a new callback for `pop_clip` that also tells it whether it was a glyph clip or rectangle clip. Then if Skrifa also does that, we can get rid of the allocation across the pipeline.

behdad:
>  Or add a new callback for `pop_clip` that also tells it whether it was a glyph clip or rectangle clip. Then if Skrifa also does that, we can get rid of the allocation across the pipeline.

@dfrg Are you open to this? A `pop_glyph_clip` that simply calls `pop_clip` by default.

behdad:
I wonder if using the Skrifa `fill_glyph` can help here...

behdad:
Closing in favor of much simpler https://github.com/harfbuzz/harfbuzz/pull/5185

PR #5181, opened 2025-03-21
[glyf] Don't access gvar/GVAR until needed

behdad:
This doesn't quite do what I want it to do, since we don't optimize all-coords-are-zero case. I have code in another branch that does that, so I suggest we land this regardless, and will be working when other branch lands.


behdad:
Nevermind. This is not necessary. My other branch already takes care of it. "Loading" the gvar accelerator by itself doesn't touch the gvar data, so it's a minor cost only.

PR #5183, opened 2025-03-21
[COLRv1] Remove specialized transform optimization

behdad:
If the font has them, it's unlikely that they are identity transforms. So, no need to try to optimize out.


behdad:
No perf benefit. And tests are failing. Closing.

PR #5186, opened 2025-03-22
[directwrite] Link to library

behdad:
Instead of loading at runtime. I just want to see how CI deals with this. Otherwise, not depending on dwrite is probably desired.

### DIFF
--- a/src/hb-directwrite.hh
+++ b/src/hb-directwrite.hh
@@ -188,25 +188,8 @@ struct hb_directwrite_global_t
 {
   hb_directwrite_global_t ()
   {
-    dwrite_dll = LoadLibraryW (L"DWrite.dll");
-
-#if defined(__GNUC__) || defined(__clang__)
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wcast-function-type"
-#endif
-
-    t_DWriteCreateFactory p_DWriteCreateFactory = (t_DWriteCreateFactory)
-			    GetProcAddress (dwrite_dll, "DWriteCreateFactory");
-
-#if defined(__GNUC__) || defined(__clang__)
-#pragma GCC diagnostic pop
-#endif
-
-    if (unlikely (!p_DWriteCreateFactory))
-      return;
-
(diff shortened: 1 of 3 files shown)


behdad:
Looks like it works... if we ever need it. Thoughts @khaledhosny ?

khaledhosny:
Fine by me. Some people were running HarfBuzz on Windows XP [and didn’t want to link to dwrite.dll](https://github.com/harfbuzz/harfbuzz/issues/3478). That was merely 3 years ago, but I was not convinced then that we need to care about Windows XP users and I’m still not convinced now.

src/hb-directwrite.hh
```diff
@@ -188,25 +188,8 @@ struct hb_directwrite_global_t
...
-
-#if defined(__GNUC__) || defined(__clang__)
-#pragma GCC diagnostic pop
-#endif
-
-    if (unlikely (!p_DWriteCreateFactory))
-      return;
-
-    HRESULT hr = p_DWriteCreateFactory (DWRITE_FACTORY_TYPE_SHARED, __uuidof (IDWriteFactory),
-					(IUnknown**) &dwriteFactory);
```
khaledhosny, line 209: We do the same dance in `test/api/test-directwrite.cc`

khaledhosny:
I’d like to have this back.

behdad:
> I’d like to have this back.

You have my support.

behdad:
Any idea why `test/api/meson.build` doesn't need to pull the dep in?

khaledhosny:
The tests depend on `libharfbuzz_dep` which in turn depends on `harfbuzz_deps` which it includes all the dependencies. So all deps other than `libharfbuzz_*` here are probably redundant (I removed them and everything builds fine):
https://github.com/harfbuzz/harfbuzz/blob/b114518310d4b8cc5612211711a478487c5940e1/test/api/meson.build#L106-L115

behdad:
> The tests depend on `libharfbuzz_dep` which in turn depends on `harfbuzz_deps` which it includes all the dependencies. So all deps other than `libharfbuzz_*` here are probably redundant (I removed them and everything builds fine):

That's what I would think as well. Okay, let's do that then.

PR #5187, opened 2025-03-22
[test] Respect MESON_EXE_WRAPPER in more test runners

behdad:
It's not working for me though. Meson doesn't seem to set WINEPATH during testing.

### DIFF
--- a/test/fuzzing/run-fuzzer-tests.py
+++ b/test/fuzzing/run-fuzzer-tests.py
@@ -2,8 +2,11 @@
 
 import pathlib
 import subprocess
+import os
 import sys
 import tempfile
+
+EXE_WRAPPER = os.environ.get("MESON_EXE_WRAPPER")
 
 
 def run_command(command):
@@ -11,6 +14,10 @@ def run_command(command):
     Run a command, capturing potentially large output in a temp file.
     Returns (output_string, exit_code).
     """
+    global EXE_WRAPPER
+    if EXE_WRAPPER:
+        command = [EXE_WRAPPER] + command
+
     with tempfile.TemporaryFile() as tempf:
         p = subprocess.Popen(command, stdout=tempf, stderr=tempf)
         p.wait()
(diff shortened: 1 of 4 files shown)


khaledhosny (approved):
It seems to work. Looks like I was mistaken and `WINE_PATH` is only set in `meson devenv` or at least that what appears to be happening on my system now. Running `meson devenv meson test` seems to work, almost all tests pass.

khaledhosny:
```
Ok:                 385
Expected Fail:      0
Fail:               2
Unexpected Pass:    0
Skipped:            83
Timeout:            1
```

behdad:
> It seems to work. Looks like I was mistaken and `WINE_PATH` is only set in `meson devenv` or at least that what appears to be happening on my system now. Running `meson devenv meson test` seems to work, almost all tests pass.

Interesting. Works for me too.

Good to merge then?

behdad:
> Ok:                 385
> Expected Fail:      0
> Fail:               2
> Unexpected Pass:    0
> Skipped:            83
> Timeout:            1

Curious: what's the fails and timeout?

khaledhosny:
> > It seems to work. Looks like I was mistaken and `WINE_PATH` is only set in `meson devenv` or at least that what appears to be happening on my system now. Running `meson devenv meson test` seems to work, almost all tests pass.
> 
> Interesting. Works for me too.
> 
> Good to merge then?

Yes.



> > Ok:                 385
> > Expected Fail:      0
> > Fail:               2
> > Unexpected Pass:    0
> > Skipped:            83
> > Timeout:            1
> 
> Curious: what's the fails and timeout?

One failure is in glib, I guess I this one runs even when I disable the tests:

```
glib:lint+no-valgrind / check-missing-install-tag.py  ERROR  0.06s  exit status 1
```

The other is test-directwrite, dwirite crashes on me with:

```
Unhandled exception: unimplemented function ntdll.dll.NtAlpcConnectPort called in 64-bit code (0x006ffffffa61e8).
```
The timeout is in libpng, so something I need to find how to disable:
```
libpng / pngtest  TIMEOUT  60.02s
```

PR #5188, opened 2025-03-22
Text rendering tests fontations

behdad:


behdad:
CoreText fails two because of the missing proper variation-selector support.
DirectWrite around seven.
Fontations fails only CMAP-4.

I suggest we fix Fontations, then enable these tests for those ot,ft,fontations.

behdad:
> Fontations fails only CMAP-4.

https://github.com/googlefonts/fontations/issues/1431

PR #5191, opened 2025-03-22
Revert "[ci] Don’t set exe_wrapper in the cross files"

khaledhosny:
This reverts commit e3dc86bb0f6ac350fb0bbfa6dcd5e9106d036c1b.


behdad:
Can we just disable the glib tests completely?

.circleci/config.yml
```diff
@@ -132,7 +132,9 @@ jobs:
     executor: win32-executor
     steps:
       - checkout
-      - run: apt update && DEBIAN_FRONTEND=noninteractive apt install -y ninja-build python3 python3-pip python3-venv git g++-mingw-w64-i686 zip
+      - run: dpkg --add-architecture i386
+      - run: apt update
+      - run: DEBIAN_FRONTEND=noninteractive apt install -y ninja-build python3 python3-pip python3-venv git g++-mingw-w64-i686 zip wine wine32
```
behdad, line 137: Do we need both `wine` and `wine32`? Or is `wine` a metapackage installing both archs? Asking for the `README.mingw` file...

khaledhosny:
> Can we just disable the glib tests completely?

We have `-Dglib:tests=disabled`, but it does seem to still allow some tests.

behdad:
> > Can we just disable the glib tests completely?
> 
> We have `-Dglib:tests=disabled`, but it does seem to still allow some tests.

How about that old trick only run the harfbuzz test suite? Or is this failing even before running tests?

khaledhosny:
> > > Can we just disable the glib tests completely?
> > 
> > 
> > We have `-Dglib:tests=disabled`, but it does seem to still allow some tests.
> 
> How about that old trick only run the harfbuzz test suite? Or is this failing even before running tests?

It was failing while building the tests, but it is fixed now. We didn’t try running the tests yet.

behdad:
LGTM. Maybe you can enable  tests in a different PR, and I'll see what I can do about the failing directwrite shape test.

khaledhosny:
> LGTM. Maybe you can enable tests in a different PR, and I'll see what I can do about the failing directwrite shape test.

I ran them in the last commit, lets see what happens.

.circleci/config.yml
khaledhosny: I don’t know to be honest. I copied it from somewhere. It does not look like `wine32` or `wine64` packages depend on `wine` package.

behdad:
I think you need:
          meson devenv -Cbuild meson test --print-errorlogs "harfbuzz:"
instead of:
          meson devenv meson test -Cbuild --print-errorlogs "harfbuzz:"

khaledhosny:
Down to two failing tests:
```
Summary of Failures:

114/464 harfbuzz:shape+in-house / directwrite                        FAIL              5.32s   exit status 1
461/464 harfbuzz:api / test-directwrite                              FAIL              0.41s   exit status 3

Ok:                 380 
Expected Fail:      0   
Fail:               2   
Unexpected Pass:    0   
Skipped:            82  
Timeout:            0   
```

behdad:
Awesome. Any idea about the API one?

khaledhosny:
> Awesome. Any idea about the API one?

It is failing on get_dw_font, which we know is broken, so I’m waiting on #5192 for this one.

khaledhosny:
> > Awesome. Any idea about the API one?
> 
> It is failing on get_dw_font, which we know is broken, so I’m waiting on #5192 for this one.

Actually this one is expects to fail. It tries to load `Bahnschrift` font (because it is a variable font that comes with Windows), but of course it is missing on Ubuntu where the cross-job is run. I guess we should skip that test if we detect wine.

behdad:
> > > Awesome. Any idea about the API one?
> > 
> > 
> > It is failing on get_dw_font, which we know is broken, so I’m waiting on #5192 for this one.
> 
> Actually this one is expects to fail. It tries to load `Bahnschrift` font (because it is a variable font that comes with Windows), but of course it is missing on Ubuntu where the cross-job is run. I guess we should skip that test if we detect wine.

Okay, let's detect WINEPATH and skip?

behdad:
> > > Awesome. Any idea about the API one?
> > 
> > 
> > It is failing on get_dw_font, which we know is broken, so I’m waiting on #5192 for this one.
> 
> Actually this one is expects to fail. It tries to load `Bahnschrift` font (because it is a variable font that comes with Windows), but of course it is missing on Ubuntu where the cross-job is run. I guess we should skip that test if we detect wine.

I wonder why it passes for me locally.

khaledhosny:
> > > > Awesome. Any idea about the API one?
> > > 
> > > 
> > > It is failing on get_dw_font, which we know is broken, so I’m waiting on #5192 for this one.
> > 
> > 
> > Actually this one is expects to fail. It tries to load `Bahnschrift` font (because it is a variable font that comes with Windows), but of course it is missing on Ubuntu where the cross-job is run. I guess we should skip that test if we detect wine.
> 
> I wonder why it passes for me locally.

You have the font? It does not pass for me locally.

khaledhosny:
> > > > Awesome. Any idea about the API one?
> > > 
> > > 
> > > It is failing on get_dw_font, which we know is broken, so I’m waiting on #5192 for this one.
> > 
> > 
> > Actually this one is expects to fail. It tries to load `Bahnschrift` font (because it is a variable font that comes with Windows), but of course it is missing on Ubuntu where the cross-job is run. I guess we should skip that test if we detect wine.
> 
> Okay, let's detect WINEPATH and skip?

I’m checking for wine in MESON_EXE_WRAPPER instead. Now I remember you have `binfmt_misc` and don’t set `exe_wrapper`.

behdad:
> > > g on Ubuntu where the cross-job is run. I guess we should skip that test if we detect wine.
> > 
> > 
> > Okay, let's detect WINEPATH and skip?
> 
> I’m checking for wine in MESON_EXE_WRAPPER instead. Now I remember you have `binfmt_misc` and don’t set `exe_wrapper`.

I've switched to `exe_wrapper` and `meson devenv` and had it working until it all stopped.

khaledhosny:
> Anyway, I suggest you use hb API to open something instead of relying on system font collection.

The idea was to load some variable font and check that we are copying variation from it, so it would require DW API to be setting variations on it. I think it can be re-write it to load a DW face, set some variations on it, then create HB font from it and check that we copied the variations.

Issue #5190, opened 2025-03-22
[directwrite] Change `hb_directwrite_font_create`?


behdad:
Currently it takes `IDWriteFont*`. As we found out, we really want `IDWriteFontFace*`. Is the fact that directwrite has been marked experimental, enough to allow changing this API?

The respective `hb_directwrite_font_get_dw_font` I suggest we deprecate and always return nullptr.

PR #5192, opened 2025-03-22
[directwrite] Clean up API / font handling and make variations work on font-funcs

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/5190

### DIFF
--- a/src/hb-directwrite-shape.cc
+++ b/src/hb-directwrite-shape.cc
@@ -65,14 +65,49 @@ struct hb_directwrite_font_data_t {};
 hb_directwrite_font_data_t *
 _hb_directwrite_shaper_font_data_create (hb_font_t *font)
 {
-  return (hb_directwrite_font_data_t *) HB_SHAPER_DATA_SUCCEEDED;
+  IDWriteFontFace *fontFace = (IDWriteFontFace *) (const void *) font->face->data.directwrite;
+
+  /*
+   * Set up variations.
+   */
+  IDWriteFontFace5 *fontFaceVariations = nullptr;
+  {
+    IDWriteFontFace5 *fontFace5;
+    if (SUCCEEDED (fontFace->QueryInterface (__uuidof (IDWriteFontFace5), (void **) &fontFace5)))
+    {
+      IDWriteFontResource *fontResource;
+      if (SUCCEEDED (fontFace5->GetFontResource (&fontResource)))
+      {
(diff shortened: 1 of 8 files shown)


behdad:
I don't know why test-directwrite is crashing...

src/hb-directwrite-shape.cc
```diff
@@ -65,14 +65,49 @@ struct hb_directwrite_font_data_t {};
...
+	  fontResource->CreateFontFace (DWRITE_FONT_SIMULATIONS::DWRITE_FONT_SIMULATIONS_NONE,
+					axis_values.arrayZ, axis_values.length, &fontFaceVariations);
+	}
+	fontResource->Release ();
+      }
+      fontFace5->Release ();
+    }
+  }
+
+  return (hb_directwrite_font_data_t *) fontFaceVariations;
```
khaledhosny, line 104: You are always returning `fontFaceVariations`, but it can be NULL.

src/hb-directwrite-shape.cc
```diff
@@ -65,14 +65,49 @@ struct hb_directwrite_font_data_t {};
 hb_directwrite_font_data_t *
 _hb_directwrite_shaper_font_data_create (hb_font_t *font)
 {
-  return (hb_directwrite_font_data_t *) HB_SHAPER_DATA_SUCCEEDED;
+  IDWriteFontFace *fontFace = (IDWriteFontFace *) (const void *) font->face->data.directwrite;
+
+  /*
+   * Set up variations.
+   */
+  IDWriteFontFace5 *fontFaceVariations = nullptr;
```
khaledhosny, line 73: Did you mean to instead do:
```suggestion
  IDWriteFontFace5 *fontFaceVariations = fontFace;
```

src/hb-directwrite-shape.cc
behdad: Only if creating failed. I removed the conditional on `font->num_coords`.

src/hb-directwrite-shape.cc
behdad: Not really. I'm always creating it. And if creation failed, I want to return nullptr.

src/hb-directwrite-shape.cc
khaledhosny: I see. No idea what is happening then.

behdad:
No idea why tests fail.

behdad:
I'm not sure why the test is failing.

PR #5203, opened 2025-03-24
[fontations] Handle recursive clip-glyphs

behdad:
This reverts 852f66a4187f59f169d0de2ea98725d1d78e8da8 in spirit, using a malloc-free implementation and keeping the fill-glyph optimization.

### DIFF
--- a/src/fontations/lib.rs
+++ b/src/fontations/lib.rs
@@ -464,6 +464,8 @@ struct HbColorPainter<'a> {
     paint_data: *mut c_void,
     color_records: &'a [ColorRecord],
     foreground: hb_color_t,
+    is_glyph_clip: u64,
+    clip_depth: u32,
 }
 
 impl HbColorPainter<'_> {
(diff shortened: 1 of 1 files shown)


src/fontations/lib.rs
```diff
@@ -612,20 +617,35 @@ impl ColorPainter for HbColorPainter<'_> {
...
                 self.paint_data,
                 glyph_id.to_u32() as hb_codepoint_t,
                 self.font,
             );
+            hb_paint_push_font_transform(self.paint_funcs, self.paint_data, self.font);
         }
     }
     fn push_clip_box(&mut self, bbox: BoundingBox) {
+        if self.clip_depth < 64 {
+            self.clip_depth += 1;
```
dfrg, line 645: I think you do need to clear the bit here since it might have been set at this depth by an earlier subgraph.

dfrg (approved):
LGTM with one nit.

src/fontations/lib.rs
behdad: Oh right.

PR #5214, opened 2025-03-27
Util env

behdad:


khaledhosny:
This fixes the crash indeed. Please drop the first two commits.

behdad:
> This fixes the crash indeed. Please drop the first two commits.

The first one I pushed to main to fix breakage. So, this is cleanest now.

khaledhosny (approved):

PR #5215, opened 2025-03-27
[util] Simplify envvar reporting

behdad:


behdad:
I'll report exit status and I think that's it for help2man readiness.

khaledhosny (approved):

Issue #5208, opened 2025-03-25
[util] Improve documentation / help


behdad:
It was brought to my attention by Karl Berry, that, our utils:

- Don't have man pages. I don't know what's the best way to address this.

- Don't specify where to report bugs.

behdad:
From Karl:

My suggestion is to use help2man. That way you only have to maintain the
--help messages instead of redundantly writing the same info in man
format.

The results would be best if you add a sentence or two to the --help
messages which say what the program does (could be stolen from
https://harfbuzz.github.io/utilities.html), and add some kind of repo
reference at the end, like:

Bug reports or other issues: https://github.com/harfbuzz/harfbuzz/

behdad:
@khaledhosny is this something you can look into?

behdad:
This gives pretty descent output:
```
help2man --help-option=--help-all build/util/hb-view | man /dev/stdin
```

behdad:
Would be nice to add a section to the help for env-vars, and exit return values.

behdad:
Looks like we currently don't list available shapers under `--shaper`. We should. We do for `--font-funcs` and `--face-loader`.

behdad:
Maybe remove the available shapers from `--version` output. `help2man` is picking it up and putting it at the end.

khaledhosny:
> Maybe remove the available shapers from `--version` output. `help2man` is picking it up and putting it at the end.

Makes sense.

PR #5216, opened 2025-03-27
[util] Towards building manpages

behdad:
TODO:
1. Only when docs are enabled?
2. Installdir
3. How to include in release?

Fixes https://github.com/harfbuzz/harfbuzz/issues/5208

### DIFF
--- a/util/meson.build
+++ b/util/meson.build
@@ -23,6 +23,8 @@ else
   help2man = disabler()
 endif
 help2man_command = [help2man, '--help-option=--help-all', '--no-info', '--output=' + '@OUTPUT@', '@INPUT@']
+help2man_env = environment()
+help2man_env.set('HB_UTIL_HELP_VERBOSE', '1')
 
 util_deps = [freetype_dep, cairo_dep, cairo_ft_dep, glib_dep]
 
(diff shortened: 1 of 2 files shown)


Eli-Zaretskii:
A comment from the peanut gallery: if you intend to produce man pages from the `--help` output, then please make `--help` be more detailed. For example, `hb-view --help-features` doesn't provide the list of features that could be used, not even the list of the important ones.  If the full list is too long, then even a pointer (like URL) to where it could be found should be a stop-gap. Bonus points for mentioning a few important ones that users of `hb-view` might want to consider.

IOW, please audit the `--help` output of the utilities to verify it includes the relevant information. It is okay to have `--help` withhold something if the man page has everything, but if they are one and the same, then producing a man page from incomplete usage information is most probably not what Karl had in mind when he asked for this.

Thanks for developing HarfBuzz.

khaledhosny (approved):

behdad:
So, these make it into release tarballs fine?

khaledhosny:
> So, these make it into release tarballs fine?

No, they are always generated at build time, so help2man is now a build requirement if one wants the man pages. The alternative is to generate them at source dir and keep them under version control (meson does not add anything to release tarball that is not under version control).

behdad:
> > So, these make it into release tarballs fine?
> 
> No, they are always generated at build time, so help2man is now a build requirement if one wants the man pages. The alternative is to generate them at source dir and keep them under version control (meson does not add anything to release tarball that is not under version control).

I see. Maybe the output is stable-enough to keep under source control. Not sure. Let's release like this and add in NEWS that distros should build with help2man...

Issue #5232, opened 2025-04-01
Misplaced marks in Lucida Grande


drott:
With "Lucida Grande.ttc", misplaced marks appear in Chrome output rendering when rolling between 8efd2d85c..a1e587b75 

![Image](https://github.com/user-attachments/assets/7e4068d4-9c84-4f75-b9f3-ad0675f53d2f)

Source test:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/virtual/text-antialias/atsui-spacing-features.html

behdad:
It's the fallback mark positioning involved, when there's a AAT deleted-glyph involved between mark and base.

drott:
Confirmed fix in Chromium roll.

PR #5233, opened 2025-04-01
Fix AAT deleted glyph marks interaction with fallback mark positioning

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/5232


behdad:
Will add test for Mac with LucidaGrande.

behdad:
> Will add test for Mac with LucidaGrande.

Done.

Issue #4653, opened 2024-04-08
Make hb_ot_math_get_constant work around Cambria Math's bug for displayOperatorMinHeight/delimitedSubFormulaMinHeight


fred-wang:
cc @khaledhosny 

OpenType doc: https://learn.microsoft.com/en-us/typography/opentype/spec/math#mathconstants-table
Microsoft bug: https://github.com/MicrosoftDocs/typography-issues/issues/1136#issuecomment-2040368757

Basically, Microsoft Word has a bug and reads `displayOperatorMinHeight` and `delimitedSubFormulaMinHeight` in the wrong order. Cambria Math seems to have the parameters inverted too, so that it still works with Microsoft Word. I don't know if because of that other math fonts were badly designed too, although I believed I reported bugs to font creators in the past.

I cannot say much about `delimitedSubFormulaMinHeight`. I'm not sure I understand where it's used for layout. Neither MathML Core not MathML implementations in browsers use it.

However, `displayOperatorMinHeight` is used by several MathML and TeX implementations to draw n-ary operators in display mode, and due to the Cambria Math bug these implementations generally ended up implementing a hack like "ensure the displaystyle is large enough, independently of the value of displayOperatorMinHeight". MathML Core and Chromium's implementation do *not* have such a hack.

So the question is: should HarfBuzz's hb_ot_math_get_constant work around Cambria Math's bug for displayOperatorMinHeight/delimitedSubFormulaMinHeight by inverting the value?

behdad:
SGTM.

fred-wang:
I'm not sure whether Microsoft plans to release a new version of Cambria Math with that bug fixed, or if other math fonts have the issue (would be interesting to check the parameters). Anyway, I'll give it a try.

behdad:
@khaledhosny What's the consensus here?

khaledhosny:
My analysis in the above Microsoft Typography issue, which was conformed by Peter Constable, is that Microsoft implementation is swapping `displayOperatorMinHeight` and `delimitedSubFormulaMinHeight`. This seems to be affecting Cambria Math only as it has too small value for `displayOperatorMinHeight`. We can detect Cambria Math and swap the values, but I’m not sure what is the best way to fingerprint it? May be use the size of the `MATH` table as well as the values for `displayOperatorMinHeight` and `delimitedSubFormulaMinHeight`?

fred-wang:
cc @icesfont

khaledhosny:
[...]

PR #5250, opened 2025-04-05
[ot-math] Workaround bad displayOperatorMinHeight in Cambria Math

khaledhosny:
See https://github.com/harfbuzz/harfbuzz/issues/4653 for details


src/hb-ot-math-table.hh
```diff
@@ -1104,6 +1104,23 @@ struct MATH
...
+  HB_INTERNAL bool is_bad_cambria (hb_font_t *font) const
+  {
+    hb::unique_ptr<hb_blob_t> table_blob {font->face->reference_table (tableTag)};
+    switch HB_CODEPOINT_ENCODE3 (table_blob->length,
+                                 get_constant (HB_OT_MATH_CONSTANT_DISPLAY_OPERATOR_MIN_HEIGHT, font),
+                                 get_constant (HB_OT_MATH_CONSTANT_DELIMITED_SUB_FORMULA_MIN_HEIGHT, font))
+    {
+      /* sha1sum:ab4a4fe054d23061f3c039493d6f665cfda2ecf5  cambria.ttc
+       * sha1sum:086855301bff644f9d8827b88491fcf73a6d4cb9  cambria.ttc
+       * sha1sum:b1e5a3feaca2ea3dfcf79ccb377de749ecf60343  cambria.ttc */
```
fred-wang, line 1117: So have you checked other popular OpenType MATH fonts have different sha1sums?

Using the length of the MATH table sounds good, but I wonder if the two constants are the best way to fingerprint Cambria Math? It seems likely that other fonts could have same values, especially if font designers that just copy what Cambria Math does...

Maybe OT::GDEF::is_blocklisted seems to use other table lengths instead, maybe we could do something similar here?

Or is the idea to rely them on these math constants so that the workaround is ignored if Microsoft fixes that in the future ? (unclear what their plan is)

src/hb-ot-math.cc
```diff
@@ -87,6 +87,20 @@ hb_position_t
...
+   * with the swapped values. */
+  if ((constant == HB_OT_MATH_CONSTANT_DISPLAY_OPERATOR_MIN_HEIGHT ||
+       constant == HB_OT_MATH_CONSTANT_DELIMITED_SUB_FORMULA_MIN_HEIGHT) &&
+      font->face->table.MATH->is_bad_cambria (font))
+  {
+    if (constant == HB_OT_MATH_CONSTANT_DISPLAY_OPERATOR_MIN_HEIGHT)
+      constant = HB_OT_MATH_CONSTANT_DELIMITED_SUB_FORMULA_MIN_HEIGHT;
+    else
+      constant = HB_OT_MATH_CONSTANT_DISPLAY_OPERATOR_MIN_HEIGHT;
+  }
```
fred-wang, line 103: Not sure how easy that will be, but I wonder if we could have a test for that in test/api/test-ot-layout-math.c

src/hb-ot-math-table.hh
khaledhosny: These checksums are here to just document what font files were checked. We check the table length, plus the values of the two constants, I found it quite unlikely that a different font will have exactly the same three values. I didn’t use the lengths of other tables because they seem irrelevant to MATH table, and an update to Cambria is likely to, say, change GSUB or GPOS while MATH table is unchanged, so we will need to add more entries to each version of Cambria now and in the future until it gets fixed.

src/hb-ot-math.cc
khaledhosny: A test will need a font with a MATH table that is exactly the same size as Cambria’s, but short of copying Cambria’s MATH table, I’m nit sure how to make such a font.

src/hb-ot-math-table.hh
fred-wang: I mean, my question was have you tried with other popular fonts like STIX Two MATH, Latin Modern MATH etc and checked they don't return true for is_bad_cambria? I agree it's unlikely to happen, but I think it's worth checking, to avoid more breakage...

src/hb-ot-math.cc
fred-wang: I think creating a font with the same constants is easy. About the table_blob->length, I'm not quite sure, I guess you can try and add dummy entries in the table but not sure you can get exactly the same length.

src/hb-ot-math-table.hh
fred-wang: If you can't get the same length, it seems you could still play with the second value overflow, making it overflow the 21 bits: https://searchfox.org/mozilla-central/rev/ca0f952fed01dc503a9cac23759da68ed137b337/gfx/harfbuzz/src/hb-algs.hh#216

src/hb-ot-math-table.hh
```diff
@@ -1104,6 +1104,23 @@ struct MATH
 		  mathVariants.sanitize (c, this));
   }
 
+  // https://github.com/harfbuzz/harfbuzz/issues/4653
+  HB_INTERNAL bool is_bad_cambria (hb_font_t *font) const
+  {
+    hb::unique_ptr<hb_blob_t> table_blob {font->face->reference_table (tableTag)};
```
behdad, line 1110: Just use `font->face->table.MATH.get_blob()`.

src/hb-ot-math.cc
khaledhosny: I’ve added a test, by padding MATH table with zeros to be exactly the expected length.

behdad:
LGTM. Thanks.

PR #5258, opened 2025-04-06
Faster tests

behdad:
Probably just closing. Want to check.


behdad:
This brings test runtime from 3s to 2s on a 12core machine. But isn't worth not seeing exactly which shape tests fail at a glance.

PR #5259, opened 2025-04-06
Simplify deps

behdad:


behdad:
This brings down the test time from 3s to 2s on a 12core machine. But is not worth the worse error message we will get if any shape test fails...

khaledhosny (approved):

PR #5260, opened 2025-04-06
Use TAP protocol for tests

behdad:
This does it for `api` test since those use glib, which is TAP-aware. If we set verbose on, it will show the actual subsets. But I found it too noisy.

I'm getting one warning I haven't been able to track down:
```
455/472 harfbuzz:api+subset / test-subset-nameids                             OK               0.09s   3 subtests passed

stdout: 14: UNKNOWN: Non-ascii character detected, ignored...This API supports ascii characters only for mac platform
Unknown TAP output lines have been ignored. Please open a feature request to
implement them, or prefix them with a # if they are not TAP syntax.
```


behdad:
The protocol thing is also from newer meson. But let's experiment and see if there's any benefits if we convert the shape and subset tests to it.

behdad:
I tried chunking up subset fuzzer tests from meson.build file such that chunks can run in parallel. Unfortunately the fs module used is meson >= 1.3 or so. :((

behdad:
> I tried chunking up subset fuzzer tests from meson.build file such that chunks can run in parallel.

If we do this, we won't need the test runner really; unless we want it to generate tap numbers.

behdad:
This turned out to be a few different thing. Each directory's changes are independent. We can pick whichever we like.

test/fuzzing/meson.build
```diff
@@ -66,11 +57,61 @@ test('repacker-fuzzer', find_program('run-fuzzer-tests.py'),
...
+
+ls = find_program('ls')
+if not ls.found()
+  # On Windows, we need to use the 'dir' command to list files in a directory.
+  # The 'ls' command is not available by default.
+  find = [find_program('dir', required: true), '/b', '/a-d']
+else
+  # On other platforms, we can use the 'ls' command.
+  find = [find_program('ls', required: true)]
+endif
```
khaledhosny, line 77: Not sure if I like this. It means tests will now fail on non-unix non-Windows systems (no idea if these actually exist and matter). What about using Python to glob the dir? A small Python program that calls `pathlib.Path().iterdir()`?

.circleci/config.yml
```diff
@@ -149,8 +149,9 @@ jobs:
     executor: win64-executor
     steps:
       - checkout
+      - run: dpkg --add-architecture i386
       - run: apt update
-      - run: DEBIAN_FRONTEND=noninteractive apt install -y ninja-build python3 python3-pip python3-venv git g++-mingw-w64-x86-64 zip wine wine64
+      - run: DEBIAN_FRONTEND=noninteractive apt install -y ninja-build python3 python3-pip python3-venv git g++-mingw-w64-x86-64 zip wine wine64 wine32:i386
```
khaledhosny, line 154: That is odd, why is it complaining about wine32 now? May be we should be using `wine64` in win64-cross-file.txt?

test/fuzzing/meson.build
khaledhosny: Or since meson really [doesn’t want us to glob paths](https://mesonbuild.com/FAQ.html#why-cant-i-specify-target-files-with-a-wildcard), may be it is time to bite the bullet and list all the fuzzing fonts in the meson file.

.circleci/config.yml
behdad: It might have been to run `dir`.

test/fuzzing/meson.build
behdad: If we do list all fonts, we can also have a test that make sure they are all in git...

Issue #5172, opened 2025-03-19
[coretext] Set kCTFontOpticalSizeAttribute to "none"?


behdad:
> I did some investigation. I cannot find any shaping change with the `ot` shaper and SFNSItalic in that commit range. However, with `coretext` shaper I do: because of commit 38db0f4c1212433fe266a53df629df4d9d9929e1, which is simply setting *all* axes on a CTFont; before we were skipping axes at their default position.
> 
> I believe that has to do with how CoreText, under *some* conditions, sets `opsz` automatically to the font ptem. And the default font ptem is 12 in CoreText, but default opsz axis is at 28 in SFNSItalic. Hence the change.
> 
> If I set `--font-ptem=X` as well as `--variations opsz=X` to both shapers, I have not been able to spot any descrepancies.
> 
> The mysterious behavior is controlled by `kCTFontOpticalSizeAttribute`, which is not documented:
> 
>   https://developer.apple.com/documentation/coretext/kctfontopticalsizeattribute
> 
> However, the CoreText headers have this to offer:
> ```
> /*!
>     @defined    kCTFontOpticalSizeAttribute
>     @abstract   The point size at which this font is intended to be used.
>     @discussion The value is a CFNumber used to activate size-specific (not linearly scaled) metrics. Starting with macOS 10.14 and iOS 12.0, the CFString "auto" can be used instead to request an optical size matching the point size. Starting with macOS 10.15 and iOS 13.0, the CFString "none" can be used instead to explicitly disable automatic optical sizing enabled by the font.
> */
> ```
> From what I've heard, they enable traking only for fonts with STAT table, and "system chain". No idea how that statement relates to the above attribute.
> 
> If we can see what Chrome expects in terms of an hb-shape command or something, that would significantly help with debugging. 

 _Originally posted by @behdad in [#5053](https://github.com/harfbuzz/harfbuzz/issues/5053#issuecomment-2734982095)_

behdad:
I'm not sure where this would go. Maybe doesn't matter since we set all variable axes.

behdad:
@khaledhosny Is this something you can help with?

khaledhosny:
See also https://chromium.googlesource.com/skia/+/9ae6732c1f4504c1dae0316f531442ff54b2c8e9/src/ports/SkFontHost_mac.cpp#1028

khaledhosny:
I tried this:
```diff
[...]

PR #5270, opened 2025-04-07
[coretext-shape] Set kCTFontOpticalSizeAttribute to "none"

khaledhosny:
Fixes https://github.com/harfbuzz/harfbuzz/issues/5172


behdad:
SFNS / SFNSItalic tests failing. I suggest we don't touch until somebody else yells.

PR #5282, opened 2025-04-11
[subset] close unicodes over bidi mirror variants during subsetting.

### LINKED ISSUE #5281: [subset] Close unicode set over Bidi_Mirroring property
See https://github.com/google/fonts/issues/9335

garretrieger:
Fixes #5281. Does the closure by default, but I've introduced a new flag and option to disable this behaviour since some users may want to get the minimal set if they know they don't need the mirrored variants.


src/hb-subset-plan.cc
```diff
@@ -209,15 +211,46 @@ _fill_unicode_and_glyph_map(hb_subset_plan_t *plan,
...
 }
 
+/*
+ * Finds additional unicode codepoints which are reachable from the input unicode set.
+ * Currently this adds in mirrored variants (needed for bidi) of any input unicodes.
+ */
+static hb_set_t
+_unicode_closure (const hb_set_t* unicodes, bool bidi_closure) {
+  // TODO: we may want to also consider pulling in reachable unicode composition and decompositions.
+  //       see: #2283
```
behdad, line 221: Full github link please.

garretrieger:
There's an open question here of how to handle the closure with inverted sets (ie. sets that say include everything except these codepoints).

If we do closure on these we may end up pulling in codepoints that were asked to be excluded in the input. So for now if I detect an inverted set I just skip the closure. However, I think probably worth some discussion if that's the right behaviour.

src/hb-subset-plan.cc
```diff
@@ -209,15 +211,46 @@ _fill_unicode_and_glyph_map(hb_subset_plan_t *plan,
...
+  if (!bidi_closure) return out;
+
+  if (out.is_inverted()) {
+    // don't closure inverted sets, they are asking to specifically exclude certain codepoints.
+    // otherwise everything is already included.
+    return out;
+  }
+
+  for (hb_codepoint_t cp : *unicodes) {
+   hb_codepoint_t mirror = hb_unicode_mirroring(hb_unicode_funcs_get_default (), cp);
```
behdad, line 232: Save `...get_default ()` to a variable.

src/hb-subset-plan.cc
```diff
@@ -209,15 +211,46 @@ _fill_unicode_and_glyph_map(hb_subset_plan_t *plan,
...
+   if (mirror != cp) {
+     out.add(mirror);
+   }
+   out.add(cp);
```
behdad, line 236: Why add `cp`? Isn't `out` already a copy of `unicodes`?

src/hb-subset.h
```diff
@@ -75,6 +75,8 @@ typedef struct hb_subset_plan_t hb_subset_plan_t;
  * substitution rules (GSUB). Since: 7.2.0.
  * @HB_SUBSET_FLAGS_OPTIMIZE_IUP_DELTAS: If set perform IUP delta optimization on the
  * remaining gvar table's deltas. Since: 8.5.0
+ * @HB_SUBSET_FLAGS_NO_BIDI_CLOSURE: If set don't pull mirrored versions of input
```
behdad, line 78: I have a preference for `do not` over `don't`.

src/hb-subset.h
```diff
@@ -96,8 +98,9 @@ typedef enum { /*< flags >*/
   HB_SUBSET_FLAGS_NO_PRUNE_UNICODE_RANGES =  0x00000100u,
   HB_SUBSET_FLAGS_NO_LAYOUT_CLOSURE =        0x00000200u,
   HB_SUBSET_FLAGS_OPTIMIZE_IUP_DELTAS	  =  0x00000400u,
+  HB_SUBSET_FLAGS_NO_BIDI_CLOSURE         =  0x00000800u,
```
behdad, line 101: I'm on the fence about the flag. I suppose it doesn't hurt.

behdad (changes requested):
LGTM with changes.

behdad:
> There's an open question here of how to handle the closure with inverted sets (ie. sets that say include everything except these codepoints).
> 
> If we do closure on these we may end up pulling in codepoints that were asked to be excluded in the input. So for now if I detect an inverted set I just skip the closure. However, I think probably worth some discussion if that's the right behaviour.

I think I agree with you very much.

Also, cc @jfkthame

src/hb-subset-plan.cc
garretrieger: Oops, reworked that while writing the code and forgot to remove the add. Fixed.

src/hb-subset-plan.cc
```diff
@@ -209,15 +211,46 @@ _fill_unicode_and_glyph_map(hb_subset_plan_t *plan,
...
+  if (out.is_inverted()) {
+    // don't closure inverted sets, they are asking to specifically exclude certain codepoints.
+    // otherwise everything is already included.
+    return out;
+  }
+
+  auto unicode_funcs = hb_unicode_funcs_get_default ();
+  for (hb_codepoint_t cp : *unicodes) {
+   hb_codepoint_t mirror = hb_unicode_mirroring(unicode_funcs, cp);
+   if (mirror != cp) {
```
behdad, line 234: Maybe throw an `unlikely` here.

behdad (approved):

PR #5283, opened 2025-04-12
[meson] Clean up warnings and fix usage of file objects

robxnano:
Hi there, this is just a small cleanup to remove some of the repeated messages and warnings in the Meson build output. The first change just suppresses the repeated (about 20) calls to `Fetching value of define "_MSC_FULL_VER" : (undefined)` when compiling for Windows, while the second gets rid of `src/meson.build:850: WARNING: Project targets '>= 0.60.0' but uses feature introduced in '1.3.0': fs.relative_to.` by checking the Meson version directly in the if statements enclosing them.

The third is for a more serious issue: `src/meson.build:1055: DEPRECATION: Project uses feature that was always broken, and is now deprecated since '1.3.0': str.format: Value other than strings, integers, bools, options, dictionaries and lists thereof..` However, the only way to get the actual paths for a files() object is to use the full_path() function which was only introduced in 1.4.0, so I had to enclose those in if statements as well.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -1042,11 +1042,17 @@ if get_option('tests').enabled()
   env.set('libs', meson.current_build_dir()) # TODO: Merge this with builddir after autotools removal
   HBSOURCES = []
   foreach f : hb_sources
+    if meson.version().version_compare('>=1.4.0')
+      f = f.full_path()
+    endif
     HBSOURCES += '@0@'.format(f)
   endforeach
   env.set('HBSOURCES', ' '.join(HBSOURCES))
   HBHEADERS = []
   foreach f : hb_headers
+    if meson.version().version_compare('>=1.4.0')
+      f = f.full_path()
+    endif
     HBHEADERS += '@0@'.format(f)
   endforeach
   env.set('HBHEADERS', ' '.join(HBHEADERS))
--- a/src/meson.build
+++ b/src/meson.build
@@ -840,15 +840,10 @@ have_gobject = conf.get('HAVE_GOBJECT', 0) == 1
 cmake_config = configuration_data()
 cmake_config_dir = cmake_package_install_dir / 'harfbuzz'
 
-have_fs_relative_to = meson.version().version_compare('>=1.3.0')
-
-if not have_fs_relative_to
-  relative_to = find_program('relative_to.py')
-endif
-
-if have_fs_relative_to
+if meson.version().version_compare('>=1.3.0')
   cmake_package_prefix_dir = fs.relative_to(get_option('prefix'), get_option('prefix') / cmake_config_dir)
 else
+  relative_to = find_program('relative_to.py')
   cmake_package_prefix_dir = run_command(relative_to, get_option('prefix'), get_option('prefix') / cmake_config_dir, check: true).stdout().strip()
 endif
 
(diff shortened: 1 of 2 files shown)


behdad:
Thank you!

src/meson.build
```diff
@@ -860,7 +855,7 @@ cmake_package_prefix_dir = '${CMAKE_CURRENT_LIST_DIR}/@0@'.format(cmake_package_
 cmake_install_includedir = get_option('includedir')
 
 if fs.is_absolute(cmake_install_includedir)
-  if have_fs_relative_to
+  if meson.version().version_compare('>=1.3.0')
```
khaledhosny, line 858: I don’t like that we now have this version check allover the place. I’d say it is a meson bug that it can’t detect that `have_fs_relative_to` is only true for `meson.version().version_compare('>=1.3.0')`.

src/meson.build
```diff
@@ -1042,11 +1042,17 @@ if get_option('tests').enabled()
   env.set('libs', meson.current_build_dir()) # TODO: Merge this with builddir after autotools removal
   HBSOURCES = []
   foreach f : hb_sources
+    if meson.version().version_compare('>=1.4.0')
+      f = f.full_path()
```
khaledhosny, line 1046: The scripts that use these lists want relative paths, not absolute ones.

src/meson.build
robxnano: I'll remove this one if you want.

src/meson.build
robxnano: Sorry, I didn't realize - since the tests all passed I assumed the scripts were able to handle them. I'll use fs.relative_to() to get back to the relative paths then.

src/meson.build
khaledhosny: I think it is OK, we will upgrade to >=1.3 one day...

src/meson.build
khaledhosny: The scripts will convert the path to absolute path by joining them with base_srcdir, so full path would have simplified things if it were available in older versions.

khaledhosny: Or may be full path here is fine. It seems that `os.path.join(a, b)` will just return `b` if it is an absolute path.

khaledhosny:
I don’t understand this link failure.

behdad:
> I don’t understand this link failure.

Somehow `hb-static.cc` is not being linked now.

src/meson.build
```diff
@@ -721,7 +721,7 @@ if get_option('tests').enabled()
   # TODO: Microsoft LINK gives the following because extern, non dllexport
   # symbols can only be used when linking against a static library
   # error LNK2019: unresolved external symbol "unsigned __int64 const * const _hb_NullPool"
-  if cpp.get_define('_MSC_FULL_VER') == ''
+  if cpp_is_microsoft_compiler
```
khaledhosny, line 724: This should be `if not` or whatever equivalent meson syntax is.

khaledhosny:
> > I don’t understand this link failure.
> 
> Somehow `hb-static.cc` is not being linked now.

Found it. We were not compiling these files for msvc because this link failure, now we inadvertently do.

robxnano:
Would you prefer if I rebased that change into the earlier commit?

behdad:
We can squash. No worries.

PR #5315, opened 2025-05-06
increase maximum set_flags range

caolanm:
This a problem for LibreOffice since:

commit 0d8717275ac5809812c254c1b0923f2fc052c582
CommitDate: Tue Apr 22 15:40:06 2025 -0600

    [buffer] Limit the extent of set_flags range

where we now see an asset of:

vcl/source/gdi/impglyphitem.cxx:319: void checkGlyphsEqual(const SalLayoutGlyphs&, const SalLayoutGlyphs&): Assertion `l1->isLayoutEquivalent(l2)' failed.

as noted in https://gerrit.libreoffice.org/c/core/+/184761

on checking in debug-builds if reusing a subset of a previous laid out string as a cache for the same subtext seen again does really give the same results.

In our test documents, empirically checking what number here works, there's one case which still fails at len 2029 but succeeds at 2030.

### DIFF
--- a/src/hb-buffer.hh
+++ b/src/hb-buffer.hh
@@ -412,7 +412,7 @@ struct hb_buffer_t
   {
     end = hb_min (end, len);
 
-    if (unlikely (end - start > 255))
+    if (unlikely (end - start >= 2048))
       return;
 
     if (interior && !from_out_buffer && end - start < 2)


behdad:
Thanks for the report.

> In our test documents, empirically checking what number here works, there's one case which still fails at len 2029 but succeeds at 2030.

What is that test case doing? This sounds like a constructed edge case not relevant to real text to me.

caolanm:
The test-cases are just documents scraped from our bugzilla. The one with this longest case is: https://bugs.documentfoundation.org/attachment.cgi?id=137495 from https://bugs.documentfoundation.org/show_bug.cgi?id=74693 which I think is a Portuguese language academic book.

On Page 8 there is a 2035 character long paragraph starting with "Ora uma análise cuidada dos dois trabalhos referidos mostrou que," which I think we layout repeatedly as we split it over lines, and try and reuse the results from the initial full paragraph.

behdad:
The question is why we call `set_flags` with a large range. Any chance you can get a backtrace for me? Thanks.

caolanm:
[backtrace-full.log](https://github.com/user-attachments/files/20092916/backtrace-full.log)

behdad:
I see what's going on. This is going through the graphite shaper, which marks the entire string as unsafe-to-break, hence the large `end - start` value.

Let me think of a fix.

caolanm:
[That](https://github.com/harfbuzz/harfbuzz/commit/089e73a76419ffc100a37216374c777198a05011) doesn't help. I see that end is clipped to len before that new -1 check

```
    end = hb_min (end, len);

    if (unlikely (end != (unsigned) -1 && end - start > 255))
      return;
```

maybe this should instead be something like:

```
    if (unlikely (end != (unsigned) -1 && end - start > 255))
      return;

    end = hb_min (end, len);
```

which does work for me

behdad:
Fixed. Thanks.

PR #5317, opened 2025-05-06
[ci/configs] Test HB_NO_DRAW and HB_NO_PAINT

khaledhosny:


behdad:
Is the revert intentional?

khaledhosny:
> Is the revert intentional?

Yes, I want to see if it will catch the issue, but apparently it doesn’t

khaledhosny:
With linking (no `-c` option to gcc), `HB_TINY` catches the issue, but I’m keeping the separate `HB_NO_DRAW` and `HB_NO_PAINT` just in case.

PR #5320, opened 2025-05-10
Skippy iter simplify

behdad:


behdad:
This is to make harfRuzz impl closer to HB.

PR #5327, opened 2025-05-14
Fix cmap4 overflow pathway.

### LINKED ISSUE #5326: Segfaults with some large subsets
We're seeing some crashes generating large subsets of large fonts. Happily some of these are open source. I've attached a script that exits with "Abort trap" and produces an empty file, and also the font the script processes. 

In the (two) examples I've traced the problem comes up in cmap serialization code. 

[crash_files.zip](https://github.com/user-attachments/files/20214149/crash_files.zip)

garretrieger:
cmap4 serialize was doing a pop_discard() on overflow, however the encoding record above it was also doing one as well. Remove the unnessecary cmap4 pop_dicard(). Note: in cmap4 there was a comment implying the subsequent pop_discard() would be a noop due to the set error, however this is not true since pop_discard() still works when there's been only an overflow error.

Fixes #5326


garretrieger:
I'll see if I can get a test added that hits a cmap4 overflow.

behdad:
LGTM.

behdad:
LGTM.

PR #5329, opened 2025-05-16
Disable OTS on the cmap4 overflow test case.

garretrieger:
The test file has been stripped of most tables to avoid checking in a very large font (since it has a high number of codepoints), which causes OTS to fail on missing required tables.


behdad:
THansk. Would be nice to make at least one bot run with ots as well.

garretrieger:
> THansk. Would be nice to make at least one bot run with ots as well.

Yeah, agreed.

PR #5333, opened 2025-05-20
cmake: always set inline hidden flag.

sezero:
This practically reverts commit ac92ed7d6875374451246a2391859fb763329adb:
minimum required cmake version is more than new enough.


sezero:
PING ?

PR #5346, opened 2025-05-28
Big misc changes

behdad:
These are from a private branch I've been carrying for some work that cannot be currently published.

- Make geometry types into templates taking float or double.
- Add little-endian integer types.
- Add `perf/hb-draw-compare` and `perf/hb-svg-compare` tools.
- Changes to paint() transform elimination.
- Add font->has_nonzero_coords and use it across.
- Extensions to avar processing for otherwise invalid mappings.

PR #5349, opened 2025-05-31
Small misc changes

behdad:
Breaking up https://github.com/harfbuzz/harfbuzz/pull/5346 into small changes, to see what change fails the fedora-valgrind bot in that PR.


behdad:
@khaledhosny Apparently the fast-path in https://github.com/harfbuzz/harfbuzz/pull/5349/commits/655b97a9020a1575bc55bf5ad86944624ed0dd49 makes the valgrind bot fail. Any chance I can get a review from you? My eyes are sore from staring at it...

khaledhosny:
All CI bots pass now.

khaledhosny:
Would this warning be related:
```
In file included from ../src/hb.hh:549,
                 from ../src/hb-subset.cc:27:
In member function ‘constexpr HBInt<BE, Type, 2>::operator Type() const [with bool BE = true; Type = short int]’,
    inlined from ‘typename Type::type OT::HBFixed<Type, fraction_bits>::to_int() const [with Type = OT::NumType<true, short int>; unsigned int fraction_bits = 14]’ at ../src/hb-open-type.hh:244:54,
    inlined from ‘Triple OT::SegmentMaps::unmap_axis_range(const Triple&) const’ at ../src/hb-ot-var-avar-table.hh:189:45,
    inlined from ‘bool OT::SegmentMaps::subset(hb_subset_context_t*, hb_tag_t) const’ at ../src/hb-ot-var-avar-table.hh:218:46:
../src/hb-algs.hh:143:63: warning: ‘*(const hb_packed_t<short unsigned int>*)((char*)&val + offsetof(OT::F2DOT14, OT::HBFixed<OT::NumType<true, short int, 2>, 14>::<unnamed>.OT::NumType<true, short int, 2>::v.HBInt<true, short int, 2>::v[0])).hb_packed_t<short unsigned int>::v’ may be used uninitialized [-Wmaybe-uninitialized]
  143 |       __builtin_bswap16 (((const hb_packed_t<uint16_t> *) v)->v)
      |                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
In file included from ../src/hb-subset.cc:54:
../src/hb-ot-var-avar-table.hh: In member function ‘bool OT::SegmentMaps::subset(hb_subset_context_t*, hb_tag_t) const’:
../src/hb-ot-var-avar-table.hh:186:13: note: ‘val’ declared here
  186 |     F2DOT14 val, unmapped_val;
      |             ^~~
```

khaledhosny:
I can reproduce the build warning and the valgrind failure locally on Debian with GCC 12.4 (fedora-valgrind bot also uses GCC 12.2, though a lower minor version) but not GCC 14. My guess is that it is a compiler bug (may be incorrectly labeling the variable as uninitialized then proceeding to do some dangerous optimization). We can just brush it off as compiler bug and update that bot to use newer GCC, your call.

khaledhosny:
The tests even fail without valgrind.

behdad:
> We can just brush it off as compiler bug and update that bot to use newer GCC

Can you please do that?

khaledhosny:
> > We can just brush it off as compiler bug and update that bot to use newer GCC
> 
> Can you please do that?

Done.

behdad:
Thanks.

PR #5350, opened 2025-06-01
Misc. CI fixes

khaledhosny:


behdad:
The names should be changed too.

khaledhosny:
> The names should be changed too.

Fixed.

khaledhosny:
I don’t know what is going on with CircleCI. Installing packages with apt seems to be randomly failing.

behdad:
> I don’t know what is going on with CircleCI. Installing packages with apt seems to be randomly failing.

Can we contact them for support? We are a paying customer by the way.

Other than that, can we at least make the bot "skip" instead of fail?

khaledhosny:
> > I don’t know what is going on with CircleCI. Installing packages with apt seems to be randomly failing.
> 
> Can we contact them for support? We are a paying customer by the way.

I have no idea how to do that.

> Other than that, can we at least make the bot "skip" instead of fail?

We don’t want builds to silently fail even if it is failing due to an external cause. Otherwise we might miss an important build configuration not being tested on CI.

behdad:
I looked for contact ways and couldn't find either.

Moving pretty much all to github sounds good. Thanks for doing this.

khaledhosny:
This unbreaks the build for now. I might migrate the remaining jobs off CircleCI later and retire it for good.

PR #5355, opened 2025-06-06
[harfruzz] Update to new API

behdad:
From https://github.com/harfbuzz/harfruzz/pull/57

### DIFF
--- a/src/rust/shape.rs
+++ b/src/rust/shape.rs
@@ -172,11 +172,9 @@ pub unsafe extern "C" fn _hb_harfruzz_shape_rs(
         hb_buffer_cluster_level_t_HB_BUFFER_CLUSTER_LEVEL_CHARACTERS => {
             harfruzz::BufferClusterLevel::Characters
         }
-        /* TODO: Enable when added to HarfRuzz
         hb_buffer_cluster_level_t_HB_BUFFER_CLUSTER_LEVEL_GRAPHEMES => {
             harfruzz::BufferClusterLevel::Graphemes
         }
-        */
         _ => harfruzz::BufferClusterLevel::default(),
     };
     hr_buffer.set_cluster_level(cluster_level);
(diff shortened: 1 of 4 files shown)


src/rust/shape.rs
```diff
@@ -65,7 +64,7 @@ fn font_coords_to_f2dot14(font: *mut hb_font_t) -> Vec<F2Dot14> {
     };
     coords
         .iter()
-        .map(|&v| F2Dot14::from_bits(v as i16))
+        .map(|&v| NormalizedCoord::from_bits(v as i16))
```
dfrg: If you make this function return `ShaperInstance`, you should be able to pass this iterator directly as the second argument to `ShaperInstance::from_coords` and avoid the collect into an intermediate allocation.

src/rust/shape.rs
```diff
@@ -215,33 +219,34 @@ pub unsafe extern "C" fn _hb_harfruzz_shape_rs(
...
-                    }
-                })
-                .collect::<Vec<_>>()
-        };
-        harfruzz::shape(face, &features, hr_buffer)
+        shaper.shape(hr_buffer, &features)
     } else {
         let shape_plan = shape_plan as *const harfruzz::ShapePlan;
-        harfruzz::shape_with_plan(face, shape_plan.as_ref().unwrap(), hr_buffer)
+        shaper.shape_with_plan(hr_buffer, shape_plan.as_ref().unwrap(), &features)
```
dfrg: We could potentially change the harfruzz API to accept an iterator for features as well which would allow users to avoid allocations here too.

src/rust/shape.rs
behdad: Thanks.

src/rust/shape.rs
behdad: I don't think that's a big deal really.

dfrg:
This LGTM now

PR #5356, opened 2025-06-07
[harfruzz] Set point_size

behdad:
Tests fail.

@dfrg Ideas?


dfrg:
I think you need to set the point size when building the shaper in `_hb_harfruzz_shape_rs` rather than when creating the plan. This seems like a bit of a footgun :/ May need to rethink things.

behdad:
> I think you need to set the point size when building the shaper in `_hb_harfruzz_shape_rs` rather than when creating the plan. This seems like a bit of a footgun :/ May need to rethink things.

Good point.

Also found the bug. Looks good so far.

PR #5382, opened 2025-07-09
[repacker] Implement table splitting for LigatureSubst

garretrieger:
Ran into cases where fonts (eg. symbol fonts) with large number of ligatures are overflowing LigatureSubst subtables. This adds support for splitting them to resolve overflows.

Also fixes a bug with handling of node duplication where there are virtual links (LigatureSubst's use virtual links to enforce ordering of the coverage table).

### DIFF
--- a/src/OT/Layout/GSUB/LigatureSet.hh
+++ b/src/OT/Layout/GSUB/LigatureSet.hh
@@ -11,11 +11,11 @@ namespace GSUB_impl {
 template <typename Types>
 struct LigatureSet
 {
-  protected:
+  public:
   Array16OfOffset16To<Ligature<Types>>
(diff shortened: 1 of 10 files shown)


behdad:
Thanks. For now:
```
src/graph/gsubgpos-graph.hh:27:10: fatal error: 'OT/Layout/GSUB/SubstLookupSubTable.hh' file not found
```

garretrieger:
> Thanks. For now:
> 
> ```
> src/graph/gsubgpos-graph.hh:27:10: fatal error: 'OT/Layout/GSUB/SubstLookupSubTable.hh' file not found
> ```

Should be fixed now.

garretrieger:
@cmyr

PR #5383, opened 2025-07-10
[subset] bug fix for AnchorFormat3

qxliu76:
Don't downgrade to format1 when subsetting, otherwise hinting device is dropped.

### DIFF
--- a/src/OT/Layout/GPOS/AnchorFormat3.hh
+++ b/src/OT/Layout/GPOS/AnchorFormat3.hh
@@ -94,7 +94,8 @@ struct AnchorFormat3
     /* in case that all axes are pinned or no variations after instantiation,
      * both var_idxes will be mapped to HB_OT_LAYOUT_NO_VARIATIONS_INDEX */
     if (x_varidx == HB_OT_LAYOUT_NO_VARIATIONS_INDEX &&
-        y_varidx == HB_OT_LAYOUT_NO_VARIATIONS_INDEX)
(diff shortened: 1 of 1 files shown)


behdad:
I don't understand the logic here.

qxliu76:
> I don't understand the logic here.

ah, maybe the code is not organized in a good way. 
So firstly we get old `x_varidx` and `y_varidx `value from Device table( `get_variation_index ()` API is called, and it returns `HB_OT_LAYOUT_NO_VARIATIONS_INDEX` for HintingDevice), we then replace the old varidx-> new varidx. Then if both new values are `NO_VARIATIONS_INDEX`, which means no variations after instancing, we downgrade format to 1.

However, in case of HintingDevice, both varidxes are always `NO_VARIATIONS_INDEX`,  we should not downgrade the format, that's why I added an instancing check here. 

And for VariationDevice, we collect and remap variation idxes during planning, when we get here, we know this table will be retained, thus no downgrade for subsetting.

Is that logic clear now or you'd recommend a better way to do it?

behdad:
I would think we'd just check the Anchor type to be VariationDevice. Maybe add a `is_variation_device` to Device? Checking for normalized-coords seems a bit icky here.

qxliu76:
Thanks! Yeah it's better to check the device type. I've got a PR out for this: #5387 , unfortunately I still didn't find a testcase that reproduces this issue.

Issue #5384, opened 2025-07-11
[fontations] Speed up `glyph_from_name`


behdad:
Currently each `hb_font_t` caches the name->gid mapping. This should be cached (through user-data) on the face.

cc https://github.com/googlefonts/fontations/pull/1559#issuecomment-3060001615

PR #5385, opened 2025-07-11
[fontations] Cache glyph_from_names hashmap on the face

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/5384

### DIFF
--- a/src/rust/Cargo.toml
+++ b/src/rust/Cargo.toml
@@ -1,6 +1,7 @@
 [package]
 name = "harfbuzz_rust"
 edition = "2021"
+rust-version = "1.75.0"
 
 [dependencies]
 skrifa = { version = "0.*", optional = true }
--- a/src/rust/font.rs
+++ b/src/rust/font.rs
@@ -5,7 +5,7 @@ use std::ffi::c_void;
 use std::mem::transmute;
 use std::ptr::null_mut;
 use std::sync::atomic::{AtomicPtr, AtomicU32, Ordering};
-use std::sync::{Mutex, OnceLock};
+use std::sync::Mutex;
 
 use skrifa::charmap::Charmap;
 use skrifa::charmap::MapVariant::Variant;
@@ -35,7 +35,6 @@ struct FontationsData<'a> {
     outline_glyphs: OutlineGlyphCollection<'a>,
     color_glyphs: ColorGlyphCollection<'a>,
     glyph_names: GlyphNames<'a>,
-    glyph_from_names: OnceLock<HashMap<String, hb_codepoint_t>>,
     size: Size,
     vert_metrics: Option<Vmtx<'a>>,
     vert_origin: Option<Vorg<'a>>,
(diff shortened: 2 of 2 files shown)


behdad:
@dfrg Do the unsafes look right to you?

dfrg:
This LGTM but I think we can use `Box` rather than `Arc` here since the reference counting is handled by `hb_face_t`.

behdad:
> This LGTM but I think we can use `Box` rather than `Arc` here since the reference counting is handled by `hb_face_t`.

Good point. Did that.

However, looks like the lifecycle management is / was broken, so it crashes actually. Debugging.

behdad:
Is `&raw mut` kosher to use, or too recent? Clippy wants me to use it.

behdad:
@dfrg PTAL.

I get clippy fail. Any suggestion for how to address that part? Thanks.

valadaptive:
> I get clippy fail. Any suggestion for how to address that part? Thanks.

Are you using the `hb_user_data_key_t *key` parameter as  a "symbol" of sorts, where the key is its address?

`&raw mut` was [stabilized in Rust 1.82](https://blog.rust-lang.org/2024/10/17/Rust-1.82.0/#native-syntax-for-creating-a-raw-pointer) last October. If you want to match fontations' MSRV of 1.75, you can't use it. (You can put `rust-version = "1.75"` under the `[package]` section of your `Cargo.toml`, which may help Clippy catch on; I'm not sure).

The alternative would be the `std::ptr::addr_of_mut!` macro, mentioned in the above blog post as the way of doing things before `&raw mut` was stabilized. In this case, I think it'd be `std::ptr::addr_of_mut!(GLYPH_FROM_NAMES_KEY)`.

I'm not sure why Clippy suggests `&raw mut GLYPH_FROM_NAMES_KEY as *mut _` as a replacement--the `&raw mut` operator already returns a raw pointer, so you should be able to just do `&raw mut GLYPH_FROM_NAMES_KEY` if you go that route.

behdad:
> > I get clippy fail. Any suggestion for how to address that part? Thanks.
> 
> Are you using the `hb_user_data_key_t *key` parameter as a "symbol" of sorts, where the key is its address?

Correct. It's just a unique identifier, assigned by the linker. The API should have been marked to take it as const, but that's water under the bridge.

> `&raw mut` was [stabilized in Rust 1.82](https://blog.rust-lang.org/2024/10/17/Rust-1.82.0/#native-syntax-for-creating-a-raw-pointer) last October. If you want to match fontations' MSRV of 1.75, you can't use it. (You can put `rust-version = "1.75"` under the `[package]` section of your `Cargo.toml`, which may help Clippy catch on; I'm not sure).

Thanks. Done.

> The alternative would be the `std::ptr::addr_of_mut!` macro, mentioned in the above blog post as the way of doing things before `&raw mut` was stabilized. In this case, I think it'd be `std::ptr::addr_of_mut!(GLYPH_FROM_NAMES_KEY)`.

That seems to fix it. Thanks.

PR #5387, opened 2025-07-11
[subset] AnchorFormat3: better check for when should downgrade to format1

qxliu76:
See also #5383

### DIFF
--- a/src/OT/Layout/GPOS/AnchorFormat3.hh
+++ b/src/OT/Layout/GPOS/AnchorFormat3.hh
@@ -91,11 +91,13 @@ struct AnchorFormat3
       }
     }
 
-    /* in case that all axes are pinned or no variations after instantiation,
-     * both var_idxes will be mapped to HB_OT_LAYOUT_NO_VARIATIONS_INDEX */
-    if (x_varidx == HB_OT_LAYOUT_NO_VARIATIONS_INDEX &&
-        y_varidx == HB_OT_LAYOUT_NO_VARIATIONS_INDEX &&
-        c->plan->normalized_coords)
+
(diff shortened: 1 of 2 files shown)


behdad:
The logic still reads problematic to me. What if one device is null, and the other non-null? We might still be able to downgrade that. I think we want something like:
```
if ((xDevice != null && !xDevice.is_variation_device) || x_varidx != NO_VARIATION ||
    (yDevice != null && !yDevice.is_variation_device) || y_vardx != NO_VARIATION)
  don't downgrade...
```
This handles the case of null fine.

qxliu76:
> The logic still reads problematic to me. What if one device is null, and the other non-null? We might still be able to downgrade that. I think we want something like:
> 
> ```
> if ((xDevice != null && !xDevice.is_variation_device) || x_varidx != NO_VARIATION ||
>     (yDevice != null && !yDevice.is_variation_device) || y_vardx != NO_VARIATION)
>   don't downgrade...
> ```
> 
> This handles the case of null fine.

Fixed. Thank you~

src/OT/Layout/GPOS/AnchorFormat3.hh
```diff
@@ -91,11 +91,18 @@ struct AnchorFormat3
       }
     }
 
-    /* in case that all axes are pinned or no variations after instantiation,
+
+    /* downgrade to format1 if:
+     * 1. both Device offsets are NULL
+     * 2. Device is VariationDevice and all axes are pinned or no variations after instantiation,
```
behdad: Now the comment is stale. Just remove it is fine.

qxliu76: Done.

PR #5389, opened 2025-07-13
const up modifier_combining_marks

caolanm:
move this symbol out of the .data section


behdad:
Thanks! Don't know how this happened.

PR #5390, opened 2025-07-15
[subset] bug fix for COLRv1 layer index closure

qxliu76:
In PaintColrLayers table, numLayers could be 0. 
The font file that could reproduce the issue(invalid new layer index(-1)) is 430MB. I'm unable to generate a small subset font file that reproduces the same issue because layers are compact after subsetting.

### DIFF
--- a/src/OT/Color/COLR/COLR.hh
+++ b/src/OT/Color/COLR/COLR.hh
@@ -178,7 +178,14 @@ struct hb_colrv1_closure_context_t :
   { glyphs->add (glyph_id); }
 
   void add_layer_indices (unsigned first_layer_index, unsigned num_of_layers)
-  { layer_indices->add_range (first_layer_index, first_layer_index + num_of_layers - 1); }
+  {
+    if (num_of_layers == 0)
+    {
+      layer_indices->add (first_layer_index);
+      return;
+    }
+    layer_indices->add_range (first_layer_index, first_layer_index + num_of_layers - 1);
+  }
 
   void add_palette_index (unsigned palette_index)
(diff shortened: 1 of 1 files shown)


behdad:
I have no clue what this code does, but sounds like if `numLayers` is 0, we should not encode anything?

behdad:
Oh, I see `numLayers` probably means `numLayersMinusOne`?

behdad:
> Oh, I see `numLayers` probably means `numLayersMinusOne`?

Nah. Doesn't make sense here.

qxliu76:
> I have no clue what this code does, but sounds like if `numLayers` is 0, we should not encode anything?

This code is for layer index closure in COLRv1. Basically we iterate all color glyphs that'll be retained and collect all layer indices that'll be retained. 
It's also my question: should we retain this layer or even the color glyph if numLayers is 0? This glyph is valid in original font file with numLayers=0.

behdad:
> It's also my question: should we retain this layer or even the color glyph if numLayers is 0? This glyph is valid in original font file with numLayers=0.

That sounds like a valid glyph that, renders nothing? @rsheeter

behdad:
I really need help here. And prefer to land soon.

rsheeter:
> That sounds like a valid glyph that, renders nothing? @rsheeter

That sounds right, it is bounded and renders nothing.

qxliu76:
I've updated my fix: 
- don't add anything during layer index closure if numLayer is 0
- during subsetting the PaintColrLayers: set first layer index to 0 if numLayers is 0.

behdad:
Looks better. Thanks.

PR #5391, opened 2025-07-16
CMakeLists.txt + HB_HAVE_DIRECTWRITE=ON : linker error on Windows : 'unresolved external symbol __imp_DWriteCreateFactory'

chr-thien:
When using HB_HAVE_DIRECTWRITE=ON then we get linker error:

harfbuzz.cc.obj : error LNK2019: unresolved external symbol __imp_DWriteCreateFactory referenced in function "public: __cdecl hb_directwrite_global_t::hb_directwrite_global_t(void)" (??0hb_directwrite_global_t@@QEAA@XZ)

Added linking to dwrite.lib on Windows.


khaledhosny:
Thanks!

PR #5392, opened 2025-07-18
[ot-font] advances and v-origin cleanup

behdad:
For now, cleans up advances only. v-origin cleanup coming.


behdad:
This is complete now. Review is gonna be hard. But the essence is this final form of the new (plural) `get_v_origins` in ot-font:

https://github.com/harfbuzz/harfbuzz/blob/d4f3566bbf4d655a2d089898e43d360773c89bc1/src/hb-ot-font.cc#L607-L735

Thanks to the test suite, valgrind, and other tools, my confidence is quite high on this. It adds ~10kb to the lib .text size, which is just at the limit of what I'm comfortable doing for fast vertical shaping. Goes away, of course, when compiled without vertical.

I'll post some before/after numbers soon.

behdad:
UPDATE: These are not the correct numbers. I forgot to update `benchmark-font` to use the new API, so it was going through a lot of fallback... Will update with new numbers.

UPDATE 2:  Something phishy observed in results. Scratch this. I'll post new numbers after I figure out (after sleep).

Bottomline is this:

Vertical origin for glyf+vmtx fonts variable goes down from 118022us to 13730us, when queried glyphs don't hit the cache whatsoever.

With the cache (testing only gids 0-255), goes down from 180us to 6us!

For reference, the same workloads spend 10183us and 4us respectively, in _v-advances_ callback.

The large numbers are from NotoSansCJK-VF.ttf.ttc for full font (64k glyphs). The small numbers are for just first 256 glyphs of it.

behdad:
To make (up!) these numbers, `benchmark-font` now takes a `--max-glyphs 256` args that have to be specified before anything else. 256 is the size of our caches, so this code path will have minimal cache misses (once per glyph).

behdad:
Okay. Here's some real numbers.

The TrueType-flavored NotoSansCJK-VF.ttf.ttc. Note that while it has vmtx, it doesn't have VVAR (that's a mistake I believe https://github.com/notofonts/noto-cjk/issues/307), so for vertical advances we have to go to gvar still. Since it's a TT-flavored font (ie. without VORG), for vertical origin we have to go to vmtx+glyf and if variable, gvar, anyway. So that operation was the slowest. Also, the operation was NOT batched in HB, causing way too many branches, and it was implemented in terms of glyph extents, which I changed to just fetch phantom points instead.

With main, benchmarking the full 64k glyphs in the font:
```
-------------------------------------------------------------------------------------------------
Benchmark                                                       Time             CPU   Iterations
-------------------------------------------------------------------------------------------------
BM_Font/glyph_h_advances/NotoSansCJK-VF.ttf.ttc/ot           84.1 us         83.6 us         8171
BM_Font/glyph_h_advances/NotoSansCJK-VF.ttf.ttc/var/ot        859 us          854 us          820
BM_Font/glyph_v_advances/NotoSansCJK-VF.ttf.ttc/ot           86.4 us         85.9 us         7962
BM_Font/glyph_v_advances/NotoSansCJK-VF.ttf.ttc/var/ot      16206 us        16116 us           41
BM_Font/glyph_v_origin/NotoSansCJK-VF.ttf.ttc/ot             2674 us         2659 us          275
BM_Font/glyph_v_origin/NotoSansCJK-VF.ttf.ttc/var/ot       117669 us       117115 us            6
```
With this branch:
```
-------------------------------------------------------------------------------------------------
Benchmark                                                       Time             CPU   Iterations
-------------------------------------------------------------------------------------------------
BM_Font/glyph_h_advances/NotoSansCJK-VF.ttf.ttc/ot           61.8 us         61.5 us        11322
BM_Font/glyph_h_advances/NotoSansCJK-VF.ttf.ttc/var/ot        786 us          781 us          895
BM_Font/glyph_v_advances/NotoSansCJK-VF.ttf.ttc/ot           68.9 us         68.6 us        10057
BM_Font/glyph_v_advances/NotoSansCJK-VF.ttf.ttc/var/ot       6847 us         6809 us           90
BM_Font/glyph_v_origins/NotoSansCJK-VF.ttf.ttc/ot            2778 us         2763 us          250
BM_Font/glyph_v_origins/NotoSansCJK-VF.ttf.ttc/var/ot        7538 us         7491 us           93
```
We see all advance numbers have improved, though not by a factor of magnitude or anything. This is because this test always misses the cache lookup. The variable v-origins callback is particularly bad, standing at 120x slower than the non-variable h-advances callback.

Now if I run the test limiting to the first 256 glyphs of the font only, using:
```
perf/benchmark-font --max-glyphs 256 NotoSansCJK-VF.ttf.ttc --benchmark_filter='(advance|origin).*/ot'
```
with main:
```
-------------------------------------------------------------------------------------------------
Benchmark                                                       Time             CPU   Iterations
-------------------------------------------------------------------------------------------------
BM_Font/glyph_h_advances/NotoSansCJK-VF.ttf.ttc/ot          0.333 us        0.331 us      2137616
BM_Font/glyph_h_advances/NotoSansCJK-VF.ttf.ttc/var/ot      0.247 us        0.246 us      2844746
BM_Font/glyph_v_advances/NotoSansCJK-VF.ttf.ttc/ot          0.335 us        0.334 us      2092783
BM_Font/glyph_v_advances/NotoSansCJK-VF.ttf.ttc/var/ot       37.9 us         37.8 us        18583
BM_Font/glyph_v_origin/NotoSansCJK-VF.ttf.ttc/ot             8.73 us         8.69 us        80239
BM_Font/glyph_v_origin/NotoSansCJK-VF.ttf.ttc/var/ot          181 us          180 us         3913
```
with this branch:
```
-------------------------------------------------------------------------------------------------
Benchmark                                                       Time             CPU   Iterations
-------------------------------------------------------------------------------------------------
BM_Font/glyph_h_advances/NotoSansCJK-VF.ttf.ttc/ot          0.250 us        0.249 us      2827732
BM_Font/glyph_h_advances/NotoSansCJK-VF.ttf.ttc/var/ot      0.237 us        0.236 us      3025730
BM_Font/glyph_v_advances/NotoSansCJK-VF.ttf.ttc/ot          0.269 us        0.267 us      2637273
BM_Font/glyph_v_advances/NotoSansCJK-VF.ttf.ttc/var/ot      0.237 us        0.236 us      2964917
BM_Font/glyph_v_origins/NotoSansCJK-VF.ttf.ttc/ot           0.535 us        0.533 us      1310684
BM_Font/glyph_v_origins/NotoSansCJK-VF.ttf.ttc/var/ot       0.518 us        0.516 us      1358074
```
We see the variable v-advance is a lot faster, thanks to a cache being used now. Moreover, the v-origin callbacks (variable or not) are just 2x slower than the advance width/height callbacks. That's the cache working.

That should be enough to sell this branch, I hope.

behdad:
Looks like the OTF fonts have it, but not TTF?

behdad:
This makes shaping vertical Latin text with NotoSansCJK-VF 2x faster.

khaledhosny (approved):
LGTM as far as I’m concerned.

PR #5401, opened 2025-07-23
[repacker] Optimize LigatureSubstFormat1::split_subtables for lookups with large numbers of ligas.

garretrieger:
Fixes multiple instances of O(n^2) behaviour. ~20x speedup for test/fuzzing/fonts/clusterfuzz-testcase-minimized-hb-subset-fuzzer-5845846876356608


behdad:
LGTM in principle. Thanks.

PR #5403, opened 2025-07-23
[VARC] Graduate out of experimental

behdad:
I didn't remove it from `HB_NO_BORING_EXPANSION`; just out of `HB_EXPERIMENTAL_API`. I'm not sure what to do. for AVAR2, we removed it from both, making it survive anyone disabling boring-expansion using `HB_NO_BORING_EXPANSION`. The idea was that Apple shipped avar2, so it's not a boring-expansion project anymore...


behdad:
I removed HB_NO_BORING_EXPANSION completely.

PR #5406, opened 2025-07-24
[kbts] Streamline buffer management

behdad:


behdad:
I did this to see if preallocating more space helps. This PR doesn't do that.

PR #5407, opened 2025-07-24
[kbts] Clean up font initialization

behdad:


src/hb-kbts.cc
```diff
@@ -44,59 +44,54 @@
...
+  size_t memory_size;
   {
-    DEBUG_MSG (KBTS, face, "kbts_font creation failed");
-    _hb_kbts_shaper_face_data_destroy ((hb_kbts_face_data_t *) kb_font);
-    kb_font = nullptr;
-    goto fail;
+    unsigned scratch_size = kbts_ReadFontHeader (kb_font, data, blob_length);
+    void *scratch = hb_malloc (scratch_size);
+    memory_size = kbts_ReadFontData (kb_font, scratch, scratch_size);
+    hb_free (scratch);
```
khaledhosny, line 83: The idea was to avoid another allocation if possible, so `scratch` was re-used if it is of enough size. (that is what `kbts_FontFromFile()` does as well).

behdad: I saw that. But for the font I was testing, memory_size was half of scratch_size... Free then malloc of a smaller size is definitely fast. We don't need to hold onto too much memory.

khaledhosny: Makes sense.

PR #5414, opened 2025-07-25
Break up `hb-subset.cc` into smaller files

### LINKED ISSUE #5393: Building hb-subset.so is surprisingly slow
I was looking at distro-wide build profiles earlier and noticed that harfbuzz was taking a surprising amount of time to compile.

Using the great ninjatracing tool, it's easy to see why...

Image

`hb-subset.cc` takes a shocking amount of time to compile.

Two thoughts:
1) I note that the CMake build has an option to disable hb-subset, would that be accepted for the Meson build too?  Am I right in thinking that the subset library is quite a niche feature that eg a normal Linux desktop wouldn't actually need?
2) I'm guessing there's some giant tables in the code that mean it takes forever to compile. Could it be split up so it can be build in pieces in parallel?

behdad:
To fix #5393

I got started on this, but somehow the tests fail. I can't spot what I messed up...

Should do more. Should add `hb-subset-table-color.cc` and `hb-subset-table-other.cc`.


behdad:
> I got started on this, but somehow the tests fail. I can't spot what I messed up...

Fixed.

> Should do more. Should add `hb-subset-table-color.cc` and `hb-subset-table-other.cc`.

Done.

Still, takes a long time. Investigating.

behdad:
@rossburton Can you test this branch and produce a graph?

behdad:
> @rossburton Can you test this branch and produce a graph?

Done. With clang. Much better:

image

rossburton:
Nice :)

PR #5423, opened 2025-07-26
[WIP] Add lookup-level cache for some lookups

behdad:
Namely, for (Chain)ContextFormat3 that uses a lot of Coverage tables.

Code to come.


behdad:
Deadend. Was supposed to speed up Duployan, but didn't.

PR #5427, opened 2025-07-26
Use float for main font scaling

behdad:
I measured `h-advances` on Mac and I measure no performance difference whatsoever. Haven't tested on Linux yet.


behdad:
Note that this is more accurate. The previous code was approximating the scaling, using just an int mul & shift approximating a mul & div.

behdad:
Humm. I'm seeing very weird results. On Linux, this makes `h-advances` over 100% slower. Wow.

behdad:
Unless I'm measuring wrong on Linux, this is a deadend.

PR #5435, opened 2025-07-29
[ci] Add a script to extract release notes from NEWS file

khaledhosny:
Add a script to extract the release notes from the NEWS files and use them when creating GitHub releases, so that we don’t need to copy them manually.

Add also a test to make sure the NEWS file always have release notes for the current release.


behdad:
LGTM.

Issue #5439, opened 2025-08-03
Wrong shaping of font LinLibertine_G with frac=2 feature enabled


t-paul:
Originally reported in https://github.com/openscad/openscad/issues/6069 it appears the shaping of text showing fractions like 73/324" is wrong with the font LinLibertine_G and the font feature `frac=2` .

The font file is available via https://www.numbertext.org/linux/ and documentation at https://www.numbertext.org/linux/fontfeatures.pdf

<img width="832" height="84" alt="Image" src="https://github.com/user-attachments/assets/eca97796-45d5-497b-83f5-3eaf25004b4e" />

`hb-view` shows correct render with `frac=1` but with `frac=2` the lower number only shows the first digit.
<img width="1524" height="1496" alt="Image" src="https://github.com/user-attachments/assets/a99e4880-d81a-4cde-a70c-d8f1d1e052a2" />

Using Firefox to render the same text shows the correct fractional display, so the font itself might not be the cause of the issue.

<img width="1292" height="782" alt="Image" src="https://github.com/user-attachments/assets/d9c454d5-0689-46e9-84e4-38de34a7c058" />

khaledhosny:
(this is a Graphite font, and AFAIK, Firefox uses libgraphite directly and does not go through HarfBuzz, so this might be an issue with how we use libgraphite).

```
./util/hb-shape LinLibertineG/LinLibertine_R_G.ttf 73/324 --features=frac=2
[seven.superior=0@316,0+944|three.superior=1+-944|horizontalbar=2+2|three.inferior=3@0,-205+628,-205|two.inferior=4@0,-2147483648+628|four.inferior=5@0,-2147483648+628]
```
There are very large negative offsets. -2147483648 is -0x80000000, I wounder where is this comming from.

behdad:
Graphite returns negative advances sometimes, FWIW

khaledhosny:
Indeed. It is returning a negative vertical advance.

khaledhosny:
With:
```diff
diff --git a/src/hb-graphite2.cc b/src/hb-graphite2.cc
index c547134d5..e2acdffb1 100644
--- a/src/hb-graphite2.cc
+++ b/src/hb-graphite2.cc
@@ -266,7 +266,7 @@ _hb_graphite2_shape (hb_shape_plan_t    *shape_plan HB_UNUSED,
   gr_segment *seg = nullptr;
   const gr_slot *is;
[...]

PR #5440, opened 2025-08-03
[graphite2] Use signed int for current advance

khaledhosny:
Graphite can return negative advances, but we were storing current x and y advances as unsigned ints. The `curradvx` and `curradvy` variables are used only when calculating the respective offset, but the advance itself it always taken from Graphite directly, so using signed here makes no difference to how advances are set, only the offsets.

Fixes https://github.com/harfbuzz/harfbuzz/issues/5439


behdad:
LGTM. Thanks.

PR #5452, opened 2025-08-12
Hr buffer cache

behdad:


behdad:
Gives a 2% speedup on benchmarks.

PR #5455, opened 2025-08-14
[rust] no_std

behdad:
Donno what good this is for, but I was curious and did it.

Binary size is ~3kb smaller.

### DIFF
--- a/src/rust/lib.rs
+++ b/src/rust/lib.rs
@@ -1,3 +1,5 @@
+#![no_std]
+
 mod hb;
 use hb::*;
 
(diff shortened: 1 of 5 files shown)


src/rust/lib.rs
```diff
@@ -6,32 +8,37 @@ mod font;
...
         hb_free(ptr as *mut c_void);
     }
 }
 
 #[global_allocator]
 static GLOBAL: MyAllocator = MyAllocator;
+
+#[panic_handler]
+fn panic(_: &core::panic::PanicInfo) -> ! {
+    unsafe { core::hint::unreachable_unchecked() }
```
behdad, line 43: I'm surprised this didn't result in lots more size / perf changes.

PR #5474, opened 2025-08-17
harfbuzz.cc hb-shape

behdad:


behdad:
This works now:
```
$ c++ -O2 src/harfbuzz.cc util/hb-shape.cc  -o hb-shape -I src -DHB_NO_FEATURES_H `pkg-config --cflags --libs glib-2.0`
```

We should put it in the `BUILD.md` / `CONFIG.md` file.

behdad:
@JCash Anything else you suggest we add to `BUILD.md` or `CONFIG.md`?

PR #5490, opened 2025-08-21
[ot-math] Fix constant scaling when checking if the font is Cambria Math

eerii:
As discussed in #4653, the workaround for Cambria Math is not working since `get_constant` returns a scaled version of `displayOperatorMinHeight` and `delimitedSubFormulaMinHeight`. Replace with a direct lookup to `minHeight`.


eerii:
Hi @behdad, would it be ok for you to review this, since you reviewed the original patch? Thank you so much :)

PR #5491, opened 2025-08-21
Avoid UB if shrink_vector(0) is called on an already-empty vector

### LINKED ISSUE #5489: Seeing a UBSan failure in 11.4.2 (related to AAT/kerx)
Trying to update to 11.4.2 in gecko, we're seeing a ubsan failure 

> hb-vector.hh:447:33: runtime error: applying non-zero offset 18446744073709551264 to null pointer

that looks to be related to the kerx-accelerator:

https://treeherder.mozilla.org/logviewer?job_id=523358026&repo=try&task=WESfImblQ3WbLZMzUerKFg.0&lineNumber=1466-1483

(I don't currently have a reduced testcase, but can try to extract one if necessary.)

jfkthame:
Fixes #5489

### DIFF
--- a/src/hb-vector.hh
+++ b/src/hb-vector.hh
@@ -444,9 +444,9 @@ struct hb_vector_t
     if (!std::is_trivially_destructible<Type>::value)
     {
       unsigned count = length - size;
-      Type *p = arrayZ + length - 1;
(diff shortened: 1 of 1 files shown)


behdad:
Thanks.

Issue #5499, opened 2025-08-25
Subsetting problems with markFilteringSets


behdad:
This affects both hb-subset and fonttools subset.

There are two problems, one is an inefficiency of the final subset, the other is a bug that changes shaping behavior. It was introduced in 8ceed6ee72edd49163714e8f4ce072f104f58277.

Basically, what we do to subset the MarkFilteringSets table of GDEF is that we walk every set and intersect its contents with the glyphset, and only retain the ones that intersect. The inefficiency is that we should also look into GSUB/GPOS retained lookups and see which mark filtering sets they refer to, and only retain those. That is, `collect_used_mark_sets` must involve GSUB/GPOS as well, not just GDEF.

The second problem is in this code, introduced in the specified commit into `hb-ot-layout-common:Lookup::subset`:
```c++
    if (lookupFlag & LookupFlag::UseMarkFilteringSet) 
    { 
      const HBUINT16 &markFilteringSet = StructAfter<HBUINT16> (subTable); 
      hb_codepoint_t *idx; 
      if (!c->plan->used_mark_sets_map.has (markFilteringSet, &idx)) 
      { 
        abort(); 
        unsigned new_flag = lookupFlag; 
        new_flag &= ~LookupFlag::UseMarkFilteringSet; 
        out->lookupFlag = new_flag; 
      } 
      else 
      { 
        if (unlikely (!c->serializer->extend (out))) return_trace (false); 
        HBUINT16 &outMarkFilteringSet = StructAfter<HBUINT16> (out->subTable); 
        outMarkFilteringSet = *idx; 
      } 
    } 
```
the problem arises where I have put an abort. The thing is, a mark-filtering-set is defined in the spec as:
> If any lookup has the USE_MARK_FILTERING_SET flag set, then the Lookup header must include the markFilteringSet field and a MarkGlyphSets table must be present in GDEF table. The lookup must ignore any mark glyphs that are **not** in the specified mark glyph set; only glyphs in the specified mark glyph set are processed.

Emphasize mine. That is, if the mark-filtering-set is empty, we CANNOT just remove the `UseMarkFilteringSet` flag. We can only do so if instead we turn the `ignoreMark` flag on. Otherwise, a font that used to skip marks, will end up not skipping marks anymore.

behdad:
I hit this with the text sequence "J̣J" and Google Sans, whereas kerning will fail to apply between the two J's after subsetting.

khaledhosny:
[...]

PR #5502, opened 2025-08-25
[subset] Fix shaping behaviour when pruning empty mark sets

khaledhosny:
See https://github.com/harfbuzz/harfbuzz/issues/5499


behdad:
LGTM. Thank you.

PR #5518, opened 2025-08-31
meson: Fix fallback to WrapDB ICU wrap >= 77.1-3

bgilbert:
The wrap switched from the transitional `icu-uc = icuuc_dep` syntax to the recommended `dependency_names = icu-uc`, which apparently breaks our fallback path because we pass a `components:` argument to `dependency()` for CMake.  Use a separate `dependency()` invocation to look for a fallback.


khaledhosny:
I’m fine with ignoring cmake-built dependencies entirely, they keep complicated the build.

bgilbert:
Okay, updated.

khaledhosny:
While at it, can you also add the ICU wrap file (I didn't know it has one).

bgilbert:
Done!

khaledhosny:
Thanks.

PR #5520, opened 2025-09-01
[GPOS/kerx] Support attaching marks to bases after them

behdad:
See comments. Fuzzer found. I'm surprised this was not discovered before though. I hope I have not missed anything.


behdad:
This is hard to test since no kerx font I know hits this case, and we don't have easy ways to create kerx test fonts (fonttools doesn't support kerx machines).

dfrg:
The code looks reasonable to me

PR #5521, opened 2025-09-02
Create man pages with useful whatis entry

aelmahmoudy:
The whatis entry (the brief description found in the NAME section) for
manual pages created by help2man is of the form:                      
                                                                      
program - manual page for program                                     
                                                                      
This conveys no information about what the program is for and is repetitive.
The short description should contain brief information about what the 
program is for to aid in searching with apropos and similar programs.

### DIFF
--- a/util/meson.build
+++ b/util/meson.build
@@ -21,7 +21,7 @@ if not get_option('docs').disabled()
 else
   help2man = disabler()
 endif
-help2man_command = [help2man, '--help-option=--help-all', '--no-info', '--output=' + '@OUTPUT@', '@INPUT@']
+help2man_command = ['sh', '-c', help2man.full_path() + ' --help-option=--help-all --no-info -n "$(@INPUT@ --help | head -n 4 | tail -n 1)" --output=$0 @INPUT@', '@OUTPUT@']
 help2man_env = environment()
 help2man_env.set('HB_UTIL_HELP_VERBOSE', '1')
 


behdad:
I like to fix this, but the approach looks rather fragile to me.

util/meson.build
```diff
@@ -21,7 +21,7 @@ if not get_option('docs').disabled()
 else
   help2man = disabler()
 endif
-help2man_command = [help2man, '--help-option=--help-all', '--no-info', '--output=' + '@OUTPUT@', '@INPUT@']
+help2man_command = [help2man, '--help-option=--help-all', '--no-info', '-n "$(' + '@OUTPUT@' + ' --help | head -n 3 | tail -n 1)"', '--output=' + '@OUTPUT@', '@INPUT@']
```
behdad: In my testing, hb-view prints the information we want on the 4th line, not 3. What's yours?

aelmahmoudy:
@behdad said:
> In my testing, hb-view prints the information we want on the 4th line, not 3. What's yours?

That's true. Fixed it

Another fix was I wrapped call to help2man inside 'sh -c' to do the 
shell expansion

Please note that I changed ***@***.***@` to ***@***.***@.1`, 
for some reason @OUTPUT@ doesn't expand with me, I don't understand why

aelmahmoudy:
I did a little workaround to be able to pass  @OUTPUT@ instead of @INPUT@.1 to help2man's --output switch

behdad:
Thanks.

Issue #5511, opened 2025-08-26
[subset] preprocess + HVAR increases HVAR table size in some cases.


garretrieger:
Reproduction:
```
$util/hb-subset --font-file=test/subset/data/fonts/NotoSansJP-VF.subset.otf --output-file=a.otf --unicodes=61,63,5009 --drop-tables+=DSIG,BASE --drop-tables-=sbix --retain-gids
$ util/hb-subset --font-file=test/subset/data/fonts/NotoSansJP-VF.subset.otf --output-file=b.otf --unicodes=61,63,5009 --drop-tables+=DSIG,BASE --drop-tables-=sbix --retain-gids --preprocess

$ ttx -l a.otf b.otf | grep HVAR

    HVAR  0x00160043        50       544
    HVAR  0x00260047        54       648
```

When preprocessing is enabled the output HVAR is 4 bytes larger.

I tracked this down to the HVAR subsetting code increasing the encoding width in the adv width map during subsetting pass as part of preprocessing (so here we're not doing any actual subsetting yet, all gids are retained). The change happens in the "index_map_subset_plan_t::remap()" function. This looks at the number of bits needed to represent the inner values and finds a case where 3 bits is needed, versus 1 bit in the original table.

garretrieger:
Some more information. The input file has null offset to the adv width mapping:

```
0001
0000 # major, minor

00000014
00000000 # adv width map
00000000
00000000
```

This means it's just the identity mapping. After running through the preprocessing step, the adv width mapping gets serialized instead of being left null:

```
0001
0000 # major, minor

00000024
00000014 # adv width map
00000000
00000000

# adv width map
00 # format
12 # entry format (byte width = 2, inner bit count = 3)
0006 # mapping count = 6
```

This sets the byte width to 2. Then finally later on once the subsetting operation is applied the byte width is retained and not shrunk so the output table ends up with width 2 instead of 1.

So I believe there's a couple of problems that can be fixed:
1. If the mapping we're going to serialize is the identity mapping, the mapping table can be omitted.
2. The serialization step should shrink the byte width if it's possible. It looks like the relevant code, will only increase the size (https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-ot-var-hvar-table.hh#L120)

behdad:
Sound right to me. Thanks for tracking it down.

qxliu76:
[...]

PR #5523, opened 2025-09-02
[subset] fix HVAR encoding width

qxliu76:
Try to fix encoding width issue in #5511.
Mainly 2 changes:
1. use pointers instead of `Null(DeltaSetIndexMap)` to avoid API calls on DeltaSetIndexMap
2. use the smallest outer bit count

Omit identity mapping needs more thinking so not taken care of in this PR.


behdad:
Thanks. I'll try to do a review soon. Is this something we need to fix in fonttools?

qxliu76:
No, we don't need to fix fonttools, now HB returns exactly the same result with fonttools(except for the part that fonttools will choose the smaller one between varstore and advmap when avdmap is identity mapping). 
Previously we have a bigger encoding width but ttx is identical so this issue is not spotted.

PR #5525, opened 2025-09-03
[rust] Debug CI fail

behdad:


behdad:
Looks like libharfbuzz_rust.a has LLVM bitcode in it because we are building with lto enabled, but the final linker is GNU ld.bfd and doesn't understand that. I don't know why I don't hit this locally, or why we didn't hit it before.

behdad:
Filed https://github.com/rust-lang/rust/issues/146145

PR #5545, opened 2025-09-12
[instancer/varstore] Add a heuristic to speed up optimization

behdad:
@garretrieger I like to merge this if we can establish that the test rebase to be done is acceptable. Thanks.


behdad:
LGTM. Thanks.

PR #5547, opened 2025-09-12
[subset] Introduce HB_SUBSET_FLAGS_RETAIN_NUM_GLYPHS.

garretrieger:
Used in conjunction with retain gids, when set the num glyphs from the input font will be preserved. Empty glyphs will be inserted as necessary to maintain the num glyphs value. Put under experimental for now.


behdad:
LGTM. Just wait till I get 11.5.0 out and we can merge.

PR #5549, opened 2025-09-12
[instancer/varstore] Only optimize instanced varstore incrementally

behdad:
See comments. This basically wipes of the varstore optimization costs out of the profiles, which saves eg. 25% total time in GoogleSansFlex instantiation, with practically no file size downside.

Tests fail. Need investigation.

Also fixes https://github.com/harfbuzz/harfbuzz/issues/5550

### DIFF
--- a/src/hb-ot-var-common.hh
+++ b/src/hb-ot-var-common.hh
@@ -1877,6 +1877,8 @@ struct item_variations_t
      * is a compromise between optimization and performance and practically
      * works very well. */
 
+    // This produces slightly smaller results in some cases.
+    encoding_objs.qsort ();
 
     /* main algorithm: repeatedly pick 2 best encodings to combine, and combine them */
     using item_t = hb_priority_queue_t<combined_gain_idx_tuple_t>::item_t;
--- a/src/hb-ot-layout-common.hh
+++ b/src/hb-ot-layout-common.hh
@@ -2333,6 +2333,11 @@ struct delta_row_encoding_t
    * needed for this region */
   struct chars_t : hb_vector_t<uint8_t>
   {
+    int cmp (const chars_t& other) const
+    {
+      return as_array ().cmp (other.as_array ());
+    }
+
     hb_pair_t<unsigned, unsigned> get_width ()
     {
       unsigned width = 0;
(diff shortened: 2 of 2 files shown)


behdad:
My confidence in this change is a bit low. @qxliu76 Any chance you can do a detailed review, as well as check the failures to see if the new output is acceptable, ie. looks correct as well as not being too much larger than the previous varstore? Thanks.

behdad:
I cleaned this up a bit more and my confident has gone up. Just needs rebasing the tests.

behdad:
So, this PR moves from a full optimizer for varstore during instancing to an incremental optimizer. I think the diff on font data binary size increments is quite acceptable:
image

Going to merge soon, but still would appreciate a review. @garretrieger @qxliu76

qxliu76:
Thank you! I'll try to review it(hopefully I can understand the algorithm:))

src/hb-ot-var-common.hh
```diff
@@ -1953,9 +1951,6 @@ struct item_variations_t
       encodings.push (std::move (encoding_objs.arrayZ[i]));
     }
 
-    /* sort again based on width, make result deterministic */
```
qxliu76, line 1977: iirc, we might need to keep this sorting to match fonttools' result. Fonttools also does sorting twice: see: [1st,](https://github.com/fonttools/fonttools/blob/03a3c8ed9e7ab35b5f219a6d25ee8081107ff2bb/Lib/fontTools/varLib/varStore.py#L601) [2nd](https://github.com/fonttools/fonttools/blob/03a3c8ed9e7ab35b5f219a6d25ee8081107ff2bb/Lib/fontTools/varLib/varStore.py#L649)

qxliu76:
LGTM, I have only one question about the final sorting. Thanks!

qxliu76:
Also a few micro optimizations in this PR are really appreciated:)

src/hb-ot-var-common.hh
behdad: Looks like I put back the sort after all.

Note that we don't match fonttools anymore, since the algorithm is different now.

The nondeterminism was originally coming from Python hash tables. Our hash table is deterministic, but good to not rely on it. But then again, in the new code I don't think we were relying on the hashtable order at all.

At any rate. There's a sort back in already.

qxliu76: oh yeah you're right, thanks very much!

PR #5552, opened 2025-09-14
Speedup map copy

behdad:


behdad:
FWIW, the last commit here sped up GoogleSansFlex subsetting by around 10%.

PR #5558, opened 2025-09-15
Fix invalid reinterpret_cast

safocl:
https://github.com/harfbuzz/harfbuzz/blob/ebcc20d085c305b1a6022f5fcfff766f9b27f8c5/src/hb-atomic.hh#L86
and the same below
with:
https://github.com/harfbuzz/harfbuzz/blob/b33f6fd213870323ecf0ae76d5f742e33af66489/src/hb-common.cc#L45
(only one example)

equivalent of this pseudo-code:
```cpp
int i;
reinterpret_cast<std::atomic<int>*>(&i)->some_atomic_op(...);
```

reinterpret_cast from non-atomic object to atomic object is not valid.

https://eel.is/c++draft/atomics.types.generic#general-3
>[Note 2: The representation of an atomic specialization need not have the same size and alignment requirement as its corresponding argument type. — end note]

https://eel.is/c++draft/basic.compound#5
>5 Two objects a and b are pointer-interconvertible if
>(5.1) they are the same object, or
>(5.2) one is a union object and the other is a non-static data member of that object ([class.union]), or
>(5.3) one is a standard-layout class object and the other is the first non-static data member of that object or any base class subobject of that object ([class.mem]), or
>(5.4) there exists an object c such that a and c are pointer-interconvertible, and c and b are pointer-interconvertible.
>If two objects are pointer-interconvertible, then they have the same address, and it is possible to obtain a pointer to one from a pointer to the other via a reinterpret_cast ([expr.reinterpret.cast]).

objects of types `int` and `std::atomic<int>` are not _pointer-interconvertible_ (an object of type `int` is not first non-static data member of `std::atomic<int>`)

https://eel.is/c++draft/basic.lval#11
>11 An object of dynamic type T-obj is type-accessible through a glvalue of type T-ref if T-ref is similar ([conv.qual]) to:
>(11.1) T-obj,
>(11.2) a type that is the signed or unsigned type corresponding to T-obj, or
>(11.3) a char, unsigned char, or std::byte type.
>If a program attempts to access ([defns.access]) the stored value of an object through a glvalue through which it is not type-accessible, the behavior is undefined

The original code resulted in **undefined behavior**.

p.s. Code formatting is done using the provided `.clang-format` file;
headers `hb.hh` and `hb-meta.hh` were recursively dependent on `hb-atomic.hh` ([gave the same error as HERE](https://github.com/clangd/clangd/issues/1407))

other:
https://stackoverflow.com/questions/8749038/how-to-use-stdatomic-efficiently

### DIFF
--- a/src/hb-atomic.hh
+++ b/src/hb-atomic.hh
@@ -149,7 +149,7 @@ struct hb_atomic_t
   hb_atomic_t () = default;
   constexpr hb_atomic_t (T v) : v (v) {}
   constexpr hb_atomic_t (const hb_atomic_t& o) : v (o.get_relaxed ()) {}
-  constexpr hb_atomic_t (hb_atomic_t&& o) : v (o.get_relaxed ()) { o.set_relaxed({}); }
+  constexpr hb_atomic_t (hb_atomic_t&& o) : v (o.get_relaxed ()) { o.set_relaxed ({}); }
 
   hb_atomic_t &operator= (const hb_atomic_t& o) { set_relaxed (o.get_relaxed ()); return *this; }
   hb_atomic_t &operator= (hb_atomic_t&& o){ set_relaxed (o.get_relaxed ()); o.set_relaxed ({}); return *this; }


safocl:
Perhaps it would be wiser to just use `std::atomic` instead?

behdad:
Thank you. I was waiting for this report to come in. :)
I'll take a look to see if this can be further simplified.

> Perhaps it would be wiser to just use `std::atomic` instead?

IIRC the problem is that it would require linking to C++ standard library, which we have managed to avoid so far.

behdad:
MSVC bots are failing.

src/hb-atomic.hh
```diff
@@ -32,8 +32,8 @@
 #ifndef HB_ATOMIC_HH
 #define HB_ATOMIC_HH
 
-#include "hb.hh"
-#include "hb-meta.hh"
+// #include "hb.hh"
```
behdad: Any reason for these? We always include `hb.hh` or another `.hh` file first.

safocl: headers hb.hh and hb-meta.hh were recursively dependent on hb-atomic.hh (https://github.com/clangd/clangd/issues/1407)

safocl: https://github.com/harfbuzz/harfbuzz/blob/43284d1f4d076b50a08c560b69a11672c94fce0c/src/hb.hh#L555

safocl:
> IIRC the problem is that it would require linking to C++ standard library, which we have managed to avoid so far.

It seems that in other parts of the code the standard library is already used, which requires linking with the implementation.

src/hb-atomic.hh
behdad: That is intentional. I'll read the clangd issue, but this has been there forever, and works with every compiler. It's a simple C-style single-inclusion that is welldefined and intentional.

behdad:
> > IIRC the problem is that it would require linking to C++ standard library, which we have managed to avoid so far.
> 
> It seems that in other parts of the code the standard library is already used, which requires linking with the implementation.

We deliberately only use parts that are header-only and don't require the library to be linked. I know it's probably nowhere defined as such, but that has worked for us so far.

safocl:
didn't keep track of const )
signature is:
```cpp
bool std::atomic<T>::compare_exchange_weak( T& expected, T desired, std::memory_order success, std::memory_order noexcept;
```

cannot be converted from `const T*` to `T*&`

https://github.com/harfbuzz/harfbuzz/blob/ebcc20d085c305b1a6022f5fcfff766f9b27f8c5/src/hb-atomic.hh#L197
The _C-style cast_ here performs a `const_cast` and may yield ub if the data is indeed constant.

safocl:
```cpp
template< class T >
struct atomic_ref;
```
https://en.cppreference.com/w/cpp/atomic/atomic_ref.html
atomic_ref can help here, I think, but it is only available since C++20

safocl:
https://github.com/harfbuzz/harfbuzz/blob/ebcc20d085c305b1a6022f5fcfff766f9b27f8c5/src/hb-algs.hh#L744
https://en.cppreference.com/w/cpp/atomic/atomic.html
> std::atomic is neither copyable nor movable.

https://eel.is/c++draft/atomics#types.int-1

>     atomic(const atomic&) = delete;
>     atomic& operator=(const atomic&) = delete;
>     atomic& operator=(const atomic&) volatile = delete;

behdad:
I see swapping atomic happening to:
```
../src/hb-bit-set.hh:49:13:   required from here
   49 |     hb_swap (a.last_page_lookup, b.last_page_lookup);
      |     ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
that can be fixed to remove the swap and set both `last_page_lookup`s to 0 it seems.

safocl:
> I see swapping atomic happening to:
> 
> ```
> ../src/hb-bit-set.hh:49:13:   required from here
>    49 |     hb_swap (a.last_page_lookup, b.last_page_lookup);
>       |     ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```
> 
> that can be fixed to remove the swap and set both `last_page_lookup`s to 0 it seems.

i do not implement explicitly _copy_ and _move_ constructors and _assignment-operators_

behdad:
> > I see swapping atomic happening to:
> > ```
> > ../src/hb-bit-set.hh:49:13:   required from here
> >    49 |     hb_swap (a.last_page_lookup, b.last_page_lookup);
> >       |     ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> > ```
> >
> > that can be fixed to remove the swap and set both `last_page_lookup`s to 0 it seems.
> 
> i do not implement explicitly _copy_ and _move_ constructors and _assignment-operators_

Just calling `set_relaxed` works, no?

behdad:
> > > I see swapping atomic happening to:
> > > ```
> > > ../src/hb-bit-set.hh:49:13:   required from here
> > >    49 |     hb_swap (a.last_page_lookup, b.last_page_lookup);
> > >       |     ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> > > ```
> > > 
> > > that can be fixed to remove the swap and set both `last_page_lookup`s to 0 it seems.
> > 
> > 
> > i do not implement explicitly _copy_ and _move_ constructors and _assignment-operators_
> 
> Just calling `set_relaxed` works, no?

In fact, leaving `last_page_lookup` untouched should also work.

behdad:
> ``
> 
> https://github.com/harfbuzz/harfbuzz/blob/ebcc20d085c305b1a6022f5fcfff766f9b27f8c5/src/hb-atomic.hh#L197
> 
> 
> The _C-style cast_ here performs a `const_cast` and may yield ub if the data is indeed constant.

Looks like the `const` is not relied on. Code compiles fine without it. I don't know why it was there in the first place.

behdad:
> > 
> 
> Looks like the `const` is not relied on. Code compiles fine without it. I don't know why it was there in the first place.

I think it precedes when we decided to declare all our atomic vars as `mutable` at callsites. I think you can remove the const.

src/hb-atomic.hh
```diff
@@ -196,9 +207,23 @@ struct hb_atomic_t<T *>
...
+  void swap( hb_atomic_t &o ){
+    T *p = o.get_acquire ();
+    o.set_relaxed (get_acquire ());
+    set_relaxed (p);
+  }
+
   std::atomic<T *> v = nullptr;
 };
 
+namespace std
```
behdad: We typically just do a
```
friend void swap (..., ...)
```
in the struct itself, and ADL seems to pick it up. See other types eg. `hb_vector_t`. Thanks.

src/hb-atomic.hh
```diff
@@ -80,27 +79,11 @@ _hb_atomic_ptr_impl_cmplexch (const void **P, const void *O_, const void *N)
 
 #include <atomic>
 
+#define STL_ATOMIC_IMPL
```
behdad: Please prefix this with `HB_`. Otherwise looks good and I'll merge when back to computer.

src/hb-atomic.hh
```diff
@@ -80,27 +79,11 @@ _hb_atomic_ptr_impl_cmplexch (const void **P, const void *O_, const void *N)
 
 #include <atomic>
```
behdad, line 80: Should this move down inside the #ifdef ?

behdad: Oh nevermind.

behdad:
Thank you and thanks for your patience.

safocl:
> Thank you and thanks for your patience.

oh, thanks for your tips!

pavel-yurkin:
cool)

PR #5559, opened 2025-09-15
fix clang-format repeated flags

safocl:
some flags were listed twice with opposite meanings


.clang-format
```diff
@@ -13,10 +13,8 @@ BraceWrapping:
   SplitEmptyFunction: false
   AfterClass: true
   AfterControlStatement: true
-  AfterEnum: false
   AfterFunction: true
   AfterNamespace: false
-  AfterStruct: true
```
behdad, line 19: We actually need the true both after enum and struct. Thanks.

PR #5562, opened 2025-09-15
[repacker] make repacker object ids stable.

garretrieger:
Prior to this change the object id for a vertex in the repacker graph is it's position in the topological ordering. As a result after each sorting the object ids change. This decouples the topological sorting from the object id assignments. A separate ordering list is retained. This simplifies handling around object ids which no longer needs to account for them shifting, and improves performance by eliminating the step of reassigned link object ids after each sort.

Benchmarks for Noto Nastaliq Urdu show a 5-10% speedup.


behdad:
Thanks. I'm not much familiar with this code, but the perf boost SGTM!

PR #5563, opened 2025-09-15
Instancer gvar consolidate mallocs

behdad:


behdad:
This significantly reduces iup allocations and speeds up GoogleSansFlex instancing with `--optimize` by over 25%.

PR #5566, opened 2025-09-16
Alloc pool

behdad:


behdad:
This significantly sped up instantiating GoogleSansFlex and RobotoFlex. By 25%.

@garretrieger Do you think you have a use for this in the repacker?

garretrieger:
> This significantly sped up instantiating GoogleSansFlex and RobotoFlex. By 25%.
> 
> @garretrieger Do you think you have a use for this in the repacker?

The repacker typically doesn't do much allocation since it's mostly just shuffling things around. We do use a persistent scratch pool during the reordering part that we swap with to avoid allocations.

qxliu76:
Nice optimization, thank you!

qxliu76:
I'm not much familiar with repacker code yet(just started looking into detailed code). It seems we have some vector allocs in the subtable splitting code, does it make sense to use this alloc pool? one example [here](https://github.com/harfbuzz/harfbuzz/blob/3ace0323195aead5aa27caab4097ae19b10648f4/src/graph/gsubgpos-graph.hh#L131).

garretrieger:
Yes some of those case might be worth switching over. That particular one is probably not worth it since the amount of work being done in that loop is quite small relative to lower levels of splitting. One place that probably makes sense and is easy to do is the new node buffer:

https://github.com/harfbuzz/harfbuzz/blob/main/src/graph/graph.hh#L1664

This is used to allocate space needed when we add new nodes into the graph.

behdad:
What's a heavy workload for the repacker?

garretrieger:
For most cases I'm guessing time is spent primarily in the distance calc and sorting methods:

https://github.com/harfbuzz/harfbuzz/blob/main/src/graph/graph.hh#L608
and
https://github.com/harfbuzz/harfbuzz/blob/main/src/graph/graph.hh#L1508

Might be worth grabbing some profiles though to confirm if that's still the case

PR #5573, opened 2025-09-22
Instancer map alloc from pool

behdad:


behdad:
While this reduces number of mallocs by a large percentage on GoogleSansFlex, it doesn't actually show up any speed difference. So, discarding it again for the sake of simplicity.

PR #5577, opened 2025-09-23
[buffer-message] Disable "more" messages by default

behdad:
This was not suppoosed to be turned on by default, as it has a performance overhead. Instead, tools like Crowbar should define HB_BUFFER_MESSAGE_MORE to 1 when building HarfBuzz.


behdad:
cc @simoncozens FYI.

behdad:
This also affects `hb-shape --verbose` output, but I think that's fine.

Issue #5601, opened 2025-10-07
Undefined behavior when using union


safocl:
https://github.com/harfbuzz/harfbuzz/blob/3f70b6987830bba1e4922cad03028cdd9d78b3a1/src/hb-aat-layout-kerx-table.hh#L909
and
https://github.com/harfbuzz/harfbuzz/blob/3f70b6987830bba1e4922cad03028cdd9d78b3a1/src/hb-aat-layout-kerx-table.hh#L873
and
https://github.com/harfbuzz/harfbuzz/blob/3f70b6987830bba1e4922cad03028cdd9d78b3a1/src/hb-aat-layout-kerx-table.hh#L865

are produce undefined behavior:
```cpp
unsigned int get_type () const { return u.header.coverage & u.header.SubtableType; } // access to u.header
return_trace (c->dispatch (u.format0, std::forward<Ts> (ds)...)); // access to u.format0
```
an access to `u.header` and a subsequent access to `u.formaN`, but either `u.formaN` or `u.header` is not an active member of the union.
https://godbolt.org/z/GKc7Kj5Y3

> In a standard-layout union with an [active member](https://eel.is/c++draft/class.union#def:active,union_member) of struct type T1, it is permitted to read a non-static data member m of another union member of struct type T2 provided m is part of the common initial sequence of T1 and T2; the behavior is as if the corresponding member of T1 were nominated[.](https://eel.is/c++draft/class.mem#general-30.sentence-1)

> If a program attempts to access ([[defns.access]](https://eel.is/c++draft/defns.access)) the stored value of an object through a glvalue through which it is not type-accessible, the behavior is undefined[.](https://eel.is/c++draft/basic.lval#11.sentence-2)

https://eel.is/c++draft/class.mem#general-30
https://eel.is/c++draft/class.prop#10
https://eel.is/c++draft/class.union#general-2
https://eel.is/c++draft/basic.lval#11.sentence-2

but if you wrap KerxSubTableHeader in a _struct_, then the _behavior_ becomes _defined_:
```cpp
template <typename KernSubTableHeader>
struct KerxSubTableFormatGeneralProtected
{
  protected:
  KernSubTableHeader		header;
};

  union {
  KerxSubTableFormatGeneral<KerxSubTableHeader> formatGeneral;
  KerxSubTableFormat0<KerxSubTableHeader>	         format0;
  KerxSubTableFormat1<KerxSubTableHeader>	         format1;
[...]

PR #5602, opened 2025-10-10
Wrap format in unions in a struct

behdad:
WIP. To fix https://github.com/harfbuzz/harfbuzz/issues/5601


safocl:
Overall, I think this is a normal fix -- everything seems to be fine (from the point of view of the implementation of the fix).

behdad:
I did more using an AI agent. Not saying this includes all instances, but a good start point.

behdad:
> I did more using an AI agent.

FWIW, I tried doing this with gemini CLI and it failed to make coherent compilable changes. ChatGPT's codex CLI managed to produce the commits in this PR, correctly.

behdad:
I think this is ready to be merged. Any further instances can be fixed later.

PR #5606, opened 2025-10-13
Fix documentation of offsets in text glyph serialization format

Hoolean:
Hello! This is just a small patch found incidentally, which only touches the docs.

Offsets are serialized when only one of `x` and `y` is non-zero too:

https://github.com/harfbuzz/harfbuzz/blob/7c2105a52e6b50e448064e82c760dad0b5f4bdb9/src/hb-buffer-serialize.cc#L302-L303

(this matches the deserialization behaviour too, and so it was just a documentation issue)


behdad:
Thanks. Might read more natural to say "if either x or y is not zero"?

Hoolean:
Yes, I ummed and ahhed about whether an 'either or' would read as inclusive or exclusive... that might only mean I've been thinking in terms of booleans too much today though 😄

Issue #5607, opened 2025-10-13
`kern`-table fallback behaviour differs from documentation


Hoolean:
Hello!

We had a head-scratcher: [the shaping and buffer output docs](https://harfbuzz.github.io/shaping-and-shape-plans.html#shaping-buffer-output) document that [the legacy `kern`-table](https://learn.microsoft.com/en-us/typography/opentype/spec/kern) is only used when `GPOS` is not present, however, in reality it can be triggered more liberally by `HB_NO_LEGACY` being undefined and the current language system containing no `kern`-feature:

https://github.com/harfbuzz/harfbuzz/blob/7c2105a52e6b50e448064e82c760dad0b5f4bdb9/src/hb-ot-shape.cc#L153-L185

This means that the fallback occurs in more common scenarios, such as in multi-script fonts compiled by [ufo2ft](https://github.com/googlefonts/ufo2ft/) that contain scripts such as Devanagari, which by default will will have a `dist` feature but no `kern` feature registered for their spacing.

Hoolean:
(thanks to @khaledhosny for helping identify this, who also notes that [the documentation became outdated when the behaviour was changed in 2018](https://github.com/harfbuzz/harfbuzz/commit/e655fd38cf20eefb1c071a52282a4caccb6f08ea))

behdad:
cc @jfkthame 

One way is to skip legacy kern if `dist` feature is available as well. Another is to disable legacy `kern` in Indic and USE shapers.

Hoolean:
Documenting some of the initial motivation here: (Common, Common) pairs often legitimately appear in `dist`, but these are also often preferred candidates for the legacy kern-table, leading to them being double-applied under the existing strategy

Hoolean:
(It is not possible to remove such pairs from the legacy kern-table without depriving older environments of them entirely)

behdad:
We already have a boolean `fallback_position` in each shaper, that is only true for default, hebrew, and arabic. I think we can reuse the same flag to enable legacy kern possibly.

behdad:
Please test https://github.com/harfbuzz/harfbuzz/pull/5609

Hoolean:
Thanks Behdad, that patch looks good to me ✅

We prepared a test matrix of:
- With `kern`-table, without `kern`-table
- With `GPOS`-table, without `GPOS`-table
- With a Latin script, a Devanagari script, and an unset script

[...]

PR #5609, opened 2025-10-13
[shape] Disable `kern` table in most shapers

behdad:
Only default, Arabic, and Hebrew keep it.

Fixes https://github.com/harfbuzz/harfbuzz/issues/5607

Needs documentation update.

### DIFF
--- a/src/hb-ot-shaper-hangul.cc
+++ b/src/hb-ot-shaper-hangul.cc
@@ -427,7 +427,7 @@ const hb_ot_shaper_t _hb_ot_shaper_hangul =
   HB_TAG_NONE, /* gpos_tag */
   HB_OT_SHAPE_NORMALIZATION_MODE_NONE,
   HB_OT_SHAPE_ZERO_WIDTH_MARKS_NONE,
-  false, /* fallback_position */
+  true, /* fallback_position */
 };
 
 


Hoolean:
Thanks Behdad, this looks great -- to confirm the extent of the change, would we expect the legacy kerntable to no longer be applied for every HarfBuzz `dist`-enabled script [listed here by ufo2ft](https://github.com/googlefonts/ufo2ft/blob/26710588b616e72d259ea1156778b983ddfea1b3/Lib/ufo2ft/featureWriters/kernFeatureWriter.py#L33-L40)?

jfkthame:
> Only default, Arabic, and Hebrew keep it.

Given that `dist` appears to be intended primarily for Indic-related scripts (per https://learn.microsoft.com/en-gb/typography/opentype/spec/features_ae#tag-dist, "Used for Indic or other Brahmi-derived scripts"), I wonder if we should keep legacy `kern` fallback enabled for Hangul as well?

Aside from that question, this seems reasonable to me.

behdad:
> > Only default, Arabic, and Hebrew keep it.
> 
> Given that `dist` appears to be intended primarily for Indic-related scripts (per https://learn.microsoft.com/en-gb/typography/opentype/spec/features_ae#tag-dist, "Used for Indic or other Brahmi-derived scripts"), I wonder if we should keep legacy `kern` fallback enabled for Hangul as well?

Sounds doable. I'll just reuse the existing bool, so it will re-enable fallback mark positioning for Hangul. I don't think that would be an issue?

PR #5610, opened 2025-10-14
[test] Add test for harfrust#278

khaledhosny:
Add test case for https://github.com/harfbuzz/harfrust/pull/278


behdad:
Lgtm. Thanks

PR #5611, opened 2025-10-15
[shape] Test fallback to `kern` table under different shapers

Hoolean:
More specifically, this asserts that the `kern` table is avoided for certain shapers, and that we avoid double-spacing in the presence of `dist` for the same.

This is a companion PR to the fix itself, which this is the base branch targeted.

(The new tests all pass for me locally but fail under `--shaper=ot`, but I'll wait to understand the implications of that before attempting any further changes)


behdad:
> (The new tests all pass for me locally but fail under `--shaper=ot`, but I'll wait to understand the implications of that before attempting any further changes)

I'm not sure what this means.

Hoolean:
Whoops, I'd typo'd the description; re-created at #5612, as requested, with that fixed too

PR #5612, opened 2025-10-16
[shape] Test fallback to `kern` table under different shapers

Hoolean:
More specifically, this asserts that the `kern` table is avoided for certain shapers, and that we avoid double-spacing in the presence of `dist` for the same.

This is a companion PR for the fix at #5609.

The new tests pass for me locally, apart from when run with `--shaper=ft`. Do I need to recompile FreeType, or should I only run with `--shaper=ot`?


Hoolean:
Same result on the CI; should I add `@font-funcs=ot,fontations` to the tests, or does it failing with `ft` indicate a real-world issue?

behdad:
I think we're double kerning on FT. I'll fix.

khaledhosny:
> Same result on the CI; should I add `@font-funcs=ot,fontations` to the tests, or does it failing with `ft` indicate a real-world issue?

It seems to be failing the `# Only legacy kern table` case for Devanagari (i.e. applying the legacy `kern` table) which is not something that should be sensitive to font functions.

Hoolean:
Indeed, I've just had a closer look at the test output and I think the exact issue is that we are still falling back to the legacy `kern`-table for Deva _in the absence of_ `GPOS`; when `GPOS` is present, we correctly only use it and avoid double kerning.

Hoolean:
(under FT)

Hoolean:
Oop, my apologies for the confusion -- I'm still learning HarfBuzz' codebase -- but I meant `@shapers` instead of `@font-funcs` above

behdad:
What's happening is that we are falling back to the font-funcs-based kerning, which only the FT font-funcs implement.

The reason this was done in the first place was such that shaping of non-SFNT fonts with kerning (Type1? I don't know if it has kerning) would work transparently through FreeType. But I think we don't care about those anymore.

I suggest deprecating the font-funcs kerning functions. Or we can just make the code not call that in this case, or not at all. I'll take a look.

jfkthame:
Type 1 fonts may indeed have kerning, specified in the AFM file -- so, separately from the PFA or PFB glyph outlines. (I don't know offhand whether FreeType handles that, but it seems possible, at least.)

The preferred solution here would be to avoid calling font-funcs-based kerning if we've decided not to apply legacy 'kern', I think.

behdad:
> The preferred solution here would be to avoid calling font-funcs-based kerning if we've decided not to apply legacy 'kern', I think.

See https://github.com/harfbuzz/harfbuzz/pull/5614

behdad:
Folded this into https://github.com/harfbuzz/harfbuzz/pull/5614
Waiting for bots to pass to merge.

Issue #5597, opened 2025-10-02
HB_BUFFER_MESSAGE_MORE changes shaping result


behdad:
> OK, this is super weird:
> 
> ```
> $ hb-shape master_ttf/NotoSansBengali-Regular.ttf 'স্কুল'
> [s-beng.half=0+745|baphala-beng.alt4=2@-580,125+0|ka-beng.part=2@-263,98+0|headline-beng.250=2@-250,0+0|uMatra-beng=2@-295,-5+0|la-beng.blws=4@-689,-214+0]
> ```
> 
> Notice here that la-beng has taken below form `la-beng.blws`. I wonder how that happened, let's trace the shaping:
> 
> ```
> $ hb-shape master_ttf/NotoSansBengali-Regular.ttf 'স্কুল' -V
> ...
> trace: end lookup 193 feature 'rclt'    buffer: [s-beng.half=0|baphala-beng.alt4=2|ka-beng.part=2|headline-beng.250=2|uMatra-beng=2|la-beng=4]
> trace: end table GSUB script tag 'bng2' buffer: [s-beng.half=0|baphala-beng.alt4=2|ka-beng.part=2|headline-beng.250=2|uMatra-beng=2|la-beng=4]
> ```
> 
> Harfbuzz bug! @behdad 

 _Originally posted by @simoncozens in [#32](https://github.com/notofonts/bengali/issues/32#issuecomment-3355582031)_

I can reproduce; enabling `HB_BUFFER_MESSAGE_MORE` makes the output to change with `--trace` vs without.

behdad:
This is the line causing the diff:
https://github.com/harfbuzz/harfbuzz/blob/a790c38b782f9d8e6f0299d2837229e5726fc669/src/OT/Layout/GSUB/Sequence.hh#L111

I have not understood why yet.

behdad:
Oof. This took some time to figure out.

PR #5617, opened 2025-10-19
[match-backtrack] Fix syllable-setting logic

behdad:
If accidentally backtrack_len() was equal idx, we were setting syllable, even though we want to match backtrack, which shouldn't match syllable.

I'm surprised how this was not exposed before.

Fixes https://github.com/harfbuzz/harfbuzz/issues/5597

### DIFF
--- a/src/hb-ot-layout-gsubgpos.hh
+++ b/src/hb-ot-layout-gsubgpos.hh
@@ -523,9 +523,16 @@ struct skipping_iterator_t
 #endif
   void reset (unsigned int start_index_)
   {
+    // For GSUB forward iterator
     idx = start_index_;
     end = c->buffer->len;
     matcher.syllable = start_index_ == c->buffer->idx ? c->buffer->cur().syllable () : 0;
+  }
+  void reset_back (unsigned int start_index_, bool from_out_buffer = false)
+  {
+    // For GSUB backward iterator
+    idx = start_index_;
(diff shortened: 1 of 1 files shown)


behdad:
@jfkthame This bug meant that most of the times we were matching backtrack limited to within the syllable. Do you think any further changes is warranted, or leave this fixed as it was originally intended?

jfkthame:
> @jfkthame This bug meant that most of the times we were matching backtrack limited to within the syllable. Do you think any further changes is warranted, or leave this fixed as it was originally intended?

Tricky one. In principle the fix seems desirable, but it does feel like there's some risk of fonts getting unwanted behavior changes because they (perhaps inadvertently) depended on the limitation.

Would be nice to run the wikipedia corpus against a wide range of fonts, and check for changed behavior, but unfortunately I don't have a setup to do that... :-(

behdad:
> Would be nice to run the wikipedia corpus against a wide range of fonts, and check for changed behavior, but unfortunately I don't have a setup to do that... :-(

I agree that's good to test. I'll see what I can do with Noto and Google Fonts collections.

behdad:
I'll write some code to try to detect fonts that might trigger this bug: iterate all per-syllable features in the font and look for ChainContext GSUB lookups that recurse to a Ligature or Multiple subst lookups and have any backtrack in the rules.

PR #5619, opened 2025-10-19
Add test-match-backtrack.py

behdad:
To test https://github.com/harfbuzz/harfbuzz/issues/5618

@jfkthame How does this look?

Issue #5618, opened 2025-10-19
Test a range of Indic / USE fonts for match-backtrack bug


behdad:
See https://github.com/harfbuzz/harfbuzz/pull/5617

behdad:
Follow in https://github.com/harfbuzz/harfbuzz/pull/5619

PR #5635, opened 2025-11-03
Match backtrack/lookahead per syllable

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/pull/5619#issuecomment-3480493524
Fixes https://github.com/harfbuzz/harfbuzz/issues/5618


behdad:
We should add tests. Ideally both for Indic shaper and USE.

cc @lianghai

behdad:
> We should add tests. Ideally both for Indic shaper and USE.

Done for Indic.

behdad:
@lianghai Is there a way you can reproduce a test font (a modification of any Noto font) that would trigger the USE shaper and try to ligate things across syllable boundary and see it fail to match?

I'll probably merge this as is, since it's a bugfix.

lianghai:
> @lianghai Is there a way you can reproduce a test font (a modification of any Noto font) that would trigger the USE shaper and try to ligate things across syllable boundary and see it fail to match?

I’m happy to create a minimal repro from https://github.com/notofonts/sinhala/blob/main/sources/NotoSerifSinhala.glyphs, but I’m not sure anymore what should be reproduced. None of these rules seem to have any effect under the sinh OTL script:

Screenshot 2025-11-03 at 20 11 52

> රකවර with the WIP repro [NotoSerifSinhala-Regular.otf.zip](https://github.com/user-attachments/files/23313720/NotoSerifSinhala-Regular.otf.zip)

Screenshot 2025-11-03 at 20 15 20

When I created the Bangla repro (https://github.com/notofonts/bengali/issues/32#issuecomment-3367682393), it was not simply ligating or matching across syllable boundaries.

behdad:
@lianghai Thanks. We just want to make sure, with USE, that rules from the foundational features don't match across syllables. For Indic I produced a test already.

lianghai:
Here you are, a minimal Sinhala (USE) test font: [NotoSerifSinhala-Regular.otf.zip](https://github.com/user-attachments/files/23343413/NotoSerifSinhala-Regular.otf.zip). Source: [NotoSerifSinhala.glyphs.zip](https://github.com/user-attachments/files/23343431/NotoSerifSinhala.glyphs.zip), derived from [notofonts/sinhala/sources/NotoSerifSinhala.glyphs](https://github.com/notofonts/sinhala/blob/7f19bfcf53f7a239aa7ed3115bc9b2b1b1a5fcc3/sources/NotoSerifSinhala.glyphs).

Note that I managed to make it actually _fail_ in [12.1.0](https://github.com/harfbuzz/harfbuzz/releases/tag/12.1.0):

```console
> hb-shape NotoSerifSinhala-Regular.otf "දාද"
[daavowelsinh=0+532|dasinh=2+532]
```

Expected:

```console
[dasinh=0+532|aavowelsignsinh=0+379|dasinh=2+532]
```

It has only the follow GSUB:

```fea
languagesystem sinh dflt;

feature akhn {
    # This rule shouldn't match, because a syllable cluster boundary exists right after aavowelsignsinh:
    sub dasinh' aavowelsignsinh' dasinh by daavowelsinh;
} akhn;
```

behdad:
That's great. Thanks Liang.

PR #5640, opened 2025-11-11
[meson] Update wrap files

Beryesa:
FreeType 2.14 loads HarfBuzz on runtime dynamically to simplify the circular dependencies, which is nice.

### DIFF
--- a/subprojects/cairo.wrap
+++ b/subprojects/cairo.wrap
@@ -1,10 +1,10 @@
 [wrap-file]
-directory = cairo-1.18.2
-source_url = https://www.cairographics.org/releases/cairo-1.18.2.tar.xz
-source_filename = cairo-1.18.2.tar.xz
-source_hash = a62b9bb42425e844cc3d6ddde043ff39dbabedd1542eba57a2eb79f85889d45a
-source_fallback_url = https://github.com/mesonbuild/wrapdb/releases/download/cairo_1.18.2-1/cairo-1.18.2.tar.xz
-wrapdb_version = 1.18.2-1
+directory = cairo-1.18.4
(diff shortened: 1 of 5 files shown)


khaledhosny:
I don’t see how simply updating the wrap changes anything, since we already pass `default_options: ['harfbuzz=disabled']` to freetype dependency:
https://github.com/harfbuzz/harfbuzz/blob/677f0b367d0ad250824fc2f7a00267c2260fe68c/meson.build#L110

Beryesa:
I was trying to fix some fallback ci warnings in gtk (and pango) that descended through harfbuzz wrap files  
The pulled harfbuzz version in gtk didn't have that change yet, but I wanted to bump here first since pango pulls from main and that could help  

Sorry for the short description, feel free to close if not useful for the upstream defaults :P

khaledhosny:
I don’t mind updating the wrap files, I just don’t see how this in itself has any effect on the circular dependency.

Beryesa:
gtk pulled old harfbuzz with freetype2 enabled, meson found the freetype2 from nearby pango's wraps (that had older freetype2 which requires harfbuzz) and for that pango itself pulls harfbuzz, from main, to satisfy freetype2 and so on.  

So I'm trying to bump in both places to the versions that already solved it to clean up; starting here _(then gtk pulls pango from main which will get harfbuzz from main)_ and walking up the line if that makes sense _(maybe doesn't 😅)_ but practically it doesn't fail since these are only absolute fallbacks, however it looked really messy in the android build that enabled it - sorry for bothering :P

khaledhosny:
I’m not convinced, but also there is nothing that prevents merging this PR.

PR #5643, opened 2025-11-14
Try to fix fuzzer found memory leak.

garretrieger:
For: https://oss-fuzz.com/testcase-detail/6083012753424384

Doesn't yet fix the issue, but cleans up some potential mishandling of in_error() vectors related to the leaked vector.


src/graph/gsubgpos-graph.hh
```diff
@@ -179,6 +179,9 @@ struct Lookup : public OT::Lookup
       if (new_sub_tables.in_error ()) return false;
       if (!new_sub_tables) continue;
       hb_pair_t<unsigned, hb_vector_t<unsigned>>* entry = all_new_subtables.push ();
+      if (all_new_subtables.in_error()) {
+        continue;
+      }
```
behdad: An alternative approach is to use `all_new_subtables.push (hb_pair (..., ...))` instead.

src/graph/split-helpers.hh
```diff
@@ -49,7 +49,9 @@ hb_vector_t<unsigned> actuate_subtable_split (Context& split_context,
     if (id == (unsigned) -1)
     {
       new_objects.reset ();
-      new_objects.allocated = -1; // mark error
+      if (!new_objects.in_error()) {
+        new_objects.allocated = -1; // mark error
```
behdad: Setting `allocated` is definitely not a good idea. We should add a method to vector for whatever this means. There's `set_error()`.

behdad: Setting allocated to -1 can cause us to lose the allocated vector and not free it. Because:
```c++
  bool is_owned () const 
  { 
    return allocated != 0 && allocated != -1; 
  } 
```
and
```c++
  void fini () 
  { 
    if (is_owned ()) 
    { 
      shrink_vector (0); 
      hb_free (arrayZ); 
    } 
    init (); 
  } 
```
whereas calling `set_error()` would do the right thing.

behdad:
This fixes it for me:
```diff
diff --git a/src/graph/split-helpers.hh b/src/graph/split-helpers.hh
index 61fd7c2d2..278c8d137 100644
--- a/src/graph/split-helpers.hh
+++ b/src/graph/split-helpers.hh
@@ -49,7 +49,7 @@ hb_vector_t<unsigned> actuate_subtable_split (Context& split_context,
     if (id == (unsigned) -1)
     {
       new_objects.reset ();
-      new_objects.allocated = -1; // mark error
+      new_objects.set_error ();
       return new_objects;
     }
     new_objects.push (id);
@@ -58,7 +58,7 @@ hb_vector_t<unsigned> actuate_subtable_split (Context& split_context,
   if (!split_context.shrink (split_points[0]))
   {
     new_objects.reset ();
-    new_objects.allocated = -1; // mark error
+    new_objects.set_error ();
   }
 
   return new_objects;
```

We should still audit the codepaths you touched in this PR.

PR #5644, opened 2025-11-14
Force gen-harfbuzzcc.py to generate POSIX-style paths to avoid annoying git diffs

Demonese:
Windows uses backslashes as path separators, causing annoying git diffs every time Meson is run.

I also checked the other c/c++ source code and confirmed that the paths in `#include` directives are in POSIX style.

`pathlib` was introduced into Python in version 3.4, and the minimum Python version used in CI and GitHub Actions is 3.10, so I it's safe to use the `pathlib` module.

```diff
PS D:\Project\github-fork\harfbuzz> git diff
diff --git a/src/harfbuzz-subset.cc b/src/harfbuzz-subset.cc
index f6a5a319c..7acbe4e9f 100644
--- a/src/harfbuzz-subset.cc
+++ b/src/harfbuzz-subset.cc
@@ -1,5 +1,5 @@
-#include "OT/Var/VARC/VARC.cc"
-#include "graph/gsubgpos-context.cc"
+#include "OT\Var\VARC\VARC.cc"
+#include "graph\gsubgpos-context.cc"
 #include "hb-aat-layout.cc"
 #include "hb-aat-map.cc"
 #include "hb-blob.cc"
diff --git a/src/harfbuzz.cc b/src/harfbuzz.cc
index 471942c55..c0e2acc0b 100644
--- a/src/harfbuzz.cc
+++ b/src/harfbuzz.cc
@@ -1,4 +1,4 @@
-#include "OT/Var/VARC/VARC.cc"
+#include "OT\Var\VARC\VARC.cc"
 #include "hb-aat-layout.cc"
 #include "hb-aat-map.cc"
 #include "hb-blob.cc"
```

### DIFF
--- a/src/gen-harfbuzzcc.py
+++ b/src/gen-harfbuzzcc.py
@@ -2,7 +2,7 @@
 
 "This tool is intended to be used from meson"
 
-import os, sys, shutil
+import os, sys, shutil, pathlib
 
 if len (sys.argv) < 3:
 	sys.exit (__doc__)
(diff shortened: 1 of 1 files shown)


khaledhosny:
Thanks!

PR #5655, opened 2025-11-26
[beyond-64k] Update to latest proposal

behdad:
This branch implements:
  - TTC header versions 1.1/2.1
  - cmap format 15, preferred over format 14
  - MAXP
  - GLYF/LOCA, cubic outlines, 24-bit components
  - HHEA/HMTX and VHEA/VMTX
  - GVAR
  - VORG version 2
  - COLR PaintGlyph2
  - Coverage/ClassDef extended formats
  - GSUB/GPOS extended subtables and version 1.2 headers
  - GDEF version 1.4
  - BASE format 4
  - JSTF version 1.1


behdad:
This is now ready for review. The main missing part from the proposed spec is the `DMAP` table, which is NOT a beyond-64k addition.

behdad:
So there was one spec divergence I needed to make, to make merging NotoSans fonts work: spec doesn't widen LookupList-to-Lookup offsets in the extended LookupList2 field and keeps them 16bit, which defeats the whole purpose. So I widened them to 32bit and merge succeeds quickly with the fontTools.merge tool.

Here's a merged NotoSans font with all scripts except for emoji & CJK:

https://drive.google.com/file/d/13pK4xKRheLEuUhQB2xvkS6pqZSSB7_iO/view

```
Unicode count: 26984
Glyph count: 106958
```

```
Table information:

Tag	Size
------------
GDEF    57051 bytes
GLYF 13001268 bytes
GPOS  1587490 bytes
GSUB  1660634 bytes
HHEA       38 bytes
HMTX   427830 bytes
LOCA   427836 bytes
MAXP       33 bytes
OS/2       96 bytes
cmap   115950 bytes
cvt       300 bytes
fpgm     3605 bytes
gasp        8 bytes
head       54 bytes
name     2568 bytes
post       32 bytes
prep      261 bytes
```

garretrieger (approved):

src/hb-ot-layout-gsubgpos.hh
anthrotype, line 2194: glyph ID is hardcoded to uint16 here, shouldn't it use 24-bit when beyond-64k?

```suggestion
      HBUINT16 d;
      typename Types::HBUINT d;
```

anthrotype, line 3353: same here for ChainRule, HBUINT16 should be Types::HBUINT I think. Otherwise every subsequent field in the serialized record gets misaligned and hb-subset may produce a corrupt output for beyond-64k fonts with glyph-keyed ChainContextPos/Subst rules

behdad:
@copilot resolve the merge conflicts in this pull request

PR #5658, opened 2025-11-28
[logo] Remove the background

khaledhosny:
Replace the white background with transparency. The PNG version is too large (1.2M), but the HEIF version is only 6K larger than the original JPEG file.


khaledhosny:
Apparently HEIF files are not supported.

PR #5659, opened 2025-11-28
[logo] Remove the background

khaledhosny:
Replace the white background with transparency. The PNG version was too large (620K), but scaling it down to 512px made it even smaller than the current JPEG file (160K vs 171K).


behdad:
Looks good! Can you do for other projects as well? Thanks.

PR #5698, opened 2026-01-02
[perf/benchmark-shape] Add Gulzar

khaledhosny:


behdad:
Thanks. Before I resisted because of it's size. But I'm fine with it now.

PR #5703, opened 2026-01-07
Build fix for avar table with HB_NO_AVAR2

drott:


src/hb-ot-var-avar-table.hh
```diff
@@ -456,6 +456,8 @@ struct avar
 
     OT::ItemVariationStore::destroy_cache (var_store_cache);
     return true;
+#else
+    return false;
```
behdad, line 460: This always returns false even for avar1 fonts, no?

qxliu76:
Thanks! I have a patch that fixes the return value for avar1 fonts: https://github.com/harfbuzz/harfbuzz/pull/5705

PR #5705, opened 2026-01-07
[subset] avar mapping: fix return value for avar1 fonts

qxliu76:


behdad:
I wonder if we should return false if font is v2 and we don't compile v2 support. Do you know what the implications would be? Would it fail the subsetter at least?

qxliu76:
> I wonder if we should return false if font is v2 and we don't compile v2 support.

Returning false would fail subsetter. If we don't compile v2 support, subsetter just skips v2 processing step, I think that's expected.

behdad:
> > I wonder if we should return false if font is v2 and we don't compile v2 support.
> 
> Returning false would fail subsetter. If we don't compile v2 support, subsetter just skips v2 processing step, I think that's expected.

Humm. I feel like failing is better than silently producing a wrong font. WDYT?

qxliu76:
Oh yeah that makes sense, I'll make a change. I thought that when NO_AVAR2 is enabled, user knows that v2 data will be dropped:)

drott:
Thanks for the right fix, and sorry I landed the previous one a bit prematurely.

PR #5716, opened 2026-01-20
[subset] fix padding in gvar table

qxliu76:
Add necessary padding if original font is using long format and subsetted font is using short format


src/hb-ot-var-gvar-table.hh
behdad, line 474: We use `clear=false` here. So the padding bytes are uncleared. Either change false to true, or better, explicitly write 0 for the padding bytes.

behdad:
Looks good with comment. Thanks.

src/hb-ot-var-gvar-table.hh
```diff
@@ -480,7 +489,10 @@ struct gvar_GVAR
 
     /* This ordering relative to the shared tuples array, which puts the glyphVariationData
        last in the table, is required when HB_SUBSET_FLAGS_IFTB_REQUIREMENTS is set */
-    char *subset_data = c->serializer->allocate_size<char> (subset_data_size, false);
+    bool add_padding = org_is_long_offset && (!long_offset);
+    if (org_is_long_offset && long_offset)
+      subset_data_size -= padding_size;
+    char *subset_data = c->serializer->allocate_size<char> (subset_data_size, true);
```
behdad: Okay I see you did the true here. Fine with me.

src/hb-ot-var-gvar-table.hh
```diff
@@ -448,10 +448,19 @@ struct gvar_GVAR
     if (it->first == 0 && !(c->plan->flags & HB_SUBSET_FLAGS_NOTDEF_OUTLINE))
       it++;
     unsigned int subset_data_size = 0;
+    bool org_is_long_offset = is_long_offset ();
```
behdad: Do we actually need the `org_is_long_offset` stuff? Can't we always count padding?

qxliu76:
I think I fixed all issues now. Thanks!

behdad:
Actually would you please make it not clear the allocation and zero the padding byte? This code is super optimized.

qxliu76:
> Actually would you please make it not clear the allocation and zero the padding byte? This code is super optimized.

sure, will do

behdad:
Much cleaner. Thanks. Please merge when bots are happy.

PR #5718, opened 2026-01-23
[gvar] Fix padding size calcs

behdad:
The previous state of is_long_offset() is irrelevant here.

Fixes https://oss-fuzz.com/testcase-detail/5920156651159552


behdad:
This can happen with fuzzed fonts mostly, but not in reality (that `is_long_offset()` is false but `long_offset` is true).

behdad:
> This can happen with fuzzed fonts mostly, but not in reality (that `is_long_offset()` is false but `long_offset` is true).

Actually it can happen with any small font if IFTB requirements flag is on.

behdad:
@qxliu76 I'm merging, but please verify the logic. Thanks.

qxliu76:
Thanks for the fix!

PR #5719, opened 2026-01-23
fix fuzzer issue: https://oss-fuzz.com/testcase-detail/6005602106277888

qxliu76:
Do not use pointers that may point outside of the allocated memory region


behdad:
Thanks.

PR #5733, opened 2026-02-06
OpenType font dependency graph support

skef:
This PR introduces an API to extract a dependency graph from an OpenType font. I wrote the vast majority of the code added to src by hand, but more recently I've been using AI coding help. In `src` the AI impact is mostly limited to the handling of contextual rules, but the test code was written by combination of AI and myself and the added documentation is pretty obviously AI-flavored. 

Given my approach on the last parts of the implementation, quality is primarily test-verified. The hb-depend-closure-parity checker in test/fuzzing compares a closure generated from the dependency graph with one generated by the subset `closure` path. If the graph is under-approximated, that is an error. If it is "unexpectedly" over-approximated, that is also an error. _Expected_ over-approximation can be reported with `-r`. Two flags on the edge of a graph warn of potential over-approximation in making use of that edge in a closure. The flags are reasonably well-documented. I have only tested extensively on fonts in the HarfBuzz repo but will work to get more testing done now that I've made the parity checker deterministic across platforms. 

I have not yet added unit tests for closure_sets because the only appropriate font in test/api/fonts is NotoNastaliqUrdu-Regular.ttf, which takes 18 seconds to generate a dependency graph for. (This is an extreme outlier.) Gulzar-Regular.ttf would be a good candidate but it is currently only in `perf/fonts` and since it is on the larger side I didn't want to just copy it there. 

**Organizational Note**: I wrote this code as part of my participation in the Web Fonts Working Group, out of concern that a strategy of repeated closure operations would not be sufficient for creating a high-quality and performant IFT encoder. I've tried to follow HarfBuzz repo style and approaches to make this close to something mergable, and therefore maintainable. I want to be clear, however, that this work implies no obligation on the part of HarfBuzz to merge or even review this code. @garretrieger has done a preliminary, informal review and suggested this draft PR might be helpful at this time.


skef:
I fixed the CI problems that I could given where the builds got to.

skef:
I took a stab at fixing the documentation error

src/hb-depend.h
```diff
@@ -43,7 +43,7 @@ HB_BEGIN_DECLS
  *
  * <warning>Highly experimental API. Subject to change.</warning>
  *
- * Since: REPLACEME
+ * Since: 12.4.0
```
khaledhosny: Should be `XSince: REPLACEME` to avoid the build error. The actual version is updated at release time when API is stabilized.

docs/harfbuzz-docs.xml
```diff
@@ -118,6 +118,7 @@
       <index id="api-index-full"><title>API Index</title><xi:include href="xml/api-index-full.xml"><xi:fallback /></xi:include></index>
       <index id="deprecated-api-index"><title>Index of deprecated API</title><xi:include href="xml/api-index-deprecated.xml"><xi:fallback /></xi:include></index>
 
+      <index id="api-index-12-4-0"><title>Index of new symbols in 12.4.0</title><xi:include href="xml/api-index-12.4.0.xml"><xi:fallback /></xi:include></index>
```
khaledhosny: This is only done at release time when we know the release number.

src/hb-depend.h
```diff
@@ -43,7 +43,7 @@ HB_BEGIN_DECLS
  *
  * <warning>Highly experimental API. Subject to change.</warning>
  *
- * Since: 12.4.0
+ * Since: REPLACEME
```
khaledhosny: **X**Since

skef: I'm digging back through the git history to try to figure out the specific handling of a REPLACEME version. I suspect I've done one or two things prematurely that need to be backed out.

skef:
OK, that last commit make the docs build correctly on my local machine. (Should probably have been trying that from the start ...)

skef:
Oh, and I think it's probably fine, even desirable, to remove the print function from the API.

skef:
One other thing on this: The code is currently inconsistent when it comes to self-referring glyphs. The GSUB code keeps those entries but the COLR code removes them. This *arguably* makes sense because of the distinct nature of those COLR self-references, but we should probably decide whether they make sense to have or to avoid and do that everywhere.

behdad:
Thanks Skef. I'm reviewing this now. High-level notes:

- Instead of `HB_DEPEND_API`, we one a `HB_NO_SUBSET_DEPEND`, that gets triggered in `hb-config.hh` by the `#ifndef HB_EXPERIMENTAL_API` branch as well as `HB_LEAN`. This also means that you shouldn't need any build system changes. This is small enough of a piece of code that a meson option is overkill.

- Most of the doesn't need to gate the implementation on the above macro. All of the `depend` methods can stay in the compiled code. Just gate the actual public API in the .cc file to not compile if not requested.

- As hinted above, I believe this should move to the subset library only. The implementation can stay as `hb_depend*` internally, but all public API should be prefixed with `hb_subset_depend`.

- The `print` functionality should definitely be removed. If possible, implement the same in an uninstalled `util/hb-subset-depend` utility that builds and prints the graph using only public API. This harness would also allow for performance & memory consumption measurement and optimizations easily.

- The `get_glyph_entry` should take a struct to fill on, instead of multiple out params.

- I wonder if `lookup_glyph` and `lookup_set` would be nicer names than `get_glyph_entry` and `get_set_from_index`.

- `ligature_set` and `context_set` should probably be renamed to `ligature_set_index` and `context_set_index` respectively.

- The depend-data as a name doesn't convey much information. Should it be renamed to depend-edge perhaps?

I'll do a line-by-line review now. Please bear with me.

src/OT/Color/COLR/COLR.hh
```diff
@@ -219,6 +220,44 @@ struct hb_colrv1_closure_context_t :
   {}
 };
 
+#ifdef HB_DEPEND_API 
+
+struct hb_colrv1_depend_context_t :
+       hb_dispatch_context_t<hb_colrv1_depend_context_t, bool>
+{
+  template <typename T>
+  return_t dispatch (const T &obj)
```
behdad, line 234: What's the return value supposed to mean?    Success?

src/OT/Color/COLR/COLR.hh
```diff
@@ -219,6 +220,44 @@ struct hb_colrv1_closure_context_t :
...
+    nesting_level_left++;
+    return true;
+  }
+  static return_t default_return_value () { return false; }
+
+  const COLR* get_colr_table () const
+  { return reinterpret_cast<const COLR *> (base); }
+
+  public:
+  const void *base;
```
behdad: Humm what is base?

behdad: Okay I see `get_colr_table()` above. I think you should keep a reference to the COLR table directly, or keep the face and access `face->table.COLR` as needed.

src/OT/Color/COLR/COLR.hh
```diff
@@ -398,6 +447,10 @@ struct NoVariable
 
 struct ColorStop
 {
+#ifdef HB_DEPEND_API
+  void dependv1 (hb_colrv1_depend_context_t * c) const {}
```
behdad, line 450: Why descend into the `ColorLine` at all?

src/OT/Color/COLR/COLR.hh
```diff
@@ -938,6 +1030,9 @@ struct PaintSweepGradient
 // Paint a non-COLR glyph, filled as indicated by paint.
 struct PaintGlyph
 {
+#ifdef HB_DEPEND_API
+  void dependv1 (hb_colrv1_depend_context_t* c) const;
+#endif
   void closurev1 (hb_colrv1_closure_context_t* c) const;
```
behdad: I wonder why these few out-of-line methods don't seem to have an `HB_EXTERNAL` and still work fine without leaking the symbol names. Can you please clean up after them? I know it's pre-existing code.

src/OT/Color/COLR/colrv1-depend.hh
```diff
@@ -0,0 +1,135 @@
...
+
+  const BaseGlyphList &baseglyph_list = colr_table->get_baseglyphList ();
+  (&baseglyph_list+baseglyph_paintrecord->paint).dispatch (c);
+}
+
+template <template<typename> class Var>
+HB_INTERNAL void PaintTransform<Var>::dependv1 (hb_colrv1_depend_context_t* c) const
+{
+  (this+src).dispatch (c);
+  (this+transform).dependv1 (c);
```
behdad: Transform, ColorLine, etc don't affect the depend graph, so can't we just not descend into them?

src/OT/Layout/GSUB/Ligature.hh
```diff
@@ -38,6 +38,37 @@ struct Ligature
...
   }
 
+#ifdef HB_DEPEND_API
+  void depend (hb_depend_context_t *c, hb_codepoint_t first) const
+  {
+    // Build the complete ligature set upfront before adding any edges
+    hb_set_t complete_ligset;
+    complete_ligset.add(first);
+    for (const auto &source : component)
+      complete_ligset.add(source);
```
behdad: I think you can just initialize the set with an iterator.

src/OT/Layout/GSUB/Ligature.hh
```diff
@@ -38,6 +38,37 @@ struct Ligature
...
+
+    hb_codepoint_t ligset = c->depend_data->new_ligature_set(complete_ligset);
```
behdad: `ligset_idx`?

src/OT/Layout/GSUB/Ligature.hh
```diff
@@ -38,6 +38,37 @@ struct Ligature
...
+
+    // Track whether any edge using this ligset was actually added
+    bool any_added = false;
+
+    // Now add all edges with the complete, immutable set
+    if (c->depend_data->add_gsub_lookup (first, c->lookup_index, ligGlyph, ligset))
+      any_added = true;
+
+    + hb_iter (component)
+    | hb_apply ([&] (const hb_codepoint_t &source) {
```
behdad: Why is this called `source` instead of eg. `target`?

src/OT/Layout/GSUB/ReverseChainSingleSubstFormat1.hh
```diff
@@ -67,6 +67,68 @@ struct ReverseChainSingleSubstFormat1
   bool may_have_non_1to1 () const
   { return false; }
 
+#ifdef HB_DEPEND_API
+  bool depend (hb_depend_context_t *c) const {
+    // Filter by intersects and parent_active_glyphs like closure does (lines 83-95)
```
behdad: Remove line numbers please.

src/OT/Layout/GSUB/ReverseChainSingleSubstFormat1.hh
```diff
@@ -67,6 +67,68 @@ struct ReverseChainSingleSubstFormat1
...
+    for (unsigned int i = 0; i < count; i++)
+    {
+      hb_set_t back_glyphs;
+      (this+backtrack[i]).intersect_set (*c->glyphs, back_glyphs);
+      if (back_glyphs.get_population () == 1)
+        context_glyphs.add (back_glyphs.get_min ());
+      else if (back_glyphs.get_population () > 1)
+      {
+        hb_codepoint_t set_idx = c->depend_data->find_or_create_context_set (back_glyphs);
+        context_glyphs.add (0x80000000 | set_idx);
```
behdad: This 0x80000... business bothers me most. Can we use a macro in the public API for it at least?

src/OT/Layout/GSUB/ReverseChainSingleSubstFormat1.hh
```diff
@@ -67,6 +67,68 @@ struct ReverseChainSingleSubstFormat1
   bool may_have_non_1to1 () const
   { return false; }
 
+#ifdef HB_DEPEND_API
+  bool depend (hb_depend_context_t *c) const {
```
behdad: So some of the `depend`s return void, others bool. What's the deal?

src/hb-depend-data.hh
```diff
@@ -0,0 +1,567 @@
...
+ *   another immediately consumes it as an "intermediate"). So a glyph matching
+ *   the coverage might not actually persist at that position when closure
+ *   traverses the rule. These edges may cause over-approximation.
+ *
+ * HB_DEPEND_EDGE_FLAG_FROM_NESTED_CONTEXT (0x02):
+ *   Edge created from a lookup called within another contextual lookup (nested
+ *   context). The outer context's requirements are not preserved on this edge,
+ *   so it may over-approximate.
+ */
+#define HB_DEPEND_EDGE_FLAG_FROM_CONTEXT_POSITION 0x01
```
behdad: Enum flags please. Check other enums.

src/hb-depend-data.hh
```diff
@@ -0,0 +1,567 @@
...
+  uint8_t flags;
+};
+
+/**
+ * edge_key_t:
+ *
+ * Key structure for edge deduplication. Combines source glyph with edge record.
+ * Uses the record's equality and hash operators for comparison.
+ */
+struct edge_key_t {
```
behdad: Some `hb_` or more prefix please. `hb_depend_edge_key_t` sounds better if not too much hassle.

src/hb-depend-data.hh
```diff
@@ -0,0 +1,567 @@
...
+/**
+ * hb_glyph_depend_record_t:
+ *
+ * Internal structure holding all dependency records for a single glyph.
+ * Contains a vector of hb_depend_data_record_t entries, indexed sequentially
+ * starting from 0.
+ */
+struct hb_glyph_depend_record_t {
+  hb_vector_t<hb_depend_data_record_t> dependencies;
+};
```
behdad, line 168: Should this simply be `using hb_depend_glyph_record_t = hb_vector<...>`?

Note the name change also, should start with `hb_depend`.

src/hb-depend-data.hh
```diff
@@ -0,0 +1,567 @@
...
+  hb_lookup_feature_record_t (hb_lookup_feature_record_t &&o) : full(o.full),
+                                                                fv_indexes(std::move(o.fv_indexes)) {}
+  hb_lookup_feature_record_t& operator = (const hb_lookup_feature_record_t &o)
+  {
+    full = o.full;
+    fv_indexes = o.fv_indexes;
+    return *this;
+  }
+
+  bool full {true};
```
behdad: I think we use `bool full = true` syntax, which I find more readable.

src/hb-depend-data.hh
```diff
@@ -0,0 +1,567 @@
...
+};
+
+/**
+ * hb_lookup_feature_record_t:
+ *
+ * Internal structure tracking which features are associated with a lookup.
+ * Used during GSUB dependency analysis to record the feature tags that
+ * activate each lookup.
+ */
+struct hb_lookup_feature_record_t {
```
behdad: Any change this name can be changed to something non-generic?

Also, `{` goes on a new line typically.

src/hb-depend-data.hh
```diff
@@ -0,0 +1,567 @@
...
+      *layout_tag = d.layout_tag;
+      *ligature_set = d.ligature_set;
+      *context_set = d.context_set;
+      if (flags) *flags = d.flags;
+      return true;
+    }
+    return false;
+  }
+
+  bool get_set_from_index(hb_codepoint_t index, hb_set_t *out)
```
behdad: This should simply return a `const hb_set_t *` pointing to the internal set. You can also make the internal sets immutable if you really care about that.

src/hb-depend-data.hh
```diff
@@ -0,0 +1,567 @@
...
+/**
+ * hb_depend_data_t:
+ *
+ * Internal container for all dependency graph data.
+ *
+ * Persistent data (retained after graph construction):
+ * - glyph_dependencies: Per-glyph edge lists
+ * - sets: Ligature sets and context sets indexed by set ID
+ *
+ * Temporary data (freed after construction via cleanup()):
```
behdad: I wonder if this deserves a separate "builder" object.

src/hb-depend-data.hh
```diff
@@ -0,0 +1,567 @@
...
+
+  hb_codepoint_t new_ligature_set (hb_set_t &set)
+  {
+    hb_codepoint_t set_index;
+
+    if (free_set_list.length > 0)
+    {
+      /* Reuse freed set */
+      set_index = free_set_list[free_set_list.length - 1];
+      free_set_list.resize (free_set_list.length - 1);
```
behdad: You can use `free_set_list.pop()`.

src/hb-depend-data.hh
```diff
@@ -0,0 +1,567 @@
...
+    }
+    else
+    {
+      /* Allocate new set */
+      set_index = sets.length;
+      hb_set_t *new_set = hb_set_create ();
+      if (unlikely (!new_set))
+        return HB_CODEPOINT_INVALID;
+      new_set->set (set);
+      sets.push (hb::unique_ptr<hb_set_t> {new_set});
```
behdad: Try `push_or_fail()` please, to avoid leaking on malloc fail.

src/hb-depend.cc
```diff
@@ -0,0 +1,431 @@
...
+ *
+ * Dependency graphs enable finding all glyphs reachable from a given input set.
+ * This is useful for font subsetting, analyzing glyph coverage, optimizing
+ * font delivery, and determining which glyphs are needed to render specific
+ * characters.
+ */
+
+hb_depend_t::hb_depend_t (hb_face_t *f)
+{
+  if (unlikely (f == nullptr)) {
```
behdad: We don't handle nullptr objects typically. It's invalid. Just crashing is fine I think.

src/hb-depend.cc
```diff
@@ -0,0 +1,431 @@
...
+  /* Extract dependencies from all relevant OpenType tables
+   * Note: cmap (UVS) dependencies are not extracted - UVS closure is handled
+   * separately via hb_font_get_variation_glyph() query API during closure computation. */
+  get_gsub_dependencies();
+  get_math_dependencies();
+  get_colr_dependencies();
+  get_glyf_dependencies();
+  get_cff_dependencies();
+
+  /* Free temporary structures - no longer needed after graph extraction */
```
behdad: Yeah I feel like this really needs a builder object instead.

src/hb-depend.cc
```diff
@@ -0,0 +1,431 @@
...
+ *    - Records each substitution (input -> output) as a dependency
+ *    - Tags with the feature tag and table tag (GSUB)
+ *    - For ligatures, assigns a ligature_set ID to group alternatives
+ *
+ * The result is a complete graph of all possible glyph substitutions
+ * that could occur through OpenType Layout features, regardless of
+ * script, language, or shaping context.
+ */
+void hb_depend_t::get_gsub_dependencies() {
+  hb_blob_t *gsub_blob = hb_sanitize_context_t ().reference_table<OT::Layout::GSUB> (face);
```
behdad: Don't reference and sanitize tables directly. Simply use `face->table.GSUB` or `face->table.GSUB.table` as needed.

src/hb-depend.cc
```diff
@@ -0,0 +1,431 @@
...
+ *    - Iterates through covered glyphs (input glyphs)
+ *    - Records each substitution (input -> output) as a dependency
+ *    - Tags with the feature tag and table tag (GSUB)
+ *    - For ligatures, assigns a ligature_set ID to group alternatives
+ *
+ * The result is a complete graph of all possible glyph substitutions
+ * that could occur through OpenType Layout features, regardless of
+ * script, language, or shaping context.
+ */
+void hb_depend_t::get_gsub_dependencies() {
```
behdad: I assumed the per-table code goes into the tables themselves.

src/hb-depend.cc
```diff
@@ -0,0 +1,431 @@
...
+    successful = false;
+    return;
+  }
+  successful = true;
+
+  /* Extract dependencies from all relevant OpenType tables
+   * Note: cmap (UVS) dependencies are not extracted - UVS closure is handled
+   * separately via hb_font_get_variation_glyph() query API during closure computation. */
+  get_gsub_dependencies();
+  get_math_dependencies();
```
behdad: These can simply be replaced with `face->table.MATH.depend(...)` or similar.

src/hb-depend.cc
```diff
@@ -0,0 +1,431 @@
...
+ * preserve glyph integrity.
+ */
+void hb_depend_t::get_glyf_dependencies()
+{
+  OT::glyf_accelerator_t glyf (face);
+  if (!glyf.has_data())
+    return;
+  for (hb_codepoint_t gid = 0; gid < glyf.get_num_glyphs (); gid++) {
+    auto glyph = glyf.glyph_for_gid(gid);
+    for (auto &item : glyph.get_composite_iterator ())
```
behdad: Please add `VARC` composite iterator here as well.

src/hb-depend.cc
```diff
@@ -0,0 +1,431 @@
...
+      // SEAC glyph found - add dependencies to base and accent
+      data.add_depend(gid, HB_TAG('C','F','F',' '), base_gid);
+      data.add_depend(gid, HB_TAG('C','F','F',' '), accent_gid);
+    }
+  }
+}
+
+hb_depend_t::~hb_depend_t ()
+{
+  hb_face_destroy(face);
```
behdad: Why does depend object need to hold onto the face?

src/hb-ot-cmap-table.hh
```diff
@@ -32,6 +32,7 @@
 #include "hb-open-type.hh"
 #include "hb-set.hh"
 #include "hb-cache.hh"
+#include "hb-depend-data.hh"
```
behdad: Why this include?

behdad:
> Gulzar-Regular.ttf would be a good candidate but it is currently only in `perf/fonts`

You can reference fonts anywhere in the tree.

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -375,7 +376,89 @@ struct hb_collect_glyphs_context_t :
   void set_recurse_func (recurse_func_t func) { recurse_func = func; }
 };
 
+#ifdef HB_DEPEND_API
+struct hb_depend_context_t :
+       hb_dispatch_context_t<hb_depend_context_t, bool>
```
behdad: Please clarify return value meaning again.

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -375,7 +376,89 @@ struct hb_collect_glyphs_context_t :
...
+  {
+    if (!active_glyphs_stack)
+      return *glyphs;
+
+    return active_glyphs_stack.tail ();
+  }
+
+  hb_set_t* push_cur_active_glyphs ()
+  {
+    hb_set_t *s = active_glyphs_stack.push ();
```
behdad: You can use the newer `push_or_fail()`.

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -1976,6 +2249,17 @@ struct ContextClosureLookupContext
...
 };
 
+#ifdef HB_DEPEND_API
+struct ContextDependLookupContext
+{
+  ContextClosureFuncs funcs;
+  ContextFormat context_format;
+  const void *intersects_data;
+  void *intersects_cache;
+  void *intersected_glyphs_cache;
```
behdad: What are these `void*` caches?

test/fuzzing/hb-depend-closure-parity-main.cc
```diff
@@ -0,0 +1,315 @@
+#include "hb-fuzzer.hh"
```
behdad, line 1: This file is pretty hard to follow.

test/fuzzing/hb-depend-closure-parity.cc
```diff
@@ -0,0 +1,1741 @@
+#include "hb-fuzzer.hh"
```
behdad, line 1: I meant this file actually.

test/fuzzing/meson.build
```diff
@@ -48,25 +53,72 @@ foreach file_name : tests
...
+  endforeach
+endif
+
 glob_cmd = find_program('glob.py', required: true)
 
 fonts_glob = run_command(glob_cmd, meson.current_source_dir() / 'fonts', check:true).stdout().strip().split('\n')
 subset_fonts_glob = run_command(glob_cmd, meson.current_source_dir() / '..' / 'subset' / 'data' / 'fonts', check:true).stdout().strip().split('\n')
 graphs_glob = run_command(glob_cmd, meson.current_source_dir() / 'graphs', check:true).stdout().strip().split('\n')
+# Use well-formed API test fonts for depend fuzzer (not malformed clusterfuzz fonts)
+api_fonts_glob = run_command(glob_cmd, meson.current_source_dir() / '..' / 'api' / 'fonts', check:true).stdout().strip().split('\n')
```
behdad: The fuzzer is going to produce malformed fonts anyway. Why restrict it here?

behdad (changes requested):

src/hb-depend.h
```diff
@@ -0,0 +1,79 @@
...
+ * XSince: REPLACEME
+ */
+
+typedef struct hb_depend_t hb_depend_t;
+
+HB_EXTERN hb_depend_t *
+hb_depend_from_face_or_fail (hb_face_t *face);
+
+HB_EXTERN hb_bool_t
+hb_depend_get_glyph_entry (hb_depend_t *depend,
```
behdad: Also a pattern we follow in other API is that array getters like this simply return the number of elements.

src/OT/Color/COLR/COLR.hh
skef: All depend and dependv1 methods are now void.

src/OT/Color/COLR/COLR.hh
skef: Removed. The context now holds a const COLR *colr directly.

src/OT/Color/COLR/COLR.hh
skef: Good point — gradient paint dependv1 methods are now empty.

src/OT/Color/COLR/COLR.hh
skef: Done — all out-of-line methods now use HB_INTERNAL.

src/OT/Color/COLR/colrv1-depend.hh
skef: Done — PaintTransform::dependv1 only dispatches to src; gradient and color paints have empty dependv1.

src/OT/Layout/GSUB/Ligature.hh
skef: Done — using + hb_iter (component) | hb_sink (complete_ligset).

src/OT/Layout/GSUB/Ligature.hh
skef: Done.

src/OT/Layout/GSUB/Ligature.hh
skef: Renamed to gid -- more neutral for total usage.

src/OT/Layout/GSUB/ReverseChainSingleSubstFormat1.hh
skef: All depend methods are now void.

src/hb-depend-data.hh
skef: Done — hb_depend_edge_flags_t with HB_MARK_AS_FLAG_T

src/hb-depend-data.hh
skef: Done.

src/hb-depend-data.hh
skef: Renamed to hb_depend_glyph_record_t and kept as a struct rather than a type alias — the struct form lets us forward-declare in hb-depend.h without pulling in hb-vector.hh.

src/hb-depend-data.hh
skef: Renamed to hb_depend_lookup_revmap_t; brace fixed.

src/hb-depend-data.hh
skef: Done.

src/hb-depend-data.hh
skef: Done — hb_depend_data_builder_t in hb-depend-data.hh.

src/hb-depend-data.hh
skef: Done.

src/hb-depend-data.hh
skef: Done.

src/hb-depend-data.hh
skef: Done — internal method returns const hb_set_t *; the public API wrapper copies into the out param.

src/hb-depend.cc
skef: Done.

src/hb-depend.cc
skef: Done.

src/hb-depend.cc
skef: Ibid.

src/hb-depend.cc
skef: Done.

src/hb-depend.cc
skef: Done.

src/hb-depend.cc
skef: I wasn't comfortable doing this. There's no closure path yet, and indeed no analogous traversal algorithm (paint doesn't quite qualify). If you want to commit to an architecture that will work for both depend and subset I can try to follow it, but absent that I feel like I would just create something that would either be replaced when the subset path is added or become a maintenance headache due to the differences.

src/hb-depend.cc
skef: It doesn't — dropped. Face is now passed as a local during construction only.

src/hb-ot-cmap-table.hh
skef: Stale include from when cmap depend extraction existed — removed.

src/hb-ot-layout-gsubgpos.hh
skef: Done — dispatch returns hb_empty_t (), documented in a comment.

src/hb-ot-layout-gsubgpos.hh
skef: Done.

src/hb-ot-layout-gsubgpos.hh
skef: Replaced with typedefs of the corresponding Closure counterparts, removing the reinterpret_casts and layout-identity comments.

test/fuzzing/hb-depend-closure-parity.cc
skef: Refactored significantly: extracted rng_prob_hit() and depend_closure_ctx_t, reduced the main processing functions from ~10
  parameters to 3, removed dead code, and reordered the file so the two closure algorithms and the comparison come first, with
  test-generation infrastructure after.

test/fuzzing/meson.build
skef: Good point — switched to subset_fonts_glob for the parity checker (well-formed fonts make sense there since it's testing
  correctness against hb_subset_plan_t), and removed api_fonts_glob entirely.

src/hb-depend.h
skef: Done — hb_subset_depend_lookup_glyph now returns the total count and takes start_offset + *entry_count IN/OUT + entries array.

src/OT/Layout/GSUB/ReverseChainSingleSubstFormat1.hh
skef: Done.

src/OT/Layout/GSUB/ReverseChainSingleSubstFormat1.hh
skef: Done — HB_DEPEND_CONTEXT_SET_FLAG macro, used in hb-depend-data.hh, hb-ot-layout-gsubgpos.hh, and
  ReverseChainSingleSubstFormat1.hh.

skef:
@behdad I think I'm caught up to your comments so far. I kept hb-depend-data.hh and hb_depend_data_t as the name/container for the compiled graph; "edge" doesn't really make sense to me here. The reason for the separation is to avoid include file cycles with hb-depend.hh. I suppose we could use "impl" instead, but I didn't find an analogy/model for this elsewhere in the repo.

src/OT/Layout/GSUB/Ligature.hh
behdad: Wouldn't `hb_set_t (complete_ligset (component); complete_ligset.add (first)` work?

Also, we typically add a space before function call opening '('. Don't worry about fixing all. Just in new code please try to follow.

skef:
@behdad I took your Ligature.hh advice but I think the comment got eaten by the rebase and change. 

The addition of the fonts designed for subset testing surfaced some bugs that I also addressed:


- Infinite recursion on self-referential GSUB lookups — `hb_depend_context_t::recurse` now uses call-stack cycle detection (add on entry, remove on exit via `lookups_seen`). Triggered by `gsub_chaining1_multiple_subrules_f1.otf` which has a ChainContextSubst that references itself.
- Out-of-range glyph IDs in `SingleSubstFormat1::depend` — the delta modulo can wrap past `num_glyphs`. The closure code produces these too but the subsetter filters them later via `_remove_invalid_gids`; better to filter at the source.
- Not filtering context position-0 class members by coverage in `context_depend_lookup` and `chain_context_depend_lookup` — `collect_class` was returning all class members without intersecting with the coverage, producing edges for glyphs that can never trigger the rule.

That testing also prompted PR #5992 and Issue #5993, the latter of which points to a potential enhancement using the code in this PR.

behdad:
> * Infinite recursion on self-referential GSUB lookups — `hb_depend_context_t::recurse` now uses call-stack cycle detection (add on entry, remove on exit via `lookups_seen`). Triggered by `gsub_chaining1_multiple_subrules_f1.otf` which has a ChainContextSubst that references itself.

Thanks. There's also the `hb-decycler.hh` you can use in that situation if you wanted to bother.

skef:
@behdad I looked into the decycler option and it seems like a win in some ways but a loss in others, given the two-level aspect. I'm going to leave it as-is but feel free to nudge a little harder.

skef:
With luck the latest commit should fix the docs error.

behdad:
> @behdad I looked into the decycler option and it seems like a win in some ways but a loss in others, given the two-level aspect. I'm going to leave it as-is but feel free to nudge a little harder.

No worries. In places that I used it, it had measurable performance improvement. So I'm curious how if we will see any in the case of fonts like NotoNastaliqUrdu that are heavily recursive. But that's not a blocker issue of course.

behdad:
@skef What's the status of this? Are you going to do the mostly mechanical changes I suggested in:
https://github.com/harfbuzz/harfbuzz/pull/5733#issuecomment-4307746438

skef:
@behdad is any of that missing? I may not have done everything absolutely down the line but I thought I addressed that stuff.

behdad:
> @behdad is any of that missing? I may not have done everything absolutely down the line but I thought I addressed that stuff.

Oh I apologize, I didn't realize that. I just read your comment responses. I'll go ahead and do a final review. Thanks.

skef:
No problem! Since it wasn't attached to a line I wasn't sure whether or where to respond.

I think the only substantial thing I punted on was the VARC traversal. Like I said I am willing to do that but I would need guidance as to the approach. Maybe just traversing like the rasterization path would work, but it seems like there's a good chance the subset code will deserialize and reserialize, and it probably makes more sense for depend to look like that code path rather than the paint codepath.

behdad:
Looks like the depend code is still being compiled into libharfbuzz. My suggestion was that `hb-depend.h` should be either renamed to `hb-subset-depend.h` and included from `hb-subset.h`, or folded into `hb-subset.h` directly. And the code compiled into libharfbuzz-subset. Thanks.

skef:
> Looks like the depend code is still being compiled into libharfbuzz. My suggestion was that hb-depend.h should be either renamed to hb-subset-depend.h and included from hb-subset.h, or folded into hb-subset.h directly. And the code compiled into libharfbuzz-subset. Thanks.

Should be done now.

behdad:
Thanks Skef. I'll review soon. Looks like docs bot is failing after your renames.

behdad:
Just a heads up that I like to land https://github.com/harfbuzz/harfbuzz/pull/5655 first, then resolve any conflicts here and land it.

This work also surfaced another issue, which is pre-existing: In other modules, the top-level header includes all subheaders, and subheaders have include guards against being included directly. However, I see `hb-subset.h` _doesn't_ include `hb-subset-serialize.h`.

I propose we fix the above, and include `hb-subset-depend.h` from `hb-subset.h` as well. Add include guards to `hb-subset-depend.h`, but perhaps NOT to `hb-subset-serialize.h` so as to not break any client's code, but warn instead. @garretrieger

skef:
I made other attempt at the docs fix, but the CI doesn't trigger automatically so ...

PR #5735, opened 2026-02-09
Add `hb-draw`

behdad:
Here's a starting point by codex.


behdad:
@twardoch

src/OT/glyf/path-builder.hh
```diff
@@ -165,7 +165,11 @@ struct path_builder_t
 				   first_oncurve.x, first_oncurve.y);
     }
     else if (first_oncurve)
-      draw_session->line_to (first_oncurve.x, first_oncurve.y);
+    {
+      if (draw_session->st.current_x != first_oncurve.x ||
+	  draw_session->st.current_y != first_oncurve.y)
+	draw_session->line_to (first_oncurve.x, first_oncurve.y);
+    }
```
behdad: I made this change to match Skrifa. But looks like now I don't match FreeType. Asked @dfrg for clarification.

khaledhosny:
The [`test/draw/data/expected/basic/cff2-var-wght900.svg`](https://github.com/harfbuzz/harfbuzz/pull/5735/changes#diff-b0f15e648cc9cfdd0bc166d4361505cd4833c21773767cc1481797b54893d087) file shows 3 `.notdef` glyphs, is this expected?

behdad:
> The [`test/draw/data/expected/basic/cff2-var-wght900.svg`](https://github.com/harfbuzz/harfbuzz/pull/5735/changes#diff-b0f15e648cc9cfdd0bc166d4361505cd4833c21773767cc1481797b54893d087) file shows 3 `.notdef` glyphs, is this expected?

I noticed that. It's the AI coming up with wrong test. I can remove or just let be.

test/draw/record-test.sh
```diff
@@ -0,0 +1,296 @@
...
+  expected_path="$expected_file"
+else
+  expected_path="$out_dir/$expected_file"
+fi
+
+mkdir -p "$(dirname "$expected_path")"
+"$hb_draw" "${options[@]}" "$subset_path" "$text" > "$expected_path"
+
+if [ "$out" = "/dev/stdout" ] || [ "$out" = "-" ]; then
+  font_field=$(/usr/bin/env python3 -c 'import os,sys; print(os.path.relpath(sys.argv[1], sys.argv[2]))' "$subset_path" "$PWD")
```
khaledhosny, line 285: Can’t `realpath` be used here?

test/shape/record-test.sh
```diff
@@ -29,14 +43,31 @@ if test "x${fontfile:0:1}" == 'x-'; then
 	exit 1
 fi
 shift
-if ! echo "$hb_subset" | grep -q 'subset'; then
-	echo "Specify hb-subset (or \"fonttools subset\"): got '$hb_subset'." >&2
-	exit 1
```
khaledhosny, line 34: Does that mean we can’t use `fonttools subset` here anymore? I think `hb-subset` is pretty good now, so probably not an issue.

util/draw-output-color.hh
```diff
@@ -0,0 +1,935 @@
...
+ *
+ * THE COPYRIGHT HOLDER SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ * FITNESS FOR A PARTICULAR PURPOSE.  THE SOFTWARE PROVIDED HEREUNDER IS
+ * ON AN "AS IS" BASIS, AND THE COPYRIGHT HOLDER HAS NO OBLIGATION TO
+ * PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
+ */
+
+/* This file is included inside draw_output_t in draw-output.hh.
+ * It provides color font rendering via the hb_paint API → SVG. */
```
khaledhosny, line 26: That should be a nice reference on how to map paint API SVG (I struggled with this previously). Now if I can get a PDF equivalent 🥺

test/shape/record-test.sh
behdad: Right. We can revert back or add an explicit `--hb-subset` if needed.

test/draw/record-test.sh
behdad: Isn't that doing the opposite? Getting the relative path?

khaledhosny: *I think* `realpath --relative-to="$PWD" "$subset_path"` does the same thing as `os.path.relpath()` call here.

behdad: Mine on Mac doesn't have that option. I think Python is more portable.

khaledhosny: Right, it is probably a GNU extension (I always use GNU coreutils, even on Mac).

khaledhosny:
`--foreground` does not seem to do anything for me. A `--background` would be nice as well, and maybe also `--ink`/`--logical`.

behdad:
> `--foreground` does not seem to do anything for me. A `--background` would be nice as well, and maybe also `--ink`/`--logical`.

- Ah, `--foreground` was only used for color fonts. Fixing.
- Added `--background`.
- I thought `ink`/`logical` would be overkill for this tool, since my primary goal is stable output. WDYT?

khaledhosny:
> I thought `ink`/`logical` would be overkill for this tool, since my primary goal is stable output. WDYT?

I’m OK with logical being the default, unlike hb-view. But text with -ve side bearings is not uncommon, so `--ink` would be nice to have.

behdad:
Maybe `--margin` also good, with 0 default?

khaledhosny:
> Maybe `--margin` also good, with 0 default?

Yes, that would be nice as well.

behdad:
New features implemented.

@khaledhosny what are your thoughts on `--ned` and `-q`? Are they useful?

khaledhosny:
> New features implemented.
> 
> @khaledhosny what are your thoughts on `--ned` and `-q`? Are they useful?

When inlining SVGs in HTML, path `id`s need to be unique, but if one is generating multiple SVG from different fonts, the `id`s can overlap, so `--ned` option can be a quick way around this. I don’t know about `--quiet` though, it is a strange choice of an option name for what it does, and I don’t know why one would need path data only, but maybe it would be useful for someone.

behdad:
I'll remove --quiet. I added it when I thought we'll always output flat glyphs only.

behdad:
Should we keep --ned or rename to --flat?

khaledhosny:
> Should we keep --ned or rename to --flat?

I have no preference.

behdad:
> > Should we keep --ned or rename to --flat?
> 
> I have no preference.

I renamed it.

Other changes pushed as well. I'm happy with the interface now.

behdad:
Finally, does it bother us that `hb-draw` also paints?

khaledhosny:
> Finally, does it bother us that `hb-draw` also paints?

We can call it `hb-paint`, thought I think draw sound more natural and groups well with shape and view.

khaledhosny (approved):

Issue #5739, opened 2026-02-10
12.3.0 rejects a cmap table of a font that was previously accepted


khaledhosny:
Reported on https://bugs.documentfoundation.org/show_bug.cgi?id=170650

Bisected to: 83dc4d1d35d8b52a45ad793e617a5bbc5401922f

Both DirectWrite and CoreText seem to accept the cmap table.

behdad:
Thanks. Yeah, I removed the code to work around overlong subtable length field. I'll think about how to ignore it instead.

PR #5740, opened 2026-02-11
[cmap/format4] Try to ignore wrong subtable length

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/5739

Not finished. See comments.


behdad:
I think this is good enough.

behdad:
Ugh. Need to debug more...

Issue #5743, opened 2026-02-15
`hb_buffer_serialize()` with `HB_BUFFER_SERIALIZE_FLAG_NO_ADVANCES` give wrong offsets with partial glyph range


khaledhosny:
Take this sample code:
```c
#include <stdio.h>
#include <hb.h>

int main(int argc, char **argv) {
    if (argc < 3) {
        fprintf(stderr, "Usage: %s <font-file> <text>\n", argv[0]);
        return 1;
    }
    const char *font_path = argv[1];
    const char *text = argv[2];

    hb_face_t *face = hb_face_create_from_file_or_fail(font_path, 0);
    hb_font_t *font = hb_font_create(face);
    hb_buffer_t *buffer = hb_buffer_create();
    hb_buffer_add_utf8(buffer, text, -1, 0, -1);
    hb_buffer_guess_segment_properties(buffer);
    hb_shape(font, buffer, NULL, 0);

    char buf[64];
    unsigned int start = 0;
    unsigned int num_glyphs = hb_buffer_get_length(buffer);
    while (start < num_glyphs) {
        unsigned int consumed;
        start += hb_buffer_serialize(buffer, start, num_glyphs,
                                     buf, sizeof(buf), &consumed,
                                     font, HB_BUFFER_SERIALIZE_FORMAT_TEXT,
                                     HB_BUFFER_SERIALIZE_FLAG_NO_ADVANCES);
        if (consumed == 0) break;
        fwrite(buf, 1, consumed, stdout);
	printf("\n");
    }

    hb_buffer_destroy(buffer);
    hb_font_destroy(font);
    hb_face_destroy(face);

    return 0;
}
```

```
$ cc example.c $(pkg-config harfbuzz --cflags --libs) -o example \
  && ./example test/api/fonts/Roboto-Regular.abc.ttf "aaaaaa"
[gid1=0|gid1=1@1114,0|gid1=2@2228,0|gid1=3@3342,0|gid1=4@4456,0
|gid1=5]
```

The x offset of the last glyph is 0 because it is the first glyph in the second call to `hb_buffer_serialize()`, but that is unexpected and confusing. `hb_buffer_serialize()` should always calculate the offsets in this case taking into account all the glyphs from the start of the buffer.

PR #5744, opened 2026-02-15
[buffer-serialize] Fix calculating absolute offsets with partial range

khaledhosny:
Fixes https://github.com/harfbuzz/harfbuzz/issues/5743

### DIFF
--- a/src/hb-buffer-serialize.cc
+++ b/src/hb-buffer-serialize.cc
@@ -114,6 +114,17 @@ _hb_buffer_serialize_glyphs_json (hb_buffer_t *buffer,
 
   *buf_consumed = 0;
   hb_position_t x = 0, y = 0;
+
+  /* Calculate the advance of the previous glyphs */
+  if (pos && (flags & HB_BUFFER_SERIALIZE_FLAG_NO_ADVANCES))
+  {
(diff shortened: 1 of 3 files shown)


behdad:
Thanks. Maybe document that this happens and suggest larger buffer for no-advance to reduce O(n^2) behavior cost.

src/hb-buffer-serialize.cc
```diff
@@ -114,6 +114,17 @@ _hb_buffer_serialize_glyphs_json (hb_buffer_t *buffer,
 
   *buf_consumed = 0;
   hb_position_t x = 0, y = 0;
+
+  /* Calculate the advance of the previous glyphs */
+  if (start > 0 && pos && (flags & HB_BUFFER_SERIALIZE_FLAG_NO_ADVANCES))
```
behdad: You don't need `start > 0`...

src/hb-buffer-serialize.cc
```diff
@@ -272,6 +283,17 @@ _hb_buffer_serialize_glyphs_text (hb_buffer_t *buffer,
...
```
behdad: Same.

behdad (approved):

behdad:
Thanks!

PR #5748, opened 2026-02-17
Lower CFF2 to CFF when subsetting to static instance

behdad:
This is good enough I think.

### DIFF
--- a/src/hb-cff-specializer.hh
+++ b/src/hb-cff-specializer.hh
@@ -171,16 +171,16 @@ specialize_commands (hb_vector_t<cs_command_t> &commands,
 
       if (dx_zero && !dy_zero)
       {
-        /* Horizontal: keep only dy */
-        cmd.op = (cmd.op == OpCode_rmoveto) ? OpCode_hmoveto : OpCode_hlineto;
+        /* Vertical movement (dx=0): keep only dy */
+        cmd.op = (cmd.op == OpCode_rmoveto) ? OpCode_vmoveto : OpCode_vlineto;
         /* Shift dy to position 0 */
         cmd.args[0] = cmd.args[1];
         cmd.args.resize (1);
       }
       else if (!dx_zero && dy_zero)
       {
-        /* Vertical: keep only dx */
-        cmd.op = (cmd.op == OpCode_rmoveto) ? OpCode_vmoveto : OpCode_vlineto;
+        /* Horizontal movement (dy=0): keep only dx */
+        cmd.op = (cmd.op == OpCode_rmoveto) ? OpCode_hmoveto : OpCode_hlineto;
         cmd.args.resize (1);  /* Keep only dx */
       }
       /* else: both zero or both non-zero, keep as rmoveto/rlineto */


behdad:
We might want to add a flag for it, but enabled by default.

behdad:
Tests need rebasing.

src/hb-subset-cff2-to-cff1.hh
```diff
@@ -0,0 +1,173 @@
...
+ * ✓ FDSelect3 format (compact range-based, 8 bytes for single-FD fonts)
+ * ✓ CID Charset with identity mapping (format 2)
+ * ✓ CharStrings with endchar operators (CFF1 requires, CFF2 doesn't)
+ * ✓ Private DICT instantiation (blend operators evaluated)
+ * ✓ Desubroutinized path (CharStrings are flattened, no subroutines)
+ * ✓ OTS validation passes
+ * ✓ HarfBuzz rendering works
+ *
+ * KNOWN LIMITATIONS:
+ * ✗ FontBBox is "0 0 0 0" (should calculate actual bounds)
```
khaledhosny: Couldn’t the extents from the `head` table be used here?

src/hb-subset-cff2-to-cff1.hh
```diff
@@ -0,0 +1,173 @@
...
+ * ✗ Width handling not verified (CFF2 uses hmtx, CFF1 encodes in CharStrings)
```
khaledhosny: What verification is needed here?

src/hb-subset-cff2-to-cff1.hh
khaledhosny: Actually this is done bellow, I think the comment is outdated.

khaledhosny:
> We might want to add a flag for it, but enabled by default.

For `hb-subset` util or compile-time one?

src/hb-subset-cff2-to-cff1.hh
```diff
@@ -0,0 +1,173 @@
...
+  // CFF1-specific additions
+  hb_vector_t<unsigned char> fontName;  // Single font name for Name INDEX
+
+  bool create (const OT::cff2_subset_plan &cff2_plan_)
+  {
+    cff2_plan = &cff2_plan_;
+
+    // Create a simple font name (CFF1 requires a Name INDEX)
+    // For subsets, we can use a generic name
+    const char *name = "CFF1Font";
```
khaledhosny: Can we use the font’s psname? Also the `CFF1Font` string is hard-coded again in TopDict below, it is probably better to set it in one place.

khaledhosny:
I’m testing this (using [`RanaKufi.otf`](https://github.com/aliftype/rana-kufi/blob/main/fonts/RanaKufi.otf) but all the outlines are deformed:
image

src/hb-subset-cff2-to-cff1.hh
behdad: Right.

src/hb-subset-cff2-to-cff1.hh
behdad: Outdated as well.

behdad:
> > We might want to add a flag for it, but enabled by default.
> 
> For `hb-subset` util or compile-time one?

HB_SUBSET_FLAG.

behdad:
> I’m testing this (using [`RanaKufi.otf`](https://github.com/aliftype/rana-kufi/blob/main/fonts/RanaKufi.otf) but all the outlines are deformed:

Thanks. Lemme add the option and check if the previous CFF2 output was correct.

behdad:
One issue I realized: CFF has unspecified overlap handling mode whereas CFF2 has non-zero winding rule. This can be a dealbreaker, or maybe it's fine to ignore the problem.

behdad:
> > > We might want to add a flag for it, but enabled by default.
> > 
> > 
> > For `hb-subset` util or compile-time one?
> 
> HB_SUBSET_FLAG.

In `fonttools subset`, this feature is off by default. You have to provide `--downgrade-cff2` to enable it. Should we do the same? @garretrieger @madig @khaledhosny @anthrotype

src/hb-subset-cff2-to-cff1.hh
behdad: I think that's unnecessary.

behdad:
> I’m testing this (using [`RanaKufi.otf`](https://github.com/aliftype/rana-kufi/blob/main/fonts/RanaKufi.otf) but all the outlines are deformed:

Fixed and addressed all other issues.

behdad:
> > > > We might want to add a flag for it, but enabled by default.
> > > 
> > > 
> > > For `hb-subset` util or compile-time one?
> > 
> > 
> > HB_SUBSET_FLAG.
> 
> In `fonttools subset`, this feature is off by default. You have to provide `--downgrade-cff2` to enable it. Should we do the same? @garretrieger @madig @khaledhosny @anthrotype

Given the overlap issue raised in https://github.com/harfbuzz/harfbuzz/pull/5748#issuecomment-3916285623, I made it default off like fonttools.

khaledhosny:
Looks good to me now.

PR #5749, opened 2026-02-18
Add `harfbuzz-raster` CPU rasterization library

behdad:


src/hb-raster.h
```diff
@@ -0,0 +1,161 @@
...
+/* ── Shared types ─────────────────────────────────────────────────── */
+
+/**
+ * hb_raster_format_t:
+ * @HB_RASTER_FORMAT_A8: 8-bit alpha-only coverage
+ * @HB_RASTER_FORMAT_BGRA32: 32-bit BGRA color
+ *
+ * Pixel format for raster images.
+ *
+ * Since: REPLACEME
```
khaledhosny:
```suggestion
 * XSince: REPLACEME
```

src/hb-raster.h
```diff
@@ -0,0 +1,161 @@
...
+ * hb_raster_extents_t:
+ * @x_origin: X coordinate of the left edge of the image in glyph space
+ * @y_origin: Y coordinate of the bottom edge of the image in glyph space
+ * @width: Width in pixels
+ * @height: Height in pixels
+ * @stride: Bytes per row; 0 means auto-calculate on input, filled on output
+ *
+ * Pixel-buffer extents for raster operations.
+ *
+ * Since: REPLACEME
```
khaledhosny:
```suggestion
 * XSince: REPLACEME
```

behdad:
Initially this was a 8-subpixel MSAA implementation, which was low quality and slow.
Swapped with a full-coverage rasterizer. Quality should be as good as FreeType.
Speed, with SIMD on Apple silicon is current 2x slower than FreeType. Not bad for a start!

behdad:
@khaledhosny This is ready for review I think.
cc @dfrg

behdad:
Now with COLRv1 support as well. I think this is ready.

behdad:
I adjusted the rasterizer curve flattening to match FreeType. It's a bit slower (~10%), but the output is much closer to FreeType. I'm very happy with this. I'm going to merge, and we can tweak the API if someone has comments on it. Thanks!

src/hb-raster.h
```diff
@@ -0,0 +1,260 @@
...
+ * @glyph_extents: glyph extents from hb_font_get_glyph_extents()
+ * @extents: (out): output raster extents
+ *
+ * Converts #hb_glyph_extents_t to #hb_raster_extents_t with no extra padding.
+ *
+ * Return value: `true` if extents are non-empty, `false` otherwise.
+ *
+ * XSince: REPLACEME
+ */
+HB_EXTERN hb_bool_t
```
khaledhosny, line 80: This function is documented both here and in the .cc file, I don’t know which one gtk-doc will pick, but the one here looks redundant/

behdad:
> I adjusted the rasterizer curve flattening to match FreeType. It's a bit slower (~10%), but the output is much closer to FreeType. I'm very happy with this. I'm going to merge, and we can tweak the API if someone has comments on it. Thanks!

To recap, in my testing, the grayscale rasterizer is over 2x faster than FreeType for a range of workloads I tested.

PR #5755, opened 2026-02-23
[buffer] Add hb_buffer_changed()

behdad:
To be called by message callbacks if they modify buffer glyph indices, to update internal caches (currently the buffer digest). This should take care of the main problem. However, it wouldn't invalidate the cached glyph class from GDEF. So as long as the glyphs are of the same type (base, ligature, mark), this should be fine.

In theory I can make it take font as well, and update the GDEF classes. It's just a bit of a layering violation so I didn't do it that way. There's at least another way I can think of to address this properly. The hb_apply_context_t can register an internal callback with the buffer to be called when the buffer_changed is called. Maybe that's cleaner. approach. I'll work on that.

### DIFF
--- a/src/hb-ot-layout-gsubgpos.hh
+++ b/src/hb-ot-layout-gsubgpos.hh
@@ -777,22 +777,29 @@ struct hb_ot_apply_context_t :
   {
     init_iters ();
     match_positions.set_storage (stack_match_positions);
+
 #ifndef HB_NO_BUFFER_MESSAGE
-    assert (buffer->changed_func == nullptr ||
-	    buffer->changed_func == buffer_changed_trampoline);
-    orig_changed_func = buffer->changed_func;
-    orig_changed_data = buffer->changed_data;
-    buffer->changed_func = buffer_changed_trampoline;
-    buffer->changed_data = this;
+    if (buffer->messaging ())
+    {
+      assert (buffer->changed_func == nullptr ||
+	      buffer->changed_func == buffer_changed_trampoline);
+      orig_changed_func = buffer->changed_func;
+      orig_changed_data = buffer->changed_data;
+      buffer->changed_func = buffer_changed_trampoline;
+      buffer->changed_data = this;
+    }
 #endif
   }
   ~hb_ot_apply_context_t ()
   {
 #ifndef HB_NO_BUFFER_MESSAGE
-    assert (buffer->changed_func == buffer_changed_trampoline);
-    assert (buffer->changed_data == this);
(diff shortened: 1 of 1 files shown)


behdad:
> There's at least another way I can think of to address this properly. The hb_apply_context_t can register an internal callback with the buffer to be called when the buffer_changed is called. Maybe that's cleaner. approach. I'll work on that.

Done.

khaledhosny:
> So as long as the glyphs are of the same type (base, ligature, mark), this should be fine.

So that is the only limitation? One can replace, ligate or decompose glyphs as long as they are of the same GDEF class, or does this only allow one-to-one replacements? Any special consideration about clusters?

behdad:
> > So as long as the glyphs are of the same type (base, ligature, mark), this should be fine.
> 
> So that is the only limitation?

I addressed that particular limitation.

> One can replace, ligate or decompose glyphs as long as they are of the same GDEF class, or does this only allow one-to-one replacements? Any special consideration about clusters?

Oh I was only thinking of one-to-one. Anything else would get really tricky. I wouldn't even try.

What are your requirements? You want to get a callback soon after cmap? Or at the end of subsitutions, or some other times? Maybe we can add something more real callbacks for modifying the buffer somehow.

khaledhosny:
> > > So as long as the glyphs are of the same type (base, ligature, mark), this should be fine.
> > 
> > 
> > So that is the only limitation?
> 
> I addressed that particular limitation.

Excellent.

> > One can replace, ligate or decompose glyphs as long as they are of the same GDEF class, or does this only allow one-to-one replacements? Any special consideration about clusters?
> 
> Oh I was only thinking of one-to-one. Anything else would get really tricky. I wouldn't even try.
> 
> What are your requirements? You want to get a callback soon after cmap? Or at the end of subsitutions, or some other times? Maybe we can add something more real callbacks for modifying the buffer somehow.

I think one-to-one is fine. The main need right now is emulating feature variation, so it is only one-to-one, the GDEF classes don’t even change. But I think @justvanrossum might have more potential use cases in mind.

justvanrossum:
My original question was rather exploratory, a "what if", and I have not decided that I definetely need this functionality. But *if* I need it, the limitation of 1-to-1 substitutions is totally fine. It's very cool to know that this is *possible*.

behdad:
I think I like this change even if you end up not using it. We just need to test it at least.

khaledhosny:
I tested this API and it works as advertised and I think it does the trick for what Fontra wants.

PR #5767, opened 2026-02-28
Avar2 instancing

behdad:
Agentic translation of FontTools PR https://github.com/fonttools/fonttools/pull/4045
I haven't checked their work.


qxliu76:
I just read your design doc again and now have a basic understanding of the algorithm, the avar2 instancing is indeed a hard problem, and you work is impressive! I'll try to review it but I plan to check fonttools' code change first to get better understanding on more details. Thanks very much for this PR!

behdad:
> I just read your design doc again and now have a basic understanding of the algorithm, the avar2 instancing is indeed a hard problem, and you work is impressive! I'll try to review it but I plan to check fonttools' code change first to get better understanding on more details. Thanks very much for this PR!

Thanks. I know this is very dense design and code but I think that's necessary for the task.

qxliu76:
Code changes mostly LGTM, nothing is blocking except for the part that `plan->has_avar2` is determined by avar->[has_v2_data()](https://github.com/harfbuzz/harfbuzz/blob/af85ffc449edf34f426312856fa000c0351b5633/src/hb-ot-var-avar-table.hh#L415), which only checks on avar table version. I think we should also check if VarStore is not empty before we set `plan->has_avar2` to true? 

I do find a few nits, see comments below. Thanks very much for your amazing work!

src/hb-ot-layout-common.hh
```diff
@@ -3490,7 +3551,15 @@ struct ItemVariationStore
     ItemVariationStore *varstore_prime = c->serializer->start_embed<ItemVariationStore> ();
     if (unlikely (!varstore_prime)) return_trace (false);
 
-    varstore_prime->serialize (c->serializer, this, inner_maps);
+    /* avar2 partial instancing: cull unreachable regions. */
+    hb_set_t dead_regions;
+    if (c->plan->has_avar2)
+      collect_dead_regions (c->plan->axes_old_index_tag_map,
```
qxliu76: should we check return value here and return early in case of false?

src/hb-ot-var-avar-table.hh
```diff
@@ -414,6 +557,39 @@ struct avar
...
+    *varidx_map = &(this+v2.varIdxMap);
+    return true;
+#else
+    return false;
+#endif
+  }
+
+  /* Apply avar v1 segment maps only, deliberately omitting avar v2 processing.
+   * Used during avar2 partial instancing to compute intermediate-space coords. */
+  bool map_coords_v1_only (float *coords, unsigned int coords_length) const
```
qxliu76: Optional, since the v1 code is the same, can we combine this one with map_coords_2_14(), with an additional argument indicating if v1 only.

src/hb-ot-var-avar-table.hh
```diff
@@ -487,14 +671,526 @@ struct avar
         hb_tag_t *axis_tag;
         if (!c->plan->axes_old_index_tag_map.has (i, &axis_tag))
           return_trace (false);
-        if (!map->subset (c, *axis_tag))
-          return_trace (false);
+
+        if (c->plan->has_avar2 &&
+            c->plan->user_axes_location.has (*axis_tag) &&
```
qxliu76: a possible minor optimization: use has(*axis_tag, *axis_location) here, then no need to use get()

src/hb-ot-var-avar-table.hh
```diff
@@ -487,14 +671,526 @@ struct avar
...
+      else if (!varidx_ref_count.set (varidx, 1))
+        return false;
+    }
+    for (unsigned i = 0; i < axisCount; i++)
+    {
+      hb_tag_t *axis_tag_ptr;
+      /* Only axes that will receive offset compensation (restricted or
+       * pinned) can contaminate a shared row. */
+      if (!c->plan->axes_index_map.has (i) ||
+          !c->plan->axes_old_index_tag_map.has (i, &axis_tag_ptr) ||
```
qxliu76: I think `!c->plan->axes_old_index_tag_map.has (i, &axis_tag_ptr)` is an error condition, we should error here and return early.

src/hb-ot-var-avar-table.hh
```diff
@@ -487,14 +671,526 @@ struct avar
...
+    }
+
+    /* 6. Add offset compensation tuples.
+     * Track processed (outer,inner) pairs to avoid adding duplicate biases
+     * when multiple axes share the same varIdx. */
+    hb_set_t processed_varidxes;
+    for (unsigned i = 0; i < axisCount; i++)
+    {
+      hb_tag_t *axis_tag_ptr;
+      if (!c->plan->axes_old_index_tag_map.has (i, &axis_tag_ptr))
```
qxliu76, line 887: same here, it's an error condition, we need to return false instead of continuing.

src/hb-ot-var-avar-table.hh
```diff
@@ -487,14 +671,526 @@ struct avar
...
+       * contribution is baked into the variation tables instead. */
+      if (!c->plan->axes_index_map.has (i))
+        continue;
+
+      Triple *new_user;
+      if (c->plan->user_axes_location.has (axis_tag, &new_user))
+      {
+        /* This axis is being restricted or pinned */
+        Triple *old_int;
+        if (!c->plan->old_intermediates.has (axis_tag, &old_int))
```
qxliu76, line 901: I guess this is also an error condition that we should just return false?

src/hb-ot-var-avar-table.hh
```diff
@@ -487,14 +671,526 @@ struct avar
...
+
+    /* Serialize DeltaSetIndexMap (push a new object). Entries cover the
+     * RETAINED axes only (self-contained pinned axes are removed from
+     * fvar), in their retained order. Count what the write loop below will
+     * actually emit: a malformed font's avar may have fewer axes than fvar,
+     * and claiming unwritten entries would map those axes to varIdx 0. */
+    unsigned retained_axis_count = 0;
+    for (unsigned i = 0; i < axisCount; i++)
+      if (c->plan->axes_index_map.has (i))
+	retained_axis_count++;
```
qxliu76: seems this block is equivalent to `unsigned retained_axis_count = c->plan->axes_index_map.get_population()`?

src/hb-ot-var-avar-table.hh
```diff
@@ -487,14 +671,526 @@ struct avar
...
+        for (unsigned w = width; w > 0;)
+        {
+          p[--w] = u;
+          u >>= 8;
+        }
+        p += width;
+      }
+
+      packed_map = c->serializer->pop_pack ();
+    }
```
qxliu76: can we use DeltaSetIndexMap[ serialize() ](https://github.com/harfbuzz/harfbuzz/blob/af85ffc449edf34f426312856fa000c0351b5633/src/hb-ot-layout-common.hh#L3654)method instead? So we just need one line code serialize_serialize() for this block

src/hb-ot-var-avar-table.hh
```diff
@@ -487,14 +671,526 @@ struct avar
...
+                           item_vars.has_long_word (),
+                           c->plan->axis_tags,
+                           item_vars.get_region_list (),
+                           item_vars.get_vardata_encodings ()))
+      {
+        c->serializer->pop_discard ();
+        return false;
+      }
+      packed_store = c->serializer->pop_pack ();
+    }
```
qxliu76: can we use serialize_serialize() instead?

src/hb-ot-var-common.hh
```diff
@@ -1790,6 +1909,79 @@ struct item_variations_t
     return true;
   }
 
+  /* Like instantiate_tuple_vars but does NOT call build_region_list().
+   * Caller can add more tuples between this call and build_region_list(). */
+  bool instantiate_tuple_vars_no_region_build (
```
qxliu76: optional, can we combine this function with [instantiate_tuple_vars()](https://github.com/harfbuzz/harfbuzz/blob/af85ffc449edf34f426312856fa000c0351b5633/src/hb-ot-var-common.hh#L1781) and pass in additional argument indicating if we want to build region list

src/hb-subset-plan-var.cc
```diff
@@ -253,34 +419,125 @@ normalize_axes_location (hb_face_t *face, hb_subset_plan_t *plan)
...
+        unsigned i = _.first;
+        hb_tag_t axis_tag = _.second.get_axis_tag ();
+        if (plan->user_axes_location.has (axis_tag))
+        {
+          /* Store intermediate-space coords for offset compensation and
+           * for avar's own segment-map renormalization and IVS rebasing. */
+          plan->old_intermediates.set (axis_tag, Triple ((double) normalized_mins[i],
+                                                         (double) normalized_defaults[i],
+                                                         (double) normalized_maxs[i]));
+          plan->avar2_axes_location.set (axis_tag, Triple ((double) normalized_mins[i],
```
qxliu76: can we remove `avar2_axes_location` and just use `old_intermediates`, seems they're exactly the same

src/hb-subset-plan-var.cc
```diff
@@ -253,34 +419,125 @@ normalize_axes_location (hb_face_t *face, hb_subset_plan_t *plan)
...
+      for (const auto& _ : + hb_enumerate (axes))
+      {
+        unsigned i = _.first;
+        hb_tag_t axis_tag = _.second.get_axis_tag ();
+        if (plan->avar2_self_contained.has (axis_tag))
+          continue;
+        plan->axes_index_map.set (i, retained_axis_idx++);
+        plan->axis_tags.push (axis_tag);
+      }
+      plan->all_axes_pinned = false;
```
qxliu76: This line seems redundant and can be deleted

src/hb-subset-plan-var.cc
```diff
@@ -55,12 +55,178 @@
...
+   /* Old-intermediate input box per axis. */
+   hb_hashmap_t<hb_tag_t, hb_pair_t<double, double>> box;
+   for (const auto &axis : axes)
+   {
+     hb_tag_t tag = axis.get_axis_tag ();
+     double lo = -1.0, hi = +1.0;
+     Triple *old_int;
+     if (plan->user_axes_location.has (tag))
+     {
+       if (!plan->old_intermediates.has (tag, &old_int)) return;
```
qxliu76: This seems to be an error condition, can we return false here? We can change the function return a bool value.

src/hb-subset-plan-var.cc
```diff
@@ -55,12 +55,178 @@
...
+       lo = hb_min (old_int->minimum, old_int->maximum);
+       hi = hb_max (old_int->minimum, old_int->maximum);
+     }
+     else if (axis.is_hidden ())
+       lo = hi = 0.0; /* private axis: intermediate is always 0 */
+     if (!box.set (tag, hb_pair (lo, hi))) return;
```
qxliu76: same, an return false here.

src/hb-subset-plan-var.cc
```diff
@@ -55,12 +55,178 @@
...
+
+   hb_vector_t<hb_hashmap_t<hb_tag_t, Triple>> regions;
+   if (!var_store->get_region_list ().get_var_regions (plan->axes_old_index_tag_map, regions))
+     return;
+
+   for (unsigned i = 0; i < axes.length; i++)
+   {
+     hb_tag_t tag = axes[i].get_axis_tag ();
+     hb_pair_t<double, double> *identity;
+     if (!box.has (tag, &identity)) return;
```
qxliu76: Ditto, return false here?

src/hb-subset-plan-var.cc
```diff
@@ -55,12 +55,178 @@
...
+       lo = hb_min (old_int->minimum, old_int->maximum);
+       hi = hb_max (old_int->minimum, old_int->maximum);
+     }
+     else if (axis.is_hidden ())
+       lo = hi = 0.0; /* private axis: intermediate is always 0 */
+     if (!box.set (tag, hb_pair (lo, hi))) return;
+   }
+
+   hb_vector_t<hb_hashmap_t<hb_tag_t, Triple>> regions;
+   if (!var_store->get_region_list ().get_var_regions (plan->axes_old_index_tag_map, regions))
```
qxliu76, line 130: ditto

src/hb-subset-plan-var.cc
```diff
@@ -55,12 +55,178 @@
...
+   for (const auto &axis : axes)
+   {
+     hb_tag_t tag = axis.get_axis_tag ();
+     double lo = -1.0, hi = +1.0;
+     Triple *old_int;
+     if (plan->user_axes_location.has (tag))
+     {
+       if (!plan->old_intermediates.has (tag, &old_int)) return;
+       lo = hb_min (old_int->minimum, old_int->maximum);
+       hi = hb_max (old_int->minimum, old_int->maximum);
```
qxliu76: I think there's no need to clamp here, they should be clamped already

src/hb-subset-plan-var.cc
```diff
@@ -55,12 +55,178 @@
...
+       else
+       {
+	 const OT::VarData &vd = var_store->get_sub_table (outer);
+	 if (inner >= vd.get_item_count ()) bounded = false;
+	 else
+	 {
+	   const OT::HBUINT8 *delta_bytes = vd.get_delta_bytes ();
+	   unsigned row_size = vd.get_row_size ();
+	   unsigned region_count = vd.get_region_index_count ();
+	   for (unsigned r = 0; r < region_count && bounded; r++)
```
qxliu76: seems no need to check `bounded` value here?

behdad:
> Code changes mostly LGTM, nothing is blocking except for the part that `plan->has_avar2` is determined by avar->[has_v2_data()](https://github.com/harfbuzz/harfbuzz/blob/af85ffc449edf34f426312856fa000c0351b5633/src/hb-ot-var-avar-table.hh#L415), which only checks on avar table version. I think we should also check if VarStore is not empty before we set `plan->has_avar2` to true?

Actually no. Current way is better: if the font / system is avar2-capable, we do want to partial-instance using the avar2 technique, since it's guaranteed to make the font not larger. The non-avar2 partial-instancing is a liability and inferior.

> I do find a few nits, see comments below. Thanks very much for your amazing work!

Thanks. I'll go over them.

qxliu76:
> Actually no. Current way is better: if the font / system is avar2-capable, we do want to partial-instance using the avar2 technique, since it's guaranteed to make the font not larger. The non-avar2 partial-instancing is a liability and inferior.

uh, I think my review comments are based on assuming non-empty VarStore, some might be ignored then:)

src/hb-ot-layout-common.hh
behdad: Why? Returning `false` is an error, no?

behdad:
@qxliu76 Thanks for the feedback. I tried addressing them all. PTAL.

qxliu76:
LGTM. Thank you! I'm actually thinking you should give us a talk on how to use AI, it seems the code quality is much better than others'(also generated by AI)

qxliu76 (approved):

qxliu76:
test failures seem not related with this PR?

behdad:
Thanks.

PR #5778, opened 2026-03-04
Add meson options to disable extra libraries

### LINKED ISSUE #5769: Add meson options to disable the new libraries
Add meson options to disable building harfbuzz-raster and harfbuzz-vector libraries, and maybe also harfbuzz-subset while at it.

khaledhosny:
* Fixes #5769

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -423,6 +423,12 @@ libs = [libharfbuzz]
 if not get_option('subset').disabled()
   libs += [libharfbuzz_subset]
 endif
+if not get_option('raster').disabled()
+  libs += [libharfbuzz_raster]
+endif
+if not get_option('vector').disabled()
+  libs += [libharfbuzz_vector]
+endif
 alias_target('libs', libs)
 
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -419,7 +419,11 @@ endif
 configure_file(output: 'config.h', configuration: conf)
 
 alias_target('lib', libharfbuzz)
-alias_target('libs', libharfbuzz, libharfbuzz_subset)
+libs = [libharfbuzz]
+if not get_option('subset').disabled()
+  libs += [libharfbuzz_subset]
+endif
```
khaledhosny, line 425: Should we add the other libs here as well?

behdad: I think so, yes.

behdad: What else is there? libharfbuzz-cairo, libharfbuzz-icu are integrations I don't care much about here. Do we have anything else? Maybe add all for completion. Up to you. Thanks.

khaledhosny: I think it is good as it is, the others are integration libraries while these are “core” libraries, if we were to make that distinction.

behdad:
LGTM. Thanks.

PR #5779, opened 2026-03-04
[vector] Move documentation to .cc when applicable

### LINKED ISSUE #5771: Clean up the new documentation
Many functions are documented in the headers, but we usually document these in the sources. There might be other minor issues as well.

khaledhosny:
* Fixes #5771


behdad:
Thank you.

PR #5782, opened 2026-03-04
[meson] Add thread_dep to harfbuzz-subset as well

khaledhosny:
Adapted patch from:
https://git.libreoffice.org/core/+/8f6d4fb1a4bbb689ca748dd033ceaa87d94a1a3d%5E%21


behdad:
LGTM.

Issue #2544, opened 2020-07-05
hb-set-fuzzer inefficiency


behdad:
It just doesn't test anything for the cast majority of fuzzer input. While fuzzer does learn the effective paths, let's help it more.

First, for instruction type, change the switch argument to: `instruction.operation % NUM_OPERATIONS`.

Then just split the available values into the two sets by looking up one integer from fuzzer and use that module size for set1 and the rest for set2.

ebraminio:
have a look at https://github.com/harfbuzz/harfbuzz/issues/2302 also maybe we can think about some fuzzer that tests both

behdad:
Yes. We'll get there. Just small fix for now.

PR #5786, opened 2026-03-05
[fuzzer] Fix hb-set-fuzzer reading pointer address instead of input data

### LINKED ISSUE #5787: hb-set-fuzzer reads pointer address instead of fuzzer input due to reinterpret_cast bug
## Summary

This is **not** a bug in harfbuzz. It is a bug in `hb-set-fuzzer` that causes the fuzzer to read the address of the `data` pointer (determined by ASLR) instead of the actual fuzzer input. As a result, the `operation` and `first_set_size` fields are effectively random constants that the fuzzer cannot control, making the fuzzer nearly useless — it achieves only **0.05% line coverage** vs **1.47%** after fix (a 29x improvement).

## Bug Description

In `test/fuzzing/hb-set-fuzzer.cc`, line 52:

```cpp
const instructions_t &instructions = reinterpret_cast<const instructions_t &> (data);
```

Here `data` is `const uint8_t *` — a pointer variable on the stack. In C++, `reinterpret_cast<T&>(x)` is equivalent to `*reinterpret_cast<T*>(&x)`. So this reads `sizeof(instructions_t)` bytes starting from the address of the pointer variable itself (`&data`, on the stack), **not** from the buffer that `data` points to.

This means:
- `instructions.operation` is determined by the low byte of the pointer address (ASLR)
- `instructions.first_set_size` is determined by the next 4 bytes of the pointer address
- The fuzzer has **zero control** over which set operation is tested or how the input is split

The compiler warned about this — `-Wstrict-aliasing` flagged this exact line — but the warning was suppressed with `#pragma GCC diagnostic ignored` in commit a470b0b2 instead of fixing the root cause.

### History

- 2020-02-06: `hb-set-fuzzer` added (commit bca9bc6b)
- 2020-07-04: Compiler warning suppressed instead of fixing the bug (commit a470b0b2)
- 2020-07-05: @behdad opened #2544 noting the fuzzer is inefficient, but did not identify the `reinterpret_cast` bug as the root cause
- 2022-02-13: #2544 closed without fixing this bug
- Present: Bug still exists

## Suggested Fix

```diff
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wstrict-aliasing"
-  const instructions_t &instructions = reinterpret_cast<const instructions_t &> (data);
-#pragma GCC diagnostic pop
+  const instructions_t &instructions = reinterpret_cast<const instructions_t &> (*data);
```

Dereference `data` to read from the input buffer. The `#pragma` is no longer needed since the fix eliminates the strict-aliasing violation.

Fix submitted in #5786.

## Experimental Evidence

10-minute fuzzing coverage comparison (libFuzzer + coverage instrumentation):

| Version | Lines Covered | Line Coverage | Functions Covered | Function Coverage |
|---------|--------------|---------------|-------------------|-------------------|
| Original | 15 | 0.05% | 2 | 0.04% |
| Fixed | 486 | 1.47% | 144 | 2.68% |
| **Improvement** | **+471** | **29.4x** | **+142** | **72x** |

The original fuzzer covers only 2 functions in 10 minutes — essentially the entry point and `create_set`. The fixed version explores all four set operations and their internal implementations.

OwenSanzas:
## Summary

`hb-set-fuzzer` has a `reinterpret_cast` bug that causes it to read the **address of the `data` pointer** (determined by ASLR) instead of the actual fuzzer input. This makes `operation` and `first_set_size` uncontrollable by the fuzzer, rendering it nearly useless.

This is **not** a bug in harfbuzz itself — it is a bug in the fuzzer harness.

Fixes #5787

## What's Wrong

```cpp
const instructions_t &instructions = reinterpret_cast<const instructions_t &> (data);
```

In C++, `reinterpret_cast<T&>(x)` is equivalent to `*reinterpret_cast<T*>(&x)`. Since `data` is a `const uint8_t *`, this reads bytes from `&data` (the pointer variable on the stack), not from the buffer `data` points to.

The compiler warned about this (`-Wstrict-aliasing`), but the warning was suppressed with `#pragma` in commit a470b0b2 instead of fixing the root cause.

## Fix

```diff
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wstrict-aliasing"
-  const instructions_t &instructions = reinterpret_cast<const instructions_t &> (data);
-#pragma GCC diagnostic pop
+  const instructions_t &instructions = reinterpret_cast<const instructions_t &> (*data);
```

Dereference `data` so the cast reads from the input buffer. The `#pragma` is no longer needed.

## Evidence

10-minute fuzzing coverage comparison:

| Version | Lines Covered | Line Coverage | Functions Covered | Function Coverage |
|---------|--------------|---------------|-------------------|-------------------|
| Original | 15 | 0.05% | 2 | 0.04% |
| Fixed | 486 | 1.47% | 144 | 2.68% |
| **Improvement** | **+471** | **29x** | **+142** | **72x** |

## Related

- #2544 noted the fuzzer was inefficient but did not identify this `reinterpret_cast` bug as the root cause.

### DIFF
--- a/test/fuzzing/hb-set-fuzzer.cc
+++ b/test/fuzzing/hb-set-fuzzer.cc
@@ -47,10 +47,7 @@ extern "C" int LLVMFuzzerTestOneInput (const uint8_t *data, size_t size)
   if (size > MAX_INPUT_SIZE)
     return 0;
 
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wstrict-aliasing"
-  const instructions_t &instructions = reinterpret_cast<const instructions_t &> (data);
-#pragma GCC diagnostic pop
+  const instructions_t &instructions = reinterpret_cast<const instructions_t &> (*data);
   data += sizeof (instructions_t);
   size -= sizeof (instructions_t);
 


behdad:
Thanks.

OwenSanzas:
Thanks for the approval!

Issue #5788, opened 2026-03-05
Shared library version name issue for AIX


shubhamhii:
In AIX, shared libraries are conventionally packaged and distributed within static archive (.a) files, following the naming convention lib<library_name>.a(lib<library_name>.so.<major_version>). 
For example, libharfbuzz.a(libharfbuzz.so.0) encapsulates the shared object libharfbuzz.so.0 within an archive member.

AIX's archive bundled shared library model enforces strict SONAME backward compatibility. Major version increments require coordinated rebuild of all dependent executables across the system, making it a deployment-breaking change.

khaledhosny:
What is the issue then?

khaledhosny:
https://github.com/harfbuzz/harfbuzz?tab=readme-ov-file#api-stability

behdad:
The soversion has not changed, and will never change. It's at 0. Always. The major version of the library is bumped for major features. We don't use semver, otherwise we would have been stuck in 0.x.y forever because of the API/ABI stability guarantees.

PR #5789, opened 2026-03-06
Fix for AIX shared library name change

shubhamhii:
#5788 

Reason:

Due to the naming change, resulting from minor version increments, causes failures in runtime dependency resolution. 
Earlier: libharfbuzz.a(libharfbuzz.so.0)
Now: libharfbuzz.a(libharfbuzz.so.0.61230.0)

Fix:

hardcoded version string to lock shared library object name for aix platform

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -640,6 +640,10 @@ else
   hb_so_version = '0'
 endif
 
+if host_machine.system() == 'aix'
+  version = '0'
+endif
+
 if get_option('fuzzer_ldflags') != ''
   extra_hb_cpp_args += ['-DHB_CUSTOM_MALLOC']
   hb_sources += hb_failing_alloc_sources


khaledhosny:
I don’t understand how setting version to zero achieves what you say it does.

shubhamhii:
In AIX , the shared libraries are archived  like below. 
```
# file libharfbuzz.a
libharfbuzz.a: archive (big format)
# ar -tv libharfbuzz.a
rwxr-xr-x   203/1     2111826 Apr 20 02:23 2023 libharfbuzz.so.0
```
So with the new meson build, the shared library became " libharfbuzz.a(libharfbuzz.so.0.61230.0) " . So to avoid this and keep the same " libharfbuzz.a(libharfbuzz.so.0)" naming , we need these changes.

khaledhosny:
Can you please explain in detail what is the issue you are having, because I' really confused about your description, and I also don't think setting version to zero is the right fix.

shubhamhii:
Hi Khaled,

The issue that we are facing with the shared object complete name.
In the latest version,  we are getting the library names as per the standard followed. Ex: libharfbuzz.so.0.61230.0

Due to the convention followed by AIX, we would like to keep the shared object name without the minor version numbering update. Ex: libharfbuzz.so.0

Hence in the patch,  we have hardcoded the value of version for AIX to remove the minor version number updates.

behdad:
How was this not a problem before?

shubhamhii:
> How was this not a problem before?

Yes, we were facing it earlier too. We were changing the name manually till now. If possible, we would like to change the shared object name format for AIX.

Thanks,
Shubham

behdad:
@khaledhosny Let's decide something here and move on.

khaledhosny:
Lets merge it, it only affects AIX. If it breaks something and someone complains loudly, they will have to figure out a fix.

PR #5798, opened 2026-03-08
raster: add PNG paint_image support via libpng

behdad:
Short codex session.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -461,10 +461,6 @@ build_summary = {
      'FreeType': conf.get('HAVE_FREETYPE', 0) == 1,
      'Fontations': conf.get('HAVE_FONTATIONS', 0) == 1,
     },
-  'Dependencies used for command-line utilities':
-    {'Cairo': conf.get('HAVE_CAIRO', 0) == 1,
-     'Chafa': conf.get('HAVE_CHAFA', 0) == 1,
-    },
   'Additional shapers':
     {'Graphite2': conf.get('HAVE_GRAPHITE2', 0) == 1,
      'WebAssembly (experimental)': conf.get('HAVE_WASM', 0) == 1,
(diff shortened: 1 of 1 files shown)


khaledhosny:
I think `hb-raster` util should also support PNG output, if possible (and also maybe, as an extra, refactor the terminal inline image support from `hb-view` and use it here).

khaledhosny:
I’m not sure what this is, but I’m noting that with Apple Color Emoji, `hb-view` rendering is smoother:
```
$ ./util/hb-raster /System/Library/Fonts/Apple\ Color\ Emoji.ttc '😅'
```
Unnamed file

```
$ ./util/hb-view /System/Library/Fonts/Apple\ Color\ Emoji.ttc '😅' --font-size=1000
```
iTerm2 PmeJ6g

behdad:
> I think `hb-raster` util should also support PNG output, if possible

Right. I wonder if, like cairo, we should add read/write from PNG functions to the library.

> (and also maybe, as an extra, refactor the terminal inline image support from `hb-view` and use it here).

Agreed.

khaledhosny:
> > I think `hb-raster` util should also support PNG output, if possible
> 
> Right. I wonder if, like cairo, we should add read/write from PNG functions to the library.

It doesn’t hurt and the libpng dependency is optional already.

behdad:
> I’m not sure what this is, but I’m noting that with Apple Color Emoji, `hb-view` rendering is smoother:

Sure. Rather expected. Our image upscaling is very primitive. I'll see if I can improve it.

behdad:
I'm also not sure if we are picking the best strike size...

behdad:
> > I’m not sure what this is, but I’m noting that with Apple Color Emoji, `hb-view` rendering is smoother:
> 
> Sure. Rather expected. Our image upscaling is very primitive. I'll see if I can improve it.

Done. Please check.

image

khaledhosny:
> > > I’m not sure what this is, but I’m noting that with Apple Color Emoji, `hb-view` rendering is smoother:
> > 
> > 
> > Sure. Rather expected. Our image upscaling is very primitive. I'll see if I can improve it.
> 
> Done. Please check.

Looks good now.

Issue #5799, opened 2026-03-08
[subset] Setting old_to_new_glyph_mapping can produce invalid CFF table


khaledhosny:
Test code:
```c
#include <stdio.h>
#include <stdlib.h>

#include <hb.h>
#include <hb-subset.h>

int
main (int argc, char **argv)
{
  if (argc < 4)
  {
    fprintf (stderr, "Usage: %s <font> <output_font> <num_glyphs>\n", argv[0]);
    return 1;
  }

  const char *font_path = argv[1];
  const char *output_path = argv[2];
  int num_glyphs = atoi (argv[3]);

  hb_face_t *face = hb_face_create_from_file_or_fail (font_path, 0);
  hb_subset_input_t *input = hb_subset_input_create_or_fail ();

  hb_set_t *glyphs = hb_subset_input_glyph_set (input);
  if (num_glyphs)
  {
    hb_set_add_range (glyphs, 0, num_glyphs - 1);

    hb_map_t *mapping = hb_subset_input_old_to_new_glyph_mapping (input);
    hb_codepoint_t gid = HB_SET_VALUE_INVALID;
    while (hb_set_next (glyphs, &gid))
    {
      if (!gid) continue;
      hb_map_set (mapping, gid, num_glyphs - gid);
    }
  }
  else
  {
    hb_set_clear (glyphs);
    hb_set_invert (glyphs);
  }

  hb_face_t *subset_face = hb_subset_or_fail (face, input);
  hb_blob_t *subset_blob = hb_face_reference_blob (subset_face);
  unsigned int length;
  const char *data = hb_blob_get_data (subset_blob, &length);

  FILE *f = fopen (output_path, "wb");
  fwrite (data, 1, length, f);
  fclose (f);

  return 0;
}
```

Testing with [`NotoSansCJKjp-Regular.otf`](https://github.com/notofonts/noto-cjk/blob/main/Sans/OTF/Japanese/NotoSansCJKjp-Regular.otf):
```
$ ./susbet-test NotoSansCJKjp-Regular.otf output.otf 400
```

The `output.otf` can’t be de-compiled with `ttx`, due to invalid local subr index:
```
$ ttx -t CFF output.otf
Dumping "output.otf" to "output.ttx"...
Dumping 'CFF ' table...
ERROR: Error in charstring 29
ERROR: Unhandled exception has occurred
Traceback (most recent call last):
[...]

PR #5800, opened 2026-03-09
[subset/cff1] Fix subsetting font dicts with old_to_new_glyph_mapping

khaledhosny:
The code was assuming old_glyph to be monotonically increasing and cashed the font dict accordingly, but that is not the case with the test font and code.

I’m not entirely sure about this fix, but not caching the font dict seems to fix the issue for me.

* Fixes https://github.com/harfbuzz/harfbuzz/issues/5799


behdad:
Looks right to me.

PR #5808, opened 2026-03-09
[arabic] Improve stch measurement pass

quangIO:
Use checked arithmetic when calculating the number of extra glyphs needed during stch processing. Includes a new hb_unsigned_add_overflows helper in hb-algs.hh.


behdad:
Thanks.

PR #5809, opened 2026-03-10
Add raster and vector fuzzers

behdad:


eXpl0it3r:
*Just posting it here, since this contains the related commit, but I can also open an separate issue if desired.*

We're getting a build error for SFML when building on macOS as Framework build:

```
hb-raster-image.cc:49:3: error: use of undeclared identifier 'png_longjmp'
```

Could it be, that we're picking up an unsupported libpng version?

https://github.com/SFML/SFML/actions/runs/23438542727/job/68191630567

behdad:
> _Just posting it here, since this contains the related commit, but I can also open an separate issue if desired._

Thanks. Please file a new issue and keep us posted if you find any clues. I'm investigating.

behdad:
> > _Just posting it here, since this contains the related commit, but I can also open an separate issue if desired._
> 
> Thanks. Please file a new issue and keep us posted if you find any clues. I'm investigating.

Do you know which version of libpng this hits?

eXpl0it3r:
SFML doesn't use libpng, so HarfBuzz must find some other "rough" libpng on the system, but I don't know the available version. And neither is it listed in the [Runner Image software list](https://github.com/actions/runner-images/blob/macos-14-arm64/20260302.0147/images/macos/macos-14-arm64-Readme.md).

I've opened a dedicated issue #5867

Issue #5810, opened 2026-03-10
Wierd symbols in harfbuzz-raster library


aelmahmoudy:
When building harfbuzz on Debian unstable, I found that libharfbuzz-raster exports those wierdly-named symbols which aren't mentioned in the NEWS file:
```
_ZN13hb_svg_defs_t13add_clip_pathE10hb_array_tIKcERK22hb_svg_clip_path_def_t
_ZN13hb_svg_defs_t14add_id_mappingEP12hb_hashmap_tI10hb_array_tIKcEjLb0EES3_j
_ZN13hb_svg_defs_tD1Ev
_ZN13hb_svg_defs_tD2Ev
_ZN17hb_raster_image_t14composite_fromEPKS_25hb_paint_composite_mode_t
_ZN17hb_raster_image_t15bytes_per_pixelE18hb_raster_format_t
_ZN17hb_raster_image_t5clearEv
_ZN17hb_raster_image_t9configureE18hb_raster_format_t19hb_raster_extents_t
_ZNK13hb_svg_defs_t13find_gradientE10hb_array_tIKcE
_ZNK13hb_svg_defs_t14find_clip_pathE10hb_array_tIKcE
_ZNK17hb_raster_image_t10get_bufferEv
```

behdad:
These are internal C++ symbols. I'm surprised our tests didn't catch them. I'll take care of them. Thanks for the report.

PR #5812, opened 2026-03-10
Fix check libraries

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/5810

### DIFF
--- a/src/hb-raster-image.hh
+++ b/src/hb-raster-image.hh
@@ -43,14 +43,14 @@ struct hb_raster_image_t
   hb_raster_extents_t  extents     = {};
   hb_raster_format_t   format      = HB_RASTER_FORMAT_A8;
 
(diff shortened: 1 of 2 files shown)


src/check_helpers.py
```diff
@@ -0,0 +1,19 @@
...
+
+# harfbuzz-icu links to libstdc++ because icu does.
+CHECK_LIBSTDCXX_LIBS = [
+    "harfbuzz",
+    "harfbuzz-subset",
+    "harfbuzz-raster",
+    "harfbuzz-vector",
+    "harfbuzz-gobject",
+    "harfbuzz-cairo",
+]
```
khaledhosny, line 19: You can even re-order the two lists and make `CHECK_SYMBOL_LIBS = CHECK_LIBSTDCXX_LIBS + ["harfbuzz-icu"]`, so we have a single list to update.

PR #5815, opened 2026-03-10
Code of ai conduct.md

behdad:


CODE_OF_AI_CONDUCT.md
```diff
@@ -0,0 +1,48 @@
...
+responsible for the changes they submit.
+
+## Requirements
+
+- Use AI as a tool, not as an authority. Verify behavior, correctness, and
+  compatibility yourself before submitting changes.
+- Keep changes narrow. Do not use AI to generate broad rewrites, drive-by
+  refactors, or style-only churn.
+- Preserve HarfBuzz's public API and ABI unless a change explicitly calls for
+  it.
```
khaledhosny, line 18: How does that fit with the API/ABI stability policy?

khaledhosny (approved):
Sounds reasonable and balanced to me.

CODE_OF_AI_CONDUCT.md
behdad: I was thinking changing unreleased API, or unstable APIs. I can try to be more explicit. Ultimately the release engineer has to review all API changes.

khaledhosny: Sounds good.

CODE_OF_AI_CONDUCT.md
```diff
@@ -0,0 +1,48 @@
+# Code of AI Conduct
```
rsheeter, line 1: A minor addition that might be nice is to note related codes and how we perceive HB to align (or not) with them and why. An appendix perhaps, since it isn't really guidance itself, just rationale for it.

CODE_OF_AI_CONDUCT.md
```diff
@@ -0,0 +1,48 @@
...
+
+This project accepts AI-assisted contributions, but contributors remain fully
+responsible for the changes they submit.
```
rsheeter: I would read this as meaning we specifically do NOT want fully independent AI contributions, e.g. I turned an agent loose and didn't look isn't appreciated. That's fine if it's intentional, just want to be sure it's on purpose.

jfkthame: That was certainly my reading of what Behdad wrote (and I'd strongly agree, fwiw). Before the HB maintainers are asked to review a contribution, a human contributor is expected to take full responsibility for its contents.

rsheeter: Fwiw I agree as well :)

behdad: Right. I expect soon there will be fuzzer-like services that would automatically file bug fixes & security fixes without human interaction. But we can deal with that when it comes up.

rsheeter: You could potentially add a carveout for if a maintainer accepts a contribution from an automaton, such as an autobugfix tool, they take personal responsibility for it? - currently it reads that the line is hard no on such things.

behdad: PTAL.

CODE_OF_AI_CONDUCT.md
behdad: I'm not sure I understand. Can you please elaborate?

rsheeter: Explain what the rationale for the policy is and note if/how it was informed by e.g. Linux policy, etc

behdad: Right. I got confused with "code" meaning computer code.

behdad: Added a link to issue instead.

behdad:
Thanks all. I think we can merge this now, and revise as needed.

behdad:
err. this got merged somehow already.

PR #5818, opened 2026-03-10
[subset] Reject malformed normalized axis ranges

behdad:
Reject invalid normalized axis triples during subset plan construction so malformed avar remapping does not reach tuple instancing.

Keep a defensive guard in the instancer solver as a backstop against future bad callers.

Tested:
- ./build/src/test-instancer-solver
- ./build/src/test-item-varstore
- ./build/src/test-tuple-varstore

Fixes: https://oss-fuzz.com/testcase-detail/6119821501857792
Assisted-by: OpenAI Codex


behdad:
This didn't came up before because the subset fuzzer used to only test pinned axes, not rebasing.

PR #5819, opened 2026-03-11
util: Add missing chafa_dep to hb-raster

heftig:


jpalus:
Got the same issue with missing `chafa_dep`. 7623e55 added include of `helper-cairo-ansi.hh` which in turn includes `chafa.h`:
https://github.com/harfbuzz/harfbuzz/blob/68d19d6964da52dfbe714527370a55b4ca43f4bc/util/helper-cairo-ansi.hh#L36-L37
and therefore `hb-raster` needs `chafa_dep` too.

khaledhosny:
Thanks!

Issue #5820, opened 2026-03-11
Support compressed SVG font data in hb-raster


mmozeiko:
See https://learn.microsoft.com/en-us/typography/opentype/spec/svg
> SVG documents within an OpenType SVG table may either be plain text or **gzip-encoded**, and applications that support the SVG table **must support both.**
> 
> The gzip format is defined in RFC 1952, “GZIP file format specification version 4.3”, available at http://www.ietf.org/rfc/rfc1952.txt. Within a gzip-encoded SVG document, the deflate compression method (defined in RFC 1951, available at http://www.ietf.org/rfc/rfc1951.txt) must be used. Thus, the first three bytes of the gzip-encoded document header must be 0x1F, 0x8B, 0x08.

Easy way to identify compressed data is from first three bytes of data listed above.

Now that harfbuzz supports libpng for rasterizing bitmap fonts, adding zlib dependency for gzip decompression should not cost much.

Example font that does svg compression is emojione-svg.otf from https://github.com/joypixels/emojione/blob/master/extras/fonts/README.md#emojione-svg-based-color-fonts

behdad:
Hahahah. @khaledhosny and I were talking about it yesterday, and we thought we'll wait until somebody asks for it.

behdad:
Also curious if you're interested in using hb-raster and where.

behdad:
Heads up: I have decided to remove SVG-font support from raster/vector libraries. SVG is a huge surface to support, and adds lots of complexity to the code for a small payoff.

PR #5824, opened 2026-03-11
[raster] Support gzip-compressed SVG documents

behdad:
The raster SVG path treated hb_ot_color_glyph_reference_svg() blobs as plain XML, even though SVG table documents may legally be gzip encoded. That caused hb-raster to reject fonts that store compressed SVG documents.

Detect gzip payloads in the raster SVG renderer, reject documents whose advertised uncompressed size exceeds HB_SVG_MAX_DOCUMENT_SIZE, and inflate the rest through zlib while keeping the same size cap enforced during decompression. Plain SVG documents are also bounded by the same 16 MiB limit.

Add a raster API regression that renders the existing plain and compressed noto_handwriting SVG fixtures and verifies identical output.

Testing:
- meson test -C build

Assisted-by: OpenAI Codex

Fixes https://github.com/harfbuzz/harfbuzz/issues/5820


meson.build
```diff
@@ -160,6 +160,7 @@ endif
 
 chafa_dep = dependency('chafa', version: chafa_min_version, required: get_option('chafa'))
 png_dep = dependency('libpng', required: false)
+zlib_dep = dependency('zlib', required: get_option('raster'))
```
khaledhosny:
```suggestion
zlib_dep = dependency('zlib', required: false)
```

khaledhosny:
It should also report zlib status in the summary next to libpng.

behdad:
PTAL. Ready from my side.

khaledhosny:
hb-vector needs to support gzip-compressed SVGs as well.

behdad:
Right.

khaledhosny:
This is probably not related, but the size of the rendered images with [`emojione-svg.otf`](https://github.com/emojione/emojione-assets/releases/download/3.1.2/emojione-svg.otf) is broken:
```
$ ./util/hb-raster emojione-svg.otf 😀😇
```
iTerm2 WL1Bw8

khaledhosny:
`noto_handwriting-untouchedsvgz.ttf` works with `hb-vector` now, but `noto_handwriting-picosvgz.ttf` produces no output, while with `emojione-svg.otf` the glyphs are truncated.

behdad:
Let's narrow down what's pre-existing bugs (I'll work on next) vs what's svgz-related.

khaledhosny:
OK

PR #5831, opened 2026-03-12
[amalgamation] Add one-with-everything source

behdad:
Generate harfbuzz-one-with-everything.cc from Meson using the same source lists as the libraries

@khaledhosny Is this useful? How should it look differently?

Assisted-by: OpenAI Codex


src/meson.build
```diff
@@ -599,6 +599,8 @@ foreach feature : features
   hb_enabled_features.set(key, conf.get('HAVE_@0@'.format(feature), false))
   hb_supported_features.set(key, 1)
 endforeach
+hb_enabled_features.set('HB_HAS_SUBSET', not get_option('subset').disabled())
+hb_supported_features.set('HB_HAS_SUBSET', 1)
```
khaledhosny, line 603: Why is this hard-coded here and not added to the loop above like the rest?

src/meson.build
```diff
@@ -1014,6 +1016,84 @@ if have_icu and not have_icu_builtin
...
+endforeach
+hb_everything_vector_inputs = []
+foreach f : hb_everything_vector_sources
+  hb_everything_vector_inputs += meson.current_source_dir() / f
+endforeach
+
+hb_everything_cairo_sources = [
+  'hb-cairo.cc',
+  'hb-cairo-utils.cc',
+]
```
khaledhosny, line 1051: Why are these hard-coded here?

khaledhosny, line 1051: To be honest, I don’t understand what is this trying to do and why subset is handled differently than raster and vector.

src/harfbuzz-one-with-everything.cc
```diff
@@ -0,0 +1,126 @@
...
+
+#ifdef HB_HAS_ICU
+#ifndef HAVE_ICU
+#define HAVE_ICU 1
+#endif
+#endif
+
+#include "harfbuzz.cc"
+
+#if defined(HB_HAS_SUBSET)
```
khaledhosny, line 68: I don’t understand the idea behind this. With `harfbuzz.cc` we would include all files, and feature that are not needed would be disabled with `hb-config.h` macros. If possible, keeping the same mechanism would be preferable here.

behdad: With `harfbuzz.cc` it's different because `hb-uniscribe.cc` is gated no `HAVE_UNISCRIBE`, etc. But hb-subset, hb-cairo, etc don't have these gates.

Also, this goes a different direction: the client provides a custom `hb-features.h`, and we configure the tree for that.

khaledhosny: Got it. Not sure I like it. Sounds like we are re-inventing build systems at this point.

behdad: Okay. I'll just land the other parts for now, that make it possible to build these files together, but not attempt a mega source file.

khaledhosny: I do like the idea, but I feel it is too complicated, and I’m not sure I like re-purposing `hb-features.h` like this. Lets land the other parts indeed, and keep this idea for a while to see if we can make it work in a different/simpler way.

PR #5835, opened 2026-03-12
[amalgamation] Add HB_HAS_SUBSET and share SVG helpers

behdad:
Add HB_HAS_SUBSET to hb-features.h and wire it through the same HAVE_* feature plumbing used by the other HB_HAS_* macros.

Refactor the raster/vector SVG helper code so the duplicated private helpers live in shared internal headers, avoiding duplicate definitions when sources are combined in a single translation unit.

Testing: meson compile -C build

Assisted-by: OpenAI Codex


src/hb-raster-svg-base.hh
```diff
@@ -218,6 +218,7 @@ hb_raster_svg_tag_is_container (hb_svg_str_t tag)
   return tag.eq ("g") || tag.eq ("a") || tag.eq ("svg") || tag.eq ("symbol");
 }
 
+
```
khaledhosny, line 221: Empty line change.

PR #5836, opened 2026-03-12
Guard VarData itemCount overflow

behdad:
Reject oversized merged VarData row sets before assigning them to the 16-bit itemCount field. This avoids truncating the serialized allocation size and prevents the subsequent row write loop from running out of bounds.

 src/hb-ot-layout-common.hh | 4 +++-
 1 file changed, 3 insertions(+), 1 deletion(-)


behdad:
This is a bandaid over a larger optimizer issue generating rowCount > 65535. I'm investigating the root cause as well.

PR #5855, opened 2026-03-18
Add start/end compose/decompose messages

### LINKED ISSUE #5853: [message] Show decompose/compose
We get `start reorder` and `end reorder` messages, but no corresponding messages for decompose and compose.

khaledhosny:
Similar to start/end reorder messages. It does not give details on what have been composed or decomposed as this seems to be tricky and one can always deduce that from the changes to buffer contents.

* Fixes #5853

### DIFF
--- a/src/hb-ot-shape-normalize.cc
+++ b/src/hb-ot-shape-normalize.cc
@@ -322,6 +322,7 @@ _hb_ot_shape_normalize (const hb_ot_shape_plan_t *plan,
   /* First round, decompose */
 
   bool all_simple = true;
+  if (buffer->message (font, "start decompose"))
   {
     buffer->clear_output ();
     count = buffer->len;
@@ -364,6 +365,7 @@ _hb_ot_shape_normalize (const hb_ot_shape_plan_t *plan,
     }
     while (buffer->idx < count && buffer->successful);
     buffer->sync ();
+    (void) buffer->message (font, "end decompose");
   }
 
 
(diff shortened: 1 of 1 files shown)


behdad:
Lgtm. Lemme know if you want more detailed messages and I'll take a look. We do that for ligation, multiplication, etc.

khaledhosny:
> Lgtm. Lemme know if you want more detailed messages and I'll take a look. We do that for ligation, multiplication, etc.

I actually attempted that, e.g.:
```diff
diff --git a/src/hb-ot-shape-normalize.cc b/src/hb-ot-shape-normalize.cc
index e7007322e..b098eaf22 100644
--- a/src/hb-ot-shape-normalize.cc
+++ b/src/hb-ot-shape-normalize.cc
@@ -454,6 +454,10 @@ _hb_ot_shape_normalize (const hb_ot_shape_plan_t *plan,
            font->get_nominal_glyph (composed, &glyph))
        {
          /* Composes. */
+         hb_codepoint_t a = buffer->out_info[starter].codepoint;
+         hb_codepoint_t b = buffer->cur().codepoint;
+         if (HB_BUFFER_MESSAGE_MORE && buffer->messaging ())
+           (void) buffer->message (font, "composing U+%04X U+%04X to U+%04X", a, b, composed);
          if (unlikely (!buffer->next_glyph ())) break; /* Copy to out-buffer. */
          buffer->merge_out_clusters (starter, buffer->out_len);
          buffer->out_len--; /* Remove the second composable. */
```

But the buffer at the time of the message wouldn’t reflect the composition, which I found a bit confusing e.g.:
```
trace: start compose    buffer: <U+0041=0|U+0301=0>
trace: composing U+0041 U+0301 to U+00C1        buffer: <U+0041=0|U+0301=0>
trace: end compose      buffer: <U+00C1=0>
```

If I move the message to the end of the block I get an assert:
```
Assertion failed: (!have_output || (out_info == info && out_len == idx)), function message_impl, file hb-buffer.cc, line 2358.
```

behdad:
Humm yeah, let's skip then. It's a tricky time to modify the buffer, which message() does sometimes.

PR #5859, opened 2026-03-18
[message] Improve start/end table GSUB/GPOS with HB_TAG_NONE

### LINKED ISSUE #5856: [message] start table GSUB is always shown even if the font has no GSUB table
In a font with no GSUB table I see:
```
trace: start table GSUB script tag '	buffer: [...]
trace: end table GSUB script tag '	buffer: [...]
```

khaledhosny:
Use `hb_tag_to_string()` so we would get an empty string in this case instead of calling `HB_UNTAG()` directly which would give us null bytes that would truncate the string.

Instead of:
```
trace: start table GSUB script tag '	buffer: [a=0]
```

We now get:
```
trace: start table GSUB script tag ''	buffer: [a=0]
```

* Fixes #5856

### DIFF
--- a/src/hb-ot-layout.cc
+++ b/src/hb-ot-layout.cc
@@ -2072,21 +2072,25 @@ inline void hb_ot_map_t::apply (const Proxy &proxy,
 void hb_ot_map_t::substitute (const hb_ot_shape_plan_t *plan, hb_font_t *font, hb_buffer_t *buffer) const
 {
   GSUBProxy proxy (font->face);
+  char tag[5];
+  hb_tag_to_string (chosen_script[0], tag);
   if (buffer->messaging () &&
-      !buffer->message (font, "start table GSUB script tag '%c%c%c%c'", HB_UNTAG (chosen_script[0]))) return;
+      !buffer->message (font, "start table GSUB script tag '%s'", tag)) return;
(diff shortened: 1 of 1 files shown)


behdad:
LGTM. Maybe just short-circuit it around there and return if it's NONE?

behdad:
Same about GPOS?

khaledhosny:
> Maybe just short-circuit it around there and return if it's NONE?

This breaks tests, we still seem to do some shaping work (e.g. insert dotted circle).

khaledhosny:
> Same about GPOS?

`hb_ot_map_t::position()` is only called when the plan `has_gpos` is true, so it gets skipped entirely in this case.

behdad:
> > Same about GPOS?
> 
> `hb_ot_map_t::position()` is only called when the plan `has_gpos` is true, so it gets skipped entirely in this case.

Still, the script might have nul or be nonsense otherwise.

khaledhosny:
> > > Same about GPOS?
> > 
> > 
> > `hb_ot_map_t::position()` is only called when the plan `has_gpos` is true, so it gets skipped entirely in this case.
> 
> Still, the script might have nul or be nonsense otherwise.

Done.

PR #5862, opened 2026-03-21
Fix for issue #5847 - Horizontal centering of glyphs in vertical runs

### LINKED ISSUE #5847: Horizontal centering of glyphs in vertical runs looks different in HarfBuzz 13.1.1
Just built the latest version of HarfBuzz (13.1.1) and noticed that the horizontal centering of glyphs in vertical runs looks different (at least for a few older fonts I tried) compared to HarfBuzz 11.2.1. The glyphs do not appear to be horizontally centered any more but are shifted by some amount in the horizontal direction (I guess to the center of the glyph's advance width, not completely sure). Attached are two small images that show the output before (HarfBuzz-11.2.1.png) and after (HarfBuzz-13.1.1.png). There are also some small differences in vertical advances after certain glyphs, although this is not the first thing that jumps out.

Went through every NEWS entry since 11.2.1 but could not find anything specifically mentioning this. It appears as if some change in HarfBuzz after 11.2.1 has abandoned the implicit horizontal centering of glyphs in vertical runs. Is the client now supposed to manually center the glyphs? 

If this is the case, isn't that a major layout regression for those relying on the legacy behavior where glyphs were balanced relative to their horizontal advance? Was this removal of the centering logic intentional or an undocumented side-effect of some other change? Thanks.

Image

Image

dejanyy:
A fix for issue #5847 that minimizes changes to existing code.

A small tweak to the hb_font_get_glyph_h_origins_default() and hb_font_get_glyph_v_origins_default() functions so their return value always matches the return value of their hb_font_get_glyph_h_origin_default() and hb_font_get_glyph_v_origin_default() counterparts. This ensures consistent behaviour in batch origin-changing callbacks, introduced on July 20, 2025 as part of commit https://github.com/harfbuzz/harfbuzz/commit/faa4472d44f43e39036f11ee38214d19c70a4034 ("Use batch origin-changing callbacks").

Without this fix, hb_font_get_glyph_h_origins_default() always returns true when font->has_glyph_h_origin_func_set() is true, and hb_font_get_glyph_v_origins_default() always returns true when font->has_glyph_v_origin_func_set() is true, which is not what their hb_font_get_glyph_h_origin_default() and hb_font_get_glyph_v_origin_default() counterparts do.

Fixes #5847

### DIFF
--- a/src/hb-font.cc
+++ b/src/hb-font.cc
@@ -406,14 +406,15 @@ hb_font_get_glyph_h_origins_default (hb_font_t *font HB_UNUSED,
 {
   if (font->has_glyph_h_origin_func_set ())
   {
+    hb_bool_t ret = true;
     for (unsigned int i = 0; i < count; i++)
     {
-      font->get_glyph_h_origin (*first_glyph, first_x, first_y, false);
+      ret &= font->get_glyph_h_origin (*first_glyph, first_x, first_y, false);
       first_glyph = &StructAtOffsetUnaligned<hb_codepoint_t> (first_glyph, glyph_stride);
       first_x = &StructAtOffsetUnaligned<hb_position_t> (first_x, x_stride);
       first_y = &StructAtOffsetUnaligned<hb_position_t> (first_y, y_stride);
     }
-    return true;
+    return ret;
   }
 
   hb_bool_t ret = font->parent->get_glyph_h_origins (count,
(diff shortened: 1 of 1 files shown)


behdad:
Thank you!

dejanyy:
FYI -- Not directly related to this, but I might have one more small fix tomorrow for apply_glyph_h_origins_with_fallback() and apply_glyph_v_origins_with_fallback(). I think I found a code path where origins[].x and origins[].y could remain uninitialized. My fix also offers a minor performance optimization by avoiding add_offset() and subtract_offset() calls when all origin values are zero.

behdad:
> FYI -- Not directly related to this, but I might have one more small fix tomorrow for apply_glyph_h_origins_with_fallback() and apply_glyph_v_origins_with_fallback(). I think I found a code path where origins[].x and origins[].y could remain uninitialized. My fix also offers a minor performance optimization by avoiding add_offset() and subtract_offset() calls when all origin values are zero.

Thanks. Looking forward to it.

PR #5863, opened 2026-03-21
Skip adjusting glyph offsets when all origin values are zero

dejanyy:
This issue is not **directly** related to #5847 but was discovered at the same time and makes sense to include in this branch. The fix affects the apply_glyph_h_origins_with_fallback() and apply_glyph_v_origins_with_fallback() functions in hb-font.hh.

There is a potential code path in these functions where origins[].x and origins[].y could remain uninitialized. This fix solves this and also offers a minor performance optimization by completely skipping adjusting glyph offsets when all origins[].x and origins[].y values are zero.

The proposed fix minimizes changes to existing code.

### DIFF
--- a/src/hb-font.hh
+++ b/src/hb-font.hh
@@ -903,15 +903,11 @@ struct hb_font_t
 	}
 	else
 	{
-	  for (unsigned j = 0; j < n; j++)
-	  {
-	    origins[j].x = 0;
-	    origins[j].y = 0;
-	  }
+	  mult = 0; /* Indicates all origins[].x and origins[].y values are 0, therefore we can skip adjusting offsets below */
 	}
       }
 
-      assert (mult == -1 || mult == +1);
+      assert (mult == -1 || mult == +1 || mult == 0);
       if (mult == +1)
         for (unsigned j = 0; j < n; j++)
 	{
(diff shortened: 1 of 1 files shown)


dejanyy:
So... I'm no longer sure if there is a potential code path in apply_glyph_h_origins_with_fallback() and apply_glyph_v_origins_with_fallback() where origins[].x and origins[].y could remain uninitialized. I might be wrong. Perhaps Behdad could check my initial commit 4ca8000 and advise.

However, the second part still stands: this fix offers a minor performance optimization by completely skipping adjusting glyph offsets when all origins[].x and origins[].y values are zero.

So I submitted a simplified commit (042cbb9).

dejanyy:
OK — I reviewed the code carefully and now understand how apply_glyph_h_origins_with_fallback() and apply_glyph_v_origins_with_fallback() work. There is no code path where origins[].x or origins[].y remain uninitialized, so my initial commit 4ca8000 was incorrect. The later commit 042cbb9, however, fixes the issue.

This pull request is only relevant if you want to apply a small micro-optimization: skip adjusting glyph offsets when all origin values are zero.

behdad:
Can you rebase on main and squash just the remaining optimization please? I would take it.

dejanyy:
I was actually just thinking of doing this. Will do.

dejanyy:
OK, I think it's done. It's now all in one simple commit de4b94a. I've also modified the commit message not to mention reading uninitialized origins[] since this part was wrong.

PR #5870, opened 2026-03-23
[fuzzing] Add targeted integer-overflow fuzzer

matejsmycka:
Add `hb-intoverflow-fuzzer` that exercises draw, paint, extents, glyph names, subset instancing, and OT layout paths with variation axes at extremes and font scales up to INT_MAX.

Adds ~22% line coverage over shape-fuzzer alone on the same corpus (53k vs 43k lines hit).

Catches signed integer overflow UB in `COLR.hh:2711` (adwaita.ttf) and `hb-geometry.hh:38,40,99,100` (varc-6868.ttf) at extreme scales.

### DIFF
--- a/test/fuzzing/hb-intoverflow-fuzzer.cc
+++ b/test/fuzzing/hb-intoverflow-fuzzer.cc
@@ -1,0 +1,308 @@
+/*
+ * Targeted integer-overflow fuzzer for HarfBuzz.
+ *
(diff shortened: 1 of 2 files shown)


behdad:
Thanks. I'll review this soon.

Historically, we've ignored int-overflow in glyph coordinates because there's nothing meaningful we can do there. How does this fuzzer address / disable those?

PR #5875, opened 2026-03-26
[message] Improve recursing message

khaledhosny:
Currently if recursed is skipped, no message is shown:
```
start lookup 56 feature 'rlig'
recursing to lookup 43 at 8
replacing glyph at 8 (single substitution)
replaced glyph at 8 (single substitution)
recursed to lookup 43
recursing to lookup 51 at 10
end lookup 56 feature 'rlig'
```

With this change we always show `recursed to` messaged, and show `skipped lookup` message is recursed lookup is skipped:
```
start lookup 56 feature 'rlig'
recursing to lookup 43 at 8
replacing glyph at 8 (single substitution)
replaced glyph at 8 (single substitution)
recursed to lookup 43
recursing to lookup 51 at 10
skipped lookup 51
recursed to lookup 51
end lookup 56 feature 'rlig'
```

### DIFF
--- a/src/hb-ot-layout-gsubgpos.hh
+++ b/src/hb-ot-layout-gsubgpos.hh
@@ -1865,22 +1865,24 @@ static inline void apply_lookup (hb_ot_apply_context_t *c,
       if (buffer->have_output)
         c->buffer->sync_so_far ();
       c->buffer->message (c->font,
-			  "recursing to lookup %u at %u",
+			  "start recursing to lookup %u at %u",
 			  (unsigned) lookupRecord[i].lookupListIndex,
 			  buffer->idx);
     }
 
-    if (!c->recurse (lookupRecord[i].lookupListIndex))
-      continue;
+    bool ret = c->recurse (lookupRecord[i].lookupListIndex);
 
     if (HB_BUFFER_MESSAGE_MORE && c->buffer->messaging ())
     {
       if (buffer->have_output)
         c->buffer->sync_so_far ();
       c->buffer->message (c->font,
-			  "recursed to lookup %u",
+			  "end recursing to lookup %u",
 			  (unsigned) lookupRecord[i].lookupListIndex);
     }
+
+    if (!ret)
+      continue;
 
     unsigned int new_len = buffer->backtrack_len () + buffer->lookahead_len ();
     int delta = new_len - orig_len;


khaledhosny:
Another form @justvanrossum suggested (in https://github.com/fontra/fontra/pull/2500#issuecomment-4115553877) would make it:
```
start lookup 56 feature 'rlig'
start recursing to lookup 43 at 8
replacing glyph at 8 (single substitution)
replaced glyph at 8 (single substitution)
end recursing to lookup 43
start recursing to lookup 51 at 10
skipped lookup 51
end recursing to lookup 51
end lookup 56 feature 'rlig'
```

I have no strong opinion either way.

It was also suggested to show why the lookup was skipped, but I don’t actually know why it was skipped.

justvanrossum:
"start recursing to lookup ..." and "end recursing to lookup ..." would be my preference, since that makes message parsing for nested grouping ~more consistent~ easier. But I understand if that is too big a change for existing users of the message API, and I can totally live with "recursing" / "recursed". Knowing that "recursed" will always be emitted would already be a great improvement.

behdad:
I'm fine with Just's proposed version. I think the only other user of this is @simoncozens, so as long as he's onboard, fine with me.

behdad:
> It was also suggested to show why the lookup was skipped, but I don’t actually know why it was skipped.

Typically what happens is that the lookup simply doesn't apply. There's no skipping as far as I know.

simoncozens:
Either way is fine, I'll adapt Crowbar.

khaledhosny:
Now it looks like this:
```
start lookup 56 feature 'rlig'
start recursing to lookup 43 at 8
replacing glyph at 8 (single substitution)
replaced glyph at 8 (single substitution)
recursed to lookup 43
end recursing to lookup 43
start recursing to lookup 51 at 10
skipped lookup 51
end recursing to lookup 51
end lookup 56 feature 'rlig'
```

src/hb-ot-layout-gsubgpos.hh
```diff
@@ -1865,19 +1865,19 @@ static inline void apply_lookup (hb_ot_apply_context_t *c,
...
     bool skipped = !c->recurse (lookupRecord[i].lookupListIndex);
 
-    if (HB_BUFFER_MESSAGE_MORE && c->buffer->messaging () && skipped)
+    if (HB_BUFFER_MESSAGE_MORE && c->buffer->messaging ())
     {
       if (buffer->have_output)
         c->buffer->sync_so_far ();
       c->buffer->message (c->font,
-			  "skipped lookup %u",
+			  "%s lookup %u", skipped ? "skipped" : "recursed to",
```
behdad: Maybe `skipped` / `recursed to` is not the right wording. Looks like "skipped" means "didn't apply" and "recursed to" means successfully applied.

khaledhosny: I used skipped similar to:
https://github.com/harfbuzz/harfbuzz/blob/e6d8ad97dce8c48f9f0d9976576018bfc9ba6001/src/hb-ot-layout.cc#L2054-L2055

But if you think a different wording is better, I can use it (do you suggest using `applied`/`didn’t apply`?) or I can drop that line entirely.

khaledhosny: I checked the font, and this the corresponding fea code:
```fea
# lookup 43
lookup tooth_rounded {
	sub behDotless-ar.medi	by behDotless-ar.medi.round;
	sub lam-ar.medi			by lam-ar.medi.round;
	sub lam-ar.medi.short	by lam-ar.medi.round.short;
	sub lam-ar.medi.hah1	by lam-ar.medi.hah1.round;
	sub lam-ar.medi.hah2	by lam-ar.medi.hah2.round;
} tooth_rounded;

# lookup 51
lookup rounded {
	sub meem-ar.init		by meem-ar.init.round;
	sub meem-ar.medi		by meem-ar.medi.round;
	sub meem-ar.fina		by meem-ar.fina.round;
	sub waw-ar.fina			by waw-ar.fina.round;
	sub heh-ar.init			by heh-ar.init.round;
} rounded;

# lookup 56
lookup rlig {
	lookupflag IgnoreMarks;
	sub [behDotless-ar.medi lam-ar.medi lam-ar.medi.hah1 lam-ar.medi.hah2]' lookup tooth_rounded
		[meem-ar.medi meem-ar.fina noonghunna-ar.fina]' lookup rounded
	;
} rlig;
```

The input (at this point is) is `[behDotless-ar.medi|noonghunna-ar.fina]`

So lookup 51 gets applied but there is no replacement for `noonghunna-ar.fina`.

I can see now how skipped is misleading. I guess it is better to not say anything then.

khaledhosny: So, how about this?
```
start lookup 56 feature 'rlig'
start recursing to lookup 43 at 8
replacing glyph at 8 (single substitution)
replaced glyph at 8 (single substitution)
end recursing to lookup 43
start recursing to lookup 51 at 10
end recursing to lookup 51
end lookup 56 feature 'rlig'
```

behdad: That `skipped` message is categorically different. It is just an informational optimization message saying that the lookup loop was not performed because we knew nothing will match.

I suggest here `applied` / `not applied`.

behdad: Nothing works. After all, if there was a substitution, you get a message for that.

khaledhosny: > Nothing works. After all, if there was a substitution, you get a message for that.

Yes, it feels cleaner this way.

Issue #5877, opened 2026-03-26
[gpu] Design `hb-gpu.h` API to absorb GLyphy library


behdad:
https://github.com/behdad/glyphy/blob/main/src/glyphy.h

UPDATE: ended up renaming `hb_gpu_glyph_t` to `hb_gpu_draw_t` in the API below.

Proposed API so far:
```c
  #ifndef HB_GPU_H                                                                                                                                                                                                 
  #define HB_GPU_H                                                                                                                                                                                                 
                                                                                                                                                                                                                   
  #include "hb.h"                                                                                                                                                                                                  
                                                                                                                                                                                                                   
  HB_BEGIN_DECLS  


  /**
   * hb_gpu_shader_lang_t:
   * @HB_GPU_SHADER_GLSL_330: Desktop OpenGL, GLSL 3.30                                                                                                                                                            
   * @HB_GPU_SHADER_GLSL_ES_300: OpenGL ES 3.0, GLSL ES 3.00
   *                                                                                                                                                                                                               
   * Shader language variant.
   *                                                                                                                                                                                                               
   * Since: REPLACEME
   */                                                                                                                                                                                                              
  typedef enum {
    HB_GPU_SHADER_GLSL_330,                                                                                                                                                                                        
    HB_GPU_SHADER_GLSL_ES_300,
  } hb_gpu_shader_lang_t;
                                                                                                                                                                                                                   
  HB_EXTERN const char * const *
  hb_gpu_shader_fragment_sources (hb_gpu_shader_lang_t  lang,                                                                                                                                                      
                                unsigned int         *count);                                                                                                                                                      
  
  HB_EXTERN const char * const *                                                                                                                                                                                   
  hb_gpu_shader_vertex_sources (hb_gpu_shader_lang_t  lang,
                              unsigned int         *count);                                                                                                                                                        
  
                                                                                                                                                                                                                   
  /**             
   * hb_gpu_glyph_t:
   *
   * An opaque GPU glyph encoder.  Accumulates glyph outlines via
   * draw callbacks, then encodes them into a compact blob for GPU                                                                                                                                                 
   * rendering.                                                                                                                                                                                                    
   *                                                                                                                                                                                                               
   * Since: REPLACEME                                                                                                                                                                                              
   */             
  typedef struct hb_gpu_glyph_t hb_gpu_glyph_t;
                                                                                                                                                                                                                   
  HB_EXTERN hb_gpu_glyph_t *
[...]

PR #5878, opened 2026-03-27
[gpu] Add hb-gpu library ported from GLyphy

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/5877

### DIFF
--- a/util/gpu/meson.build
+++ b/util/gpu/meson.build
@@ -22,8 +22,6 @@ gpu_demo_sources = files(
   'trackball.c',
 )
 
-gpu_demo_generated = []
-
 glew_dep = dependency('glew', required: false)
 glfw_dep = dependency('glfw3', required: false)
 gl_dep = dependency('gl', required: false)
--- a/util/meson.build
+++ b/util/meson.build
@@ -219,7 +219,7 @@ if not get_option('gpu').disabled()
 
   if have_gpu_demo_deps and conf.get('HAVE_GLIB', 0) == 1
     hb_gpu_util = executable('hb-gpu-demo',
-      ['hb-gpu-demo.cc'] + gpu_demo_sources + gpu_demo_generated,
+      ['hb-gpu-demo.cc'] + gpu_demo_sources,
       cpp_args: cpp_args,
       include_directories: [incconfig, incsrc],
       dependencies: [util_deps, gl_dep, glew_dep, glfw_dep],


src/hb-gpu-shaders.cc
```diff
@@ -0,0 +1,95 @@
...
+ * shader language.  The returned array and strings are static and
+ * must not be freed.
+ *
+ * The caller should pass all @count strings to `glShaderSource()`
+ * (or the equivalent) when compiling the fragment shader.
+ *
+ * Return value: (transfer none) (array length=count):
+ * An array of source strings, or `NULL` if @lang is unsupported.
+ *
+ * Since: REPLACEME
```
khaledhosny: This should be XSince (here and elsewhere).

util/gpu/meson.build
```diff
@@ -0,0 +1,53 @@
...
+
+gpu_demo_generated += custom_target('default font',
+  input: 'default-font.ttf',
+  output: 'default-font.h',
+  capture: true,
+  command: [ hexify, 'static const char default_font[]', '@INPUT@' ])
+
+glew_dep = dependency('glew', required: true)
+glfw_dep = dependency('glfw3', required: true)
+gl_dep = dependency('gl', required: true)
```
khaledhosny: We should add warp files for these.

util/meson.build
```diff
@@ -213,3 +213,16 @@ else
   hb_info = disabler()
   hb_subset = disabler()
 endif
+
+if not get_option('gpu').disabled()
+  subdir('gpu')
+
+  hb_gpu_util = executable('hb-gpu',
```
khaledhosny: Since this is a demo, I’d call it `hb-gpu-demo`, and disable building it by default (even when `gpu` is enabled) as it has 3 dependencies and we don’t even install it.

khaledhosny:
The linux job needs to install the demo dependencies, and disable the demo for other jobs.

src/hb-gpu-shaders.cc
behdad: Yeah I should add to AGENTS.md.

util/meson.build
behdad: Not auto-detect? I'm fine with that. What meson API do you suggest?

behdad:
> The linux job needs to install the demo dependencies, and disable the demo for other jobs.

Just did that for linux job.

src/hb-gpu-shaders.cc
khaledhosny: It is there already. Maybe it needs stronger words :)

util/meson.build
khaledhosny: Autodetect is fine as well. But if you want to go the other route, then add a `gpu_demo`.

behdad: Let's keep auto for now.

behdad:
This is ready as far as I'm concerned.

src/hb-gpu.cc
```diff
@@ -34,24 +34,206 @@
...
+ * allow buffer reuse; call hb_gpu_draw_reset() before each new
+ * glyph.
+ *
+ * # Atlas setup
+ *
+ * All encoded blobs are uploaded into a single GL_TEXTURE_BUFFER
+ * backed by a GL_RGBA16I format buffer.  Each glyph occupies a
+ * contiguous range of texels at a known offset:
+ *
+ * |[<!-- language="C" -->
```
khaledhosny: `language="C"` looks odd in the HTML generated by GTK-Doc, that is why I use `language="plain"`.

src/hb-gpu.cc
```diff
@@ -34,24 +34,206 @@
...
+ * glShaderSource (vert_shader, n, vert_sources, NULL);
+ *
+ * // Same for fragment shader
+ * ]|
+ *
+ * # Vertex shader
+ *
+ * The vertex library provides one function:
+ *
+ * |[<!-- language="GLSL" -->
```
khaledhosny: I’d use `language="plain"` here too (for any code sample, really).

src/hb-gpu.cc
```diff
@@ -34,24 +34,206 @@
...
+ *
+ * This expands each glyph quad by approximately half a pixel on
+ * screen to ensure proper antialiasing coverage at the edges.
+ * It must be called in the vertex shader before computing
+ * gl_Position.
+ *
+ * Per-vertex attributes for each glyph quad (2 triangles, 6
+ * vertices, 4 unique corners):
+ *
+ * - **position** (vec2): object-space vertex position, computed from
```
khaledhosny: GTK-Doc seems to not support the `**bold**` formatting.

src/hb-gpu.cc
```diff
@@ -34,24 +34,206 @@
...
+ *   the glyph's bounding box.  For corner (cx, cy) where cx,cy
+ *   are 0 or 1:
+ *   `pos.x = pen_x + scale * lerp(extent_min_x, extent_max_x, cx)`
+ *   `pos.y = pen_y - scale * lerp(extent_min_y, extent_max_y, cy)`
```
khaledhosny: This gets rendered as one line. It should use the `|[ ]|` blocks instead.

util/gpu/meson.build
```diff
@@ -12,41 +12,18 @@ gpu_demo_sources = files(
...
 
-gpu_demo_shaders = [
-  [ 'demo-fragment.glsl', 'demo-fragment-glsl.h', 'demo_fragment_glsl' ],
-  [ 'demo-vertex.glsl', 'demo-vertex-glsl.h', 'demo_vertex_glsl' ],
-  [ 'default-text.txt', 'default-text.h', 'default_text' ],
-]
-
-gpu_stringize = find_program('stringize')
-
 gpu_demo_generated = []
```
khaledhosny: This is unused now.

PR #5883, opened 2026-03-30
Build hb-gpu demo on mingw CI

behdad:


.github/workflows/crossbuild-mingw.yml
```diff
@@ -62,7 +64,30 @@ jobs:
         Libs.private: -lgdi32
         Cflags: -I\${includedir}
         EOF
-        sudo cp /tmp/glfw3.pc /usr/${TRIPLET}/lib/pkgconfig/
+        sudo cp /tmp/glfw3.pc ${SYSROOT}/lib/pkgconfig/
+
+        # GLEW (build from source with MinGW)
```
khaledhosny: There seem to be a wrap file for it, so you can add wrap file and let meson do the building.

.ci/build-win.sh
```diff
@@ -10,10 +10,44 @@ fi
...
+  TRIPLET=i686-w64-mingw32
+fi
+
 BUILD=build-win${ARCH}
 INSTALL=install-win${ARCH}
 DIST=harfbuzz-win${ARCH}
 
+# Install GLFW for MinGW if not already present
+if [ ! -f /usr/${TRIPLET}/include/GLFW/glfw3.h ]; then
+  GLFW_VER=3.4
```
khaledhosny: same here

khaledhosny:
Meson is still not building glfw/glew. I think because they are not required deps, I think we need a gpu_demo option, or handle it with the utilities option and make them required deps then.

behdad:
> Meson is still not building glfw/glew. I think because they are not required deps, I think we need a gpu_demo option, or handle it with the utilities option and make them required deps then.

I'm doing just that.

.ci/build-win.sh
```diff
@@ -33,6 +33,7 @@ meson setup \
 	-Dchafa=disabled \
 	-Dgdi=enabled \
 	-Ddirectwrite=enabled \
+	-Dgpu=enabled \
```
khaledhosny: You need `-Dgpu_demo=enabled` here as well.

khaledhosny: Also, `gpu` is enabled by default, so you don’t need to enable it here.

util/gpu/meson.build
```diff
@@ -33,15 +33,19 @@ if host_machine.system() == 'windows'
...
+gpu_demo_opt = get_option('gpu_demo')
+have_gpu_demo_deps = false
+if not gpu_demo_opt.disabled()
+  glew_dep = dependency('glew', required: false)
+  glfw_dep = dependency('glfw3', required: false)
+  if host_machine.system() == 'darwin'
+    gl_dep = dependency('appleframeworks', modules: ['OpenGL'], required: false)
+    metal_dep = dependency('appleframeworks', modules: ['Metal', 'QuartzCore', 'Foundation', 'AppKit'], required: false)
+  else
+    gl_dep = dependency('gl', required: false)
```
khaledhosny:
```suggestion
  glew_dep = dependency('glew', required: gpu_demo_opt)
  glfw_dep = dependency('glfw3', required: gpu_demo_opt)
  if host_machine.system() == 'darwin'
    gl_dep = dependency('appleframeworks', modules: ['OpenGL'], required: gpu_demo_opt)
    metal_dep = dependency('appleframeworks', modules: ['Metal', 'QuartzCore', 'Foundation', 'AppKit'], required: gpu_demo_opt)
  else
    gl_dep = dependency('gl', required: gpu_demo_opt)
```

PR #5904, opened 2026-04-10
Fix: Add safety limits to hb-repacker to prevent algorithmic complexity attacks

XenomorphSk:
This patch introduces hard limits on iterations and graph sizes during the repacking process to mitigate potential OOM and infinite loops on malformed font subsets.


behdad:
Can you add these constants to hb-limits.hh?

behdad:
Thanks.

PR #5906, opened 2026-04-12
[vector] Add PDF fragment render APIs for embedding

behdad:
Adds two new public APIs for generating the PDF-format output of hb-vector without the PDF document wrapping, so consumers (e.g. LibreOffice) can embed glyph rendering directly into their own PDF documents (for example as Type 3 font CharProcs).

Paint side (hb_vector_paint_render_pdf_fragment):
  Returns the raw PDF content stream (drawing operators only,
  referring to resources by symbolic names /GS0 /SH0 /Im0).
  Stream objects that cannot be inlined as dictionary values
  (XObject images + SMasks, Type 6 Coons-patch shadings) are
  handed to a consumer callback in dependency order; the callback
  returns the PDF object id the consumer assigned, which HarfBuzz
  then substitutes into any later stream bodies and into the
  final resource dictionary.  Non-stream objects (Functions,
  simple ExtGState, Type 2/3 shadings) are inlined into the
  resource dict, so most glyphs need zero callback invocations.

Draw side (hb_vector_draw_render_pdf_fragment):
  Trivial — outline glyphs have no resources and no streams;
  returns just the path + fill-operator bytes.

Both reset the underlying context on success, matching the semantics of hb_vector_{draw,paint}_render.

Added test/api/test-vector.cc with five cases covering:
  - draw fragment (outline glyph)
  - paint fragment with no streams (non-sweep COLRv1)
  - NULL resources out-param accepted
  - paint fragment with a stream (sweep gradient → Type 6 shading)
  - NULL callback with streams aborts cleanly

All 241 tests pass.


behdad:
This lgtm if it works for you.

khaledhosny:
I think my change broke something, but also turns out the API on main is simpler to use and results in even smaller PDFs. Lets close this for now.

PR #5908, opened 2026-04-13
Use some RAII error handling

behdad:


src/hb-zlib.cc
```diff
@@ -38,27 +39,25 @@ hb_blob_decompress_gzip (hb_blob_t *blob,
                              (size_t) max_output_len);
   char *output = (char *) hb_malloc (allocated);
   if (!output)
-  {
-    inflateEnd (&stream);
     return nullptr;
-  }
+  auto output_guard = hb_make_scope_guard ([&]() { hb_free (output); });
```
khaledhosny, line 43: Why do some places use `hb_make_scope_guard` and others use the macro `HB_SCOPE_GUARD`?

behdad: The ones that need to call .release on the guard don't use the macro.

PR #5915, opened 2026-04-13
Return size_t rather than unsigned int from get_size()

jfkthame:


behdad:
Thanks. There seems to be more places? linux bot failing with such suggestions.

jfkthame:
Ugh, I see lots of "comparison of integers of different signs".... curious why, as `unsigned int` and `size_t` are both unsigned, so why wasn't that an issue previously? I'll see if I can clean up... not sure if I currently have a working build locally...

behdad:
Now CIFuzz unhappy.

behdad:
Before, unsigned int was being promoted to int64_t for comparison maybe.

jfkthame:
> Before, unsigned int was being promoted to int64_t for comparison maybe.

Yeah, makes sense. I've tried replacing the use of int64_t for `vertex_len` vars in the /graph/ files with size_t -- is that sound, or is there some reason we need those to be signed?

behdad:
> > Before, unsigned int was being promoted to int64_t for comparison maybe.
> 
> Yeah, makes sense. I've tried replacing the use of int64_t for `vertex_len` vars in the /graph/ files with size_t -- is that sound, or is there some reason we need those to be signed?

I think so. @garretrieger

PR #5920, opened 2026-04-13
Add `hb-gpu-paint` COLRv1 renderer

### LINKED ISSUE #5919: [gpu] Add support for COLR table (v0 & v1)
### Context

Currently `hb-gpu` only directly supports rendering ordinary single-color glyphs. Since the library was first added I've experimented and successfully got a working implementation of rendering color fonts in both COLRv0 and COLRv1 formats. Mr. @behdad has then invited me to contribute support for COLR table to `hb-gpu` itself.

- My [blog post](https://leduyquang753.com/blog/2026/4/4/rendering-arbitrary-scale-emojis-using-the-slug-algorithm) describes an outline of my approach.
- I made [a modified version](https://leduyquang753.com/demos/color-fonts-using-slug-algorithm/) of `hb-gpu`'s web demo to show the color font rendering in action. The source code is [here](https://github.com/leduyquang753/harfbuzz/tree/color-font-test) for reference.

The formats in a nutshell:

- COLRv0: Multiple solid-color glyphs stacked on top of each other. Layers are retrievable through `hb_ot_color_glyph_get_layers`.
- COLRv1: A render tree of layers, each layer has clip glyphs, fills (solid color or gradient), affine transforms, and blending equations. Drawing commands are processed by HarfBuzz through `hb_font_paint_glyph`.

[OpenType specification of the COLR table on Microsoft's website](https://learn.microsoft.com/en-us/typography/opentype/spec/colr)

### Considerations and challenges

- Interface:
  - Probably new functions will be added to return multiple layers for each color glyph. Each layer's data may consist of different encoded data blobs for the clip glyph, gradient, transforms, layering,...
  - Feature flags or different fragment shader sources for enabling or disabling color glyph rendering logic.
- Interaction with other features, for example whether to apply fake bold, italic to the color glyphs.
- Performance: Some optimization of the fragment shader will be needed compared to my hacky implementation which is full of branches and slows down noticeably on weaker machines.

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/5919


behdad:
@copilot resolve the merge conflicts in this pull request

behdad:
I pushed a high-level design of the painting.

behdad:
I've pushed the infrastructure and a COLRv0 already.

behdad:
I ended up implementing COLRv1 similar to @leduyquang753's approach.

See the demo:

https://behdad.org/hb-gpu-demo/?font=https://raw.githubusercontent.com/googlefonts/noto-emoji/main/fonts/Noto-COLRv1.ttf&text=%F0%9F%98%80%F0%9F%98%81%F0%9F%98%82%F0%9F%A4%A3%F0%9F%98%83%F0%9F%98%84%F0%9F%98%85%F0%9F%98%86%F0%9F%98%89%F0%9F%98%8A%F0%9F%98%8B%F0%9F%98%8E%F0%9F%98%8D%F0%9F%98%98%F0%9F%A5%B0%F0%9F%98%97%F0%9F%98%99%F0%9F%98%9A%E2%98%BA%EF%B8%8F%F0%9F%99%82%F0%9F%A4%97%F0%9F%A4%A9

It should be reviewed and perhaps simplified. But it's pretty solid already.

behdad:
This is fairly complete now.

src/hb-gpu-paint.cc
```diff
@@ -0,0 +1,1361 @@
...
+  { c->unsupported = true; hb_free (heap_stops); return; }
+  float mn, mx;
+  normalize_stops (stops, got, &mn, &mx);
+  float dx = x1 - x0, dy = y1 - y0, dr = r1 - r0;
+  float cx0 = x0 + mn * dx, cy0 = y0 + mn * dy, cr0 = r0 + mn * dr;
+  float cx1 = x0 + mx * dx, cy1 = y0 + mx * dy, cr1 = r0 + mx * dr;
+  /* Bake current transform into centers; approximate the radius
+   * scale with sqrt(|det|) (the uniform-scale assumption -- a
+   * non-uniform scale turns the circle into an ellipse which the
+   * shader does not model). */
```
leduyquang753: Was there a major issue that you didn't go with passing the inverse transform into the fragment shader? I do see radial gradients with non-uniform scale be used in _Segoe UI emoji_ and this makes the result deviate a bit from how it's supposed to look like.

behdad: No reason, just didn't get to it and then forgot. I'll work on it.

Can you review this PR for me please, and suggest improvements based on your experience?

behdad: > Was there a major issue that you didn't go with passing the inverse transform into the fragment shader? I do see radial gradients with non-uniform scale be used in _Segoe UI emoji_ and this makes the result deviate a bit from how it's supposed to look like.

Which emoji in Segoe UI Emoji can I use for testing? I have this implemented now.

behdad:
Okay, please take another look.

src/hb-gpu.cc
```diff
@@ -320,15 +336,18 @@
...
- *   `dot (foreground.rgb, vec3 (1.0 / 3.0))`.
+ * - brightness: brightness of the color that will end up on
+ *   screen for this fragment, in [0, 1], computed as
+ *   `dot (straight_color.rgb, vec3 (1.0 / 3.0))`.
  *
  * - ppem: pixels per em at this fragment, computed as
  *   `1.0 / max (fwidth (v_texcoord).x, fwidth (v_texcoord).y)`.
  *
- * Example:
+ * Example -- draw path:
```
khaledhosny: The `--` does get converted to en-dash in the HTML output, in case you want to use a real en (or even em) dash.

khaledhosny (approved):

src/hb-gpu-paint.cc
leduyquang753: > Which emoji in Segoe UI Emoji can I use for testing? I have this implemented now.

`🥳` uses radial gradients on the eyes, the shadows of the eyes, the mouth and the party horn.

behdad: Thanks. I haven't been able to get my hands on a Segoe UI Emoji COLRv1 font. Can you please test it at https://behdad.org/hb-gpu-demo/

Thanks

behdad: I managed to test it. Seems to work. Thank!

behdad:
@leduyquang753 I implemented up to 3 nested clips as well.

leduyquang753:
I noticed that each color glyph here is wholly blended in the fragment shader rather than being rendered as multiple quads, which I suppose is to make it easier to integrate by the end user. I guess the bigger bottleneck right now to be addressed is the large number of branch divergences in the fragment shader in general. Running both my original demo and the version here on an Intel Core i5-12500 for instance slows down to about 10 FPS when the glyphs take up the majority of a 1080p screen.

behdad:
> I noticed that each color glyph here is wholly blended in the fragment shader rather than being rendered as multiple quads, which I suppose is to make it easier to integrate by the end user. I guess the bigger bottleneck right now to be addressed is the large number of branch divergences in the fragment shader in general. Running both my original demo and the version here on an Intel Core i5-12500 for instance slows down to about 10 FPS when the glyphs take up the majority of a 1080p screen.

Yes, I made that decision to use integration much easier. 10fps is really bad though. What does your implementation get?

leduyquang753:
> Yes, I made that decision to use integration much easier. 10fps is really bad though. What does your implementation get?

The same about 10 FPS, as I stated for both implementations. Ordinary text still render fast when they cover the whole screen, also for both implementations.

behdad:
> > Yes, I made that decision to use integration much easier. 10fps is really bad though. What does your implementation get?
> 
> The same about 10 FPS, as I stated for both implementations. Ordinary text still render fast when they cover the whole screen, also for both implementations.

Thanks. Another issue I'm facing is that the demo takes 80s to start under wine. This got really this bad as the shader grew. I hope real Windows is not like that. Are you on windows? When we make a release, it would create Windows binaries someone can test.

leduyquang753:
> Thanks. Another issue I'm facing is that the demo takes 80s to start under wine. This got really this bad as the shader grew. I hope real Windows is not like that. Are you on windows? When we make a release, it would create Windows binaries someone can test.

Yes, I use Windows. It started nearly instantly when I tried the in-progress implementation 12 hours ago.

behdad:
> > Thanks. Another issue I'm facing is that the demo takes 80s to start under wine. This got really this bad as the shader grew. I hope real Windows is not like that. Are you on windows? When we make a release, it would create Windows binaries someone can test.
> 
> Yes, I use Windows. It started nearly instantly when I tried the in-progress implementation 12 hours ago.

Thanks. Would be great if you can test the current state as well.

leduyquang753:
I just tried with the latest version and it still started quickly.

Also just another interesting data point that when there are just 2 emojis covering a 1440p screen, at 165 Hz refresh rate, the demo consumes 40% of the NVIDIA RTX 5070.

image

PR #5921, opened 2026-04-13
[gpu] Extract shared shader helpers into hb_gpu_shader_source()

behdad:
The draw-renderer fragment shaders defined utilities (HB_GPU_UNITS_PER_EM, the hb_gpu_atlas uniform, hb_gpu_fetch) that any future GPU renderer will also need.  The draw-renderer vertex shader likewise contained only the generic hb_gpu_dilate helper. If a paint renderer were added today and a caller linked both fragment sources together, hb_gpu_fetch would be defined twice.

Move these shared helpers into a new family of sources and expose them via a new public entry point:

  hb_gpu_shader_source (stage, lang);

The existing hb_gpu_draw_shader_source() now returns only the draw-specific fragment code; it returns an empty string for the vertex stage because all current vertex logic is shared.  Callers concatenate the shared source first, then the draw-specific source, then their own main():

  [#version, hb_gpu_shader_source (...), hb_gpu_draw_shader_source
  (...), main]

Additions:
 - src/hb-gpu-fragment.{glsl,msl,wgsl,hlsl} — shared fragment
 - src/hb-gpu-vertex.{glsl,msl,wgsl,hlsl}   — shared vertex,
   renamed from hb-gpu-draw-vertex.*
 - src/hb-gpu-shaders.cc — hb_gpu_shader_source() dispatcher

Removals:
 - src/hb-gpu-draw-vertex.{glsl,msl,wgsl,hlsl} — moved to common

Shared sources are Behdad-only copyright; the Eric Lengyel / Slug attribution stays on the draw-specific fragment sources where the Slug algorithm actually lives.

Updates: test/api/test-gpu.cc, the Metal/D3D11/WebGPU/GL demo shaders, the hb-gpu SECTION documentation, meson.build, CMakeLists.txt, gen-gpu-shader-artifacts.py decl_map, harfbuzz-world.cc, and docs/harfbuzz-sections.txt.


behdad:
In preparation for https://github.com/harfbuzz/harfbuzz/pull/5920

behdad:
This is just moving existing code around.

PR #5925, opened 2026-04-15
[draw] Add line/rect/circle drawing helpers

behdad:
And add `--show-extents` to `hb-gpu` util.


khaledhosny:
Should `hb-raster`/`hb-vector` use these helpers now, too?

khaledhosny (approved):

behdad:
> Should `hb-raster`/`hb-vector` use these helpers now, too?

The new API is on hb-draw now, so all draw consumers automatically benefit.

khaledhosny:
> > Should `hb-raster`/`hb-vector` use these helpers now, too?
> 
> The new API is on hb-draw now, so all draw consumers automatically benefit.

I meant the utils, they seem to have their own code to draw extents.

behdad:
> > > Should `hb-raster`/`hb-vector` use these helpers now, too?
> > 
> > 
> > The new API is on hb-draw now, so all draw consumers automatically benefit.
> 
> I meant the utils, they seem to have their own code to draw extents.

Right. Needs some new API to work through paint API as well. Currently it's draw-only (in hb-gpu as well; extents are not draw in the paint path).

PR #5927, opened 2026-04-15
Remove flat option from hb-vector SVG output

behdad:
Wasn't working correctly. Remove now that we have svg_set_id_prefix().


khaledhosny:
LGTM.

PR #5931, opened 2026-04-16
[vector/raster/gpu] Add *_glyph_or_fail() variants across all three

### LINKED ISSUE #5926: paint_or_fail() in new libraries?
On hb-font, we have:
```
void hb_font_draw_glyph();
hb_bool_t hb_font_draw_glyph_or_fail();
void hb_font_paint_glyph();
hb_bool_t hb_font_paint_glyph_or_fail();
```
The _or_fail() variants are clear. The draw() one silently does nothing if there's no outlines for the glyph. The paint() tries paint_or_fail() first and cascades to draw() if the paint failed.

Sometimes a cascade like above is useful, sometimes not. This becomes a question in hb-raster, hb-vector, and hb-gpu, each of which have their own draw_t and paint_t objects that expose similar draw/paint API.

So, for example, for GPU, if user chose to use paint(), it's desirable that monochrome glyphs will still be drawn by that same call. And that's how `hb_gpu_paint_glyph()` currently works: just calls `hb_font_paint_glyph()` underneath.

In hb-raster and hb-vector however, the _paint_glyph() call currently internally calls `hb_font_paint_glyph_or_fail()` and returns the status. For hb-raster sometimes this is not a problem and user has to call draw_glyph; They have to deal with the different output format anyway, but at other times it's desirable that paint() still draws the monochrome glyph.

The hb-vector situation is more interesting: again, it was designed such that user tries paint() first, if failed, try draw(). But that's inconvenient at times, and if user doesn't care much they should be able to just use paint(). However, the SVG/PDF output might be different and slightly larger if paint fallback was used instead of draw.

All three should be aligned. The easiest would be to add a `paint_glyph_or_fail()`. It's three more entry points, but probably the right fix. At that point, should we also add `draw_glyph_or_fail()` for completeness? Not sure.

behdad:
Consolidate the {vector,raster,gpu}_{draw,paint}_glyph APIs to match the hb_font_{draw,paint}_glyph / _or_fail pattern:

  * Each existing *_glyph() public function now returns void. For the draw variants this is purely a return-value drop. For the paint variants the void entry additionally gains a synthesize fallback for non-color glyphs (calls hb_font_paint_glyph instead of _or_fail) so a mono glyph still produces output via a foreground-colored outline.

  * New *_glyph_or_fail() variants return hb_bool_t, reporting exactly what hb_font_{draw,paint}_glyph_or_fail() would. Callers that care about distinguishing color paint from synthesized outline now have a clean boolean signal.

Refactors the internal impls behind a shared static helper parameterised by a `fallible` flag; the two public entries delegate, passing true (for _or_fail) or false (for the void alias).

For GPU specifically, paint_glyph_or_fail no longer folds the encoder's unsupported flag into its return -- that state was always better reported from hb_gpu_paint_encode() (which returns NULL), and conflating the two prevented callers from telling "font has no paint data" from "encoder hit a limit".

While in there, remove a dead branch at the end of hb_vector_paint_glyph's SVG case (both inner blocks had already returned; the tail was unreachable).

Updates docs/harfbuzz-sections.txt with the six new symbols and the NEWS note with the Changed API callouts.  Updates in-tree callers (hb-vector-svg-all, hb-raster-all, test-gpu, util/vector-output.hh) to the _or_fail variants where they previously read the return value.

Fixes: https://github.com/harfbuzz/harfbuzz/issues/5926

### DIFF
--- a/src/hb-gpu-draw.cc
+++ b/src/hb-gpu-draw.cc
@@ -966,24 +966,25 @@ hb_gpu_draw_get_scale (const hb_gpu_draw_t *draw,
 }
 
 /**
- * hb_gpu_draw_glyph:
+ * hb_gpu_draw_glyph_or_fail:
  * @draw: a GPU shape encoder
  * @font: font to draw from
  * @glyph: glyph ID to draw
  *
  * Convenience wrapper that draws a single glyph outline into the
- * encoder using hb_font_draw_glyph().  Calls hb_gpu_draw_set_scale()
- * with the font's scale before walking the outline.
+ * encoder using hb_font_draw_glyph_or_fail().  Calls
+ * hb_gpu_draw_set_scale() with the font's scale before walking the
+ * outline.
  *
  * Return value: `true` if the glyph was drawn, `false` if the font
  * has no outlines for @glyph.
  *
- * Since: 14.0.0
+ * XSince: REPLACEME
  **/
 hb_bool_t
-hb_gpu_draw_glyph (hb_gpu_draw_t  *draw,
-		   hb_font_t      *font,
-		   hb_codepoint_t  glyph)
+hb_gpu_draw_glyph_or_fail (hb_gpu_draw_t  *draw,
+			   hb_font_t      *font,
+			   hb_codepoint_t  glyph)
 {
   int x_scale, y_scale;
   hb_font_get_scale (font, &x_scale, &y_scale);
@@ -992,6 +993,25 @@ hb_gpu_draw_glyph (hb_gpu_draw_t  *draw,
   return hb_font_draw_glyph_or_fail (font, glyph,
 				     hb_gpu_draw_get_funcs (),
 				     draw);
+}
+
+/**
+ * hb_gpu_draw_glyph:
+ * @draw: a GPU shape encoder
+ * @font: font to draw from
+ * @glyph: glyph ID to draw
+ *
+ * Draws a single glyph outline into the encoder.  Equivalent to
+ * hb_gpu_draw_glyph_or_fail() with the return value ignored.
+ *
+ * Since: 14.0.0
+ **/
+void
+hb_gpu_draw_glyph (hb_gpu_draw_t  *draw,
+		   hb_font_t      *font,
+		   hb_codepoint_t  glyph)
+{
+  hb_gpu_draw_glyph_or_fail (draw, font, glyph);
 }
 
 
--- a/src/hb-gpu-paint.cc
+++ b/src/hb-gpu-paint.cc
@@ -348,7 +348,7 @@ emit_clip_sub_blob (hb_gpu_paint_t *c,
   {
     /* Fast path: feed the glyph outline straight into the draw
      * encoder with no adapter. */
-    ok = hb_gpu_draw_glyph (c->scratch_draw, clip.font, clip.glyph);
+    ok = hb_gpu_draw_glyph_or_fail (c->scratch_draw, clip.font, clip.glyph);
   }
   else
   {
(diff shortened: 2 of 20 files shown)


src/hb-gpu-draw.cc
```diff
@@ -966,24 +966,25 @@ hb_gpu_draw_get_scale (const hb_gpu_draw_t *draw,
...
  * @draw: a GPU shape encoder
  * @font: font to draw from
  * @glyph: glyph ID to draw
  *
  * Convenience wrapper that draws a single glyph outline into the
- * encoder using hb_font_draw_glyph().  Calls hb_gpu_draw_set_scale()
- * with the font's scale before walking the outline.
+ * encoder using hb_font_draw_glyph_or_fail().  Calls
+ * hb_gpu_draw_set_scale() with the font's scale before walking the
+ * outline.
```
khaledhosny, line 977: Not strictly relate to this PR, but I’m thinking that `_or_fail()` API should try document the cases where it would fail and if enumerating that is not practical, at least give some examples.

src/hb-gpu-paint.cc
```diff
@@ -1156,47 +1156,68 @@ hb_gpu_paint_get_scale (const hb_gpu_paint_t *paint,
...
+ * @font: font to paint from
+ * @glyph: glyph ID to paint
+ *
+ * Feeds @glyph into the encoder.  Unlike
+ * hb_gpu_paint_glyph_or_fail(), non-color glyphs are handled
+ * transparently: harfbuzz synthesizes a single foreground-colored
+ * layer from the outline, which our callbacks turn into a single
+ * LAYER_SOLID op.
+ *
+ * Since: 14.0.0
```
khaledhosny, line 1207: This was already a new API before this change, so it should be `XSince: REPLACEME`.

src/hb-raster-paint.cc
```diff
@@ -2167,13 +2149,69 @@ hb_raster_paint_glyph (hb_raster_paint_t *paint,
...
+ * @glyph: glyph ID to paint
+ * @pen_x: glyph origin x in font coordinates (pre-transform)
+ * @pen_y: glyph origin y in font coordinates (pre-transform)
+ *
+ * Paints one glyph at (@pen_x, @pen_y).  Unlike
+ * hb_raster_paint_glyph_or_fail(), glyphs with no color paint data
+ * fall back to a synthesized foreground-colored outline, so any
+ * glyph with an outline or bitmap image produces output.
+ *
+ * Since: 13.0.0
```
khaledhosny, line 2203: It had `XSince: REPLACEME` before.

src/hb-vector-paint.cc
```diff
@@ -1328,15 +1328,16 @@ hb_vector_paint_custom_palette_color (hb_paint_funcs_t *pfuncs HB_UNUSED,
  *
  * Return value: `true` if glyph paint data was emitted, `false` otherwise.
  *
- * Since: 13.0.0
+ * XSince: REPLACEME
  */
```
khaledhosny, line 1332: This doc comment block should be removed since `hb_vector_paint_glyph_or_fail` is below and it has its doc comment.

src/hb-vector-paint.cc
```diff
@@ -1465,28 +1480,69 @@ hb_vector_paint_glyph (hb_vector_paint_t *paint,
...
+ * @pen_x: glyph x origin before context transform.
+ * @pen_y: glyph y origin before context transform.
+ * @extents_mode: extents update mode.
+ *
+ * Paints one glyph into @paint.  Unlike
+ * hb_vector_paint_glyph_or_fail(), glyphs with no color paint
+ * data fall back to a synthesized foreground-colored outline,
+ * so any glyph with an outline or bitmap image produces output.
+ *
+ * Since: 13.0.0
```
khaledhosny, line 1532: I _think_ if we are changing the return value, then we should reset the version as the old API is not strictly compatible. If we do this, we should do it to all the other function that changed their return value.

util/vector-output.hh
```diff
@@ -234,6 +234,21 @@ struct vector_output_t : output_options_t<>, view_options_t
...
+     * otherwise auto-detect via the font's color tables.
+     * hb_vector_paint_glyph synthesizes paint from outlines
+     * for mono fonts, so --paint works even without color. */
+    hb_face_t *face = hb_font_get_face (font);
+    bool font_has_color = hb_ot_color_has_paint (face) ||
+			  hb_ot_color_has_layers (face) ||
+#ifndef HB_NO_SVG
+			  hb_ot_color_has_svg (face) ||
+#endif
+			  hb_ot_color_has_png (face);
```
khaledhosny, line 247: Worth having as a public API? I have something in LibreOffice for PDF embedding, but it only checks the formats its PDF writer is prepared to support, but I imagine some HarfBuzz clients using that to decide whether to use paint or draw.

util/helper-cairo.hh
```diff
@@ -119,6 +120,24 @@ helper_cairo_create_scaled_font (const font_options_t *font_opts,
...
+   * otherwise auto-detect: route mono fonts through the
+   * cheaper outline path (NO_COLOR) and color fonts through
+   * the paint pipeline (cairo's default color mode). */
+  hb_face_t *hb_face = hb_font_get_face (font);
+  bool font_has_color = hb_ot_color_has_paint (hb_face) ||
+			hb_ot_color_has_layers (hb_face) ||
+#ifndef HB_NO_SVG
+			hb_ot_color_has_svg (hb_face) ||
+#endif
+			hb_ot_color_has_png (hb_face);
```
khaledhosny, line 134: We use it twice internally now, so at least an internal helper.

behdad: Well some backends handle SVG, others don't. Actually should we remove the has_svg here? Also, some backends handle bitmap, others don't. So I think a unified API is less useful here.

util/vector-output.hh
behdad: Maybe we can add per-consumer (raster / vector / gpu) "is font color" API. Not sure yet.

src/hb-vector-paint.cc
behdad: Right.

src/hb-gpu-draw.cc
behdad: or_fail() simply means there's no color tables. Yeah we should document better.

util/helper-cairo.hh
khaledhosny: Right!

Also yes, if we are not supporting SVG table any more, then `has_svg()` makes less sense as we will be drawing the fallback monochrome glyphs actually.

behdad: Okay, so all _has_svg calls from our utils need to go.

util/vector-output.hh
khaledhosny: Lets skip that for now, it is not really important, and clients might have to consider different tables differently (e.g. prefer to hand-roll COLRv0 table instead of going through paint).

behdad:
I accidentally merged this before addressing review comments. Pushed those to main now.

PR #5938, opened 2026-04-17
Fix ector pdf gradient opacity

behdad:


behdad:
Depends on https://github.com/harfbuzz/harfbuzz/pull/5937 landing first, then this becomes just two commits.

khaledhosny (approved):

PR #5939, opened 2026-04-17
[vector/raster] Set foreground/background API where it was missing

behdad:


behdad:
I'm merging this since I'm blocked on it. But please review @khaledhosny Thanks.

util/view-options.hh
```diff
@@ -109,9 +109,13 @@ struct view_options_t
     has_background = false;
     if (back && *back)
     {
-      if (!parse_color (back,
-			background_color.r, background_color.g,
-			background_color.b, background_color.a))
+      if (0 == g_ascii_strcasecmp (back, "none"))
```
khaledhosny, line 112: We need to document this special value somewhere.

src/hb-vector-draw.cc
```diff
@@ -692,7 +700,31 @@ hb_vector_draw_glyph_or_fail (hb_vector_draw_t *draw,
 					draw->x_scale_factor,
 					draw->y_scale_factor,
 					xx, yx, xy, yy, tx, ty);
-      hb_buf_append_str (&draw->body, "\"/>\n");
+      hb_buf_append_str (&draw->body, "\"");
+      /* Per-glyph fill from current foreground. */
```
khaledhosny, line 704: How does this handle overlapping glyphs, would the background of the next glyph cover the outlines of the previous one?

src/hb-raster-paint.cc
```diff
@@ -117,7 +117,10 @@ ensure_initialized (hb_raster_paint_t *c)
 			    hb_color_get_green (c->background),
 			    hb_color_get_red (c->background),
 			    hb_color_get_alpha (c->background));
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcast-align"
     uint32_t *pixels = (uint32_t *) hb_raster_image_get_buffer (root);
```
khaledhosny, line 122: The commit message says `Use hb_memcpy instead of casting`, but the commit is actually only ignoring the warning.

behdad: Oops yeah I changed it and amended without rewriting. Sorry.

src/hb-vector-draw.cc
behdad: Humm? Background is emitted once.

Or do you mean we are filling diff glyphs separately so translucent overlapping glyphs add up instead of not? That's an astute observation.

behdad: Postponed. Documented the limitation.

PR #5941, opened 2026-04-17
vector touchups

behdad:
Trying to clean up everything left.


behdad:
Each commit is going to be a separate change, so review separately. Thanks.

src/hb-vector.cc
```diff
@@ -39,3 +39,95 @@
...
+ * DAMAGE.
+ *
+ * THE COPYRIGHT HOLDER SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ * FITNESS FOR A PARTICULAR PURPOSE.  THE SOFTWARE PROVIDED HEREUNDER IS
+ * ON AN "AS IS" BASIS, AND THE COPYRIGHT HOLDER HAS NO OBLIGATION TO
+ * PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
+ *
+ * Author(s): Behdad Esfahbod
+ */
```
khaledhosny, line 66: This is a duplicate license header.

src/hb-vector.cc
```diff
@@ -39,3 +39,95 @@
...
+
+#include "hb.hh"
```
khaledhosny, line 68: Duplicate include (other includes should be moved up, too).

PR #5950, opened 2026-04-19
More review of AI code

behdad:
- [x] hb-raster
- [x] hb-vector
- [x] hb-gpu


behdad:
@khaledhosny Our enums typically have an _INVALID entry. `hb_vector_format_t` has it too. `hb_gpu_shader_lang_t` doesn't. Should I add one? The INVALID is useful to carry "is not set" information in clients.

khaledhosny:
> @khaledhosny Our enums typically have an _INVALID entry. `hb_vector_format_t` has it too. `hb_gpu_shader_lang_t` doesn't. Should I add one? The INVALID is useful to carry "is not set" information in clients.

I think it makes sense, yes.

behdad:
@khaledhosny This is ready for review. Thanks.

khaledhosny (approved):

PR #5958, opened 2026-04-21
[hb-config.hh] Tweak HB_TINY / HB_NO_DRAW / HB_NO_CFF relationship

### LINKED ISSUE #5955: Building fails with `HB_TINY` and undefining `HB_NO_CFF` without undefining `HB_NO_DRAW`
In harfbuzzjs repo, we build `harfbuzz-subset.wasm` with `-DHB_TINY` and the following config override files:
```
#undef HB_NO_CFF
#undef HB_NO_OT_FONT_CFF
#undef HB_NO_SUBSET_CFF
#undef HB_NO_SUBSET_LAYOUT
#undef HB_NO_VAR
#undef HB_NO_STYLE
```

But the build fails with:

> wasm-ld: error: /var/folders/kg/rc0yqzr55d54sf3ygnqlbt8r0000gp/T/emscripten_temp_fqfj8ui5/harfbuzz-subset.o: undefined symbol: OT::cff2::accelerator_t::get_extents(hb_font_t*, unsigned int, hb_glyph_extents_t*) const

Unless we add to the config override:
```
#undef HB_NO_DRAW
```

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/5955

### DIFF
--- a/src/hb-config.hh
+++ b/src/hb-config.hh
@@ -136,7 +136,7 @@
 #endif
 
 #ifdef HB_NO_DRAW
-#define HB_NO_CFF
+#define HB_NO_OT_FONT_CFF
 #define HB_NO_OUTLINE
 #define HB_NO_PAINT
 #endif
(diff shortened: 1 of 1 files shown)


khaledhosny:
How is that intended to be used, should I change the harfbuzzjs override file to be a `last` one?

khaledhosny:
> How is that intended to be used, should I change the harfbuzzjs override file to be a `last` one?

This seems to work, thanks!

src/hb-config.hh
```diff
@@ -136,7 +136,7 @@
 #endif
 
 #ifdef HB_NO_DRAW
-#define HB_NO_CFF
+#define HB_NO_OT_FONT_CFF
```
behdad, line 139: @khaledhosny I thought this might fix it for you. The other change was a last-resort. Did you use the last.h, or this fixed it?

khaledhosny: I had to use last.h

behdad: Okay lemme look into it tomorrow more.

PR #5968, opened 2026-04-28
Allow line-ending changes in gen-gpu-shader-artifacts.py

### LINKED ISSUE #5967: gen-gpu-shader-artifacts.py line endings issue
Hi, I ran into an issue on the vcpkg build servers where one of the Windows builds randomly failed in gen-gpu-shader-artifacts.py with "hb-gpu-fragment-glsl.hh is out of date; regenerate with a writable source tree"

I think there is an issue in the `OSError` handler in gen-gpu-shader-artifacts.py: the script operates on files in text mode, but the final validation check when the copy back to source fails is a binary check. If the `git clone` was done with a non-native line ending and the source tree is non-writable (or the copy randomly fails, as seems to have happened on the build agent in this case), the output file will end up with different line endings and the check will fail.

ajtribick:
Fixes #5967

Avoids failing the build when the repository is checked out with non-native line endings and the source tree is non-writable. An alternative strategy might be to detect the line endings from the source and use that to write the output, but that feels unnecessarily complex.

### DIFF
--- a/src/gen-gpu-shader-artifacts.py
+++ b/src/gen-gpu-shader-artifacts.py
@@ -38,6 +38,7 @@ src_copy = os.path.join(CURRENT_SOURCE_DIR, hh)
 try:
     shutil.copyfile(OUTPUT, src_copy)
 except OSError:
-    import filecmp
-    if not filecmp.cmp(OUTPUT, src_copy, shallow=False):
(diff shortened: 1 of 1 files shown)


behdad:
Thanks.

PR #5969, opened 2026-04-29
[shape] Mongolian: force NNBSP/MVS to .notdef when font lacks neighbors

### LINKED ISSUE #4503: Consider providing option for restricting space synthesis to stay within unicode-range
See details in https://crbug.com/1499787

When shaping `--unicodes 180e,1821,202f,1836,1822` with Noto Sans Mongolian subsetted to latin - the U+202F space is synthesized with the space from the ASCII range, which then leads to breaking shaping down the line when shaping with the Mongolian subset. 

Space synthetisation is generally useful and we don't want to plainly switch it off, but it would be useful if HarfBuzz could be told to stay within a specified unicode range or call back to the client to ask "can_synthesize_space_for?" with the codepoint from the input buffer for which a space is about to be synthesized. 

Your thoughts are welcome on this issue.

Satsrag:
In Mongolian, U+202F (NARROW NO-BREAK SPACE) and U+180E (MONGOLIAN VOWEL SEPARATOR) carry script-meaningful semantics that depend on the adjacent letters. A font that has a glyph for U+202F or U+180E but does not cover the adjacent Mongolian letters cannot shape the cluster correctly.

Two real-world examples on macOS:

  1. Helvetica and SF Pro carry U+202F (for French typography) but contain no Mongolian letters. With Chrome's default cascade, NNBSP gets resolved to Helvetica's narrow space glyph, which prevents the rest of the cluster from being re-shaped by the actual Mongolian font; the surrounding Mongolian letters end up in their isolated forms instead of the proper init/final variants.

  2. A unicode-range Latin slice cut from Noto Sans Mongolian keeps U+0020 but drops U+202F, U+180E, and Mongolian letters. hb_ot_hide_default_ignorables() then masks an unmapped U+180E with the U+0020 invisible glyph, again short-circuiting cascade.

Detect the situation at glyph resolution time. When shaping Mongolian script and the current codepoint is U+202F or U+180E, inspect the adjacent input characters: if any neighbor is in script Mongolian and the active font does not provide a glyph for that neighbor, the font cannot complete the cluster, so return .notdef instead of the font's own glyph (or any synthesized fallback). The cluster is also marked
unsafe-to-concat because the decision is context-dependent.

This mirrors what Firefox does in gfxTextRun::FindFontForChar (treating these characters as cluster extenders during font selection) but at the shaper level, so every shaper-driven font cascade -- Chrome, Edge, Linux WebKit, Pango, Skia, etc. -- benefits without per-engine plumbing.

The hide_default_ignorables() path is also adjusted to preserve .notdef for default-ignorables in Mongolian script when the font genuinely lacks the codepoint, so MVS forced to .notdef in normalize survives until the caller can act on it.

All other scripts retain the existing behavior: NNBSP still falls back to a half-advance space, MVS is still hidden via the U+0020 invisible glyph, French-style "word NNBSP !" still renders correctly with Helvetica.

Tested with `meson test -C build` (240/240 passing). Adds regression tests in test/shape/data/in-house/tests/ mongolian-cascade.tests covering three scenarios:

  A. A Latin-only font that has neither U+202F, U+180E nor any
     Mongolian letter -- the original issue scenario.

  B. A subset of a Mongolian font that keeps only U+0020,
     U+202F, U+180E and U+180B-U+180D (no Mongolian letters)
     -- the macOS Helvetica / SF Pro situation. In Mongolian
     context the font's real U+202F / U+180E glyph is bypassed
     for .notdef; standalone U+202F (no Mongolian neighbors)
     still gets the font's own `uni202F.nobreak` glyph; Latin
     context is unchanged. The trap font is added at
     test/shape/data/in-house/fonts/
     e0196734431776b0d1f147c6222fb9028d40e7e0.ttf.

  C. The full Mongolian font already in the test corpus -- the
     fix must NOT change behaviour because the font can shape
     the cluster, so HarfBuzz uses its real NNBSP / MVS
     glyphs and the surrounding Mongolian letters get their
     proper context-aware variant forms.

Manually verified on macOS that the patched HarfBuzz makes a Helvetica-then-Noto-Sans-Mongolian cascade produce the same shape output as shaping with Noto Sans Mongolian alone, including the context-dependent NNBSP and MVS variant forms.

Fixes: https://github.com/harfbuzz/harfbuzz/issues/4503
Assisted-by: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


src/hb-ot-shape-normalize.cc
```diff
@@ -153,6 +153,48 @@ decompose_current_character (const hb_ot_shape_normalize_context_t *c, bool shor
   hb_codepoint_t u = buffer->cur().codepoint;
   hb_codepoint_t glyph = 0;
 
+  /* In Mongolian, NARROW NO-BREAK SPACE (U+202F) and MONGOLIAN VOWEL
+   * SEPARATOR (U+180E) carry script-meaningful semantics that depend on
+   * the adjacent letters: a font that has a glyph for U+202F or U+180E
+   * (e.g. Helvetica or San Francisco for French typography, or a
+   * unicode-range Latin slice cut from Noto Sans Mongolian) but does
+   * not cover the adjacent Mongolian letters cannot shape the cluster
+   * correctly. Force .notdef in that case so the caller can cascade to
```
drott, line 162: What do we break compared to current status-quo if no such font is (edit:) found?

Satsrag: My understanding (please correct me if I'm wrong):                          
              
  - No Mongolian font in cascade: status quo already renders the Mongolian letters as .notdef; this PR just makes U+202F/U+180E .notdef too. A few more tofu boxes in an already-broken cluster.                                   
  - Mongolian font available: status quo shapes with wrong letter variants — visible but full of errors, not readable. This PR cascades correctly.       

And in practice macOS/Windows/Linux/Android/iOS all ship a Mongolian font by default, so the first case is rare.                                                                          

Does this match your concern, or is there a scenario I'm missing?

drott: > Does this match your concern, or is there a scenario I'm missing?

I think so. So with existing fonts, we would not worsen the shape result? Or do we need the fonts to be newer than what is currently installed as system fonts on Mac or Windows?

Satsrag: > > Does this match your concern, or is there a scenario I'm missing?
> 
> I think so. So with existing fonts, we would not worsen the shape result? Or do we need the fonts to be newer than what is currently installed as system fonts on Mac or Windows?

Yes, existing system fonts work as-is — no font updates needed. The fix is in the shaper itself: it just lets the cascade fall through to the already-installed Mongolian font (e.g. Mongolian Baiti on Windows, Noto Sans Mongolian on macOS), which already shapes correctly. So users benefit immediately without updating anything.

One disclaimer: I provided the high-level approach, but the implementation was done with help from Claude Code. The overall direction should be sound, but I'd really appreciate a careful review of the code-level details there may be edge cases or style issues I'm not in the best position to catch.

khaledhosny:
IMO, HarfBuzz is the wrong level to address this. Chrome must have some code that keep grapheme clusters together when doing font fallback (i.e. if the font has a partial coverage for a grapheme cluster, it tries the fallback font for the whole cluster), if does not it should. While this is not a case of partial grapheme clusters, it is conceptually the same, and Chrome should not use the spaces from a font and the surrounding Mongolian characters from a different font.

khaledhosny:
Another reason this should be addressed in Chrome, if the first font has the space characters (i.e. they are not synthesized by HarfBuzz), Mongolian shaping would fail the same way even with this change.

khaledhosny:
> Another reason this should be addressed in Chrome, if the first font has the space characters (i.e. they are not synthesized by HarfBuzz), Mongolian shaping would fail the same way even with this change.

Actually the change here rejects the code point even if the font supports it, which is really not something HarfBuzz should be doing at all. HarfBuzz has no concept of a font cascading list, it deals with one font at a time, so it should never reject a code point it is asked to shape.

drott:
> IMO, HarfBuzz is the wrong level to address this. Chrome must have some code that keep grapheme clusters together when doing font fallback (i.e. if the font has a partial coverage for a grapheme cluster, it tries the fallback font for the whole cluster), if does not it should.

We do this, but we let a primary font capture and render space characters - also we allow space synthetisation with the primary font. It would be tricky and a bit of overhead to special case whitespace in the shaper driven approach. We don't really want to postpone whitespace shaping to after non-whitespace stages are done.

khaledhosny:
> > IMO, HarfBuzz is the wrong level to address this. Chrome must have some code that keep grapheme clusters together when doing font fallback (i.e. if the font has a partial coverage for a grapheme cluster, it tries the fallback font for the whole cluster), if does not it should.
> 
> We do this, but we let a primary font capture and render space characters - also we allow space synthetisation with the primary font. It would be tricky and a bit of overhead to special case whitespace in the shaper driven approach. We don't really want to postpone whitespace shaping to after non-whitespace stages are done.

I don’t know the relevant Chrome code to know how feasible is this, but I think something along these lines should work: If this is U+202F or U+180E and the resolved script is Mongolian and the font has no Mongolian support (or the surrounding code points returned `.notdef` by HarfBuzz) try the next fallback font for these as well, unless this is the last font).

Satsrag:
Thanks for the architectural feedback. There's already a Chromium issue for this —https://crbug.com/1499787 — but it hasn't seen movement, which is why I attempted a HarfBuzz-side fix. Mongolian users on Chromium-based browsers have been hitting this in input fields, search boxes, and comment forms for many years.

Test strings that reproduce in this very comment view (system default font hits the case directly):
- macOS ≥ 15.1: ᠠᠨ᠎ᠠ᠎ᠶᠢᠨ — U+1820 U+1828 U+180E U+1820 U+180E U+1836 U+1822 U+1828
- macOS < 15.1 / Windows: ᠠᠨ᠎ᠠ ᠶᠢᠨ — U+1820 U+1828 U+180E U+1820 U+202F U+1836 U+1822 U+1828                                                                                                                                                           

Reference images of expected rendering below. 

Untitled

Really hoping we can find some path forward here — happy to reshape this PR if there's a HarfBuzz-side approach that would be acceptable.

Satsrag:
Following up on @khaledhosny's suggestion above, I drafted a Blink-side                                                                                                                                
patch implementing it. Browseable here:                                                                                                                                                                

[0001-mongolian-cascade.patch](https://github.com/user-attachments/files/27230336/0001-mongolian-cascade.patch)                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                         
Please take a look and help land it if it looks fine. Thanks!                                                                                                                                         
                                                                                                                                                                                                         
Thanks for the guidance so far.                                                                                                                                                                        
                                                                                                                                                              
  cc @drott @khaledhosny        
                                                       
```c
diff --git a/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_face.cc b/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_face.cc
index eb5939a..930c297 100644
--- a/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_face.cc
+++ b/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_face.cc
@@ -398,6 +398,26 @@ unsigned HarfBuzzFace::UnitsPerEmFromHeadTable() {
   return hb_face_get_upem(face);
 }
 
+bool HarfBuzzFace::HasOpenTypeMongolianSupport() {
+  if (harfbuzz_font_data_->mongolian_gsub_ ==
+      HarfBuzzFontData::GsubScriptSupport::kUnknown) {
+    hb_face_t* face =
+        hb_font_get_face(harfbuzz_font_data_->unscaled_font_.get());
+    DCHECK(face);
+    // OpenType registers a single Mongolian script tag, 'mong'
+    // (Unicode block U+1800-U+18AF / ISO 15924 'Mong').
+    // https://learn.microsoft.com/en-us/typography/opentype/spec/scripttags
+    unsigned int script_index = 0;
+    const bool found = hb_ot_layout_table_find_script(
+        face, HB_OT_TAG_GSUB, HB_TAG('m', 'o', 'n', 'g'), &script_index);
+    harfbuzz_font_data_->mongolian_gsub_ =
+        found ? HarfBuzzFontData::GsubScriptSupport::kPresent
+              : HarfBuzzFontData::GsubScriptSupport::kNotPresent;
+  }
+  return harfbuzz_font_data_->mongolian_gsub_ ==
+         HarfBuzzFontData::GsubScriptSupport::kPresent;
+}
+
 Glyph HarfBuzzFace::HbGlyphForCharacter(UChar32 character) {
   hb_codepoint_t glyph = 0;
   HarfBuzzGetNominalGlyph(harfbuzz_font_data_->unscaled_font_.get(),
diff --git a/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_face.h b/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_face.h
index 4469f7b..99cb3c6 100644
--- a/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_face.h
+++ b/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_face.h
@@ -83,6 +83,13 @@ class PLATFORM_EXPORT HarfBuzzFace final
   unsigned UnitsPerEmFromHeadTable();
   Glyph HbGlyphForCharacter(UChar32 character);
 
+  // Returns true if this font's GSUB declares the OpenType 'mong'
+  // script tag. Used to decide whether the font can correctly shape
+  // clusters that contain U+202F NNBSP or U+180E MVS in Mongolian
+  // context. The result is cached on the underlying HarfBuzzFontData.
+  // crbug.com/1499787
+  bool HasOpenTypeMongolianSupport();
+
   hb_codepoint_t HarfBuzzGetGlyphForTesting(UChar32 character,
                                             UChar32 variation_selector);
 
diff --git a/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_font_data.h b/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_font_data.h
index 9d637b5..2e2c171 100644
--- a/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_font_data.h
+++ b/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_font_data.h
@@ -101,6 +101,14 @@ struct HarfBuzzFontData final : public GarbageCollected<HarfBuzzFontData> {
   SpaceGlyphInOpenTypeTables space_in_gsub_ =
       SpaceGlyphInOpenTypeTables::kUnknown;
 
+  enum class GsubScriptSupport { kUnknown, kPresent, kNotPresent };
+
+  // Whether this font's GSUB advertises the OpenType 'mong' script
+  // tag. Lazily computed; used to decide whether the font can
+  // correctly shape clusters around U+202F NNBSP and U+180E MVS in
+  // Mongolian context. crbug.com/1499787, harfbuzz:4503
+  GsubScriptSupport mongolian_gsub_ = GsubScriptSupport::kUnknown;
+
   Member<OpenTypeVerticalData> vertical_data_;
   Member<const UnicodeRangeSet> range_set_;
 };
diff --git a/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_shaper.cc b/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_shaper.cc
index 8cd3a8a..1f0fcca 100644
--- a/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_shaper.cc
+++ b/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_shaper.cc
@@ -188,6 +188,15 @@ FontFallbackPriority ApplyFontVariantEmojiOnFallbackPriority(
   return curr_font_fallback_priority;
 }
 
+// Returns true if `c` is a Mongolian context-sensitive format character
+// (NARROW NO-BREAK SPACE or MONGOLIAN VOWEL SEPARATOR) whose cluster
+// requires the surrounding letters' font for correct shaping.
+// crbug.com/1499787, harfbuzz:4503
+inline bool IsMongolianContextSpace(UChar c) {
+  return c == uchar::kNarrowNoBreakSpace ||
+         c == uchar::kMongolianVowelSeparator;
+}
+
 }  // namespace
 
 enum ReshapeQueueItemAction {
@@ -592,6 +601,23 @@ void HarfBuzzShaper::ExtractShapeResults(
       // We revisit when HarfBuzz decides how to solve this more generally.
       // https://github.com/harfbuzz/harfbuzz/issues/2889
       glyph_result = kNotDef;
+    } else if (current_run_script == USCRIPT_MONGOLIAN &&
+               !IsLastFontToShape(fallback_stage) &&
+               IsMongolianContextSpace(text_[current_cluster]) &&
+               !current_font->PlatformData()
+                    .GetHarfBuzzFace()
+                    ->HasOpenTypeMongolianSupport()) {
+      // U+202F (NNBSP) and U+180E (MVS) carry script-meaningful semantics
+      // in Mongolian that depend on adjacent letter shaping. If the active
+      // font's GSUB does not declare a Mongolian script tag, it cannot
+      // shape the cluster correctly even if it has a glyph for the format
+      // character itself, so force fallback. Without this, fonts that carry
+      // U+202F for French typography (e.g. Helvetica, SF Pro) or carry
+      // U+1820 for fallback coverage but no Mongolian shaping (e.g. Heiti
+      // TC on macOS) "capture" the NNBSP / MVS cluster and prevent the
+      // surrounding Mongolian letters from being re-shaped together with
+      // it. crbug.com/1499787, harfbuzz:4503
+      glyph_result = kNotDef;
     } else if (glyph_id == kUnmatchedVSGlyphId) {
       fallback_stage = ChangeStageToVS(fallback_stage);
       glyph_result = kNotDef;
diff --git a/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_shaper_test.cc b/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_shaper_test.cc
index 1de050c..434613a 100644
--- a/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_shaper_test.cc
+++ b/third_party/blink/renderer/platform/fonts/shaping/harfbuzz_shaper_test.cc
@@ -800,6 +800,59 @@ TEST_F(HarfBuzzShaperTest, IdeographicSpace) {
   EXPECT_EQ(run_font_data.size(), 1u);
 }
 
+// crbug.com/1499787, harfbuzz:4503
+// In Mongolian, U+202F (NARROW NO-BREAK SPACE) carries script-meaningful
+// semantics that depend on the adjacent letters. Fonts like Helvetica or
+// SF Pro on macOS carry U+202F for French typography but no Mongolian
+// letters; without the cascade fix, the primary font "captures" NNBSP
+// and prevents the surrounding Mongolian letters from being shaped
+// together with it, splitting the run into multiple font segments.
+//
+// Locale is intentionally set to English (or any non-Mongolian) to
+// reproduce the real bug: a `<p>...</p>` with no `lang` attribute and
+// no Mongolian font-family. Setting "mn" would bias the cascade
+// straight to Noto Sans Mongolian and trivially pass the assertion
+// without exercising the fix.
+TEST_F(HarfBuzzShaperTest, MongolianNNBSPCascade) {
+  font_description.SetLocale(LayoutLocale::Get(AtomicString("en")));
+  Font* font = MakeGarbageCollected<Font>(font_description);
+
+  String string(
+      u"\u1820"    // MONGOLIAN LETTER A
+      u"\u202f"    // NARROW NO-BREAK SPACE
+      u"\u1822");  // MONGOLIAN LETTER I
+  HarfBuzzShaper shaper(string);
+  const ShapeResult* result = shaper.Shape(font, TextDirection::kLtr);
+  HeapVector<ShapeResult::RunFontData> run_font_data;
+  result->GetRunFontData(&run_font_data);
+  // All three codepoints must end up in the same font run -- the
+  // Mongolian fallback, not the Latin primary for U+202F.
+  EXPECT_EQ(run_font_data.size(), 1u);
+}
+
+// crbug.com/1499787, harfbuzz:4503
+// Same idea for U+180E (MONGOLIAN VOWEL SEPARATOR). The four-codepoint
+// input "a-n-MVS-a" is a real Mongolian usage pattern; its shaping
+// produces context-dependent variant forms (n.fina1 before MVS, a.fina2
+// after MVS) that only the Mongolian font can deliver. As above,
+// locale stays non-Mongolian to avoid biasing the cascade and
+// short-circuiting the test.
+TEST_F(HarfBuzzShaperTest, MongolianMVSCascade) {
+  font_description.SetLocale(LayoutLocale::Get(AtomicString("en")));
+  Font* font = MakeGarbageCollected<Font>(font_description);
+
+  String string(
+      u"\u1820"    // MONGOLIAN LETTER A
+      u"\u1828"    // MONGOLIAN LETTER NA
+      u"\u180e"    // MONGOLIAN VOWEL SEPARATOR
+      u"\u1820");  // MONGOLIAN LETTER A
+  HarfBuzzShaper shaper(string);
+  const ShapeResult* result = shaper.Shape(font, TextDirection::kLtr);
+  HeapVector<ShapeResult::RunFontData> run_font_data;
+  result->GetRunFontData(&run_font_data);
+  EXPECT_EQ(run_font_data.size(), 1u);
+}
+
 #if BUILDFLAG(IS_MAC) || BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_WIN)
 TEST_F(HarfBuzzShaperTest, SystemEmojiVS15) {
   ScopedSystemFallbackEmojiVSSupportForTest scoped_feature_system_emoji_vs(
diff --git a/third_party/blink/renderer/platform/wtf/text/character_names.h b/third_party/blink/renderer/platform/wtf/text/character_names.h
index 336d6ab..01a51ae 100644
--- a/third_party/blink/renderer/platform/wtf/text/character_names.h
+++ b/third_party/blink/renderer/platform/wtf/text/character_names.h
@@ -107,6 +107,7 @@ const UChar kEthiopicPrefaceColon = 0x1366;
 const UChar kEthiopicNumberHundred = 0x137B;
 const UChar kEthiopicNumberTenThousand = 0x137C;
 const UChar kMongolianFreeVariationSelectorTwo = 0x180C;
+const UChar kMongolianVowelSeparator = 0x180E;
 const UChar kMongolianLetterA = 0x1820;
 
 // U+2***
@@ -133,6 +134,7 @@ const UChar kRightToLeftEmbedding = 0x202B;
 const UChar kPopDirectionalFormatting = 0x202C;
 const UChar kLeftToRightOverride = 0x202D;
 const UChar kRightToLeftOverride = 0x202E;
+const UChar kNarrowNoBreakSpace = 0x202F;
 const UChar kPerMilleSign = 0x2030;
 const UChar kPerTenThousandSign = 0x2031;
 const UChar kOverline = 0x203E;
```

drott:
> I don’t know the relevant Chrome code to know how feasible is this, but I think something along these lines should work: If this is U+202F or U+180E and the resolved script is Mongolian and the font has no Mongolian support (or the surrounding code points returned `.notdef` by HarfBuzz) try the next fallback font for these as well, unless this is the last font).

I agree the best outcome seems connected to knowledge about font fallback. We can also not efficiently identify what is the "last font with any Mongolian coverage", last resort on Mac OS for example is the Mac OS Last Resort font which just shows which Unicode code block a glyph is in. I'd generally prefer not to go down to this level of granularity outside the shaper, and trying to fix up shape results with special cases in Blink. Is there no generic solution we can implement at the shaper level? 

Or I might need to go back and consider if there would be an option to do whitespace handling differently, while keeping efficiency.

Best outcomes are generally achieved if the content is formatted well with a suitable primary font. (I see the bug also applies to omnibox shaping, there is no control over font choice there.)

khaledhosny:
The shaper is working as intended, it is Chrome that is segmenting the text and sending incomplete (for Mongolian) chunks to HarfBuzz. Addressing this in HarfBuzz would be fixing a segmentation issue in the shaper (which does not do segmentation), not the other way around.

khaledhosny:
I think in general, it would be a good practice to always make spaces use the same font as the surrounding text as much as possible. A font might have wider or narrower white-space and it is generally part of its design, but since almost all fonts have space glyph, it is guaranteed to be taken from the first font on the stack, regardless of what font the rest of characters are taking from.

This is outright broken for Mongolian, but it is also less than ideal for any other script. Some fonts even have substitutions or kerning involving space, which will not work when the space glyph is taken from a different font. This have certainly affected me before more than once, with spacing looking off because space and the surrounding text are taken from two different fonts.

Satsrag:
Hi @drott — friendly check-in on progress here.

For reference, here's how other engines handle the same Mongolian space-class capture bug — might be useful as design reference for the Chromium-side approach:

- **Pango** (`pango/itemize.c::consider_as_space`): broadest — defers any `gc=Zs` / `Cf` / variation-selector to the neighbour font via `font=NULL` + trailing-space promotion. Empirically passes pure-Mong NNBSP and U+0020 cases.
- **Firefox** (`gfxTextRun.cpp::FindFontForChar`): per-character heuristics — NNBSP-specific `if` branch with forward + backward neighbour-font lookup; `IsClusterExtender + IsDefaultIgnorable` for ZWJ / ZWNJ / FVS.
- **Minikin** (`FontCollection.cpp::itemize`): `stickyAllowlist` (NNBSP, hyphens, …) + `doesNotNeedFontSupport` (ZWJ, ZWNJ, BiDi controls, variation selectors).
- **Chromium / Skia / Flutter** (current upstream): cascade-after-shape, no Mongolian-specific protection — currently affected.

Skia / Flutter has **no equivalent patch in flight** (verified against `google/skia` mainline `OneLineShaper.cpp::sortOutGlyphs`), so without an HB-side fix they stay structurally broken indefinitely.

Is there anything I can do to help unblock things — either here on the HB PR or on the Chromium-side patch ([crbug.com/1499787](https://crbug.com/1499787))? Happy to share more test data, run cross-engine validation, or write up specific aspects.

Satsrag:
@drott @khaledhosny — based on the discussion here, I filed a Chromium CL:

  https://chromium-review.googlesource.com/c/chromium/src/+/7849879

  Could you take a look when you have time?
  
  Design notes: https://resonant-cheesecake-be228e.netlify.app/mongolian-shaping-fix.html

  Thanks!

behdad:
I'm hesitant to make a change like this in HB. I suggest you move the discussion to relevant downstream projects instead. Thanks.

PR #5978, opened 2026-05-06
Add Flutter to list of users in README.md

metaspook:
Flutter is a free, open-source, Google-developed framework for creating natively compiled applications for mobile, web, and desktop from a single codebase using the Dart language. It is unique because it draws its own UI using the Skia or Impeller graphics engines rather than relying on native platform components, ensuring consistent design across platforms.

Why Flutter uses HarfBuzz?

Complex Text Shaping: 
HarfBuzz is a text-shaping engine that converts Unicode text into correctly shaped and positioned glyphs.

Ensuring Proper Display: 
It is essential for rendering complex scripts (e.g., Arabic, Indic, Tamil) where characters change shape based on neighbors.

Cross-Platform Consistency: 
Because Flutter draws its own UI, it needs a reliable, embedded engine like HarfBuzz to handle text rendering identically across iOS, Android, and Desktop, regardless of the operating system's native capabilities

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ steel background.
 
 ## Users
 
-HarfBuzz is used in Android, Chrome, ChromeOS, Firefox, GNOME, GTK+, KDE,
+HarfBuzz is used in Android, Chrome, ChromeOS, Firefox, Flutter, GNOME, GTK+, KDE,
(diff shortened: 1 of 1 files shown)


behdad:
Thank you.

PR #5983, opened 2026-05-13
Fix warnings in embedder build with newer rustc

drott:
Wrap no_mangle in unsafe() to address compiler warnings with newer rust compilers/editions.


drott:
Rustc from 1.82 should support this - so does not conflict with requirements from Cargo.toml.

PR #5986, opened 2026-05-18
build: Fix building hb-gpu on Windows

fanc999-1:
Hi,

This attempts to update the Meson build files to fix building `hb-gpu` on Windows:

* Use `-DNOMINMAX` when building on Visual Studio-style compilers, so that `std::min` and `std::max` can be used properly.
* Correct the dependency for linking to `D3DCompiler_47.dll`-the library to look for should be `D3DCompiler.lib` (or so for MinGW).

With blessings, thank you!

### DIFF
--- a/util/gpu/meson.build
+++ b/util/gpu/meson.build
@@ -64,6 +64,6 @@ if not gpu_demo_opt.disabled()
     metal_dep = disabler()
   endif
   d3d11_dep = meson.get_compiler('cpp').find_library('d3d11', required: false)
-  d3dcompiler_dep = meson.get_compiler('cpp').find_library('d3dcompiler_47', required: false)
+  d3dcompiler_dep = meson.get_compiler('cpp').find_library('d3dcompiler', required: false)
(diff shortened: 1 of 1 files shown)


util/meson.build
```diff
@@ -221,16 +221,20 @@ if not get_option('gpu').disabled()
   subdir('gpu')
 
   if have_gpu_demo_deps and conf.get('HAVE_GLIB', 0) == 1
+    extra_cpu_demo_cpp_args = []
```
behdad: 'cpu' -> 'gpu'.

behdad:
> * Use `-DNOMINMAX` when building on Visual Studio-style compilers, so that `std::min` and `std::max` can be used properly.

I already fixed the min/max issue in main by not using `std` versions. Please update?

> * Correct the dependency for linking to `D3DCompiler_47.dll`-the library to look for should be `D3DCompiler.lib` (or so for MinGW).

Looks good to me. I'm away from my mingw setup though, I assume it works on cross-compiler still.

util/meson.build
fanc999-1: Oops, although this is now obsolete.

PR #5988, opened 2026-05-18
[algs] Saturating arithmetic for get_size() on 32-bit

behdad:
On 32-bit platforms size_t is unsigned int, so the typical get_size() computation `min_size + count * static_size` silently wraps when the product exceeds UINT_MAX.  The truncated value then flows into sanitize / serialize as a small but plausible size, defeating bounds checks.

Add three helpers in hb-algs.hh:

  hb_unsigned_mul_saturate (a, b, ...)
  hb_unsigned_add_saturate (a, b, ...)
  hb_unsigned_mul_add_saturate (a, b, c)   /* a*b + c */

They saturate to SIZE_MAX on overflow.  Variadic overloads left-fold so 3+ arguments work.  On 64-bit (size_t wider than unsigned int), all three reduce to plain a*b / a+b — the inputs to size computations are at most unsigned int so the result fits in size_t without overflow.

Convert every get_size() that does arithmetic over variable inputs:

  hb-open-type.hh        UnsizedArrayOf, ArrayOf,
                         HeadlessArrayOf, ArrayOfM1,
                         sorted-array-with-header, CFFIndex
  hb-ot-layout-common.hh CmapSubtableFormat-style,
                         VarRegionList, VarData, MultiVarData,
                         DeltaSetIndexMapFormat01
  hb-ot-hdmx-table.hh    DeviceRecord, hdmx
  hb-ot-cff-common.hh    Charset0, Charset wrapper
  hb-ot-cff1-table.hh    Encoding, Charset wrapper
                         (also widened unsigned accumulator
                         to size_t)
  hb-ot-os2-table.hh     OS/2 (widened accumulator to size_t)
  hb-ot-var-common.hh    TupleVariationHeader,
                         TupleVariationData accumulator
  hb-ot-var-hvar-table.hh DeltaSetIndexMap
  hb-vector.hh           hb_vector_t::get_size

When saturation kicks in, check_range() and extend_size() naturally reject the resulting value (SIZE_MAX truncated to unsigned is UINT_MAX, far past any real blob; extend_size already rejects sizes >= INT_MAX).

Assisted-by: Claude <noreply@anthropic.com>

### DIFF
--- a/src/hb-algs.hh
+++ b/src/hb-algs.hh
@@ -1215,6 +1215,62 @@ hb_unsigned_add_overflows (unsigned int a, unsigned int b, unsigned *result = nu
   return b > (unsigned int) -1 - a;
 }
 
+/* Saturating arithmetic on size_t.  On platforms where size_t is wider than
+ * unsigned int (i.e. 64-bit), the inputs to get_size()-style computations
+ * (counts and static_sizes which are at most 32-bit) cannot overflow size_t,
+ * so these reduce to plain arithmetic.  On 32-bit platforms (size_t is
+ * unsigned int), they saturate to SIZE_MAX so sanitize/serialize callers
+ * naturally reject the resulting size as out-of-range. */
+
+static inline size_t
+hb_unsigned_mul_saturate (size_t a, size_t b)
+{
+  if (sizeof (size_t) > sizeof (unsigned int))
+    return a * b;
+#if hb_has_builtin(__builtin_mul_overflow)
+  size_t result;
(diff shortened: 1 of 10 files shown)


jfkthame (approved):
Yes, this looks reasonable, thanks.

(Initially I was a little concerned whether just using saturating arithmetic could lead to under-allocation at a callsite that uses the size to allocate space -- e.g. during subsetting -- but since extend_size() will reject SIZE_MAX it looks fine.)

behdad:
Right. It repurposes SIZE_MAX to mean overflow.

PR #5989, opened 2026-05-18
Cross-platform clippy-friendly casts in FFI

drott:
Bindgen generates a std::os::raw::c_char as return type for functions like hb_blob_get_data. Depending on platform/arch, this can be an i8 or u8. In the Rust side code, receiving the data, these are interpreted as u8. Perform the pointer casts in a way that compiles without clippy warnings about unnecessary casts when the c_char type was already u8.


behdad:
Thanks. Please merge when bots happy.

PR #5992, opened 2026-05-18
[subset] Fix MATH closure to handle transitive glyph dependencies

skef:
I found this while adding test support for the subset test fonts in #5733. The parity checker showed a discrepancy with some subsets of STIXTwoMath and in this case I think the closure code needs updating.

The problem is straightforward -- MATH glyph constructions can reference glyphs that themselves have MATH constructions, so you need to loop or recurse at some level. The particular intervention here was based on looking at the various analogous cases (GSUB, glyf, COLRv1) and choosing a semantic that seemed closest to those other cases. 

I also tried to add some tests to capture the problem.


khaledhosny:
> MATH glyph constructions can reference glyphs that themselves have MATH constructions, so you need to loop or recurse at some level.

I’m not sure if this recursion is useful. Using `MATH` construction does not apply such recursion (unlike, say, GSUB), so if the glyph in not accessible through other means (`cmap` or `GSUB`) its own construction will be inaccessible.

skef:
@khaledhosny you're right, and I had misunderstood the nature of this problem, but there is still a problem with the subsetting path. 

The easiest (and perhaps only) way to see this with STIXTwoMath is to run `hb-subset --unicodes=U+21D4` on it, which produces no output. This is because When ⇐ is retained (as an assembly part of ⇔), its construction is copied, which references uni21D0.endr, but that glyph isn't in the subset.

This PR represents one way of fixing the problem, by continuing to add glyphs until everything referenced has been included. 

The other way would be to strip constructions for glyphs included in this way. Here's the specific advice on that a little bird told me:

> In `MathVariants::subset`, only emit constructions for glyphs that were in the set *before* MATH closure (i.e. from cmap/GSUB), not for glyphs added by MATH closure. The glyph outlines are still retained (needed for drawing the assembly), but their own MATH constructions are dropped, so there are no dangling references. This would mean filtering `collect_coverage_and_indices` by `_glyphset_gsub` instead of `_glyphset_mathed` — the other MATH subset methods (italics correction, accent attachment, kern info) would continue using `_glyphset_mathed` since those records are needed for all retained glyphs.

khaledhosny:
I see. I’d be inclined to match fonttools subset output.

skef:
Should I close this and open an issue?

behdad:
> Should I close this and open an issue?

That would work, sure. Thanks.

skef:
Closed in favor of #5998

PR #5995, opened 2026-05-21
Add HB_SCRIPT_KATAKANA_OR_HIRAGANA

### LINKED ISSUE #1473: Add KATAKANA_OR_HIRAGANA script
Blink currently collates Katakana to Hiragana because both will be mapped to `kana` in OpenType, and reducing the number of script runs, even only for this case, is surprisingly a big win for the performance of shaping Japanese text.

Adding `hrkt` script will help the situation clearer from the interface perspective. Appreciate your considerations.

xfq:
Fix #1473.

### DIFF
--- a/docs/harfbuzz-sections.txt
+++ b/docs/harfbuzz-sections.txt
@@ -142,6 +142,7 @@ hb_script_from_iso15924_tag
 hb_script_to_iso15924_tag
 hb_script_from_string
 hb_script_get_horizontal_direction
+HB_SCRIPT_KATAKANA_OR_HIRAGANA
 hb_language_from_string
 hb_language_to_string
 hb_language_get_default
--- a/docs/harfbuzz-sections.txt
+++ b/docs/harfbuzz-sections.txt
@@ -142,7 +142,6 @@ hb_script_from_iso15924_tag
 hb_script_to_iso15924_tag
 hb_script_from_string
 hb_script_get_horizontal_direction
-HB_SCRIPT_KATAKANA_OR_HIRAGANA
 hb_language_from_string
 hb_language_to_string
 hb_language_get_default
--- a/src/hb-ot-tag.cc
+++ b/src/hb-ot-tag.cc
@@ -43,8 +43,10 @@ hb_ot_old_tag_from_script (hb_script_t script)
     case HB_SCRIPT_INVALID:		return HB_OT_TAG_DEFAULT_SCRIPT;
     case HB_SCRIPT_MATH:		return HB_OT_TAG_MATH_SCRIPT;
 
-    /* KATAKANA and HIRAGANA both map to 'kana' */
+    /* Hiragana, Katakana, and their collective script all map to 'kana'. */
     case HB_SCRIPT_HIRAGANA:		return HB_TAG('k','a','n','a');
+    case HB_SCRIPT_KATAKANA_OR_HIRAGANA:
+					return HB_TAG('k','a','n','a');
 
     /* Spaces at the end are preserved, unlike ISO 15924 */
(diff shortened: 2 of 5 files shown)


behdad:
Thank you. This looks good to me. @jfkthame comments?

behdad (approved):

khaledhosny:
If we do this, should we support other similar [script code aliases](https://www.unicode.org/iso15924/faq.html#:~:text=Questions-,What%20are%20script%20code%20aliases) as well?

behdad:
> If we do this, should we support other similar [script code aliases](https://www.unicode.org/iso15924/faq.html#:~:text=Questions-,What%20are%20script%20code%20aliases) as well?

The only reason the KATAKANA_OR_HIRAGANA is useful is because they both map to the same OpenType script tag. So, by combining them, the runs can mix and match and still perform correct OpenType shaping, and kerning in between. It wouldn't make sense for the other two listed there as it would break OpenType shaping them.

khaledhosny:
OK. So why not add it directly to the hb_script_t enum? It is already an ISO 15924 script code: https://unicode.org/iso15924/iso15924-codes.html

behdad:
> OK. So why not add it directly to the hb_script_t enum? It is already an ISO 15924 script code: https://unicode.org/iso15924/iso15924-codes.html

`hb_script_t` is technically speaking Unicode scripts. I feel a bit hesitant to do it. Because eg. glib wouldn't have an equivalent.

This is more of a hack for systems that want to combine those two scripts into one run. In a way like `HB_SCRIPT_MYANMAR_ZAWGYI`, which we don't bother exposing either. Donno. Maybe we shouldn't expose this one in the public header either.

khaledhosny:
> This is more of a hack for systems that want to combine those two scripts into one run. In a way like `HB_SCRIPT_MYANMAR_ZAWGYI`, which we don't bother exposing either. Donno. Maybe we shouldn't expose this one in the public header either.

That would work for me. Or give it a name that distinguish it from regular `hb_script_t`, say, `HB_SCRIPT_ALIAS_KATAKANA_OR_HIRAGANA`.

behdad:
Let's do without an alias for now and ask implementations to use another method to create one (HB_TAG, etc).

jfkthame:
Seems like there should be a public value that clients can pass to `hb_buffer_set_script` if they wish?

behdad:
> Seems like there should be a public value that clients can pass to `hb_buffer_set_script` if they wish?

Right. To avoid the public API discussion I was suggesting that clients that do this use either HB_TAG() directly, or hb_script_from_iso15924_tag or something.

What's your suggestion?

jfkthame:
> > Seems like there should be a public value that clients can pass to `hb_buffer_set_script` if they wish?
> 
> Right. To avoid the public API discussion I was suggesting that clients that do this use either HB_TAG() directly, or hb_script_from_iso15924_tag or something.
> 
> What's your suggestion?

Personally, I'd lean towards having an explicitly-defined constant, but I guess it's not essential. It does feel a bit odd (IMO) to have `hb_script_from_iso15924_tag` return a value that isn't defined as one of the `hb_script_t` enum constants, but if you want to avoid putting it in the public API, I can live with that.

Your point about `hb_script_t` constants corresponding to Unicode scripts does make sense, though. Hmm, now I'm questioning my inclination in the previous paragraph....

So, seems fine I guess. I don't have any better ideas!

behdad:
There's precedent for Zwagyi. Letting clients that know what they are doing do it.

behdad:
@xfq Can you please remove the public macro? Would that work for your usecase?

PR #5997, opened 2026-05-22
Some CMake updates for Windows builds

fanc999-1:
Hi,

This attempts to fix up CMake builds, in particular for Windows, for the following:

* Fix up building `hb-cpu` on Windows by including the needed `demo-renderer-d3d11.cc` source file, and linking to `d3d11.lib` and `D3DCompiler.lib` (or so on MinGW).
* Fix up introspection builds when ICU is enabled, by grabbing the proper library name for ICU-UC, so that we can feed that into `g-ir-scanner`.

With blessings, thank you!


behdad:
Thanks.

PR #5999, opened 2026-05-25
[rust] Windows build fixes

drott:
Cast and add `#[allow(clippy::unnecessary_cast)]` as the C compiler chooses different signedness for the `hb_script_t` and `hb_tag_t` enum types depending on architecture. Add casts where needed and add clippy suppressions to avoid warnings when compiled on systems where the enum is unsigned by default.

### DIFF
--- a/src/rust/shape.rs
+++ b/src/rust/shape.rs
@@ -131,7 +131,9 @@ pub unsafe extern "C" fn _hb_harfrust_shape_plan_create_rs(
 ) -> *mut c_void {
     let font_data = font_data as *const HBHarfRustFontData;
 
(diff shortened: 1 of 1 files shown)


drott:
It would be best if we could define a strict type for the enums on the header API level, but that is only possible in C++ as far as I can see. `rust-bindgen` then takes the signedness from clang - and makes it platform dependent, `u32` on Linux, `i32` on Linux. Cast where needed and suppress clippy warnings.

behdad:
Thanks. Yeah in C we're limited to platform-specific unfortunately.

PR #6001, opened 2026-05-26
[harfrust] Update shape glue for harfrust 0.8

behdad:
Update the Rust bridge and Cargo dependency to harfrust 0.8. Switch shaping calls to ShapeOptions, keep the cached shaper path, and remove an unused face_data argument from the shaping ABI.

Tested: meson compile -C build
Assisted-by: OpenAI Codex


dfrg (approved):
LGTM

PR #6003, opened 2026-05-26
[mort] Fix contextual offsets

### LINKED ISSUE #5887: Text with hyphens does not shape correctly with MacOS CharcoalCY font
Using `hb_shape` on the MacOS (MacOS Tahoe 26.3.1) CharcoalCY font with text that includes hyphens creates invalid substitutions from the `mort` table, eg:
```
hb-shape "/System/Library/PrivateFrameworks/FontServices.framework/Versions/A/Resources/Fonts/ApplicationSupport/CharcoalCY.dfont" "Empty-handed" -v
```
```
1: (Empty-handed)
1: <U+0045=0|U+006D=1|U+0070=2|U+0074=3|U+0079=4|U+002D=5|U+0068=6|U+0061=7|U+006E=8|U+0064=9|U+0065=10|U+0064=11>
1: [gid1028|gid257|gid257|gid257|gid257|space|gid257@512,0|gid257@512,0|gid257@512,0|gid257@512,0|gid257@512,0|gid257@512,0]
```
The font only has 251 glyphs, so all of the replaced glyph indexes (other than `space`) are not glyphs that are in the font. 

After doing a lot of debugging, I've determined that the `ContextualSubtable::driver_context_t` that is meant to replace `--` with an Em dash is indexing off the end of the substitution table, leading to invalid replacements.

`ftxdumperfuser -t mort` dumps the offending subtable as this:
```
<mortSubtable subtableIndex="3" length="68" coverage="0x00002001" subtableType="1" subFeatureFlags="0x00000008" >
	<contextualSubtable>
		<ft:classTable nClasses="5" classTableOffset="0x0000000A" >
			<class name="c4">
				<glyph glyphRefID="36" />
			</class>
		</ft:classTable>
		<ft:stateTable nStates="3" stateTableOffset="0x00000010" >
			<stateRow index="s0" >
				<stateEntry class="c0" entry="e1" />
				<stateEntry class="c1" entry="e1" />
				<stateEntry class="c2" entry="e1" />
				<stateEntry class="c3" entry="e1" />
				<stateEntry class="c4" entry="e2" />
			</stateRow>
			<stateRow index="s1" >
				<stateEntry class="c0" entry="e1" />
				<stateEntry class="c1" entry="e1" />
				<stateEntry class="c2" entry="e1" />
				<stateEntry class="c3" entry="e1" />
				<stateEntry class="c4" entry="e2" />
			</stateRow>
			<stateRow index="s2" >
				<stateEntry class="c0" entry="e1" />
				<stateEntry class="c1" entry="e1" />
				<stateEntry class="c2" entry="e1" />
				<stateEntry class="c3" entry="e1" />
				<stateEntry class="c4" entry="e3" />
			</stateRow>
		</ft:stateTable>
		<contextualEntryTable nEntries="3" entryTableOffset="0x00000020" >
			<contextualEntry entry="e1" nextState="s0" mark="no" advance="yes" />
			<contextualEntry entry="e2" nextState="s2" mark="yes" advance="yes" />
			<contextualEntry entry="e3" nextState="s0" mark="no" advance="yes" subMarkOffset="131056" subCurrentOffset="131058" />
		</contextualEntryTable>
		<contextualSubstitutionTable glyphCount="2" tableOffset="0x00000038">
			<substitutionGlyph offset="56" glyphRefID="199" />
		</contextualSubstitutionTable>
	</contextualSubtable>
</mortSubtable>
```

I believe there are two issues:
1. The `markOffset` and `currentOffset` fields of the `ContextualSubtable::Entry` structure should be `HBINT16` instead of `HBUINT16` for the `mort` table. The bottom of the `State Tables` section [here](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6Tables.html#StateTables) indicates that the offsets are signed and can be negative. In this particular case, they are negative, but are being interpreted as positive before being applied. Note that `ftxdumperfuser` is also incorrectly reporting these as positive, though the documentation clearly states that they can be negative. The raw values are `0xfff8` for `subMarkOffset` and `0xfff9` for `subCurrentOffset`, which when interpreted as signed give valid indexes into the substitution table.
2. It seems that any glyph beyond the number of glyphs in the class table cause the substitution table to be indexed beyond its limit, which is what seems to be causing the invalid glyph IDs reported above.

I have locally added code to interpret the offsets as signed and not apply a replacement glyph if the index is beyond the length of the substitution table and it fixes the issue without adding regressions in my local test data or in the harfbuzz unit-tests. I believe the desire is to do no runtime bounds-checking during the shaping phase though, so I'm not sure sure if this is the _correct_ solution, but it seems to work.

behdad:
Fixes https://github.com/harfbuzz/harfbuzz/issues/5887

Plus a minor other fix while in there.

### DIFF
--- a/src/hb-aat-layout-common.hh
+++ b/src/hb-aat-layout-common.hh
@@ -228,6 +228,11 @@ struct hb_aat_apply_context_t :
   void replace_glyph_inplace (unsigned i, hb_codepoint_t glyph)
   {
     buffer->info[i].codepoint = glyph;
+    if (glyph == DELETED_GLYPH)
+    {
+      buffer->scratch_flags |= HB_BUFFER_SCRATCH_FLAG_AAT_HAS_DELETED;
+      _hb_glyph_info_set_aat_deleted (&buffer->info[i]);
+    }
     if (likely (using_buffer_glyph_set))
       buffer_glyph_set->add (glyph);
 #ifndef HB_NO_OT_LAYOUT
--- a/src/hb-aat-layout-common.hh
+++ b/src/hb-aat-layout-common.hh
@@ -1139,29 +1139,35 @@ struct ObsoleteTypes
   typedef ClassTable<HBUINT16> ClassTypeWide;
 
   template <typename T>
-  static int offsetToIndex (int offset,
+  static int offsetToIndex (int64_t offset,
 			    const void *base,
 			    const T *array)
   {
+    int64_t array_offset = (const char *) array - (const char *) base;
+    int bad_index = INT_MAX / T::static_size;
+
     /* https://github.com/harfbuzz/harfbuzz/issues/3483 */
     /* If offset is less than base, return an offset that would
      * result in an address half a 32bit address-space away,
      * to make sure sanitize fails even on 32bit builds. */
-    if (unlikely (offset < int ((const char *) array - (const char *) base)))
-      return INT_MAX / T::static_size;
+    if (unlikely (offset < array_offset))
+      return bad_index;
 
     /* https://github.com/harfbuzz/harfbuzz/issues/2816 */
-    return (offset - int ((const char *) array - (const char *) base)) / T::static_size;
+    int64_t index = (offset - array_offset) / T::static_size;
+    if (unlikely (index > bad_index))
+      return bad_index;
+    return index;
   }
   template <typename T>
-  static int byteOffsetToIndex (int offset,
+  static int byteOffsetToIndex (int64_t offset,
 				const void *base,
 				const T *array)
   {
     return offsetToIndex (offset, base, array);
   }
   template <typename T>
-  static int wordOffsetToIndex (int offset,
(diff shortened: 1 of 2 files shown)


src/hb-aat-layout-common.hh
```diff
@@ -1134,31 +1139,31 @@ struct ObsoleteTypes
...
+  static int offsetToIndex (int offset,
+			    const void *base,
+			    const T *array)
   {
     /* https://github.com/harfbuzz/harfbuzz/issues/3483 */
     /* If offset is less than base, return an offset that would
      * result in an address half a 32bit address-space away,
      * to make sure sanitize fails even on 32bit builds. */
-    if (unlikely (offset < unsigned ((const char *) array - (const char *) base)))
+    if (unlikely (offset < int ((const char *) array - (const char *) base)))
```
behdad: This needs careful reviewing.

PR #6011, opened 2026-05-30
[rust] Use new harfrust GlyphFlags getter

behdad:
Faster.


dfrg (approved):
Always fun to delete code. LGTM!

PR #6017, opened 2026-06-02
Generate Rust language tags as bytes

behdad:
For https://github.com/harfbuzz/harfrust/pull/380


src/gen-tag-table.py
```diff
@@ -922,7 +922,9 @@ def get_matching_language_name_rust (intersection, candidates):
 			continue
 		commented_out = len (language) == 3 and len (tags) == 1 and language == tags[0].lower ()
 		for tag in tags:
-			print ('%sLangTag { language: "%s", \ttag: %s },' % ('//  ' if commented_out else '    ', language, hb_tag_rust (tag)), end='')
+			language_bytes = (language.encode ('ascii') + b'\0' * 4)[:4]
+			language_bytes = ''.join (chr (b) if 32 <= b < 127 and chr (b) not in '\\"' else f'\\x{b:02x}' for b in language_bytes)
```
drott: As mentioned in the HarfRust side change. Can we avoid the trailing `\x00` with a Rust macro, for example? They look a bit noisy in the generated source.

drott (approved):

PR #6020, opened 2026-06-05
[util] Fix Unicode command line arguments on Windows

moi15moi:
GLib g_option_context_parse() converts argv strings from the system ACP via g_locale_to_utf8(), but Windows cmd deliver arguments encoded in the system codepage (ex: CP1252), which cannot represent most Unicode codepoints. This caused non ascii character to be corrupted

For example, for this command, it rendered garbage instead of the emoji:
```
hb-view.exe "C:\Windows\Fonts\seguiemj.ttf" "😋" --output-file=output.png
```

Fix this in all util binaries (hb-ft-raster, hb-info, hb-raster, hb-shape, hb-subset, hb-vector, hb-view) by:
1. Calling g_win32_get_command_line() to obtain correctly UTF-8-encoded arguments from GetCommandLineW().
2. Parsing with g_option_context_parse_strv() which skips the g_locale_to_utf8() conversion for G_OPTION_ARG_STRING entries.
3. Adding G_OPTION_FLAG_FILENAME to the G_OPTION_REMAINING callback entry in main-font-text.hh, so the callback path also skips the locale conversion in strv_mode (see https://gitlab.gnome.org/GNOME/glib/-/blob/8ef2f3ad402daf9fb94615ddef63ceef978fd76f/glib/goption.c#L1305).

### DIFF
--- a/util/hb-ft-raster.cc
+++ b/util/hb-ft-raster.cc
@@ -332,5 +332,12 @@ int
 main (int argc, char **argv)
 {
   using main_t = main_font_text_t<shape_consumer_t<ft_raster_output_t>, font_options_t, shape_text_options_t>;
+#ifdef _WIN32
+  gchar **argv_utf8 = g_win32_get_command_line ();
+  int ret = batch_main<main_t, true> ((int) g_strv_length (argv_utf8), argv_utf8);
+  g_strfreev (argv_utf8);
+  return ret;
+#else
   return batch_main<main_t, true> (argc, argv);
+#endif
 }
--- a/util/hb-info.cc
+++ b/util/hb-info.cc
@@ -1437,5 +1437,12 @@ retry:
 int
 main (int argc, char **argv)
 {
+#ifdef _WIN32
+  gchar **argv_utf8 = g_win32_get_command_line ();
+  int ret = batch_main<info_t> ((int) g_strv_length (argv_utf8), argv_utf8);
+  g_strfreev (argv_utf8);
+  return ret;
+#else
   return batch_main<info_t> (argc, argv);
+#endif
 }
--- a/util/hb-raster.cc
+++ b/util/hb-raster.cc
@@ -38,5 +38,12 @@ int
 main (int argc, char **argv)
 {
   using main_t = main_font_text_t<shape_consumer_t<raster_output_t>, font_options_t, shape_text_options_t>;
+#ifdef _WIN32
+  gchar **argv_utf8 = g_win32_get_command_line ();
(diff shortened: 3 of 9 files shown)


parasol-aser:
**`util/options.hh:274` (`option_parser_t::parse`, Windows branch)** — switching from the count-based `g_option_context_parse()` to `g_strdupv(*argv)` + `g_option_context_parse_strv()` makes the parser rely on the argv array being NULL-terminated, but two callers pass arrays sized by `argc` with no terminator: `hb-subset.cc::parse_face` (`g_memdup2(argv, argc * sizeof *)`, hit on every normal `hb-subset.exe in.ttf out.ttf`) and `batch.hh`'s `args[64]` (filled to `argc`, never terminated, hit via `--batch` on all tools). On Windows, `g_strdupv` scans past the last real entry — a heap OOB read in the subset case and an uninitialized-stack read in batch, so a crash (or stray bytes injected as extra argv). It's local-only (operator-supplied argv), so the practical impact is a Windows crash/UB rather than anything remote, but these paths were memory-safe before this change.

Fix: NULL-terminate before the strv API — in `batch.hh` reserve a slot and set `args[argc] = nullptr` after the loop, and in `parse_face` allocate `argc + 1` pointers with a trailing `NULL` (or keep a count-based parse on the Windows branch).

For `batch.hh`:

```suggestion
      char *args[64];
      argc = 0;
      args[argc++] = argv[0];
      char *p = buf, *e;
      args[argc++] = p;
      while ((e = strchr (p, ';')) && argc < (int) ARRAY_LENGTH (args) - 1)
      {
	*e++ = '\0';
	while (*e == ';')
	  e++;
	args[argc++] = p = e;
      }
      args[argc] = nullptr;
```

behdad:
Thanks. Is this ready, or needs updates as per the latest comment?

Can the duplicated logic move to a function and called?

Marking REMAINING as FILENAME wouldn't interfere with text argument later?

moi15moi:
I will need to fix the code as @parasol-aser said.
I just tested the hb-view and i didnt catch the 2 problems he said.

moi15moi:
> Can the duplicated logic move to a function and called?

I'm not sure how I can do that since the batch_main call is different for all files.

> Marking REMAINING as FILENAME wouldn't interfere with text argument later?

No, i dont see how it could interfere in glib source code.

behdad:
I clean this up in https://github.com/harfbuzz/harfbuzz/pull/6022
Please check & test.
I'm also building on wine to test.
Thanks.

moi15moi:
Yes, your fix does also work

PR #6022, opened 2026-06-07
Follow up to PR 6020 [util] Fix Unicode command line arguments on Windows

behdad:


behdad:
Fix works on mingw/wine. Thanks!

PR #6029, opened 2026-06-11
[cmake]: Weak assignment of the C++ standard

UVV-gh:
Some projects may need higher C++ standard than C++11. In order not to force the requirement for the whole project, the C++ standard can be set at the configure time with -DCMAKE_CXX_STANDARD=20 for example. The project default takes precedence only when there's no assignment provided in the command line by setting the default cache value.

There's no need to additionally check for the compiler support when CMAKE_CXX_STANDARD_REQUIRED is already set.

This is a follow-up from [this PR](https://github.com/harfbuzz/harfbuzz/pull/5344)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,9 @@ if(POLICY CMP0063)
   cmake_policy(SET CMP0063 NEW)
 endif()
 
-set(CMAKE_CXX_STANDARD 11)
+if(NOT DEFINED CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 11)
+endif()
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 ## Disallow in-source builds, as CMake generated make files can collide with autotools ones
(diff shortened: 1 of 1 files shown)


behdad:
Thanks. Our cmake build is community-maintained as we developers don't know it. But here's what codex has to say about your change:

  Main issue: `set(CMAKE_CXX_STANDARD 11 CACHE STRING "C++ standard")` creates the cache entry early. The later ICU>=75.1 change:
```
  set(CMAKE_CXX_STANDARD 17 CACHE STRING "C++ standard")
```
  will not overwrite that existing cache entry without FORCE, so CMake builds with ICU 75.1+ will stay at C++11. That regresses the current
  behavior, where set(CMAKE_CXX_STANDARD 17) at CMakeLists.txt:354 does affect all later targets because targets are created after that point.

  The intent is good: user-provided -DCMAKE_CXX_STANDARD=20 should not be clobbered, and removing the manual -std=c++11 flag is reasonable
  because CMAKE_CXX_STANDARD_REQUIRED handles compiler support. But the ICU minimum needs to remain effective.

  Suggested shape:
```
  if (NOT DEFINED CMAKE_CXX_STANDARD)
    set(CMAKE_CXX_STANDARD 11)
  endif()
  set(CMAKE_CXX_STANDARD_REQUIRED ON)

  ...

  if (ICU_VERSION VERSION_GREATER_EQUAL 75.1)
    if (NOT CMAKE_CXX_STANDARD OR CMAKE_CXX_STANDARD LESS 17)
      set(CMAKE_CXX_STANDARD 17)
    endif ()
  endif ()
```
  That preserves explicit higher standards like 20, while still enforcing the ICU 75.1+ minimum. CI is partly still running; completed checks
  I saw were mostly green, but this CMake/ICU cache case is not covered by the visible C++ versions workflow.

UVV-gh:
Thanks for the suggestion.

Issue #4850, opened 2024-09-03
ligatures with hangul does not working correctly


cloudchamb3r:
I'm using [d2coding ligature font](https://github.com/user-attachments/files/16843174/d2codingliga.ttf.zip)
everything works well, except for hangul ligature. 

i found by enabling calt feature(disabled in `hb-ot-shaper-hangul.cc`) this issue can be resolved, but also founded enabling calt features makes some rendering issue on some old hangul characters, such as U+11A2.

![liga ko](https://github.com/user-attachments/assets/05c3ae12-84db-4ba0-ac2d-fcd5b648ab85)
![liga jp](https://github.com/user-attachments/assets/540561f1-8445-4bca-a558-fa7115b009e1)
![liga en](https://github.com/user-attachments/assets/360bf121-cac4-4b02-9b82-cc74ffc03f0b)
![liga cn](https://github.com/user-attachments/assets/0b63dab1-e74a-44fe-85ea-9de91ed63b31)

PR #6030, opened 2026-06-12
Apply 'calt' feature to non-Hangul characters in Hangul runs

j0j1j2:
Follow-up to #4850 (now discussion #4853).

The Hangul shaper disables `calt` for the whole run since ac53443f1 (2014): Uniscribe doesn't apply it for Hangul, and fonts at the time (Noto Sans CJK / Source Han Sans) had their jamo lookups in `calt`. The side effect is that Latin/Common characters itemized into a Hangul run lose `calt` too, so programming ligatures (D2Coding, JetBrains Mono, ...) don't work in mixed Hangul/Latin text.

In the discussion behdad asked whether those fonts have been fixed since. They have: current Noto Sans CJK KR (2.004) keeps its jamo lookups in `ljmo`/`vjmo`/`tjmo`, and its `calt` only contains a contextual spacing rule for spaces between syllables which the blanket disable is suppressing as well.

This PR keeps `calt` off for Hangul characters (old jamo-in-calt fonts stay safe, and Hangul itself still matches Uniscribe/DWrite), but re-adds it as a masked feature enabled on the non-Hangul characters of the run. Explicit `calt=0` / `calt=1` from the user now also takes precedence; previously it was silently ignored.

JetBrains Mono, `가=>`:

```
before: [.notdef=0+600|equal=1+600|greater=2+600]
after:  [.notdef=0+600|SPC=1+600|equal_greater.liga=2+600]
```

The existing hangul-jamo tests (whose fonts have the old jamo-in-calt structure) pass unchanged. Added hangul-calt tests with a small test font covering the masked behavior and the user-feature overrides.


src/hb-ot-shaper-hangul.cc
```diff
@@ -42,34 +42,56 @@ enum {
...
 };
 
 static const hb_tag_t hangul_features[HANGUL_FEATURE_COUNT] =
 {
   HB_TAG_NONE,
   HB_TAG('l','j','m','o'),
   HB_TAG('v','j','m','o'),
-  HB_TAG('t','j','m','o')
+  HB_TAG('t','j','m','o'),
+  HB_TAG('c','a','l','t')
```
behdad: No need to add this to the group. You can just define a tag for it and use it directly I think. No?

src/hb-ot-shaper-hangul.cc
```diff
@@ -42,34 +42,56 @@ enum {
...
 
-  for (unsigned int i = FIRST_HANGUL_FEATURE; i < HANGUL_FEATURE_COUNT; i++)
+  for (unsigned int i = FIRST_HANGUL_FEATURE; i <= LAST_HANGUL_JAMO_FEATURE; i++)
     map->add_feature (hangul_features[i]);
 }
 
 static void
 override_features_hangul (hb_ot_shape_planner_t *plan)
 {
+  /* If the user requested 'calt' globally, respect that and leave the
```
behdad: Don't worry about this. The generic shaper overrides shaper features with user features.

src/hb-ot-shaper-hangul.cc
```diff
@@ -42,34 +42,56 @@ enum {
...
+
   /* Uniscribe does not apply 'calt' for Hangul, and certain fonts
    * (Noto Sans CJK, Source Sans Han, etc) apply all of jamo lookups
-   * in calt, which is not desirable. */
+   * in calt, which is not desirable.
+   *
+   * Don't kill it for the entire run though; re-add it as a masked
+   * feature defaulting to off, enabled only on non-Hangul characters
+   * (marked in preprocess_text_hangul), so that jamo stay out of reach
+   * of such lookups while embedded Latin/Common characters keep
```
behdad: Latin shouldn't ever make it to this shaper.

src/hb-ot-shaper-hangul.cc
```diff
@@ -118,6 +140,23 @@ data_destroy_hangul (void *data)
 
 #define isHangulTone(u) (hb_in_range<hb_codepoint_t> ((u), 0x302Eu, 0x302Fu))
 
+/* Approximation of Script=Hangul, for deciding which characters keep
```
behdad: I don't like this function. I suggest either just turning it off for Jamo. Or something based on the General_Category. Either works for me.

behdad (changes requested):
Thanks. Generally looking good. Added a few comments.

src/hb-ot-shaper-hangul.cc
j0j1j2: Ok, I Dropped it from hangul_features. Instead of that calt gets its own mask now.

src/hb-ot-shaper-hangul.cc
j0j1j2: Removed. calt=0/1 still work through the generic path.

src/hb-ot-shaper-hangul.cc
j0j1j2: I Fixed the wording. Thank you.

src/hb-ot-shaper-hangul.cc
j0j1j2: Switched to clearing the mask on jamo only, function's gone. Only covers conjoining jamo though, not compat jamo (U+3130~318F) fine to leave those? They don't hit the jamo lookups.

j0j1j2:
@behdad  Thanks, addressed all four and repushed. it's just the hangul shaper now. One thing I'd like you to double-check: I only exclude conjoining jamo, not compat jamo. Could you confirm that's the behavior you want?

behdad:
This looks good now. Thanks. @jfkthame ?

jfkthame:
> This looks good now. Thanks. @jfkthame ?

LGTM. (I didn't test it personally, but I see there's test coverage, and the idea makes sense.)

PR #6032, opened 2026-06-15
[subset] fix bug in math table

### LINKED ISSUE #5998: Subsetting STIXTwoMath with U+21D4 fails due to MATH table referencing absent glyphs
Subsetting STIXTwoMath-Regular.ttf with U+21D4 (⇔) fails:

```
hb-subset --unicodes=U+21D4 -o /dev/null test/subset/data/fonts/STIXTwoMath-Regular.ttf
# -> error: Operation failed.
```

Dropping MATH succeeds, confirming the MATH table is the problem:

```
hb-subset --unicodes=U+21D4 --drop-tables=MATH -o /dev/null test/subset/data/fonts/STIXTwoMath-Regular.ttf
# -> succeeds
```

The cause: MATH closure adds assembly parts of ⇔ to the retained set, including ⇐ (uni21D0) and ⇒ (uni21D2). The MATH subsetter then copies these glyphs' own constructions, which reference further glyphs (uni21D0.endr, uni21D0.endl) that aren't in the subset. The serializer fails on the unmappable glyph IDs.

Math engines don't recursively stretch assembly parts, so the referenced glyphs aren't functionally needed. The fix should strip or skip MATH constructions for glyphs that were only added via MATH closure, not from the input set. This would match fonttools subset behavior.

See #5992 for prior discussion.

qxliu76:
1. do Math closure before gsub closure, this also matches fonttools
2. don't keep constructions for glyphs that are only added during math closure, since we don't do recursive closure over math table.

This fixes #5998, now the result matches fonttools'


khaledhosny (commented):
Thanks! Can you split it into two commits, one per fix? (the test can be added to the last commit). LGTM otherwise.

qxliu76:
Done.

khaledhosny (approved):

qxliu76:
There's another diff in the coverage table that fonttools always writes a Coverage table even it's empty, HB just set coverage offset to 0. I'm not sure if we want to do that in HB.

behdad:
> There's another diff in the coverage table that fonttools always writes a Coverage table even it's empty, HB just set coverage offset to 0. I'm not sure if we want to do that in HB.

FT also used to write 0, but @anthrotype changed it.

The spec says writers should always write non-zero offset, but readers should always accept 0 offset. Which means, 0 is safe to use, and indeed for 8 years we've used it without any problem. I'm inclined to continue doing that.

anthrotype:
FWIW the spec actually uses the words "must" for the compiler's side and "should" and "non-conformant font-data" for the parser's:

> For cases in which subtable offset fields are not documented as permitting NULL values, font compilers must include a subtable of the indicated format, even if it is a header stub without further data (for example, a coverage table with no glyph IDs). Applications parsing font data should, however, anticipate non-conformant font data that has a NULL subtable offset where only a non-NULL value is expected.

https://learn.microsoft.com/en-us/typography/opentype/spec/otff#data-types

But I'm not gonna die on a 4-bytes hill 😄

I checked OTS and for MATH in particular it only rejects a table containing NULL-offset coverage when glyph count != 0, which isn't the case for your hb-subset font.

behdad:
> But I'm not gonna die on a 4-bytes hill 😄

Same here. I'll defer to subset owners @garretrieger and @qxliu76.

garretrieger:
> > But I'm not gonna die on a 4-bytes hill 😄
> 
> Same here. I'll defer to subset owners @garretrieger and @qxliu76.

For me, it's a mild preference to match fonttools behaviour as that makes the fonttools comparison tests easier.

qxliu76:
I'll work on the coverage diff, but that will be in a different PR. Is this PR OK to merge now?

behdad:
Lg

PR #6051, opened 2026-07-06
[graphite2] Fix cluster values when the buffer was reversed

mimbrown:
## Summary

When shaping with the `graphite2` shaper and the buffer's direction differs from the script's natural horizontal direction, cluster values can violate the documented cluster contract: a merged cluster group takes the value of its **visually first** character instead of the **minimum** of the characters it covers. Some input cluster values are then entirely absent from the output, so consumers that tile text by cluster boundaries (e.g. Skia's skparagraph) see input positions covered by no cluster and assert/crash.

This is not theoretical: Arabic-Indic digits and the Arabic subtending marks U+0600–U+0603 are bidi class AN, which always resolves to an even (LTR-shaped) embedding level — so any bidi-itemizing layout engine shapes these runs with `direction=ltr` while the script direction is RTL, which is exactly the configuration that triggers the internal buffer reversal. We hit this in production (Flutter engine built with graphite2, Awami Nastaliq, plain `U+0600` + digits text).

## Reproduction (before this fix)

```
$ hb-shape --shaper=graphite2 --direction=ltr --unicodes "0600,0661" AwamiNastaliq-Regular.ttf
[absOne.small=1@331,450+1600|absNumberSign=1@-1600,0+0]
```

Both glyphs have cluster **1**; no glyph carries cluster 0, so the character at index 0 (U+0600) belongs to no cluster group. Expected per the [cluster contract](https://harfbuzz.github.io/working-with-harfbuzz-clusters.html): the merged group covers characters {0, 1}, so its value must be 0. All of U+0600–U+0603 followed by digits reproduce this with Awami Nastaliq; `--cluster-level=1` behaves the same. The same text via the `ot` shaper (Scheherazade New), and the same font/text in native RTL, are both correct.

## Analysis

When `direction` differs from the script's horizontal direction, `_hb_graphite2_shape` calls `hb_buffer_reverse_clusters (buffer)` up front, after which original cluster values **decrease** along `buffer->info`. The cluster-building loop then assigns each group's output value solely from its base (first) char (`clusters[0].cluster = buffer->info[0].cluster` and `c->cluster = buffer->info[c->base_char].cluster`), and never recomputes it when the group later grows (`num_chars` extension from `gr_slot_after`, or the merge-down loop). For a non-reversed buffer the base char holds the minimum, so this is correct; for a reversed buffer the minimum lives at the *end* of the span and is never picked up.

The fix computes each group's value as the minimum of its endpoint values once the groups are final — cluster values are monotone along the buffer in either case, so the minimum is always at an end. This runs before the loop that rewrites `buffer->info`, while original cluster values are still readable.

## Validation

- Differential fuzz against the unfixed code (~7,500 random combos of subtending marks, Arabic-Indic/European digits, Arabic letters, marks, controls; both directions): the old code violates cluster invariants (uncovered run start, out-of-range) in 237 cases, the fixed code in **0**. Every changed output is in the reversed-buffer path only (`rtl` outputs are byte-identical), and no cluster value ever increased — changes are exactly the min-merge corrections, including mid-run groups (e.g. `[0,1,3,3]` → `[0,1,2,2]` where characters {2,3} merged).
- `meson test --suite=shape` fully green (6,156 baseline subtests unaffected; positions unchanged everywhere).
- The added test fails on current `main` (clusters `[1,1]`) and passes with the fix; the third test line pins the already-correct native-RTL behavior. The tests carry `--shaper=graphite2` in options, so `run-tests.py` skips them cleanly on builds without graphite2.

## Test font

The suite previously had no graphite2 shape tests. The test font is a subset of [Awami Nastaliq](https://software.sil.org/awami/) (SIL, OFL 1.1) covering U+0600–U+0603 and the Arabic-Indic digits, with the graphite tables (`Silf`/`Glat`/`Gloc`/`Feat`/`Sill`) passed through unmodified and GIDs retained. It is renamed ("Nastaliq Cluster Test") because "Awami" and "SIL" are Reserved Font Names under the OFL; copyright and license name records are preserved. Happy to adjust or drop the test if you'd rather handle graphite fixtures differently.


mimbrown:
Context on this PR: my team and I ran into this issue real-world with the `U+0600`–`U+0603` marks, using Awami Nastaliq. Originally we thought this was a skia issue, and had a local skia patch which added some defensive code, which fixed the issue we were facing. But digging into it with Fable 5 turned up this analysis. As Claude said, if not having graphite tests is intentional than I can remove the tests. I can't vouch that this fix is the right one; but I can vouch that the bug is real.

behdad:
Thanks. I haven't fully studied your report. But if anything, the code should be gated on cluster-value being monotone.

mimbrown:
Good point, thanks — done in a6b0f1f. The fixup is now gated on `HB_BUFFER_CLUSTER_LEVEL_IS_MONOTONE (buffer->cluster_level)`, following the precedent of the coretext shaper's out-of-order cluster fixup; at the non-monotone levels the historic behavior (base_char's cluster value) is preserved.

I went with the level check rather than scanning the actual values, since that's what `merge_clusters` / buffer-verify / coretext all key on, and a value scan would min-rewrite a `CHARACTERS`-level buffer whose values happen to be ordered. Happy to change to a literal monotonicity scan if that's what you had in mind.

Verified on the repro case (`U+0600,U+0661`, `--direction=ltr`, graphite2 shaper): levels 0/1 now give clusters `[0,0]`, levels 2/3 keep `[1,1]` as before; full shape suite still green.

src/hb-graphite2.cc
```diff
@@ -383,6 +383,19 @@ _hb_graphite2_shape (hb_shape_plan_t    *shape_plan HB_UNUSED,
...
+   * of the cluster values of the characters it covers, and those values
+   * are monotone along the buffer (decreasing if the buffer was reversed
+   * above), so the minimum is at one end of the covered range.  base_char
+   * alone is wrong when the buffer was reversed and the cluster covers
+   * more than one character.  At non-monotone cluster levels keep the
+   * historic behavior of using base_char's value. */
+  if (HB_BUFFER_CLUSTER_LEVEL_IS_MONOTONE (buffer->cluster_level))
+    for (unsigned int i = 0; i < ci; ++i)
+      if (clusters[i].num_chars)
+	clusters[i].cluster = hb_min (buffer->info[clusters[i].base_char].cluster,
```
behdad, line 396: Can't you just call `merge_clusters` with the correct range here?

behdad: @mimbrown

mimbrown: @behdad apologies for the delay. Agreed that's cleaner; but Claude is wanting to also add a second loop to re-copy the clusters into the scratch array, based on the fact that `merge_clusters` can affect things out-of-order: "a later group's merge can extend backward and lower an earlier group's characters."

```diff
-  if (HB_BUFFER_CLUSTER_LEVEL_IS_MONOTONE (buffer->cluster_level))
-    for (unsigned int i = 0; i < ci; ++i)
-      if (clusters[i].num_chars)
-     clusters[i].cluster = hb_min (buffer->info[clusters[i].base_char].cluster,
-                                   buffer->info[clusters[i].base_char + clusters[i].num_chars - 1].cluster);
+  for (unsigned int i = 0; i < ci; ++i)
+    if (clusters[i].num_chars)
+      buffer->merge_clusters (clusters[i].base_char,
+                           clusters[i].base_char + clusters[i].num_chars);
+  for (unsigned int i = 0; i < ci; ++i)
+    if (clusters[i].num_chars)
+      clusters[i].cluster = buffer->info[clusters[i].base_char].cluster;
```

I am unable to prove this second loop is actually necessary so wanted your input before we go this way. If you're fine with this I'll go with it.

behdad: The effect that Claude doesn't like, is actually the intended behavior of `merge_clusters`. Just leave it as that. No second loop.

mimbrown: Ok just wanting to make sure we're understanding each other: I know that's the intended behavior of `merge_clusters`. Claude is wanting to make sure the *locally scoped* scratch `clusters` array is re-synced with the `buffer`'s clusters after `merge_clusters` has done its operations across the buffer. Is that necessary or not?

behdad: Oh. What are `clusters` later used for? Lemme check.

behdad: I think you actually need to move the whole block to after we set the buffer clusters. Only then we should merge.

behdad: @mimbrown Let me know what you think. Thanks.

behdad: Does this make sense to you:
https://github.com/harfbuzz/harfbuzz/pull/6156

NBruderman: Any news with this? @mimbrown

mimbrown: @behdad @NBruderman again apologies for the delay; some personal life stuff is making me slow to respond. I think let's close this in favor of #6156. That still fixes the original issue; Claude had some concerns, but we followed through to make sure they were unreachable.

behdad: So you think https://github.com/harfbuzz/harfbuzz/pull/6156 is a good fix? I closed it because I wasn't sure.

behdad: > So you think #6156 is a good fix? I closed it because I wasn't sure.

Okay, codex thinks that one is not enough either, but going to finish it there.

behdad:
Continued in https://github.com/harfbuzz/harfbuzz/pull/6156

PR #6056, opened 2026-07-08
Fix hb-vector rendering of COLRv0 layers with huge font scales

khaledhosny:
When painting a COLRv0 glyph at large scale, The output showed the glyph clipped. This is because COLRv0 (and outline fallback) layers were painted with a sequence of push_clip_glyph + unbounded color + pop_clip. For the unbounded clipping a huge 2*32k sized rect was used. With a huge font scale, this rect was not enough and results in clipped glyphs. COLRv1 is not affected because clipping is done in font units (font transform push/pop).

To avoid the clipping issue entirely (and allow clients to optimize this common paint operation), a new fill_glyph callback is introduced and used in harfbuzz-vector. The default implementation uses the decomposed sequence, so existing clients are unaffected.


khaledhosny:
The test failure was a latent bug, fixed in https://github.com/harfbuzz/harfbuzz/pull/6057

behdad:
Okay, I don't have to like this but the approach is solid. Thanks for working on it.

khaledhosny:
> Okay, I don't have to like this but the approach is solid. Thanks for working on it.

I think I never liked the push_clip_glyph + unbounded color + pop_clip approach to v0 layers (I remember commenting about it when the paint API was being implemented; it felt wasteful to me), so we probably will have to agree to disagree on this one.

behdad:
> > Okay, I don't have to like this but the approach is solid. Thanks for working on it.
> 
> I think I never liked the push_clip_glyph + unbounded color + pop_clip approach to v0 layers (I remember commenting about it when the paint API was being implemented; it felt wasteful to me), so we probably will have to agree to disagree on this one.

Skia actually detects this pattern on COLRv1 graphs and uses fill-glyph as well. Bonus points for implementing that!

behdad:
can we in our own renderer, detect if a clip_glyph's node's child is a paint_solid, then emit a fill_glyph for it?

khaledhosny:
> can we in our own renderer, detect if a clip_glyph's node's child is a paint_solid, then emit a fill_glyph for it?

I’ll give it a try.

khaledhosny:
> > can we in our own renderer, detect if a clip_glyph's node's child is a paint_solid, then emit a fill_glyph for it?
> 
> I’ll give it a try.

Done. PTAL.

src/hb-paint.h
```diff
@@ -157,6 +157,29 @@ typedef hb_bool_t (*hb_paint_color_glyph_func_t) (hb_paint_funcs_t *funcs,
...
+ * @funcs: paint functions object
+ * @paint_data: The data accompanying the paint functions in hb_font_paint_glyph()
+ * @glyph: the glyph ID
+ * @font: the font
+ * @is_foreground: whether the color is the foreground
+ * @color: The color to use, unpremultiplied
+ * @user_data: User data pointer passed to hb_paint_funcs_set_fill_glyph_func()
+ *
+ * A virtual method for the #hb_paint_funcs_t to fill a glyph's
+ * shape with a solid color.
```
behdad: Should we talk about the fallback behavior here?

behdad (approved):
Minor docs request. Looks great otherwise. Thanks.

khaledhosny:
Is it good now?

behdad:
Yes. Thanks.

Issue #6065, opened 2026-07-14
Consider adding API to hb_subset_input that converts an input to an equivalent hb-subset command line


garretrieger:
Would be useful for producing command line reproductions of cases using the hb-subset api.

- Should try to produce a compact command line, for example use ranges where possible in the sets.
- API will likely need  to take a preallocated buffer to fill with the output, or could allocate and return a hb_blob_t.

behdad:
Sounds good to me. Is a bit of a bummer because `hb-subset` cmdline tool's syntax is not part of the library. But I'm fine with it.

PR #6071, opened 2026-07-14
Add new hb_subset_input_to_command_line_or_fail().

garretrieger:
This generates an equivalent hb-subset command line string for a given subset input. For #6065 

Assisted by gemini.

### DIFF
--- a/src/hb-subset-input.cc
+++ b/src/hb-subset-input.cc
@@ -744,9 +744,8 @@ _append_arg (hb_vector_t<char> &sb, const char *arg)
   unsigned len = strlen (arg);
   sb.extend (hb_bytes_t (arg, len));
 }
-
 static void
-_format_number_set (const hb_set_t *set, bool is_hex, hb_vector_t<char> &out)
+_format_number_set (const hb_set_t *set, hb_vector_t<char> &out, unsigned base = 10)
 {
   hb_codepoint_t first = HB_SET_VALUE_INVALID, last = HB_SET_VALUE_INVALID;
   bool first_range = true;
@@ -760,14 +759,14 @@ _format_number_set (const hb_set_t *set, bool is_hex, hb_vector_t<char> &out)
     int len;
     if (first == last)
     {
-      if (is_hex)
+      if (base == 16)
         len = snprintf (buf, sizeof (buf), "%X", first);
       else
         len = snprintf (buf, sizeof (buf), "%u", first);
     }
     else
     {
-      if (is_hex)
+      if (base == 16)
         len = snprintf (buf, sizeof (buf), "%X-%X", first, last);
       else
         len = snprintf (buf, sizeof (buf), "%u-%u", first, last);
@@ -837,13 +836,16 @@ _format_set_arg (const hb_set_t *set,
                  Formatter &&formatter,
                  hb_vector_t<char> &sb)
 {
+  char opt_prefix[64];
+  snprintf (opt_prefix, sizeof (opt_prefix), "--%s", opt_name);
+
   if (hb_set_is_inverted (set))
   {
     hb_set_t *excluded = hb_set_copy (set);
     hb_set_invert (excluded);
 
     hb_vector_t<char> opt;
-    opt.extend (hb_bytes_t (opt_name, strlen (opt_name)));
+    opt.extend (hb_bytes_t (opt_prefix, strlen (opt_prefix)));
     opt.extend (hb_bytes_t ("=*", 2));
     opt.push ('\0');
     if (opt.arrayZ) _append_arg (sb, opt.arrayZ);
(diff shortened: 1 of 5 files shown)


src/hb-subset-input.cc
```diff
@@ -734,6 +734,343 @@ hb_subset_axis_range_to_string (hb_subset_input_t *input,
...
+  }
+  return true;
+}
+
+/**
+ * hb_subset_input_to_string_or_fail:
+ * @input: a #hb_subset_input_t object.
+ *
+ * Produces a command line string representation of the given subset input
+ * suitable for use with the `util/hb-subset` command line tool.
```
behdad: Remove `util/`.

src/hb-subset-input.cc
```diff
@@ -734,6 +734,343 @@ hb_subset_axis_range_to_string (hb_subset_input_t *input,
...
+hb_subset_input_to_string_or_fail (hb_subset_input_t *input)
+{
+  if (unlikely (!input || input->in_error ()))
+    return nullptr;
+
+  hb_vector_t<char> sb;
+
+  struct flag_option_t {
+    hb_subset_flags_t flag;
+    const char *option;
```
behdad: Use a fixed size buffer: `const option[32]` or something long enough.

src/hb-subset-input.cc
```diff
@@ -734,6 +734,343 @@ hb_subset_axis_range_to_string (hb_subset_input_t *input,
...
+  };
+  static const flag_option_t flag_options[] = {
+    { HB_SUBSET_FLAGS_NO_HINTING, "--no-hinting" },
```
behdad: You don't need all the repeated `--` here, no?

src/hb-subset-input.cc
```diff
@@ -734,6 +734,343 @@ hb_subset_axis_range_to_string (hb_subset_input_t *input,
...
+    {
+      _append_arg (sb, fo.option);
+    }
+  }
+
+  hb_subset_input_t *default_input = hb_subset_input_create_or_fail ();
+  if (unlikely (!default_input))
+    return nullptr;
+
+  // Serialize all set flags if they are not equal to their default setting:
```
behdad: "flags" -> "sets"?

src/hb-subset-input.cc
```diff
@@ -734,6 +734,343 @@ hb_subset_axis_range_to_string (hb_subset_input_t *input,
...
+  if (!_is_set_default (input->sets.unicodes, default_input->sets.unicodes))
+    _format_number_set_arg (input->sets.unicodes, "--unicodes", true, sb);
```
behdad: Again, the repeated `--` can move into the function, no?

src/hb-subset-input.cc
```diff
@@ -734,6 +734,343 @@ hb_subset_axis_range_to_string (hb_subset_input_t *input,
...
+
+  if (!_is_set_default (input->sets.glyphs, default_input->sets.glyphs))
+    _format_number_set_arg (input->sets.glyphs, "--gids", false, sb);
```
behdad: Can you add two macros for handling number/tag sets each?

src/hb-subset-input.cc
```diff
@@ -734,6 +734,343 @@ hb_subset_axis_range_to_string (hb_subset_input_t *input,
...
+      opt.push ('\0');
+      if (opt.arrayZ) _append_arg (sb, opt.arrayZ);
+    }
+  }
+}
+
+static void
+_format_number_set_arg (const hb_set_t *set,
+                        const char *opt_name,
+                        bool is_hex,
```
behdad: can we take `base` instead of `is_hex`. maybe as an optional last arg?

src/hb-subset-input.cc
```diff
@@ -734,6 +734,343 @@ hb_subset_axis_range_to_string (hb_subset_input_t *input,
...
+  if (unlikely (!_format_glyph_map_arg (input->glyph_map, sb)))
+    return nullptr;
+
+  if (unlikely (!_format_axes_location_arg (input->axes_location, sb)))
+    return nullptr;
+
+  if (unlikely (sb.in_error ()))
+    return nullptr;
+
+  char *buf = (char *) hb_malloc (sb.length + 1);
```
behdad: Use `sb.steal()`.

src/hb-subset-input.cc
```diff
@@ -837,13 +836,16 @@ _format_set_arg (const hb_set_t *set,
                  Formatter &&formatter,
                  hb_vector_t<char> &sb)
 {
+  char opt_prefix[64];
+  snprintf (opt_prefix, sizeof (opt_prefix), "--%s", opt_name);
```
behdad: No need to produce this value.

src/hb-subset-input.cc
```diff
@@ -1050,35 +1052,36 @@ hb_subset_input_to_string_or_fail (hb_subset_input_t *input)
   {
     if (input->flags & fo.flag)
     {
-      _append_arg (sb, fo.option);
+      char buf[36];
+      snprintf (buf, sizeof (buf), "--%s", fo.option);
```
behdad, line 1053: Same here.

behdad:
LGTM. Thanks!

src/hb-subset-input.cc
garretrieger: This one needs to be appended as a single string.

PR #6076, opened 2026-07-17
[paint] pass color_glyph its own user_data

Nussu06:
hb_paint_funcs_t::color_glyph dispatches with user_data->push_clip_glyph where every other wrapper in the file passes its own slot, so a client that registers color_glyph and push_clip_glyph with different context objects has the push_clip_glyph context delivered to its color_glyph callback and casts it to the wrong type. Both COLR.hh:2874 and hb-ft-colr.hh:417 reach it, so it fires on any color glyph paint; hb_paint_funcs_destroy already reads user_data->color_glyph, so the slot is allocated and freed correctly and just never handed to the callback. I swept the other 17 wrappers here plus hb-draw.hh and hb-font.hh and this is the only mismatch. The added test registers the two callbacks with distinct contexts and aborts on the unpatched tree.

### DIFF
--- a/src/hb-paint.hh
+++ b/src/hb-paint.hh
@@ -93,7 +93,7 @@ struct hb_paint_funcs_t
   { return func.color_glyph (this, paint_data,
                              glyph,
                              font,
-                             !user_data ? nullptr : user_data->push_clip_glyph); }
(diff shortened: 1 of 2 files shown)


behdad:
Ouch. Thanks!

PR #6077, opened 2026-07-17
[subset-repack] bug fixes

qxliu76:
1. fix possible OOB accesses, basically make a deep copy before we mutate an object. 
    Also tried to make as_table() return a const table pointer, gemini helped with template thing, I couldn't make it work in a cleaner way..
3. fix is_shared() API


src/graph/graph.hh
```diff
@@ -1085,8 +1097,27 @@ struct graph_t
       return -1;
     }
 
-    clone->obj.head = child.obj.head;
-    clone->obj.tail = child.obj.tail;
+    if (deep_copy)
+    {
+      unsigned table_size = child.obj.tail - child.obj.head;
+      if (table_size)
```
jfkthame: What if `table_size` is zero? (Unexpected, probably, but should we nevertheless set `child.obj.head` and `.tail` safely here?)

qxliu76: I think the `child.obj.head` and `.tail` would be nullptr then? See object_t constructor: https://github.com/harfbuzz/harfbuzz/blob/ac9227e0a234add05f21907ed1cfe25ecaab7c7a/src/hb-serialize.hh#L74

garretrieger: Probably the easiest thing to do would be to assign the existing head/tail pointers. Since size is 0 duplication isn't needed and this will ensure that the head pointer points at valid memory address (nothing should be attempting to access the memory either way).

src/graph/graph.hh
garretrieger, line 265: Should clear the virtual_parents set here, in the event that the single parent is also a virtual link

garretrieger, line 313: If virtual_parents size() is >0 then we need to remap the value in it as well.

src/graph/graph.hh
jfkthame: > I think the `child.obj.head` and `.tail` would be nullptr then? See object_t constructor:
> 
> [harfbuzz/src/hb-serialize.hh](https://github.com/harfbuzz/harfbuzz/blob/ac9227e0a234add05f21907ed1cfe25ecaab7c7a/src/hb-serialize.hh#L74)
> 
> Line 74 in [ac9227e](/harfbuzz/harfbuzz/commit/ac9227e0a234add05f21907ed1cfe25ecaab7c7a)
>  object_t () = default;

It looks like the `clone` object is created by `vertices_.push()` above; but does `push()` with no argument _value-initialize_ the new element (which would zero-initialize the pointers), or does it _default-initialize_ it (which would leave them as garbage)? That's not obvious from here...

Looking into `hb_vector_t::push`, it appears the answer is that it will zero-initialize (_not_ value-initialize, nor default-initialize!) the newly-appended element, so it turns out this would actually be ok. But personally, I'd vote for being more explicit here.

src/graph/graph.hh
```diff
@@ -437,19 +426,23 @@ struct graph_t
     }
   };
 
-  template <typename T>
```
garretrieger, line 440: Instead of using a bool you could use a enum:
```
enum vertex_mutability_t {
  Immutable,
  Mutable
};
```

Would make the type more descriptive (eg. vertex_and_table_t<T, Mutable>)

src/graph/graph.hh
```diff
@@ -712,32 +705,51 @@ struct graph_t
...
+
+    if (!r.table->sanitize (*(r.vertex), std::forward<Ts>(ds)...))
+      return vertex_and_table_t<T, is_const> ();
+
+    return r;
   }
 
   template <typename T, typename ...Ts>
-  vertex_and_table_t<T> as_table_from_index (unsigned index, Ts... ds)
+  vertex_and_table_t<T, true> as_table_from_index (unsigned index, Ts... ds) const
```
garretrieger: This is duplicated from the above more generic version, is it needed?

src/graph/graph.hh
```diff
@@ -1065,7 +1077,7 @@ struct graph_t
   /*
    * Creates a copy of node_idx and returns it's new index.
    */
-  unsigned duplicate (unsigned node_idx)
+  unsigned duplicate (unsigned node_idx, bool deep_copy = false)
```
garretrieger: deep_copy is a bit ambigous here (for example it implies to me that children are also copied), suggest naming something like "copy_table" instead.

src/graph/graph.hh
qxliu76: Done

garretrieger (approved):

behdad:
I had to revert this, since some fuzzer test was timed out. Might be irrelevant to this one though. @qxliu76 can you open a new PR based on master so we can see the effect? I wonder if your increase of vertices limit plus this, has the effect. Thanks.

PR #6082, opened 2026-07-20
[raster] Reject non-finite texel coords in hb_raster_paint_image

Nussu06:
both pixel loops in hb_raster_paint_image gate the sampler with a plain range test on ix/iy, and every comparison in it is false for NaN, so NaN coordinates walk straight into hb_raster_sample_bilinear_premul and hit (int) floorf. only x1/y1 get clamped there, so x0/y0 index src[] with no lower bound at all. the determinant guard upstream has the same shape (fabsf (det) < 1e-10f is also false for NaN), so a transform that overflowed to infinity is accepted and inv_det = 1/inf = 0 turns inf * 0 into NaN. nested PaintScale/PaintTransform reach this since hb_raster_paint_push_transform composes multiplicatively. on x86-64 (int) floorf (NaN) is INT_MIN and the index becomes 0xfffffffd80000000, a wild read of the glyph bitmap; aarch64 saturates to 0 and merely renders garbage. checking isfinite in the guard matches what evaluate_color_line, normalize_gradient_t and lookup_gradient_lut already do for gradient parameters in this file. the added test trips ubsan at hb-raster-paint.cc:59 without the fix and is clean with it; full suite green locally.

### DIFF
--- a/src/hb-raster-paint.cc
+++ b/src/hb-raster-paint.cc
@@ -946,7 +946,8 @@ hb_raster_paint_image (hb_paint_funcs_t *pfuncs HB_UNUSED,
 	float ix = (gx - img_x) / img_sx;
 	float iy = (float) (src_height - 1) - (gy - img_y) / img_sy;
 
-	if (ix < 0.f || iy < 0.f ||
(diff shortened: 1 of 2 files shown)


behdad:
Thanks.

PR #6087, opened 2026-07-21
[colr] clamp alpha and radius deltas before unsigned conversion

Nussu06:
color_alpha in hb_paint_context_t multiplies the palette color alpha by the paint alpha and returns unsigned, but that alpha is an F2DOT14 field (PaintSolid, PaintVarSolid, ColorStop) plus a variation delta, so a font can make it negative and the conversion is undefined; patching a PaintSolid alpha to -1.0 in test/api/fonts/test_glyphs-glyf_colr_1.ttf and painting every glyph gives "COLR.hh:125: -255 is outside the range of representable values of type 'unsigned int'" under ubsan. PaintRadialGradient::subset has the same conversion for radius0/radius1 while the x/y siblings next to it already go through (int); `hb-subset --unicodes='*' --instance=GRR0=-1000 test/api/fonts/test_glyphs-glyf_colr_1_variable.ttf` reports it at COLR.hh:811 and GRR1 at :814. clamping is done in the callee since both values come straight off the wire and every caller would otherwise need the same guard; it also drops the wrap that turned an over-range alpha into a garbage byte in HB_COLOR and a negative radius delta into a 16-bit radius near 0xffff. full meson test suite passes and the three ubsan reports are gone.

### DIFF
--- a/src/OT/Color/COLR/COLR.hh
+++ b/src/OT/Color/COLR/COLR.hh
@@ -122,7 +122,9 @@ public:
   static unsigned
   color_alpha (hb_color_t color, float alpha)
   {
-    return roundf (hb_color_get_alpha (color) * alpha);
+    /* Font-supplied alpha is F2DOT14 plus variation deltas, so it can be
+     * negative or greater than one; converting that to unsigned is UB. */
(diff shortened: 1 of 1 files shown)


behdad:
Thanks.

PR #6090, opened 2026-07-22
[raster] clamp coordinates before converting to int

Nussu06:
hb_raster_draw_set_glyph_extents and hb_raster_paint_set_glyph_extents are line-for-line copies that transform the four corners of the caller's hb_glyph_extents_t and then hand the resulting floats to (int) floorf / (int) ceilf. the members are int32, so any magnifying transform or scale factor puts the corners well outside the int range, which makes the conversion undefined, and the width/height are then computed as ex1 - ex0 in int, which overflows once the saturated endpoints sit at opposite ends. emit_segment has the same shape one level down, scaling by HB_RASTER_ONE_PIXEL before the int32 cast. ubsan on test/fuzzing/fonts/clusterfuzz-testcase-minimized-harfbuzz_fuzzer-6712347260092416 reports float-cast-overflow at hb-raster-draw.cc:385-389 and :486-489 and hb-raster-paint.cc:2084-2087, plus signed-integer-overflow at hb-raster-draw.cc:399 and hb-raster-paint.cc:2098; hb_raster_draw_set_glyph_extents and hb_raster_paint_set_glyph_extents also reach them directly from the public api, which is what the added test does. saturating in double keeps INT32_MIN/INT32_MAX exact and gives NaN a defined landing spot, and it is the same shape the file already uses for the int64 cell coordinates. the remaining float-to-int conversions in the backend take values the callers already range-check (the bilinear sampler) or that are bounded by the surface (push_clip_rectangle), and none of them showed up in the ubsan run, so i left them alone.

### DIFF
--- a/src/hb-raster.hh
+++ b/src/hb-raster.hh
@@ -29,6 +29,17 @@
 
 #include "hb.hh"
 
+/* Convert a coordinate to int32_t, saturating instead of relying on the
+ * undefined behavior of an out-of-range float-to-int conversion.  NaN
+ * saturates to INT32_MIN. */
+static HB_ALWAYS_INLINE int32_t
+hb_raster_clamp_to_int32 (float v)
+{
+  return (int32_t) hb_clamp ((double) v,
+			     (double) INT32_MIN,
+			     (double) INT32_MAX);
+}
+
 /* Shared pixel helpers (used by paint and image compositing). */
 
 static HB_ALWAYS_INLINE uint8_t
--- a/src/hb-raster.hh
+++ b/src/hb-raster.hh
@@ -29,17 +29,6 @@
 
 #include "hb.hh"
 
-/* Convert a coordinate to int32_t, saturating instead of relying on the
- * undefined behavior of an out-of-range float-to-int conversion.  NaN
- * saturates to INT32_MIN. */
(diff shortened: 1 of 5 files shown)


src/hb-raster.hh
```diff
@@ -29,6 +29,17 @@
 
 #include "hb.hh"
 
+/* Convert a coordinate to int32_t, saturating instead of relying on the
+ * undefined behavior of an out-of-range float-to-int conversion.  NaN
+ * saturates to INT32_MIN. */
+static HB_ALWAYS_INLINE int32_t
+hb_raster_clamp_to_int32 (float v)
```
behdad: This can be moved to be named `hb_clamp_to_int32()`, no? Or even: `hb_clamp_to<int32_t> ()`.

Nussu06: Good call. Pulled it into hb-algs.hh as a generic hb_clamp_to<T> (saturating to hb_int_min/hb_int_max of T, NaN to the minimum) and switched the three raster sites to hb_clamp_to<int32_t>. Rebuilt with the sanitizers and the regression test plus the fuzz seed are still clean.

behdad: Thanks!

PR #6099, opened 2026-07-25
Add hb_malloc2/hb_realloc2 for overflow-checked array allocation

thalha-a9:
Add `hb_malloc2(nmemb, size)` and `hb_realloc2(ptr, nmemb, size)` that check for multiplication overflow before allocating, without paying the cost of zero-initialization.

**Changes:**

- Add `hb_malloc2` / `hb_realloc2` to the public API (`hb-common.h` / `hb-common.cc`)
- Convert all `hb_malloc(count * sizeof(T))` sites to `hb_malloc2(count, sizeof(T))` (12 files)
- Convert `hb_realloc(ptr, count * sizeof(T))` to `hb_realloc2` in `hb-vector.hh` and `hb-buffer.cc`
- Clean up now-redundant manual `hb_unsigned_mul_overflows` checks in `hb-vector.hh` and `hb-buffer.cc`
- Fix raw `malloc`/`free` → `hb_malloc2`/`hb_free` in `main.cc` and `wasm/graphite/shape.cc`

### DIFF
--- a/src/OT/glyf/glyf-helpers.hh
+++ b/src/OT/glyf/glyf-helpers.hh
@@ -94,7 +94,7 @@ _add_loca_and_head (hb_subset_context_t *c,
   unsigned num_offsets = c->plan->num_output_glyphs () + 1;
   unsigned entry_size = use_short_loca ? 2 : 4;
 
-  char *loca_prime_data = (char *) hb_malloc (entry_size * num_offsets);
+  char *loca_prime_data = (char *) hb_malloc2 (num_offsets, entry_size);
 
   if (unlikely (!loca_prime_data)) return false;
 
--- a/src/hb-algs.hh
+++ b/src/hb-algs.hh
@@ -39,6 +39,19 @@
 #include <functional>
 #include <new>
 
+static HB_ALWAYS_INLINE void *
+hb_malloc2 (size_t nmemb, size_t size)
+{
+  if (size && nmemb > SIZE_MAX / size) return nullptr;
+  return hb_malloc (nmemb * size);
+}
+static HB_ALWAYS_INLINE void *
+hb_realloc2 (void *ptr, size_t nmemb, size_t size)
+{
+  if (size && nmemb > SIZE_MAX / size) return nullptr;
+  return hb_realloc (ptr, nmemb * size);
+}
+
 /*
  * Flags
  */
--- a/src/hb-cairo-utils.cc
+++ b/src/hb-cairo-utils.cc
@@ -237,7 +237,7 @@ _hb_cairo_get_color_stops (hb_cairo_context_t *c,
     return false;
   if (len > *count)
   {
-    *stops = (hb_color_stop_t *) hb_malloc (len * sizeof (hb_color_stop_t));
+    *stops = (hb_color_stop_t *) hb_malloc2 (len, sizeof (hb_color_stop_t));
     if (unlikely (!*stops))
       return false;
     allocated = true;
--- a/src/hb-gpu-paint.cc
+++ b/src/hb-gpu-paint.cc
@@ -877,7 +877,7 @@ hb_gpu_paint_emit_sweep (hb_gpu_paint_t  *c,
   hb_color_stop_t *heap_stops = nullptr;
   if (count > 16)
   {
-    heap_stops = (hb_color_stop_t *) hb_malloc (count * sizeof (hb_color_stop_t));
+    heap_stops = (hb_color_stop_t *) hb_malloc2 (count, sizeof (hb_color_stop_t));
     if (unlikely (!heap_stops)) { c->unsupported = true; return; }
     stops = heap_stops;
   }
--- a/src/hb-map.hh
+++ b/src/hb-map.hh
@@ -204,7 +204,7 @@ struct hb_hashmap_t
 
     unsigned int power = hb_bit_storage (hb_max (hb_max ((unsigned) population, new_population) * 2, 4u));
     unsigned int new_size = 1u << power;
-    item_t *new_items = (item_t *) hb_malloc ((size_t) new_size * sizeof (item_t));
+    item_t *new_items = (item_t *) hb_malloc2 ((size_t) new_size, sizeof (item_t));
     if (unlikely (!new_items))
     {
       successful = false;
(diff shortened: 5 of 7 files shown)


thalha-a9:
Hi — this is a straightforward conversion of 10 `hb_malloc(count * sizeof(T))` sites to `hb_calloc(count, sizeof(T))`, which checks for integer overflow in the size computation before allocating.

The wasm/graphite sites also fix raw `malloc`/`free` calls → `hb_calloc`/`hb_free` for API consistency.

All buffers are filled immediately after allocation, so the zero-initialization from `hb_calloc` is harmless. Happy to adjust anything if needed.

behdad:
Thanks. I prefer NOT to pay for the zero-initialization. How about we add a `hb_malloc2()` to do the multiplication overflow? We already have multiplication overflow code scattered at other `hb_malloc` sites, we can clean those up as well.

thalha-a9:
Done, reworked as per your suggestion.

Added `hb_malloc2(nmemb, size)` and `hb_realloc2(ptr, nmemb, size)` that check for multiplication overflow without zero-initializing. Converted all sites to use these instead of `hb_calloc`.

Also cleaned up the scattered manual `hb_unsigned_mul_overflows` checks that are now redundant (in `hb-vector.hh` and `hb-buffer.cc`), and fixed raw `malloc`/`free` calls in `main.cc` and `wasm/graphite/shape.cc` to use the project allocator.

Builds clean with cmake.

behdad:
Any ideas why bots unhappy?

thalha-a9:
Fixed, the test failure was in `test-buffer.c:419` (`test_buffer_allocation`). 

When I removed the manual `hb_unsigned_mul_overflows` guard in `hb_buffer_t::enlarge`, the growth loop `new_allocated += (new_allocated >> 1) + 32` could wrap around on large sizes before reaching `hb_realloc2`. Added an unsigned overflow check inside the loop (detect when `new_allocated` wraps below `prev`) and `goto done` on wrap.

Pushed, CI should be green now.

behdad:
Humm. Seems failing still. Thanks for looking.

thalha-a9:
Found it, and it was my fault, not the bots'. Two independent issues, both fixed in the force-push to `f81fb12`.

**1. `test_buffer_allocation` (test-buffer.c:419) - 64-bit only**

The tell was that `win32` and `msvc-2025-x86` passed while every 64-bit job failed.

I had replaced the `hb_unsigned_mul_overflows()` guards in `hb_buffer_t::enlarge()` and `hb_vector_t::_realloc()` with the new helpers. Those are **not** equivalent:

- `hb_unsigned_mul_overflows (unsigned int count, unsigned int size, unsigned *result)` enforces that the resulting *byte count fits in 32 bits*.
- `hb_malloc2`/`hb_realloc2` do a `size_t` check.

On LP64, `hb_buffer_pre_allocate (b, UINT_MAX/20 + 1)` wants `214748365 * 20 = 4294967300` bytes. That overflows 32 bits but is a perfectly legal `size_t`, so under overcommit the allocation succeeded and the buffer reported success where the test expects failure. On ILP32 `size_t` is 32-bit, so the check happened to still fire - hence the split.

Since `hb_buffer_t::len`/`::allocated` are `unsigned int` and `hb_vector_t::allocated` is `int`, that 32-bit cap is a real invariant, not incidental. It can't be folded into `hb_malloc2` without also making the helpers 32-bit-limited, which would be wrong for their general use.

So I've **reverted `hb-buffer.cc` and `hb-vector.hh` entirely** — both were already overflow-safe upstream and didn't need touching. The new helpers are now used only at the call sites that genuinely had an unchecked `n * sizeof (T)`.

**2. `docs` job**

`hb_malloc2` and `hb_realloc2` are new public symbols and weren't listed in `docs/harfbuzz-sections.txt`, so gtkdoc-check flagged them as undocumented. Added. I also had `Since: 11.2.0` copy-pasted in — corrected to `14.3.0`.

**Also dropped:** the `src/wasm/graphite/shape.cc` hunk. That module is compiled into the wasm sandbox against `hb-wasm-api.h` and can't link libharfbuzz's `hb_malloc`, so `malloc`/`free` are correct there.

Remaining diff is now just the two helpers plus these call sites:

- `src/OT/glyf/glyf-helpers.hh`
- `src/hb-cairo-utils.cc`
- `src/hb-gpu-paint.cc`
- `src/hb-map.hh`
- `src/hb-ot-post-table.hh`
- `src/hb-shape.cc`
- `src/main.cc` (also switched a stray `malloc`/`free` pair to the `hb_` allocators)

The workflows need maintainer approval to run - could you kick them off when you get a chance? Happy to split the helpers into their own commit if you'd prefer that for review.

behdad:
Thanks. For the sake of simplicity, I suggest the `hb_malloc2` etc be internal API, not public. They can live in `hb-algs.hh` and marked `ALWAYS_INLINE` even.

behdad:
> Thanks. For the sake of simplicity, I suggest the `hb_malloc2` etc be internal API, not public. They can live in `hb-algs.hh` and marked `ALWAYS_INLINE` even.

And they can call `hb_malloc` / `hb_realloc` instead of calling the impl method directly.

thalha-a9:
Done — moved `hb_malloc2`/`hb_realloc2` to `hb-algs.hh` as `static HB_ALWAYS_INLINE`, calling `hb_malloc`/`hb_realloc`. Removed from public header, docs sections, and `hb-common.cc`. Squashed to a single commit.

behdad:
Can you undo any main.cc changes? That file is not important for anything. We should kill it some time. Thanks.

thalha-a9:
Done, reverted main.cc changes.

thalha-a9:
All checks green except one pre-existing flake - `raster-fuzzer-chunk-2` timed out (SIGTERM after 30s). 245/246 tests passed, 0 failures. Unrelated to this change.

khaledhosny:
> All checks green except one pre-existing flake - `raster-fuzzer-chunk-2` timed out (SIGTERM after 30s). 245/246 tests passed, 0 failures. Unrelated to this change.

There are no flaky tests. All tests must pass.

thalha-a9:
On it!

thalha-a9:
@khaledhosny You're right, apologies for the misleading wording. All 27 checks are passing now, 0 failures, 0 timeouts.

@behdad Ready to merge whenever you get a chance!

PR #6100, opened 2026-07-25
[subset] Add hb_subset_cff_get_glyph_cid()

khaledhosny:
Exposes cff1::glyph_to_sid()

Subsetting preserves the CIDs of the input font, so the glyph indices of a subset font are generally not its CIDs. Consumers that address glyphs by CID need this mapping, e.g. PDF producers embedding the subset as a CIDFontType0 font, where CIDs are mapped to glyphs through the charset.

### DIFF
--- a/docs/harfbuzz-sections.txt
+++ b/docs/harfbuzz-sections.txt
@@ -991,6 +991,7 @@ hb_subset_plan_unicode_to_old_glyph_mapping
 hb_subset_plan_new_to_old_glyph_mapping
 hb_subset_plan_old_to_new_glyph_mapping
 hb_subset_preprocess
+hb_subset_cff_get_glyph_cid
 hb_subset_flags_t
 hb_subset_input_t
 hb_subset_sets_t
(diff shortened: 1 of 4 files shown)


behdad:
Does the subsetting `old_to_new` glyphID APIs not address this usecase?

khaledhosny:
This is about CID mapping of CID-keyed CFF table, which is independent of glyph IDs.

behdad:
> This is about CID mapping of CID-keyed CFF table, which is independent of glyph IDs.

Right. I was just curious if you already have the gid->CID of the original font, then you can map it to new gid->CID. I'm just trying to understand the dataflow better.

khaledhosny:
> > This is about CID mapping of CID-keyed CFF table, which is independent of glyph IDs.
> 
> Right. I was just curious if you already have the gid->CID of the original font, then you can map it to new gid->CID. I'm just trying to understand the dataflow better.

To git the original mapping I’d need to parse CFF charset, and this is what I’m trying to move away from.

Alternative to this API, we can add a subsetter option to convert the CFF charset to an identity one, and CID would be equal to glyphs ids and we wouldn’t need any mapping.

khaledhosny:
> Alternative to this API, we can add a subsetter option to convert the CFF charset to an identity one, and CID would be equal to glyphs ids and we wouldn’t need any mapping.

https://github.com/harfbuzz/harfbuzz/pull/6101 for the alternative approach.

khaledhosny:
Closing in favor of https://github.com/harfbuzz/harfbuzz/pull/6101

PR #6101, opened 2026-07-26
[subset] Add HB_SUBSET_FLAGS_CFF_IDENTITY_CHARSET

khaledhosny:
When set and subsetting a CID-keyed CFF font, the output charset is written as identity: CID = new GID.


behdad:
This looks better to me. @garretrieger @qxliu76

qxliu76:
LGTM. But to be honest, I do not have sufficient knowledge or prior experience with CFF code.:)

garretrieger (approved):

PR #6115, opened 2026-07-30
[GPOS] Apply mark-attachment cross-direction offsets immediately

### LINKED ISSUE #4596: Difference mark positioning interaction with cursive attachemnt between Harfbuzz and Uniscribe/DirectWrite/CoreText
In this attached font, there are two lookups a mark positioning lookup and a cursive attachment one, associated with `mark` and `curs` features, respectively.

The font is designed so that marks always have the same vertical positioning, regardless of whether the base glyph sets on the baseline or not. To achieve this, the mark lookup is ordered before the cursive lookup.

This works in Uniscribe/DirectWrite as well as CoreText, but not in HarfBuzz. HarfBuzz seems to always move the mark with the base glyph regardless of the order of the lookups. If I’m not mistaken, this is an intentional feature of HarfBuzz, but there are two issues here:
1. This causes compatibility issues between HarfBuzz and the other implementations,
2. There does not seem to be a way to achieve the desired result here using HarfBuzz.

Test font: [Quraan-Font-bug-report.otf.zip](https://github.com/harfbuzz/harfbuzz/files/14379650/Quraan-Font-bug-report.otf.zip)
Test string: "سَیَقُوْلُ  سَقَمُنَ"
HarfBuzz:
hb
DirectWrite:
dw
CoreText:
ct

(This was brought to my attention by @Zishan-Nasar, who provided the test font)

khaledhosny:
* Fixes https://github.com/harfbuzz/harfbuzz/issues/4596, see the issue for more details.

Instead of deferring mark positioning to the end of GPOS application, we now apply cross-direction offsets (y-direction in horizontal text) immediately and keep applying main-direction (x-direction in horizontal text) as before.

This is still not 100% compatible with DirectWrite or Core Text (they apply both directions immediately), but enough to fix the reported issue.

Based on Behdad’s patch below, but fixes the Noto Nastaliq regression by resolving the base offset through the whole cursive attachment chain since cursive offsets are relative the previous glyph and get finalized only at end of GPOS:
https://github.com/harfbuzz/harfbuzz/issues/4596#issuecomment-2276300688

The remaining failure is not a regression, the new expectation now match DirectWrite and Core Text.

The tests include a subsetted bug font, as well as two constructed fonts where curs lookup come before mark ones, once with kern before mark and with kern after mark, to show that lookup order affects y but not x-direction.


behdad:
LGTM. Thanks.

PR #6117, opened 2026-07-30
[ot] Add hb_ot_fetch_bits() and hb_ot_fetch_number()

### LINKED ISSUE #1920: Expose OS/2 fsSelection

### LINKED ISSUE #6110: Add API to get font bbox
I need this for writing bbox when embedding fonts in PDF. Should it be separate API or should I use for the the reserved slots in `hb_font_extents_t`?

khaledhosny:
* Fixes https://github.com/harfbuzz/harfbuzz/issues/1920
* Fixes https://github.com/harfbuzz/harfbuzz/issues/6110


khaledhosny:
@behdad Should we add macros to give nice names for the various bits? e.g. `HB_OT_FS_TYPE_INSTALLABLE` or `HB_OT_FS_SELECTION_USE_TYPO_METRICS`? I’m split, they are nicer to use but there is too many of them (23 one).

behdad:
> @behdad Should we add macros to give nice names for the various bits? e.g. `HB_OT_FS_TYPE_INSTALLABLE` or `HB_OT_FS_SELECTION_USE_TYPO_METRICS`? I’m split, they are nicer to use but there is too many of them (23 one).

Nah. Too many, and C-only.

khaledhosny:
> > @behdad Should we add macros to give nice names for the various bits? e.g. `HB_OT_FS_TYPE_INSTALLABLE` or `HB_OT_FS_SELECTION_USE_TYPO_METRICS`? I’m split, they are nicer to use but there is too many of them (23 one).
> 
> Nah. Too many, and C-only.

Exactly!

src/hb-config.hh
```diff
@@ -62,6 +62,7 @@
 #define HB_NO_BUFFER_SERIALIZE
 #define HB_NO_BUFFER_VERIFY
 #define HB_NO_BITMAP
+#define HB_NO_BITS
```
behdad: This should be `HB_NO_OT_BITS` but also the bit vs number bothers me. 

Maybe we should call it `hb_ot_fetch_number()` and `hb_ot_fetch_bits()` and call it `HB_NO_OT_FETCH`. Reads better.

behdad (approved):
Approved with comment. Thanks.

src/hb-config.hh
khaledhosny: Sounds good. I renamed everything except the enums. Lets me know if you want to rename these as well.

behdad: I think I'm fine with that, instead of adding a `_FETCH` in the middle of all.

khaledhosny:
The msan failure seems to be happening on main as well.

behdad:
Yes. I'm on it soon. I don't know why it started failing.

khaledhosny:
> Yes. I'm on it soon. I don't know why it started failing.

Apparently, the bot was not actually doing anything. It has thousands of glib failure that were ignored, and now the VM dies.

behdad:
Yes, because glib isn't built for sanitizer.

src/meson.build
```diff
@@ -313,6 +314,7 @@ hb_base_headers = files(
   'hb-map.h',
   'hb-ot-color.h',
   'hb-ot-deprecated.h',
+  'hb-ot-fetch.h',
```
Andarwinux, line 317: missed in CMakeLists.txt?

PR #6123, opened 2026-07-30
Respect HB_NO_AAT_SHAPE in KernSubTable::collect_glyphs

garretrieger:
Sanitize is skipped for subtable 1 and 3 when HB_NO_AAT_SHAPE is set, so it's not safe to run collect_glyphs in this case. This mirrors dispatch().


behdad:
LGTM.

PR #6130, opened 2026-07-31
[subset] Convert repacker graph traversals to BFS.

garretrieger:
None of the traversals require DFS order so BFS can be used instead which eliminates the possibility of stack overflow. For the queue we re-use the existing ordering_scratch_ space to avoid allocating any additional memory. As a bonus this fixes a bug in find_32bit_roots() that was the result of DFS traversal order.

All of the directed traversal methods are unified to a single common implementation.

### DIFF
--- a/src/graph/graph.hh
+++ b/src/graph/graph.hh
@@ -928,11 +928,21 @@ struct graph_t
     return true;
   }
 
-  // BFS graph traversal starting at start_idx, op() will be called for each edge (plus once for the root).
-  // if it returns false the children of the edges destination will not be traversed.
-  // Does not implement a visited set, it's expected that op() will handle that as needed.
-  template <typename Op>
-  void traverse_directed_bfs (unsigned start_idx, Op&& op)
+  // BFS graph traversal starting at start_idx.
+  //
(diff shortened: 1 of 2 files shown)


behdad:
cc @qxliu76 for Rust port.

qxliu76:
> cc @qxliu76 for Rust port.

Thank you! I'll work on this.

src/graph/graph.hh
```diff
@@ -919,41 +928,113 @@ struct graph_t
     return true;
   }
 
-  void find_subgraph (unsigned node_idx, hb_map_t& subgraph)
+  // BFS graph traversal starting at start_idx, op() will be called for each edge (plus once for the root).
+  // if it returns false the children of the edges destination will not be traversed.
+  // Does not implement a visited set, it's expected that op() will handle that as needed.
+  template <typename Op>
+  void traverse_directed_bfs (unsigned start_idx, Op&& op)
```
behdad: Maybe op can be renamed to something more clear, and documented better? The rest looks good to me. Thanks!

garretrieger: Renamed to visit_edge and added more detailed documentation.

PR #6140, opened 2026-08-05
[subset] avar2 partial-instancing fixes: never-downgrade NULL VarStore, exact reachable ranges

behdad:
Four improvements to the avar2 partial-instancing code, following the review of the sibling fontTools work (fonttools/fonttools#4045) and an audit of this port against it:

- **Never downgrade avar2 / fix NULL-VarStore failure.** An avar version 2 table with a NULL varStore offset is legal and maps like plain v1, but \`hb-subset --instance\` failed outright on such fonts (the planner errored against the missing store). Per the never-downgrade policy, such fonts stay on the avar2 path: mapping entries that don't resolve to a real store row get runtime "no variation" semantics, and offset compensation creates delta rows on demand in an empty store. The instance stays version 2 and bit-exact. Comes with a subset test (a null-VarStore variant of RobotoDelta) that fails outright without the fix. fontTools now implements the same policy (fonttools/fonttools@bf6f82741).

- **Exact reachable ranges for culling.** The per-axis reachable-range bound used per-region interval arithmetic — sound, but loose enough that culling rarely fires on real parametric fonts. The delta is piecewise multilinear in the axis coordinates, so its exact extremes lie on the grid of per-axis tent breakpoints; enumerate that grid when small (real fonts: exact, microseconds) with the interval bound kept as the conservative fallback for crafted many-axis rows. Ports fontTools' \`VarStore.getExtremes\` rewrite.

- **Kink-complete offset-error estimation.** The estimator that picks between offset-compensation knot sets (and gates the approximation warning) sampled a uniform 257-point grid only; now it also samples the user-space preimages of every residual kink, so the worst kink can't fall between samples.

- **Strict comparisons in region culling.** The support scalar is 1 (not 0) at v == peak, so a boundary-peaked tent is live at the range edge; the ±1-unit range padding already masked this in practice, but the predicate is now correct on its own terms.

Cross-validated against fontTools: instancing the avar2 Amstelvar fixture and comparing final normalized coordinates through the spec pipeline agrees within 1 F2Dot14 unit at all sampled locations, with no live gvar variations culled. Full meson suite passes (264/264).


qxliu76:
Thanks! How did you find out those issues? (I'm unable to figure out those so I'm curious). Really appreciate your continuous efforts to improve this code. 
I'll take a look, hopefully my memory about this isn't fuzzy yet. I might need some time to re-familiarize myself with the details.

behdad:
> Thanks! How did you find out those issues? (I'm unable to figure out those so I'm curious). Really appreciate your continuous efforts to improve this code. I'll take a look, hopefully my memory about this isn't fuzzy yet. I might need some time to re-familiarize myself with the details.

Thank you. Cosimo and Marc did Fable reviews of the FontTools counterpart and found issues. I addressed those using Fable, then asked it to update HarfBuzz work to match.

https://github.com/fonttools/fonttools/pull/4045

behdad:
@qxliu76 Any luck with this? Thanks.

qxliu76:
> @qxliu76 Any luck with this? Thanks.

oops really sorry for my delay, I'll make sure to review it tomorrow.

src/hb-ot-var-avar-table.hh
```diff
@@ -404,6 +404,35 @@ _avar2_map_new_mapping (const hb_vector_t<AxisValueMap> &mappings, double v)
            + (double) mappings.arrayZ[len - 1].coords[1].to_float ();
 }
 
+/* Inverse of _avar2_map_new_mapping: pull an output coordinate back to a
+ * preimage input coordinate. For a non-strictly-monotone mapping this picks
+ * one preimage, which is fine for its only use (augmenting the error
+ * estimator's sample set). */
+static inline double
+_avar2_unmap_new_mapping (const hb_vector_t<AxisValueMap> &mappings, double v)
```
qxliu76, line 414: This code is pretty similar to https://github.com/harfbuzz/harfbuzz/blob/dfdc088c4d7c5d31dd5b13070b919b51f6c21ea8/src/hb-ot-var-avar-table.hh#L383, I wonder if we could do something like https://github.com/harfbuzz/harfbuzz/blob/dfdc088c4d7c5d31dd5b13070b919b51f6c21ea8/src/hb-ot-var-avar-table.hh#L245

src/hb-subset-plan-var.cc
```diff
@@ -136,45 +136,178 @@
...
+       hb_pair_t<double, double> *input;
+       double blo = -1.0, bhi = +1.0;
+       if (box.has (t, &input)) { blo = input->first; bhi = input->second; }
+       grid_tags.push (t);
+       grid_points.push (hb_vector_t<double> ());
+       if (grid_tags.in_error () || grid_points.in_error ()) return -1;
+       auto &points = grid_points[grid_points.length - 1];
+       points.push (blo);
+       points.push (bhi);
+       if (points.in_error () || !grid_pos.set (t, grid_tags.length - 1)) return -1;
```
qxliu76: seems we don't need to use `grid_tags` vector , it's only used to increment `pos`, I think we can use `grid_pos.get_population()` instead?

src/hb-subset-plan-var.cc
```diff
@@ -136,45 +136,178 @@
...
+      * VarStore.getExtremes. */
+     constexpr unsigned MAX_GRID = 1u << 14;
+     bool exact = true;
+     double dmin = 0.0, dmax = 0.0; /* delta extremes, F2Dot14 units */
+     double vmin = 0.0, vmax = 0.0; /* identity + delta extremes, coord units */
+
+     /* Grid axes: the target axis (identity term), plus every axis a valid
+      * tent of an active region references. */
+     hb_vector_t<hb_tag_t> grid_tags;
+     hb_hashmap_t<hb_tag_t, unsigned> grid_pos;
```
qxliu76: can we move `grid_pos` and `grid_points `out of the for loop to save some allocs?

src/hb-subset-plan-var.cc
```diff
@@ -136,45 +136,178 @@
...
+	 /* Tents the runtime ignores (constant scalar 1) add no breakpoints. */
+	 if (tent.middle == 0.0 ||
+	     tent.minimum > tent.middle || tent.middle > tent.maximum ||
+	     (tent.minimum < 0.0 && tent.maximum > 0.0))
+	   continue;
+	 int pos = add_grid_axis (_.first);
+	 if (pos < 0) return false;
+	 hb_pair_t<double, double> *input;
+	 double blo = -1.0, bhi = +1.0;
+	 if (box.has (_.first, &input)) { blo = input->first; bhi = input->second; }
```
qxliu76: this piece of code is duplicate with the code in `add_grid_axis`, I think we can just use `blo = points[0], bhi=points[1]` ?

src/hb-subset-plan-var.cc
```diff
@@ -136,45 +136,178 @@
...
+       if (grid_size > MAX_GRID) { exact = false; break; }
+     }
+
+     if (exact)
+     {
+       unsigned target_pos = 0; /* the target axis was added first */
+       hb_vector_t<unsigned> odometer;
+       if (!odometer.resize (grid_tags.length)) return false;
+       bool first = true;
+       for (;;)
```
qxliu76: I don't have a strong opinion here, but I'd prefer to use `while(true)`.

src/hb-ot-var-avar-table.hh
behdad: Done in 5202b1d779. `_avar2_map_new_mapping()` now takes from/to offsets, and the inverse is a one-line call with `(1, 0)`, matching the `SegmentMaps::unmap_float()` pattern.

src/hb-subset-plan-var.cc
behdad: Done in 5202b1d779. Removed `grid_tags`; new positions and the odometer size now come from `grid_pos.get_population()`.

src/hb-subset-plan-var.cc
behdad: Done in 5202b1d779. `grid_pos` and `grid_points` now live outside the per-axis loop and are reset for allocation reuse. While auditing this path, I also made the grid-size cutoff overflow-safe so a crafted axis product cannot wrap past the fallback guard.

src/hb-subset-plan-var.cc
behdad: Done in 5202b1d779. The code now reuses `points[0]` and `points[1]` as `blo` and `bhi`, avoiding the duplicate box lookup.

src/hb-subset-plan-var.cc
behdad: Done in 5202b1d779. Switched the odometer loop to `while (true)`.

behdad:
@qxliu76 PTAL.

qxliu76 (approved):

qxliu76:
LGTM, thanks!

PR #6141, opened 2026-08-07
Sanitize values before emitting them as code in src/gen-tag-table.py.

garretrieger:
To prevent malicious input files from being able to insert arbitrary code into the generated output.


behdad:
So you regenerated tag tables with this PR and it produces the same bytes?

garretrieger:
> So you regenerated tag tables with this PR and it produces the same bytes?

I did with the latest version of the input files and there was some small diffs (due to changes in the input), I'll go find the exact versions that it was generated from and verify there's no change.

behdad:
Thanks. Please commit the small changes as well, if they re cosmetic onyl.

garretrieger:
> Thanks. Please commit the small changes as well, if they re cosmetic onyl.

Alright, added the updated version. Just some comments that are changed.

PR #6149, opened 2026-08-11
In text-rendering-tests/update.py escape test names sourced from https://github.com/OpenType/fonttest before putting them into meson.build.

garretrieger:
Ran update.py and the output is identical to the currently checked in file.


behdad:
Thanks. Does it change the output?

garretrieger:
> Thanks. Does it change the output?

Output is exactly the same.

PR #6152, opened 2026-08-12
[subset] Handle invalid avar2 delta-set indices

behdad:
## Summary

- treat out-of-range avar2 delta-set indices as zero during subset planning and rewriting, matching runtime evaluation
- normalize invalid output mappings to `NO_VARIATION_INDEX`
- add an end-to-end regression for an empty non-NULL VarStore with an implicit index map

This ports the applicable invalid-index behavior identified in https://github.com/fonttools/fonttools/pull/4150. HarfBuzz already handles the other culling-bound cases there through conservative interval bounds, F2Dot14 padding, and symmetric endpoint clamping.

## Testing

- `meson test -C build` (263/263 passed)
- `meson test -C build --suite subset` (120/120 passed)
- `HB_TINY` amalgamation compile

### DIFF
--- a/src/hb-subset-plan-var.cc
+++ b/src/hb-subset-plan-var.cc
@@ -138,42 +138,38 @@
 
      double dmin = 0.0, dmax = 0.0;
      uint32_t varidx = varidx_map->map (i);
+     /* Runtime treats out-of-range delta-set indices as zero. */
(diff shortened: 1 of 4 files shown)


src/hb-subset-plan-var.cc
qxliu76: invalid` var_idx` set to` HB_OT_LAYOUT_NO_VARIATIONS_INDEX` already, I think we can remove these checks.

PR #6156, opened 2026-08-13
[graphite2] Fix cluster values when the buffer was reversed

behdad:
## Summary

- carry forward the Graphite2 reversed-buffer cluster fix and regression data from #6051
- preserve each Graphite glyph's covered input cluster while building scratch data
- assign those clusters to the HarfBuzz output glyphs, then call `buffer->merge_clusters()` over each output glyph range
- keep the Graphite layout-group index separate from `info.cluster` and use it to emit advances, so legal repeated or gapped client cluster labels cannot suppress a later Graphite group's advance
- validate Graphite slot association bounds before indexing the input buffer
- preserve the historical behavior at non-monotone cluster levels

This supersedes #6051 while preserving the original contributor's commits and authorship, and addresses the ordering requested in https://github.com/harfbuzz/harfbuzz/pull/6051#discussion_r3717886480.

## Root cause and regression

HarfBuzz cluster values are opaque client labels, whereas Graphite layout groups determine where group advances are emitted. They can coincide, but they are not the same identity. After `merge_clusters()` coalesced two Graphite groups carrying repeated input cluster values, the positioning loop used `info.cluster` as its group boundary and skipped the later group's advance.

The API regression shapes U+0660 U+0600 U+0600 with input clusters 0,3,3 in RTL at cluster level 0. Before the fix, the output clusters were correctly merged to 0 but total x advance fell from 3200 to 1269. The new separate group index preserves both the merged cluster semantics and the complete 3200 advance.

## Testing

- `meson setup build --reconfigure -Dgraphite2=enabled`
- `meson test -C build test-graphite2 --print-errorlogs` (1/1 passed)
- `meson test -C build --suite shape --print-errorlogs` (3/3 passed; 7,554 subtests)
- `meson test -C build --print-errorlogs` (253/253 passed)
- original manual Graphite2 repro at cluster levels 0-3; levels 0/1 produce `[0,0]`, levels 2/3 preserve `[1,1]`


behdad:
On top of https://github.com/harfbuzz/harfbuzz/pull/6051; codex work.

behdad:
Assessment: this PR has the right overall model for HarfBuzz clusters, but it currently conflates the merged HarfBuzz cluster value with the Graphite layout group used to apply advances. HarfBuzz cluster labels are opaque and may already be repeated or gapped. If merge_clusters() coalesces two Graphite layout groups into one HarfBuzz cluster, the positioning loop sees no cluster transition and drops the later group advance. A concrete case with the fixture font is U+0660 U+0600 U+0600, input clusters 0,3,3, RTL, cluster level 0: the current PR returns total x advance 1269 instead of 3200. PR #6051 avoids that loss but does not correctly extend pre-existing HarfBuzz clusters, so it is not the general fix either.

The way forward is to preserve the Graphite layout-group identity separately from info.cluster, use that identity to drive positioning/advance emission, and reserve info.cluster for HarfBuzz cluster semantics. I am preparing that change plus a regression test with repeated custom cluster values and will report full test results here.

behdad:
Implemented the assessment in commit 42427d6b2ea1d996fbbbeb48d3207666b0c49c57.

The positioning loop now keys on a separately preserved Graphite layout-group index instead of the merged HarfBuzz cluster label. The repeated-cluster regression that failed with total x advance 1269 now returns the correct 3200 while retaining output cluster 0 for every glyph. I also added bounds validation for Graphite slot source associations.

Validation:
- targeted Graphite2 API regression: 1/1 passed
- shape suite: 3/3 passed, 7,554 subtests
- full configured suite: 247/247 passed

behdad:
Synced the branch with current `main` in 1e048ddd900390cf708fe76540bcf4b8176f219c. The only conflict was the newer Graphite2 saturating-position work; the resolution preserves both that arithmetic and this PR's separate layout-group identity.

Post-sync validation:
- targeted Graphite2 API regression: 1/1 passed
- shape suite: 3/3 passed, 7,554 subtests
- full configured suite: 253/253 passed

behdad:
Squashed as requested. The PR branch is now one commit directly on current `main`: 4e784b6fc1c1ee1b226d0d32138488d0a31195e9.

The squashed commit has the same tree as the fully validated merge result. I reran the targeted Graphite2 API regression (1/1) and shape suites (3/3, 7,554 subtests) after squashing; the identical tree passed the full configured suite (253/253). The PR is mergeable, and CI has restarted on the clean commit.

PR #6163, opened 2026-08-16
[JSTF] Expose extender glyphs via hb_ot_layout_get_extender_glyphs

hjanuschka:
The JSTF table is already parsed but never exposed to clients. Extender glyphs are the most widely populated part of it and are useful for cursive-script justification implementations that want to prefer a font-provided extender over the plain U+0640 TATWEEL fallback.

## What

Adds a new accessor:

```c
unsigned
hb_ot_layout_get_extender_glyphs (hb_face_t      *face,
                                  hb_script_t     script,
                                  hb_language_t   language,
                                  unsigned        start_offset,
                                  unsigned       *glyphs_count /* IN/OUT */,
                                  hb_codepoint_t *glyphs       /* OUT */);
```

Follows the standard iterator idiom used by `hb_ot_layout_get_attach_points` and friends: pass `NULL` for size probing, use the in/out `count` for pagination.

Also enables the `JSTF` table accelerator, which was compiled but not registered as a core table (`//HB_OT_CORE_TABLE (OT, JSTF)` in `hb-ot-face-table-list.hh`).

## Scope

`hb_script_t` + `hb_language_t` are taken as inputs (like the `_get_baseline2` / `_get_font_extents2` accessors) and converted internally to the `hb_tag_t` pair the layout tables use, via `choose_base_tags`.

Only `script_tag` is actually consulted for extenders: per the OT spec, extender glyphs live directly on `JstfScript` and do not vary by language. `JstfLangSys` carries per-language justification priorities, but exposing those is both harder and less common among fonts than the (already rare) JSTF table itself; that can be a separate accessor later.

## Ecosystem

JSTF adoption in the wild is sparse. An audit of a typical Windows font collection found the table only in Arial and Times variants -- and those two ship the **same** table (their Arabic glyphs are visually identical; see the discussion in [crbug 41136866#c5](https://issues.chromium.org/issues/41136866#comment5)). Third-party font uptake is expected to be similar or lower. Both fonts in the audit only populated extender-glyph info, nothing else the table can carry.

That's exactly the argument for a small, well-scoped accessor rather than a broader JSTF API: expose the one useful piece, keep the door open for more if usage ever grows.

## Tests

`meson test -C build`:
- New `test/api/test-ot-layout.c :: test_ot_layout_get_extender_glyphs`
- New minimal fixture `test/api/fonts/jstf.ttf` (740 bytes; JSTF table with 5 Arabic extender glyphs)
- Exercises both the size-probe path (`NULL` count/array) and paginated retrieval
- All 71 api subtests pass locally

## Consumer

No consumer in HarfBuzz clients yet. A follow-up in Chromium's `text-justify: kashida` work would use this to prefer font-provided extenders over the plain tatweel fallback when a JSTF table is present, falling back cleanly (as today) for the ~99.99% of fonts without one.


behdad:
Generally looks good to me. I was under the impression that we do this already. I let Khaled review though.

behdad:
@khaledhosny Can you review this please?

src/hb-ot-layout.cc
```diff
@@ -2782,6 +2782,47 @@ hb_ot_layout_lookup_get_optical_bound (hb_font_t      *font,
   }
   return ret;
 }
+
+/**
+ * hb_ot_layout_get_extender_glyphs:
+ * @face: The #hb_face_t to work on
+ * @script: a script tag, usually HB_SCRIPT_ARABIC for Arabic script.
+ * @language: a language tag, doesn't matter right now given how JSTF is formed.
```
khaledhosny:
```suggestion
 * @language: a language tag, currently unused.
```

src/hb-ot-layout.cc
```diff
@@ -2782,6 +2782,47 @@ hb_ot_layout_lookup_get_optical_bound (hb_font_t      *font,
   }
   return ret;
 }
+
+/**
+ * hb_ot_layout_get_extender_glyphs:
+ * @face: The #hb_face_t to work on
+ * @script: a script tag, usually HB_SCRIPT_ARABIC for Arabic script.
```
khaledhosny:
```suggestion
 * @script: a script tag to get extender glyphs for.
```

src/hb-ot-layout.cc
```diff
@@ -2782,6 +2782,47 @@ hb_ot_layout_lookup_get_optical_bound (hb_font_t      *font,
...
+ * @glyphs: (out caller-allocates) (array length=glyphs_count) (nullable): A glyphs buffer.
+ *          Extender glyphs associated with the script tag.
+ *
+ * Fetches extender glyphs for a given script and language. Extender glyphs
+ * are used in certain scripts to extend the length of a word for
+ * justification purposes, such as variants of tatweel (U+0640) for the
+ * Arabic script.
+ *
+ * The default tatweel glyph is not returned as a fallback; the client should
+ * handle it, which is the only extender glyph needed for most fonts.
```
khaledhosny: This needs to be re-worded. We don’t usually do such fallback unless the function has an explicit fallback promise (i.e. ends with `_with_fallback`).

src/hb-ot-layout.cc
```diff
@@ -2782,6 +2782,47 @@ hb_ot_layout_lookup_get_optical_bound (hb_font_t      *font,
...
+HB_EXTERN unsigned
+hb_ot_layout_get_extender_glyphs (hb_face_t      *face,
+				  hb_script_t     script,
+				  hb_language_t   language,
+				  unsigned        start_offset,
+				  unsigned       *glyphs_count /* IN/OUT.  May be NULL */,
+				  hb_codepoint_t *glyphs       /* OUT.     May be NULL */)
+{
+  hb_tag_t script_tag, language_tag;
+  choose_base_tags (script, language, &script_tag, &language_tag);
```
khaledhosny: `choose_base_tags()` is wrapped inside `#ifndef HB_NO_BASE ... #endif`, so this code will fail in builds that define `HB_NO_BASE`. This function shiuld be extracted out and renamed that it is not specific to `BASE` table code.

src/hb-ot-layout.cc
```diff
@@ -2782,6 +2782,47 @@ hb_ot_layout_lookup_get_optical_bound (hb_font_t      *font,
...
+  // Only script_tag is used. Extender glyphs live directly on JstfScript
+  // and do not vary by language per the OT spec; JstfLangSys carries
+  // per-language justification priorities, but exposing those is both
+  // harder and less common among fonts than the (already rare) JSTF table
+  // itself.
```
khaledhosny: I don’t think we need all this comment. A short note that extender glyphs are per-script in JSTF table should be enough.

src/hb-ot-layout.cc
hjanuschka: Reworded to state that only glyphs listed in the JSTF table are returned, without implying a fallback contract.

khaledhosny:
The code looks OK, but with the nits above.

However, if the intent is to support _kashida_ justification, then I question the usefulness of this API. As noted there are only two fonts with `JSTF` table, and all they have is a lost of extender glyphs (that are simple rectangles, so they provide no much value over repeating the _tatweel_ glyph).

Implementing _kashida_ justification needs to resolve three issues, and `JSTF` table does not help with any of them:
1. Where to insert _kashida_ in a word. There are aesthetic and orthographic considerations that govern where a _kashida_ can be inserted in a word. `JSTF` table has no way of conveying such information. As it happens, I recently released [a library that attempts to answer this question](https://github.com/aliftype/raqim-kashida) (there is also a [web demo](https://aliftype.com/raqim-kashida/english/)).
2. Where to insert a _kashida_ in a line (the _kashidas_ can be distributed in a several ways to maintain balance and to avoid clustering them in certain parts of the line). There also can’t be two _kashidas_ in the same word (or rather joined run). How many _kashidas_ can be inserted in each position. Etc.
3. Re-shape the text after inserting the _kashida_ so that the font can *respond* to the inserted character as some fonts have curved _kashida_, so multiple ones need to form a ligature. It would seem extender glyphs would help here by having multiple sized curved glyphs, but that is not enough, since glyphs before and after the _kashida_ need to change to account for the length and the curvature.

I’d say forget about `JSTF`, it is a dead-end, and try to figure out these three questions (I have a justification engine that does this and a bit more, but I don’t know when it will be ready for a public release).

hjanuschka:
Thanks. I should clarify the scope, because this PR is not an attempt to build kashida justification on JSTF. It came out of an ongoing Chromium implementation (attempt):

- https://issues.chromium.org/issues/41431127
- https://chromium-review.googlesource.com/c/chromium/src/+/7957320
- https://static.januschka.com/i-41431127/index.html

That WIP already tackles the three problems independently of JSTF:

1. It chooses one insertion point per word using Unicode joining data and the IE/DirectWrite priority scheme, including ligature and mark handling.
2. It distributes the line budget across words, keeps one elongation point per word, and sends any remainder to normal word spacing.
3. It inserts U+0640 only into the shaping input, reshapes the run, remaps the result to the original text, and has a font-safety gate for faces where tatweel insertion is unsuitable.

This JSTF accessor was only a small, separable step for a possible later experiment with font-provided extender glyphs. Chromium does not consume it yet, and the implementation does not depend on it. 


Your point that the extenders in the few existing JSTF fonts add little over repeated tatweel may mean that this optional experiment is not worthwhile. 

i have no feelings, happy to close, its maintainer call :)

khaledhosny:
> * https://static.januschka.com/i-41431127/index.html

Reading the text here. Re-`safe-to-insert-tatweel`: this is not what it is for, it is a an escape hatch for implementations that insert _kashida_ after shaping (nearly all of them today) as it tells you whether inserting _kashida_ at this index would separate two glyphs that should not be separated (e.g. they were contextually substituted). If you insert _kashida_ in the text and reshape then you should not use it at all. To skip fonts that do not support _kashida_, check of the font has tatweel glyph at all and if it has a +ve advance width.

Also note that MS rule are in adequate for many fonts, and even in the example you show it inserts a _kashida_ in `بي` which breaks its correct shape in Amiri. In [my blog post](https://aliftype.com/blog/introducing-raqim-kashida/english) I try to explain why a single set of _kashida_ insertion rules is not enough.

> i have no feelings, happy to close, its maintainer call :)

I’ll leave that to Behdad.

behdad:
> > i have no feelings, happy to close, its maintainer call :)
> 
> I’ll leave that to Behdad.

If you're currently just experimenting, I'd wait. But if/when you're gonna need it in Chromium, I think it belongs to HB indeed.

PR #6170, opened 2026-08-17
Fix arithmetic overflows in IUP handling.

garretrieger:
Fixes https://oss-fuzz.com/testcase-detail/4688270404157440.


behdad:
Is this using double instead of int for index values? Can you please explain what's overflowing?

garretrieger:
> Is this using double instead of int for index values? Can you please explain what's overflowing?

I went with double's since the values are operated on and compared to with tolerance which is a double.

PR #6176, opened 2026-08-18
[subset] don't return early on null offset in BASE table

qxliu76:
I noticed this issue while fixing empty VariationDevice in Skera(https://github.com/googlefonts/fontations/pull/2057). `serialize_subset() `call would return false if the offset is null. Spec says these offsets "may be NULL".  
No testcase for this issue.


behdad:
LGTM. Thanks.

qxliu76:
Seems all tests are clean. Is it ready to merge?

PR #6180, opened 2026-08-19
Guard against underflow in calculating val.last_arg_offset in top dict processing.

garretrieger:


behdad:
I think this is unnecessary as a correctness/security fix. The unsigned wrap is reachable, but the wrapped value is dead.

Proof:

1. `last_arg_offset` is read only in `cff1_top_dict_op_serializer_t::serialize()`, and only for `OpCode_ROS`.
2. Processing `ROS` pops three arguments. If fewer exist, `argStack` enters the error state and `process_op()` returns at the `env.in_error()` check before `dictval.add_op()`. Thus an argument-less/undersupplied `ROS` is never recorded.
3. If those three arguments do exist, parsing an argument has set `env.last_offset` after its first byte. That byte is necessarily after `dictval.opStart`, so `env.last_offset > dictval.opStart` holds for every recorded `ROS`.
4. Argument-less operators such as `FontBBox` can record `UINT_MAX` here, but their serializers never inspect `last_arg_offset`.
5. The unsigned wrap itself is defined C++ behavior, not UB.

I also checked the interpreter directly on current main:

```text
bare FontBBox:      ok=1 values=1 last_arg_offset=4294967295
bare ROS:           ok=0 values=0
three-argument ROS: ok=1 values=1 last_arg_offset=2
```

So this patch only changes unused metadata for malformed non-`ROS` entries; it cannot affect a successfully parsed `ROS` or subset output. If the goal is to satisfy a particular analyzer rule, calculating the field only for `ROS` and asserting the invariant would express the actual contract better than clamping it to zero.

behdad:
I was skeptical that this is actually needed, especially since unsigned underflow is not UB in C. So I asked codex to prove it unnecessary.

behdad:
Simplified it. @garretrieger PTAL.

garretrieger (commented):
lgtm

PR #6182, opened 2026-08-19
[depend] Deduplicate ligature dependency data

behdad:
## Summary

- walk the already-deduplicated ligature component set instead of revisiting the raw component array
- intern equivalent ligature component sets through the existing content-based dependency-set map
- preserve the unused-set reuse path by discarding only newly interned sets

## Root cause

Dependency extraction built a complete component set for every ligature, but then ignored that set for traversal and allocated a separate copy for every ligature record. Malformed GSUB data could therefore trigger millions of redundant edge lookups and hundreds of thousands of equivalent set allocations.

## Impact

On `clusterfuzz-testcase-minimized-hb-subset-fuzzer-5845846876356608`:

- runtime: about 0.91s to 0.24s
- peak RSS: about 292MB to 80MB
- stored component sets: 210,702 to 76
- retained edges: 2,652,906 to 396,546

The dependency graph remains semantically equivalent: the API models ligature requirements as sets, so identical component sets can share one index and repeated component visits do not add information.

## Validation

- `meson test -C build --print-errorlogs`
- 270/270 tests passed, including dependency API tests, closure parity, failing-allocation fuzzing, and all dependency-fuzzer chunks

### DIFF
--- a/src/OT/Layout/GSUB/Ligature.hh
+++ b/src/OT/Layout/GSUB/Ligature.hh
@@ -50,27 +50,26 @@ struct Ligature
       return;
     }
 
-    hb_codepoint_t ligset_idx = c->depend_data->new_ligature_set(complete_ligset);
+    bool ligset_created;
+    hb_codepoint_t ligset_idx = c->depend_data->find_or_create_set (complete_ligset,
+                                                                    &ligset_created);
     if (unlikely (ligset_idx == HB_CODEPOINT_INVALID))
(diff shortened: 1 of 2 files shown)


behdad:
cc @skef

skef:
Yes, seems sensible to me

behdad:
> Yes, seems sensible to me

Thanks for checking.

PR #6184, opened 2026-08-19
[cff] Reject non-SEAC charstrings from their suffix

behdad:
## What changed

CFF1 SEAC dependency discovery now examines the charstring suffix before invoking the Type 2 interpreter.

The fast path requires a final `endchar`, recognizes every Type 2 number encoding that could end immediately before it, and retains the full interpreter for ambiguous operands and `callsubr`/`callgsubr`. Other suffixes cannot carry SEAC operands and return early.

## Why

Dependency discovery previously interpreted every CFF1 glyph charstring to inspect the operand stack at `endchar`. Most glyphs end in a stack-clearing path operator, so this needlessly expanded charstrings and subroutines. On malformed fuzz fonts, the per-glyph 200,000-operation limit multiplied across thousands of glyphs.

The same internal SEAC lookup is used during ordinary CFF subsetting, so it benefits as well.

## Performance

- depend-fuzzer chunk 4: 1.31s → 0.04s
- 100 SourceHan dependency scans: 1.38s → 0.36s
- 50 full SourceHan subsets: 2.56s → 2.05s

## Validation

- `meson compile -C build`
- `meson test -C build --print-errorlogs --no-rebuild`
- All 270 tests pass

### DIFF
--- a/src/hb-ot-cff1-table.cc
+++ b/src/hb-ot-cff1-table.cc
@@ -595,12 +595,41 @@ struct cff1_cs_opset_seac_t : cff1_cs_opset_t<cff1_cs_opset_seac_t, get_seac_par
   }
 };
 
+static bool
+_cff1_charstring_may_end_in_seac (const hb_ubytes_t &str)
+{
+  if (unlikely (!str.length || str[str.length - 1] != OpCode_endchar))
+    return false;
+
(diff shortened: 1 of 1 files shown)


behdad:
cc @skef

PR #6185, opened 2026-08-19
[depend] Speed up contextual dependency collection

behdad:
## What changed

- add `--compile` for construction-only profiling, `-q` / `--quiet` for retrieval without printing, and `--no-glyph-names` for numeric output
- memoize completed contextual GSUB lookup-recursion states per top-level lookup
- include the active glyph set, ordered recursion ancestry, and context set in the cache key; omit edge flags because they affect neither traversal nor edge identity
- encode the common root and single-active-glyph states directly in the memoization key
- intern ordered recursion paths directly from parent-path and lookup IDs instead of hashing and copying ancestry sets
- cache position-specific contextual edge labels by rule, lookup record, and stable active-glyph identity
- skip side-effect-free input-position expansion when every record label is already cached
- skip chained-context backtrack, lookahead, and input-position preparation when a nonempty rule's record labels are already cached
- reuse context-set indices resolved by the all-cached precheck instead of probing the cache twice
- retain content-interned active-glyph identities across top-level lookups
- combine edge deduplication lookup and insertion, then store exact edge locations in a compact dedicated table
- cache contextual position glyph sets, keep the bounded covered-position bitmap inline, and recycle scratch sets and vectors by active recursion depth
- retain active-glyph set storage and reuse set-intersection compaction workspace
- store lookup feature tags in one compact owned vector with per-lookup offsets, sorting and deduplicating once after collection
- add an inversion-aware exact set-intersection helper and use it to avoid allocation-heavy temporary set subtraction
- expose the helper as the public `hb_set_intersects()` API
- consume stable `hb_set_t` iteration 64 bitmap bits at a time while preserving scalar and inverted-set behavior
- compute iterator length lazily when a consumer actually requests it
- intersect coverage-format-2 ranges and insert `hb_set_t` results one 64-bit bitmap word at a time
- remove unused contextual stack bookkeeping
- run each Noto Nastaliq dependency closure-parity fixture as its own named one-font test

## Why

Complex contextual lookup graphs can reach the same completed subgraph through many parent paths. The existing active-stack set prevents cycles but forgets states after unwinding, so Noto Nastaliq repeatedly expands equivalent recursion states. Memoization removes that repeated traversal.

After memoization, retained scratch storage, compact lookup-feature storage, bitmap-word set operations, and compact edge deduplication reduce the remaining allocation and bookkeeping cost. Contextual rules also rebuilt the same position-specific edge label on nearly every visit; caching that label by its exact stable inputs avoids repeated set scans, intersections, and content hashing without changing traversal or label semantics.

## Performance

On the full 529,108-byte `NotoNastaliqUrdu-Regular.ttf`:

- old implementation with full graph output: 7:44.77 wall, 462.58s user
- new construction path, measured over 30 runs: 103.53ms task-clock, 105.64ms elapsed, and 1.208 billion instructions
- new implementation with full graph output, measured over 10 runs: 575.80ms task-clock and 578.30ms elapsed
- new peak RSS: about 30MB

Construction is about 4,400x faster than the original end-to-end baseline. The apples-to-apples full-output run is about 804x faster. The full formatted graph is 150,835,628 bytes. Old and new stdout are byte-for-byte identical with SHA-256 `cb59827fa48f3efa96e3275b249b671aa5546779d39bfc6c2dff49075bc3d2dc`; stderr is empty for both.

The follow-up storage changes reduced allocation/reallocation calls from 5.26 million to 95,308 and cumulative allocated bytes from 480MB to about 51.7MB. Context-label caching, lazy input expansion, cached chained-context preparation, and flag-independent memoization reduced construction from about 209ms and 2.77 billion instructions to about 106ms and 1.21 billion instructions. From the first memoized implementation, construction fell from 374ms and 4.71 billion instructions to about 106ms and 1.21 billion instructions.

The internal 64-bit set iterator provides a smaller general benefit without adding public API. Ten-iteration full-Unicode subset runs improved by about 1.0% on Noto Nastaliq, 1.4% on Amiri, 1.9% on Source Serif Variable, and 2.7% on Noto COLRv1. A trial conversion of very sparse contextual sets regressed and was discarded.

## Validation

- full Meson suite: 272/272 passed
- public and internal exact set-intersection coverage for regular and inverted pairs
- word-at-a-time set insertion coverage for regular and inverted sets
- all dependency fuzzer chunks passed
- Clang 18 ASan/UBSan dependency fuzzer chunks: 9/9 passed after flattening lookup-feature storage
- `depend-closure-parity-noto-nastaliq-bold` and `depend-closure-parity-noto-nastaliq-regular` each pass as independent one-font tests
- remaining graph-vs-legacy closure fixtures pass in two balanced chunks
- exact old-vs-new graph output across 513 fuzz fonts, 99 subset fixtures that completed with the old implementation, and representative production fonts
- byte-identical graph output for Noto Nastaliq, Noto Sans Duployan, Gulzar, IranNastaliq, Noto-COLRv1, Amiri, and SourceSerifVariable-Roman
- byte-identical full-Unicode subset output before/after the iterator change for Noto Nastaliq, Amiri, Source Serif Variable, and Noto COLRv1
- an independent adversarial audit found the memoization key complete and no successful-construction correctness counterexample; OOM allocation timing changes but failure remains safe

### DIFF
--- a/docs/harfbuzz-sections.txt
+++ b/docs/harfbuzz-sections.txt
@@ -872,6 +872,7 @@ hb_set_symmetric_difference
 hb_set_invert
 hb_set_is_inverted
 hb_set_is_equal
+hb_set_intersects
 hb_set_is_subset
 hb_set_next
 hb_set_next_range
--- a/src/OT/Layout/Common/CoverageFormat2.hh
+++ b/src/OT/Layout/Common/CoverageFormat2.hh
@@ -151,9 +151,22 @@ struct CoverageFormat2_4
       if (unlikely (range.first < last))
         break;
       last = range.last;
-      for (hb_codepoint_t g = range.first - 1;
-	   glyphs.next (&g) && g <= last;)
-	intersect_glyphs << g;
+      hb_codepoint_t cursor = range.first & -hb_bit_page_t::ELT_BITS;
+      cursor = cursor ? cursor - 1 : HB_SET_VALUE_INVALID;
+      uint64_t bits;
+      while (glyphs.next_bits (&cursor, &bits) && cursor <= last)
+      {
+	if (range.first > cursor)
+	  bits &= UINT64_MAX << (range.first - cursor);
+	unsigned int end = last - cursor;
+	if (end < hb_bit_page_t::ELT_BITS - 1)
+	  bits &= (uint64_t (1) << (end + 1)) - 1;
+	while (bits)
+	{
+	  intersect_glyphs << cursor + hb_ctz (bits);
+	  bits &= bits - 1;
+	}
+      }
     }
   }
 
--- a/src/OT/Layout/Common/CoverageFormat2.hh
+++ b/src/OT/Layout/Common/CoverageFormat2.hh
@@ -139,6 +139,19 @@ struct CoverageFormat2_4
     return false;
   }
 
+  static void add_intersected_bits (hb_set_t &set, hb_codepoint_t base, uint64_t bits)
+  { set.add_bits (base, bits); }
+
+  template <typename IterableOut>
+  static void add_intersected_bits (IterableOut&& out, hb_codepoint_t base, uint64_t bits)
+  {
+    while (bits)
+    {
+      out << base + hb_ctz (bits);
+      bits &= bits - 1;
+    }
+  }
+
   template <typename IterableOut,
 	    hb_requires (hb_is_sink_of (IterableOut, hb_codepoint_t))>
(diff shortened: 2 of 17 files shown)


behdad:
cc @skef

behdad:
While at it, this addes `hb_set_intersects()` (which was needed in the patch). Lemme know if you prefer that separated @khaledhosny. Thanks.

garretrieger (approved):

behdad:
Merging this since it touches places I'm working on improving as a separate PR. Please review regardless. Thanks.

PR #6188, opened 2026-08-20
[GDEF] Cache mark-filtering sets as bit-pages

behdad:
Speed up shaping of fonts that make heavy use of `UseMarkFilteringSet` lookups, Noto Sans Duployan being the extreme case (403 lookups, 344 of them mark-filtering; 4388 ChainContextFormat3 subtables).

Mark-filtering-set checks ran on every skipping-iterator step as a set-digest pre-filter plus a Coverage binary search in the GDEF table. The GDEF accelerator now caches each mark set: a set whose glyphs span fewer than 512 gids (102/123 sets in Duployan, 16/30 in Gulzar) is flattened into a single `hb_bit_page_t` biased at its first glyph — exact O(1) answers, no table access. Wider sets (e.g. possible with 24-bit gids) keep a set-digest in the same union as a pre-filter ahead of the out-of-line table fallback; either way one fixed 68-byte entry per set.

Measured with interleaved A/B against main (`perf stat -e instructions,task-clock`, `hb-shape -O '' --text-file perf/texts/duployan.txt`):

| Font | Instructions | Wall |
|---|---|---|
| Noto Sans Duployan / duployan.txt | **−18.0%** | **−18.5%** (5.25s → 4.27s) |
| Amiri, Gulzar, Nastaliq, Devanagari, Roboto | flat (±0.2%) | flat within alignment noise |

Also adds an `HB_NEVER_INLINE` macro, used to keep the table fallback out of the inlined hot path.

All 269 meson tests pass.

Profiling notes: after this change ~70% of Duployan shaping time remains in `skipping_iterator_t::prev`/`next` step count itself (backtrack walks over ignorable runs, re-done per subtable); follow-up work planned separately.

### DIFF
--- a/src/OT/Layout/GDEF/GDEF.hh
+++ b/src/OT/Layout/GDEF/GDEF.hh
@@ -978,7 +978,39 @@ struct GDEF
       }
 
 #ifndef HB_NO_GDEF_CACHE
-      table->get_mark_glyph_sets ().collect_coverage (mark_glyph_sets);
+      /* Flatten each mark set whose glyphs span fewer than PAGE_BITS
+       * gids into a single bit-page biased at its first glyph, for
+       * O(1) exact answers.  Wider sets, or on allocation failure,
+       * fall back to the table. */
+      {
+	hb_vector_t<hb_set_t> sets;
+	table->get_mark_glyph_sets ().collect_coverage (sets);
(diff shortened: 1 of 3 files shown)


dfrg (approved):
Nice one. LGTM

PR #6194, opened 2026-08-21
[ot-layout] Skip the per-subtable digest retest for single-subtable lookups

behdad:
Found while investigating why HarfRust's OT apply loop now measures cheaper than ours: its driver tests only the lookup-level digest, while our `hb_applicable_t::apply` retests the per-subtable digest after `apply_forward` already passed the accelerator digest. Since the accelerator digest is the union of the subtable digests, the retest is a guaranteed-true evaluation whenever the lookup has a single subtable — which is the overwhelming majority of lookups in practice.

This adds digest-free entry points to `hb_applicable_t` and dispatches `count == 1` lookups straight to the apply function. Multi-subtable lookups keep the per-subtable tests, where they do real filtering. Purely a work reduction — the digest is an optimization filter, so behavior is unchanged by construction; verified byte-identical outputs on Roboto, NotoNastaliqUrdu, Amiri, and NotoSansDevanagari over the benchmark texts, and `test-shape`/`test-shape-plan`/`test-aat-layout` pass.

Instruction counts (single pass of the benchmark texts): **Roboto −2.3%, NotoNastaliqUrdu −3.6%**. Wall clock on my machine is within noise (the removed tests pipeline well); the win is real work removed either way.

### DIFF
--- a/src/hb-ot-layout-gsubgpos.hh
+++ b/src/hb-ot-layout-gsubgpos.hh
@@ -1453,6 +1453,10 @@ struct hb_accelerate_subtables_context_t :
     {
       return digest.may_have (c->buffer->cur().codepoint) && apply_func (obj, c, nullptr);
     }
+    bool apply_no_digest (hb_ot_apply_context_t *c) const
+    {
+      return apply_func (obj, c, nullptr);
+    }
 #else
     bool apply (hb_ot_apply_context_t *c) const
     {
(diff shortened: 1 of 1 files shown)


behdad:
*[Written by Claude Code, posting from Behdad's account.]*

Fixed the fuzzing crash: invalid subtables don't collect and leave their accelerator entries zeroed, relying on the empty digest to gatekeep the null apply funcs — the fast path skipped that gate. It now checks the function pointer instead (single-subtable path only). Local shape-fuzzer chunks and shape tests pass.

behdad:
*[Written by Claude Code, posting from Behdad's account.]*

Reworked per Behdad's suggestion: instead of a runtime null-func guard, the zeroed (uncollected) tail entries are now trimmed at accelerator create time — collection fills a prefix, so a tail `count--` loop removes exactly the invalid entries. The fast path needs no guard, the multi-subtable loop stops iterating dead entries, and the crash class is gone structurally. Fuzzer chunks + shape tests pass; slightly better numbers than the guard version (Roboto −2.3%, Nastaliq −3.7% instructions).

PR #6201, opened 2026-08-24
[harfrust] Store shape-plan data directly

behdad:
Store the HarfRust plan directly on `hb_shape_plan_t`, like the existing OT plan data, instead of putting it in generic user data.

The data is created lazily because construction needs the font instance, published atomically for shared plans, and destroyed with the shape plan. This removes one mutex-backed user-data lookup per shape.

There is no HarfRust-specific field on `hb_buffer_t`; the reusable Rust buffer remains owned by generic user data.

Memory cost: 8 bytes per shape plan when HarfRust is enabled, with one fewer user-data allocation for plans.

Measured against current `main` with the clang release build and `hb-shape -O "" -n 10 --shaper harfrust`:

| Font / text | Instructions | Cycles |
|---|---:|---:|
| Roboto / en-prince | -0.11% | -0.29% |
| Roboto / en-words | -0.73% | -1.75% |

Cycle results are noisier than retired instructions.

Validation:

- `test-shape-harfrust`: 4 subtests pass
- HarfRust and OT output hashes match on the benchmark corpus
- `git diff --check`

### DIFF
--- a/src/hb-harfrust.cc
+++ b/src/hb-harfrust.cc
@@ -100,7 +100,6 @@ _hb_harfrust_shape_plan_create_rs (const void *font_data,
 				   const hb_feature_t *features,
 				   unsigned int        num_features);
 
-
 extern "C" void
 _hb_harfrust_shape_plan_destroy_rs (void *data);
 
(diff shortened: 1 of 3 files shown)


behdad:
This is not an option for buffer, but for other shape plans we do. Lemme send back to agent.

behdad:
Updated accordingly: there is no buffer field now. The HarfRust plan is stored and owned directly by `hb_shape_plan_t`; the Rust buffer stays in generic user data.

behdad:
Fixed the CI symbol-visibility regression by marking the out-of-line internal shape-plan destructor HB_INTERNAL, matching the existing internal destructor pattern. Force-pushed; the local check-symbols test now passes.

PR #6202, opened 2026-08-24
[harfrust] Shape the cached Rust buffer in place

behdad:
Depends on harfbuzz/harfrust#448.

Borrow the reusable Rust UnicodeBuffer allocation in place for the duration of shaping. This removes the per-shape Box::from_raw / move-out / move-back / Box::into_raw sequence for the 304-byte buffer. A scoped result guard from HarfRust clears the buffer on drop, including post-shape early returns; the pre-shape import failure clears it explicitly.

The existing public shape() API remains unchanged. This change is independent of #6201 and is based directly on main after #6200.

Measured with the clang release build and hb-shape -O "" -N count, changing only the old boxed bridge versus this in-place bridge:

| benchmark | instructions |
|---|---:|
| Menlo / en-prince | -0.56% |
| Lucida Grande / en-prince | -0.41% |
| Geeza Pro / fa-prince | -0.15% |
| Devanagari Sangam / hi-words | -0.64% |
| Roboto / en-prince | -0.33% |
| Roboto / en-words | -2.19% |
| Source Serif Variable / react-dom | -0.62% |
| Noto Sans Devanagari / hi-words | -1.02% |
| Amiri / fa-prince | -0.05% |
| Noto Nastaliq Urdu / fa-prince | -0.03% |
| Noto Nastaliq Urdu / fa-words | -0.31% |

Paired cycle measurements for Roboto were -0.51% on prince and -3.94% on words. Callgrind on Roboto words attributes about 301 fewer instructions per shape, including about 236 removed from the bridge function itself and about 65 from memcpy.

No new allocation or persistent memory is added.

Validation:

- HarfBuzz HarfRust API tests: 4 subtests passed
- all eleven benchmark outputs match the OT shaper
- HarfRust: 6,271 tests and clippy with warnings denied

### DIFF
--- a/src/hb-harfrust.cc
+++ b/src/hb-harfrust.cc
@@ -112,7 +112,7 @@ _hb_harfrust_shape_plan_destroy_rs (void *data);
 extern "C" hb_bool_t
 _hb_harfrust_shape_rs (const void         *font_data,
 		       const void         *rs_shape_plan,
-		       const void         *rs_buffer,
+		       void               *rs_buffer,
 		       hb_font_t          *font,
 		       hb_buffer_t        *buffer,
 		       const hb_codepoint_t *pre_context,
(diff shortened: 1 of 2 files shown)


behdad:
Closing with harfbuzz/harfrust#448. The in-place bridge depended on a temporary borrowed typestate API that is not worth adding ahead of the planned UnicodeBuffer/GlyphBuffer unification.

PR #6204, opened 2026-08-24
[depend] Bound context inputCount to HB_MAX_CONTEXT_LENGTH

Nussu06:
context_depend_recurse_lookups tracks covered sequence positions in a fixed hb_bit_page_t, but the dependency-graph path never bounded inputCount the way the shaping apply path does, so a crafted GSUB contextual rule with glyphCount past the page (512) plus a lookup record whose sequenceIndex lands in a higher page element than endIndex makes hb_bit_page_t::add_range memset a wrapped negative length past the stack. It is reachable from hb_subset_depend_from_face_or_fail and the depend fuzzer, where ASan reports a stack-buffer-overflow at hb-bit-page.hh:160. Cap inputCount at HB_MAX_CONTEXT_LENGTH before the page is used, matching the apply path, since longer rules never apply during shaping.


behdad:
Thanks. Can we also add protection to the bitset code?

Nussu06:
Done. Added a guard in hb_bit_page_t::add_range and del_range so the multi-element branch only runs when la < lb. When the two endpoints map to element pointers out of order the memset span would otherwise go negative and wrap, so now the page is left untouched instead. Fuzzing and set tests stay green.

PR #6209, opened 2026-08-25
[arabic] Enable Win1256 fallback on all platforms

behdad:
The Win1256 fallback is a static GSUB recipe selected by a narrow glyph-ID signature; it does not call a Windows API. Enable it outside Windows too, while retaining HB_NO_WIN1256 as an explicit opt-out.

Tested: meson test -C build; build with -DHB_NO_WIN1256

Assisted-by: Codex <codex@openai.com>

behdad: I don't quite remember why we disabled it in non-Windows systems to begin with. @khaledhosny any insights?

### DIFF
--- a/src/hb-ot-shaper-arabic-fallback.hh
+++ b/src/hb-ot-shaper-arabic-fallback.hh
@@ -231,11 +231,7 @@ struct arabic_fallback_plan_t
   OT::hb_ot_layout_lookup_accelerator_t *accel_array[ARABIC_FALLBACK_MAX_LOOKUPS];
 };
 
-#if defined(_WIN32) && !defined(HB_NO_WIN1256)
-#define HB_WITH_WIN1256
-#endif
-
-#ifdef HB_WITH_WIN1256
+#ifndef HB_NO_WIN1256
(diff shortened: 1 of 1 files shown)


khaledhosny:
I don’t remember either. I think the whole thing was something found/suggested by @jfkthame so maybe he has some insight.

behdad:
> I don’t remember either. I think the whole thing was something found/suggested by @jfkthame so maybe he has some insight.

Wasn't it libass or something? Or that was just the PUA stuff?

khaledhosny:
> > I don’t remember either. I think the whole thing was something found/suggested by @jfkthame so maybe he has some insight.
> 
> Wasn't it libass or something? Or that was just the PUA stuff?

Libass was the PUA stuff, this one is older IIRC.

behdad:
Mozilla requested it indeed: https://bugzilla.mozilla.org/show_bug.cgi?id=1045139
However, my memory serves me that you would need to get the Uniscribe nominal_glyph mapping for this to be detected. I don't think our font-funcs impls have the legacy code in there. At any rate, no harm enabling?

khaledhosny:
Sounds reasonable

PR #6210, opened 2026-08-25
[subset] Support VARC tables

behdad:
This adds VARC table subsetting support.

- close over variable-composite component glyphs
- filter VARC coverage and glyph records and remap component glyph IDs
- copy retained-GID records byte-for-byte without decoding or rewriting them
- expose VARC component dependencies through the subset dependency API
- fail when retained VARC would be incompatible with variation instancing or passthrough glyph-ID compaction
- allow callers to explicitly drop VARC during instancing

Tests:
- meson test -C build --suite subset --print-errorlogs
- meson test -C build --print-errorlogs test-subset-varc test-ot-depend
- git diff --check


behdad:
FontTools counterpart: https://github.com/fonttools/fonttools/pull/4162

NEWS
```diff
@@ -3,6 +3,8 @@ Thursday, August 13, 2026
 =====================================
 - Various fuzzing and build fixes.
 - Various subsetting fixes.
+- Support subsetting `VARC` tables. Subsetting with variation instancing is not
```
khaledhosny: These are the release notes of the already released 14.3.1

behdad: Err. Sorry about that. Fixed.

src/OT/Var/VARC/VARC-subset.hh
```diff
@@ -0,0 +1,299 @@
...
+
+  hb_vector_t<unsigned char> record_data;
+  hb_vector_t<hb_ubytes_t> records;
+  bool retain_gids = c->plan->flags & HB_SUBSET_FLAGS_RETAIN_GIDS;
+  if (unlikely (!records.alloc_exact (old_indices.length)))
+    return_trace (fail ());
+
+  bool auxiliary_indices_unchanged = true;
+  for (hb_codepoint_t old_index : condition_indices)
+    auxiliary_indices_unchanged &= condition_map.get (old_index) == old_index;
```
garretrieger: small optimizations, these loops could break early once a change is detected and you could skip checking subsequent loops as well.

test/api/test-subset-varc.c
```diff
@@ -0,0 +1,309 @@
...
+static void
+test_subset_varc_retain_gids (void)
+{
+  hb_face_t *face = hb_test_open_font_file ("fonts/varc-ac00-ac01.ttf");
+  hb_subset_input_t *input = hb_subset_input_create_or_fail ();
+  hb_set_add (hb_subset_input_unicode_set (input), 0xAC00);
+  hb_subset_input_set_flags (input, HB_SUBSET_FLAGS_RETAIN_GIDS);
+
+  hb_face_t *subset = hb_subset_test_create_subset (face, input);
+  g_assert_cmpuint (hb_face_get_glyph_count (subset), ==, 7);
```
garretrieger, line 247: Probably also want to check that for draw equivalence using the old gid + the subset, in order to verify that retain gids kept the gids unchanged.

src/OT/Var/VARC/VARC.cc
```diff
@@ -1,15 +1,214 @@
...
+  return true;
+}
+
+bool
+VARC::closure_glyphs (hb_set_t *glyphset) const
+{
+  if (unlikely (glyphset->in_error ()))
+    return false;
+
+  hb_set_t visited;
```
garretrieger: Instead of two additional sets (visited, pending) I think this could be made to work using just pending (moved to outside the loop) and glyphset as visited.

garretrieger (approved):

PR #6211, opened 2026-08-26
[depend] avoid inputCount underflow in context depend loops

Nussu06:
context_depend_lookup and chain_context_depend_lookup build their input-position glyph sets with a `for (i = 0; i < inputCount - 1; i++)` loop, so a Context/ChainContext Rule whose inputCount is 0 (which is what a null rule offset resolves to through the Null pool) underflows the bound to UINT_MAX and reads input[i] far past the object. hb_subset_depend_from_face_or_fail reaches this on a crafted GSUB with a ContextFormat1 RuleSet holding a null rule offset, and ASan reports a global-buffer-overflow past _hb_NullPool at gsubgpos.hh:2899. Both loops are rewritten as `i + 1 < inputCount`, identical for inputCount >= 1 and matching the guarded inputCount handling the rest of the file already uses; a minimized seed is added under test/fuzzing/fonts.


behdad:
Thank you.

PR #6214, opened 2026-08-27
[limits] Add rendering work budgets

### LINKED ISSUE #6213: Design: threaded work budget to bound COLR/VARC/CFF → outline traversal (fixes vector/raster/gpu fuzzer timeouts)
**Design proposal** for bounding adversarial outline-traversal work, prompted by three OSS-Fuzz timeouts in the `hb-vector`, `hb-raster`, and `hb-gpu` fuzzers (e.g. testcase `4886959609413632`). All three share one shape: a COLR/VARC paint walk issues many glyph draws, and glyf's per-draw structural `edge_count` resets on every draw, so the two 16384 caps multiply (~112M glyf loads measured). A per-draw structural cap cannot stop this; one aggregate work counter, shared in place through nested outline loads, can.

The full design is below. It has been through several rounds of review. An implementation plan will follow as a comment. Feedback welcome — particularly on the public API surface, the policy/live state model, and the third-party-backend scope boundary.

---

# Full design: threaded work budget for COLR/VARC/CFF → outline traversal

Status: DESIGN ONLY (not yet implemented).
Goal: eliminate three OSS-Fuzz timeouts (hb-vector, hb-raster, hb-gpu fuzzers) whose
common shape is a paint walk (COLR or VARC) issuing many glyph draws, each
re-expanding a composite-glyph "diamond bomb" in `glyf Glyph::get_points`. Measured:
~112M glyf loads for one testcase; worst single glyph 5505 loads over 4 distinct
subglyphs. Testcase IDs to name in the regression: 4886959609413632 (vector) + the
raster and gpu equivalents.

---

## 1. Root cause

    hb_*_paint_glyph  (COLR)  or  VARC
      -> N × push_clip_glyph / fill_glyph   (N bounded by COLR edge_count ~16384)
        -> hb_font_draw_glyph               (PUBLIC boundary: font, gid, funcs, data)
          -> glyf/cff/VARC outline traversal (per-draw structural cap RESET each draw)

Two independent structural caps multiply: COLR's `edge_count` bounds N; glyf's
`edge_count` (16384) bounds ONE draw and resets per draw. Backends keep a per-session
work budget (`work_left`), but it is charged only for OUTPUT (segments / flattened
curves / bytes). Degenerate composites traverse enormously while emitting ~nothing,
so the session budget is never charged for traversal → no aggregate bound.

Fix: charge outline traversal against the SAME finite counter that bounds the whole
paint/draw operation, shared IN PLACE through nested loads. A per-draw structural cap
cannot stop the multiplication; one aggregate counter can. Structural caps are
RETAINED (they bound different failure modes, esp. when a client sets UNLIMITED).

---

### Client-facing (one conceptual budget)

    HB_BUDGET_DEFAULT   = INT64_MIN   // "use the backend's built-in cap"
    HB_BUDGET_UNLIMITED = INT64_MAX   // "do not bound" (client opt-in ONLY)

    hb_bool_t hb_paint_set_budget       (pfuncs, paint_data, int64_t cap);
    int64_t   hb_paint_get_budget       (pfuncs, paint_data);            // the CAP (symmetric)
    int64_t   hb_paint_get_budget_remaining (pfuncs, paint_data);        // live remaining
    // identical trio on the draw side: hb_draw_set_budget / _get_budget / _get_budget_remaining

- `set_budget(cap)` sets the client's POLICY: a concrete value, `DEFAULT`, or
  `UNLIMITED`. Returns FALSE if the backend does not implement it (budget NOT enforced
  — a support probe independent of any returned counter value). **The public wrapper
  CLAMPS a negative non-sentinel `cap` to `0` before storing/dispatching** (only
  `INT64_MIN`=DEFAULT and `INT64_MAX`=UNLIMITED are special; every other value is
  clamped into `[0, INT64_MAX-1]`). So the backend never sees a negative cap, `live`
  never starts at a huge negative, and `spend` cannot signed-overflow toward
  `INT64_MIN`. A stray negative becomes "no budget", not a silent no-op.
- `get_budget()` returns the cap the client set (setter/getter symmetry). Returns
  `DEFAULT` if none was set / unimplemented.
- `get_budget_remaining()` returns the live remaining (`< 0` ⇒ exhausted). When the
  backend does not track a live counter (unimplemented, or a per-draw local that was
  discarded), it returns `resolve(policy)` — the resolved cap, ≥ 0, "full budget,
  depletion untracked" — NOT the `DEFAULT` sentinel (`INT64_MIN` would read as
  exhausted under the `< 0` check). Falls back to `INT64_MAX` if there is no policy.

### Internal (two fields on the backend's DATA object)

    int64_t  policy;   // persistent: DEFAULT / UNLIMITED / concrete. Survives clear.
    int64_t  live;     // the running counter for the current session. ALWAYS concrete.
    unsigned active;   // nesting depth; a no-clear backend resolves `live` only at 0→1
                       // (backstop so a nested re-entry never re-arms — §3). Also lets
                       // raster's lazy DEFAULT be resolved on first charge (§3).

Why two: a session spans `clear → draw → … → render` (multiple API calls), so the
running counter can't be a stack local; and `clear` must restore the cap after `live`
has depleted, which one field can't do (it can't be both the untouched cap and the
depleting remainder). The two fields are HB's internal storage in its OWN paint
objects — the client still sees one budget.

Consequences (this is what makes the model simple and safe):
- **`live` is never a sentinel** → no `INT64_MIN - n` UB, no resolve-before-spend
  hazard, no need for saturating arithmetic. Manually-driven callbacks are safe.
- Sentinels live ONLY in `policy`. The hot path only ever sees a concrete `live`.
- `0` stays an observable exact-fit result (never a policy value).

Third-party backends are NOT forced into this split. Most implement no budget
callbacks → per-draw finite fallback (§4). One that opts in may keep a single running
field it reset at its own session boundary; the policy/live split is HB's internal
choice, not part of the protocol.

---

## 3. Lifecycle: resolve at the real session boundary

The session boundary depends on the backend kind — this MUST be stated per kind or
`live` never re-arms:

- **Stateful backends with a clear/render lifecycle (vector/raster/gpu)**: session =
  `clear → render`. A client may accumulate many glyphs into one paint object before
  render; the aggregate spans all of them. (They already reset `work_left` in
  `hb_*_paint_clear`, not per glyph.) Resolve `live` at `clear`/first-use.
- **Backends with NO clear/render lifecycle** (the generic `hb_font_paint_glyph` path,
  custom `paint_data`, bare `hb_font_draw_glyph`): each top-level `hb_font_*_glyph`
  call IS its own session — resolve `live` at the start of the call. This still bounds
  the bomb: the multiplication is WITHIN one `paint_glyph`'s COLR tree (its many nested
  draws share the one `live`); separate top-level calls are separate user operations,
  not a multiplication. Without this rule `live` would resolve once and deplete
  permanently with no reset path.

    on clear / first use after clear|render:   live = resolve (policy)
    during the session (all glyphs, all nested draws):  charge `live` in place
    render / encode:                            ends the session (live frozen at remaining)
    set_budget(v):                              policy = v; live = resolve(v)  (RECHARGE NOW)
    reset():                                    policy = DEFAULT  (restore user config)

`set_budget` RESOLVES IMMEDIATELY — it sets policy AND re-arms `live` now, not
deferred to the next `clear`. This is the intuitive "set my budget to X" and, as a
side effect, gives the no-clear/generic path a reset handle (a client can always
recharge). `clear` re-arms `live` from the stored policy; `reset` sets policy=DEFAULT.

    resolve (policy):
      policy == HB_BUDGET_UNLIMITED (INT64_MAX)  -> INT64_MAX
      policy == HB_BUDGET_DEFAULT   (INT64_MIN)  -> <backend built-in finite default>   // §8
      else                                       -> policy    // 0 or positive (negatives clamped at set_budget)

- **ONLY the two sentinels are special.** `policy` is `0`, positive, or a sentinel —
  `set_budget` already clamped negatives to `0` (§2). `0` is a valid concrete cap
  ("no budget, draw nothing"). `resolve` must NOT map `<= 0` to the default —
  otherwise a transfer that seeds an EXHAUSTED remainder of `0` would recharge a full
  default budget and the bomb survives (codex #2). And a transfer must not START when
  `seeded <= 0` (skip it; the child would just abort on first spend).
- `clear` restores the concrete cap (or a fresh default) — "reset to policy." `reset`
  restores POLICY to DEFAULT.
- **Nesting is scoped to HB backends that re-enter, i.e. cairo.** A *generic* client
  never re-enters `hb_font_paint_glyph` mid-walk — COLR's own nesting recurses inside
  one `hb_paint_context_t`, not through the public entry — so no HB-tracked nesting is
  needed there (a generic client that opts in manages its own session). The one
  re-entry is cairo (`cairo_show_glyphs` → `hb_cairo_render_color_glyph` builds a fresh
  stack `hb_cairo_context_t`), and that is HB's own backend: HB attaches the session
  budget to a SCOPED OWNER that survives re-entry (the `hb_cairo_font_t`/scaled-font or
  the render context), so the nested context reads the parent's remaining. The `active`
  depth (§2) is HB's backstop in these structs; it is never required of a generic
  client.
- **No in-house path runs unlimited by default** — `resolve` is finite unless
  `policy == UNLIMITED` (explicit client opt-in only).
- **Raster's dynamic default resolves LAZILY** — `max(HB_BUDGET_RASTER_PAINT,
  passes·w·h)` needs `w×h`, which is NOT known at `clear` (set later in
  `ensure_initialized`). So DEFAULT resolution is deferred/dirty: resolved on the
  first charged work after surface init, before any charge — not at `clear` (codex #4).
  It never overwrites a client-set concrete/unlimited policy.

---

## 4. How the live counter reaches the traversal: the paint→draw transfer

The aggregate is threaded across the paint→child boundary by a **budget transfer**:
the parent seeds the child from its own remaining, lets the child charge, then debits
ITSELF by the consumption. The child side is pure save/restore over the PUBLIC API;
the parent debits its OWN `live` internally (it owns its `paint_data`). At every
`push_clip_glyph`/`fill_glyph` (`child` = the draw or a nested paint):

    if (my_live <= 0) return;                    // exhausted: don't start the transfer (§3)
    seeded = my_live;                            // internal: my own remaining
    saved  = child.get_budget ();                // PUBLIC: save the child's client cap
    if (!child.set_budget (seeded)) {            // PUBLIC seed; FALSE ⇒ child doesn't participate
      hb_font_draw_glyph (...);                  //   child self-bounds per-draw (HB_BUDGET_GLYPH)
      my_live -= HB_BUDGET_GLYPH;                //   conservative debit so the parent still depletes
      return;
    }
    hb_font_draw_glyph (font, gid, draw_funcs, draw_data);   // child charges child.live
    r = child.get_budget_remaining ();           // PUBLIC: what the child had left
    r = clamp (r, 0, seeded);                    // NEVER trust an out-of-range r (fail-open guard)
    child.set_budget (saved);                    // PUBLIC: restore the child's client cap
    my_live -= seeded - r;                       // internal: debit by consumption ∈ [0, seeded]

Fail-safe (codex #1): the parent must never trust an uncooperative child. If
`set_budget` returns FALSE the child ran unbudgeted per-draw (its own `HB_BUDGET_GLYPH`
local), so debit that conservative amount and move on. Otherwise CLAMP `r` to
`[0, seeded]` so consumption is always in `[0, seeded]` — an unsupported/garbage
`get_budget_remaining()` (e.g. `INT64_MAX`) can then never INCREASE the parent's
budget.

Why this needs NO `set_budget_remaining` and NO write-live op: the ONLY cross-object
interaction is with the child — save→seed→read→restore over public
`get_budget`/`set_budget`/`get_budget_remaining`. `set_budget`'s immediate-resolve
(§3) is fine: the child's cap is restored afterward, its `live` between transfers is
irrelevant, its CLIENT policy preserved. The parent never needs a public op on itself.

A **bare** `hb_font_draw_glyph` (no parent transfer) uses a LOCAL discarded budget
with NO write-back, so it never touches the sink's cap — the snapshot write-back
happens ONLY inside a transfer, where the parent's save/restore masks the transient
policy change. So no cap corruption anywhere, with no write-remaining callback.

- **In-house backends** ("or internal variable access"): parent and child are HB's
  own code, so the save/restore collapses to sharing — `child.live` aliases the
  session `live`; glyf/CFF and the child's own output/flatten all charge that one
  field in place (no divergence). No public round-trip.
- **Third-party paint backend**: does exactly the block above — owns its `live` for
  the debit, uses the public child calls for the seed/read/restore. So it too gets
  full aggregate bounding. The protocol is genuinely composable (below).
- **Nested paint (cairo color-glyph re-entry)**: the outer paint transfers into the
  fresh nested context the same way, so the nested continues the budget instead of
  re-arming (§3 nesting).
- **Bare `hb_font_draw_glyph`** (no paint): resolve a LOCAL finite counter at entry —
  `resolve(get_budget(data))` if the draw funcs expose a cap, else `HB_BUDGET_GLYPH`.
  Finite, single-glyph; `edge_count` also caps it. Not a multiplication, so per-draw
  is enough.

### Scope: third-party is opt-in via the public protocol

A third-party paint backend that performs the save/restore transfer + its own `live`
debit gets full aggregate bounding — no HB-private hooks needed. A backend that does
NOT (just calls `hb_font_draw_glyph` with no transfer) gets `HB_BUDGET_GLYPH` per
draw — no aggregate, bombable as today, no regression. HB still won't smuggle it in
transparently (no `thread_local`; see `_hb_CrapPool`, `hb-null.hh`) — but the public
protocol gives third parties a real path in.

---

## 5. The spend primitive

Reference-based, always-inline, in `hb-limits.hh` (self-contained). `live` is always
concrete, so the primitive needs no sentinel branch and no saturation:

    HB_ALWAYS_INLINE static bool
    hb_budget_spend (int64_t &budget, int64_t cost, int64_t mult = 1)
    {
      budget -= cost * mult;         // cost = HB_BUDGET_N weight; mult = count (e.g. #points)
      return budget >= 0;            // false ⇒ exhausted, caller aborts
    }

- Reference, never a pointer — null can't reach it.
- `cost` is REQUIRED (a ladder weight, §6); `mult` defaults to 1. Overflow-safe by
  construction: negative caps are clamped to `0` at `set_budget` (§2), so `budget`
  starts at `0`, positive, or `INT64_MAX`; `cost*mult` is bounded by structural caps
  (`cost ≤ 1024`, `mult ≤ points×tuples` etc. — worst ~1e13 ≪ 1e18); and the precharge
  guards (§12) latch exhaustion (`<= 0` stops charging), so `budget` never marches
  toward `INT64_MIN`. No sentinel branch, no saturation needed.
- Charge sites read: `if (unlikely (!hb_budget_spend (budget, HB_BUDGET_16))) return false;`
- Threading param through `get_points`/`get_path`/`get_extents_at`/`cff::get_path`
  stays `int64_t *budget = nullptr` (the `edge_count` idiom, Glyph.hh:320) so existing
  non-session callers are untouched; the entry backs an absent budget with a stack
  `HB_BUDGET_GLYPH` local and passes its reference down. Null never reaches `spend`.

---

## 6. Charge model — fixed power-of-2 weight ladder

**Budget unit ≈ one machine cycle** — a sizing INTUITION, not accounting. It only
buys two things: RELATIVE op weights, and translating a cap to a rough wall-clock
ceiling (cap ÷ ~GHz ≈ seconds vs. the 61s timeout + ASAN slowdown). Absolute cycles
are not tracked.

**Weights: a fixed ladder in `hb-limits.hh`, name = value** (greppable, transparent):

    HB_BUDGET_1, HB_BUDGET_2, HB_BUDGET_4, … HB_BUDGET_1024        // == 1,2,4,…,1024

Weights are FIXED forever; ALL tuning lives in the caps (§8). A site's weight states
relative op cost; the cap states total allowed. Re-tuning = picking another
`HB_BUDGET_N` or resizing a cap, never editing a constant's value. Optionally alias
important sites (e.g. `HB_BUDGET_GLYF_LOAD = HB_BUDGET_16`) so code self-documents.

Charges — flat for O(1) steps, `count × weight` where work scales with data:

| charge          | site                                              | weight (estimate) |
|-----------------|---------------------------------------------------|-------------------|
| per-output      | draw emitters / path_builder (per segment)        | `HB_BUDGET_1` |
| per-flatten     | raster/gpu curve flattening (per segment)         | `HB_BUDGET_1` |
| per-load        | `get_points`, at the `edge_count++` site          | `HB_BUDGET_16` |
| gvar            | delta apply — `points × tuples/regions`           | proportional |
| CFF2 blend      | `operands × region count` (the real inner loop)   | proportional |
| terminal points | `get_path_at`/`get_extents_at` over N points      | `N × HB_BUDGET_1` |
| per-cff-op      | charstring interp loop                             | `HB_BUDGET_1`; callsubr/gsubr `HB_BUDGET_8` |
| VARC component  | per component/record processed                    | `HB_BUDGET_4` |

- Flat per-load catches the DEGENERATE bomb (many loads, ~0 points). Proportional
  charges catch point/region-heavy work. gvar and CFF2 blend are NOT point-only —
  charge the real inner loop (points×regions / operands×regions), per codex.
- Final weights/caps from the §10 measurement.
- All charges go through `hb_budget_spend`. (Backends' existing per-output/flatten
  decrements ARE these charges, retargeted to the shared `live`.)

### `edge_count` is retained (not redundant)

`edge_count` bounds ONE glyph's composite DAG (16384 loads, resets per draw); the
budget bounds the whole operation. Decisive case: under a client-set `UNLIMITED`,
`spend` never stops, so `edge_count` is the only thing preventing a single malicious
composite (a DAG, not a cycle; depth ≤ 64 ⇒ up to ~2⁶⁴ loads) from hanging.
"Unlimited" must mean "lift my session cap," not "let a font DoS me." Use the exact
existing off-by-one (`> HB_MAX_GRAPH_EDGE_COUNT` before increment), or fix it
separately.

---

## 7. Threading + the forwarding-adaptor checklist

The paint→draw transfer (§4) assumes OT charges the SAME `draw.live` the paint seeded.
But several layers replace the draw funcs/data before OT sees them, and a paint→draw
callback can run with no `hb_font_*_glyph` boundary at all. Every forwarding pen/
context must carry the same live reference (or re-perform the transfer), or a
cumulative counter silently becomes per-draw. Checklist:

- **Generic entry** `hb_ot_draw_glyph_or_fail` / paint: thread `int64_t &budget` into
  `VARC->get_path`, `glyf->get_path`, `cff2/cff1->get_path`. Construct the budget
  BEFORE the `hb_draw_session_t` (whose dtor emits a final `close_path` that charges —
  §9), and close the session in an inner scope.
- **Synthetic slant/embolden**: draws into `hb_outline_recording_pen` first, then
  replays (`hb-font.hh:689-735`). OT sees the recorder, not the sink — establish the
  budget above this switch or carry it in the recording context.
- **Parent-font adaptor**: child-font draw redirects through `_hb_draw_funcs_default`
  + adaptor (`hb-font.cc:734-762`) — must forward budget access.
- **paint-extents**: draw_data is a stack `hb_paint_extents_counting_sink_t`, not the
  context holding `work_left` (`hb-paint-extents.cc:68-72,151-168`) — the sink needs a
  link to the live counter.
- **vector**: sink already carries a work pointer → point it at `&live`.
- **raster**: use `external_work ? external_work : &flatten_work_left`; returning only
  `external_work` splits standalone traversal from flattening.
- **gpu**: `hb_gpu_draw_t` has no work pointer today; the identity scratch-draw AND the
  transforming pen must both link to `&live`, else one gets a fresh per-glyph fallback.
- **VARC pen**: forward the pointer through the transforming pen and final draw
  session, all inside the budget lifetime.
- **cairo**: static draw funcs + raw `cairo_t *` at several sites; the cairo paint
  context owns the session budget; the transfer (§4) passes the live reference. Nested
  color-glyph: `hb_cairo_paint_color_glyph` → `cairo_show_glyphs` can re-enter
  `hb_cairo_render_color_glyph` and build a FRESH stack `hb_cairo_context_t`
  (`hb-cairo.cc:192-215,680-689`) — a pointer only in the outer context won't cross;
  attach the budget to the scoped render context so nesting propagates.

Missing one adaptor is a silent correctness bug (cumulative → per-draw). This becomes
an implementation checklist with a test per path (§11).

---

## 8. Backend constants (`hb-limits.hh`, all `HB_BUDGET_` prefixed)

Session caps are DERIVED `#define`s: `HB_BUDGET_GLYPH × <glyphs-per-session>`, so only
a small integer FACTOR per backend is hand-tuned (§10):

    // a session allows ~N glyph-equivalents; only N (per backend) is measured (§10):
    #define HB_BUDGET_VECTOR_PAINT (HB_BUDGET_GLYPH * HB_BUDGET_VECTOR_PAINT_GLYPHS)

- `HB_BUDGET_GLYPH` = finite fallback for a single-glyph / no-session call in ANY
  outline format (glyf/CFF/VARC) — format-agnostic, hence GLYPH not GLYF. Distinct
  from the `HB_BUDGET_DEFAULT` sentinel (an actual value). It is a MEASURED
  compatibility target, NOT a closed form: a tight `edge_count×16 + maxPoints×1`
  (~462k) is WRONG — §6 charges gvar as `points × tuples`, so a heavily-variable glyph
  is ~10M+ (200k points × dozens of tuples), an order of magnitude above that, before
  CFF2 blend or VARC aggregation. Size it from the §10 legit corpus to exceed the
  worst legitimate glyph across glyf load+points, gvar tuple/point work, CFF
  charstring+seac+subrs, CFF2 blend operand×region, and VARC leaf aggregation — with
  margin. (The budget may still intentionally truncate pathological-but-valid glyphs;
  that is a documented compatibility target, not a theoretical guarantee.)
- Rename existing caps into the namespace: `HB_VECTOR_MAX_PAINT_WORK →
  HB_BUDGET_VECTOR_PAINT`, `…RASTER…`, `HB_PAINT_EXTENTS_MAX_WORK →
  HB_BUDGET_PAINT_EXTENTS`, `HB_GPU_PAINT_MAX_WORK → HB_BUDGET_GPU_PAINT`,
  `HB_VARC_MAX_WORK → HB_BUDGET_VARC` (retained as VARC's aggregate leaf cap). Add
  `HB_BUDGET_CAIRO_PAINT`.
- Non-budget structural caps keep unit names (`HB_GLYF_MAX_POINTS`, `HB_CFF_MAX_OPS`,
  `HB_MAX_GRAPH_EDGE_COUNT`, `HB_RASTER_MAX_DRAW_EDGES`, `HB_RASTER_MAX_BUFFER_SIZE`).
- **Public** (draw/paint header): ONLY `HB_BUDGET_DEFAULT`/`HB_BUDGET_UNLIMITED` + the
  callback typedefs/setters. The weight ladder and caps are internal. The numeric
  scale is NOT a public contract — clients prefer the sentinels; a concrete `N` is an
  advanced, version-specific knob (JS needs BigInt for the int64 sentinels).

---

## 9. Per-outline-format accounting

- **glyf**: per-load `HB_BUDGET_16` at the `edge_count++` site; gvar proportional
  (points × tuples); terminal points `N × HB_BUDGET_1`. Thread `int64_t *budget`
  through `get_points`/`get_path[_at]`/`get_extents_at`.
- **CFF**: `cff1/cff2::get_path` take `int64_t *budget`; charge per charstring op and
  callsubr/gsubr. CFF1 `seac` calls `_get_path()` for base AND accent
  (`hb-ot-cff1-table.cc:519-556`) — the budget must live in the CFF path param /
  environment so `process_seac` forwards it into BOTH recursions (charging only
  `interpret()` is insufficient). CFF2 blend charges `operands × regions`.
- **VARC**: charge VARC's OWN component/record traversal + variation work, not just
  leaves — a condition-skipped / malformed / no-outline component graph can consume
  work while reaching few charged leaves, and multiplies through COLR. Retain VARC
  structural edge/depth caps; forward the same pointer to all glyf/CFF leaves.

### Lifetime correctness

The threaded budget is a plain `int64_t &` — `&data->live` in-house (identity
transfer), a stack `HB_BUDGET_GLYPH` local for a bare draw, or the `local` inside the
per-draw snapshot for a third-party value-only sink (§4). Rules:

- **Snapshot holder (third-party value sinks only)**: it has a self-referential
  pointer (`ptr == &local`); **delete copy/assignment, construct in place** (no
  return-by-value factory). Its ctor reads `draw.get_budget_remaining()` into `local`;
  its dtor writes `local` back via `set_budget` on ALL `get_points` return paths. In-
  house backends use `&data->live` directly and have no such object.
- Construct the budget/local BEFORE the `hb_draw_session_t` (whose dtor emits a final
  `close_path` that charges, `hb-draw.hh:176-182`) and close the session in an inner
  scope, so the reference outlives every emitter.
- The same order applies to forwarding sessions built inside VARC / GPU / synthetic
  recorders — the `&live` must stay valid until every pen has flushed.
- No callback-pair validation: it is the user's responsibility to register both
  `get_budget`/`set_budget` (and `get_budget_remaining`) sensibly; a partial set
  simply means the backend does not participate.

---

## 10. Defaults & tuning (measure, don't assert)

- Weights FIXED (§6); tune CAPS only. A bomb aborts after ~`cap / load-weight` loads.
- Use weighted totals `Σ(count_i × weight_i)`, not raw load counts. (Correcting §
  arithmetic: a 16M cap at load-weight 16 aborts ≈ 1M loads ≈ ~0.9% of the 112M bomb —
  ~112× load-count headroom, not "6.7×".)
- Reproduce all THREE testcases on the i386 ASAN target; confirm each finishes WELL
  under 61s, reporting weighted totals + elapsed so later tuning distinguishes a cap
  change from a cost-model change.
- Build a LEGITIMATE corpus (large composite glyf; CJK/complex; variable gvar with
  many tuples; VARC deep/broad; CFF1 seac + subr-heavy; CFF2 variation-heavy; COLRv0/v1
  many layers/gradients/nested; vector/raster/gpu/cairo; synthetic slant/embolden;
  child/parent). Record max + percentiles of weighted consumption per op/backend;
  pick caps with explicit margin. Measure 32/64-bit optimized AND i386 ASAN.

---

## 11. Testing

Deterministic tests (assert weighted-work / load ceilings, not just wall-clock):
- **Semantics**: draw+paint set/get round-trips; DEFAULT/UNLIMITED/concrete; exact-0
  as a valid "no budget" cap (NOT recharged to default); already-negative; repeated
  calls after exhaustion (no runaway); support probe; the save/restore transfer
  (child cap preserved, parent debited); nested/cairo re-entry continues (not
  re-armed); clear/reset/render/encode lifecycle; raster lazy dynamic default.
- **Propagation** (one per §7 adaptor): generic entry vs each convenience wrapper;
  manually-driven draw/paint funcs; synthetic recorder; parent-font adaptor; vector;
  raster local+external; gpu identity+transformed; paint-extents; cairo wrapper +
  color-glyph re-entry.
- **Formats**: glyf composite diamond with asserted ceiling; VARC component-heavy +
  condition-skipped; CFF1 seac forwarding; CFF subr-heavy; CFF2 blend
  operands×regions; gvar proportional; structural caps active under `UNLIMITED`.
- **Build/robustness**: `HB_NO_DRAW`/`HB_NO_PAINT`/`HB_NO_VAR`/`HB_NO_VAR_COMPOSITES`/
  `HB_NO_CFF`; amalgam `harfbuzz.cc`; GCC/Clang/MSVC; failing-malloc; C + C++ header
  compile + symbol/API tests. Name the three OSS-Fuzz IDs.

---

## 12. Status semantics (narrowed) & compatibility

- Budget status is BUDGET-only, not a global success oracle: `< 0` a charge exceeded
  the amount; `0` no positive-cost work remains (prior work may have fit exactly);
  `> 0` capacity remains. `remaining ≥ 0` does NOT imply "every op succeeded" — ops
  fail for OOM / malformed / structural caps with budget left. Success/failure stays
  in existing return values + backend error state.
- Guards (DECIDED — precharge): the budget's public `remaining` semantics use `< 0`
  (so `0` = exact-fit success). But the existing backend guards stay PRECHARGE checks —
  `if (work_left <= 0) skip` — i.e. "can I start positive-cost work?" (`0` correctly
  blocks the next op, skipping more work). They are NOT blanket-rewritten to `< 0`.
  This supersedes the earlier "harmonize all `<= 0` to `< 0`" and is compatible with
  the public `0`-as-success contract. Prefer precharge everywhere the cost is known;
  POSTCHARGE only where cost is learned after emission (serializers) — there, one
  bounded overshoot or reserve/rollback, not a mechanical guard flip.
- Draw failure semantics CHANGE: `hb_font_draw_glyph_or_fail` documents `false` as "no
  outline"; a finite budget can now abort with an outline present. Paint callbacks are
  void → a paint op may report data yet produce partial output. State that exhaustion
  is best-effort truncation (glyf returns false mid-load; emitters stop) and how a
  caller detects it (`get_budget_remaining() < 0`). Update NEWS + the return-value docs.
- Compatibility: "no expected change below the documented structural/work ceilings,"
  NOT an absolute "no behavior change." `hb_budget_t` + outline-format charges are
  shared infrastructure — under `HB_NO_PAINT`/`HB_NO_DRAW`, the core outline/extents
  charges still compile; enumerate what remains rather than "compiles out with the
  backend."

---

## 13. Scope

- IN: OT glyf/CFF/VARC; the paint→draw transfer (§4) + forwarding adaptors (§7); vector,
  raster, gpu, cairo, paint-extents; public draw+paint `set/get_budget[_remaining]`;
  `hb_budget_t` + `hb_budget_spend`; the weight ladder + derived caps.
- BEST-EFFORT / OUT: external-library font backends (FreeType, CoreText, DirectWrite,
  fontations) — HB can't instrument their internal glyph loading; they bound HB-side
  decomposition only, best-effort, with the limitation stated. No `thread_local`.

---

## 14. Docs obligations

Public declarations + callback typedef docs; `XSince: REPLACEME`; `NEWS` (new APIs +
changed draw/paint failure semantics); `docs/harfbuzz-sections.txt`;
`docs/harfbuzz-docs.xml`; API tests for all six generic functions (set/get/
get_remaining × draw/paint); language-binding
notes for the 64-bit sentinels; reduced-feature guards consistent between
declaration and definition.

---

## 15. Commit plan

Ordered to **prove the fix before freezing the public surface** — what actually kills
the three reports is small (glyf per-load charge + one shared session counter in the
in-house paint backends), so land and verify that first; commit the API/constants
only once the i386-ASAN measurement (§10) confirms the caps. Every intermediate build
keeps finite defaults — no commit silently runs a bounded path as unlimited.

1. `hb_budget_spend` + weight ladder in `hb-limits.hh`; glyf per-load + gvar/terminal
   charges threaded through `get_points`/`get_path[_at]` (internal, no public API yet).
2. in-house paint backends (vector/raster/gpu/paint-extents) share one session `live`
   via the paint→draw transfer (§4) + the §7 adaptors; lifecycle (clear/reset). **Verify all
   three OSS-Fuzz testcases on i386 ASAN + the §10 legit corpus; fix caps from data.**
3. freeze the public draw/paint `set/get_budget[_remaining]` API + policy/live state +
   sentinels/constants; docs (§14).
4. VARC + CFF accounting (seac forwarding, CFF2 blend operand×region); cairo.
5. remaining in-house backends (best-effort) + reduced-feature guards.
6. deterministic regressions, fuzz seeds, tuning data.

behdad:
Fixes the multiplicative outline-traversal work described in #6213.

Outline traversal now spends from a weighted work budget, and HarfBuzz's
in-house renderers share that budget across nested glyph draws instead of
restarting a per-glyph allowance. This covers glyf, CFF1/CFF2, VARC, synthetic
outlines, paint extents, vector, raster, GPU, and Cairo color-glyph re-entry.

The change also adds the public draw/paint budget protocol:

- `hb_{draw,paint}_{set,get}_budget()`
- `hb_{draw,paint}_get_budget_remaining()`
- finite defaults plus explicit `HB_BUDGET_DEFAULT` and
  `HB_BUDGET_UNLIMITED` policies

The three supplied ClusterFuzz testcases are included as regression seeds.

Testing:

- full Meson suite after every commit (275/275 at HEAD)
- CMake C++11 build with Cairo, subset, raster, vector, and GPU
- C++11 amalgam builds: normal, `HB_TINY`, and 32-bit
- reduced-feature Meson builds: `HB_TINY`, `HB_NO_COLOR`, and
  `HB_NO_VAR_COMPOSITES`
- ASan+UBSan runs of all three regression seeds
  - vector: 0.01s
  - raster: 0.01s
  - GPU: 0.38s

Closes #6213.


garretrieger (approved):
Did a high level review and lgtm.

behdad (commented):
## Claude's review of #6214

*(Posted by Behdad on Claude's behalf — the assessment and wording below are Claude's.)*

This is a faithful, well-structured implementation of the #6213 threaded-work-budget design. The core pattern — one live counter owned by the top-level renderer, borrowed in place by nested glyf/CFF/VARC traversal via a `get_budget_remaining` pointer callback — is applied consistently across glyf, CFF1/CFF2, gvar, VARC, synthetic outlines, paint-extents, vector, raster, GPU, and Cairo. The `policy` (persistent) vs `budget_remaining` (live) split, the `recharge_budget()` helper, and the `HB_BUDGET_DEFAULT`/`HB_BUDGET_UNLIMITED` sentinels all match what we settled on. NEWS, section index, and public docstrings are in place, and the docstrings correctly avoid implementation detail.

**Independent verification of the central claim.** I ran all three minimized ClusterFuzz seeds against the HEAD-built `build/src/libharfbuzz.so.0` (confirmed via `LD_DEBUG=libs`):

| seed | fuzzer | time |
|------|--------|------|
| 4886959609413632 | vector | 0.85s |
| 6328536122851328 | raster | 0.35s |
| 5336988644081664 | gpu    | 0.49s |

All complete sub-second — the multiplicative outline-traversal blowup is gone. I did **not** re-run the full Meson suite, the reduced-feature builds, or the ASan/UBSan/TSan runs; those claims are from the PR, not independently reproduced here. Everything else below is a code read.

**Things I checked that are correct:**
- The paint-extents `budget_initialized` trick (distinguishing a zero-filled COLR scratch from a client-selected `0`) is sound — `COLR.hh:2786` always calls `scratch.paint_extents.clear()` before the first `push_clip_glyph`, so the guard fires.
- glyf's terminal points charge runs *before* the consumer emits, so an out-of-budget glyph is skipped cleanly rather than emitting a partial path.
- The Cairo `active` depth counter correctly recharges only on the outermost callback, and storing the budget on the `scaled_font` is safe because Cairo serializes user-font glyph init under the scaled-font mutex (the same lock the TSan-skip commit is about).
- raster-draw's flatten loops write the local `work_left` back to the shared counter on their single return path — no lost updates.

**Findings** (none are blockers; ranked):
1. **CFF interpreter hot-loop branch** (perf) — per-op cost branch added to the innermost loop on the normal path for every CFF font; hoistable. See inline.
2. **Flat `HB_BUDGET_GLYPH` default for session-spanning vector/PDF draw** (possible correctness) — the byte-denominated cap could truncate a legitimate large multi-glyph document. See inline — this is the one I'd most like an answer on.
3. **`hb_budget_spend` computes `cost * mult` in `unsigned int`** (defensive) — unreachable overflow today, but the design called for 64-bit here. See inline.
4. **`set_budget` write-through clobbers a borrowed counter** (latent) — unreachable, but worth removing rather than papering over. See inline.
5. Assorted nits, batched in one inline comment.

Nice work overall — the shape of the code is clean and the fix demonstrably holds.

src/hb-cff-interp-cs-common.hh
```diff
@@ -880,11 +887,22 @@ struct cs_interpreter_t : interpreter_t<ENV>
...
   {
     SUPER::env.set_endchar (false);
+    SUPER::env.set_budget (budget);
 
     bool ret = true;
     unsigned max_ops = HB_CFF_MAX_OPS;
     for (;;) {
-      OPSET::process_op (SUPER::env.fetch_op (), SUPER::env, param);
+      op_code_t op = SUPER::env.fetch_op ();
+      unsigned int cost = op == OpCode_callsubr || op == OpCode_callgsubr ?
```
behdad: **(Claude, via Behdad)** This per-op `cost` computation adds two comparisons + a branch to the CFF charstring interpreter's innermost loop — the hot path for *every* CFF font, not just adversarial ones — replacing the old single post-loop `*budget -= (HB_CFF_MAX_OPS - max_ops)`.

Per-op charging is the right call (post-charging lets a charstring burn ~200k ops before anyone notices), but the branch is hoistable: charge `HB_BUDGET_1` unconditionally here, and add the extra `HB_BUDGET_8 - HB_BUDGET_1` inside the `callsubr`/`callgsubr` handlers, where a branch on the opcode already exists. Same semantics, no new hot-loop branch — worth an A/B given the ongoing shape-perf work.

(Same loop, minor: `spend_budget()`'s `!budget ||` null-test also runs per op; on the OT path `budget` is always non-null so it predicts well, but it's the same instruction-budget concern.)

src/hb-limits.hh
```diff
@@ -159,55 +159,42 @@
...
+#ifndef HB_BUDGET_GLYPH
+#define HB_BUDGET_GLYPH ((int64_t) 1 << 26)
+#endif
+
+/* Precharge COST * MULT.  Callers bound COST and MULT structurally, and live
+ * budgets are concrete non-negative values when a session starts. */
+static HB_ALWAYS_INLINE bool
+hb_budget_spend (int64_t &budget, unsigned int cost, unsigned int mult = 1)
+{
+  budget -= cost * mult;
```
behdad: **(Claude, via Behdad)** `cost * mult` is evaluated in `unsigned int`, so it wraps at 2³² *before* the widening to `int64_t`. No call site can reach that today — the worst case is VARC's `axisValues.length (≤ 4096) × coords.length (≤ 65535) ≈ 2.7e8` — so this is not a live bug. But the whole point of this primitive is hardening, and the #6213 design specified 64-bit arithmetic here. One cast removes the reasoning burden permanently:

```c
budget -= (int64_t) cost * mult;
```

Cheap insurance against a future charge site or a relaxed structural bound.

src/hb-limits.hh
```diff
@@ -159,55 +159,42 @@
...
+#define HB_BUDGET_64	64u
+#define HB_BUDGET_128	128u
+#define HB_BUDGET_256	256u
+#define HB_BUDGET_512	512u
+#define HB_BUDGET_1024	1024u
+
+/* The common finite default for one top-level glyph rendering session.
+ * Nested outline and paint work shares the same live counter. */
+#ifndef HB_BUDGET_GLYPH
+#define HB_BUDGET_GLYPH ((int64_t) 1 << 26)
```
behdad, line 189: **(Claude, via Behdad)** This `2²⁶` lands as the single finite default for VARC and *all* in-house draw/paint backends — a flat per-session cap. The #6213 design instead described derived caps (`HB_BUDGET_GLYPH × <glyph count>`, factor measured).

That's fine for a one-glyph paint session, but the vector/PDF **draw** backends run one session across a whole document of many glyphs, and `end_path_command` charges `path.length - before` — i.e. **serialized output bytes**, not outline points (`hb-vector-draw.hh:82`). Is a legitimate large multi-glyph SVG/PDF guaranteed to stay under 2²⁶ bytes of path data? If not, this could silently truncate valid output, which is exactly the regression class CLAUDE.md warns about.

Was dropping the per-glyph-count scaling deliberate for these session-spanning backends, or should vector/PDF get a larger (or output-size-derived) default the way raster keeps its surface-scaled one?

src/hb-raster-draw.cc
```diff
@@ -1091,6 +1102,30 @@ hb_raster_close_path (hb_draw_funcs_t *dfuncs HB_UNUSED,
...
 }
 
+static hb_bool_t
+hb_raster_draw_set_budget (hb_draw_funcs_t *, void *draw_data,
+			   int64_t budget, void *)
+{
+  auto *draw = (hb_raster_draw_t *) draw_data;
+  draw->budget = budget;
+  draw->recharge_budget ();
+  if (draw->external_budget)
```
behdad: **(Claude, via Behdad)** When `external_budget` is set (a scratch draw borrowing a paint session's counter), `set_budget()` writes the fresh recharge back into the parent's shared counter — clobbering the aggregate. It's unreachable today: borrowing scratch draws aren't user-addressable, and nothing calls `set_budget` on them internally. So this is a question, not a bug — is the write-through needed at all? If a borrowing draw can never legitimately receive a public `set_budget`, dropping the `if (external_budget)` block (here and the mirror at `hb-gpu-draw.cc:249`) removes a latent foot-gun rather than papering over it.

src/hb-raster-draw.cc
```diff
@@ -519,9 +530,9 @@ hb_raster_draw_set_clip_box (hb_raster_draw_t *draw,
 
 void
 hb_raster_draw_set_external_work (hb_raster_draw_t *draw,
```
behdad: **(Claude, via Behdad)** Minor nits, batched (non-blocking):

- The setter is still named `hb_raster_draw_set_external_work` but now sets `external_budget` — stale name.
- `flatten_{quadratic,cubic}_recursive` still guard on `work_left <= 0`, while the *Consolidate* commit harmonized the aggregate guards to `< 0` (0 = exact fit succeeds). Harmless here (local degrade-to-chord heuristic), just an outlier from the stated boundary rule.
- NEWS drops the trailing blank line at EOF — unrelated whitespace churn; probably worth restoring to keep the diff minimal.

src/hb-cff-interp-cs-common.hh
behdad: Hoist the accounting such that there's no budget code inside the hot op loop at all.

src/hb-limits.hh
behdad: Sure, cast it.

src/hb-limits.hh
behdad: SVG/PDF are *really* to render a single glyph. They are not full document APIs. So I'm fine with keeping this.

src/hb-raster-draw.cc
behdad: Scratch draws are ours to manage.

src/hb-raster-draw.cc
behdad: We should remove the separate `external_work` and unify on the shared budget API instead.

behdad:
**Claude's follow-up design note** *(posted by Behdad on Claude's behalf — wording is Claude's).*

Capturing a design point for a follow-up change, so the discussion hanging off the `external_budget` / `set_external_work` thread has a home. Two problems are tangled together in the raster (and GPU) budget wiring:

**1. Layering.** `hb_raster_paint` reaches into `hb_raster_draw` internals in two places — `hb_raster_draw_set_external_work()` to inject a counter pointer, and `hb_raster_draw_get_edge_work()` to read back scanline cost. Both are `HB_INTERNAL` cross-module calls. This PR *adds a public draw/paint budget protocol* precisely so these two modules can talk through it instead of poking each other's internals.

**2. Dimensional conflation.** The one session counter charges both outline-traversal work (`HB_BUDGET_GLYPH` scale — glyf loads, CFF ops, flatten subdivisions) and pixel/scanline fill work (≈ edges × rows, a much larger scale). That conflation is exactly why raster-paint has to inflate its default with `HB_BUDGET_RASTER_PAINT_PASSES × w × h`. Two different resources should not share one counter.

**Proposed follow-up — two budgets:**

- *Outline budget* = the public draw/paint budget added in this PR (`HB_BUDGET_GLYPH`, cross-backend). raster-paint seeds raster-draw via `hb_draw_set_budget()`; glyf/CFF/flatten charge it inside the child; raster-paint reads it back via `hb_draw_get_budget_remaining()`. Pure public API — no raster-draw internals touched.
- *Pixel budget* = a separate, larger, surface-scaled session budget owned by raster-paint, charged per glyph from the **area of the returned mask image** (`hb_raster_image_get_extents()`, public) — no `get_edge_work()` peek.

Both stay session-aggregate, so COLR still cannot multiply either dimension across a paint walk.

Net effect: removes *both* internal calls, un-conflates the two costs (so `HB_BUDGET_GLYPH` goes back to purely outline work and the surface-scaling moves onto the pixel budget where it belongs — which also retires the awkward lazy surface-sized default), and makes raster/GPU dogfood the public budget API this PR is built around. Same split then carries to GPU (curve budget vs. fill/encode cost).

Deferring this to a follow-up; addressing the four smaller review findings first.

src/hb-cff-interp-cs-common.hh
behdad: **(Claude, via Behdad)** Fixed in 115c3c0028 — charge one unit unconditionally in the loop, and add the remaining call weight inside the callsubr/callgsubr handlers. Total per-call-operator charge is unchanged (1 + 7 = 8), the subroutine is still not entered once the budget is exhausted, and the hot loop is now branch-free.

src/hb-limits.hh
behdad: **(Claude, via Behdad)** Fixed in 565cbc3768 — `budget -= (int64_t) cost * mult;`.

src/hb-raster-draw.cc
behdad: **(Claude, via Behdad)** NEWS end-of-file churn fixed in 148fcb3c79. Withdrawing the flatten `work_left <= 0` point — that guard is actually correct: a subdivision is positive-cost work, and per the public `hb_draw_get_budget_remaining` doc zero means no positive-cost work can start, so `<= 0` is the right boundary there. The `set_external_work` rename is subsumed by the raster budget refactor in the follow-up design note (the function is removed there).

src/hb-limits.hh
behdad: **(Claude, via Behdad)** Confirmed real, but it's the same output-vs-outline dimensional issue as the raster pixel budget, not a nit: `hb_vector_draw_glyph_or_fail` calls `new_path()` (not clear), so multiple accumulated glyphs share one flat, byte-denominated budget until clear/render. A proper fix touches the design's vector session boundary, so folding it into the dimensional-separation follow-up (see the design note comment on the PR) rather than a standalone patch. Leaving this thread open to track it.

src/hb-raster-draw.cc
behdad: **(Claude, via Behdad)** Deferred: this write-through disappears entirely in the raster budget refactor (the `external_budget` field and `set_external_work` are both removed, replaced by public `set_budget`/`get_budget_remaining` transfer). See the follow-up design note comment on the PR. Leaving open to track.

src/hb-cff-interp-cs-common.hh
behdad: **(Claude, via Behdad)** Update: the approach evolved past the per-op hoist. Final version is 83eafa8ca6 — the loop's existing `max_ops` counter is capped at `min(HB_CFF_MAX_OPS, remaining_budget)` and the consumed ops are charged post-loop, so one counter enforces both the structural cap and the live budget with *no* budget check in the hot loop, and the op count can't overshoot. (A byte-length precharge was tried and rejected: a 110-byte charstring can execute 200K ops via subr reuse, so length undercounts execution ~900×.)

src/hb-raster-draw.cc
behdad: **(Claude, via Behdad)** Resolved by the two-budget/decoupling refactor. The `external_budget` field and its `set_budget` write-through are removed from both backends: raster in dfee0d4380 (outline vs pixel split, seed/readback via the public draw-budget API) and GPU in 76eccd5f27 (single outline-derived budget, same public transfer — the GPU backend never rasterizes, so no pixel dimension applies). Neither backend reaches into the rasterizer/encoder internals anymore.

src/hb-limits.hh
behdad: **(Claude, via Behdad)** Resolved in 7eec305e71. The flat outline default was only a symptom of conflation. vector output is now bounded by a separate flat document-size cap (HB_VECTOR_MAX_DOCUMENT_SIZE, 16MB) on the output buffer — covering path serialization, sweep-gradient meshes, and (newly) embedded bitmaps, none of which the outline budget could see. With output bounded structurally, the outline work budget (HB_BUDGET_GLYPH) reverts to pure outline traversal. vector is one-glyph-per-context, so a flat cap suffices rather than a per-glyph-count derived budget.

khaledhosny (approved):

PR #6217, opened 2026-08-27
Add missing error handling for fallible operations in the repacker.

garretrieger:
Assisted by Gemini


behdad:
CI Fuzz is failing. Also, can all those -1's be named constants?

PR #6219, opened 2026-08-27
Add a result type that can be used for functions returning success or error values.

garretrieger:
Error handling in repacker is in rough shape. Many of the methods are fallible, but often the return values aren't being checked for error state.

To help address this I thought it might be useful to introduce a result type similar to Result<> in rust (or std::expected) that would force callers to deal with errors in the returns.

This PR introduces hb_result_t<OkVal, ErrVal>.

Assisted by: Gemini


src/hb-result.hh
```diff
@@ -0,0 +1,465 @@
...
+ * PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
+ *
+ * Google Author(s): Garret Rieger
+ */
+
+#ifndef HB_RESULT_HH
+#define HB_RESULT_HH
+
+#include "hb.hh"
+#include "hb-meta.hh"
```
behdad: `hb-meta.hh` is always included by `hb.hh`.

src/hb-result.hh
```diff
@@ -0,0 +1,465 @@
...
+
+#include <new>
```
behdad: As discussed, let's do without this header.

src/hb-result.hh
```diff
@@ -0,0 +1,465 @@
...
+#include <utility>
```
behdad: Also already included by `hb.hh`.

src/hb-result.hh
```diff
@@ -0,0 +1,465 @@
...
+
+// Generic error code useful for describing common error scenarios that
+// occur in harfbuzz code.
+enum error_code_t {
+  ALLOCATION_FAILURE,
+  INVARIANT_VIOLATED,
+  OFFSET_OVERFLOW,
+  LIMIT_EXCEEDED,
+};
```
behdad: I'm a bit ambivalent about this. At any rate, needs some namespacing at least.

src/hb-result.hh
```diff
@@ -0,0 +1,465 @@
...
+
+  template <hb_enable_if ((!hb_is_same (E, hb_result_err_t<E>)))>
+  bool operator != (E e) const
+  {
+    return is_ok_ || err_ != e;
+  }
+};
+
+template <typename U, typename T, typename E>
+static inline bool operator == (const hb_result_ok_t<U>& o, const hb_result_t<T, E>& r)
```
behdad, line 466: Typically we do these as `friend` functions inside the type itself.

src/hb-result.hh
```diff
@@ -0,0 +1,465 @@
...
+  template <typename F = E,
+            hb_enable_if ((std::is_constructible<E, F>::value))>
+  hb_result_t (hb_result_err_t<F> e) : is_ok_ (false)
+  {
+    new (std::addressof (err_)) E (std::move (e.error));
+  }
+
+  // Construct directly from error E (when T != E)
+  template <hb_enable_if ((!hb_is_same (T, E)))>
+  hb_result_t (E e) : is_ok_ (false)
```
behdad: What happens if they are convertible? Eg. both integers? Probably a compiler error, which is fine for now.

src/hb-result.hh
```diff
@@ -0,0 +1,465 @@
...
+    }
+    else
+    {
+      destroy ();
+      is_ok_ = o.is_ok_;
+      if (is_ok_)
+        new (std::addressof (val_)) T (o.val_);
+      else
+        new (std::addressof (err_)) E (o.err_);
+    }
```
behdad, line 145: Can this block be added as a `init(is_ok_)`?

src/hb-result.hh
```diff
@@ -0,0 +1,465 @@
...
+#include "hb-meta.hh"
+
+#include <new>
+#include <utility>
+
+// Generic error code useful for describing common error scenarios that
+// occur in harfbuzz code.
+enum error_code_t {
+  ALLOCATION_FAILURE,
+  INVARIANT_VIOLATED,
```
behdad: I'm a bit allergic to this value. We typically crash hard on invariant violations instead of hiding them in errors.

src/hb-result.hh
behdad: I suggest we do without a generic error code for the first cut.

behdad (approved):
Generally good to me. I lef

src/hb-result.hh
garretrieger: Sounds good, I'll add a graph specific one then (when I update graph to use result) and get rid of this one.

garretrieger: Removed

src/hb-result.hh
garretrieger: Removed

src/hb-result.hh
garretrieger: If the operator's are declared inside the class as friend functions then you need to also repeat the definitions inside of the void specialization. So probably better to leave them out here so they only need to be defined once.

src/hb-result.hh
garretrieger: Ah good catch, tried out `hb_result_t<int64_t, int32_t> r(1);` which actually compiles and happens to treat it as an error init. I'll see if I can fix to make that a compilation error.

behdad: `hb_is_convertible` might be useful.

src/hb-result.hh
garretrieger: Added init()

src/hb-result.hh
```diff
@@ -0,0 +1,483 @@
...
+  hb_result_t (hb_result_t&& o) noexcept (std::is_nothrow_move_constructible<T>::value &&
+                                     std::is_nothrow_move_constructible<E>::value)
+      : is_ok_ (o.is_ok_)
+  {
+    init (std::move (o));
+  }
+
+  hb_result_t& operator = (const hb_result_t& o)
+  {
+    if (this == &o) return *this;
```
behdad, line 132: This is a bit unusual to have here.

behdad, line 132: Oh nevermind. It's `operator =`. I thought it's `operator ==`.

src/hb-result.hh
```diff
@@ -0,0 +1,483 @@
...
+
+  void destroy ()
+  {
+    if (is_ok_)
+      val_.~T ();
+    else
+      err_.~E ();
+  }
+
+  void init(const hb_result_t& o) {
```
behdad, line 94: Shouldn't these inits just copy is_ok from `o` first? That sounds more natural.

src/hb-result.hh
```diff
@@ -0,0 +1,483 @@
...
+            hb_enable_if ((std::is_constructible<E, const F&>::value))>
+  hb_result_t (const hb_result_err_t<F>& e) : is_ok_ (false)
+  {
+    new (std::addressof (err_)) E (e.error);
+  }
+
+  // Construct directly from error E (when T and E are not convertible into each other)
+  template <typename F = E,
+            hb_enable_if ((hb_is_same (F, E) &&
+                           !(hb_is_convertible (T, E) && hb_is_convertible (E, T))))>
```
behdad: What happens if T is convertible to E but not vice versa?

src/hb-result.hh
```diff
@@ -0,0 +1,483 @@
...
+
+  template <typename U>
+  T value_or (U&& default_value) const &
+  {
+    if (is_ok_) return val_;
+    return std::forward<U> (default_value);
+  }
+
+  template <typename U>
+  T value_or (U&& default_value) &&
```
behdad, line 243: Maybe add a `value_or_default()` as well?

src/hb-result.hh
garretrieger: Made this || so convertibility in either direction disables direct construction.

behdad:
Bots failing. But LGTM after that.

PR #6225, opened 2026-08-31
[ft] Serialize access to the static FreeType library

### LINKED ISSUE #6224: hb-ft.cc may have some functions that are not thread-safe
When I use the drawtext module of ffmpeg (drawtext call harfbuzz, harfbuzz call freetype), it is found that some functions in hb-ft.cc are not thread-safe, which will lead to program crash.
Some functions of hb-ft.cc will call reference_ft_library to get the global static variable static_ft_library, such as hb_ft_font_set_funcs. static_ft_library is not protected by locks. When static_ft_library is used to call some thread-unsafe functions of freetype, the critical section protection will be lost.

Called thread-unsafe function:
FT_Done_Library and FT_Reference_Library maybe use static_ft_library.FT_Library.
FT_New_Face,FT_New_Memory_Face and PT_Done_Face maybe use static_ft_library.FT_Library.modules[FT_MODULE_FONT_DRIVER].faces_list.

behdad:
## Summary

- serialize access to HarfBuzz's shared static `FT_Library`
- lock creation and destruction of faces owned by that library without affecting caller-owned FreeType libraries
- add a multithreaded regression test covering blob-backed faces and `hb_ft_font_set_funcs()`

## Testing

- `meson test -C build` (272/272 passed)

Fixes #6224

### DIFF
--- a/src/hb-ft.cc
+++ b/src/hb-ft.cc
@@ -88,12 +88,15 @@
 
 using hb_ft_advance_cache_t = hb_cache_t<16, 24, 8, false>;
 
+static void _hb_ft_face_destroy_static (void *data);
+
 struct hb_ft_font_t
 {
   int load_flags;
   bool symbol; /* Whether selected cmap is symbol cmap. */
   bool unref; /* Whether to destroy ft_face when done. */
   bool transform; /* Whether to apply FT_Face's transform. */
+  bool static_library; /* Whether ft_face uses static_ft_library. */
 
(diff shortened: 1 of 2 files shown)


RaistlinZW:
I used my test program to test the new hb-ft.cc‎. The program doesn't crash any more. I tested it 10 times. Before that, the test program almost certainly crashed.

behdad:
Thanks.

Issue #29, opened 2014-03-06
Oriya not shaping correctly in Firefox


ammazzarella:
Hi,

Testing my typeface in both Firefox and Safari, I've discovered some problems in the shaping of the Reph feature. The window on top is Safari and it shapes it correctly, the window below is Firefox and it doesn't. In Indesign everything works fine. 

![screen shot 2014-03-06 at 12 34 11](https://f.cloud.github.com/assets/6872728/2345116/032939cc-a52f-11e3-9482-aec6e9b62dfd.png)

Any idea on why this happens?

Alessia

PR #6227, opened 2026-09-02
[ot-shape] Enable fallback mark positioning for unanchored marks in GPOS fonts

woodham1:
When shaping combining character sequences where a base glyph lacks an explicit OpenType GPOS anchor (such as geometric symbols, mathematical characters, or Kaomoji like U+25E1 LOWER HALF CIRCLE with U+0308 COMBINING DIAERESIS '◡̈'), HarfBuzz previously disabled fallback mark positioning entirely whenever the font possessed a GPOS table (c->plan->apply_gpos). Because standard text fonts (such as Roboto and Noto Sans Symbols) contain GPOS tables for alphabetic glyphs but omit mark-to-base anchors for symbols, this caused unanchored combining marks to be left unpositioned, rendering detached or displaced (e.g. '◡ ¨').

Refusing implicit heuristic mark attachment when an explicit anchor is undeclared is non-compliant with the Unicode Standard:
1. Conformance Requirement C7 (Unicode Standard, Section 3.6 Combination): "A system that claims to support combining marks shall be able to apply combining marks to any base character." The standard explicitly states that combining marks are valid with any base character (including symbols) and must not be prohibited or discarded merely because a font lacks pre-declared linguistic anchors.
2. Section 5.13 (Implementation Guidelines - Rendering Nonspacing Marks): Specifies that when font tables (such as OpenType GPOS) do not define an explicit attachment point for a given base-plus-mark combination, rendering implementations are expected to apply fallback positioning heuristics (such as centering the mark over the base character's advance or bounding box) rather than leaving the mark detached or spacing.
3. Unicode Standard Annex #29 (UAX #29, Section 3): A base character followed by nonspacing marks forms an atomic Extended Grapheme Cluster (Rule GB9) that must be rendered as an integrated unit.

This change enables fallback_mark_positioning whenever the script allows it, while checking buffer->pos[i].attach_type() to guarantee that marks already positioned by GPOS remain untouched. Any unattached marks that GPOS skipped now undergo fallback heuristic centering over the base glyph.

### DIFF
--- a/src/hb-ot-shape-fallback.cc
+++ b/src/hb-ot-shape-fallback.cc
@@ -30,6 +30,7 @@
 
 #include "hb-ot-shape-fallback.hh"
 #include "hb-kern.hh"
+#include "OT/Layout/GPOS/Common.hh"
 
 static unsigned int
 recategorize_combining_class (hb_codepoint_t u,
@@ -195,6 +196,9 @@ zero_mark_advances (hb_buffer_t *buffer,
   for (unsigned int i = start; i < end; i++)
     if (_hb_glyph_info_get_general_category (&info[i]) == HB_UNICODE_GENERAL_CATEGORY_NON_SPACING_MARK)
     {
+      if (buffer->pos[i].attach_type())
+	continue;
+
       if (adjust_offsets_when_zeroing)
       {
 	buffer->pos[i].x_offset = hb_saturate_sub (buffer->pos[i].x_offset,
@@ -368,6 +372,9 @@ position_around_base (const hb_ot_shape_plan_t *plan,
   for (unsigned int i = base + 1; i < end; i++)
     if (_hb_glyph_info_get_modified_combining_class (&info[i]))
     {
+      if (buffer->pos[i].attach_type())
+	continue;
+
       if (num_lig_components > 1) {
 	unsigned int this_lig_id = _hb_glyph_info_get_lig_id (&info[i]);
 	int this_lig_component = _hb_glyph_info_get_lig_comp (&info[i]) - 1;
--- a/src/hb-ot-shape.cc
+++ b/src/hb-ot-shape.cc
@@ -203,8 +203,10 @@ hb_ot_shape_planner_t::compile (hb_ot_shape_plan_t           &plan,
 #endif
 					      );
 
-  plan.fallback_mark_positioning = plan.adjust_mark_positioning_when_zeroing &&
-				   script_fallback_position;
+  /* Fallback mark positioning should run for any unattached marks (Unicode
+   * Section 3.6 / 5.13), even if GPOS is present. Marks attached by GPOS
+   * will be skipped during fallback positioning. */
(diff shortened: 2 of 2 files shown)


behdad:
This is something we have deliberately decided not to do. It fights against the font designers, who might intentionally leave marks unattached when they are positioned correctly. Unicode conformance does not apply here since Unicode doesn't specify rendering.

cc @khaledhosny @jfkthame

woodham1:
¯\_(ツ)_/¯ I was just trying to fix it because when my sister sends me smiley face on ios, it breaks compatibility with my android device and it renders improperly. But I can understand why you'd want to keep it that way.

behdad:
I'll wait for Jonathan / Khaled's input. What you're seeing sounds like might be an Android itemization problem.

behdad:
Care to share examples of text & misrendering? Thanks.

woodham1:
Android end: 

Iphone end:

khaledhosny:
> 2\. Section 5.13 (Implementation Guidelines - Rendering Nonspacing Marks): Specifies that when font tables (such as OpenType GPOS) do not define an explicit attachment point for a given base-plus-mark combination, rendering implementations are expected to apply fallback positioning heuristics (such as centering the mark over the base character's advance or bounding box) rather than leaving the mark detached or spacing.

I wad curious because this does not sound like something Unicode would say, so I checked and found no such text in this Unicode spec section. Unicode core spec is available as online HTML pages now, please provide a direct link to this text.

khaledhosny:
> This is something we have deliberately decided not to do. It fights against the font designers, who might intentionally leave marks unattached when they are positioned correctly. Unicode conformance does not apply here since Unicode doesn't specify rendering.
> 
> cc @khaledhosny @jfkthame

Indeed. Also with current fallback, a font can easily opt out by having any GPOS feature for the scriopt, but with the proposed change, every unanchored pair will get fallback position with no simple way to opt out. It is not our job to fix font limitations or second-guess font designers.

woodham1:
Yeah, I thought the agent was hallucinating too, and that chrome and ios were out of spec in their rendering, but when I was reading the way the inside out rule was being applied to the baseline glyph, I thought, yeah, that kind of makes sense. 

You're saying that the render is in spec, and that it's not your job; that makes sense, and it seems to align with what [behdad](https://github.com/behdad) mentioned before on how you don't want to fight against the font designers. I guess this is just a raw implementation decision. 

Fair enough, I just hope that this trend doesn't continue.

jfkthame:
Agree with Behdad & Khaled that this isn't our job.

I'm curious what the iPhone is doing here; I don't see `:)` getting auto-converted to a rotated form on my phone. Maybe this is new in the latest iOS update?

Personally, I'd view this behavior as a bug (though no doubt some consider it a feature) in the messaging app. If I enter `:)` in a text message, that's exactly what should be sent. If I wanted a "real" emoji rather than an ASCII-style emoticon, there's an emoji input method for that.

PR #6229, opened 2026-09-03
[unicode] Treat U+180F FVS4 as Default_Ignorable

dylanpulver:
U+180F MONGOLIAN FREE VARIATION SELECTOR FOUR (Unicode 14.0) has
Default_Ignorable_Code_Point=Yes, but `is_default_ignorable()` covered only
U+180B..180E, so FVS4 was never hidden and leaked a `.notdef` into the output.

8c65442069 ("[Unicode 14] Handle U+180F FVS4") added the U+180F case to
`_hb_glyph_info_set_unicode_props()` and to the Arabic shaper, but that branch
sits inside `if (unicode->is_default_ignorable (u))`, so it has been unreachable
since it was written. Widening the range makes it live. U+180B..180F are all
Default_Ignorable per `DerivedCoreProperties.txt`, so it stays contiguous.

On a font with no Mongolian coverage, every other default ignorable
(U+180B..180E, U+FE00, U+200B) shapes to `[gid1=0+1114]`. Only FVS4 leaked:

    $ hb-shape test/api/fonts/Roboto-Regular.abc.ttf -u U+0061,U+180F
    [gid1=0+1114|gid0=0@-908,0+0]   # before (gid0 = .notdef)
    [gid1=0+1114]                   # after

### Verified locally

    meson setup build -Dtests=enabled
    meson compile -C build
    meson test -C build

254/254 OK, 0 failures (macOS arm64, clang).

Added two cases to `mongolian-variation-selector.tests`, which covered FVS1,
FVS2, FVS3 and MVS but not FVS4: `U+180F`, plus `U+1810` MONGOLIAN DIGIT ZERO
to pin the top of the range. Restoring U+180B..180E fails in-house 6174/6176;
widening to U+180B..1810 fails it the same way.

Not verified locally: reduced-feature and MSVC configs.

### DIFF
--- a/src/hb-unicode.hh
+++ b/src/hb-unicode.hh
@@ -176,7 +176,7 @@ HB_UNICODE_FUNCS_IMPLEMENT_CALLBACKS_SIMPLE
 	case 0x03: return unlikely (ch == 0x034Fu);
 	case 0x06: return unlikely (ch == 0x061Cu);
 	case 0x17: return hb_in_range<hb_codepoint_t> (ch, 0x17B4u, 0x17B5u);
-	case 0x18: return hb_in_range<hb_codepoint_t> (ch, 0x180Bu, 0x180Eu);
+	case 0x18: return hb_in_range<hb_codepoint_t> (ch, 0x180Bu, 0x180Fu);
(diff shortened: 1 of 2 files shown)


behdad:
Thanks. cc @jfkthame

behdad:
Thanks!

PR #6231, opened 2026-09-04
OT: Fuse redundant per-glyph shaping passes

behdad:
## Summary

- initialize each glyph mask while computing Unicode properties
- rebuild the buffer digest while assigning GDEF glyph properties on the normal OpenType path
- preserve the existing post-morx digest rebuild and glyph-class ordering

The changes are separate commits so their performance can be measured independently. The position-buffer optimization was intentionally left out because generic HarfBuzz font callbacks do not initialize complete position records. The fast glyph-mapping path is already single-pass in HarfBuzz.

## Testing

- meson compile -C build
- meson test -C build test-buffer test-shape in-house aots --print-errorlogs
- 25 buffer, 4 shape, 800 AOTS, and 7,272 in-house subtests passed

### DIFF
--- a/src/hb-buffer.hh
+++ b/src/hb-buffer.hh
@@ -385,11 +385,6 @@ struct hb_buffer_t
   }
   /* Advance idx without copying to output. */
   void skip_glyph () { idx++; }
-  void reset_masks (hb_mask_t mask)
-  {
-    for (unsigned int j = 0; j < len; j++)
-      info[j].mask = mask;
-  }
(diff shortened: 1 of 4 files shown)


behdad:
Two ideas lifted from https://github.com/harfbuzz/harfrust/pull/463
Reduces two loops over the buffer. Shows 1/2% speedup in benchmarks.

dfrg (approved):
LGTM!

PR #6235, opened 2026-09-04
Convert repacker and graph code to utilize to hb_result_t.

garretrieger:
Converts all error handling in the repacker/graph code to use hb_result_t.

Subset benchmarks are neutral to this change.

Assisted by: Gemini


behdad:
LGTM when bots happy.

PR #6240, opened 2026-09-05
[subset] Emit a valid empty shared-tuples offset

behdad:
Partial instancing can compile away all shared tuples while retaining glyph
variation data. The gvar serializer currently emits a zero
`sharedTuplesOffset` in that case. Strict consumers reject the null offset and
discard the remaining glyph variations.

Point the offset at the end of the glyph-variation offsets array, where the
empty shared-tuples array resides. Apply this to both the instancing serializer
and the regular subsetting path.

Tests:

- `HB_FONT_FUNCS=fontations meson test -C build --no-rebuild test-subset-avar2`
- `meson test -C build --no-rebuild` (273 passed, 1 skipped)

### DIFF
--- a/src/hb-ot-var-gvar-table.hh
+++ b/src/hb-ot-var-gvar-table.hh
@@ -379,7 +379,8 @@ struct gvar_GVAR
     bool long_offset = glyph_var_data_size > 0x1FFFEu || force_long_offsets;
     out->flags = long_offset ? 1 : 0;
 
-    HBUINT8 *glyph_var_data_offsets = c->allocate_size<HBUINT8> ((long_offset ? 4 : 2) * (num_glyphs + 1), false);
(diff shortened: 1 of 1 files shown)


behdad:
The spec says:

> The shared tuples array provides a set of variation-space positions that can be referenced by variation data for any glyph. The shared tuples array follows the GlyphVariationData offsets array at the end of the 'gvar' header. This data is simply an array of tuple records, each representing a position in the font’s variation space.

https://learn.microsoft.com/en-us/typography/opentype/spec/gvar#shared-tuples-array

so one can argue that a zero `sharedTuplesOffset` is wrong even if `sharedTupleCount` is zero. read-fonts / Skrifa have such an interpretation.

cc @dfrg @cmyr

behdad:
@garretrieger the patch can probably be done in a cleaner way using the serializer. Feel free to rework. Thanks.

behdad:
cc @qxliu76

garretrieger (approved):

PR #6244, opened 2026-09-07
[limits] Restore rendering budget test runtime

behdad:
## Problem

The unified glyph-work budget raised the effective outline allowance for
VARC, raster, vector, and GPU rendering to `1 << 26`. Pathological fuzz
inputs consume that full allowance, increasing the full test-suite runtime
from under three seconds to about six seconds.

Separately, splitting raster outline and pixel budgets stopped charging
scanline edge work before rendering. Complex COLR masks could therefore do
expensive edge sweeps before their rendered-mask area was charged.

## Fix

- Restore the common outline budget to `1 << 24`.
- Keep the raster pixel-work floor independently at `1 << 26`.
- Precharge estimated scanline work against the pixel budget before
  rendering a raster mask.

No public API or ABI changes are involved.

## Results

- Before: 6.00-6.02s
- After: 2.82-2.83s across three consecutive full-suite runs
- Tests: 276 passed, 1 skipped

### DIFF
--- a/src/hb-limits.hh
+++ b/src/hb-limits.hh
@@ -186,7 +186,7 @@
 /* The common finite default for one top-level glyph rendering session.
  * Nested outline and paint work shares the same live counter. */
 #ifndef HB_BUDGET_GLYPH
-#define HB_BUDGET_GLYPH ((int64_t) 1 << 26)
+#define HB_BUDGET_GLYPH ((int64_t) 1 << 24)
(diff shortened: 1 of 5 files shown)


behdad:
So this whole new budget needs to be tested in the wild, but this is a good fix anyway.

khaledhosny (approved):

PR #6248, opened 2026-09-09
Reapply "[rust] Pin skrifa"

behdad:
This reverts commit 9279d1d1a16129d45f30c905231dbebf83c5f685.


behdad:
The CI fails seem unrelated.

